PR #3640, opened 2021-07-29
Implement dependency resolution through package registry

mattt:
This PR builds on top of https://github.com/apple/swift-package-manager/pull/3635 to implement dependency resolution through a registry service.

### Motivation:

Implement [SE-0292](https://github.com/apple/swift-evolution/blob/main/proposals/0292-package-registry-service.md).

### Modifications:

- [x] Add `PackageRegistry` library and test target containing the `RegistryManager` client implementation.
- [x] Add `RegistryPackageContainer` class to `PackageGraph` module
- [x] Update `Workspace` to provide `RegistryPackageContainer` if the package is declared with a registry identifier (**Currently WIP with hardcoded registry URL**)
- [x] Update `PubGrubDependencyResolver` to support registry packages

### Result:

With this change, Swift Package Manager will be able to resolve package dependencies through a configured registry service.

PR #3750, opened 2021-09-16
Additional refactoring of ManagedDependencies

mattt:
As suggested in https://github.com/apple/swift-package-manager/pull/3698#discussion_r709598802, this PR provides additional refactoring of `ManagedDependencies` by rebasing changes in #3640 against #3698.

### Motivation:

Improve convenience and correctness of `Workspace` APIs.

### Modifications:

- Converted `ManagedDependency` from class to structure and made `ManagedDependency.State` an indirect enumeration. It makes sense for this to be a value type and has the added benefit of automatic `Equatable` synthesis.
- Removed `ManagedDependency.packageIdentity` property. This was a helper for `self.packageRef.identity` and only used in a couple places.
- Consolidated dependency path determination into dedicated APIs. There are a few places where we determine the location of a repository checkout or edited package by appending to a location. To ensure consistent routing (i.e. [case normalization](https://github.com/apple/swift-package-manager/pull/3749)), these now go through a single API.
 - <del>Removed `ManagedDependency.subpath` property. The location of a managed dependency is determined by the workspace (except in the case of an unmanaged edited dependency, which stores its path separately), so we don't need to store it separately.</del> <ins>I wasn't able to get this working. To be considered for a future refactoring pass.</ins>

### DIFF
--- a/Sources/Workspace/ManagedDependency.swift
+++ b/Sources/Workspace/ManagedDependency.swift
@@ -19,9 +19,9 @@ extension Workspace {
     ///
     /// Each dependency will have a checkout containing the sources at a
     /// particular revision, and may have an associated version.
-    final public class ManagedDependency {
+    public struct ManagedDependency: Equatable {
         /// Represents the state of the managed dependency.
-        public enum State: Equatable {
+        public indirect enum State: Equatable {
             /// The dependency is a managed checkout.
             case checkout(CheckoutState)
 
@@ -34,19 +34,6 @@ extension Workspace {
 
             // The dependency is a local package.
             case local
-
-            public static func == (lhs: Workspace.ManagedDependency.State, rhs: Workspace.ManagedDependency.State) -> Bool {
-                switch (lhs, rhs) {
-                case (.local, .local):
-                    return true
-                case (.checkout(let lState), .checkout(let rState)):
-                    return lState == rState
-                case (.edited(let lBasedOn, let lUnmanagedPath), .edited(let rBasedOn, let rUnmanagedPath)):
-                    return lBasedOn?.packageRef == rBasedOn?.packageRef && lUnmanagedPath == rUnmanagedPath
-                default:
-                    return false
-                }
-            }
         }
 
         /// The package reference.
--- a/Sources/Workspace/ManagedDependency.swift
+++ b/Sources/Workspace/ManagedDependency.swift
@@ -57,10 +57,6 @@ extension Workspace {
         /// The checked out path of the dependency on disk, relative to the workspace checkouts path.
         public let subpath: RelativePath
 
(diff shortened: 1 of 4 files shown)


Sources/SPMTestSupport/MockWorkspace.swift
```diff
@@ -399,7 +399,7 @@ public final class MockWorkspace {
             workspace.state.artifacts.add(artifact)
         }
 
-        try workspace.state.saveState()
+        try workspace.state.save()
```
tomerd, line 402: <3

Sources/Workspace/ManagedDependency.swift
```diff
@@ -14,195 +14,160 @@ import SourceControl
...
     }
 }
 
-
 // MARK: - ManagedDependencies
 
-/// A collection of managed dependencies.
-public final class ManagedDependencies {
-    /// The dependencies keyed by the package URL.
-    private var dependencyMap: [String: ManagedDependency]
```
tomerd, line 153: @mattt this is where I got in trouble with my refactoring. to make incremental progress, I suggest we keep the map based on location right now (ie [String: ManagedDependency]) and then I will find a way to refactor / fix the manifest loading logic in Workspace so we can finally change this to [PackageIdentity: ManagedDependency]

wdyt?

mattt: @tomerd That works for me. To help keep our history comprehensible, could you patch #3698 to isolate the call site changes and force push? (Maybe you have a local branch that with unsquashed commits for adb61b3?)

Did #3749 resolved any of your trouble on manifest loading? We might be able to keep keying `dependencies` on `PackageIdentity` since it's equivalent to what we were doing before, comparing the `lowercased()` versions, like in `subscript(forNameOrIdentity:)`

tomerd:
@matt #3698 should be ready for you to test against, when you are happy with it I will merge it as the rest of our tests pass

Sources/Workspace/ManagedDependency.swift
tomerd: this is now done, I went for a bit of a different approach. #3749 helps in various other places, but he the case is more subtle with how manifests are loaded (long story)

mattt:
@tomerd Thanks! I just rebased against the latest changes in #3698. This is ready to go in as soon as CI is ✅.

mattt:
@swift-ci Please smoke test

mattt:
@swift-ci Please smoke test

mattt:
@swift-ci Please smoke test

mattt:
@swift-ci Please smoke test

tomerd (approved):

tomerd:
nice. thank you

PR #3723, opened 2021-09-07
refactor diagnostics

tomerd:
motivation: refactor the diagnostics frameworks to seperate between emitting diagnostics and collecting/handling them, this helps creating an API to allow emitting diagnostics from any point at the code without having to pass in a DiagnosticsEngine everywhere

changes:
* add newe ObservabilitySystem API to abstract features like diagnostics, metrics and tracing
* add new DiagnosticsEmitter API design to [only] emit diagnostics
* adjust callsites to always create DiagnosticsEngine using the ObservabilitySystem so that the system are bridged
* convert GraphLoading subsystem to use DiagnosticsEmitter instead of DiagnosticsEngine
* move the Diagnostic printing code closer the Diganostic handling code in SwiftTool and clean it up
* adjust all tests and test utilities to use ObservabilitySystem

note: the code includes several API to allow bridging between the DiagnosticsEngine using the ObservabilitySystem APIs so that we can transition in steps and control the scope of the cahnges. The APIs are marked as deprecated to help trasnition, but are fairly noisy until we make more progress.

PR #3751, opened 2021-09-17
add metrics API to the observability system 

tomerd:
motivation: ability to track duration of operations

changes:
* introduce scoped metrics API to the observability system, focused on timers
* added tests

### DIFF
--- a/Sources/Basics/Observability.swift
+++ b/Sources/Basics/Observability.swift
@@ -54,19 +54,23 @@ public class ObservabilitySystem {
             description: "top scope",
             parent: .none,
             metadata: .none,
-            diagnosticsHandler: factory.diagnosticsHandler
+            diagnosticsHandler: factory.diagnosticsHandler,
+            metricsHandler: factory.metricsHandler
         )
     }
 
-    private struct NOOPFactory: ObservabilityFactory, DiagnosticsHandler {
+    private struct NOOPFactory: ObservabilityFactory, DiagnosticsHandler, MetricsHandler {
         var diagnosticsHandler: DiagnosticsHandler  { self }
+        var metricsHandler: MetricsHandler { self }
 
         func handleDiagnostic(scope: ObservabilityScope, diagnostic: Diagnostic) {}
+        func handleTimer(scope: ObservabilityScope, label: String, duration: DispatchTimeInterval) {}
     }
 }
 
 public protocol ObservabilityFactory {
     var diagnosticsHandler: DiagnosticsHandler { get }
+    var metricsHandler: MetricsHandler { get }
 }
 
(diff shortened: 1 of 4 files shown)


tomerd:
follow up to #3723

Sources/Basics/Observability.swift
```diff
@@ -54,19 +54,23 @@ public class ObservabilitySystem {
...
         )
     }
 
-    private struct NOOPFactory: ObservabilityFactory, DiagnosticsHandler {
+    private struct NOOPFactory: ObservabilityFactory, DiagnosticsHandler, MetricsHandler {
         var diagnosticsHandler: DiagnosticsHandler  { self }
+        var metricsHandler: MetricsHandler { self }
 
         func handleDiagnostic(scope: ObservabilityScope, diagnostic: Diagnostic) {}
+        func handleTimer(scope: ObservabilityScope, label: String, duration: DispatchTimeInterval) {}
```
abertelrud, line 67: Adding the start time in addition to the duration (or, really, recording any two of: start time, duration, or end time) would let us plot a timeline of what happened and when, which would be really great.  Could the start time of he activity be added here?  I realize that this is not consistent across hosts, but on a single host as we're talking about here, just the wall clock time would be tremendously useful.

Sources/Basics/Observability.swift
```diff
@@ -134,6 +143,16 @@ public final class ObservabilityScope: DiagnosticsEmitterProtocol {
         self.diagnosticsHandler.handleDiagnostic(scope: self, diagnostic: diagnostic)
     }
 
+    // metrics
+
+    public func makeTimer(label: String) -> Timer {
```
abertelrud, line 148: I wonder if this would be a time to pass in a start time.  It's not always `Date.now` since the metrics might come from a subsystem, but that seems like the default.  And then when "ending" the recording (the way I'm thinking about this) it would also allow a end time (or duration) but default it to `Date.now`.

Sources/Basics/Observability.swift
```diff
@@ -307,6 +326,40 @@ public struct Diagnostic: CustomStringConvertible, Equatable {
     }
 }
 
+// MARK: - Metrics
+
+public protocol MetricsHandler {
+    func handleTimer(scope: ObservabilityScope, label: String, duration: DispatchTimeInterval)
+}
+
+public struct Timer: TimerProtocol {
```
abertelrud, line 335: I'm not sure this is a good name since `Timer` in the everyday world is usually more like an alarm clock or some kind of callback.  I don't have a good candidate name in mind though.

tomerd:
we will explore a different path here

PR #3753, opened 2021-09-17
Fix a unit test that is writing to the checkout of SwiftPM itself

abertelrud:
This was never a good thing to do, but now that we have v2 of Package.resolved, it is an actively harmful thing to do if the IDE or toolchain being used to build SwiftPM and run the tests is older (e.g. using the 5.5 toolchain but working on main).

What happens is that after the test is run, using the older toolchain on the checkout (such as running `swift build`) results in a "unsupported schema version 2" error for the Package.resolved.

There's not need at all to run this test against SwiftPM itself — a sample package created for the purpose in the usual temporary directory is all that's needed.

rdar://83222890

### DIFF
--- a/Tests/WorkspaceTests/WorkspaceTests.swift
+++ b/Tests/WorkspaceTests/WorkspaceTests.swift
@@ -3806,40 +3806,48 @@ final class WorkspaceTests: XCTestCase {
 
     // This verifies that the simplest possible loading APIs are available for package clients.
     func testSimpleAPI() throws {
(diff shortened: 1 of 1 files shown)


abertelrud:
@swift-ci please smoke test

tomerd (approved):

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

PR #3754, opened 2021-09-17
convert the PackageLoading module to the new diagnostics apis

tomerd:
motivation: adapt new diagnostics apis

changes: refactor the PackageLoading module to use new diagnostics api

### DIFF
--- a/Sources/PackageLoading/ModuleMapGenerator.swift
+++ b/Sources/PackageLoading/ModuleMapGenerator.swift
@@ -8,9 +8,10 @@
  See http://swift.org/CONTRIBUTORS.txt for Swift project authors
  */
 
-import TSCBasic
-import PackageModel
+import Basics
 import Foundation
+import PackageModel
+import TSCBasic
 
 /// Name of the module map file recognized by the Clang and Swift compilers.
 public let moduleMapFilename = "module.modulemap"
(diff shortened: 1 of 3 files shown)


tomerd:
@swift-ci please smoke test

Sources/PackageLoading/ModuleMapGenerator.swift
```diff
@@ -86,8 +87,14 @@ public struct ModuleMapGenerator {
     }
     
     /// Inspects the file system at the public-headers directory with which the module map generator was instantiated, and returns the type of module map that applies to that directory.  This function contains all of the heuristics that implement module map policy for package targets; other functions just use the results of this determination.
-    public func determineModuleMapType(diagnostics: DiagnosticsEngine) -> ModuleMapType {
+    public func determineModuleMapType(observabilityScope: ObservabilityScope) -> ModuleMapType {
         // The following rules are documented at https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md#creating-c-language-targets.  To avoid breaking existing packages, do not change the semantics here without making any change conditional on the tools version of the package that defines the target.
+
+        let diagnosticsEmitter = observabilityScope.makeDiagnosticsEmitter {
```
abertelrud, line 93: Is this part about creating a diagnostics emitter temporary until the old diagnostics engine has been removed, or will every client function create its own emitter in the future as well?

tomerd: in most cases, code would emit directly from the `ObservabilityScope`, however, there are cases where you may want to customize metadata further which you can do in one of two ways: either create a `DiagnosticsEmitter` with additional metadata , or specific the additional metadata on the `emit()` call itself. in this case we choose the former since the code emit in many places so its more efficient to hold a emitter with metadata set.

abertelrud: Makes sense.  It will be interesting to see how this can be simplified if we have task locals or similar.  It seems compelling to want to wrap a closure and add metadata to any diagnostics it emits without quite so much ceremony.

tomerd: agreed. when we migrate the code to 5.5/5.6 we would be able to take advantage of task locals

abertelrud: Nice!  Right, it wasn't so much a commentary on this PR as thinking ahead.  This looks good to me.

abertelrud (approved):

Sources/PackageLoading/ModuleMapGenerator.swift
tomerd: https://github.com/Kitura/Kitura/issues/1559

PR #3756, opened 2021-09-18
convert the build modules to the new diagnostics apis

tomerd:
motivation: adapt new diagnostics apis

changes:
* refactor the build  modules and relevant call sites to use new diagnostics api
* clean up assumptions about local file system as a default argument
* adjust call-sites and test

### DIFF
--- a/Sources/PackageLoading/Target+PkgConfig.swift
+++ b/Sources/PackageLoading/Target+PkgConfig.swift
@@ -56,7 +56,7 @@ public struct PkgConfigResult {
 }
 
 /// Get pkgConfig result for a system library target.
-public func pkgConfigArgs(for target: SystemLibraryTarget, diagnostics: DiagnosticsEngine, fileSystem: FileSystem = localFileSystem, brewPrefix: AbsolutePath? = nil) -> [PkgConfigResult] {
+public func pkgConfigArgs(for target: SystemLibraryTarget, brewPrefix: AbsolutePath? = .none, fileSystem: FileSystem, observabilityScope: ObservabilityScope) -> [PkgConfigResult] {
     // If there is no pkg config name defined, we're done.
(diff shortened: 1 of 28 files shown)


tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

abertelrud (approved):

tomerd:
rebased on https://github.com/apple/swift-package-manager/pull/3777

tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

tomerd:
@abertelrud @neonichu this is rebased on #3777 and ready to go

Sources/PackageLoading/Target+PkgConfig.swift
```diff
@@ -252,17 +253,26 @@ public func removeDefaultFlags(cFlags: [String], libs: [String]) throws -> ([Str
...
+        metadata.targetName = targetName
+        return metadata
+    }
+}
 
-    public init(pcFile: String, target: String) {
-        self.pcFile = pcFile
-        self.target = target
+extension ObservabilityMetadata {
+    public var pcFile: String? {
```
abertelrud, line 266: Nice.

abertelrud (approved):

Sources/PackageLoading/Target+PkgConfig.swift
tomerd: with the current architecture the clients would need to know what to look for in the metadata, and there is no tie in back to the scope. these extensions however are usually internal such that only if you import the right module you can access them. happy to do something better if you have an idea in mind.

neonichu (approved):

tomerd:
@swift-ci please smoke test

PR #3757, opened 2021-09-18
fix PkgConfigHint test on systems that do not have pkg-config installed

tomerd:
motivation: fix CI

changes: change assumption in test about single diagnostic emitted, as there may be more on systems that do not have pkg-config installed


tomerd:
@swift-ci please smoke test

abertelrud (approved):

PR #3761, opened 2021-09-23
make error message more specific when failing to create build system with LLBuild

tomerd:
motivation: clearer error message to help diagnose when LLBuild fails

changes: instead of throwing a generic error, throw a more detailed one


tomerd:
@swift-ci please smoke test

abertelrud (approved):

PR #3764, opened 2021-09-26
refactor location abstractions

tomerd:
motivation: reduce dependency on string location information, in support of registry based dependencies

changes:
* add location information to PackageReference::Kind
* seperare out sourceControl dependencies to local and remote, where remote carries AbsolutePath and rmeote carries URL
* adjust call sites
* adjust tests and test utilities

rdar://82954076


Sources/PackageModel/Manifest/PackageDependencyDescription.swift
```diff
@@ -22,17 +27,21 @@ public enum PackageDependency: Equatable {
...
             case exact(Version)
             case range(Range<Version>)
             case revision(String)
             case branch(String)
         }
+
+        public enum Location: Equatable, Encodable {
+            case local(AbsolutePath)
+            case remote(Foundation.URL)
+        }
```
tomerd, line 44: 👀  this is another key change

Sources/PackageModel/PackageReference.swift
```diff
@@ -17,55 +18,99 @@ import TSCUtility
...
-
-        /// A remote package.
-        case remote
+        case fileSystem(AbsolutePath)
+
+        /// A local source package.
+        case localSourceControl(AbsolutePath)
+
+        /// A remote source package.
+        case remoteSourceControl(Foundation.URL)
```
tomerd, line 32: 👀  this is the key change

Sources/SourceControl/Repository.swift
```diff
@@ -28,22 +43,36 @@ public struct RepositorySpecifier: Hashable, Codable {
...
+
+        public var description: String {
+            switch self {
+            case .path(let path):
+                return path.pathString
+            case .url(let url):
+                return url.absoluteString
+            }
+        }
+    }
```
tomerd, line 70: 👀  this new enum is another key change

tomerd:
@mattt this is the second PR designed to help with landing SE-0292, I think I prefer to take this one even before #3698

wdyt?

tomerd:
@abertelrud @neonichu @elsh @yim-lee this PR includes refactoring we discussed several times with regards to how location is managed, and that is required to land SE-0292. its actually a pretty small change in the model itself but it has huge impact on call sites, especially in tests. given the large number of the lines changed, please focus on the core model changes which are highlighted in comments

tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

Sources/PackageGraph/RepositoryPackageContainer.swift
```diff
@@ -395,39 +400,41 @@ public class RepositoryPackageContainerProvider: PackageContainerProvider {
         on queue: DispatchQueue,
         completion: @escaping (Result<PackageContainer, Swift.Error>) -> Void
     ) {
-        // If the container is local, just create and return a local package container.
-        if package.kind != .remote {
-            return queue.async {
-                let container = LocalPackageContainer(
+        if let repositorySpecifier = package.repositorySpecifier {
```
tomerd, line 403: @mattt here you would need to add a case for the registry, and probably we can rename RepositoryPackageContainerProvider to something more generic that deals with all PackageContainer types, perhaps rename to DefaultPackageContainerProvider

Sources/PackageLoading/ManifestJSONParser.swift
```diff
@@ -154,59 +154,84 @@ enum ManifestJSONParser {
...
+            let identity = try identityResolver.resolveIdentity(for: url)
+            return .remoteSourceControl(
+                identity: identity,
+                nameForTargetDependencyResolutionOnly: name,
+                url: url,
+                requirement: requirement,
+                productFilter: .everything
+            )
+        } else {
+            throw StringError("invalid location: \(location)")
```
tomerd, line 220: with the new model, we can do things here much more safely

Sources/PackageModel/PackageReference.swift
```diff
@@ -9,63 +9,108 @@
...
+    // FIXME: we should not need this
     public var name: String
 
-    /// The path of the package.
+    /// The location of the package.
     ///
     /// This could be a remote repository, local repository or local package.
-    public let location: String
+    // FIXME: we should not need this
+    //@available(*, deprecated)
```
tomerd, line 70: after landing this PR, there will be follow up to clean up call sites that still depend on location string. there are very few now, mostly for display so we are getting there

Sources/SourceControl/Repository.swift
```diff
@@ -8,16 +8,31 @@
...
 
+import Foundation
 import TSCBasic
 
 /// Specifies a repository address.
-public struct RepositorySpecifier: Hashable, Codable {
+public struct RepositorySpecifier: Hashable {
     /// The URL of the repository.
-    public let url: String
+    // FIXME: transition url to location
```
tomerd, line 17: this is for follow up PR

Sources/SourceControl/RepositoryManager.swift
```diff
@@ -550,9 +550,19 @@ extension RepositoryManager.RepositoryHandle {
         guard let status = Status(rawValue: repository.status) else {
             throw StringError("unknown status :\(repository.status)")
         }
+        // FIXME: encode the type
```
tomerd, line 553: this is for follow up PR, its not too bad

Sources/PackageModel/PackageIdentity.swift
```diff
@@ -42,8 +41,15 @@ public struct PackageIdentity: CustomStringConvertible {
...
     /// Creates a package identity from a URL.
     /// - Parameter url: The package's URL.
-    public init(url: String) { // TODO: Migrate to Foundation.URL
-        self.description = Self.provider.init(url).description
+    public init(url: Foundation.URL) {
+        self.init(urlString: url.absoluteString)
+    }
+
+    /// Creates a package identity from a URL.
+    /// - Parameter url: The package's URL.
```
yim-lee, line 49:
```suggestion
    /// - Parameter urlString: The package's URL.
```

Sources/PackageModel/PackageReference.swift
yim-lee: What's the difference between `fileSystem` and `localSourceControl`? Do we do anything to ensure `remoteSourceControl` is indeed remote (e.g., what if URL scheme is `file://`?), or it doesn't matter?

tomerd:
> What's the difference between fileSystem and localSourceControl? Do we do anything to ensure remoteSourceControl is indeed remote (e.g., what if URL scheme is file://?), or it doesn't matter?

this is exactly one of the goals of this PR, `fileSystem` is a local dependency that is not necessarily a git repo, while localSourceControl must be a git repo. with this distinction the downstream code is much safe. we do not validate the prefix for URLs put into remoteSourceControl, that is a good idea to add

tomerd:
@swift-ci please smoke test

tomerd:

Sources/PackageModel/PackageReference.swift
mattt: Is there an opportunity for us to reuse the `SourceControl.Location` enumeration here?

https://github.com/apple/swift-package-manager/blob/51fe9f947e18fe3daccf3c2aaf6391136ba2b801/Sources/PackageModel/Manifest/PackageDependencyDescription.swift#L41-L44

mattt:
> @mattt this is the second PR designed to help with landing SE-0292, I think I prefer to take this one even before #3698. wdyt?

If #3698 is ready to go, I think it'd be great to lock in our gains with that. But you're closer to the source material and, so I defer to you on the best order of operations.

Sources/PackageModel/PackageReference.swift
tomerd: yes, my plan is to further consolidate / cleanup after this initial but big PR lands.

neonichu (approved):

Sources/Commands/Describe.swift
```diff
@@ -109,7 +109,12 @@ fileprivate struct DescribedPackage: Encodable {
             case .fileSystem(let settings):
                 self = .fileSystem(identity: settings.identity, path: settings.path)
             case .sourceControl(let settings):
-                self = .sourceControl(identity: settings.identity, location: settings.location, requirement: settings.requirement)
+                switch settings.location {
+                case .local(let path):
+                    self = .sourceControl(identity: settings.identity, location: path.pathString, requirement: settings.requirement)
```
abertelrud, line 114: Should we do a follow-on to separate out the described location similarly?

Sources/PackageGraph/LocalPackageContainer.swift
```diff
@@ -30,27 +30,35 @@ public final class LocalPackageContainer: PackageContainer {
...
     private let manifest = ThreadSafeBox<Manifest>()
 
     private func loadManifest() throws -> Manifest {
         try manifest.memoize() {
+            let path: AbsolutePath
+            switch self.package.kind {
+            case .root(let _path), .fileSystem(let _path):
+                path = _path
+            default:
+                throw InternalError("invalid package type \(package.kind)")
```
abertelrud, line 46: Could we add the package identity to the message here (or is it perhaps already part of the metadata?).  That would make it much easier to try to figure out errors coming in from the field that have this message.

Sources/PackageGraph/LocalPackageContainer.swift
```diff
@@ -80,8 +88,14 @@ public final class LocalPackageContainer: PackageContainer {
...
         fileSystem: FileSystem = localFileSystem
-    ) {
-        assert(URL.scheme(package.location) == nil, "unexpected scheme \(URL.scheme(package.location)!) in \(package.location)")
+    ) throws {
+        //assert(URL.scheme(package.location) == nil, "unexpected scheme \(URL.scheme(package.location)!) in \(package.location)")
+        switch package.kind {
+        case .root, .fileSystem:
+            break
+        default:
+            throw InternalError("invalid package type \(package.kind)")
```
abertelrud, line 97: Same comment as the other message.  If it's already part of the context for whatever handler wraps this call, so it will show up as metadata for a diagnostic, then that's even better.

Sources/PackageModel/Manifest/PackageDependencyDescription.swift
```diff
@@ -22,17 +27,21 @@ public enum PackageDependency: Equatable {
...
         public enum Requirement: Equatable, Hashable {
             case exact(Version)
             case range(Range<Version>)
             case revision(String)
             case branch(String)
         }
+
+        public enum Location: Equatable {
+            case local(AbsolutePath)
+            case remote(Foundation.URL)
```
abertelrud, line 43: In the past I think we had some issues with Foundation representing `git@...` URLs, at least on Darwin.  I agree that using URL instead of String here makes more sense conceptually, but do we think there might be side effects?  We should of course report them if there, but this might be something to watch for.

Sources/PackageModel/Manifest/PackageDependencyDescription.swift
```diff
@@ -135,37 +140,73 @@ public enum PackageDependency: Equatable {
...
+            .init(
+                identity: identity,
+                nameForTargetDependencyResolutionOnly: nameForTargetDependencyResolutionOnly,
+                path: path,
+                productFilter: productFilter
+            )
+        )
+    }
+
+    public static func localSourceControl(identity: PackageIdentity,
```
abertelrud, line 153: This is for a Git repository stored at a `file://` path?

Sources/PackageModel/PackageReference.swift
abertelrud: Great, this will do a lot to clarify the meaning of the code.

abertelrud (approved):

Sources/Commands/Describe.swift
tomerd: yes we could, less critical at this point imo

Sources/PackageGraph/LocalPackageContainer.swift
tomerd: sure. this is an internal error so a programmer error and not user one, so less important IMO

Sources/PackageModel/Manifest/PackageDependencyDescription.swift
tomerd: this PR helps remove the logic that depends on checking if there is a scheme or not, and instead uses the type: `localSourceControl` vs `remoteSourceControl`, so ideally this should no longer be an issue regardless

abertelrud: I see.  Yes, if SwiftPM is no longer asking those same questions but mainly using the URL as a container of URL-formatter strings, then that might go better.

Sources/Commands/Describe.swift
abertelrud: Yes, absolutely.

Sources/PackageModel/Manifest/PackageDependencyDescription.swift
tomerd: yes this is the goal - make the concrete location opaque to logic, so it only gets used when we pass it to git, where we use absoluteString to write out the full value

is there a specific use case you recall that we need to add a test for to make sure this works correctly in that sense?

PR #3772, opened 2021-09-30
reduce duplicate code in test tool

tomerd:
motivation: code cleanup

changes: remove almost identitical code in the two variants of the test method, and replace with a shared implementation that takes an output redirection and either prints or collects the output

### DIFF
--- a/Sources/Commands/SwiftTestTool.swift
+++ b/Sources/Commands/SwiftTestTool.swift
@@ -8,18 +8,16 @@ See http://swift.org/LICENSE.txt for license information
 See http://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
 
-import class Foundation.ProcessInfo
-
 import ArgumentParser
 import Basics
-import TSCBasic
-import SPMBuildCore
 import Build
-import TSCUtility
+import class Foundation.ProcessInfo
 import PackageGraph
-import Workspace
-
+import SPMBuildCore
+import TSCBasic
 import func TSCLibc.exit
+import TSCUtility
+import Workspace
 
 private enum TestError: Swift.Error {
     case invalidListTestJSONData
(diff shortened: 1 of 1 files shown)


tomerd:
@swift-ci please smoke test

Sources/Commands/SwiftTestTool.swift
```diff
@@ -652,61 +641,48 @@ final class TestRunner {
...
+        var stderr: [UInt8] = []
+
+        func makeOutput() -> String {
+            return String(bytes: stdout + stderr, encoding: .utf8)?.spm_chuzzle() ?? ""
         }
-        return (success, output)
-    }
 
-    /// Executes and returns execution status. Prints test output on standard streams.
-    private func test(testAt testPath: AbsolutePath) -> Bool {
```
tomerd, line 686: this is the main change, there use to be two variants of test function, one which collects the output and one that prints to standard streams. this consolidates this into one function that always collects, and prints when needed

Sources/Commands/SwiftTestTool.swift
```diff
@@ -652,61 +641,48 @@ final class TestRunner {
...
-    /// Executes the tests without printing anything on standard streams.
-    ///
-    /// - Returns: A tuple with first bool member indicating if test execution returned code 0
-    ///            and second argument containing the output of the execution.
-    private func test(testAt testPath: AbsolutePath) -> (Bool, String) {
-        var output = ""
-        var success = false
-        do {
-            // FIXME: The environment will be constructed for every test when using the
-            // parallel test runner. We should do some kind of caching.
```
tomerd, line 664: took care of this FIXME while at it, now the env is constructed once and injected by the caller

neonichu (approved):

abertelrud (approved):

PR #3774, opened 2021-09-30
`swift build` isn't passing `-fmodules` to C files in targets that contain one or more C++ files

abertelrud:
This change considers each Clang invocation separately w.r.t. passing `-fmodules` or not.

### Motivation:

In mixed C and C++ targets, C files get treated as C++ even there are any C++ files at all in a Clang target.  This means that if modules are disabled due to a Clang target having one or more C++ files, they are disabled for the whole target, which isn't correct.

### Modifications:

This changes the decision about whether to enable modules to be on a per-Clang-invocation basis.  The same restrictions are before are kept, i.e. Clang modules are disabled on Windows and Android due to SDK issues and on Darwin for C++ files only.

### Result:

The language of each source file is considered individually when deciding whether to enable or disable Clang modules.

rdar://83703185


abertelrud:
@swift-ci please smoke test

abertelrud:
@neonichu Any concerns?  There's more work to be done here, but I think we'll want to revisit how these command args are generated to achieve that, and so this is a spot fix.

abertelrud:
@swift-ci Please test source compatibility

abertelrud:

abertelrud:

abertelrud:
@DougGregor @benlangmuir @neonichu WDYT?  This was the follow-on issue that @benlangmuir saw.

neonichu (approved):

benlangmuir (commented):
The `-fmodules` option used to come before the following, and now it comes after:
* `self.buildSettingsFlags()`
* `buildParameters.toolchain.extraCCFlags`
* `buildParameters.flags.cCompilerFlags`

If a user/package specified `-fno-modules` using unsafe package build settings, or else using `-Xcc` on the command-line, I think it would change the behaviour for .m files, since the last argument wins when there are contradictory flags.

I don't know if this is a real concern or not, so I'll leave this with you. Otherwise the change LGTM.

abertelrud:
Thanks @benlangmuir that's a good point.  That is probably something to fix, though it will mean a larger diff.  I'll adjust the PR.

PR #3775, opened 2021-10-01
`ProductFilter` type is not `Decodable` for the `everything` case

abertelrud:
This fixes a case of not being able decode a `ProductFilter` that had been encoded.

### Motivation:

`ProductFilter` in the SwiftPM project model is declared as `Codable` and it should be possible to decode what was encoded.

### Discussion:

The `ProductFilter.everything` is encoded as the `.none` case of a `[String]?`, which gets encoded by the `SingleValueEncodingContainer` as JSON `null`.  But then `SingleValueDecodingContainer` doesn't decode that `null` as the `.none` case of `[String]?`, instead throwing an error.

This change uses `decodeNil` to check for `null`, but it seems to me that there might be a bug in `SingleValue{Encoding,Decoding}Container` here, since it doesn't seem symmetrical.  I would expect consistent handling of `[String]?` for encoding and decoding.

So there is follow-up there on a possible `SingleValueDecodingContainer` issue.

## Modifications:

- use `decodeNil()` to check for the "no specific filter" case
- add a unit test

### Result:

`ProductFilter` is properly decodable.

rdar://83647022

### DIFF
--- a/Sources/PackageModel/Product.swift
+++ b/Sources/PackageModel/Product.swift
@@ -247,11 +247,10 @@ extension ProductFilter: Codable {
 
     public init(from decoder: Decoder) throws {
         let container = try decoder.singleValueContainer()
-        let optionalSet: Set<String>? = try container.decode([String]?.self).map { Set($0) }
-        if let set = optionalSet {
-            self = .specific(set)
-        } else {
+        if container.decodeNil() {
             self = .everything
+        } else {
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

tomerd (approved):

SDGGiesbrecht:
Ah, interesting. Good catch! 👍

PR #3776, opened 2021-10-01
[5.5] `ProductFilter` type is not `Decodable` for the `everything` case (#3775)

abertelrud:
This is the 5.5 cherry-pick of #3775.  See that PR for the details.


abertelrud:
@swift-ci please test

tomerd (approved):

PR #3778, opened 2021-10-02
fix workspace to notify delegate on any result of manifest loading

tomerd:
motivation: bug where delegate was only notified on success but not on error

changes:
* udpate logic to notify on both error and success cases
* added unit test


tomerd:
@swift-ci please smoke test

abertelrud:
Thanks for fixing this, this is great!

abertelrud (approved):

abertelrud:
Taking the liberty to merge this, since it looks good and tests pass, and I'm eager to start adopting it.

PR #3777, opened 2021-10-02
remove global observability scope

tomerd:
motivation: better fit with long running process likes IDEs

changes:
* remove the global bootstrap and scope
* add observabilityScope argument to functions that need to emit diagnostics
* adjust call sites and tests

PR #3752, opened 2021-09-17
 convert test tool to new diagnostics apis

tomerd:
motivation: adapt new diagnostics apis

changes: refactor SwiftTestTool to use new diagnostics api, including relevant scopes and timer measuring duration of tests

PR #3779, opened 2021-10-02
migrate SwiftTestTool to Observability APIs

tomerd:
motivation: adopt new Observability APIs

changes: replace DiagnosticsEngine with Observability APIs in SwiftTestTool

### DIFF
--- a/Sources/Commands/SwiftTestTool.swift
+++ b/Sources/Commands/SwiftTestTool.swift
@@ -204,7 +204,7 @@ public struct SwiftTestTool: SwiftCommand {
     
     public func run(_ swiftTool: SwiftTool) throws {
         // Validate commands arguments
-        try validateArguments(diagnostics: swiftTool.observabilityScope.makeDiagnosticsEngine())
+        try self.validateArguments(observabilityScope: swiftTool.observabilityScope)
 
         switch options.mode {
         case .listTests:
@@ -301,7 +301,7 @@ public struct SwiftTestTool: SwiftCommand {
                 toolchain: toolchain,
                 testEnv: testEnv,
                 outputStream: swiftTool.outputStream,
-                diagnostics: swiftTool.observabilityScope.makeDiagnosticsEngine()
+                observabilityScope: swiftTool.observabilityScope
             )
 
             // Finally, run the tests.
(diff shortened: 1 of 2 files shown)


tomerd:
needs #3777 first

replaces #3752

Examples/package-info/Sources/package-info/main.swift
```diff
@@ -13,22 +14,16 @@ let packagePath = AbsolutePath(#file).parentDirectory.parentDirectory.parentDire
 
 // There are several levels of information available.
 // Each takes longer to load than the level above it, but provides more detail.
-let diagnostics = DiagnosticsEngine(handlers: [{ print($0)}])
+
+let observability = ObservabilitySystem({ print("\($0): \($1)") })
```
abertelrud, line 18: Small suggestion: would it make sense to replace with named parameters now that there’s more than one?  I can guess that the first one is the location and the second is the message, but the first might also be severity.  Names would make it clearer in this case.

Examples/package-info/Sources/package-info/main.swift
```diff
@@ -13,22 +14,16 @@ let packagePath = AbsolutePath(#file).parentDirectory.parentDirectory.parentDire
...
-guard !diagnostics.hasErrors else {
-    fatalError("error loading package: \(diagnostics)")
-}
-
-let graph = try workspace.loadPackageGraph(rootPath: packagePath, diagnostics: diagnostics)
-guard !diagnostics.hasErrors else {
-    fatalError("error loading graph: \(diagnostics)")
-}
+
+let manifest = try tsc_await { workspace.loadRootManifest(at: packagePath, observabilityScope: observability.topScope, completion: $0) }
```
abertelrud, line 22: Do we lose the context here for where the diagnostic is being emitted from?

tomerd: not sure i follow, in what way do we loose context?

abertelrud: I just meant that the previous error messages said  "error loading manifest” etc.  It looks like that code has been removed though, so do they just say “error” now?  Or is the 'error loading manifest” what the `$0` is in the observability handler?

I was simply noticing the removal of a string containing context visible to the user.

tomerd: so this specific code is an example of how to use the workspace apis, and not part of SwiftPM itself

the example used to check for "hasError" on the diagnostics, for cases where a diagnostics was emitted, but that check is not required and actually an anti-pattern. if a critical error was found the code would exit on the throw, otherwise diagnostics will be printed via the handler configured at the top of the example

tomerd: @abertelrud note this specific change is part of https://github.com/apple/swift-package-manager/pull/3777 which this PR builds on top, so probably worth reviewing and merging https://github.com/apple/swift-package-manager/pull/3777 first, then I will rebase this PR on top which will make the diff much smaller

abertelrud: Thanks for pointing that out.  Yes, I noticed that after leaving this comment.  I’ll review that one first.

abertelrud (approved):

tomerd:
@swift-ci please smoke test

tomerd:
@abertelrud @neonichu this is rebased on #3777 and ready to go

PR #3784, opened 2021-10-04
Add SWIFT_MANIFEST_PATH environment variable

pcbeard:
Add SWIFT_MANIFEST_PATH environment variable

### Motivation:

This passes the path of the Package.swift to the manifest to allow it to examine the file system relative to itself, without relying on the current working directory.

### Modifications:

1 line change. Creates a new environment variable `"SWIFT_MANIFEST_PATH"` that is set to the path of the `Package.swift` that has just been compiled.

### Result:

Manifests will be able to use this path to compute locations of files relative to the manifest file. Sandboxed manifests already have access to read the file system, this just makes it easier for them to do so in a reliable way.


abertelrud:
This will work for the root package, but not for package dependencies sourced by semantic versions in Git repositories.  For versioned dependencies, the package manifest is extracted from the bare Git clone and evaluated without regard to the file system.  Once a specific commit is chosen, the files at that commit are actually checked out into the file system.  Doing a full checkout for each potential revision was quite expensive in the past, and involves a lot of file system churn, so I don't think we want to go back to that.

One of the problems here is that the package manifest does double duty in specifying the dependencies on other packages as well as defining how to build the package itself given that the dependencies exist.

It seems to me that SwiftPM should vend an API to package manifests through which they can get various information about the context in which the package is used.  Part of such an API could also be a FileSystem API (as SwiftPM has internally) that can be backed by either the real file system or by the Git repo, as needed.  SwiftPM already does that internally, so it would be a matter of wiring it up for the package manifest to use (and accounting for running in a separate process, etc).

On that note there should also be a better way for manifests to get contextual information rather than just the checking for env vars with magic names that it does today.  Some of the work being done for package plugins could provide a starting point here.

abertelrud:
I think it would make sense to say that for root packages, local packages, and any other case in which a local file system hierarchy exists when the manifest is evaluated, it makes sense to set the initial working directory to the package directory.  But really, the manifest should have access to SwiftPM's FileSystem API and it would be transparently backed by either the file system or a view onto the Git repo file system.

pcbeard:
@abertelrud I've updated the patch to specify the working directory of the manifest process to the package root. I've tested this on macOS for now. Will test on Linux and Windows later on.

pcbeard:
I see two ways forward:

1. Add prefix code to the manifest that changes the working directory of the manifest process to a value passed-in as an argument or as an environment variable. Something that might look like this:

```
import class Foundation.FileManager
import class Foundation.ProcessInfo
do {
    if let packageRoot = ProcessInfo.processInfo.environment["SWIFT_PACKAGE_ROOT"] {
        FileManager.default.changeCurrentDirectoryPath(packageRoot)
    }
}
```
2. Stick with the minimal solution of passing this information as simply the environment variable.

abertelrud:
Thanks @pcbeard!  I'm not stoked on the additional environment variable, but if you're prepending code to the manifest anyway, would it make sense to provide this as part of a `context` struct as I alluded to earlier?  For a long time now we've wanted to provide manifests with access to a struct of read-only information, which could include the package directory and an accessor for getting at contextual information.  The package directory is the first of those, and then we could extend that to include things like toolchain version and input parameters later.  It wouldn't necessary be a lot more work than this — I do think we'd want to define the struct in PackageDescription, which I could help with, and then it gets filled in here (or possibly through other means than prepending to the manifest source, e.g. initialized on first use similar to `targetBuildContext` for plugins).

It would be slightly less hacky and set a better future direction, I think, and not a lot more work.  Is that something you'd want to help turn this into?

abertelrud:
One disadvantage would be the introduction of a new variable, let's call it just `context`, which package authors would have to know about.  But if declared in `PackageDescription`, it would be properly documented and code-completable, etc, and not unlike the other symbols that are used in a package manifest.  A later larger change could then potentially adopt `@main` for package manifests, but I'm _not_ suggesting that here.

PR #3785, opened 2021-10-04
Serialize workspace state deterministically

randy-becker:
The `ManagedDependencies` and `ManagedArtifacts` collections use dictionaries as their underlying storage, so enumerating their elements is nondeterministic.

### Motivation:

In CI, we hash the contents of the workspace state, and use that as part of a cache key to avoid re-building unnecessarily, e.g. when pre-merge and post-merge sources are identical.  Thus, nondeterminism in serializing the workspace state contributed to extremely high cache invalidation rates.

### Modifications:

Sort artifacts and dependencies by package identity, which seems to be unique and Comparable.

### Result:

I ran `swift-package resolve` on SPM itself, and verified by inspection that the resulting dependencies in `.build/workspace-state.json` were sorted as expected.

### DIFF
--- a/Sources/Workspace/WorkspaceState.swift
+++ b/Sources/Workspace/WorkspaceState.swift
@@ -154,6 +154,11 @@ fileprivate struct WorkspaceStateStorage {
         struct Container: Codable {
             var dependencies: [Dependency]
             var artifacts: [Artifact]
+
+            init(dependencies: [Dependency], artifacts: [Artifact]) {
+                self.dependencies = dependencies.sorted { $0.packageRef.identity < $1.packageRef.identity }
+                self.artifacts = artifacts.sorted { $0.packageRef.identity < $1.packageRef.identity }
+            }
         }
 
         struct Dependency: Codable {
--- a/Sources/Workspace/WorkspaceState.swift
+++ b/Sources/Workspace/WorkspaceState.swift
@@ -146,19 +146,14 @@ fileprivate struct WorkspaceStateStorage {
         init (dependencies: Workspace.ManagedDependencies, artifacts: Workspace.ManagedArtifacts) {
             self.version = 4
             self.object = .init(
-                dependencies: dependencies.map { .init($0) },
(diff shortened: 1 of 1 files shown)


randy-becker:
@swift-ci please smoke test

neonichu (approved):

Sources/Workspace/WorkspaceState.swift
```diff
@@ -154,6 +154,11 @@ fileprivate struct WorkspaceStateStorage {
         struct Container: Codable {
             var dependencies: [Dependency]
             var artifacts: [Artifact]
+
+            init(dependencies: [Dependency], artifacts: [Artifact]) {
+                self.dependencies = dependencies.sorted { $0.packageRef.identity < $1.packageRef.identity }
```
abertelrud, line 159: Sorting these seems like a good idea, but since this constructor takes ordered inputs already, does it make sense to sort them here?  I would expect it to be done at wherever the Dictionary is getting iterated over (admittedly I don't know this code so it could be that I'm missing something there, but in general, for any type that takes ordered initialization parameters, I would expect order to be preserved).

randy-becker: Ah, good point.  I did it here because it clearly provides the stable serialization guarantee regardless of what the caller does, but in fact there's only one caller, a few lines up.  I'll update the PR accordingly.

I also considered changing the underlying Collection conformance extensions to provide stable ordering, but that seemed broader than what we need.

randy-becker:
@swift-ci please smoke test

tomerd (approved):

Sources/Workspace/WorkspaceState.swift
tomerd: I think doing this in the serialization is the right place, since the focus on consistency of the serialization for, not the external data mode 

but I would love to see a unit test :D

tomerd:
thanks @randy-becker, could you follow on with a PR to add a unit test please

PR #3787, opened 2021-10-04
Add PackageRegistryManager

mattt:
A partial tree of #3640 containing additive changes that don't affect how dependencies are downloaded or resolved. This PR itself shouldn't affect the behavior of Swift Package Manager, and is intended to serve as a backstop should we need to rollback any changes related to package resolution. 

### Motivation:

Improve velocity of registry implementation and reduce risk of merging future changes. 

### Modifications:

- Add `RegistryManager` and associated tests
- Add `SourceArchiver` used by `RegistryManager`
- Extract `Registry` from `RegistryConfiguration`
- Add `scopeAndName` property to `PackageIdentity`
- Raise error when dependency is declared using invalid package identifier
- Rename `identity` to `id` in `PackageDescription` to match SE-0292

### Result:

No changes.


mattt:
@swift-ci Please smoke test

Sources/PackageRegistry/RegistryManager.swift
```diff
@@ -0,0 +1,296 @@
...
+                "Accept": "application/vnd.swift.registry.v1+zip"
+            ]
+        )
+
+        request.options.authorizationProvider = authorizationProvider
+
+        client.execute(request, progress: nil) { result in
+            switch result {
+            case .success(let response):
+                if response.statusCode == 200,
```
tomerd, line 229: iirc one can configure the request with acceptable response codes to avoid such check, tho you do make other checks here as well

Sources/PackageRegistry/RegistryManager.swift
```diff
@@ -0,0 +1,296 @@
...
+                    completion(.failure(RegistryError.invalidResponse))
+                }
+            case .failure(let error):
+                completion(.failure(error))
+            }
+        }
+    }
+}
+
+private extension String {
```
tomerd, line 274: move to Basics?

Sources/PackageRegistry/SourceArchiver.swift
```diff
@@ -0,0 +1,72 @@
...
+            return
+        }
+
+        guard fileSystem.isDirectory(destinationPath) else {
+            completion(.failure(FileSystemError(.notDirectory, destinationPath)))
+            return
+        }
+
+        // TODO: consider calling `libarchive` or some other library directly instead of spawning a process
+        DispatchQueue.global(qos: .userInitiated).async {
```
tomerd, line 55: use .sharedConcurrent as queue

Sources/PackageRegistry/SourceArchiver.swift
```diff
@@ -0,0 +1,72 @@
...
+    public var supportedExtensions: Set<String> { ["zip"] }
+
+    /// The file-system implementation used for various file-system operations and checks.
+    private let fileSystem: FileSystem
+
+    /// Creates a `SourceArchiver`.
+    ///
+    /// - Parameters:
+    ///   - fileSystem: The file-system to used by the `SourceArchiver`.
+    public init(fileSystem: FileSystem = localFileSystem) {
```
tomerd, line 35: trying to avoid defaulting to localFileSystem where we can

Sources/PackageRegistry/SourceArchiver.swift
```diff
@@ -0,0 +1,72 @@
...
+        self.fileSystem = fileSystem
+    }
+
+    public func extract(
```
tomerd, line 39: I think there is some code in workspace that also does similar things (for binary artifacts). maybe we can reuse / converge on one implementation?

Sources/Workspace/Workspace.swift
```diff
@@ -296,11 +305,17 @@ public class Workspace {
             delegate: delegate.map(WorkspaceRepositoryManagerDelegate.init(workspaceDelegate:)),
             cachePath: sharedRepositoriesCacheEnabled ? location.sharedRepositoriesCacheDirectory : .none
         )
+        let registryManager = customRegistryManager ?? registries.map { configuration in
+            RegistryManager(configuration: configuration,
+                            identityResolver: identityResolver,
+                            authorizationProvider: authorizationProvider?.httpAuthorizationHeader(for:),
+                            diagnosticEngine: nil)
```
tomerd, line 312: its unlikely to have a diagnosticEngine at this point since workspace is long lived. probably worth removing from the ctor and moving the public methods that need it

mattt: Thanks for calling this out. I have a note to myself to move `diagnosticEngine` to parameters in the instance methods themselves. Working on that now.

mattt: Fixed by 9d1c129b

Sources/PackageRegistry/SourceArchiver.swift
mattt: Fixed by b1a641f5

Sources/PackageRegistry/SourceArchiver.swift
mattt: Fixed by 05c639a9

Sources/PackageRegistry/SourceArchiver.swift
mattt: There's `ZipArchiver` in TSCUtility, which I based this type on: https://github.com/apple/swift-tools-support-core/blob/f9bbd6b80d67408021576adf6247e17c2e957d92/Sources/TSCUtility/Archiver.swift#L34

(Hence some of the callouts from copy-pasted code, like dispatching to the global queue)

I discuss the rationale for this type in the doc comments, but the main takeaway is that this is probably only useful for the registry. I'd also be interested in getting some more 👀 on this, especially from security folks.  

As far as converging on a single implementation, there may be an opportunity to make the delegated system call and directory prefix stripping a configurable property. But at that point, I think you may as well have two separate implementations of the `Archiver` protocol.

Sources/PackageRegistry/RegistryManager.swift
mattt: Yeah, if it were just the status check or a range of status codes, I would've gone with that. But it's less code overall to do it this way here.

Sources/Commands/SwiftTool.swift
```diff
@@ -571,6 +586,7 @@ public class SwiftTool {
                 sharedConfigurationDirectory: sharedConfigurationDirectory
             ),
             mirrors: self.getMirrorsConfig(sharedConfigurationDirectory: sharedConfigurationDirectory).mirrors,
+            registries: try? self.getRegistriesConfig(sharedConfigurationDirectory: sharedConfigurationDirectory).configuration,
```
tomerd, line 589: `try?` intentional? should it throw instead?

Sources/PackageDescription/PackageDependency.swift
```diff
@@ -537,16 +539,21 @@ extension Package.Dependency {
...
 
     // intentionally private to hide enum detail
     @available(_PackageDescription, introduced: 999)
     private static func package(
-        identity: String,
+        id: String,
         requirement: Package.Dependency.RegistryRequirement
     ) -> Package.Dependency {
-        return .init(identity: identity, requirement: requirement)
+        let pattern = #"\A[a-zA-Z\d](?:[a-zA-Z\d]|-(?=[a-zA-Z\d])){0,38}\.[a-zA-Z0-9](?:[a-zA-Z0-9]|[-_](?=[a-zA-Z0-9])){0,99}\z"#
```
tomerd, line 551: @yim-lee 👀

Sources/PackageModel/PackageIdentity.swift
```diff
@@ -149,6 +160,11 @@ extension PackageIdentity {
             self = scope
         }
 
+        fileprivate init?(_ substring: String.SubSequence?) {
+            guard let substring = substring else { return nil }
+            self.init(String(substring))
+        }
```
tomerd, line 166: `let identity = sequence.map{ String($0) }` at the call site too verbose?

Sources/PackageRegistry/CMakeLists.txt
```diff
@@ -7,12 +7,16 @@
...
+  RegistryConfiguration.swift
+  RegistryManager.swift
+  SourceArchiver.swift)
 target_link_libraries(PackageRegistry PUBLIC
   TSCBasic
   PackageLoading
   PackageModel
-  TSCUtility)
+  TSCUtility
+  Basics)
```
tomerd, line 19: we try to keep them in alphabetical oder

tomerd:
lgtm, couple of small things

Sources/PackageDescription/PackageDependency.swift
yim-lee: 👍

Sources/PackageModel/PackageIdentity.swift
mattt: The alternative to this is:

```swift
guard components.count == 2,
              let scope = components.first.flatMap({ Scope(String($0)) }),
              let name = components.last.flatMap({ Name(String($0)) })
        else { return nil }
```

This seemed to me less clear, but really it's a subjective call. Totally up to you which convention you'd prefer to have in the code base.

Sources/Commands/SwiftTool.swift
mattt: If we throw here, package resolution would fail for anyone who doesn't have a `.swiftpm/configuration/registries.json` file, even if they only use URL-based dependencies. The check for valid registry configuration is pushed back to be done only when necessary.

mattt:
@swift-ci Please smoke test

Sources/PackageRegistry/SourceArchiver.swift
tomerd: cc @andyp-apple and @robertlacroix

mattt:
@swift-ci Please smoke test

mattt:
@swift-ci Please smoke test

mattt:
@swift-ci Please smoke test

tomerd (approved):

PR #3788, opened 2021-10-04
Add a test for WorkspaceState serialization stability

randy-becker:
Check that serialized workspace state indeed has stable ordering for dependencies.

### Motivation:

Follow-up from https://github.com/apple/swift-package-manager/pull/3785.

### Modifications:

Adds a new test case for `WorkspaceState` that loads a mis-sorted state file, then re-saves it.

### Result:

Because the underlying issue is non-deterministic, this test has a false-negative rate of roughly 50%, i.e. it will pass half the time, even when it should fail.  Repeating the test can improve those odds, but I wan't able to do so reliably from within the test itself vs. multiple `swift test` invocations.

### DIFF
--- a/Tests/WorkspaceTests/WorkspaceStateTests.swift
+++ b/Tests/WorkspaceTests/WorkspaceStateTests.swift
@@ -13,17 +13,12 @@ import TSCBasic
 import XCTest
 
 final class WorkspaceStateTests: XCTestCase {
-    private var fs: FileSystem! = nil
-
-    override func setUp() {
-        super.setUp()
-
-        fs = InMemoryFileSystem()
-    }
-
     func testSavedDependenciesAreSorted() throws {
+        let fs = InMemoryFileSystem()
+
         let buildDir = AbsolutePath("/.build")
         let statePath = buildDir.appending(component: "workspace-state.json")
+
         try fs.writeFileContents(statePath) {
             $0 <<<
                 """
--- a/Tests/WorkspaceTests/WorkspaceStateTests.swift
+++ b/Tests/WorkspaceTests/WorkspaceStateTests.swift
@@ -1,0 +1,85 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2021 Apple Inc. and the Swift project authors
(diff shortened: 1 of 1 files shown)


randy-becker:
@swift-ci please smoke test

tomerd (approved):

Tests/WorkspaceTests/WorkspaceStateTests.swift
```diff
@@ -0,0 +1,85 @@
...
+@testable import Workspace
+import XCTest
+
+final class WorkspaceStateTests: XCTestCase {
+    private var fs: FileSystem! = nil
+
+    override func setUp() {
+        super.setUp()
+
+        fs = InMemoryFileSystem()
```
tomerd, line 21: probably do this at a test by test basis, as we dont always use the in-memory one

randy-becker:
@swift-ci please smoke test

Tests/WorkspaceTests/WorkspaceStateTests.swift
```diff
@@ -0,0 +1,80 @@
...
+
+        let state = WorkspaceState(dataPath: buildDir, fileSystem: fs)
+        try state.save()
+
+        let serialized = try fs.readFileContents(statePath).description
+
+        let argpRange = try XCTUnwrap(serialized.range(of: "swift-argument-parser"))
+        let yamsRange = try XCTUnwrap(serialized.range(of: "yams"))
+
+        XCTAssertTrue(argpRange.lowerBound < yamsRange.lowerBound)
```
abertelrud, line 78: IIRC there is hash randomization that potentially causes the hash order to be different for each run, so in theory this should pass 50% of the time or fail 50% of the time even without sorting, which is better than accidentally passing all the time.  A potentially more consistent way of doing it might be programmatically generating the input with, say, 100 unique input strings, and then checking that the sorted set of unique appear in increasing order.  I don't think it has to be done for this PR since we'll still get a signal from sporadic failures, just a thought for a possible improvement in the future.

abertelrud (approved):

PR #3790, opened 2021-10-05
add support for local netrc file

tomerd:
motivation: SE-0292 defines that local .netrc files should be honored with priority over user home directory .netrc files

changes:
* add lookup for local .netrc file before defaulting to user home directory
* deprecate Workspace.Configuration.Netrc since it now redundant
* add tests

### DIFF
--- a/Sources/Basics/FileSystem+Extensions.swift
+++ b/Sources/Basics/FileSystem+Extensions.swift
@@ -129,4 +129,8 @@ extension FileSystem {
     public func writeFileContents(_ path: AbsolutePath, string: String) throws {
         return try self.writeFileContents(path, bytes: .init(encodingAsUTF8: string))
     }
+
+    public func writeFileContents(_ path: AbsolutePath, provider: () -> String) throws {
+        return try self.writeFileContents(path, string: provider())
+    }
 }
--- a/Sources/Commands/SwiftTool.swift
+++ b/Sources/Commands/SwiftTool.swift
@@ -498,8 +498,8 @@ public class SwiftTool {
     func getAuthorizationProvider() throws -> AuthorizationProvider? {
         var providers = [AuthorizationProvider]()
         // netrc file has higher specificity than keychain so use it first
-        if let workspaceNetrc = try self.getNetrcConfig()?.get() {
-            providers.append(workspaceNetrc)
+        if let netrcConfigFile = try self.getNetrcConfigFile() {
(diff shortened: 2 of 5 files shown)


tomerd:
@swift-ci please smoke test

Sources/Commands/SwiftTool.swift
```diff
@@ -498,34 +498,38 @@ public class SwiftTool {
...
+        }
 
-            return .init(path: configuredPath, fileSystem: localFileSystem)
-        } else {
-            let defaultPath = localFileSystem.homeDirectory.appending(component: ".netrc")
-            guard localFileSystem.exists(defaultPath) else { return nil }
-
-            return .init(path: defaultPath, fileSystem: localFileSystem)
+        // TODO: replace multiroot-data-file with explicit overrides
+        let localPath = try (options.multirootPackageDataFile ?? self.getPackageRoot()).appending(component: ".netrc")
```
tomerd, line 526: @mattt not sure what location you had in mind for the local case. for this PR I assumed the package's root directory, but we can consider other locations like `<root>/.swiftpm/configuration/.netrc`

Sources/Basics/FileSystem+Extensions.swift
```diff
@@ -129,4 +129,8 @@ extension FileSystem {
     public func writeFileContents(_ path: AbsolutePath, string: String) throws {
         return try self.writeFileContents(path, bytes: .init(encodingAsUTF8: string))
     }
+
+    public func writeFileContents(_ path: AbsolutePath, provider: () -> String) throws {
```
mattt, line 133: Could we consolidate this into the existing API by changing the original signature to use an `@autoclosure`?

Sources/Commands/SwiftTool.swift
mattt: That's perfect. I think `<root>/.netrc` is the right place to look for a local override.

As for `.swiftpm/configuration`, I think that makes more sense for configuration that's specific to Swift Package Manager. Since `.netrc` files are a de facto standard, I wouldn't expect to see them there.

Tests/CommandsTests/SwiftToolTests.swift
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+import Basics
+@testable import Commands
+import SPMTestSupport
+import TSCBasic
+import XCTest
+
+final class SwiftToolTests: XCTestCase {
+    func testNetrcLocations() throws {
```
mattt, line 18: 🤩

mattt (approved):

PR #3795, opened 2021-10-07
Add accept header to download request

jimmya:
Add `Accept: application/octet-stream` header to binary download request to allow use of Github API.

### Motivation:

I want to be able to download Github release assets as binary dependencies. However the [Github API states](https://docs.github.com/en/rest/reference/repos#get-a-release-asset) that in order to download an asset a `Accept: application/octet-stream` header should be present, otherwise the API JSON response is returned.

### Modifications:

When downloading a binary artifact an `Accept: application/octet-stream` header is added to the request.

### Result:

On each download download request of an artifact the `Accept: application/octet-stream` header is added.

### DIFF
--- a/Sources/Workspace/Workspace.swift
+++ b/Sources/Workspace/Workspace.swift
@@ -2025,7 +2025,9 @@ extension Workspace {
             let archivePath = parentDirectory.appending(component: artifact.url.lastPathComponent)
 
             group.enter()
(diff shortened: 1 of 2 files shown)


tomerd:
@swift-ci please smoke test

tomerd:
thanks @jimmya

PR #3798, opened 2021-10-08
protect from concurrency issues when extracting binary target archives

tomerd:
motivation: protect from edge cases when binary artifact archives are downloaded and extracted twice at the same time

changes:
* make the temp extraction directory name more unique
* move the temp extraction directory creation closer to extraction, to minimize the window of another process interrupting
* add file lock on the binary artifact
* add test to make sure binary artifacts are not duplicated due to transitive depednecies (they are not)

rdar://83516383


tomerd:
@swift-ci please smoke test

abertelrud (approved):

tomerd:
@swift-ci please smoke test

PR #3800, opened 2021-10-09
determin the version Package.resolved based on the minim tools versio…

tomerd:
…n of the root manifests

motivation: the introduction of v.2 of package reslved is causing confusion when using newer SwiftPM on older packages

changes:
* change the resolve file storage to take tools version and only write v.2 when using tools-version >= 5.6
* compute the minimum tools version for the root manifests and pass it to the pin store when saving
* add and adjust tests

rdar://83960284

### DIFF
--- a/Sources/PackageGraph/PinsStore.swift
+++ b/Sources/PackageGraph/PinsStore.swift
@@ -92,8 +92,13 @@ public final class PinsStore {
         self._pins.clear()
     }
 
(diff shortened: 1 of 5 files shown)


tomerd:
@swift-ci please smoke test

abertelrud (approved):

neonichu (approved):

PR #3801, opened 2021-10-11
Fix bug: https://bugs.swift.org/browse/SR-15304

DimaRU:
### Motivation:

When parsing the pkg-config file, the `-framework` flag from the `libs:` section is handled incorrectly, which leads to an error at the link stage of the package build. The `-framework` flag must have an associated value which is mistakenly filtered out.

Detail explanation in [https://bugs.swift.org/browse/SR-15304](https://bugs.swift.org/browse/SR-15304)

### Modifications:

Change "-framework" flag processing in `func allowlist`

### Result:

Correct pkg-config file parsing.


tomerd:
@swift-ci please smoke test

PR #3802, opened 2021-10-12
[Collections] Update certificate config and improve logging

yim-lee:
- Apple package collection is signed using Swift Package Collection cert
  only, hence removing Distribution cert from the config
- Add more info to logging so we know which cert policy leads to
  specific warning/error.


yim-lee:
@swift-ci please smoke test

PR #3804, opened 2021-10-13
Stop renaming main symbol for WASI

kateinoigakukun:
Stop renaming main symbol for WASI due to lack of support in the linker

### Motivation:

The latest SwiftPM rename main entry point name of executable target to avoid conflicting "main" with test target since swift-tools-version >= 5.5.
The main symbol is renamed to "{{module_name}}_main" and it's renamed again to be "main" when linking the executable target. The former renaming is done by Swift compiler, and the latter is done by linker, so SwiftPM passes some special linker flags for each platform.
But SwiftPM assumed that wasm-ld supports it by returning an empty array instead of nil even though wasm-ld doesn't support it yet.

### Modifications:

This patch changed `linkerFlagsForRenamingMainFunction` to return nil when targeting WASI to indicate it's not supported.

However, for a long-term solution, this renaming feature should be supported on wasm-ld because it's not a limitation of WebAssembly.

See also: https://github.com/swiftwasm/swift/issues/3645
CC: @MaxDesiatov

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -748,7 +748,7 @@ public final class SwiftTargetBuildDescription {
             // we can rename the symbol unconditionally.
             // No `-` for these flags because the set of Strings in driver.supportedFrontendFlags do
             // not have a leading `-`
-            if buildParameters.shouldRenameEntrypointFunctionName,
+            if buildParameters.canRenameEntrypointFunctionName,
                buildParameters.linkerFlagsForRenamingMainFunction(of: target) != nil {
                 args += ["-Xfrontend", "-entry-point-function-name", "-Xfrontend", "\(target.c99name)_main"]
             }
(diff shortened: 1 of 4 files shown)


abertelrud:
Thank you for this fix!  I'm actually wondering if it's ever correct for this function to return an empty array.  This looks like a logic error from the start.  Should it perhaps always return nil if the linker doesn't support it, not just for WASI?

abertelrud (approved):

kateinoigakukun:
@swift-ci please smoke test

tomerd:
thanks @kateinoigakukun

abertelrud (approved):

tomerd:
> Having a separate boolean makes sense to me, would be interested in hearing what @neonichu and @tomerd think as well.

same: +1

MaxDesiatov:
@swift-ci please smoke test

abertelrud (approved):
Thanks, this looks good to me!

PR #3806, opened 2021-10-13
fix flaky tests that assume order on DispatchQueue.async

tomerd:
motivation: stable tests

changes:
* update ManifestLoader to call willParse delegate call on the same queue as wilLoad, for correctnessa dn consistency
* update manifest loading tests to wait on delegate calls since they may come out of order: ManifestLoaded uses the same queue to call the delegates as it does for calling the delegate.

rdar://79415639
rdar://84181195


tomerd:
@swift-ci please smoke test

neonichu (approved):

abertelrud (approved):

PR #3807, opened 2021-10-13
Add release notes for SwiftPM 5.5

friedbunny:
Add release notes for SwiftPM 5.5. This is an initial draft and is almost certainly incomplete, so it needs a bit of help from everyone. 🙇

### DIFF
--- a/Documentation/ReleaseNotes/5.5.md
+++ b/Documentation/ReleaseNotes/5.5.md
@@ -1,0 +1,25 @@
+# SwiftPM 5.5 Release Notes
+
+### Package Collections
+
+You can now use package collections to discover, manage, and publish curated lists of packages. Package collections are authored as static JSON documents and can be published to the web or distributed to local file systems. SwiftPM and Xcode locally share the same package collections, so either interface can manage your added collections.
+
+To try the official Apple Swift Packages collection, run:
+
+```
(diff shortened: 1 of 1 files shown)


Documentation/ReleaseNotes/5.5.md
```diff
@@ -0,0 +1,25 @@
+# SwiftPM 5.5 Release Notes
+
+### Package Collections
```
yim-lee, line 3: 👍

abertelrud:
Thanks @friedbunny for starting this document.  According to `CHANGELOG.md`, both of these were also in 5.5:
```
* [#3410]

  In a package that specifies a minimum tools version of 5.5, `@main` can now be used in a single-source file executable as long as the name of the source file isn't `main.swift`.  To work around special compiler semantics with single-file modules, SwiftPM now passes `-parse-as-library` when compiling an executable module that contains a single Swift source file whose name is not `main.swift`.
```

```
* [#3316]

  Test targets can now link against executable targets as if they were libraries, so that they can test any data strutures or algorithms in them.  All the code in the executable except for the main entry point itself is available to the unit test.  Separate executables are still linked, and can be tested as a subprocess in the same way as before.  This feature is available to tests defined in packages that have a tools version of `5.5` or newer. 
```

abertelrud (approved):

PR #3810, opened 2021-10-19
fix pacakge describe to handle binary targets

tomerd:
motivation: package describe command fails for packages with binary targets

changes:
* PackageBuilder requires remote artifacts to be set to compute the model, as such make it a required argument
* do not call PackageBuilder directly from describe and format commands, instead use  Workspace::loadRootPackage API for loading a root package
* update Workspace::loadRootPackage to compute artifacts


tomerd:
@swift-ci please smoke test

ykhandelwal913:
How can i test this version locally? as currently with latest xcode 14beta 3, we are getting swift package manager version 5.7. How can i update the SPM version locally to validate ?

ykhandelwal913:
@tomerd 
How can i test this version locally? as currently with latest xcode 14beta 3, we are getting swift package manager version 5.7. How can i update the SPM version locally to validate ?

PR #3811, opened 2021-10-20
Fix spaces in -DCMAKE_Swift_FLAGS

keith:
Previously this passed something like `-DCMAKE_Swift_FLAGS=
-module-cache-path "foo"` on Linux which did not actually propagate
correctly.

PR #3812, opened 2021-10-20
[DNM] just testing

WowbaggersLiquidLunch:
It appears that I can't summon the CI bot in this repo, but can in the swift repo. Just testing it out here.


WowbaggersLiquidLunch:
ok so it works here, but for some reason it doesn't work for #3811...

PR #3813, opened 2021-10-20
Add C++2b to the CXXLanguageStandard enum

benrimmington:
[C++2b][] (tentatively ISO C++ 2023) is available in Clang 13+.

* Copy the details from the [LangStandards.def][] file.
* Update the internal documentation.

[C++2b]: <https://clang.llvm.org/cxx_status.html#cxx23>
[LangStandards.def]: <https://github.com/apple/llvm-project/blob/apple/main/clang/include/clang/Basic/LangStandards.def>

### DIFF
--- a/Sources/PackageDescription/LanguageStandardSettings.swift
+++ b/Sources/PackageDescription/LanguageStandardSettings.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2017 - 2020 Apple Inc. and the Swift project authors
+ Copyright (c) 2017 - 2021 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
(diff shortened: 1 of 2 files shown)


tomerd:
@swift-ci please smoke test

neonichu (approved):

Sources/PackageDescription/LanguageStandardSettings.swift
```diff
@@ -142,6 +146,10 @@ public enum CXXLanguageStandard: String, Encodable {
     /// ISO C++ 2020 DIS with GNU extensions.
     @available(_PackageDescription, introduced: 5.4)
     case gnucxx20 = "gnu++20"
+
+    /// Working draft for ISO C++ 2023 DIS with GNU extensions.
+    @available(_PackageDescription, introduced: 999.0)
```
tomerd, line 151: ditto

PR #3814, opened 2021-10-20
improve treatment of directories with extensions

tomerd:
motivation: treat directories with extensions more correctly and intuitivly: only treat well known directory extensions specially

changes:
* define list of known directory extensions (eg .xcassets)
* only treat well known directory extensions specially
* add and adjust tests

rdar://82933763


tomerd:
@swift-ci please smoke test

PR #3815, opened 2021-10-20
replace TSC signal handler with Dispatch's

tomerd:
motivation: DispatchSignalSource is a supported cross platform signal handing API, we do not need a custom one in TSC

changes:
* replace use TSC's InterruptHandler with DispatchSignalSource
* fix an issue where some git processes did not get registered with the ProcessSet for termination candidates

### DIFF
--- a/Sources/Commands/SwiftTool.swift
+++ b/Sources/Commands/SwiftTool.swift
@@ -26,6 +26,10 @@ import TSCUtility
 import Workspace
 import XCBuildSupport
 
+#if os(Windows)
+import WinSDK
+#endif
+
 typealias Diagnostic = TSCBasic.Diagnostic
 
 private class ToolWorkspaceDelegate: WorkspaceDelegate {
@@ -257,6 +261,11 @@ extension SwiftCommand {
 }
 
 public class SwiftTool {
+    #if os(Windows)
+    // unfortunately this is needed for C callback handlers used by Windows shutdown handler
+    static var shutdownRegistry: (processSet: ProcessSet, buildSystemRef: BuildSystemRef)?
+    #endif
+
     /// The original working directory.
     let originalWorkingDirectory: AbsolutePath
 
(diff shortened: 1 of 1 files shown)


tomerd:
@swift-ci please smoke test

neonichu (approved):

Sources/Commands/SwiftTool.swift
```diff
@@ -340,14 +337,16 @@ public class SwiftTool {
...
-#if !os(macOS)
-            self.options._buildSystem = .native
-#endif
-
             let processSet = ProcessSet()
             let buildSystemRef = BuildSystemRef()
-            interruptHandler = try InterruptHandler {
+
+            // trap SIGINT to terminate sub-processes, etc
+            signal(SIGINT, SIG_IGN)
```
compnerd, line 344: I think that this is going to break - this needs to be in a `!os(Windows)` block.
```suggestion
#if !os(WIndows)
            signal(SIGINT, SIG_IGN)
#endif
```

tomerd: @compnerd does Windows n require calling `signal()`or the moral equivalent?

compnerd:
I *think* that this might work:
```
diff --git a/Sources/Commands/SwiftTool.swift b/Sources/Commands/SwiftTool.swift
index 184d117d..17baf7e6 100644
--- a/Sources/Commands/SwiftTool.swift
+++ b/Sources/Commands/SwiftTool.swift
@@ -26,6 +26,10 @@ import TSCUtility
 import Workspace
 import XCBuildSupport
 
+#if os(Windows)
+import WinSDK
+#endif
+
 typealias Diagnostic = TSCBasic.Diagnostic
 
 private class ToolWorkspaceDelegate: WorkspaceDelegate {
@@ -297,6 +301,10 @@ public class SwiftTool {
     /// The current build system reference. The actual reference is present only during an active build.
     let buildSystemRef: BuildSystemRef
 
+#if os(Windows)
+    static var storage: (ProcessSet, BuildSystemRef)?
+#endif
+
     /// The execution status of the tool.
     var executionStatus: ExecutionStatus = .success
 
@@ -341,6 +349,22 @@ public class SwiftTool {
             let buildSystemRef = BuildSystemRef()
 
             // trap SIGINT to terminate sub-processes, etc
+#if os(Windows)
+            _ = SetConsoleCtrlHandler({ _ in
+                // Terminate all processes on receiving an interrupt signal.
+                SwiftTool.storage?.0.terminate()
+                SwiftTool.storage?.1.buildSystem?.cancel()
+
+                // Reset the handler.
+                _ = SetConsoleCtrlHandler(nil, false)
+
+                // Exit as if by signal()
+                TerminateProcess(GetCurrentProcess(), 3)
+
+                return true
+            }, true)
+            SwiftTool.storage = (processSet, buildSystemRef)
+#else
             signal(SIGINT, SIG_IGN)
             let interruptSignalSource = DispatchSource.makeSignalSource(signal: SIGINT)
             interruptSignalSource.setEventHandler {
@@ -351,10 +375,7 @@ public class SwiftTool {
                 processSet.terminate()
                 buildSystemRef.buildSystem?.cancel()
 
-#if os(Windows)
-                // Exit as if by signal()
-                TerminateProcess(GetCurrentProcess(), 3)
-#elseif os(macOS) || os(OpenBSD)
+#if os(macOS) || os(OpenBSD)
                 // Install the default signal handler.
                 var action = sigaction()
                 action.__sigaction_u.__sa_handler = SIG_DFL
@@ -376,6 +397,7 @@ public class SwiftTool {
 #endif
             }
             interruptSignalSource.resume()
+#endif
 
             self.processSet = processSet
             self.buildSystemRef = buildSystemRef
```

PR #3817, opened 2021-10-21
fix regression in test output

tomerd:
motivation: fix a regression in test output when running in parallel

changes:
* only emit diagnostics when writeToOutputStream is set to true, otherwise accumalate error info
* ignore ProcessSetError.cancelled as we have a specific handler for when the process is signaled


tomerd:
@swift-ci please smoke test

PR #3824, opened 2021-10-28
[Optimus] Simplify resource_bundle_accessor.swift

Kyle-Ye:
Simplify resource_bundle_accessor.swift

### Modifications:

Change "?:" into more swifty "??" operator

### Result:

Simplify the generated resource_bundle_accessor.swift file

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -686,7 +686,7 @@ public final class SwiftTargetBuildDescription {
 
                 let preferredBundle = Bundle(path: mainPath)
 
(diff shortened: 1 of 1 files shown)


abertelrud:
> Using packages in Xcode uses Xcode's own build system, not SwiftPM's and therefore also a different implementation of this generator. Changing the generator in SwiftPM won't affect Xcode at this time.

In addition to this, the conceptual idea is that each build system should generate an accessor that matches how that build system packaged the resources when it built the artifact.  So on different platforms there could be very different ways of accessing them, and as long as the semantics are preserved, it should be seen as an implementation detail.

PR #3825, opened 2021-10-29
improve build complete message

tomerd:
motivation: when no work needs to be done the output is `[0/0] Build complete!` which is confusing

changes: treat the build complete step as a discrete step incrementing the step + total counts

### DIFF
--- a/Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
+++ b/Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
@@ -623,8 +623,8 @@ final class BuildOperationBuildSystemDelegateHandler: LLBuildBuildSystemDelegate
         queue.sync {
             if success {
                 self.progressAnimation.update(
-                    step: self.taskTracker.finishedCount,
-                    total: self.taskTracker.totalCount,
+                    step: self.taskTracker.finishedCount+1,
+                    total: self.taskTracker.totalCount+1,
                     text: "Build complete!")
             }
             self.progressAnimation.complete(success: success)
--- a/Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
+++ b/Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
@@ -621,13 +621,12 @@ final class BuildOperationBuildSystemDelegateHandler: LLBuildBuildSystemDelegate
 
     func buildComplete(success: Bool) {
         queue.sync {
-            if success {
-                self.progressAnimation.update(
-                    step: self.taskTracker.finishedCount+1,
-                    total: self.taskTracker.totalCount+1,
-                    text: "Build complete!")
-            }
             self.progressAnimation.complete(success: success)
+            if success {
+                self.progressAnimation.clear()
+                self.outputStream <<< "Build complete!\n"
+                self.outputStream.flush()
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


neonichu (approved):

Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
```diff
@@ -623,8 +623,8 @@ final class BuildOperationBuildSystemDelegateHandler: LLBuildBuildSystemDelegate
         queue.sync {
             if success {
                 self.progressAnimation.update(
-                    step: self.taskTracker.finishedCount,
-                    total: self.taskTracker.totalCount,
+                    step: self.taskTracker.finishedCount+1,
+                    total: self.taskTracker.totalCount+1,
```
WowbaggersLiquidLunch, line 627: Nit: I think the convention is leaving white space around infix operators.

WowbaggersLiquidLunch:
Just throwing in an idea for an alternative solution: Would a message like GNU Make's "nothing to be done for target/product" work too?

I'm not sure if it's a valid concern, but I feel the user might misunderstand "[1/1]" to mean that something is changed or something new is built.

tomerd:
> I'm not sure if it's a valid concern, but I feel the user might misunderstand "[1/1]" to mean that something is changed or something new is built.

i dont have a strong opinion. the reason we added the 'build complete' message was because SwiftPM was silently exiting in the no-op use case which was confusing. the 'build conplete' message itself is consistent across the no-op and regular builds. @neonichu @abertelrud opinions?

abertelrud:
> > I'm not sure if it's a valid concern, but I feel the user might misunderstand "[1/1]" to mean that something is changed or something new is built.
> 
> i dont have a strong opinion. the reason we added the 'build complete' message was because SwiftPM was silently exiting in the no-op use case which was confusing. the 'build conplete' message itself is consistent across the no-op and regular builds. @neonichu @abertelrud opinions?

I agree that `[1/1]` feels a bit misleading.  I also think that `[0/0]` is confusing.  So the suggestion of saying something like "Everything already up to date" or something to that effect makes sense to me.

tomerd:
> I agree that [1/1] feels a bit misleading. I also think that [0/0] is confusing. So the suggestion of saying something like "Everything already up to date" or something to that effect makes sense to me.

the challange with something like "Everything already up to date" is that it is lacking context, it needs to say something about the build as it may be printed when running `swift test`, `swift run` or other operations than perform a build first.

Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
```diff
@@ -621,13 +621,12 @@ final class BuildOperationBuildSystemDelegateHandler: LLBuildBuildSystemDelegate
...
-                self.progressAnimation.update(
-                    step: self.taskTracker.finishedCount,
-                    total: self.taskTracker.totalCount,
-                    text: "Build complete!")
+                self.progressAnimation.clear()
+                self.outputStream <<< "Build complete!\n"
+                self.outputStream.flush()
             }
-            self.progressAnimation.complete(success: success)
         }
```
tomerd, line 630: @abertelrud @neonichu updated per our discussion ^^

tomerd:
@swift-ci please smoke test

tomerd:
UX is now:

```
❯ ~/code/swift/swift-package-manager/.build/debug/swift-package clean
~/code/swift/swift-tools-support-core ± ● main
❯ ~/code/swift/swift-package-manager/.build/debug/swift-build
[101/101] Merging module TSCTestSupport
Build complete!
```

```
~/code/swift/swift-tools-support-core ± ● main
❯ ~/code/swift/swift-package-manager/.build/debug/swift-build
Build complete!
```

PR #3833, opened 2021-11-03
improve the availablity message for deprecate manifest APIs

tomerd:
motivation: existing message is generic and does not provide an actionable fix

changes: add information to the message to make it more actionable


tomerd:
@swift-ci please smoke test

neonichu (approved):

PR #3835, opened 2021-11-03
emit information diagnostics for dependency resolution

tomerd:
motivation: provide transperancy / debug information when needed

changes:
* add a new dependency resolver delegate that emits debug diagnostics for resover callbacks, wire it into the observability system
* this information will only show when running in verbose mode
* deprecate TracingDependencyResolverDelegate which was a bespoke solution serving the same need
* add and adjust tests

### DIFF
--- a/Sources/PackageGraph/DependencyResolver.swift
+++ b/Sources/PackageGraph/DependencyResolver.swift
@@ -8,6 +8,7 @@
  See http://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
 
+import Basics
 import Dispatch
 import PackageModel
 import TSCBasic
@@ -43,6 +44,7 @@ public protocol DependencyResolverDelegate {
     func solved(result: [DependencyResolver.Binding])
 }
 
+@available(*, deprecated, message: "user verbosity flags instead")
 public struct TracingDependencyResolverDelegate: DependencyResolverDelegate {
     private let stream: OutputByteStream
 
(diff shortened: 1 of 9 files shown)


tomerd:
UX:

```
~/code/swift/sourcekit-lsp ± ● main
❯ ~/code/swift/swift-package-manager/.build/debug/swift-package update --verbose | grep DependencyResolver
info: [DependencyResolver] derived 'swift-argument-parser' requirement '0.4.1..<0.5.0'
info: [DependencyResolver] derived 'yams' requirement '4.0.0..<4.1.0'
info: [DependencyResolver] derived 'swift-crypto' requirement '1.1.4..<1.2.0'
info: [DependencyResolver] resolving 'swift-crypto'
info: [DependencyResolver] resolved 'swift-crypto' @ '1.1.6'
info: [DependencyResolver] resolving 'swift-argument-parser'
info: [DependencyResolver] resolved 'swift-argument-parser' @ '0.4.4'
info: [DependencyResolver] resolving 'yams'
info: [DependencyResolver] resolved 'yams' @ '4.0.6'
info: [DependencyResolver] solved 'yams' (https://github.com/jpsim/Yams.git) at '4.0.6'
info: [DependencyResolver] solved 'swift-argument-parser' (https://github.com/apple/swift-argument-parser.git) at '0.4.4'
info: [DependencyResolver] solved 'swift-crypto' (https://github.com/apple/swift-crypto.git) at '1.1.6'
info: [DependencyResolver] solved 'swift-package-manager' (https://github.com/apple/swift-package-manager.git) at 'main'
info: [DependencyResolver] solved 'swift-tools-support-core' (https://github.com/apple/swift-tools-support-core.git) at 'main'
info: [DependencyResolver] solved 'sourcekit-lsp' (/Users/tomerd/code/swift/sourcekit-lsp) at 'unversioned'
info: [DependencyResolver] solved 'swift-driver' (https://github.com/apple/swift-driver.git) at 'main'
info: [DependencyResolver] solved 'indexstore-db' (https://github.com/apple/indexstore-db.git) at 'main'
info: [DependencyResolver] solved 'swift-llbuild' (https://github.com/apple/swift-llbuild.git) at 'main'
info: [DependencyResolver] dependency resolution complete!
```

tomerd:
@swift-ci please smoke test

Sources/PackageGraph/DependencyResolver.swift
```diff
@@ -99,6 +101,53 @@ public struct TracingDependencyResolverDelegate: DependencyResolverDelegate {
...
+    public init (observabilityScope: ObservabilityScope) {
+        self.observabilityScope = observabilityScope.makeChildScope(description: "DependencyResolver")
+    }
+
+    public func willResolve(term: Term) {
+        self.debug("resolving '\(term.node.package.identity)'")
+    }
+
+    public func didResolve(term: Term, version: Version, duration: DispatchTimeInterval) {
+        self.debug("resolved '\(term.node.package.identity)' @ '\(version)'")
```
abertelrud, line 116: Is this final, or is there any case in which the resolver might go back and change the version later on?

abertelrud (approved):

Sources/PackageGraph/DependencyResolver.swift
tomerd: I think it is final, but not 100% to be honest. in any case this PR does not change the resolver semantics, it only routes the information to the observability scope and logs it there in a similar fashion to how it did for the bespoke file based logging

abertelrud: Right, understood.  I was thinking more along the lines of whether an IDE could use this callback to, for example, mark packages has having been resolved to a version even while the package resolution proceeds with other packages.  I don't see a problem with this PR per se, other than potential for confusion if there are multiple lines of output suggesting that the package is being resolved to different versions.  Even that case seems pretty understandable to me in a CLI sense though, since a later line of output clearly supersedes an earlier one.  Thanks!

tomerd: > Right, understood. I was thinking more along the lines of whether an IDE could use this callback to, for example, mark a dependency has having been resolved to a version even while the dependency resolution proceeds to resolve other dependencies. I don't see a problem with this PR per se, other than potential for confusion if there are multiple lines of output suggesting that the package is being resolved to different versions. Even that case seems pretty understandable to me in a CLI sense though, since a later line of output clearly supersedes an earlier one. Thanks!

iirc we already have callbacks for making the package is resolved, which we use in the CLI

abertelrud: Right, I see `didComputeVersion()` on WorkspaceDelegate — which I think was added earlier this year — and I know that `didResolve()` calls through to that one, and that seems very useful!  But seeing this here raised the question of how "definitive" this is, i.e. whether it is only ever called once.  From the PubGrub code it looks as though it might be definitive, but figured I'd ask.  In any case, just trying to understand the semantics of this one and not a comment on you change in this PR.  Thanks!

neonichu (approved):

PR #3840, opened 2021-11-05
[NFC] Replace type-erased pattern matching with typed if-else blocks

WowbaggersLiquidLunch:
Motivation: Concrete types are better than `Any`.

Result: This removes an unreachable `default` case, and should result in a bit less runtime cost.


WowbaggersLiquidLunch:
@swift-ci please smoke test

elsh (approved):

PR #3842, opened 2021-11-08
Make the `@main` entry points for plugins be `async` so they can call other `async` APIs

abertelrud:
Make the `@main` entry point for plugins be `async`.

### Motivation:

This allows the `@main` entry point for plugins to call APIs marked `async`.

### Modifications:

- mark the protocols in the plugin API as `async`
- use `await` when calling those APIs
- add a required `-rpath` when compiling plugins on macOS older than 12.0 (not needed on other platforms)

rdar://83206272

*Note:*  This builds on https://github.com/apple/swift-package-manager/pull/3898 and will be easier to review once that one is merged.

### DIFF
--- a/Sources/Workspace/DefaultPluginScriptRunner.swift
+++ b/Sources/Workspace/DefaultPluginScriptRunner.swift
@@ -161,6 +161,18 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
                 macOSPackageDescriptionPath = runtimePath.appending(component: "libPackagePlugin.dylib")
             }
 
+            #if os(macOS)
+            // On macOS earlier than 12, add an rpath to the directory that contains the concurrency fallback library.
+            if #available(macOS 12.0, *) {
+                // Nothing is needed; the system has everything we need.
+            }
+            else {
+                // Add an `-rpath` so the Swift 5.5 fallback libraries can be found.
+                let swiftSupportLibPath = self.toolchain.swiftCompilerPath.parentDirectory.parentDirectory.appending(components: "lib", "swift-5.5", "macosx")
+                command += ["-Xlinker", "-rpath", "-Xlinker", swiftSupportLibPath.pathString]
+            }
+            #endif
+
             // Use the same minimum deployment target as the PackageDescription library (with a fallback of 10.15).
             #if os(macOS)
             let triple = self.hostTriple
--- a/Sources/PackagePlugin/Plugin.swift
+++ b/Sources/PackagePlugin/Plugin.swift
@@ -53,7 +53,7 @@ extension Plugin {
     
     /// Main entry point of the plugin — sets up a communication channel with
     /// the plugin host and runs the main message loop.
-    public static func main() throws {
+    public static func main() async throws {
         // Duplicate the `stdin` file descriptor, which we will then use for
         // receiving messages from the plugin host.
         let inputFD = dup(fileno(stdin))
@@ -93,7 +93,7 @@ extension Plugin {
         // indicating that we're done.
         while let message = try pluginHostConnection.waitForNextMessage() {
             do {
-                try handleMessage(message)
+                try await handleMessage(message)
             }
             catch {
                 // Emit a diagnostic and indicate failure to the plugin host,
@@ -104,7 +104,7 @@ extension Plugin {
         }
     }
     
-    fileprivate static func handleMessage(_ message: HostToPluginMessage) throws {
+    fileprivate static func handleMessage(_ message: HostToPluginMessage) async throws {
         switch message {
 
         case .performAction(let wireInput):
@@ -147,7 +147,7 @@ extension Plugin {
                 }
                 
                 // Invoke the plugin to create build commands for the target.
-                let generatedCommands = try plugin.createBuildCommands(context: context, target: target)
+                let generatedCommands = try await plugin.createBuildCommands(context: context, target: target)
                 
                 // Send each of the generated commands to the host.
                 for command in generatedCommands {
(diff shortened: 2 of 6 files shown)


tomerd:
lgtm, I think we need to set the tools-version to 5.5?

abertelrud:
> lgtm, I think we need to set the tools-version to 5.5?

Oh right, that.  I thought we already did, but I guess only up to 5.4.  Is anything blocking that at this point?

abertelrud:
@swift-ci please smoke test

tomerd:
> Oh right, that. I thought we already did, but I guess only up to 5.4. Is anything blocking that at this point?

no, we should be fine to go to 5.5

abertelrud:
> > Oh right, that. I thought we already did, but I guess only up to 5.4. Is anything blocking that at this point?
> 
> no, we should be fine to go to 5.5

@neonichu @tomerd What are the next steps here for moving SwiftPM's tools version up to 5.5?  I can open a separate PR if we're clear to do it (we probably need to coordinate with SourceKit-LSP though).

abertelrud:
> > > Oh right, that. I thought we already did, but I guess only up to 5.4. Is anything blocking that at this point?
> > 
> > 
> > no, we should be fine to go to 5.5
> 
> @neonichu @tomerd What are the next steps here for moving SwiftPM's tools version up to 5.5? I can open a separate PR if we're clear to do it (we probably need to coordinate with SourceKit-LSP though).

Actually it looks as if SourceKit-LSP is using tools version 5.3 today, and yet SwiftPM has already moved to 5.4.  How is that not already broken in `main` of both packages?

abertelrud:
SwiftPM is now at 5.5, updated and testing again.

abertelrud:
@swift-ci please smoke test

abertelrud:
Looks like macOS tests are not finding `libswift_Concurrency.dylib` in the toolchain (they do seem to be getting the `@rpath` prefixes correct).  I'm seeing something different locally, where not even the prefix is correct, so will try to replicate with the same version as the CI.

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

abertelrud:
`async` is still unhappy on macOS:
```
dyld: lazy symbol binding failed: can't resolve symbol _$ss13_runAsyncMainyyyyYaKcF in /private/var/folders/_8/79jmzf2142z2xydc_01btlx00000gn/T/Miscellaneous_Plugins.JM9GfQ/MySourceGenPlugin/.build/plugins/cache/MySourceGenBuildToolPlugin because dependent dylib /usr/lib/swift/libswift_Concurrency.dylib could not be loaded
dyld: can't resolve symbol _$ss13_runAsyncMainyyyyYaKcF in /private/var/folders/_8/79jmzf2142z2xydc_01btlx00000gn/T/Miscellaneous_Plugins.JM9GfQ/MySourceGenPlugin/.build/plugins/cache/MySourceGenBuildToolPlugin because dependent dylib /usr/lib/swift/libswift_Concurrency.dylib could not be loaded
```
The linux thing is something else though.

abertelrud:

abertelrud:
@swift-ci please smoke test

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -157,6 +157,16 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
...
 
+            // On macOS earlier than 12, add an rpath to the directory that contains the concurrency fallback library.
+            if #available(macOS 12.0, *) {
+                // Nothing is needed; the system has everything we need.
+            }
+            else {
+                // Add an `-rpath` so the Swift 5.5 fallback libraries can be found.
+                let swiftSupportLibPath = self.toolchain.swiftCompilerPath.parentDirectory.parentDirectory.appending(components: "lib", "swift-5.5", "macosx")
+                command += ["-Xlinker", "-rpath", "-Xlinker", swiftSupportLibPath.pathString]
+            }
```
tomerd, line 168: 1. do we need to care about other darwin based system beyond macOS?
2. should we test for compiler version and not assume 5.5?

abertelrud: That's a great point — I'll update the condition to include iOS 15 at least (I don't know if tvOS or watchOS are interesting).  Also, that seems like perhaps a problem with this PR in general, because we only want to add `-rpath` on Darwin on macOS, not on non-Darwin.  I'll revise.

abertelrud: I guess the original reason I did just macOS here is that the entire function is full of `#if macOS()` so will in practice not work on other Darwin platforms.  Note that macOS is the only one that allows spawning subprocesses — this code won't work on iOS, tvOS, and watchOS.

abertelrud:
@swift-ci please smoke test

neonichu (approved):

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: That's a good question.  @DougGregor would the compiler version for the async compatibility libraries be updated for the currently used compiler, i.e. should we base this on what version of the compiler happens to be installed?  Or would newer ones still use the `swift-5.5` subdirectory because the async feature was introduced in Swift 5.5?

abertelrud:
@please test macos

abertelrud:
@swift-ci please test macos

abertelrud:
@swift-ci please smoke test macos

abertelrud:
@swift please smoke test macOS platform

abertelrud:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke tes

tomerd:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci smoke test

abertelrud:
@swift-ci smoke test

tomerd (approved):

PR #3850, opened 2021-11-10
[Collections] Call VACUUM in async queue

yim-lee:
Motivation:
I noticed that when I run `package-collection remove` in CLI there is a warning in the output:

```
Failed to 'VACUUM' the database: SQL logic error
```

Modifications:
- Not sure what the real cause is, but the problem doesn't occur if there is a sleep between `DELETE` and `VACUUM`, so perhaps `VACUUM` fails because other operations are not done yet. Instead of calling `VACUUM` synchronously, do that in an async queue, which is probably more appropriate since `VACUUM` can be time consuming.
- Fix a bug in the same file where we didn't do `ROLLBACK` when a transaction fails.


yim-lee:
@swift-ci please smoke test

neonichu (approved):

tomerd (approved):

PR #3852, opened 2021-11-10
Update CODEOWNERS

tomerd:
add  @elsh to CODEOWNERS

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -11,13 +11,25 @@
 # N: Ankit Aggarwal
 # E: ankit_aggarwal@apple.com
 # D: Package Manager
-# 
+
 # N: Anders Bertelrud
 # E: anders@apple.com
 # D: Package Manager
- 
+
+# N: Boris Buegling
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -24,4 +24,4 @@
 
 # The following lines are used by GitHub to automatically recommend reviewers.
 
-* @abertelrud @neonichu @tomerd @friedbunny
+* @abertelrud @neonichu @tomerd @elsh
```
abertelrud, line 27: Intentional replacement?

neonichu (approved):

CODEOWNERS
tomerd: added. I think it makes sense to keep Daniel and Ankit info in that section since they show up in commit history, correct?

tomerd: > Intentional replacement (summary says to add)?

yes I believe @friedbunny would be less involved in these going forward

PR #3853, opened 2021-11-10
remove resolution retry on resolution failure

tomerd:
motivation: resolution retry + removing of package resolved in failure is difficult to justify

changes: remove the retry code from package resolution and "give up" after one attempt without mutating the package resolved file

### DIFF
--- a/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
+++ b/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
@@ -1394,14 +1394,14 @@ private final class ContainerProvider {
     /// Get the container for the given identifier, loading it if necessary.
     func getContainer(for package: PackageReference, completion: @escaping (Result<PubGrubPackageContainer, Error>) -> Void) {
         // Return the cached container, if available.
-        if let container = self.containersCache[package] {
+        if let container = self.containersCache[package], package.equalsIncludingLocation(container.package) {
             return completion(.success(container))
         }
 
         if let prefetchSync = self.prefetches[package] {
             // If this container is already being prefetched, wait for that to complete
             prefetchSync.notify(queue: .sharedConcurrent) {
-                if let container = self.containersCache[package] {
+                if let container = self.containersCache[package], package.equalsIncludingLocation(container.package) {
                     // should be in the cache once prefetch completed
                     return completion(.success(container))
                 } else {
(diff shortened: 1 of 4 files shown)


tomerd:
@neonichu @abertelrud do you agree with this change?

note there is a unit tests that actually checks for this behavior which is naturally now failing. my default posture is we should remove the test, but it does have the following somewhat interesting comment:

```
        // In this test, we get into a state where add an entry in the resolved
        // file for a transitive dependency whose URL is later changed to
        // something else, while keeping the same package identity.
        //
        // This is normally detected during pins validation before the
        // dependency resolution process even begins but if we're starting with
        // a clean slate, we don't even know about the correct urls of the
        // transitive dependencies. We will end up fetching the wrong
        // dependency as we prefetch the pins. If we get into this case, it
        // should kick off another dependency resolution operation which will
        // have enough information to remove the invalid pins of transitive
        // dependencies.
```

wdyt?

abertelrud:
> @neonichu @abertelrud do you agree with this change?

In theory I think this is a good change since it removes some source of confusion and "magic" behavior.

> note there is a unit tests that actually checks for this behavior which is naturally now failing. my default posture is we should remove the test, but it does have the following somewhat interesting comment:
> 
> ```
>         // In this test, we get into a state where add an entry in the resolved
>         // file for a transitive dependency whose URL is later changed to
>         // something else, while keeping the same package identity.
>         //
>         // This is normally detected during pins validation before the
>         // dependency resolution process even begins but if we're starting with
>         // a clean slate, we don't even know about the correct urls of the
>         // transitive dependencies. We will end up fetching the wrong
>         // dependency as we prefetch the pins. If we get into this case, it
>         // should kick off another dependency resolution operation which will
>         // have enough information to remove the invalid pins of transitive
>         // dependencies.
> ```
> 
> wdyt?

This is an interesting comment though.  So now we would just get a failure, and then the user would need to reresolve?

tomerd:
@swift-ci please smoke test

tomerd:

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
```diff
@@ -1394,14 +1394,14 @@ private final class ContainerProvider {
...
-        if let container = self.containersCache[package] {
+        if let container = self.containersCache[package], package.equalsIncludingLocation(container.package) {
             return completion(.success(container))
         }
 
         if let prefetchSync = self.prefetches[package] {
             // If this container is already being prefetched, wait for that to complete
             prefetchSync.notify(queue: .sharedConcurrent) {
-                if let container = self.containersCache[package] {
+                if let container = self.containersCache[package], package.equalsIncludingLocation(container.package) {
```
tomerd, line 1404: @abertelrud @neonichu fixed the logic so now the test passes, its actually the right thing to do anyways

tomerd:
@neonichu @abertelrud this is ready, also passes the package compatibility test

neonichu (approved):

abertelrud (approved):
Fantastic!  LGTM!

PR #3856, opened 2021-11-11
Call didLoadManifest on failure

krzyzanowskim:
Call didLoadManifest on failure as documented in `WorkspaceDelegate.didLoadManifest`

> The workspace has loaded a package manifest, either successfully or not. The manifest is nil if an error occurs, in which case there will also be at least one error in the list of diagnostics (there may be warnings even if a manifest is loaded successfully).


krzyzanowskim:
nvm. looks like it's fixed on `main` already

PR #3858, opened 2021-11-11
Fix Plugin Resolution

SDGGiesbrecht:
Spun out of #3838. This contains only the pieces related to fixing plugins, which contained bugs affecting both resolution modes.

I added two tests that are independent of the resolution mode, one to make sure plugin executables are usable even when they are not attached to any product (which was the primary thing broken), and the other to make sure that such executables are not accessible outside of the plugin itself.

To make the second test work even with legacy full resolution, I needed to explicitly block synthesized products from dependency links (in `PackageGraph+Loading.swift`). That should also make it more resilient during hypothetical future refactors, because it does not rely on the order of the loading process.

@abertelrud


SDGGiesbrecht:
The diff of the `PackageBuilder.swift` portion is much more legible [under `-w`](https://github.com/apple/swift-package-manager/pull/3858/files?w=1#diff-68f661bfe880d4bf7059718964ee104add3839e7e2f5bc8bfd644ab706ff7149), because it is mostly indentation changes caused by the `if` statement moving.

abertelrud:
Thanks a bunch of taking the time to split this out @SDGGiesbrecht!  This looks good to me.

abertelrud (approved):

abertelrud:
@swift-ci please smoke test

abertelrud:
Note: We'll probably see some async-related failures in the Linux self hosted test due to unrelated reasons.  Should clear up with a later toolchain, but shouldn't hold up merging this if everything else passes.

neonichu (approved):

abertelrud:
@swift-ci please smoke test

SDGGiesbrecht:
> Basics/Errors.swift:14: Fatal error: unknown checkout state sourceControl

Is that what #3865 is about?

tomerd:
> Is that what #3865 is about?

yes

tomerd:

abertelrud:
I did some extra testing on this, and it seems to work great, also in a client of libSwiftPM.  Thanks @SDGGiesbrecht.

PR #3859, opened 2021-11-11
Bump minimum swift-crypto v2 version to 2.0.1

yim-lee:
Motivation:

Test and upgrade to the latest swift-crypto [2.0.1](https://github.com/apple/swift-crypto/releases/tag/2.0.1) release, which
includes a BoringSSL update: https://github.com/apple/swift-crypto/pull/95

Modifications:

Tests that use real certs, which execute the OCSP code path, crash on Linux (`swift:5.5` and `swift:5.5-focal` docker images). (rdar://85280061)

The crash happens on L351 of `CertificatePolicy.swift` (`BoringSSLOCSPClient.checkStatus`) because `response`, an `OCSP_RESPONSE` pointer, is `NULL`. It is `NULL` because the logic in BoringSSL's `crypto/asn1/asn1_lib.c` has changed (https://github.com/google/boringssl/commit/ee510f58895c6a88b59f1b66a94939d08ebdfe5b), so the return value needs to be handled differently in `PackageCollectionsSigningLibc/asn1/a_d2i_fp.c`.

I also don't think this change is correct: https://github.com/google/boringssl/commit/efab69bf7323009da8257f47e7e0976a43e12c64#diff-dc711de78524fcd1edcc461a96ca36b80cac1bf03a67287a9b434f772a6f4e3dR170.

- Guard against `NULL` OCSP response in `BoringSSLOCSPClient`.
- Update logic in `PackageCollectionsSigningLibc/asn1/a_d2i_fp.c`:
  - Add `ASN1_get_object_without_inf` and `asn1_get_length_without_inf`, copied from [BoringSSL](https://github.com/google/boringssl/blob/ee510f58895c6a88b59f1b66a94939d08ebdfe5b/crypto/asn1/asn1_lib.c) but reverting the ["0x80 fix"](https://github.com/google/boringssl/commit/efab69bf7323009da8257f47e7e0976a43e12c64#diff-dc711de78524fcd1edcc461a96ca36b80cac1bf03a67287a9b434f772a6f4e3dR170), such that we use the same logic regardless of swift-crypto version. We can probably remove them later when we switch over to swift-crypto 2.x completely (unless I'm correct the 0x80 fix).
  - Remove `inf` in `asn1_d2i_read_bio`.
  - Note that there are `ASN1_get_object_with_inf` and `asn1_get_length_with_inf` too. We can use these instead to keep the same behavior as swift-crypt 2.0.0 and less. I will delete these if reviewers are ok with the current approach.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -54,7 +54,7 @@ automatic linking type with `-auto` suffix appended to product's name.
 let autoProducts = [swiftPMProduct, swiftPMDataModelProduct]
 
 let useSwiftCryptoV2 = ProcessInfo.processInfo.environment["SWIFTPM_USE_SWIFT_CRYPTO_V2"] != nil
-let minimumCryptoVersion: Version = useSwiftCryptoV2 ? "2.0.0" : "1.1.4"
+let minimumCryptoVersion: Version = useSwiftCryptoV2 ? "2.0.1" : "1.1.4"
 var swiftSettings: [SwiftSetting] = []
 if useSwiftCryptoV2 {
     swiftSettings.append(.define("CRYPTO_v2"))
--- a/Sources/PackageCollectionsSigning/Certificate/CertificatePolicy.swift
+++ b/Sources/PackageCollectionsSigning/Certificate/CertificatePolicy.swift
@@ -346,15 +346,17 @@ private struct BoringSSLOCSPClient {
                         return
                     }
 
-                    let response = d2i_OCSP_RESPONSE_bio(bio, nil)
+                    guard let response = d2i_OCSP_RESPONSE_bio(bio, nil),
+                          let basicResp = OCSP_response_get1_basic(response) else {
+                        results.append(.failure(OCSPError.responseConversionFailure))
+                        return
+                    }
                     defer { OCSP_RESPONSE_free(response) }
-                    let basicResp = OCSP_response_get1_basic(response)
                     defer { OCSP_BASICRESP_free(basicResp) }
 
                     // This is just the OCSP response status, not the certificate's status
                     guard OCSP_response_status(response) == OCSP_RESPONSE_STATUS_SUCCESSFUL,
-                        CCryptoBoringSSL_OBJ_obj2nid(response?.pointee.responseBytes.pointee.responseType) == NID_id_pkix_OCSP_basic,
-                        let basicRespData = basicResp?.pointee.tbsResponseData.pointee else {
+                        CCryptoBoringSSL_OBJ_obj2nid(response.pointee.responseBytes.pointee.responseType) == NID_id_pkix_OCSP_basic else {
                         results.append(.failure(OCSPError.badResponse))
                         return
                     }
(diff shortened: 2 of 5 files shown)


yim-lee:
@swift-ci please smoke test

yim-lee:
@Lukasa I would appreciate if you can take a look at this as well. Thanks!

tomerd:
linux CI failure on asyncTest expected until CI gets newer toolchain cc @shahmishal

shahmishal:
We have a new toolchain, please re-request testing.

tomerd:

Package.swift
```diff
@@ -54,7 +54,7 @@ automatic linking type with `-auto` suffix appended to product's name.
 let autoProducts = [swiftPMProduct, swiftPMDataModelProduct]
 
 let useSwiftCryptoV2 = ProcessInfo.processInfo.environment["SWIFTPM_USE_SWIFT_CRYPTO_V2"] != nil
```
buttaface, line 56: Are there any plans to update `build-script` and the CI to swift-crypto v2? Since this variable is not set, I think [it's still testing against 1.1.5](https://github.com/apple/swift/blob/2ff10a57f40da9a72a897b36fe0e15e533748e08/utils/update_checkout/update-checkout-config.json#L97).

Sources/PackageCollectionsSigning/Certificate/CertificatePolicy.swift
```diff
@@ -346,15 +346,17 @@ private struct BoringSSLOCSPClient {
                         return
                     }
 
-                    let response = d2i_OCSP_RESPONSE_bio(bio, nil)
+                    guard let response = d2i_OCSP_RESPONSE_bio(bio, nil),
+                          let basicResp = OCSP_response_get1_basic(response) else {
```
Lukasa, line 350: This has introduced a memory leak on the failure path if `response` is non-nil` but `basicResp` is `nil`.

Lukasa (commented):
Looks fine aside from the memory leak.

yim-lee:
@swift-ci please smoke test

Sources/PackageCollectionsSigning/Certificate/CertificatePolicy.swift
yim-lee: Amended https://github.com/apple/swift-package-manager/pull/3859/commits/a1c79c42564fdad57f2ccb2e91d246d6ad289dcb

Package.swift
yim-lee: > Since this variable is not set, I think it's still testing against 1.1.5

Right, SwiftPM is still using swift-crypto v1.

> Are there any plans to update build-script and the CI to swift-crypto v2?

No specific timeline yet. So far I've been testing locally (to do additional test cases that CI can't cover) whenever there's a new swift-crypto release. When we are ready to upgrade to v2 we'll probably remove this env var and stick with `build-script`.

neonichu (approved):

PR #3860, opened 2021-11-11
Fix PubGrub Container

SDGGiesbrecht:
Spun out of #3838. This contains only the pieces related to `PubGrubPackageContainer` and the bug in its caching. The relevant code branches are traversed and tested under both resolution modes. (But I think the context that flushes out the problems never occurs naturally under legacy resolution.)


SDGGiesbrecht:
This time too, [-w](https://github.com/apple/swift-package-manager/pull/3860/files?w=1) is helpful to filter out indentation‐only changes.

tomerd:
@swift-ci please smoke test

tomerd:

tomerd (approved):

PR #3862, opened 2021-11-11
Improve Resolver Determinism

SDGGiesbrecht:
Spun out of #3838. This makes sure the order of resolution diagnostics is deterministic.

Unlike the first iteration in #3838, this does it properly and preserves the user’s order (which turned out to be much easier than I expected). It also adds a test that would catch any regression, regardless of the resolution mode.

### DIFF
--- a/Sources/PackageGraph/PackageGraph+Loading.swift
+++ b/Sources/PackageGraph/PackageGraph+Loading.swift
@@ -230,7 +230,7 @@ private func createResolvedPackages(
             metadata: package.diagnosticsMetadata
         )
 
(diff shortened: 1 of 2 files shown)


tomerd:
@swift-ci please smoke test

tomerd (approved):

neonichu (approved):

abertelrud:
Thank you @SDGGiesbrecht, this is great!  Might seem like a small thing but I think it can help understandability in a subtle way.

abertelrud (approved):

abertelrud:
@neonichu and @tomerd all clear to merge?

PR #3865, opened 2021-11-12
state file parsing errors should not assert

tomerd:
motivation: state parsing is a "legitimate" error that we handle gracefuly and should not assert on

changes: use StringError instead of InternalError when encountering invalid kind/type attribute


tomerd:
@swift-ci please smoke test

yim-lee (approved):

neonichu (approved):

PR #3866, opened 2021-11-13
[Collections] Remove VACUUM

yim-lee:
`VACUUM` causes more problems than benefits, so removing it.

rdar://85368153


yim-lee:
@swift-ci please smoke test

abertelrud (approved):

PR #3871, opened 2021-11-15
Fix Identity Resolution

SDGGiesbrecht:
Spun out of #3838. This contains only the pieces related to identities.

The first commit adds a test against an existing bug in the logic that connects the dependency declarations of a target to their respective packages. Named packages (tools versions 5.2 to 5.5) are not looked up correctly, and the resolver ends up looking for them everywhere, as though the manifest were 5.1 or older. I do not think it can lead to anything breaking, but it does lead to excessive work in the resolver, and in rare cases, to more dependencies being fetched than necessary (If two dependencies have products with the same name, and only one of them is actually used, both will be fetched anyway). The test is only meaningful with target‐based resolution engaged, as legacy resolution overrides the logic with a request for `.everything` anyway.

The second commit fixes the bug by repairing the lookup.

The last commit updates several stale tests that were looking for old spellings of diagnostics which changed with the introduction of identities. Two of the four do not actually depend on target‐based resolution, and they were re‐enabled.

### DIFF
--- a/Sources/PackageModel/Manifest.swift
+++ b/Sources/PackageModel/Manifest.swift
@@ -175,12 +175,12 @@ public final class Manifest {
         #if ENABLE_TARGET_BASED_DEPENDENCY_RESOLUTION
         // If we have already calculated it, returned the cached value.
         if let dependencies = self._requiredDependencies[productFilter] {
-            return self.dependencies
+            return dependencies
         } else {
             let targets = self.targetsRequired(for: productFilter)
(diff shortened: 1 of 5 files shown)


tomerd:
@swift-ci please smoke test

tomerd:

tomerd (approved):

abertelrud (approved):

tomerd (approved):

PR #3874, opened 2021-11-18
remove SourceArchiver

tomerd:
motivation: SourceArchiver was introduced in RegistryClient but adds a system depedency on bsdtar

changes:
* remove SourceArchiver
* change RegistryClient to use ZipArchiver
* add utility to strip first level of a directory and use that in RegistryClient
* add tests


tomerd:
@swift-ci please smoke test

Sources/PackageRegistry/SourceArchiver.swift
```diff
@@ -1,72 +0,0 @@
...
-
-        guard fileSystem.isDirectory(destinationPath) else {
-            completion(.failure(FileSystemError(.notDirectory, destinationPath)))
-            return
-        }
-
-        // TODO: consider calling `libarchive` or some other library directly instead of spawning a process
-        DispatchQueue.sharedConcurrent.async {
-            do {
-                let result = try Process.popen(args: "bsdtar",
```
tomerd, line 57: `bsdtar` is a new system dependency and the reason it was used as explained above (see code comment) was for stripping the top level directory. this PR basically rolls back that and uses `unzip` which SwiftPM already relies on (and as such is already a system dependency) and handles stripping programmatically after the unzipping

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -426,7 +427,10 @@ public final class RegistryClient {
                             // TODO: Bail if archive contains relative paths or overlapping files
                             archiver.extract(from: downloadPath, to: destinationPath) { result in
                                 defer { try? fileSystem.removeFileTree(downloadPath) }
-                                completion(result.mapError { error in
+                                completion(result.tryMap {
+                                    // strip first level component
+                                    try fileSystem.stripFirstLevel(of: destinationPath)
```
tomerd, line 432: @yim-lee we would need to adopt this also in the compatibility test suite and update the fixtures accordingly

tomerd:
@swift-ci please smoke test

yim-lee (approved):

PR #3875, opened 2021-11-18
Increment SwiftPM tools version to 5.5 and minimum deployment target to 10.15.4 on macOS

abertelrud:
Increment SwiftPM tools version to 5.5 and minimum deployment target to 10.15.4 on macOS

### Motivation:

Allow use of Swift 5.5 features such as automatic `Codable` synthesis.  Minimum deployment target of 10.15.4 allows use of FileHandle improvements such as `read(upToCount:)`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.4
+// swift-tools-version:5.5
 
 /*
  This source file is part of the Swift.org open source project
--- a/Package.swift
+++ b/Package.swift
@@ -63,7 +63,7 @@ if useSwiftCryptoV2 {
 let package = Package(
     name: "SwiftPM",
     platforms: [
-        .macOS(.v10_15),
+        .macOS("10.15.4"),
         .iOS(.v13)
     ],
     products:


abertelrud:
@swift-ci please smoke test

abertelrud:
@benlangmuir would it be possible also for SourceKit-LSP to require 10.15.4 on macOS?  Seems like a small change compared with 10.15, and the reason was to allow the new Swift-friendlier FileHandle APIs (that throw Swift.Errors instead of raising ObjC exceptions).

I can open a SourceKit-LSP PR if that isn't a no-go.

benlangmuir:
> @benlangmuir would it be possible also for SourceKit-LSP to require 10.15.4 on macOS? Seems like a small change compared with 10.15, and the reason was to allow the new Swift-friendlier FileHandle APIs (that throw Swift.Errors instead of raising ObjC exceptions).

No concerns from me, we've generally bumped our deployment target whenever swiftpm decides to, I think.  CCing @akyrtzi  and @ahoppen in case they want to protest.

akyrtzi:
We can follow swiftpm on this, I don't have any concern.

abertelrud:
Great, thanks all for the comments.

abertelrud:
https://github.com/apple/sourcekit-lsp/pull/436
@swift-ci please smoke test

tomerd (approved):

neonichu (approved):

PR #3876, opened 2021-11-19
Fix archive-source command

yim-lee:
Motivation:
Source archives created using the `archive-source` command doesn't
include top-level directory.

Modification:
The trailing slash `/` in the `--prefix` argument to `git archive`,
which is the underlying command for `archive-source`, is required to
create top level directory. Without it the prefix is simply added to all the files and
dirs in the archive.


yim-lee:
@swift-ci please smoke test

tomerd (approved):

PR #3878, opened 2021-11-19
Fix CommandsTests.PackageToolTests/testArchiveSource by removing git version specific error message

pcbeard:
Make tests pass with different version of git installed.

### Motivation:

On my system, I have a more recent version of git installed (git version 2.33.1) than comes with Xcode (git version 2.30.1 (Apple Git-130)). This causes the unit test  CommandsTests.PackageToolTests/testArchiveSource to fail.

### Modifications:

Remove the test of the version-specific string.

### Result:

Test will pass on more platforms. This addresses https://bugs.swift.org/browse/SR-15500


tomerd (approved):

tomerd:
@swift-ci please smoke test

abertelrud (approved):

neonichu (approved):

PR #3882, opened 2021-11-24
Test the "false" path of `<`

WowbaggersLiquidLunch:
`XCTAssertGreaterThan` only asserts that the "true" path of `>` works, which implies that the "true" path of `<` works, but it doesn't test the "false" path. `XCTAssertFalse` covers the "false" path of `<`.

Motivation: better test coverage

Modifications: added a `XCTAssertFalse(<)` for each `XCTAssertGreaterThan`.

Result: better test coverage

### DIFF
--- a/Tests/PackageDescriptionTests/VersionTests.swift
+++ b/Tests/PackageDescriptionTests/VersionTests.swift
@@ -44,6 +44,10 @@ class VersionTests: XCTestCase {
         )
     }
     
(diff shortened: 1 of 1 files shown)


WowbaggersLiquidLunch:
@swift-ci please smoke test

WowbaggersLiquidLunch:
@swift-ci please smoke test

abertelrud (approved):

neonichu (approved):

tomerd:
thanks @WowbaggersLiquidLunch

PR #3884, opened 2021-11-30
Fix iOS minimum deployment annotation and fix compilation error on iOS from plugin changes

abertelrud:
Fix iOS minimum deployment annotation and fix compilation error on iOS from plugin changes.

### Motivation:

Fixes a build break on iOS.

### Modifications:

- set the iOS deployment target to 13.4 to match the macOS 10.15.4 set earlier — this provides access to Swift-friendly Foundation.FileHandle APIs
- conditionalize use of Process APIs that aren't available on iOS, watchOS, tvOS

### Result:

Be able to build on iOS again.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -64,7 +64,7 @@ let package = Package(
     name: "SwiftPM",
     platforms: [
         .macOS("10.15.4"),
-        .iOS(.v13)
+        .iOS("13.4")
     ],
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #3886, opened 2021-11-30
Revert "Initial implementation of Command Plugins"

compnerd:
Reverts apple/swift-package-manager#3855


compnerd:
@swift-ci please test

abertelrud:
@compnerd Is there a log where I can see how this failed?  It should be possible to fix this forward.

tomerd:
@abertelrud @compnerd is this addressed by #3888 and can be close?

compnerd:
@tomerd - I want to keep this open until we know that it does fix the build, otherwise, I think that we should roll back to get the builds green again - they have been broken for ~12 hours now.

abertelrud:
> @abertelrud @compnerd is this addressed by #3888 and can be close?

In my view we can close this, since I'm confident that we can fix this forward.  We are still testing #3888 but I don't see a circumstance in which we'd need to revert this commit.

PR #3888, opened 2021-11-30
Fix a build failure in plugins on Windows

abertelrud:
Windows UCRT doesn't have `setlinebuf` but does have `setvbuf` (which `setlinbuf` is just a cover for).

### Motivation:

Fixes a Windows build failure.

### Modifications:

- change `setlinebuf` to `setvbuf`

### Result:

It works on Windows too.

### DIFF
--- a/Sources/PackagePlugin/Plugin.swift
+++ b/Sources/PackagePlugin/Plugin.swift
@@ -82,7 +82,7 @@ extension Plugin {
         }
         
         // Turn off full buffering so printed text appears as soon as possible.
-        setlinebuf(stdout)
+        setvbuf(stdout, nil, _IOLBF, 0)
         
         // Open a message channel for communicating with the plugin host.
         pluginHostConnection = PluginHostConnection(


tomerd (approved):
seems fine to me, @compnerd to confirm

abertelrud:
> seems fine to me, @compnerd to confirm

Thanks for the quick review.  When reworking how the plugin communication works I stuck to Foundation and FileHandle as much as a I could in order to be portable, but there were a couple of places where I used calls that are in UCRT, based on the Microsoft documentation.  This particular call is to make sure printouts from long-running plugins get delivered right away so they can be show in SwiftPM.

neonichu (approved):

compnerd:
No, this doesn't seem to resolve the issue(s):

```
S:\SourceCache\swift-package-manager\Sources\Workspace\DefaultPluginScriptRunner.swift:360:26: error: ambiguous use of 'close()'
                    try? outputHandle.close()
                         ^
S:\SourceCache\swift-corelibs-foundation\Sources\Foundation\FileHandle.swift:634:17: note: found this candidate
    public func close() throws {
                ^
S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:1020:17: note: found this candidate
    public func close() throws {
                ^
S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\WritableByteStream.swift:63:10: note: found this candidate
    func close() throws { }
         ^
```

abertelrud:
> No, this doesn't seem to resolve the issue(s):
> 
> ```
> S:\SourceCache\swift-package-manager\Sources\Workspace\DefaultPluginScriptRunner.swift:360:26: error: ambiguous use of 'close()'
>                     try? outputHandle.close()
>                          ^
> S:\SourceCache\swift-corelibs-foundation\Sources\Foundation\FileHandle.swift:634:17: note: found this candidate
>     public func close() throws {
>                 ^
> S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\Process.swift:1020:17: note: found this candidate
>     public func close() throws {
>                 ^
> S:\SourceCache\swift-tools-support-core\Sources\TSCBasic\WritableByteStream.swift:63:10: note: found this candidate
>     func close() throws { }
>          ^
> ```

Thanks from trying this out @compnerd.  It does get past the other error, and this one is independent (a different target and source file).  I'm getting set up with a local Windows development environment and will add the fix for this other issue.

abertelrud:
> Process.swift

It's because of this in ToolsSupportCore:
```
#if os(Windows)
extension FileHandle: WritableByteStream {
    public var position: Int {
        return Int(offsetInFile)
    }

    public func write(_ byte: UInt8) {
        write(Data([byte]))
    }

    public func write<C: Collection>(_ bytes: C) where C.Element == UInt8 {
        write(Data(bytes))
    }

    public func flush() {
        synchronizeFile()
    }

    public func close() throws {
        closeFile()
    }
}
#endif
```

Since Foundation.FileHandle already implements `close()`, this sets up an ambiguity.  This seems like an unfortunate thing for TSC to be doing actually since presumably any use of Foundation.FileHandle in a source file that also imports TSCBasic would be affected by this.

@compnerd @sergiocampama  I don't know the history here, should that `close()` be public?

compnerd:
I don't remember tbh, but, if the things build without it 🤷‍♂️ (testing the removal of that)

abertelrud:
> I don't remember tbh, but, if the things build without it 🤷‍♂️ (testing the removal of that)

I'm looking at a workaround as well, but having those public just seems to "poison the well" for everyone.

compnerd:
> > I don't remember tbh, but, if the things build without it 🤷‍♂️ (testing the removal of that)
> 
> I'm looking at a workaround as well, but having those public just seems to "poison the well" for everyone.

Well, a build of SPM and swift-driver without that change seems to pass ...

compnerd (approved):
Conditionally approving, we still need to fix the second error.

abertelrud:
> > > I don't remember tbh, but, if the things build without it 🤷‍♂️ (testing the removal of that)
> > 
> > 
> > I'm looking at a workaround as well, but having those public just seems to "poison the well" for everyone.
> 
> Well, a build of SPM and swift-driver without that change seems to pass ...

Great, thank you for trying it out!  (I'm still in the middle of setting up a Windows test env).

So what are the next steps, do you think it can be removed?

abertelrud:
> Conditionally approving, we still need to fix the second error.

Seems to me that could merge this which gets us further, and then decide if the fix for the other one is in TSC or another PR here.  The failures seem unrelated.

compnerd:
I think that removal is reasonable - we can build swift-driver, swift-package-manager, swift-tools-support-core without the single function, and that would unblock the builds and then obviates the need for the revert.  We can address any fallout subsequently.

abertelrud:
> I think that removal is reasonable - we can build swift-driver, swift-package-manager, swift-tools-support-core without the single function, and that would unblock the builds and then obviates the need for the revert. We can address any fallout subsequently.

Sounds like a good approach.  Thanks @compnerd.  I'll merge this, since it seems a good piece of the solution regardless.

PR #3889, opened 2021-11-30
address race condition in platforms where sqlite is not safe to access concurrently

tomerd:
motivation: on some platforms sqlite is not safe to access concuurently, so never assume it is safe

changes: always run the process code inside the sqlite access lock

rdar://85856852


tomerd:
@swift-ci please smoke test

abertelrud (approved):

PR #3891, opened 2021-11-30
Convert RepositoryHandle to a struct

pcbeard:
This solves a data race by preventing other threads from concurrently reading a handle while its being updated.

### Motivation:

This data race can happen if multiple thread call into RepositoryManager.lookup(). This was uncovered by
allowing more concurrency with the new implementation of TSCBasic.Process.popen().

### Modifications:

By converting to a struct threads can no longer simultaneously access a  `RepositoryHandle` which is now a value. Consequently, it no longer needs a lock to protect its internal state. When it gets updated, a new copy of it is written into the `RepositoryManager.repositories` dictionary.

Fixed up the unit tests to deal with the value nature of RepositoryHandle.

### Result:

This fixes a data race that's observable using the command:

```bash
swift test --sanitize=thread --filter CommandsTests.BuildToolTests/testNonReachableProductsAndTargetsFunctional
```

### DIFF
--- a/Sources/SourceControl/RepositoryManager.swift
+++ b/Sources/SourceControl/RepositoryManager.swift
@@ -129,98 +129,96 @@ public class RepositoryManager {
         // Dispatch the action we want to take on the serial queue of the handle.
         self.lookupQueue.addOperation {
             // First look for the handle.
-            let handle = self.getHandle(for: repository)
+            var handle = self.getHandle(for: repository)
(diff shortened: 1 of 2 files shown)


tomerd:
alternative: https://github.com/apple/swift-package-manager/pull/3893

pcbeard:
Looks like another approach will be used.

tomerd:
thanks @pcbeard

PR #3897, opened 2021-12-01
Fix a bug in the plugin context where the top-level package was not always the root package

abertelrud:
Fix a bug with the information provided in the context to plugins when there are multiple packages involved.

### Modifications:

- fix the serialization logic to always end up with the right package id
- extend a unit test to check for this

### DIFF
--- a/Sources/SPMBuildCore/PluginInvocation.swift
+++ b/Sources/SPMBuildCore/PluginInvocation.swift
@@ -885,7 +885,7 @@ struct PluginScriptRunnerInputSerializer {
             return nil
         }
         
-        // We only get this far if we are serializing the target. If so we also serialize its dependencies.
+        // We only get this far if we are serializing the target. If so we also serialize its dependencies. This needs to be done before assigning the next wire ID for the target we're serializing, to make sure we end up with the correct one.
         let dependencies: [PluginScriptRunnerInput.Target.Dependency] = try target.dependencies(satisfying: buildEnvironment).compactMap {
             switch $0 {
             case .target(let target, _):
@@ -973,6 +973,11 @@ struct PluginScriptRunnerInputSerializer {
             }
         }
 
+        // Serialize the dependency packages. This needs to be done assigning the next wire ID to the package we're serializing, to make sure we end up with the correct one.
+        let dependencies = try package.dependencies.map {
+            PluginScriptRunnerInput.Package.Dependency(packageId: try serialize(package: $0))
+        }
+
         // Assign the next wire ID to the package and append a serialized Package record.
         let id = packages.count
         packages.append(.init(
@@ -984,7 +989,7 @@ struct PluginScriptRunnerInputSerializer {
                 major: package.manifest.toolsVersion.major,
                 minor: package.manifest.toolsVersion.minor,
                 patch: package.manifest.toolsVersion.patch),
-            dependencies: try package.dependencies.map{ .init(packageId: try serialize(package: $0)) },
+            dependencies: dependencies,
             productIds: try package.products.compactMap{ try serialize(product: $0) },
             targetIds: try package.targets.compactMap{ try serialize(target: $0) }))
         packagesToIds[package] = id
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -188,19 +188,22 @@ class PluginTests: XCTestCase {
     
     func testCommandPluginInvocation() throws {
         try testWithTemporaryDirectory { tmpPath in
-            // Create a sample package with a library target and a plugin.
+            // Create a sample package with a library target and a plugin. It depends on a sample package.
             let packageDir = tmpPath.appending(components: "MyPackage")
-            try localFileSystem.writeFileContents(packageDir.appending(component: "Package.swift")) {
+            try localFileSystem.writeFileContents(packageDir.appending(components: "Package.swift")) {
```
tomerd, line 193: ?

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -263,9 +293,11 @@ class PluginTests: XCTestCase {
 
             // Load the package graph.
             let packageGraph = try workspace.loadPackageGraph(rootInput: rootInput, observabilityScope: observability.topScope)
-            XCTAssert(observability.diagnostics.isEmpty, "\(observability.diagnostics)")
-            XCTAssert(packageGraph.packages.count == 1, "\(packageGraph.packages)")
-            let package = try XCTUnwrap(packageGraph.packages.first)
+            XCTAssert(!observability.hasErrorDiagnostics, "\(observability.diagnostics)")
+            XCTAssert(!observability.hasWarningDiagnostics, "\(observability.diagnostics)")
```
tomerd, line 297: there is also XCTAssertNoDiangostics

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -984,7 +989,7 @@ struct PluginScriptRunnerInputSerializer {
                 major: package.manifest.toolsVersion.major,
                 minor: package.manifest.toolsVersion.minor,
                 patch: package.manifest.toolsVersion.patch),
-            dependencies: try package.dependencies.map{ .init(packageId: try serialize(package: $0)) },
```
tomerd, line 987: I fail to see the difference between this and the fix above, can you help me see it?

tomerd (approved):

Sources/SPMBuildCore/PluginInvocation.swift
abertelrud: This moves the serialization of the other packages up before the next wire ID is assigned, so that it refers to the correct package.  There are more comments about how this works in this source file but I'll add some commentary here to make this clear.

Tests/FunctionalTests/PluginTests.swift
abertelrud: To make it consistent with all the other calls in this function.  This one just happens to have a single component.  I can change it back.

Tests/FunctionalTests/PluginTests.swift
abertelrud: Thanks, didn't know about that one.  The problem here was that I don't actually care about info and debug diagnostics (there might be some of those).  I saw the `problemsOnly` flag to XCTAssertNoDiags, that's great.

abertelrud: In this case we get things like `- [debug]: [DependencyResolver] solved 'mypackage'`

Sources/SPMBuildCore/PluginInvocation.swift
abertelrud: Done

Tests/FunctionalTests/PluginTests.swift
abertelrud: Done

Tests/FunctionalTests/PluginTests.swift
abertelrud: Changed it

abertelrud:
@swift-ci please smoke test

abertelrud:
@tomerd are you fine with me landing this at this point?  I made the changes you suggested.

neonichu (approved):

PR #3898, opened 2021-12-01
Adjust command plugin implementation to match SE-0332 amendments

abertelrud:
This adjusts the implementation to match https://github.com/apple/swift-evolution/pull/1496

Specifically:
- removes the `outputPath` parameter to the plugin again
- renamed `outcome` property to `result` in the tests
- change the `PluginCommandIntent` and `PluginPermission` enums into opaque structs, with static constructor functions for the enum cases

### DIFF
--- a/Sources/SPMBuildCore/PluginInvocation.swift
+++ b/Sources/SPMBuildCore/PluginInvocation.swift
@@ -20,7 +20,7 @@ public typealias Diagnostic = Basics.Diagnostic
 
 public enum PluginAction {
     case createBuildToolCommands(target: ResolvedTarget)
-    case performCommand(targets: [ResolvedTarget], arguments: [String], outputPath: AbsolutePath?)
+    case performCommand(targets: [ResolvedTarget], arguments: [String])
 }
 
 extension PluginTarget {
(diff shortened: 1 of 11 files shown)


Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -493,16 +499,14 @@ public struct PluginInvocationTestResult: Encodable {
             public var tests: [Test]
             public struct Test: Encodable {
                 public var name: String
-                public var output: String
-                public var status: Status
+                public var outcome: Outcome
                 public var duration: Double
-                public enum Status: String, CaseIterable, Encodable {
+                public enum Outcome: String, Encodable {
```
tomerd, line 504: Result?

abertelrud: I had that initially, actually, but avoid it in the proposal since result is also the name of the outer structure.  But "result" seems like a better name even so.  This matches the proposal so if we change it that will need to be an amendment.

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

abertelrud:
This is ready for review.

neonichu (approved):

PR #3899, opened 2021-12-01
Build local and remote tools depended on by command plugins before running the plugins

abertelrud:
When a command plugin has a dependency on an executable, it needs to be built or brought-up-to-date and included in the tool map.

### Modifications:

- extended the SwiftPackageTool logic to build these tools first
- other small fixes in the SwiftPackageTool logic for running plugins
- include them in the tools map sent to the command plugin
- extend a unit test to include this local and remote built tools

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -922,16 +922,11 @@ extension SwiftPackageTool {
             let plugin = matchingPlugins[0]
             swiftTool.observabilityScope.emit(info: "Running plugin \(plugin)")
             
-            // The proposal calls for applying plugins to regular targets.
-            func isEligible(target: ResolvedTarget) -> Bool {
-                return ![.plugin, .snippet, .systemModule].contains(target.type)
-            }
-            
             // In SwiftPM CLI, we have only one root package.
             let package = packageGraph.rootPackages[0]
 
             // If no targets were specified, default to all the applicable ones in the package.
-            let targetNames = targetNames.isEmpty ? package.targets.filter(isEligible).map(\.name) : targetNames
+            let targetNames = targetNames.isEmpty ? package.targets.filter(\.isEligibleForPluginCommand).map(\.name) : targetNames
             
             // Find the targets (if any) specified by the user. We expect them in the root package.
             var targets: [String: ResolvedTarget] = [:]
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -921,17 +922,25 @@ extension SwiftPackageTool {
             let plugin = matchingPlugins[0]
             swiftTool.observabilityScope.emit(info: "Running plugin \(plugin)")
             
-            // Find the targets (if any) specified by the user. We expect them in the root package.
+            // The proposal calls for applying plugins to regular targets.
+            func isEligible(target: ResolvedTarget) -> Bool {
+                return ![.plugin, .snippet, .systemModule].contains(target.type)
+            }
```
tomerd, line 928: can this be made into an extension on ResolvedTarget?

Tests/CommandsTests/PackageToolTests.swift
```diff
@@ -1170,27 +1170,37 @@ final class PackageToolTests: CommandsTestCase {
     func testCommandPlugin() throws {
```
tomerd, line 1170: is it possible to add a test for the binary dependency case?

tomerd (approved):
looks good, some small suggestions

Tests/CommandsTests/PackageToolTests.swift
abertelrud: Definitely, I actually have that in a later PR, but can bring that here.

Sources/Commands/SwiftPackageTool.swift
abertelrud: 👍

abertelrud:
> looks good, some small suggestions

Thanks, implemented both.

abertelrud:
@swift-ci please smoke test

abertelrud:
This failure is in the updated unit test code, and only non-Darwin.  Will adjust it.

abertelrud:
@swift-ci please smoke test

abertelrud:
@tomerd are you fine with me landing this at this point?  I made the changes you suggested.

neonichu (approved):

PR #3900, opened 2021-12-02
Don't use scoped imports when building with library evolution enabled

pcbeard:
Fixes a compiler warning.

### Motivation:

Nobody likes warnings.

### Modifications:

Just use `import Foundation`

### Result:

Warnings are gone.


tomerd (approved):

tomerd:
@swift-ci please smoke test

PR #3901, opened 2021-12-02
Mirrors config migration should not modify user's file

yim-lee:
Motivation:
The migration from `.swiftpm/config` to `.swiftpm/configuration/mirrors.json` moves `.swiftpm/config` and leads user to think that the file gets deleted. (rdar://85917746)

Modifications:
- Copy `.swiftpm/config` to `.swiftpm/configuration/mirrors.json` instead of moving it.
- Emit warning that `.swiftpm/config` has been deprecated.
- Don't perform the migration if `.swiftpm/configuration/mirrors.json` already exists to prevent error. (rdar://85917836)


yim-lee:
@swift-ci please smoke test

tomerd (approved):

PR #3907, opened 2021-12-06
[Registry] Package release 'url' is optional

yim-lee:
Per the [API spec](https://github.com/apple/swift-package-manager/blob/main/Documentation/Registry.md#41-list-package-releases), a release's `url` field in the "list package releases" (`GET /{scope}/{name}`) API response is optional.

### DIFF
--- a/Sources/PackageRegistry/RegistryClient.swift
+++ b/Sources/PackageRegistry/RegistryClient.swift
@@ -586,10 +586,10 @@ extension RegistryClient {
             }
 
             public struct Release: Codable {
-                public var url: String
+                public var url: String?
                 public var problem: Problem?
 
-                public init(url: String, problem: Problem? = .none) {
+                public init(url: String?, problem: Problem? = .none) {
                     self.url = url
                     self.problem = problem
                 }


yim-lee:
@swift-ci please smoke test

tomerd:
@yim-lee what is the behavior if both are empty? or is the spec that one of them must be filled? if so, should we validate it on the client side?

yim-lee:
> what is the behavior if both are empty? 

Both being empty shouldn't cause problem. Quoting the spec,

"A server MAY specify a URL for a release using the url key. A client SHOULD locate a release using the value of the url key, if one is provided. Otherwise, the client SHOULD locate a release by expanding the URI Template /{scope}/{name}/{version} on the originating host."

AFAIU, `url` is not the archive URL--it just points to the release [details/JSON], so IMO it's not relevant to `RegistryClient`.  `RegistryClient` can always use `/{scope}/{name}/{version}`, which is a required API, to get the same.

> or is the spec that one of them must be filled? 

No. `url` is more like an "alternative" URL, and `problem` is set only if there is a problem, so both can be missing. The spec says `problem` SHOULD be filled if a release is unavailable. I think we should change the wording to MUST to make it required.

tomerd (approved):

abertelrud (approved):

PR #3908, opened 2021-12-06
Apply sandbox to commands generated by build tools (on platforms that support sandboxing)

abertelrud:
Sandboxing was already applied to the running of plugins, but for `buildTool` plugins, they were not to the commands they provide for the build system to run.

### Motivation:

Since the plugin can configure commands in any way it wants to, they need to be just as sandboxed as the plugin.

### Modifications:

This change applies the sandbox to the regular build commands and to the prebuild commands returned by `buildTool` plugins.

It also fixes a related bug that caused the sandboxing of plugins in general to ignore the `disable-sandbox` option, which is documented to disable sandboxing for all subprocesses.

Since many Foundation functions requires that the non-persistent temporary directory be writable, this also adds a new strictness level that allows this, and uses that when invoking the build commands.  Without this, many build commands become unusable from command plugins without disabling the sandbox completely.  In fact, half of the unit test fixture plugins (which use Foundation) fail without it.

Another approach is to set `TMPDIR` in the environment, though that counts on having the tool honor that setting.  In testing, it doesn't seem that all tools do that.

Specifically:
- add sandboxing of build commands and prebuild commands on platforms that support it
- make `disable-sandbox` work as advertised for plugin invocations also
- add a sandbox strictness level that allows writing to non-persistent temporary directories (but not caches, as the pre-5.3 manifest behavior)
- fix potential path quoting problems in the sandbox profile
- rename some parameters and properties to clarify that the build-related plugin results are specific to build tool plugins

Note that these changes are split into six separate commits, which might be easier to review individually.

rdar://74665029

### DIFF
--- a/Sources/Basics/Sandbox.swift
+++ b/Sources/Basics/Sandbox.swift
@@ -63,7 +63,7 @@ fileprivate func macOSSandboxProfile(
 
     // Allow writing only to certain directories.
     var writableDirectoriesExpression = writableDirectories.map {
-        "(subpath \"\(resolveSymlinks($0).pathString)\")"
+        "(subpath \(resolveSymlinks($0).quotedAsSubpathForSandboxProfile))"
     }
     // The following accesses are only needed when interpreting the manifest (versus running a compiled version).
     if strictness == .manifest_pre_53 {
@@ -75,7 +75,7 @@ fileprivate func macOSSandboxProfile(
     else if strictness == .writableTemporaryDirectory {
         writableDirectoriesExpression.append("(subpath \"/private/tmp\")")
         if let tmpDir = try? TSCBasic.determineTempDirectory() {
-            writableDirectoriesExpression += ["(subpath \"\(resolveSymlinks(tmpDir).pathString)\")"]
+            writableDirectoriesExpression += ["(subpath \(resolveSymlinks(tmpDir).quotedAsSubpathForSandboxProfile))"]
         }
     }
 
@@ -88,6 +88,16 @@ fileprivate func macOSSandboxProfile(
     }
 
     return contents
+}
+
+fileprivate extension AbsolutePath {
+    /// Private computed property that returns a version of the path as a string quoted for use as a subpath in a .sb sandbox profile.
+    var quotedAsSubpathForSandboxProfile: String {
+        return "\"" + self.pathString
+            .replacingOccurrences(of: "\\", with: "\\\\")
+            .replacingOccurrences(of: "\"", with: "\\\"")
+            + "\""
+    }
 }
 
 extension TSCUtility.Platform {
--- a/Sources/Basics/Sandbox.swift
+++ b/Sources/Basics/Sandbox.swift
@@ -29,6 +29,7 @@ public enum Sandbox {
     public enum Strictness: Equatable {
         case `default`
         case manifest_pre_53 // backwards compatibility for manifests
+        case writableTemporaryDirectory
     }
 }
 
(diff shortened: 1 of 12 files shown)


Sources/Basics/Sandbox.swift
```diff
@@ -82,6 +90,15 @@ fileprivate func macOSSandboxProfile(
...
 }
 
+fileprivate extension String {
+    var quotedAsSubpathForSandboxProfile: String {
+        return "\"" + self
+            .replacingOccurrences(of: "\\", with: "\\\\")
+            .replacingOccurrences(of: "\"", with: "\\\"")
+            + "\""
+    }
+}
```
tomerd, line 100: should this be a private extension on AbsolutePath?

Sources/Build/BuildOperation.swift
```diff
@@ -29,8 +29,11 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
     /// The closure for loading the package graph.
     let packageGraphLoader: () throws -> PackageGraph
     
-    /// The closure for invoking plugins in the package graph.
-    let pluginInvoker: (PackageGraph) throws -> [ResolvedTarget: [BuildToolPluginInvocationResult]]
+    /// The closure for invoking build tool plugins in the package graph.
+    let buildToolPluginInvoker: (PackageGraph) throws -> [ResolvedTarget: [BuildToolPluginInvocationResult]]
+
+    /// Whether to sandbox commands from build tool plugins.
+    public let enableSandboxForPluginCommands: Bool
```
tomerd, line 36: nite: should this be disableXXX since the default is true?

Sources/Build/LLBuildManifestBuilder.swift
```diff
@@ -33,6 +33,9 @@ public class LLBuildManifestBuilder {
     /// The build plan to work on.
     public let plan: BuildPlan
 
+    /// Whether to sandbox commands from build tool plugins.
+    public let enableSandboxForPluginCommands: Bool
```
tomerd, line 37: same nit

Sources/Commands/SwiftRunTool.swift
```diff
@@ -122,7 +122,7 @@ public struct SwiftRunTool: SwiftCommand {
                 buildParameters: buildParameters,
                 cacheBuildManifest: false,
                 packageGraphLoader: graphLoader,
-                pluginInvoker: { _ in [:] },
+                buildToolPluginInvoker: { _ in [:] },
```
tomerd, line 125: would this not be a problem for running your app in REPL? not urgent to fix now, but if it is a problem we should add a FIXME and radar

tomerd (approved):
nice, couple of suggestions

Sources/Commands/SwiftRunTool.swift
abertelrud: Yes, this was a TODO left over from earlier implementation.  Should add a Radar number for sure.

abertelrud:
> nice, couple of suggestions

Thanks — I am actually also trying the separate approach of setting `TMPDIR` to point inside the plugin output directory.  All of Foundation seems to honor it, but one concern is that it essentially turns it into a reboot-persisted temporary directory (since the plugin output directories aren't cleared on reboot as `/tmp` is).  Also, I don't think all tools honor it, though in theory they should.

Sources/Basics/Sandbox.swift
abertelrud: That's a good point — I'll do that.

Longer-term there should be some more organized way of doing quoting / escaping for a variety of purposes.  We could also consider using custom interpolators for it, since it doesn't matter so much what the string originally was.

This is also not the most elegant implementation of this, but it works and is called infrequently enough that we can pend that until we do something more comprehensive.

But for here I'll change it to extend AbsolutePath.

abertelrud:
> > nice, couple of suggestions
> 
> Thanks — I am actually also trying the separate approach of setting `TMPDIR` to point inside the plugin output directory. All of Foundation seems to honor it, but one concern is that it essentially turns it into a reboot-persisted temporary directory (since the plugin output directories aren't cleared on reboot as `/tmp` is). Also, I don't think all tools honor it, though in theory they should.

This doesn't seem like a reliable approach.  In think that allowing write permission to the non-persistent temporary directory seems alright conceptually for a sandbox, but we can refine over time.

Sources/Basics/Sandbox.swift
abertelrud: Changed.

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

PR #3913, opened 2021-12-07
Provide better and more detailed error messages when things go wrong compiling and running plugins

abertelrud:
Provide better and more detailed error messages when things go wrong compiling and running plugins

### Motivation:

This will help debug exceptional conditions, including in CI.  Regular syntax errors in plugins etc were already being reported.

### Modifications:

- add more specific cases to the errors reported by the DefaultPluginScriptRunner
- wrap all of the compiler command line construction in a do-catch block and convert the `try?` to  a `try` rather than paper over a failure to get minimum deployment target on Darwin
- emit the textual output of the compiler as a `info` to the observability scope
- emit more specific error messages, such as the signal number that ended the plugin invocation


abertelrud:
@swift-ci please smoke test

neonichu (approved):

PR #3915, opened 2021-12-07
Only recompile plugin scripts to executables when needed

abertelrud:
This is an optimization that skips the compilation step of a package plugin if a compiled version already exists, if none of the input files have changed, and if none of the compiler flags have changed.

### Motivation:

Better performance by avoiding unnecessary work.

### Modifications:

- compute hash of the input conditions and store it next to the compiled executable
- if the compiled executable exists and the hash is the same since the next time, skip compiling
- add a field to the compilation result structure to reflect whether the compilation was cached
- extend a unit test to verify that diagnostics (e.g. warnings) are still emitted

### Result:

No functional change, but less time spent compiling plugins.

rdar://74663095

### DIFF
--- a/Sources/Workspace/DefaultPluginScriptRunner.swift
+++ b/Sources/Workspace/DefaultPluginScriptRunner.swift
@@ -49,6 +49,22 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
             completion: completion)
     }
 
+    /// A synchronous version of `compilePluginScript()`.
+    public func compilePluginScript(
+        sources: Sources,
+        toolsVersion: ToolsVersion,
+        observabilityScope: ObservabilityScope
+    ) throws -> PluginCompilationResult {
+        // Call the asynchronous version. In our case we don't care which queue the callback occurs on.
+        return try tsc_await { self.compilePluginScript(
+            sources: sources,
+            toolsVersion: toolsVersion,
+            observabilityScope: observabilityScope,
+            callbackQueue: DispatchQueue.sharedConcurrent,
+            completion: $0)
+        }
+    }
+
     /// Public protocol function that starts evaluating a plugin by compiling it and running it as a subprocess. The tools version controls the availability of APIs in PackagePlugin, and should be set to the tools version of the package that defines the plugin (not the package containing the target to which it is being applied). This function returns immediately and then repeated calls the output handler on the given callback queue as plain-text output is received from the plugin, and then eventually calls the completion handler on the given callback queue once the plugin is done.
     public func runPluginScript(
         sources: Sources,
@@ -107,6 +123,9 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
     ) {
         // FIXME: Much of this is similar to what the ManifestLoader is doing. This should be consolidated.
         do {
+            // We could name the executable anything, but using the plugin name makes it more understandable.
+            let execName = sources.root.basename.spm_mangledToC99ExtendedIdentifier()
+
             // Get access to the path containing the PackagePlugin module and library.
             let runtimePath = self.toolchain.swiftPMLibrariesLocation.pluginAPI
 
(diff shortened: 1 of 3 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -179,35 +182,80 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
...
+                            diagnosticsFile: diagnosticsFile,
+                            compiledExecutable: executableFile,
+                            wasCached: false)
+                        guard $0.exitStatus == .terminated(code: 0) else {
+                            throw DefaultPluginScriptRunnerError.compilationFailed(result)
+                        }
+
+                        // We only get here if the compilation succeeded. In this case, we also write out the hash of the inputs before returning the result.
+                        if let compilerInputsHash = compilerInputsHash {
+                            try? localFileSystem.writeFileContents(hashFile, string: compilerInputsHash)
```
tomerd, line 243: do we want to emit any diagnostics if writing fails?

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -179,35 +182,80 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
...
+                }
+            }
+            if compilationNeeded {
+                // We need to recompile the executable, so we do so asynchronously.
+                Process.popen(arguments: command, environment: toolchain.swiftCompilerEnvironment, queue: callbackQueue) {
+                    // We are now on our caller's requested callback queue, so we just call the completion handler directly.
+                    dispatchPrecondition(condition: .onQueue(callbackQueue))
+                    completion($0.tryMap {
+                        // Emit the compiler output as observable info.
+                        let compilerOutput = ((try? $0.utf8Output()) ?? "") + ((try? $0.utf8stderrOutput()) ?? "")
```
tomerd, line 228: is try? hiding error we should report?

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -179,35 +182,80 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
...
+                    try stream <<< localFileSystem.readFileContents(sourceFile).contents
+                }
+                return stream.bytes.sha256Checksum
+            }()
+
+            // If we already have a compiled executable, then compare its hash with the new one. If they're the same, we don't need to compile.
+            var compilationNeeded = true
+            let hashFile = executableFile.parentDirectory.appending(component: execName + ".inputhash")
+            if localFileSystem.exists(executableFile) {
+                if (try? localFileSystem.readFileContents(hashFile)) == compilerInputsHash {
```
tomerd, line 217: is try? hiding error we should report?

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -179,35 +182,80 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
...
-
-                    // We return a PluginCompilationResult for both the successful and unsuccessful cases (to convey diagnostics, etc).
-                    let result = PluginCompilationResult(compilerResult: $0, diagnosticsFile: diagnosticsFile, compiledExecutable: executableFile)
-                    guard $0.exitStatus == .terminated(code: 0) else {
-                        throw DefaultPluginScriptRunnerError.compilationFailed(result)
-                    }
-                    return result
-                })
+            // Hash the command line and the contents of the source files to decide whether we need to recompile the plugin executable. We have a bounded set of input conditions, and can control what we hash.
+            let compilerInputsHash: String? = try? {
```
tomerd, line 204: is try? hiding error we should report?

Tests/SPMBuildCoreTests/PluginInvocationTests.swift
```diff
@@ -294,20 +314,140 @@ class PluginInvocationTests: XCTestCase {
...
             XCTAssert(result.compiledExecutable.components.contains("plugin-cache"), "\(result.compiledExecutable.pathString)")
             XCTAssert(result.diagnosticsFile.suffix == ".dia", "\(result.diagnosticsFile.pathString)")
             let contents = try localFileSystem.readFileContents(result.diagnosticsFile)
             let diags = try SerializedDiagnostics(bytes: contents)
             XCTAssertEqual(diags.diagnostics.count, 1)
             let warningDiag = try XCTUnwrap(diags.diagnostics.first)
             XCTAssertTrue(warningDiag.text.hasPrefix("variable \'unused\' was never used"), "\(warningDiag)")
+
+            // Recompile the command plugin again without changing its source code.
+            let recompilationResult = try tsc_await { pluginScriptRunner.compilePluginScript(
```
tomerd, line 334: nit: create a test only private extension for PluginScriptRunner that is an sync, throwing version of compilePluginScript (which does the `try tsc_await` dance)

tomerd (approved):
looks good, couple of questions

neonichu (approved):

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: Agreed.  Fixed.

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: Not in this case.  The API here is weird; in practice this will only throw if we don't collect the output, which we do.  This could be force unwrap except that we try to avoid those.

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: Agreed.  Fixed.

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: Agreed.  Fixed.

Tests/SPMBuildCoreTests/PluginInvocationTests.swift
abertelrud: Sounds good.  Fixed.

abertelrud: I also cleaned up the rest of the test to be more readable.  Could use some refactoring of the test, but will do in a later PR.

abertelrud:
@swift-ci please smoke test

PR #3916, opened 2021-12-08
Initial version of the `--list` option for command plugins

abertelrud:
Initial version of the `--list` option for command plugins.  This shows each command, along with the target and package that provides them.  A later change can easily refine this to also list the intent (separately from the command).

### Motivation:

This is useful and is a part of the proposal SE-0332.

### Modifications:

- add a `--list` option to `swift package plugin`
- extend a unit test

### Result:

Invoking `swift package plugin --list` now shows all available command plugins, initially in the form:

`‘mycmd’ (plugin ‘MyPlugin’ in package ‘MyPackage’)`

We could refine this to also group by intent, and should also show the description where there is a custom intent.  Suggestions are welcome.

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -882,39 +882,48 @@ extension SwiftPackageTool {
                 help: "Target(s) to which the plugin command should be applied")
         var targetNames: [String] = []
 
-        @Argument(help: "Name of the command plugin to invoke")
-        var command: String
+        @Flag(name: .customLong("list"),
+              help: "List the available plugin commands")
+        var listCommands: Bool = false
 
-        @Argument(help: "Any arguments to pass to the plugin")
-        var arguments: [String] = []
-        
-        func findPlugins(matching command: String, in graph: PackageGraph) -> [PluginTarget] {
-            // Find all the available plugin targets.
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -882,39 +882,48 @@ extension SwiftPackageTool {
...
-            
-            // Find the plugins that match the command.
-            swiftTool.observabilityScope.emit(info: "Finding plugin for command '\(command) '")
-            let matchingPlugins = findPlugins(matching: command, in: packageGraph)
+
+            // List the available plugins, if asked to.
+            if listCommands {
+                let allPlugins = availableCommandPlugins(in: packageGraph)
+                for plugin in allPlugins.sorted(by: { $0.name < $1.name }) {
+                    guard case .command(let intent, _) = plugin.capability else { return }
```
tomerd, line 901: nit: use filter?

tomerd (approved):
nice

neonichu (approved):

Sources/Commands/SwiftPackageTool.swift
abertelrud: I wasn't sure how to do that while also unpacking the intent and permissions (which aren't yet used in this PR but should be).  How would that look?  I'd like to switch to filter if I can.

PR #3917, opened 2021-12-08
Add the ability to pass an output path to command plugins

abertelrud:
Add the ability to pass an output path to command plugins.  This output path is given to the plugin and is added to the sandbox as a place to which it can write.

This is a proposed amendment to the SE-0332 proposal currently being reviewed.

### Motivation:

Like all plugins, command plugins are run in a sandbox that only provides write access to temporary directories.  But it's quite common for a plugin to generate output, so there needs to be a way for a user to provide the path to the command and for SwiftPM to add that as a writable path to the sandbox.  This path can be either a directory or a file — it's up to the plugin to decide what it creates at that path.

### Modifications:

- extend the plugin protocol to have a `outputPath` parameter
- add compatibility glue to allow existing plugins to continue working
- add the optional output path field to the transport structure to be sent to the plugin
- add the `--outputPath` option to the `swift package plugin` subcommand
- extend unit tests

### Result:

Command plugins can access the optional `outputPath` parameter, and have write permission to it.

### DIFF
--- a/Sources/PackagePlugin/Protocols.swift
+++ b/Sources/PackagePlugin/Protocols.swift
@@ -99,13 +99,49 @@ public protocol CommandPlugin: Plugin {
         arguments: [String]
     ) throws
 
+    /// Invoked by SwiftPM to perform the custom actions of the command.
+    func performCommand(
+        /// The context in which the plugin is invoked. This is the same for all
+        /// kinds of plugins, and provides access to the package graph, to cache
+        /// directories, etc.
+        context: PluginContext,
+
+        /// The targets to which the command should be applied. If the invoker of
+        /// the command has not specified particular targets, this will be a list
+        /// of all the targets in the package to which the command is applied.
+        targets: [Target],
+
+        /// Any literal arguments passed after the verb in the command invocation.
+        arguments: [String],
+
+        /// Optional output path to which the command is allowed to write.
+        outputPath: Path?
+    ) throws
+
     /// A proxy to the Swift Package Manager or IDE hosting the command plugin,
     /// through which the plugin can ask for specialized information or actions.
     var packageManager: PackageManager { get }
 }
 
 extension CommandPlugin {
-    
+
+    public func performCommand(
+        context: PluginContext,
+        targets: [Target],
+        arguments: [String]
+    ) throws {
+        try self.performCommand(context: context, targets: targets, arguments: arguments, outputPath: .none)
+    }
+
+    public func performCommand(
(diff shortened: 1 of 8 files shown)


abertelrud:
@swift-ci please smoke test

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -576,7 +576,7 @@ fileprivate extension FileHandle {
     func writePluginMessage(_ message: HostToPluginMessage) throws {
         // Encode the message as JSON.
         let payload = try JSONEncoder().encode(message)
```
tomerd, line 578: unrelated to this PR, but should the JSONEncoder be shared across call instead of created every time?

Sources/PackagePlugin/Protocols.swift
```diff
@@ -99,13 +99,49 @@ public protocol CommandPlugin: Plugin {
...
+
+    public func performCommand(
+        context: PluginContext,
+        targets: [Target],
+        arguments: [String],
+        outputPath: Path?
+    ) throws {
+        try self.performCommand(context: context, targets: targets, arguments: arguments)
+    }
+
```
tomerd, line 144: since this is a new API, do we need both?

tomerd (approved):
lgtm, some small questions

neonichu (approved):

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: It's pretty lightweight, but it is a class, so yes, it would be better to share it as it's at least one heap allocation.  From the implementation it doesn't look as if state can bleed from one encode to the next, which was my main concern.  I'll make a note to address this in another PR.

abertelrud: Also, for context, the number of messages for a typical plugin is anywhere from two to half a dozen, a bit more if there are diagnostics or a lot of build commands defined.  So it would matter but probably not measurably.  Should still be fixed.  Thanks for pointing it out.

Sources/PackagePlugin/Protocols.swift
abertelrud: Yes, for a time, since I didn't want to break things like the repository of sample code that matches what's in the proposal.  It should be removed in final 5.6 though.

abertelrud:
@swift-ci please smoke test

PR #3920, opened 2021-12-08
Simplify the adding of temporary directories to the sandbox profile when strictness is 'writableTemporaryDirectory'

abertelrud:
Specifically, we add regular and the Foundation temporary directories, but not anything passed in through the environment.

### Motivation:

This reduces the amount of magic and unpredictability regarding what's in the profile.  Client code that wants to add additional directories can do so using the array of writable paths that's already supported.

### Modifications:

- change the sandbox profile creation to just add `"/tmp"` and `NSTemporaryDirectory()`

### DIFF
--- a/Sources/Basics/Sandbox.swift
+++ b/Sources/Basics/Sandbox.swift
@@ -74,14 +74,9 @@ fileprivate func macOSSandboxProfile(
     }
     // Optionally allow writing to temporary directories (a lot of use of Foundation requires this).
     else if strictness == .writableTemporaryDirectory {
-        // Add the standard and Foundation temporary directories, and the one determined by TSC (which also taked into account environment variables).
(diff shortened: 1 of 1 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
This is a refinement of https://github.com/apple/swift-package-manager/pull/3918

tomerd (approved):

zoza2244:
> Specifically, we add regular and the Foundation temporary directories, but not anything passed in through the environment.
> 
> ### Motivation:
> 
> This reduces the amount of magic and unpredictability regarding what's in the profile.  Client code that wants to add additional directories can do so using the array of writable paths that's already supported.
> 
> ### Modifications:
> 
> - change the sandbox profile creation to just add `"/tmp"` and `NSTemporaryDirectory()`
>

PR #3929, opened 2021-12-10
optimize manifest requests for registry based depedendcies

tomerd:
motivation: improve performance by reducing redundant http calls

changes:
* save main manifest on first request
* cache known manifest across calls in RegistryPackageContainer
* add RegistryPackageContainer tests
* fix a bug in workspace where custom tools version is not propogated to the toolsLoader

rdar://86218181

### DIFF
--- a/Sources/Workspace/Workspace.swift
+++ b/Sources/Workspace/Workspace.swift
@@ -290,7 +290,7 @@ public class Workspace {
     ) throws {
         // defaults
         let currentToolsVersion = customToolsVersion ?? ToolsVersion.currentToolsVersion
-        let toolsVersionLoader = ToolsVersionLoader()
+        let toolsVersionLoader = ToolsVersionLoader(currentToolsVersion: currentToolsVersion)
         let manifestLoader = try customManifestLoader ?? ManifestLoader(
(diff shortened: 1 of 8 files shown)


tomerd:
@swift-ci please smoke test

Sources/Workspace/Workspace.swift
```diff
@@ -290,7 +290,7 @@ public class Workspace {
     ) throws {
         // defaults
         let currentToolsVersion = customToolsVersion ?? ToolsVersion.currentToolsVersion
-        let toolsVersionLoader = ToolsVersionLoader()
+        let toolsVersionLoader = ToolsVersionLoader(currentToolsVersion: currentToolsVersion)
```
tomerd, line 293: customToolsVersion should only be used in tests, we should probably  remove it from the public constructor and move it into an internal one

yim-lee (approved):

Sources/Workspace/Workspace.swift
tomerd: addressed in https://github.com/apple/swift-package-manager/pull/3933

PR #3931, opened 2021-12-10
preserve old formatting of Package.resolved in tools-version < 5.6

tomerd:
motivation: We moved to JSONEncoder in 5.6 which formats JSON in a non-traditional way (add space around the field name colon), and different from previous version of SwiftPM. We can make the change tools-version dependent.

changes: user older JSON encoder when tools-version < 5.6, preserving previous format

rdar://86341850


tomerd:
@swift-ci please smoke test

abertelrud (approved):

PR #3937, opened 2021-12-11
Disable TOFU feature temporarily

yim-lee:
It continues to cause failures in the source compatibility test.


yim-lee:
@swift-ci please smoke test

tomerd (approved):

PR #3946, opened 2021-12-15
Actually remove the `--output` option from `swift package plugin`

abertelrud:
A previous commit removed the parameter to the plugin invocation and made the option not do anything, but this removes it from `swift package plugin` as well.

### Motivation:

This is to align with SE-0332.

### Modifications:

- remove the command line option (it was still present but didn't do anything)


abertelrud:
@swift-ci please smoke test

neonichu (approved):

tomerd (approved):

PR #3951, opened 2021-12-16
make tests that check stderr more robust

tomerd:
motivation: broken tests when warnings are present in stderr

changes: use XCTAssertMatch and check for error instead of straight comparison


tomerd:
@swift-ci please smoke test

yim-lee (approved):

PR #3940, opened 2021-12-13
[5.6] Load manifests asynchronously [cherry pick]

pcbeard:
This is a cherry pick of PR #3894 on to the release/5.6 branch.

PR #3954, opened 2021-12-17
reduce concurrent manifest loading test contention

tomerd:
motivation: with manfiest loading async, the concurrent loading tests introduce significant contention on cpu since the underlying queue in unbounded

changes: use semaphore to limit the concurrency to max active cpus


tomerd:
cc @pcbeard

tomerd:
@swift-ci please smoke test

abertelrud (approved):

abertelrud:
This only affects `main`, is that correct?  (not 5.6)

tomerd:
> This problem being fixed here only affects main, is that correct? (not 5.6)

at this point yes, #3940 will bring async manifest loading into 5.6

tomerd:
@pcbeard should we pull this into the #3940 cherry-pick?

PR #3955, opened 2021-12-17
move manifest validation that requires source control to Workspace

tomerd:
motivation: all access to git operations should be done via the repositoriesProvider handed to the workspace, but some was hard wired to using Git

changes:
* move git related validation from manfeist JSON parser and graph loader to workspace and use the workspace repositoryProvider to perform them
* add methods on RepositoryProvider to hande validation of directory and ref formats.
* remove PackageLoading depedency on SourceControl which was a problem for a while and only needed for these validations
* add and adjust tests

rdar://84393257


tomerd:
@swift-ci please smoke test

abertelrud (approved):
LGTM

PR #3957, opened 2021-12-18
Errors thrown at top level of plugin should be reported as diagnostics

abertelrud:
When a plugin throws an uncaught error, it should be reported as a diagnostic rather than causing the standard Swift output.

### Motivation:

This is a better user experience than the very technical output that is shown otherwise.

### Details

As part of making this work well, we also simplify the way the result code is received from the plugin.  It was a bit awkward to have two result codes (the one from the last message as well as the exit code from the process) so now we just use the exit code.  This is also most robust to any plugin that decides to do "exit(1)" on its own from inside the code.  If/when we have long-running multi-action plugins we can revisit this.

We also make sure that at least one error message is emitted from any plugin that exits with an error code but that does not emit an error diagnostic.

This also reworks the unit test for command invocations to be more scalable.

### Modifications:

- catch errors thrown by plugins and report them as error diagnostics
- get rid of the plugin return message with the separate bool and use the exit code of the process; this is also robust against plugins that decide to do `exit(1)` instead of treating that as internal error
- make sure there is a failure diagnostic when the plugin returns an error code but didn't emit an error
- significantly rework the unit test for command plugins — it should be factored out into separate tests but now at least is much more extendable within the test

### DIFF
--- a/Sources/PackagePlugin/Plugin.swift
+++ b/Sources/PackagePlugin/Plugin.swift
@@ -92,7 +92,15 @@ extension Plugin {
         // Handle messages from the host until the input stream is closed,
         // indicating that we're done.
         while let message = try pluginHostConnection.waitForNextMessage() {
-            try handleMessage(message)
+            do {
+                try handleMessage(message)
+            }
+            catch {
+                // Emit a diagnostic and indicate failure to the plugin host,
+                // and exit with an error code.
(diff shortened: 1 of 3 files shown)


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #3964, opened 2021-12-22
Fix an ordering issue of a print statement and a diagnostic arriving right at the exit of a plugin

abertelrud:
There's no strong ordering between the free form output of a plugin and the diagnostics it emits, but for things like thrown exceptions etc, it makes a bit more sense to show the diagnostic last.

Perhaps the test shouldn't really depend on the order here — we will see if this addresses the issue in the CI (it seems specific to Linux and I was never able to reproduce it in Docker).

### Motivation:

Fix a flaky test in CI, but also to make the order of output from the end of the plugin execution make a little more sense when it exits by throwing an error.

### Modifications:

- after process termination, read the last text output before any diagnostics
- I also made the unit test text a little less whimsical, and got rid of the emoji annotating the plugin output, and added printing out of the received diagnostics

rdar://86729106

### DIFF
--- a/Sources/Workspace/DefaultPluginScriptRunner.swift
+++ b/Sources/Workspace/DefaultPluginScriptRunner.swift
@@ -485,11 +485,11 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
             // Close the output handle through which we talked to the plugin.
             try? outputHandle.close()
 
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

tomerd (approved):

neonichu (approved):

PR #3967, opened 2021-12-24
make package.resolved more deterministic in edge cases

tomerd:
motivation: in some cases like two URLs with casing differences, package.resolved could be non-deterministic

changes:
* base managed dependencies lookup on canonical location (case insensitive, ignore .git suffix)
* refactor how package.resolved is saved to test for non material differences
* adjust call sites that manipulate the pin store to be more efficient about when to save the resolved file
* add warning when expected dependency that should be stored in the resolved file is not found
* adjust call sites
* add and adjust tests


tomerd:
@swift-ci smoke test

tomerd:

Sources/Workspace/Workspace.swift
```diff
@@ -1326,25 +1326,86 @@ extension Workspace {
...
+            } else {
+                observabilityScope.emit(warning: "required dependency \(dependency.identity) (\(dependency.locationString)) was not found in managed dependencies and will not be recorded in resolved file")
+            }
+        }
+
+        // FIXME
+        // load the pin store from disk so we can compare for any changes
+        // this is needed as we want to avoid re-writing the resolved files
+        // unless absolutely required
+        guard let storedPinStore = observabilityScope.trap({ try self.pinsStore.load() }) else {
```
abertelrud, line 1350: So if this fails, such as because of an I/O error, would it be better to try to write the new file anyway?  I guess I'm questioning whether returning here is the right thing.

tomerd: are you suggesting we try rewrite if we cant read? I can go with that too

abertelrud: Yes, exactly (I should have been more clear).  Since the only reason we're reading it is to avoid unnecessary changes, it seems reasonable to write it out if the read fails.  Of course, if it's a permission issue, the write may still fail, but that's a separate issue.

tomerd: updated https://github.com/apple/swift-package-manager/pull/3967/commits/e2375162f06a66a1e0cd91b11aecd67e2520d8e0

tomerd:
@swift-ci please smoke test

tomerd:

neonichu (approved):

abertelrud (approved):

Sources/Workspace/Workspace.swift
```diff
@@ -1326,25 +1326,61 @@ extension Workspace {
...
+                        break
+                    }
+                }
             }
+        } else {
+            needsUpdate = true
         }
 
-        observabilityScope.trap{
+        // exist early is there is nothing to do
```
abertelrud, line 1372:
```suggestion
        // exit early if there is nothing to do
```

PR #3970, opened 2022-01-04
Mark `Platform.openbsd` as available since 5.6

abertelrud:
Mark `Platform.openbsd` as available since 5.6.

### Motivation:

Allows this new enum to be used in packages marked with 5.6.

### Modifications:

- Change 999.0 to 5.6 availability and add a CHANGELOG entry.

This will also be nominated to the `release/5.6` branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ Note: This is in reverse chronological order, so newer entries are added to the 
 
 Swift 5.6
 -----------
+* `openbsd` has been added to the set of possible platform values for packages that specify a 5.6 tools version.
+
 * [#3649]
 
   Semantic version dependencies can now be resolved against Git tag names that contain only major and minor version identifiers.  A tag with the form `X.Y` will be treated as `X.Y.0`. This improves compatibility with existing repositories.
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci smoke test

tomerd (dismissed):

tomerd (commented):
what level of support do we offer for this platform? is actually fully supported?

tomerd (commented):
what level of support do we offer for this platform? is actually fully supported?

abertelrud:
That's a good question — I will take a look at the original PR.  I just noticed it as being 999.0 when changing the command plugin APIs, but we should leave as 999.0 if there isn't actually any support.

abertelrud:
Looks like this was introduced in https://github.com/apple/swift-package-manager/pull/3572, where the statement is that SwiftPM now builds on OpenBSD.  But since that's not an officially supported platform AFAIK, probably the right thing is to leave it as an experimental feature at 999.0 (which means SwiftPM 5.6 won't support it, but main toolchains will).

abertelrud:
Closing this PR, leaving the annotation as 999.0.

PR #3971, opened 2022-01-04
Don't emit "unhandled files" warnings for files that are handled by build commands from package plugins

abertelrud:
Filter out source paths for which build tool commands were created by plugins and only emit warnings for any remaining paths.

### Motivation:

Source files for which build tool plugins emit commands are actually handled, and we shouldn’t generate warnings for them.

### Modifications:

- move the logic to emit warnings from TargetSourceBuilder until after the plugins have run
- filter out any paths that appear as input dependencies in build commands generated by plugins

In the future we might consider more overt ways for plugins to express what types of files should be considered handled by the plugin.

Note that libSwiftPM clients will need to emit these diagnostics after they run the plugins.

rdar://74663980


abertelrud:
@swift-ci smoke test

Sources/Build/BuildOperation.swift
```diff
@@ -209,6 +209,36 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
...
+                
+                // Subtract out any that were inputs to any commands generated by plugins.
+                if let result = buildToolPluginInvocationResults[target] {
+                    let handledFiles = result.flatMap{ $0.buildCommands.flatMap{ $0.inputFiles } }
+                    unhandledFiles.subtract(handledFiles)
+                }
+                if unhandledFiles.isEmpty { continue }
+                
+                // Emit a diagnostic if any remain. This is kept the same as the previous message for now, but this could be improved.
+                let diagnosticsEmitter = self.observabilityScope.makeDiagnosticsEmitter {
```
abertelrud, line 228: This preserves the same metadata as now.

Sources/Build/BuildOperation.swift
```diff
@@ -209,6 +209,36 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
         let prebuildCommandResults = try graph.reachableTargets.reduce(into: [:], { partial, target in
             partial[target] = try buildToolPluginInvocationResults[target].map { try self.runPrebuildCommands(for: $0) }
         })
+
+        // Emit warnings about any unhandled files in authored packages. We do this after applying build tool plugins, once we know what files they handled.
+        for package in graph.rootPackages where package.manifest.toolsVersion >= .v5_3 {
```
abertelrud, line 214: This applies only to root packages as today.

Sources/PackageModel/Target.swift
```diff
@@ -452,6 +453,7 @@ public final class ClangTarget: Target {
             type: type,
             sources: sources,
             resources: resources,
+            others: others,
```
abertelrud, line 456: This passes the list of others through for Clang targets as well, which wasn’t being done before.  Warnings were already being emitted for unknown files in Clang targets, but now that we’re doing it later, we have to pass them through (which is correct anyway, otherwise plugins wouldn’t see them for example).  I haven’t found other side effects with this.

Tests/CommandsTests/PackageToolTests.swift
```diff
@@ -1080,23 +1090,30 @@ final class PackageToolTests: CommandsTestCase {
...
+                            // Create and return a build command that uses all the `.foo` files in the target as inputs, so they get counted as having been handled.
+                            let fooFiles = (target as? SourceModuleTarget)?.sourceFiles.compactMap{ $0.path.extension == "foo" ? $0.path : nil } ?? []
+                            return [ .buildCommand(displayName: "A command", executable: "/bin/echo", arguments: ["Hello"], inputFiles: fooFiles) ]
                         }
 
-                        enum Error : Swift.Error {
-                        case inconsistentDisplayName(String)
-                        }
                     }
+                    extension String : Error {}
```
abertelrud, line 1103: This just makes it a bit more convenient to throw errors from the test.

tomerd (approved):

tomerd:
nice

PR #3972, opened 2022-01-04
Support `--allow-writing-to-package-directory` as described in SE-0332 and `--allow-writing-to-directory` to allow arbitrary directories

abertelrud:
### Motivation:

This adds an option that was spelled out in the proposal.  Also adds an additional useful option to specify arbitrary directories to allow the plugin to write to, which is helpful for plugins that have their own output arguments.

### Modifications:

- adds `--allow-writing-to-package-directory` as described in SE-0332
- also adds `--allow-writing-to-directory`, which takes a relative or absolute path as an argument
- also adds `--block-writing-to-temporary-directory` to provide control over whether the temporary directories can be written to

It also adds in the error to message that is shown if writing to the package directory isn't allowed, which includes the stated reason the plugin provided for why it needs this access.

rdar://86786404

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -886,6 +886,18 @@ extension SwiftPackageTool {
               help: "List the available plugin commands")
         var listCommands: Bool = false
 
+        @Flag(name: .customLong("allow-writing-to-package-directory"),
+              help: "Allow the plugin to write to the package directory")
+        var allowWritingToPackageDirectory: Bool = false
+
+        @Option(name: .customLong("allow-writing-to-directory"),
+              help: "Allow the plugin to write to an additional directory")
+        var additionalAllowedWritableDirectories: [String] = []
+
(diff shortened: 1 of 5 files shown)


abertelrud:
@swift-ci please smoke test

neonichu (approved):

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -969,6 +978,31 @@ extension SwiftPackageTool {
...
             let outputDir = pluginsDir.appending(component: "outputs")
 
+            // Determine the set of directories under which plugins are allowed to write. We always include the output directory.
+            var writableDirectories = [outputDir]
+            if allowWritingToPackageDirectory {
+                writableDirectories.append(package.path)
+            }
+            if !blockWritingToTemporaryDirectory {
+                writableDirectories.append(AbsolutePath("/tmp"))
+                writableDirectories.append(AbsolutePath(NSTemporaryDirectory()))
```
tomerd, line 988: I think we recently added tempDirectory to FileSystem, maybe we can use it here?

PR #3974, opened 2022-01-05
gracefully fail workspace state loading when duplicate entries found

tomerd:
motivation: some older versions of SwiftPM wrote duplicate entries into workspace-state.json which can cause a crash due to use of Dictionary::uniquingKeysWith

changes:
* do not use Dictionary::uniquingKeysWith for non-santized input such as workspace state
* add tests

rdar://86857825


tomerd:
@swift-ci please smoke test

abertelrud (approved):

PR #3977, opened 2022-01-05
Allow package plugins to be invoked using just `swift package`, and not require `swift package plugin`.

abertelrud:
Allow package plugins to be invoked using just `swift package`, and not require `swift package plugin`.  The `plugin` subcommand remains, especially for things like `--list` and future plugin-specific subcommands.

### Motivation:

This is part of the SE-0332 proposal.  Requiring the `plugins` spelling was always temporary.

### Modifications:

- set `plugin` as the default subcommand for `swift package`
- fix the error messages to be clearer, and to use the swift-argument-parser `ValidationError` where appropriate (this also emits the help text)
- extend unit test to check that both with `plugin` and without `plugin` spelling work

### Result:

Invoking `swift package foo` will fall back on `swift package plugin foo` if `foo` is not a built-in command.

rdar://87077988


abertelrud:
@swift-ci please test

tomerd (approved):

abertelrud:
@swift-ci please smoke test

PR #3983, opened 2022-01-07
Don't prefix each line of command plugin output with the name of the plugin

abertelrud:
### Motivation:

It is noisy and makes the plugin commands look very different from the built in commands, which is opposite of the long-term goal.

### Modifications:

- change the output string to not include the prefix

rdar://86855690


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #3984, opened 2022-01-07
Include the default command search paths for plugin tool lookup

abertelrud:
### Motivation:

This makes it possible for tools to use regular commands expected to be in `/usr/bin` etc and which should appropriate to embed as an artifact archive in a package.

### Modifications:

- add the standard set of search paths to the list we already pass down to the plugin (and which includes the toolchain etc)


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #3985, opened 2022-01-07
fix dependency resolution edge cases issues

tomerd:
motivation: dependency resolution could be incorrect in some edge cases when a pre-release requirement is specified, and not report the issue to the user

changes:
* depedency resolustion should not assume 0.0.0 is the correct minimal requirment, instead it should use the specified version
* fix dependency term computation which used incorrect version requirment for the the parent node
* add validation at the end of package udpate API to make sure no missing packages exist after resolution, and fail the operation if such exists. this mirrors the behavior of package resolve API

### DIFF
--- a/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
+++ b/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
@@ -1216,41 +1216,47 @@ internal final class PubGrubPackageContainer {
 
             // Since the pinned version is most likely to succeed, we don't compute bounds for its
             // incompatibilities.
-            return try Array(constraints.map { (constraint: PackageContainerConstraint) -> [Incompatibility] in
-                guard case .versionSet(let vs) = constraint.requirement else {
+            return try constraints.flatMap { constraint -> [Incompatibility] in
+                // We only have version-based requirements at this point.
+                guard case .versionSet(let constraintRequirement) = constraint.requirement else {
                     throw InternalError("Unexpected unversioned requirement: \(constraint)")
                 }
                 return try constraint.nodes().map { dependencyNode in
                     var terms: OrderedSet<Term> = []
+                    // the package version requirement
                     terms.append(Term(node, .exact(version)))
-                    terms.append(Term(not: dependencyNode, vs))
+                    // the dependency's version requirement
+                    terms.append(Term(not: dependencyNode, constraintRequirement))
                     return try Incompatibility(terms, root: root, cause: .dependency(node: node))
                 }
-            }.joined())
+            }
         }
 
-        let (lowerBounds, upperBounds) = try self.computeBounds(for: node,
-                                                                constraints: constraints,
-                                                                startingWith: version)
+        let (lowerBounds, upperBounds) = try self.computeBounds(
+            for: node,
+            constraints: constraints,
+            startingWith: version
+        )
 
-        return try constraints.flatMap { constraint in
+        return try constraints.flatMap { constraint -> [Incompatibility] in
+            // We only have version-based requirements at this point.
+            guard case .versionSet(let constraintRequirement) = constraint.requirement else {
+                throw InternalError("Unexpected unversioned requirement: \(constraint)")
+            }
             return try constraint.nodes().map { constraintNode in
                 var terms: OrderedSet<Term> = []
-                let lowerBound = lowerBounds[constraintNode] ?? "0.0.0"
-                let upperBound = upperBounds[constraintNode] ?? Version(version.major + 1, 0, 0)
-                assert(lowerBound < upperBound)
-
-                // We only have version-based requirements at this point.
-                guard case .versionSet(let vs) = constraint.requirement else {
-                    throw InternalError("Unexpected unversioned requirement: \(constraint)")
-                }
-
-                let requirement: VersionSetSpecifier = .range(lowerBound ..< upperBound)
-                terms.append(Term(node, requirement))
-                terms.append(Term(not: constraintNode, vs))
 
                 // Make a record for this dependency so we don't have to recompute the bounds when the selected version falls within the bounds.
-                self.emittedIncompatibilities[constraintNode] = requirement.union(emittedIncompatibilities[constraintNode] ?? .empty)
+                if let lowerBound = lowerBounds[constraintNode], let upperBound = upperBounds[constraintNode] {
+                    assert(lowerBound < upperBound)
+                    let requirement: VersionSetSpecifier = .range(lowerBound ..< upperBound)
+                    self.emittedIncompatibilities[constraintNode] = requirement.union(emittedIncompatibilities[constraintNode] ?? .empty)
+                }
+
+                // the package version requirement
+                terms.append(Term(node, .exact(version)))
+                // the dependency's version requirement
+                terms.append(Term(not: constraintNode, constraintRequirement))
 
                 return try Incompatibility(terms, root: root, cause: .dependency(node: node))
             }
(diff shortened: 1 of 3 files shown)


tomerd:
@swift-ci please smoke test

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
```diff
@@ -1216,41 +1216,47 @@ internal final class PubGrubPackageContainer {
...
-                let requirement: VersionSetSpecifier = .range(lowerBound ..< upperBound)
-                terms.append(Term(node, requirement))
-                terms.append(Term(not: constraintNode, vs))
-
-                // Make a record for this dependency so we don't have to recompute the bounds when the selected version falls within the bounds.
-                self.emittedIncompatibilities[constraintNode] = requirement.union(emittedIncompatibilities[constraintNode] ?? .empty)
+                // the package version requirement
+                terms.append(Term(node, .exact(version)))
+                // the dependency's version requirement
+                terms.append(Term(not: constraintNode, constraintRequirement))
```
tomerd, line 1259: this is the heart of the change, there a couple of things going on here that seem wrong:

1: `requirement` is used for two different purposes: to record the dependency version (via ` self.emittedIncompatibilities`) but also as the resolved version for `node` (via `terms.append(Term(node, requirement))`), which iiuc does not really make sense as it mixed the dependency (`constraintNode`) version with the package (`node`) version.

2.  `requirement` is looked up on the bounds computed for the dependencies using `constraintNode` but if not found, defaults to 0.0.0 < version(major+1, 0, 0). the defaulting is wrong in two ways: it does not work for pre-release versions since it strips the rep-release extensions and it is based on `version` which is the version of `node` not `constraintNode`

the changes is to always set the `node` term to the exact version requested which works for pre-release and non-pre-release. it does however break a bunch of tests that recorded the 0.0.0 <  version(major+1, 0, 0) behavior, but maybe that is okay since they are testing for wrong behavior?

@SDGGiesbrecht very interested in your 👀 on this, as you are probably more familiar with this code / logic than I am. I actually hope I am wrong and this is not as wrong as I think it is :D

Sources/Workspace/Workspace.swift
```diff
@@ -793,7 +793,9 @@ extension Workspace {
         guard let pinsStore = observabilityScope.trap({ try self.pinsStore.load() }) else { return nil }
 
         // Ensure we don't have any error at this point.
-        guard !observabilityScope.errorsReported else { return nil }
+        guard !observabilityScope.errorsReported else {
+            return nil
+        }
```
tomerd, line 798: formatting only

Sources/Workspace/Workspace.swift
```diff
@@ -824,17 +826,28 @@ extension Workspace {
...
 
         if dryRun {
-            return observabilityScope.trap { return try computePackageStateChanges(root: graphRoot, resolvedDependencies: updateResults, updateBranches: true, observabilityScope: observabilityScope) }
+            return observabilityScope.trap {
+                return try self.computePackageStateChanges(root: graphRoot, resolvedDependencies: updateResults, updateBranches: true, observabilityScope: observabilityScope)
+            }
```
tomerd, line 836: formatting only

Sources/Workspace/Workspace.swift
```diff
@@ -824,17 +826,28 @@ extension Workspace {
...
 
         // Load the updated manifests.
         let updatedDependencyManifests = try self.loadDependencyManifests(root: graphRoot, observabilityScope: observabilityScope)
+        // If we have missing packages, something is fundamentally wrong with the resolution of the graph
+        let stillMissingPackages = updatedDependencyManifests.computePackages().missing
+        guard stillMissingPackages.isEmpty else {
+            let missing = stillMissingPackages.map{ $0.description }
+            observabilityScope.emit(error: "exhausted attempts to resolve the dependencies graph, with '\(missing.joined(separator: "', '"))' unresolved.")
+            return nil
+        }
```
tomerd, line 850: added validation as `package update` could fail silently without this check

Sources/Workspace/Workspace.swift
```diff
@@ -1715,7 +1728,7 @@ extension Workspace {
             let dependenciesToLoad = dependenciesRequired.map{ $0.createPackageRef() }.filter { !loadedManifests.keys.contains($0.identity) }
             let dependenciesManifests = try temp_await { self.loadManagedManifests(for: dependenciesToLoad, observabilityScope: observabilityScope, completion: $0) }
             dependenciesManifests.forEach { loadedManifests[$0.key] = $0.value }
-            return pair.item.dependenciesRequired(for: pair.key.productFilter).compactMap { dependency in
+            return dependenciesRequired.compactMap { dependency in
```
tomerd, line 1731: small cleanup as `dependenciesRequired` is already defined above as `pair.item.dependenciesRequired(for: pair.key.productFilter)`

Tests/PackageGraphTests/PubgrubTests.swift
```diff
@@ -496,7 +496,7 @@ final class PubgrubTests: XCTestCase {
         ])
     }
 
-    func testCycle1() {
+    func _testCycle1() {
```
tomerd, line 499: this test crashes now on a precondition and will be fixed together with the other failing tests if this change is good

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
```diff
@@ -1216,41 +1216,47 @@ internal final class PubGrubPackageContainer {
...
                     var terms: OrderedSet<Term> = []
+                    // the package version requirement
                     terms.append(Term(node, .exact(version)))
-                    terms.append(Term(not: dependencyNode, vs))
+                    // the dependency's version requirement
+                    terms.append(Term(not: dependencyNode, constraintRequirement))
                     return try Incompatibility(terms, root: root, cause: .dependency(node: node))
                 }
-            }.joined())
+            }
```
tomerd, line 1232: bit of cleanup, uses flatMap instead of creating an array and then joining it

Sources/Workspace/Workspace.swift
abertelrud: This is a great improvement too — the silent failures cause so much confusion.

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
abertelrud: I don't have much knowledge here, but I would not be surprised if some of the tests are conceptually wrong and just verifying the implemented behavior.  So I wouldn't be concerned about having to change the tests.

SDGGiesbrecht: Taking a closer look now. I will get back to you.

SDGGiesbrecht: I believe both your observations are correct, @tomerd, and same goes for your solutions.

Whatever situation used to hit the defaults will now result in nothing recorded into `emittedIncompatibilities`. At first that concerned me, but after tracking where that leads, it feels safe, even if a little odd. The only two times `emittedIncompatibilities` is read from are:
- To skip processing a node that has already been handled. Such nodes will be re‐processed now, but I don’t think that would be a problem.
- To not trust a pin if we have already encountered anything odd. This too only prevents a shortcut, as the whole graph will be traversed either way, just on the one hand not necessarily trying the most probable solution first when obvious, or on the other hand not preloading alternatives when overestimating the probability that a solution will pan out.

I would be curious to know _how_ the defaults were reachable in the first place, and if it could be done from the wild or only by mocked tests that bypass earlier checks.

> I actually hope I am wrong and this is not as problematic as I think it is :D

Yes, that code appears to have been embarrassingly invalid. I think it snuck by everyone for so long because without the proper incompatibility here, it would just fail to record a short‐circuit opportunity and merely be slower, and because the nonsensical extra incompatibility would usually end up so far afield that it would be completely disjoint from the actual problem space.

tomerd:
@swift-ci please smoke test

tomerd:

tomerd:

abertelrud (approved):

PR #3989, opened 2022-01-08
Update CHANGELOG to describe manifest API changes

tomerd:
motivation: document user visible changes in change log

changes: Update CHANGELOG to include deprecation of dependency requirement enum calling convention


abertelrud (approved):
Yay!  I'm so happy about this change actually.

PR #3992, opened 2022-01-08
avoid deprecation warnings from test discovery

tomerd:
motivation: do not prduce warning when tests are marked as deprecated, since they are designed to test deprecated functionality

changes:
* mark generated test-discovery wrappers as psuedo-deprecated so they dont produce deprecation warnings
* use @main for the test discovery entry point instead of main.swift which helps avoid deprecation warnings at the entry point
* cleanup test discovery generate  names to be easier to reason about
* add and adjust tests for throwing variants
* add test for deprecation use case

### DIFF
--- a/Fixtures/Miscellaneous/TestDiscovery/Async/Tests/AsyncTests/SwiftTests.swift
+++ b/Fixtures/Miscellaneous/TestDiscovery/Async/Tests/AsyncTests/SwiftTests.swift
@@ -7,6 +7,9 @@ class AsyncTests: XCTestCase {
     func testAsync() async {
     }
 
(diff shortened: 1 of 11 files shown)


tomerd:
@swift-ci please smoke test

tomerd:
@Lukasa iirc you filed a jira/radar on this? could not find it! do you happen to have a link so I can associate with the fix?

abertelrud (approved):

Lukasa:
Sorry for the delay, @weissi filed the issue: https://bugs.swift.org/browse/SR-12008.

PR #3993, opened 2022-01-08
Make the symbol graph extraction requested by plugins more robust

abertelrud:
This includes reworking the SymbolGraphExtract type that so it's more suitable for being called in the context of plugins.

### Motivation:

The symbol graph extraction support for plugins called into the existing SymbolGraphExtract type, but this type was more suited to use from the command line action, and this caused some bugs for its use from plugins.

This PR refactors SymbolGraphExtract to hoist the assumptions about things like destination path out into the caller, and adjusts the call sites.

### Modifications

- factor out assumptions and console output from SymbolGraphExtract to the caller
- when used for plugins, avoid generate symbol graph information for targets other than the one asked for
- use a unique symbol graph output directory for each package/target combination
- add a unit test with better checking for different symbol graph generation usage
- adjust call sites for the `dump-symbol-graph` command as well as its use from plugins
- made verbose mode pass `-v` to `dump-symbol-graph` so it prints additional information too

### Results

Same behavior for `dump-symbol-graph` except that verbose mode is now also passed on to that tool, which is in the spirit of enabling verbose more.  For package plugin use, there are now separate output paths per combination of package and target, and only those symbol graphs that are asked for are emitted.

rdar://87234110&86785765

This will be nominated for inclusion in SwiftPM 5.6.

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -531,7 +531,7 @@ extension SwiftPackageTool {
     struct DumpSymbolGraph: SwiftCommand {
         static let configuration = CommandConfiguration(
             abstract: "Dump Symbol Graph")
-        static let defaultMinimumAccessLevel = AccessLevel.public
+        static let defaultMinimumAccessLevel = SymbolGraphExtract.AccessLevel.public
 
         @OptionGroup(_hiddenFromHelp: true)
         var swiftOptions: SwiftToolOptions
@@ -542,7 +542,7 @@ extension SwiftPackageTool {
         @Flag(help: "Skip members inherited through classes or default implementations.")
         var skipSynthesizedMembers = false
 
-        @Option(help: "Include symbols with this access level or more. Possible values: \(AccessLevel.allValueStrings.joined(separator: " | "))")
+        @Option(help: "Include symbols with this access level or more. Possible values: \(SymbolGraphExtract.AccessLevel.allValueStrings.joined(separator: " | "))")
         var minimumAccessLevel = defaultMinimumAccessLevel
 
         @Flag(help: "Skip emitting doc comments for members inherited through classes or default implementations.")
@@ -552,23 +552,35 @@ extension SwiftPackageTool {
         var includeSPISymbols = false
 
         func run(_ swiftTool: SwiftTool) throws {
-            let symbolGraphExtract = try SymbolGraphExtract(
-                tool: swiftTool.getToolchain().getSymbolGraphExtract())
-
             // Build the current package.
             //
             // We turn build manifest caching off because we need the build plan.
             let buildOp = try swiftTool.createBuildOperation(cacheBuildManifest: false)
             try buildOp.build()
 
-            try symbolGraphExtract.dumpSymbolGraph(
-                buildPlan: buildOp.buildPlan!,
-                prettyPrint: prettyPrint,
-                skipSynthesisedMembers: skipSynthesizedMembers,
+            // Configure the symbol graph extractor.
+            let symbolGraphExtractor = try SymbolGraphExtract(
+                tool: swiftTool.getToolchain().getSymbolGraphExtract(),
+                skipSynthesizedMembers: skipSynthesizedMembers,
                 minimumAccessLevel: minimumAccessLevel,
                 skipInheritedDocs: skipInheritedDocs,
-                includeSPISymbols: includeSPISymbols
-            )
+                includeSPISymbols: includeSPISymbols,
+                prettyPrintOutputJSON: prettyPrint)
+
+            // Run the tool once for every library and executable target in the root package.
+            let buildPlan = buildOp.buildPlan!
+            let symbolGraphDirectory = buildPlan.buildParameters.dataPath.appending(component: "symbolgraph")
+            let targets = buildPlan.graph.rootPackages.flatMap{ $0.targets }.filter{ $0.type == .library || $0.type == .executable }
+            for target in targets {
+                print("-- Emitting symbol graph for", target.name)
+                try symbolGraphExtractor.extractSymbolGraph(
+                    target: target,
+                    buildPlan: buildPlan,
+                    verbose: verbosity != .concise,
+                    outputDirectory: symbolGraphDirectory)
+            }
+
+            print("Files written to", symbolGraphDirectory.pathString)
         }
     }
 
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -564,8 +564,7 @@ extension SwiftPackageTool {
                 skipSynthesizedMembers: skipSynthesizedMembers,
                 minimumAccessLevel: minimumAccessLevel,
                 skipInheritedDocs: skipInheritedDocs,
-                includeSPISymbols: includeSPISymbols,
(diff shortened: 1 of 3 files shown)


abertelrud:
@swift-ci smoke test

abertelrud:
@swift-ci smoke test

Sources/Commands/SymbolGraphExtract.swift
```diff
@@ -1,97 +1,79 @@
...
-            if skipSynthesisedMembers { args.append("-skip-synthesized-members") }
-            if minimumAccessLevel != SwiftPackageTool.DumpSymbolGraph.defaultMinimumAccessLevel {
-                args += ["-minimum-access-level", minimumAccessLevel.rawValue]
-            }
-            if skipInheritedDocs { args.append("-skip-inherited-docs") }
-            if includeSPISymbols { args.append("-include-spi-symbols") }
-
-            print("-- Emitting symbol graph for", target.name)
-            try runTool(args)
+        try localFileSystem.createDirectory(outputDirectory, recursive: true)
```
tomerd, line 45: should this support abstract file system instead of assuming local?

Tests/CommandsTests/PackageToolTests.swift
```diff
@@ -1442,4 +1442,85 @@ final class PackageToolTests: CommandsTestCase {
             }
         }
     }
+
+    func testCommandPluginSymbolGraphCallbacks() throws {
+        try testWithTemporaryDirectory { tmpPath in
+            // Create a sample package with a library, and executable, and a plugin.
+            let packageDir = tmpPath.appending(components: "MyPackage")
+            try localFileSystem.writeFileContents(packageDir.appending(components: "Package.swift")) {
+                $0 <<< """
```
tomerd, line 1451: you can now pass string directly,  o need to go through the byte stream

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -552,23 +552,35 @@ extension SwiftPackageTool {
...
+            // Run the tool once for every library and executable target in the root package.
+            let buildPlan = buildOp.buildPlan!
+            let symbolGraphDirectory = buildPlan.buildParameters.dataPath.appending(component: "symbolgraph")
+            let targets = buildPlan.graph.rootPackages.flatMap{ $0.targets }.filter{ $0.type == .library || $0.type == .executable }
+            for target in targets {
+                print("-- Emitting symbol graph for", target.name)
+                try symbolGraphExtractor.extractSymbolGraph(
+                    target: target,
+                    buildPlan: buildPlan,
+                    verbose: verbosity != .concise,
```
tomerd, line 579: iirc we have been trying to get rid of using the global verbosity flag in favor of diagnostics severity level, and SwiftTool has that available already

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -552,23 +552,35 @@ extension SwiftPackageTool {
...
+            // Configure the symbol graph extractor.
+            let symbolGraphExtractor = try SymbolGraphExtract(
+                tool: swiftTool.getToolchain().getSymbolGraphExtract(),
+                skipSynthesizedMembers: skipSynthesizedMembers,
                 minimumAccessLevel: minimumAccessLevel,
                 skipInheritedDocs: skipInheritedDocs,
-                includeSPISymbols: includeSPISymbols
-            )
+                includeSPISymbols: includeSPISymbols,
+                prettyPrintOutputJSON: prettyPrint)
```
tomerd, line 568: should this be an enumeration instead? so output: .json(pretty: true)

tomerd (approved):
few suggestions

Sources/Commands/SwiftPackageTool.swift
tomerd: SwiftTool::logLevel basically, having verbose: SwiftTool::logLevel  <= .info

abertelrud: Sounds good.  In this case I just hoisted the condition from inside of SymbolGraphExtract up to the caller, but that solution sounds even better.

Tests/CommandsTests/PackageToolTests.swift
abertelrud: Great, I didn't realize.  All the other tests use the stream, but I can change this one over.

Sources/Commands/SymbolGraphExtract.swift
abertelrud: Since the tool only works locally there didn't seem to be much point, and could be misleading since only the local file system will work.

Sources/Commands/SwiftPackageTool.swift
abertelrud: Perhaps — what would the other enum cases be?  Or would it just be JSON for now?

It does sound cleaner... I tried to keep this diff bounded by just moving these properties from the function args to the struct properties to make it easier to configure, but this seems reasonable to change.

abertelrud: I really like this actually — it makes conceptual sense even if there is just one enum case for now.

abertelrud:
@swift-ci smoke test

tomerd (approved):

Tests/CommandsTests/PackageToolTests.swift
abertelrud: There seems to be a subtle difference, where for the block-and-stream version, the parent directory of the output file is automatically created, but for the others (string, bytes, etc) it's not.  I'm inclined to leave this as is just in the interesting of getting more bugs fixed and then deal with all the tests in a separate PR, possibly also modifying the other calls to at least optionally create the parent directories of the output file.

abertelrud:
Error is:
```
15:32:48 error: unspecified("toolchain is invalid: could not find swift-symbolgraph-extract at expected path /Users/buildnode/jenkins/workspace/swift-package-manager-PR-macos-smoke-test/branch-main/buildbot_incremental/swiftpm-macosx-x86_64/x86_64-apple-macosx/release/swift-symbolgraph-extract")
```
I'm going to have to consider the best of of fixing this one — it's only an issue during testing, but it would be better to have the tests always running rather than conditionalize on availability of this tool.  It looks like `UserToolchain` uses the path of the Swift compiler to determine the toolchain path (which is reasonable) but during smoke tests cannot find this tool there.

abertelrud:
@swift-ci smoke test

abertelrud:
> Error is:
> 
> ```
> 15:32:48 error: unspecified("toolchain is invalid: could not find swift-symbolgraph-extract at expected path /Users/buildnode/jenkins/workspace/swift-package-manager-PR-macos-smoke-test/branch-main/buildbot_incremental/swiftpm-macosx-x86_64/x86_64-apple-macosx/release/swift-symbolgraph-extract")
> ```
> 
> I'm going to have to think about the best way of fixing this one — it's only an issue during testing, but it would be better to have the tests always running rather than conditionalize on availability of this tool. It looks like `UserToolchain` uses the path of the Swift compiler to determine the toolchain path (which is reasonable) but during smoke tests cannot find this tool there.

I ended up making the test conditional on the presence of the `swift-symbolgraph-extract` tool.  There are cases in which we could find a different version of it in the installed toolchain, but as these test will find it in both the self tests and also in the full-build case, it seems reasonable to skip the test if the tool isn't available at all.

PR #3994, opened 2022-01-08
cleanup use test-discovery main file name

tomerd:
motivation: small follow up cleanup to #3992 which consolidates the use of a share definition for the test discovery main file name

changes: use the shared definition for the main file name used for test discovery everywhere


tomerd:
@swift-ci please smoke test

abertelrud (approved):

PR #3996, opened 2022-01-08
Enforce read-only-ness of package directory even when it's inside writable directories

abertelrud:
Enforce read-only-ness of package directory even when it's inside `/tmp` and other writable directories to the number of reduce special cases.

### Motivation

The package directory should always be read-only unless specifically requested to be writable, even if it happens to be in `/tmp` or other implicitly writable locations.

### Details

The sandbox supports a list of `writableDirectories`, which are treated as path component prefixes after package resolution.  The are applied to the sandbox rules after the default-deny and everything-is-read-only rules, carving out cones of writability in specific areas of the file system.

But if the package directory is under one of these directories that is allowed by default, then it is implicitly writable even in the absence `--allow-writing-to-package-directory`.  It also makes unit tests awkward, since they usually create temporary test fixtures inside the temporary directory.

### Modifications:

- extend `Sandbox` with a list of read-only locations
- add function and parameter documentation to Sandbox
- thread through the list of read only directories
- pass the package directory as a specific read-only directory unless it's covered by an explicit `writable-directory` parameter
- switch back to using `strictness` for temporary-directory writability (cleaner)
- remove the flag to make temporary directory read-only that was added mostly for unit tests

### Result:

There are fewer edge cases with respect to when the package directory is writable.

### DIFF
--- a/Sources/Basics/Sandbox.swift
+++ b/Sources/Basics/Sandbox.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See http://swift.org/LICENSE.txt for license information
@@ -13,13 +13,25 @@ import TSCBasic
 import TSCUtility
 
 public enum Sandbox {
+
+    /// Applies a sandbox invocation to the given command line (if the platform supports it),
+    /// and returns the modified command line. On platforms that don't support sandboxing, the
+    /// command line is returned unmodified.
+    ///
+    /// - Parameters:
+    ///   - command: The command line to sandbox (including executable as first argument)
+    ///   - strictness: The basic strictness level of the standbox.
+    ///   - writableDirectories: Paths under which writing should be allowed.
+    ///   - readOnlyDirectories: Paths under which writing should be denied, even if they would have otherwise been allowed by either the strictness level or paths in `writableDirectories`.
+
     public static func apply(
         command: [String],
+        strictness: Strictness = .default,
         writableDirectories: [AbsolutePath] = [],
-        strictness: Strictness = .default
+        readOnlyDirectories: [AbsolutePath] = []
     ) -> [String] {
         #if os(macOS)
-        let profile = macOSSandboxProfile(writableDirectories: writableDirectories, strictness: strictness)
+        let profile = macOSSandboxProfile(strictness: strictness, writableDirectories: writableDirectories, readOnlyDirectories: readOnlyDirectories)
         return ["/usr/bin/sandbox-exec", "-p", profile] + command
         #else
         // rdar://40235432, rdar://75636874 tracks implementing sandboxes for other platforms.
@@ -27,9 +39,13 @@ public enum Sandbox {
         #endif
     }
 
+    /// Basic strictness level of a sandbox applied to a command line.
     public enum Strictness: Equatable {
+        /// Blocks network access and all file system modifications.
         case `default`
+        /// More lenient restrictions than the default, for compatibility with SwiftPM manifests using a tools version older than 5.3.
         case manifest_pre_53 // backwards compatibility for manifests
+        /// Like `default`, but also makes temporary-files directories (such as `/tmp`) on the platform writable.
         case writableTemporaryDirectory
     }
 }
@@ -38,8 +54,9 @@ public enum Sandbox {
 
 #if os(macOS)
 fileprivate func macOSSandboxProfile(
+    strictness: Sandbox.Strictness,
     writableDirectories: [AbsolutePath],
-    strictness: Sandbox.Strictness
+    readOnlyDirectories: [AbsolutePath]
 ) -> String {
     var contents = "(version 1)\n"
 
@@ -80,10 +97,20 @@ fileprivate func macOSSandboxProfile(
         }
     }
 
+    // Emit rules for paths under which writing is allowed. Some of these expressions may be regular expressions and others literal subpaths.
     if writableDirectoriesExpression.count > 0 {
         contents += "(allow file-write*\n"
         for expression in writableDirectoriesExpression {
             contents += "    \(expression)\n"
+        }
+        contents += ")\n"
+    }
+
+    // Emit rules for paths under which writing should be disallowed, even if they would be covered by a previous rule to allow writing to them. A classic case is a package which is located under the temporary directory, which should be read-only even though the temporary directory as a whole is writable.
+    if readOnlyDirectories.count > 0 {
+        contents += "(deny file-write*\n"
+        for path in readOnlyDirectories {
+            contents += "    (subpath \(resolveSymlinks(path).quotedAsSubpathForSandboxProfile))\n"
         }
         contents += ")\n"
     }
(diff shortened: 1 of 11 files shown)


Sources/Basics/Sandbox.swift
```diff
@@ -13,23 +13,39 @@ import TSCBasic
...
+    /// Applies a sandbox invocation to the given command line (if the platform supports it),
+    /// and returns the modified command line. On platforms that don't support sandboxing, the
+    /// command line is returned unmodified.
+    ///
+    /// - Parameters:
+    ///   - command: The command line to sandbox (including executable as first argument)
+    ///   - strictness: The basic strictness level of the standbox.
+    ///   - writableDirectories: Paths under which writing should be allowed.
+    ///   - readOnlyDirectories: Paths under which writing should be denied, even if they would have otherwise been allowed by either the strictness level or paths in `writableDirectories`.
+
```
abertelrud, line 26:
```suggestion
```

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -999,6 +988,10 @@ extension SwiftPackageTool {
                 writableDirectories.append(AbsolutePath(pathString, relativeTo: swiftTool.originalWorkingDirectory))
             }
 
+            // Make sure that the package path is read-only unless it's covered by any of the explicitly writable directories.
+            let readOnlyDirectories = writableDirectories.contains{ package.path.isDescendantOfOrEqual(to: $0) } ? [] : [package.path]
+            print("readOnlyDirectories: \(readOnlyDirectories)")
```
abertelrud, line 993:
```suggestion
```

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -895,10 +895,6 @@ extension SwiftPackageTool {
               help: "Allow the plugin to write to an additional directory")
         var additionalAllowedWritableDirectories: [String] = []
 
-        @Flag(name: .customLong("block-writing-to-temporary-directory"),
```
abertelrud, line 898: This flag was just added, so nobody is depending on it yet, and it was mostly for unit tests — it seems much clearer to always make the package directory read-only, even for packages checked out to `/tmp` or other implicitly writable places.

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -979,13 +975,6 @@ extension SwiftPackageTool {
             if allowWritingToPackageDirectory {
                 writableDirectories.append(package.path)
             }
-            if !blockWritingToTemporaryDirectory {
-                writableDirectories.append(AbsolutePath("/tmp"))
```
abertelrud, line 983: The temporary directories are now again handled by the strictness.

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -979,13 +975,6 @@ extension SwiftPackageTool {
...
-                writableDirectories.append(AbsolutePath(NSTemporaryDirectory()))
-            }
-            if allowWritingToPackageDirectory {
```
abertelrud, line 986: This was actually redundant with lines 975-977 above.

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -979,13 +975,6 @@ extension SwiftPackageTool {
...
-                writableDirectories.append(package.path)
-            }
```
abertelrud, line 988: This was actually redundant with lines 975-977 above.

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -999,6 +988,9 @@ extension SwiftPackageTool {
                 writableDirectories.append(AbsolutePath(pathString, relativeTo: swiftTool.originalWorkingDirectory))
             }
 
+            // Make sure that the package path is read-only unless it's covered by any of the explicitly writable directories.
+            let readOnlyDirectories = writableDirectories.contains{ package.path.isDescendantOfOrEqual(to: $0) } ? [] : [package.path]
+
```
abertelrud, line 993: This one is subject to debate.  It avoids marking the package path as readonly if it's covered by any of the directories marked as explicitly writable.  Maybe what should be done here in the long term is to have a single list of directory-specific permissions, each of which could be either `block` or `allow` and then the caller can craft their sandbox directories as they want to.  For example, it might make sense to allow a plugin to modify only a subdirectory of the package.  We can extend this in the future — probably better to keep it simple for now.

Tests/CommandsTests/PackageToolTests.swift
```diff
@@ -1374,8 +1374,9 @@ final class PackageToolTests: CommandsTestCase {
             let packageDir = tmpPath.appending(components: "MyPackage")
             try localFileSystem.writeFileContents(packageDir.appending(components: "Package.swift")) {
                 $0 <<< """
-                // swift-tools-version: 999.0
+                // swift-tools-version: 5.6
```
abertelrud, line 1377: This is an incidental fix that will be needed once we mark the 5.6 branch as no longer a development branch (so it stops accepting 999).  The PR that added this was still open when all the plugins were changed from 999 to 5.6 after evolution proposal acceptance.

abertelrud:
@swift-ci please smoke test

Sources/Basics/Sandbox.swift
```diff
@@ -13,23 +13,38 @@ import TSCBasic
...
     }
 
+    /// Basic strictness level of a sandbox applied to a command line.
     public enum Strictness: Equatable {
+        /// Blocks network access and all file system modifications.
         case `default`
+        /// More lenient restrictions than the default, for compatibility with SwiftPM manifests using a tools version older than 5.3.
         case manifest_pre_53 // backwards compatibility for manifests
+        /// Like `default`, but also makes temporary-files directories (such as `/tmp`) on the platform writable.
         case writableTemporaryDirectory
```
tomerd, line 48: would it better to have the call site pass the temp dirs as writable locations instead of defining a new strictness option for that? alternatively, should this be an argument on "default" strictness instead of it's own case?

Sources/Commands/SwiftPackageTool.swift
tomerd: starting simple makes sense to me

tomerd (approved):

Sources/Commands/SwiftPackageTool.swift
abertelrud: 👍

Sources/Basics/Sandbox.swift
abertelrud: I had that (passing in the temporary directories from the client) initially actually, but that pushes the platform-specific nature of these paths onto the client.  In addition to the paths, there could actually be a varying number of them on different platforms.  But I agree that this shouldn't really be part of the strictness, but rather an option on the default strictness as you suggest.  I think there are more of those in the future, such as whether or not to allow network, etc.

So yes, in retrospect I think it was a mistake to add the temporary directory handling as a separate strictness level, and it's something to be unwound in the future.

Really what I'd like to do is to change the `Sandbox` struct to a `SandboxProfile` or some other name that suggests that it's a set of options, and then that profile could be configured once and passed around so it can be applied to multiple calls later without having to thread through all the individual parameters (and perhaps also cached using operating system calls based on the contents of that profile).  Then as we add more parameters, none of the intermediate code has to change.  The code that calls the subprocess could just apply the same profile to one command after another.  I'm hoping to get to that improvement soonish, in particular since I'm hopeful that it could allow compilation of the profile on macOS just once and the reuse of the compiled profile, saving time when starting sandboxed subprocesses.

At that point it would make sense to get rid of strictness, I think, and just have a set of individually controllable options, and then a convenience initializer to create a `SandboxProfile` that matches the SwiftPM 5.3 semantics.  That special case could then exist outside of the `SandboxProfile` code itself as an extension, which seems more correct than leaking it into the sandbox API.

All of course for a future PR, but hopefully not too long in the future.

PR #4000, opened 2022-01-09
Don't filter out `.docc` bundles from the list of target source files provided to plugins

abertelrud:
Make sure that `.docc` bundles (and in the future, any other files that have built-in rules that mark them as ignored) are still passed through to plugins.  Note that this is separate from files marked as excluded by the manifest, with are filtered out at an earlier stage and never seen by plugins or the internal logic.

### Motivation:

Plugins such as documentation generations need to be able to see `.docc` files without having to look in the file system.  It was a bug that ignored files were not being passed through, as all other files were.

### Background

There was a special rule added in 5.5 to mark .docc bundles as ignored, so that there wouldn't be warnings about packages that contained them (since Xcode could deal with them but SwiftPM couldn't).  But that also caused `.docc` files to become filtered out of the source file lists given to plugins, since they weren't included from the array of other files in the target.

This change keeps ignored files from causing unhandled-files warnings, but it passes them along to plugins as "unknown" role, just other general files.  It does this by keeping them in an additional array.

### Modifications:

- keep track of files that were ignored by rules (note that this is different from excluded files)
- pass those along to the plugin
- update unit tests

### Comments

We should consolidate the separate arrays of sources, resources, other files, and now ignored files.  This PR is intended to be nominated to 5.6 so I'm keeping it simple by just adding another one, but will put up a PR for that restructuring in main sometime in the next month.

rdar://86785502

### DIFF
--- a/Sources/PackageLoading/PackageBuilder.swift
+++ b/Sources/PackageLoading/PackageBuilder.swift
@@ -810,7 +810,7 @@ public final class PackageBuilder {
             fileSystem: self.fileSystem,
             observabilityScope: self.observabilityScope
         )
-        let (sources, resources, headers, others) = try sourcesBuilder.run()
+        let (sources, resources, headers, ignored, others) = try sourcesBuilder.run()
 
         // Make sure defaultLocalization is set if the target has localized resources.
         let hasLocalizedResources = resources.contains(where: { $0.localization != nil })
@@ -868,6 +868,7 @@ public final class PackageBuilder {
                 type: targetType,
                 sources: sources,
                 resources: resources,
+                ignored: ignored,
                 others: others,
                 dependencies: dependencies,
                 swiftVersion: try swiftVersion(),
@@ -902,6 +903,7 @@ public final class PackageBuilder {
                 type: targetType,
                 sources: sources,
                 resources: resources,
+                ignored: ignored,
                 dependencies: dependencies,
                 buildSettings: buildSettings
             )
--- a/Sources/SPMBuildCore/PluginInvocation.swift
+++ b/Sources/SPMBuildCore/PluginInvocation.swift
@@ -824,6 +824,9 @@ struct PluginScriptRunnerInputSerializer {
         targetFiles.append(contentsOf: try target.underlyingTarget.resources.map {
             .init(basePathId: try serialize(path: $0.path.parentDirectory), name: $0.path.basename, type: .resource)
         })
+        targetFiles.append(contentsOf: try target.underlyingTarget.ignored.map {
+            .init(basePathId: try serialize(path: $0.parentDirectory), name: $0.basename, type: .unknown)
+        })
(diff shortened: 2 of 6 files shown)


abertelrud:
@swift-ci smoke test

Sources/PackageLoading/PackageBuilder.swift
```diff
@@ -810,7 +810,7 @@ public final class PackageBuilder {
             fileSystem: self.fileSystem,
             observabilityScope: self.observabilityScope
         )
-        let (sources, resources, headers, others) = try sourcesBuilder.run()
+        let (sources, resources, headers, ignored, others) = try sourcesBuilder.run()
```
abertelrud, line 813: Adding another tuple member keeps diffs minimal for this PR but this is getting out-of-hand.  I plan to put up a PR to refactor the results coming back from the `TargetSourcesBuilder` but want to keep risk minimal to fix this bug.

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -824,6 +824,9 @@ struct PluginScriptRunnerInputSerializer {
         targetFiles.append(contentsOf: try target.underlyingTarget.resources.map {
             .init(basePathId: try serialize(path: $0.path.parentDirectory), name: $0.path.basename, type: .resource)
         })
+        targetFiles.append(contentsOf: try target.underlyingTarget.ignored.map {
```
abertelrud, line 827: This is the line that actually fixes the bug, which is to include the "ignored" files as source file of unknown role.  Note again that this is different from files marked "excluded" in the manifest — those are never passed to the internal build log and not to plugins either.

Tests/CommandsTests/PackageToolTests.swift
```diff
@@ -1232,6 +1232,17 @@ final class PackageToolTests: CommandsTestCase {
                 public func Foo() { }
                 """
             }
+            try localFileSystem.writeFileContents(packageDir.appending(components: "Sources", "MyLibrary", "test.docc")) {
+                $0 <<< """
```
tomerd, line 1236: nit: you can pass string directly no need in byte stream

Sources/PackageLoading/PackageBuilder.swift
tomerd: yea seems like this requires a bit of refactoring, even a struct to replace the tuple seems nicer

tomerd (approved):

Tests/CommandsTests/PackageToolTests.swift
abertelrud: Thanks — I did try that in the previous PR where you mentioned it, but it would require manual creation of intermediate directories, so this seems better.  What's I'd like to do as follow-up is add an optional parameter about whether to create intermediate directories to the ToolSupportCore methods, and then switch over to use them.  The trade-off would be to use a string here but to explicitly have to create all directories.

PR #4002, opened 2022-01-10
Allow running tests from a command plugin

abertelrud:
This wraps up the last piece for SE-0332, including support for filtering and for capturing code coverage while the tests are running.

### Motivation:

This is specified in SE-0332 (it's the last part of SE-0332 that wasn't yet merged).

### Modifications:

- add support to running tests in a similar manner as `SwiftTestTool`, reusing much of the implementation of `swift test`
- factor out some of the `SwiftTestTool` methods so they they can be invoked on request from plugins too
- add a unit test

### Comment

This uses most of the same implementation that existed in a very specify way in SwiftTestTool.swift, factoring it out into a new shared `TestingSupport.swift` file and making it more generic.  Future refactoring should move it down to a new module that provides the shared test support for use from everywhere.

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -1126,12 +1126,12 @@ final class PluginDelegate: PluginInvocationDelegate {
         // Run the build in the background and call the completion handler when done.
         DispatchQueue.sharedConcurrent.async {
             completion(Result {
-                return try self.doBuild(subset: subset, parameters: parameters)
+                return try self.performBuildForPlugin(subset: subset, parameters: parameters)
             })
         }
     }
     
-    private func doBuild(subset: PluginInvocationBuildSubset, parameters: PluginInvocationBuildParameters) throws -> PluginInvocationBuildResult {
+    private func performBuildForPlugin(subset: PluginInvocationBuildSubset, parameters: PluginInvocationBuildParameters) throws -> PluginInvocationBuildResult {
         // Configure the build parameters.
         var buildParameters = try self.swiftTool.buildParameters()
         switch parameters.configuration {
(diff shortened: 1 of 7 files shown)


abertelrud:
@swift-ci smoke test

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -1207,8 +1207,7 @@ final class PluginDelegate: PluginInvocationDelegate {
     }
     
     private func doTest(subset: PluginInvocationTestSubset, parameters: PluginInvocationTestParameters) throws -> PluginInvocationTestResult {
-        // FIXME: To implement this we should factor out a lot of the code in SwiftTestTool.
-        throw StringError("Running tests from a plugin is not yet implemented")
+        return try SwiftTestTool.doTest(subset: subset, parameters: parameters, swiftTool: swiftTool)
```
abertelrud, line 1210: The implementation is in SwiftTestTool, where the existing support code is.  That code should really be moved down to one of the library modules where it can be used by both SwiftTestTool and the other code that needs it.  SwiftTestTool should just be a thin layer that defines command line actions that then call down into the library code that can run the tests without assumption about the context.

Sources/Commands/SwiftTestTool.swift
```diff
@@ -1139,3 +1140,141 @@ private extension SwiftTool {
...
+        // Return the results to the plugin. We only consider the test run a success if no test failed.
+        return PluginInvocationTestResult(
+            succeeded: (numFailedTests == 0),
+            testTargets: testTargetResults,
+            codeCoverageDataFile: codeCoverageDataFile?.pathString)
+    }
+}
+
+fileprivate extension DispatchTimeInterval {
+    var asSeconds: Double? {
```
abertelrud, line 1266: We should consider consolidating this code with the other `descriptionInSeconds` extension, making all of the translation public.  It was somewhat surprising that there was no library method for getting seconds out of a `DispatchTimeInterval`.

Sources/Workspace/UserToolchain.swift
```diff
@@ -223,12 +223,12 @@ public final class UserToolchain: Toolchain {
 
     /// Returns the path to llvm-cov tool.
     public func getLLVMCov() throws -> AbsolutePath {
-        return try UserToolchain.getTool("llvm-cov", binDir: self.destination.binDir)
+        return try UserToolchain.getTool("llvm-cov", binDir: self.swiftCompilerPath.parentDirectory)
```
abertelrud, line 226: This one and the one below were actually wrong and didn't work in debug builds of SwiftPM, as these tools come from the toolchain and are not built by SwiftPM.  The new code is the same as for the symbol graph extraction tool and the other similar tools.

Tests/CommandsTests/PackageToolTests.swift
```diff
@@ -1541,4 +1541,130 @@ final class PackageToolTests: CommandsTestCase {
...
+
+    func testCommandPluginTestingCallbacks() throws {
+        // Depending on how the test is running, the `llvm-profdata` and `llvm-cov` tool might be unavailable.
+        try XCTSkipIf((try? UserToolchain.default.getLLVMProf()) == nil, "skipping test because the `llvm-profdata` tool isn't available")
+        try XCTSkipIf((try? UserToolchain.default.getLLVMCov()) == nil, "skipping test because the `llvm-cov` tool isn't available")
+
+        try testWithTemporaryDirectory { tmpPath in
+            // Create a sample package with a library, a command plugin, and a couple of tests.
+            let packageDir = tmpPath.appending(components: "MyPackage")
+            try localFileSystem.createDirectory(packageDir, recursive: true)
```
abertelrud, line 1553: We should have a version of `writeFileContents(_:string:)` that can create intermediate directories when needed.

In the long run it would be useful for unit tests to have a way of creating a file system using something like result builders, to lay out contents in the file system like SwiftUI lays out views on the screen.  We could then have a library of generator for common kinds of files.  For now, using the string versions of `writeFileContents` seems better than the stream block version even though we have to create the directories ourselves.

abertelrud:
Unit test failure looks unrelated to these diffs, and maybe a result of `Update the toolchain name (commit: 656884b)` (from the bot log).

abertelrud:
@swift-ci smoke test

Sources/Commands/SwiftPackageTool.swift
tomerd: agreed this seems like a bit of an abstraction leak. how difficult is it to refactor to make this better layered?

Sources/Commands/SwiftTestTool.swift
tomerd: there is a SE proposal that would address this in the future. for now I suggest we move such extensions to the Basics module so they are reusable

Sources/Commands/SwiftTestTool.swift
```diff
@@ -1139,3 +1140,141 @@ private extension SwiftTool {
         return parameters
     }
 }
+
+extension SwiftTestTool {
```
tomerd, line 1144: should we define a new target "Test" or "PackageTest" (the build one is call "Build"), move this code there, and then reuse from both call sites?

tomerd: or even to workspace if we must

Sources/Commands/SwiftTestTool.swift
```diff
@@ -1139,3 +1140,141 @@ private extension SwiftTool {
...
+
+    public static func doTest(subset: PluginInvocationTestSubset, parameters: PluginInvocationTestParameters, swiftTool: SwiftTool) throws -> PluginInvocationTestResult {
```
tomerd, line 1146: should this return something more generic than PluginInvocationTestResult and then the call site does the translation? this feels fairly leaky at the abstraction level

Sources/Commands/SwiftTestTool.swift
tomerd: the other alternative (maybe better) it to move all this code into SwiftPackageTool which handles plugins already (so that the fact doTest returns PluginInvocationTestResult is less leaky), and move the shared utilities (eg getTestSuites) to some static helper that all tools can use, e.g. TestUtilities that lives in the tools area or even defined in SwiftTool.swift

Tests/CommandsTests/PackageToolTests.swift
tomerd: one interesting thing I found was that `writeFileContents(_ path: AbsolutePath, body: (WritableByteStream) -> Void) throws` creates the directory while `func writeFileContents(_ path: AbsolutePath, bytes: ByteString)`. its a subtle difference and not super clear to me why one does it over the other, but we should probably may the behavior the same across the board, maybe by adding an optional argument `createDirectory` with no default such that the test code can create an extension that sets it to true. wdyt?

tomerd: cc @neonichu

abertelrud: Right, that's exactly what I had in mind (the extra parameter).

Sources/Commands/SwiftPackageTool.swift
abertelrud: It doesn't seem difficult, but will take a little time and introduce some risk to the existing test functionality.  I think the best would be to rewrite the testing functionality — it's a bit of a mess, with layers of changes.  Rewriting the testing seemed to me to be appropriate to do as a clean-up task, but I didn't want to block the ability of plugins to invoke tests on that.

Tests/CommandsTests/PackageToolTests.swift
abertelrud: Since this PR uses the string forms I don't think we should block it on changing `writeFileContents` though.

Sources/Commands/SwiftTestTool.swift
abertelrud: That seems reasonable.  That would be short of the rewrite that this SwiftTestTool needs but still not add too much to the problem.

Sources/Commands/SwiftTestTool.swift
abertelrud: Right, that's a good point.  I originally started with this being in SwiftPackageTool but then kept it as it moved over.  Really the plugin callbacks should be in their own separate file, actually, but I was hesitant to "pull on that thread" as part of this PR.  But it's certainly a better place to end up.

I'll move this over as per your previous comment though — that seems better for this PR, and then switching SwiftTestTool to use the same code in a further followup.

Tests/CommandsTests/PackageToolTests.swift
tomerd: probably a radar to track the desired refactoring?

tomerd (approved):

abertelrud:
FYI:  Working on some of the other feedback here.  Will mark as WIP while I do.  Should be done later today.

Tests/CommandsTests/PackageToolTests.swift
abertelrud: Definitely.  I'll file a bug on bugs.swift.org.

Sources/Commands/SwiftTestTool.swift
abertelrud: Factored this out into a separate source file and moved the main logic back into SwiftPackageTool.  Future improvements would be to move it down into its own module.

Tests/CommandsTests/PackageToolTests.swift
abertelrud: Filed https://bugs.swift.org/browse/SR-15719

abertelrud:
@swift-ci smoke test

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -1206,9 +1206,122 @@ final class PluginDelegate: PluginInvocationDelegate {
         }
     }
     
-    private func doTest(subset: PluginInvocationTestSubset, parameters: PluginInvocationTestParameters) throws -> PluginInvocationTestResult {
-        // FIXME: To implement this we should factor out a lot of the code in SwiftTestTool.
-        throw StringError("Running tests from a plugin is not yet implemented")
+    func doTest(subset: PluginInvocationTestSubset, parameters: PluginInvocationTestParameters) throws -> PluginInvocationTestResult {
```
tomerd, line 1209: nit: invokeTestsForPlugin

Sources/Commands/TestHelpers.swift
```diff
@@ -0,0 +1,136 @@
...
+import SPMBuildCore
+import TSCBasic
+import Workspace
+
+
+/// Locates XCTestHelper tool inside the libexec directory and bin directory.
+/// Note: It is a fatalError if we are not able to locate the tool.
+///
+/// - Returns: Path to XCTestHelper tool.
+func xctestHelperPath(swiftTool: SwiftTool) throws -> AbsolutePath {
```
tomerd, line 21: would it be better encapsulations to have these as static functions on a new "TestHelper" struct rather than global functions?

abertelrud: I'll make this change (and call it `TestSupport`).  I kept them as free functions to minimize the change on the call sites, but it's better to keep them inside a type.

abertelrud:
@swift-ci smoke test

tomerd (approved):

PR #4005, opened 2022-01-12
Rework the `Sandbox` enum into a `SandboxProfile` struct that can separate out point of configuration from point of application

abertelrud:
This change turns the stateless Sandbox enum into a SandboxProfile struct that can be configured and passed around well before being applied, and it makes the configuration more flexible in several regards:
- there is a list of path rules allowing a mixed order of `allow` and `deny` rules
- the choice of allowing writing to temporary directories is independent of anything other choice
- the defaults are built into the initializer rather than a separate `strictness` parameter — this means that they can be queried once the SandboxProfile has been created and are expressed in terms of the choices available to all sandbox profiles.

### Motivation

Separating out the point of configuration from the point of application of the sandbox makes it easier to configure.

The ergonomic changes described above also make it easier to use.

Having the sandbox profile be a struct that generates the platform specifics as needed also provides a place with which to associate any cached/compiled representation for profiles that are largely static.

### Details

As cleanup, this commit also removes the pre-SwiftPM 5.3 specialities which were specific to running the package manifest in an interpreter rather than compiling and executing it.  This functionality is no longer relevant since it isn't possible to run any manifest in the interpreter.

In this commit, the call sites have been adjusted so that they use the modified SandboxProfile API, but still construct the profiles on-the-fly as before.  A future change could make them instead be configured at an early point and then applied later when the sandboxed process is actually launched.  Another future change could add the sandbox profile to Process as a property so that there is no API assumption that applying a sandbox necessarily involves just modifying the command line.


Sources/Basics/Sandbox.swift
```diff
@@ -20,32 +20,34 @@ public enum Sandbox {
...
+        case readonly(path: AbsolutePath)
+        case writable(path: AbsolutePath)
+    }
+
     /// Basic strictness level of a sandbox applied to a command line.
     public enum Strictness: Equatable {
-        /// Blocks network access and all file system modifications.
-        case `default`
+        /// Blocks network access and all file system modifications except, optionally, temporary locations.
+        case `default`(writableTemporaryDirectory: Bool)
```
abertelrud, line 48: This came out of a previous PR discussion but I'm not 100% happy with how it works, ergonomically.  Ideally I think all the sandbox options should be properties, and then the default and pre-5.3-manifest behaviors should be initializers that return a filled-in Sandbox struct.  Whether or not to allow writing to temporary directories would then be a property like the others.

Tests/CommandsTests/PackageToolTests.swift
```diff
@@ -1361,6 +1370,7 @@ final class PackageToolTests: CommandsTestCase {
                 let output = try result.utf8Output() + result.utf8stderrOutput()
                 XCTAssertEqual(result.exitStatus, .terminated(code: 0), "output: \(output)")
                 XCTAssertMatch(output, .contains("This is MyCommandPlugin."))
+                XCTAssertMatch(output, .contains("successfully created file at path \(resolveSymlinks(packageDir).appending(components: ".build", "plugins", "outputs", "Foo"))"))
```
abertelrud, line 1373: This is the check that was incomplete in the previous test, and would have caught the problem.

abertelrud:
@swift-ci smoke test

Sources/Basics/Sandbox.swift
tomerd: I agree!

abertelrud: Sounds great.  I was hesitant to go too far down that road for something that I want to nominate in 5.6, but will take another pass at this.  I don't think it will add a lot of risk.  Thanks for taking a look quickly!

abertelrud: So what I'd actually like to do is to call it `SandboxProfile`, make it a struct, and have it contain the configuration options.  Then applying that profile to a command line would be a member function (not static).  I would still like to leave the whole part of having it be a parameter to the Process configuration as a later cleanup, and for now stay with the assumption that it's a transformation on the command line.

tomerd: this sounds great

abertelrud:
@swift-ci smoke test

Tests/BasicsTests/SandboxTests.swift
```diff
@@ -1,60 +1,73 @@
...
         }
     }
 
     func testNetworkNotAllowed() throws {
         #if !os(macOS)
-        try XCTSkipIf(true, "test is only supported on macOS")
+        try XCTSkipIf(true, "sandboxing is only supported on macOS")
         #endif
 
-        let command = Sandbox.apply(command: ["ping", "-t", "1", "localhost"], strictness: .default, writableDirectories: [])
```
abertelrud, line 30: This is blocked anyway if the firewall is on and "stealth mode" is enabled, so this isn't an adequate test of the sandbox.

Tests/BasicsTests/SandboxTests.swift
```diff
@@ -86,72 +100,106 @@ final class SandboxTest: XCTestCase {
...
+            XCTAssertThrowsError(try Process.checkNonZeroExit(arguments: deniedCommand)) { error in
+                guard case ProcessResult.Error.nonZeroExit(let result) = error else {
+                    return XCTFail("invalid error \(error)")
+                }
+                XCTAssertMatch(try! result.utf8stderrOutput(), .contains("Operation not permitted"))
+            }
+        }
+    }
+
+    func testWritingToWritableInsideReadOnlyAllowed() throws {
```
abertelrud, line 179: This adds a specific test case for a writable directory inside of a read-only directory.

Sources/PackageLoading/ManifestLoader.swift
```diff
@@ -899,8 +899,8 @@ public final class ManifestLoader: ManifestLoaderProtocol {
                     // We only allow the permissions which are absolutely necessary.
                     if self.isManifestSandboxEnabled {
                         let cacheDirectories = [self.databaseCacheDir, moduleCachePath].compactMap{ $0 }
-                        let strictness: Sandbox.Strictness = toolsVersion < .v5_3 ? .manifest_pre_53 : .default
-                        cmd = Sandbox.apply(command: cmd, strictness: strictness, writableDirectories: cacheDirectories)
+                        let sandboxProfile = SandboxProfile(pathRules: cacheDirectories.map{ .writable($0) })
+                        cmd = sandboxProfile.apply(to: cmd)
```
abertelrud, line 903: The notes in the Sandbox profile generator indicate that the pre-5.3 rules were needed for interpreting the manifest rather than compiling and running it.  This isn't supported anymore, and the operations affected were being able to write to some Clang locations and to set `sysctl`s.  Those don't seem like operations that we want to keep supporting.  Do we know of any packages that required the < 5.3 strictness?

abertelrud:
@swift-ci smoke test

tomerd (approved):
<3 this is great

abertelrud:
@swift-ci please test package compatility

abertelrud:

abertelrud:

abertelrud:
Hmm...
```
11:01:23 
/Users/buildnode/jenkins/workspace/swift-package-manager-source-compat-suite-PR-macos/swift-driver/Sources/SwiftDriver/Utilities/System.swift:60:32: error: multi-line string literal content must begin on a new line
11:01:23       var quoted: String = #"""#
```

abertelrud:
Moving this one aside temporarily while preparing a much smaller fix for the problem at hand.  That fix can then be nominated for inclusion in 5.6.  Then I'll bring this back since it's what we want to do moving forward.

abertelrud:

abertelrud:
@swift-ci please smoke test

abertelrud:
Opened a new PR here:  https://github.com/apple/swift-package-manager/pull/4307 that replaces this one

PR #4006, opened 2022-01-12
[wip] dependency resolution performance

tomerd:
motivation: improve dependency resolution performance

changes:
* add informational diagnostics on manifest cache hit/miss
* replace manifest loader operation queue with semaphore
* audit and clarify the use of manifest loader dispatch queues for delegate callbacks vs. completion handlers
* disable expensive optimization in pubgrub resolver

### DIFF
--- a/Tests/PackageCollectionsSigningTests/CertificatePolicyTests.swift
+++ b/Tests/PackageCollectionsSigningTests/CertificatePolicyTests.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See http://swift.org/LICENSE.txt for license information
@@ -20,7 +20,7 @@ class CertificatePolicyTests: XCTestCase {
     func test_RSA_validate_happyCase() throws {
         try skipIfUnsupportedPlatform()
 
-        fixture(name: "Collections") { directoryPath in
+        fixture(name: "Collections", createGitRepo: false) { directoryPath in
             let certPath = directoryPath.appending(components: "Signing", "Test_rsa.cer")
             let certificate = try Certificate(derEncoded: Data(try localFileSystem.readFileContents(certPath).contents))
 
@@ -40,7 +40,7 @@ class CertificatePolicyTests: XCTestCase {
     func test_EC_validate_happyCase() throws {
         try skipIfUnsupportedPlatform()
 
-        fixture(name: "Collections") { directoryPath in
+        fixture(name: "Collections", createGitRepo: false) { directoryPath in
             let certPath = directoryPath.appending(components: "Signing", "Test_ec.cer")
             let certificate = try Certificate(derEncoded: Data(try localFileSystem.readFileContents(certPath).contents))
 
@@ -60,7 +60,7 @@ class CertificatePolicyTests: XCTestCase {
     func test_validate_untrustedRoot() throws {
         try skipIfUnsupportedPlatform()
 
-        fixture(name: "Collections") { directoryPath in
+        fixture(name: "Collections", createGitRepo: false) { directoryPath in
             let certPath = directoryPath.appending(components: "Signing", "Test_rsa.cer")
             let certificate = try Certificate(derEncoded: Data(try localFileSystem.readFileContents(certPath).contents))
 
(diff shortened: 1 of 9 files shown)


Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
```diff
@@ -1232,11 +1232,11 @@ internal final class PubGrubPackageContainer {
             }
         }
 
-        let (lowerBounds, upperBounds) = try self.computeBounds(
+        /*let (lowerBounds, upperBounds) = try self.computeBounds(
             for: node,
             constraints: constraints,
             startingWith: version
-        )
+        )*/
```
tomerd, line 1239: computation of the bounds here is _very_ expensive and it needs to evaluate manifests throughout the dependency closure. afaict this is only used for the `emittedIncompatibilities` which is itself a performance optimization. disabling this gave 3x performance improvement on my initial tests

tomerd: cc @SDGGiesbrecht as this reflects our previous discussion about this optimization

tomerd:
@swift-ci please smoke test

tomerd:

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
SDGGiesbrecht: 🤷 If it’s not actually faster, it’s not actually faster.

If you remove this, then `emittedIncompatibilities` is never written to and thus always empty, which in turn means...
- The line with `if emittedIncompatibilities[node]?.contains(version) != true {` is always true, and the contents of the `if` may as well be run unconditionally.
- The line with `emittedIncompatibilities.isEmpty` is also always true, although not necessarily the other conditions grouped with it.

While looking at it, I also noticed that [this section](https://github.com/apple/swift-package-manager/blob/15d19824817917b7faf7680c5a2d8a4bd459c16e/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift#L1219-L1232) and [this section](https://github.com/apple/swift-package-manager/blob/15d19824817917b7faf7680c5a2d8a4bd459c16e/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift#L1241-L1266) are now basically identical—comments and all. So you can probably reduce [this](https://github.com/apple/swift-package-manager/blob/15d19824817917b7faf7680c5a2d8a4bd459c16e/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift#L1207-L1233) to just the following and let it naturally fall through into the other, identical section.

```swift
if version == pinnedVersion {
  if self.emittedPinnedVersionIncompatibilities[node] ?? false { return [] }
  self.emittedPinnedVersionIncompatibilities[node] = true
} 
```

It might be worth checking whether those remaining lines are actually helpful for performance. If not, you could get rid of `emittedPinnedVersionIncompatibilities` too.

SDGGiesbrecht: This particular optimization predates me. The [commit that added it](https://github.com/apple/swift-package-manager/commit/aee02eb246442bcc28119101b89feee1d54110b4) contains a bit of an explanation and a radar link (to which I lack access). It might be worth checking that radar for specific examples to double‐check for performance regressions. Also, if @aciidb0mb3r is still around out there somewhere, he might have something to add.

Tests/PackageLoadingTests/PD_4_2_LoadingTests.swift
```diff
@@ -864,6 +864,7 @@ class PackageDescription4_2LoadingTests: PackageDescriptionLoadingTests {
                     identityResolver: identityResolver,
                     fileSystem: localFileSystem,
                     observabilityScope: observability.topScope,
+                    delegateQueue: .global(),
```
abertelrud, line 867: Just curious:  what's the reason that `sharedConcurrent` is sometimes used and sometimes `global()`?

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
abertelrud: Could it have an impact on the quality of the diagnostics when there are conflicts?

Tests/PackageLoadingTests/PD_4_2_LoadingTests.swift
tomerd: this was an unrelated change that I removed now and will add in a separate PR

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
tomerd: we have a fairly robust set of tests for dependency resolution diagnostics which _have been_ impacted by #3985 but not by this. afaict the code here was added in https://github.com/apple/swift-package-manager/commit/aee02eb246442bcc28119101b89feee1d54110b4 as an attempt to improve diagnostics but because it is so expensive it is also cached. we already re-did the part that impacted the diagnostics in #3985 by taking a bit of a different approach and such this PR is purely about performance in not running that expensive calculation any more as it no longer has any functional value

tomerd:
@swift-ci please smoke test

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
tomerd: this is now https://github.com/apple/swift-package-manager/pull/4007, let take the discussion there?

PR #4009, opened 2022-01-13
Sandbox blocks output to default plugin output directory when it's under `<pkgdir>/.build`

abertelrud:
The sandbox rules introduced in #3996 made the entire package directory read-only, but that isn't appropriate when `.build` is inside the package directory.

The rules for applying read-only directories were intended to shadow the implicit rules added by specifying a writable temporary directory, not to block the ones specified by explicit writable directories.  This fixes that ordering, and adds the missing unit test that would have caught the problem.

A future change reworks Sandbox completely so that it becomes a struct that can carry around the sandbox profile configuration until it is applied — this change is intended to be small enough to be nominatable

### Changes

- in the generated sandbox profile, swap the order of the explicit readable and writable properties
- add a new unit test for this case (existing ones check the other cases)

rdar://87417780


abertelrud:
@swift-ci please smoke test

tomerd (approved):

Sources/Basics/Sandbox.swift
```diff
@@ -114,6 +113,15 @@ fileprivate func macOSSandboxProfile(
         contents += ")\n"
     }
 
+    // Emit rules for paths under which writing is allowed, even if they are descendants directories that are otherwise read-only.
+    if writableDirectories.count > 0 {
```
tomerd, line 117: there seems to be duplicate code block that write these out. is one redundant?

abertelrud: They look almost the same, but the L99 one is emitting the expressions for the implicit writable directories, which includes both temporary directories and the Clang cache directories for pre-5.3 compatibility (the former are subpaths and the latter regexes).  The problem was that the explicitly specified writable directories were emitted as part of that same batch of directories, followed by the explicitly writable ones.  But really the correct order is:
1. implicitly writable (temporary directories and the Clang cache directories)
2. explicitly readable
3. explicitly writable

In the new PR https://github.com/apple/swift-package-manager/pull/4005 this is made a lot more clear with a bit less redundancy.

abertelrud:
@swift-ci please smoke test

compnerd:
@abertelrud @tomerd - this seems to have regressed the windows build: https://ci-external.swift.org/job/oss-swift-windows-toolchain-x86_64-vs2019/619/console

PR #4010, opened 2022-01-13
Add support for package index as metadata and search provider

yim-lee:
rdar://85617104, rdar://85617156, rdar://85617198

TODOs:
- [x] Remove `fatalError`
- [x] Add tests
- [x] Caching


Sources/PackageCollections/PackageIndex.swift
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    func getPackageMetadata(
+        identity: PackageIdentity,
+        location: String?,
+        callback: @escaping (Result<PackageCollectionsModel.PackageMetadata, Error>) -> Void
+    ) {
+        self.runIfConfigured(callback: callback) { _, _ in
+            // TODO: call package index's get metadata API
+            fatalError("Not implemented: \(#function)")
```
yim-lee, line 87: We should probably have some "fake" implementation at least so we can write tests?

Sources/PackageCollections/PackageIndex.swift
```diff
@@ -0,0 +1,186 @@
...
+                        )
+                    },
+                    watchersCount: package.watchersCount,
+                    readmeURL: package.readmeURL,
+                    license: package.license,
+                    authors: package.authors,
+                    languages: package.languages
+                )
+                
+                let url = try? temp_await { callback in self.get(callback: callback) }
```
tomerd, line 173: temp_await needed?

Sources/PackageCollections/PackageIndex.swift
```diff
@@ -0,0 +1,186 @@
...
+        let callback = self.makeAsync(callback)
+        
+        guard self.enabled else {
+            return callback(.failure(PackageIndexError.featureDisabled))
+        }
+        
+        self.configurationStorage.update(index: .init(url: nil)) { result in
+            callback(result.tryMap { () })
+        }
+    }
```
tomerd, line 66: not sure if we need the set/unset to be part of the API for PackageIndex. in other places in SwiftPM similar business logic modules get a readonly copy of the configuration in the initializer so they can read the configuration. the CLI / Xcode deal with set/unset using the "storage" type data structures. in this case this means the CLI deals with PackageIndexConfigurationStorage, and PackageIndex initializer would take PackageIndexConfiguration which is a readonly representation of the active configuration

Sources/PackageCollections/Providers/PackageMetadataProvider.swift
```diff
@@ -16,23 +16,18 @@ import TSCBasic
...
-
-    /// Returns `AuthTokenType` for a package.
-    ///
-    /// - Parameters:
-    ///   - location: The package's location
-    func getAuthTokenType(for location: String) -> AuthTokenType?
+    func get(
+        identity: PackageIdentity,
+        location: String,
+        callback: @escaping (Result<PackageCollectionsModel.PackageBasicMetadata, Error>, PackageMetadataProviderContext?) -> Void
```
yim-lee, line 29: In additional to the new APIs, I am making major changes around `PackageMetadataProviderContext` as well. 

The sole(?) purpose of `PackageMetadataProviderContext` is a way of telling the caller any auth-related warning (e.g., it's better to configure auth token for a higher API limit) or error (e.g., permission denied). i.e., `PackageMetadataProviderContext` is not just for error cases only. `PackageMetadataProviderContext` was constructed separately in previous logic, now it's part of the `get` response.

Also, `PackageMetadataProviderContext` has a new `name` property so the caller would know where the metadata comes from.

Sources/PackageCollections/PackageIndexAndCollections.swift
```diff
@@ -0,0 +1,284 @@
...
+    public func getPackageMetadata(
+        identity: PackageIdentity,
+        location: String?,
+        collections: Set<PackageCollectionsModel.CollectionIdentifier>?,
+        callback: @escaping (Result<PackageCollectionsModel.PackageMetadata, Error>) -> Void
+    ) {
+        // Get metadata using both package index and collections
+        let sync = DispatchGroup()
+        let results = ThreadSafeKeyValueStore<Source, Result<PackageCollectionsModel.PackageMetadata, Error>>()
+        DispatchQueue.sharedConcurrent.async {
```
tomerd, line 166: since getPackageMetadata is already async, invoking it in  the DispatchQueue.sharedConcurrent.async block seems redundant?

Sources/PackageCollections/PackageIndexAndCollections.swift
```diff
@@ -0,0 +1,284 @@
...
+            // This uses either package index or "alternative" (e.g., GitHub) as metadata provider,
+            // then merge the supplementary metadata with data coming from collections. The package
+            // must belong to at least one collection.
+            self.collections.getPackageMetadata(identity: identity, location: location, collections: collections) { result in
+                defer { sync.leave() }
+                results[.collections] = result
+            }
+        }
+        
+        sync.notify(queue: .sharedConcurrent) {
```
tomerd, line 185: cool

Sources/PackageCollections/PackageIndexAndCollections.swift
```diff
@@ -0,0 +1,284 @@
...
+                    case .success(let collectionsSearchResult):
+                        let items = indexSearchResult.items.map { item in
+                            PackageCollectionsModel.PackageSearchResult.Item(
+                                package: item.package,
+                                collections: collectionsSearchResult.items.first(where: {
+                                    item.package.identity == $0.package.identity && item.package.location == $0.package.location
+                                })?.collections ?? [],
+                                indexes: item.indexes
+                            )
+                        }
```
tomerd, line 252: does this "decorate" the index search results with information about which collection the result belongs to?

Sources/PackageCollections/PackageIndex.swift
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// MARK: - PackageMetadataProvider conformance
+
+extension PackageIndex: PackageMetadataProvider {
+    func get(
+        identity: PackageIdentity,
+        location: String,
+        callback: @escaping (Result<PackageCollectionsModel.PackageBasicMetadata, Error>, PackageMetadataProviderContext?) -> Void
```
tomerd, line 146: is PackageMetadataProviderContext used outside the successful case?

yim-lee: Yes, see https://github.com/apple/swift-package-manager/pull/4010#discussion_r783556211

Sources/PackageCollections/PackageIndex.swift
yim-lee: This means package index URL doesn't change until next time it gets initialized, correct?

Sources/PackageCollections/PackageIndex.swift
yim-lee: No, if config (i.e., package index URL) doesn't change during lifetime of `PackageIndex`.

Sources/PackageCollections/PackageIndexAndCollections.swift
```diff
@@ -0,0 +1,284 @@
...
+    public func setIndex(
+        url: Foundation.URL,
+        callback: @escaping (Result<Void, Error>) -> Void
+    ) {
+        self.index.set(url: url, callback: callback)
+    }
+
+    public func unsetIndex(callback: @escaping (Result<Void, Error>) -> Void) {
+        self.index.unset(callback: callback)
+    }
```
tomerd, line 141: similarly to the comment above the set/unset should not be part of this API imo, and should only be exposed on the storage data structure

Sources/PackageCollections/PackageIndex.swift
tomerd: as an example you see the mirrors passed into workspace, and the mirrors storage which is used to set/unset that is only used by the CLI (and IDEs)

Tests/PackageCollectionsTests/GitHubPackageMetadataProviderTests.swift
```diff
@@ -394,3 +361,11 @@ internal extension GitHubPackageMetadataProvider {
         )
     }
 }
+
+private extension GitHubPackageMetadataProvider {
```
tomerd, line 365: cool

Sources/PackageCollections/PackageIndex.swift
tomerd: yea that is a goo idea, maybe even a real http client with fake routes and payloads

Sources/PackageCollections/PackageIndex.swift
tomerd: > This means package index URL doesn't change until next time it gets initialized, correct?

yes, which is think is fine imo. the CLI will initialize this on every call anyways, and the Xcode integration layer could do the same if it needs to

Sources/PackageCollections/API.swift
```diff
@@ -248,3 +251,192 @@ public enum PackageCollectionError: Equatable, Error {
...
+    /// Package index support is disabled
+    case featureDisabled
+    /// No package index configured
+    case notConfigured
+}
+
+// MARK: - Package index and collection combined
+
+// It's on purpose that the protocol does not conform to `PackageCollectionsProtocol` and `PackageIndexProtocol`,
+// so that we can exclude APIs that we believe are not required and have more flexibility around return types.
```
tomerd, line 326: 👍 

unless we need it for mocking / testing, maybe we dont even need a dedicated protocol and the concrete impl is enough

Sources/PackageCollections/PackageCollections.swift
```diff
@@ -26,14 +26,21 @@ public struct PackageCollections: PackageCollectionsProtocol {
...
     // initialize with defaults
     public init(configuration: Configuration = .init(), observabilityScope: ObservabilityScope) {
+        self.init(configuration: configuration, customMetadataProvider: nil, observabilityScope: observabilityScope)
+    }
+    
+    init(
+        configuration: Configuration = .init(),
+        customMetadataProvider: PackageMetadataProvider?,
+        observabilityScope: ObservabilityScope
+    ) {
```
tomerd, line 43: maybe collapse the two initializers into one by giving customMetadataProvider default value?

Sources/PackageCollections/PackageIndex.swift
```diff
@@ -0,0 +1,186 @@
...
+        callback: @escaping (Result<PackageCollectionsModel.PaginatedPackageList, Error>) -> Void
+    ) {
+        self.runIfConfigured(callback: callback) { _, _ in
+            // TODO: cap `limit`
+            // TODO: call package index's list API
+            fatalError("Not implemented: \(#function)")
+        }
+    }
+
+    private func runIfConfigured<T>(
```
tomerd, line 113: nice

Sources/PackageCollections/PackageIndex.swift
yim-lee: amended https://github.com/apple/swift-package-manager/pull/4010/commits/c1d364a022d867b5bfcd9f69006a1fc714001414

Sources/PackageCollections/PackageIndex.swift
yim-lee: amended https://github.com/apple/swift-package-manager/pull/4010/commits/c1d364a022d867b5bfcd9f69006a1fc714001414

Sources/PackageCollections/PackageIndexAndCollections.swift
yim-lee: amended https://github.com/apple/swift-package-manager/pull/4010/commits/c1d364a022d867b5bfcd9f69006a1fc714001414

Sources/PackageCollections/API.swift
yim-lee: removed https://github.com/apple/swift-package-manager/pull/4010/commits/c1d364a022d867b5bfcd9f69006a1fc714001414

Sources/PackageCollections/PackageCollections.swift
```diff
@@ -43,41 +50,41 @@ public struct PackageCollections: PackageCollectionsProtocol {
...
-         metadataProvider: PackageMetadataProvider,
-         customSearchProvider: PackageSearchProvider? = nil) {
+         metadataProvider: PackageMetadataProvider
+    ) {
         self.configuration = configuration
         self.observabilityScope = observabilityScope
         self.storageContainer = (storage, false)
         self.collectionProviders = collectionProviders
         self.metadataProvider = metadataProvider
-        self.searchProvider = customSearchProvider ?? StorageBackedPackageSearchProvider(storage: storage.collections)
```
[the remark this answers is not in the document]
yim-lee: It's an extra layer that in hindsight seems redundant and makes things seem more convoluted than needed.

Sources/PackageCollections/PackageIndex.swift
yim-lee: https://github.com/apple/swift-package-manager/pull/4010/commits/cd27dc5fc04e4392ca27e8f487634a794cf98c07

Sources/PackageCollections/PackageCollections.swift
yim-lee: I want to keep `customMetadataProvider` internal though

yim-lee:
@swift-ci please smoke test

yim-lee:

Sources/PackageCollections/Model/PackageList.swift
```diff
@@ -0,0 +1,26 @@
...
+ Copyright (c) 2022 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+ */
+
+extension PackageCollectionsModel {
+    /// A representation of paginated list of packages.
+    public struct PaginatedPackageList {
```
tomerd, line 13: cool

Sources/PackageCollections/PackageCollections.swift
```diff
@@ -425,7 +417,7 @@ public struct PackageCollections: PackageCollectionsProtocol {
                         let metadata = Model.PackageMetadata(
                             package: Self.mergedPackageMetadata(package: packageSearchResult.package, basicMetadata: basicMetadata),
                             collections: packageSearchResult.collections,
-                            provider: PackageMetadataProviderContext(authTokenType: authTokenType, isAuthTokenConfigured: isAuthTokenConfigured)
+                            provider: provider
```
tomerd, line 420: nice cleanup ^^

Sources/PackageCollections/PackageIndex+Configuration.swift
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// MARK: - Utility
+
+private extension PackageIndexConfiguration {
+    init(_ from: StorageModel.Index) throws {
+        let url: Foundation.URL?
+        switch from.url {
+        case .none:
+            url = nil
```
tomerd, line 117: a bit surprised that URL is optional in this context, but I assume there are reasons when this makes sense?

Sources/PackageCollections/PackageIndex+Configuration.swift
```diff
@@ -0,0 +1,149 @@
...
+                throw SerializationError.invalidPath(path)
+            }
+        }
+
+        self.init(
+            url: url,
+            searchResultMaxItemsCount: from.searchResultMaxItemsCount,
+            cacheDirectory: cacheDirectory,
+            cacheTTLInSeconds: from.cacheTTLInSeconds,
+            cacheSizeInMegabytes: from.cacheSizeInMegabytes
```
tomerd, line 141: nit:  cacheMaxSizeInMegabytes

Sources/PackageCollections/PackageIndex.swift
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    func getPackageMetadata(
+        identity: PackageIdentity,
+        location: String?,
+        callback: @escaping (Result<PackageCollectionsModel.PackageMetadata, Error>) -> Void
+    ) {
+        self.runIfConfigured(callback: callback) { url, callback in
+            if let cached = try? self.cache?.get(key: identity.description),
+               cached.dispatchTime + DispatchTimeInterval.seconds(self.configuration.cacheTTLInSeconds) > DispatchTime.now() {
```
tomerd, line 68: nit: flipping the condition terms over can save roundtrip to the cache when expired

Sources/PackageCollections/PackageIndex.swift
```diff
@@ -0,0 +1,250 @@
...
+                return callback(.success((package: cached.package, collections: [], provider: self.createContext(host: url.host, error: nil))))
+            }
+            
+            // TODO: rdar://87582621 call package index's get metadata API
+            let metadataURL = url.appendingPathComponent("packages").appendingPathComponent(identity.description)
+            self.httpClient.get(metadataURL) { result in
```
tomerd, line 74: for later PRs: we would probably want to set a retry / exponential backoff strategy

Sources/PackageCollections/PackageIndex.swift
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    func listPackages(
+        offset: Int,
+        limit: Int,
+        callback: @escaping (Result<PackageCollectionsModel.PaginatedPackageList, Error>) -> Void
+    ) {
+        self.runIfConfigured(callback: callback) { url, callback in
+            // TODO: rdar://87582621 call package index's list API
+            guard let listURL = URL(string: url.appendingPathComponent("packages").absoluteString + "?offset=\(offset)&limit=\(limit)") else {
```
tomerd, line 138: nit: use URLComponent to construct the query arguments

Sources/PackageCollections/PackageIndexAndCollections.swift
```diff
@@ -0,0 +1,300 @@
...
+                callback(.success(
+                    PackageCollectionsModel.PackageMetadata(
+                        package: metadataResult.package,
+                        collections: collectionsResult.success?.collections ?? [],
+                        provider: metadataResult.provider
+                    )
+                ))
+            case .failure(let indexError):
+                switch collectionsResult {
+                case .success(let metadataResult):
```
tomerd, line 215: considers switching over both results to avoid the nested switch, e.g.

```
switch (indexResult, collectionsResult) {
case (.success(let metadataResult), _):
 ...
case (.failure(let indexError), .success(let metadataResult)):
...
case (.failure(let indexError), .failure(let collectionsError))::
...
}
```

Sources/PackageCollections/Providers/GitHubPackageMetadataProvider.swift
```diff
@@ -167,23 +168,72 @@ struct GitHubPackageMetadataProvider: PackageMetadataProvider {
...
-    func getAuthTokenType(for location: String) -> AuthTokenType? {
-        guard let baseURL = Self.apiURL(location) else {
+    
+    private func errorCallback(
+        _ error: Error,
+        apiHost: String?,
+        callback: @escaping (Result<Model.PackageBasicMetadata, Error>, PackageMetadataProviderContext?) -> Void
+    ) {
+        callback(.failure(error), self.createContext(apiHost: apiHost, error: error))
+    }
```
tomerd, line 185: would we need add something similar to this to the index client, or perhaps because it has no planned auth we dont really need to?

tomerd (approved):
this is great. some nits / ideas inline

Sources/PackageCollections/PackageIndex+Configuration.swift
yim-lee: This is to handle the case when there is no configuration at all or when user wants to disable the feature.

Sources/PackageCollections/PackageIndex.swift
yim-lee: I don't follow?

tomerd: hmm, my bad. read it wrong

Sources/PackageCollections/Providers/GitHubPackageMetadataProvider.swift
yim-lee: > perhaps because it has no planned auth we dont really need to?

That's right. Package index client sets `PackageMetadataProviderContext` too, but since its errors are never authn related they don't need to be added to the context.

Sources/PackageCollections/PackageIndex.swift
yim-lee: ok

yim-lee:
@swift-ci please smoke test

yim-lee:
@swift-ci please smoke test

neonichu (approved):

yim-lee:

yim-lee:
@swift-ci please smoke test

PR #4012, opened 2022-01-13
[5.6] dependency resolution performance (#4007)

tomerd:
5.6 cherry pick of #4007

Explanation:

Fixed dependency resolution issues causing failure on edge cases such with pre-release versions. as a result of that fix, the bounds computation done as part of the resolution is no longer requires which removes a significant performance bottleneck. this PR removes the bounds computation yielding 3x performance improvement on medium to large projects. 

The bounds computation was originally added in [aee02eb](https://github.com/apple/swift-package-manager/commit/aee02eb246442bcc28119101b89feee1d54110b4) in an attempt to improve diagnostics when graph fail to be resolve. https://github.com/apple/swift-package-manager/pull/3985 changes how the underlying data for these diagnostics is captured which yield even better diagnostics. as such this PR is purely about performance gain by not running that expensive calculation any more since it no longer has any functional value.

Scope of Issue: 

This impacts the performance of package dependency resolution, triggered by `swift package update` or when loading the package in Xcode.

Reason for Nominating to 5.6 : 

This yield significant improvement in performance.

Risk: 

Low risk; The risky change was https://github.com/apple/swift-package-manager/pull/3985 which is already part of 5.6. This PR is a followup realization that because of https://github.com/apple/swift-package-manager/pull/3985 we do not need to run that expensive bounds computation at all.

Reviewed By: @neonichu @abertelrud 

Automated Testing: Unit test, package compatibility test suite

Dependencies: None

Impact on CI: None

How to Verify:  Resolve [non-trivial] package dependencies and make sure the resolved versions (eg content of  Package.resolved) are identical but that the performance is better

### DIFF
--- a/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
+++ b/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
@@ -1035,13 +1035,6 @@ internal final class PubGrubPackageContainer {
     /// Reference to the pins map.
     private let pinsMap: PinsStore.PinsMap
 
-    /// The map of dependencies to version set that indicates the versions that have had their
-    /// incompatibilities emitted.
-    private var emittedIncompatibilities = ThreadSafeKeyValueStore<DependencyResolutionNode, VersionSetSpecifier>()
-
-    /// Whether we've emitted the incompatibilities for the pinned versions.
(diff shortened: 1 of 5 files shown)


tomerd:
@swift-ci test

tomerd:
@swift-ci test linux (ci infra issue)

tomerd:
@swift-ci test linux

tomerd:
@swift-ci please test

abertelrud (approved):

airspeedswift (approved):

PR #4014, opened 2022-01-13
Package command plugins should have the same initial working directory as the invocation

abertelrud:
### Motivation

This allows relative paths passed as parameters to work.  For build tool plugins we pass in the package directory so that relative paths in configuration files work there as well.  This also matches the behavior of many build systems.  The proposal didn't specify the initial working directory and usually all paths are absolute, but having predictable initial working directories follows the principle of least surprise.

### Changes

- pass through the working directory as a parameter to the plugin invocation
- for build tool invocations, make it the package directory
- for command invocations, keep the original working directory

rdar://86831942


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #4017, opened 2022-01-13
Limit concurrent HTTP requests when downloading binary dependencies

plu:
Limit the concurrent HTTP requests for downloading binary dependencies to 8 concurrent requests.

### Motivation:

We have a fairly large Package.swift with 50 binary dependencies. In total the size of these 50 zip files is 400 MB. On a slower internet connection (16 MBit) the `swift package resolve` command errors out with some HTTP timeouts. It seems the main problem is that all 50 HTTP requests are executed at the same time.

### Modifications:

Using some `DispatchSemaphore` to limit the concurrently downloaded binary dependencies to 8 concurrent requests. 

### Result:

In our project with the 50 binary dependencies it successfully completes the `swift package resolve` command. It's still open for discussion if the `max 8 concurrent downloads` is the right number. Also not sure if we should introduce another queue for this, reusing `callbackQueue` doesn't feel exactly right. However I found another [TODO](https://github.com/apple/swift-package-manager/blob/main/Sources/Basics/HTTPClient.swift#L152-L153) that's also just using `callbackQueue` and wondering if there should be a separate queue.

### DIFF
--- a/Sources/Workspace/Workspace.swift
+++ b/Sources/Workspace/Workspace.swift
@@ -2270,8 +2270,12 @@ extension Workspace {
             }
         }
 
+        // download max n files concurrently
+        let semaphore = DispatchSemaphore(value: 8)
+
         // finally download zip files, if any
         for artifact in (zipArtifacts.map{ $0 }) {
+            semaphore.wait()
             group.enter()
             defer { group.leave() }
 
@@ -2297,7 +2301,10 @@ extension Workspace {
                         totalBytesToDownload: totalBytesToDownload)
                 },
                 completion: { downloadResult in
-                    defer { group.leave() }
+                    defer {
+                        group.leave()
+                        semaphore.signal()
+                    }
 
                     switch downloadResult {
                     case .success:
--- a/Sources/Workspace/Workspace.swift
+++ b/Sources/Workspace/Workspace.swift
@@ -2291,6 +2291,7 @@ extension Workspace {
             headers.add(name: "Accept", value: "application/octet-stream")
             var request = HTTPClient.Request.download(url: artifact.url, headers: headers, fileSystem: self.fileSystem, destination: archivePath)
             request.options.authorizationProvider = self.authorizationProvider?.httpAuthorizationHeader(for:)
+            request.options.retryStrategy = .exponentialBackoff(maxAttempts: 3, baseDelay: .milliseconds(50))
             request.options.validResponseCodes = [200]
             self.httpClient.execute(
                 request,
--- a/Sources/Workspace/Workspace.swift
+++ b/Sources/Workspace/Workspace.swift
@@ -2271,7 +2271,7 @@ extension Workspace {
         }
 
         // download max n files concurrently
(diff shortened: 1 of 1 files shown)


plu:
See also: https://bugs.swift.org/browse/SR-15725

plu:
@swift-ci Please smoke test

Sources/Workspace/Workspace.swift
```diff
@@ -2270,8 +2270,12 @@ extension Workspace {
             }
         }
 
+        // download max n files concurrently
+        let semaphore = DispatchSemaphore(value: 8)
```
tomerd, line 2274: why 8?

tomerd:
thanks @plu, couple of follow up questions

Sources/Workspace/Workspace.swift
tomerd: we have `Concurrency.maxOperations` which maps to active cores, maybe we should use that or at least use that as the max ie `min(8, Concurrency.maxOperations)` if 8 is the right number

plu: 8 was just a random number that I came up with. `Concurrency.maxOperations` makes a lot more sense, thank you for the hint!

Sources/Workspace/Workspace.swift
```diff
@@ -2270,8 +2270,12 @@ extension Workspace {
...
+
         // finally download zip files, if any
         for artifact in (zipArtifacts.map{ $0 }) {
+            semaphore.wait()
```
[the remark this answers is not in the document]
plu: Hmm, maybe we should do `group.enter()` and `semaphore.wait()` only once after the `guard`, when we're sure we're actually executing the asynchronous HTTP request? Probably I'm just missing some detail about the lines before the `guard` though, not understanding the full picture why they also should be part of the `DispatchGroup`. What do you think?

plu:
Thank you for the quick review @tomerd. I made one change already, and had another question related to the code around the `DispatchGroup`.

Sources/Workspace/Workspace.swift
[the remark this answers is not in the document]
plu: I just pushed 97ccc8d - which is what I meant with my earlier comment. If this is wrong for some reason that I don't see yet, please let me know and I'll revert it and make the change you suggested in your first comment.

tomerd:
@swift-ci please smoke test

tomerd (approved):

tomerd:
@plu would you please also create a cherry-pick into the release/5.6 branch

plu:
Thank you for the review/approval, @tomerd. Here's the PR that cherry-picks the commit into `release/5.6`: #4024

plu:
Is this something that Xcode is using directly, or should I file some radar for Xcode? I'm asking because we're also facing this issue when we do `File -> Packages -> Resolve Package Versions` in Xcode.

tomerd:
@plu this will be included in Xcode in the next release cycle

PR #4018, opened 2022-01-13
[5.6] Sandbox blocks output to default plugin output directory when it's under `<pkgdir>/.build`

abertelrud:
This is the 5.6 nomination of #4009.

**Explanation:**  The sandbox rules introduced in #3996 made the entire package directory read-only, but that isn't appropriate when `.build` is inside the package directory.  Swapping the order so that the sandbox applies the allow-write rules after the deny-write rules allows `.build` (or whatever the output directory is) to be written to while keeping the rest of the package directory readonly to the plugin.

**Scope of Issue:**  This affects SwiftPM users who run command plugins that write to the plugin working directory and who are not customizing the intermediate-files directory using the `--build-path` option.

**Reason for Nominating to 5.6:**  This is a common use case in new functionality (command plugins).

**Risk:**  Low — although the sandbox is used for manifest loading as well as running plugins, the ability to add custom writable locations is only used in the new functionality of command plugins.

**Reviewed By:**  @tomerd

**Automated Testing:**  A new unit test covers this case.

**Dependencies:**  None

**Impact on CI:**  None

**How to Verify:**  Use a command plugin that writes intermediate files to the plugin working directory (as opposed to /tmp) and check that it works even when the SwiftPM intermediates directory is inside the package directory.

rdar://87417780

### DIFF
--- a/Sources/Basics/Sandbox.swift
+++ b/Sources/Basics/Sandbox.swift
@@ -21,8 +21,8 @@ public enum Sandbox {
     /// - Parameters:
     ///   - command: The command line to sandbox (including executable as first argument)
     ///   - strictness: The basic strictness level of the standbox.
-    ///   - writableDirectories: Paths under which writing should be allowed.
-    ///   - readOnlyDirectories: Paths under which writing should be denied, even if they would have otherwise been allowed by either the strictness level or paths in `writableDirectories`.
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please test

tomerd (approved):

airspeedswift (approved):

PR #4015, opened 2022-01-13
[5.6] Mark the SwiftPM 5.6 branch as a release branch

abertelrud:
Mark the SwiftPM 5.6 branch as a release branch.

### Motivation

As happens with every release, the development bit in SwiftPM needs to be turned off in the release branch so that it disallows use of 999.0 package manifest tools versions.  This also causes `swift package --version` to show the version without a `-dev` suffix.

### Modifications

- change `development` from `true` to `false` in SwiftVersion.swift

rdar://87535197

PR #4019, opened 2022-01-13
Fix some availability annotations in PackageDescription APIs that say "999"

abertelrud:
These should be marked as 5.6, where they were introduced.  They were missed because they used "999" instead of "999.0" (both are valid) and therefore didn't show up in the too-narrowly-focus grep to change them earlier.

### Motivation:

When marked as 999.0, these APIs will be available only on main SwiftPM and not on release branches.

### Modifications:

- change the "999" to "5.6" for relevant PackageDescription entries

### Result:

Packages can use these APIs even in the release build of SwiftPM 5.6.

This has already been changed in the 5.6 branch in #4015 and this is just the back-cherry-pick to main.

### DIFF
--- a/Sources/PackageDescription/PackageDependency.swift
+++ b/Sources/PackageDescription/PackageDependency.swift
@@ -488,7 +488,7 @@ extension Package.Dependency {
     /// - Parameters:
     ///     - id: The identity of the package.
     ///     - version: The minimum version requirement.
-    @available(_PackageDescription, introduced: 999)
+    @available(_PackageDescription, introduced: 5.6)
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift smoke test

tomerd (approved):

tomerd (changes requested):
actually these should not be included in 5.6, they are still not a public feature

abertelrud:
I see — thanks for pointing that out.  I will remove them then also from https://github.com/apple/swift-package-manager/pull/4015

PR #4021, opened 2022-01-14
gracefully handle pre-existing binary artifact from failed downloads

tomerd:
motivation: when binary artifact download fails it may leave behind a partial or otherwise corrupt file

changes:
* remove pre-existing archive file before attempting to download the same file again
* add test

### DIFF
--- a/Tests/WorkspaceTests/WorkspaceTests.swift
+++ b/Tests/WorkspaceTests/WorkspaceTests.swift
@@ -6502,6 +6502,131 @@ final class WorkspaceTests: XCTestCase {
         }
     }
 
+    func testArtifactDownloadArchiveExists() throws {
+        let sandbox = AbsolutePath("/tmp/ws/")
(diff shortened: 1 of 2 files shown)


tomerd:
@swift-ci please smoke test

Tests/WorkspaceTests/WorkspaceTests.swift
```diff
@@ -6502,6 +6502,131 @@ final class WorkspaceTests: XCTestCase {
         }
     }
 
+    func testArtifactDownloadArchiveExists() throws {
+        let sandbox = AbsolutePath("/tmp/ws/")
+        let fs = InMemoryFileSystem()
+        // this relies on internal knowledge of the destination path construction
```
abertelrud, line 6508: This seems fine for a unit test.  We could always extract into a static property for the subpath if we feel it's important enough.

abertelrud (approved):

PR #4022, opened 2022-01-14
[5.6] Package command plugins should have the same initial working directory as the invocation

abertelrud:
5.6 nomination cherry-pick of https://github.com/apple/swift-package-manager/pull/4014

**Explanation:**  The initial working directory of invoked plugin processes isn't defined, but is currently set to the cache directory.  For command plugins this is not a particularly useful default, especially since any relative paths passed as command line arguments should be relative to whatever working directory the command is invoked from (which is not necessarily the package directory).  Before command plugins initially become in SwiftPM 5.6, we should fix this to use the original working directory of the `swift package <verb>` invocation, and as follow-up we should clarify this in SE-0332.

**Scope of Issue:**  This applies to command plugins that accept relative paths as arguments.  SE-0332 leaves formal declaration of plugin parameters to the future, but specifies that argument strings should be passed on to the plugin, and it's likely that some plugins will want to receive relative paths as arguments.

**Reason for Nominating to 5.6:**  We should start with a useful behavior in the first release of SwiftPM with command plugins rather than change the behavior later.

**Risk:**  Very low. By definition there are no current command plugins that depend on the current behavior, and the code change is straightforward.

**Reviewed By:**  @tomerd

**Automated Testing:**  A new unit test checks that the initial directory is the expected one.

**Dependencies:**  None

**Impact on CI:**  None

**How to Verify:**  Write a command plugin and do file operations based on the plugin process' initial working directory.  It should be the directory from which the `swift package <verb>` invocation was made.  See the unit test as an example.

rdar://86831942

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -1050,6 +1050,7 @@ extension SwiftPackageTool {
                 package: package,
                 buildEnvironment: buildEnvironment,
                 scriptRunner: pluginScriptRunner,
+                workingDirectory: swiftTool.originalWorkingDirectory,
                 outputDirectory: outputDir,
                 toolSearchDirectories: toolSearchDirs,
                 toolNamesToPaths: toolNamesToPaths,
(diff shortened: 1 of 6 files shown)


abertelrud:
@swift-ci please test

tomerd (approved):

airspeedswift (approved):

PR #4028, opened 2022-01-14
Don't emit a "this package dependency is unused" warning for packages that vend command plugins

abertelrud:
An existing warning about dependencies on packages from which no products are used shouldn't apply if the package has any command plugin products.  The warning assumes that the only use of products is to link against them or use them as build tool plugins, but that's not correct for command plugins, where the dependency is only used to make the plugin available for the user to invoke.

### Motivation

With command plugins it's perfectly reasonable to depend on a package without explicitly having a reference to any of its products, so suppress the warning in that case.

### Modifications:

- added a check to skip the warning if the depended-on-package vends at least one command plugin product
- add a unit test to check the absence of this diagnostic, and the continued presence of diagnostics about unused built tool plugins

Existing unit tests check that the warning still shows up for regular products.

rdar://86787186

### DIFF
--- a/Sources/PackageGraph/PackageGraph+Loading.swift
+++ b/Sources/PackageGraph/PackageGraph+Loading.swift
@@ -182,11 +182,27 @@ private func checkAllDependenciesAreUsed(_ rootPackages: [ResolvedPackage], obse
                 continue
             }
 
+            // Skip this check if this dependency contains a command plugin product.
+            if dependency.products.contains(where: \.isCommandPlugin) {
+                continue
+            }
+
+            // Otherwise emit a warning if none of the dependency package's products are used.
             let dependencyIsUsed = dependency.products.contains(where: productDependencies.contains)
             if !dependencyIsUsed && !observabilityScope.errorsReportedInAnyScope {
                 observabilityScope.emit(.unusedDependency(dependency.identity.description))
             }
         }
+    }
+}
+
+fileprivate extension ResolvedProduct {
+    /// Returns true if and only if the product represents a command plugin target.
+    var isCommandPlugin: Bool {
+        guard type == .plugin else { return false }
+        guard let target = underlyingProduct.targets.compactMap({ $0 as? PluginTarget }).first else { return false }
+        guard case .command = target.capability else { return false }
+        return true
     }
 }
 
--- a/Tests/FunctionalTests/PluginTests.swift
+++ b/Tests/FunctionalTests/PluginTests.swift
@@ -468,4 +468,156 @@ class PluginTests: XCTestCase {
             }
         }
     }
+
+    func testUnusedPluginProductWarnings() throws {
+        // Test the warnings we get around unused plugin products in package dependencies.
+        try testWithTemporaryDirectory { tmpPath in
+            // Create a sample package that uses three packages that vend plugins.
+            let packageDir = tmpPath.appending(components: "MyPackage")
+            try localFileSystem.createDirectory(packageDir, recursive: true)
+            try localFileSystem.writeFileContents(packageDir.appending(component: "Package.swift"), string: """
+                // swift-tools-version: 5.6
+                import PackageDescription
+                let package = Package(
+                    name: "MyPackage",
+                    dependencies: [
+                        .package(name: "BuildToolPluginPackage", path: "VendoredDependencies/BuildToolPluginPackage"),
+                        .package(name: "UnusedBuildToolPluginPackage", path: "VendoredDependencies/UnusedBuildToolPluginPackage"),
+                        .package(name: "CommandPluginPackage", path: "VendoredDependencies/CommandPluginPackage")
+                    ],
+                    targets: [
+                        .target(
+                            name: "MyLibrary",
+                            path: ".",
+                            plugins: [
+                                .plugin(name: "BuildToolPlugin", package: "BuildToolPluginPackage")
+                            ]
+                        ),
+                    ]
+                )
+                """)
+            try localFileSystem.writeFileContents(packageDir.appending(component: "Library.swift"), string: """
+                public var Foo: String
+                """)
+
+            // Create the depended-upon package that vends a build tool plugin that is used by the main package.
+            let buildToolPluginPackageDir = packageDir.appending(components: "VendoredDependencies", "BuildToolPluginPackage")
+            try localFileSystem.createDirectory(buildToolPluginPackageDir, recursive: true)
+            try localFileSystem.writeFileContents(buildToolPluginPackageDir.appending(component: "Package.swift"), string: """
+                // swift-tools-version: 5.6
+                import PackageDescription
+                let package = Package(
+                    name: "BuildToolPluginPackage",
+                    products: [
+                        .plugin(
+                            name: "BuildToolPlugin",
+                            targets: ["BuildToolPlugin"])
+                    ],
+                    targets: [
+                        .plugin(
+                            name: "BuildToolPlugin",
+                            capability: .buildTool(),
+                            path: ".")
+                    ]
+                )
+                """)
+            try localFileSystem.writeFileContents(buildToolPluginPackageDir.appending(component: "Plugin.swift"), string: """
+                import PackagePlugin
+                @main struct MyPlugin: BuildToolPlugin {
+                    func createBuildCommands(context: PluginContext, target: Target) throws -> [Command] {
+                        return []
+                    }
+                }
+                """)
+
+            // Create the depended-upon package that vends a build tool plugin that is not used by the main package.
+            let unusedBuildToolPluginPackageDir = packageDir.appending(components: "VendoredDependencies", "UnusedBuildToolPluginPackage")
+            try localFileSystem.createDirectory(unusedBuildToolPluginPackageDir, recursive: true)
+            try localFileSystem.writeFileContents(unusedBuildToolPluginPackageDir.appending(component: "Package.swift"), string: """
+                // swift-tools-version: 5.6
+                import PackageDescription
+                let package = Package(
+                    name: "UnusedBuildToolPluginPackage",
+                    products: [
+                        .plugin(
+                            name: "UnusedBuildToolPlugin",
+                            targets: ["UnusedBuildToolPlugin"])
+                    ],
+                    targets: [
+                        .plugin(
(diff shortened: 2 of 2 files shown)


abertelrud:
Putting into draft mode because I think we might want to treat build tool plugins and command plugins differently.  For the former we might still want to warn if there is no use at all of the plugin.

abertelrud:
@swift-ci smoke test

tomerd (approved):

abertelrud:
I should also comment that if we revisit package how dependencies are expressed in the future, I think it would be useful to distinguish between different kinds of dependencies.  Dependencies on plugins (and indeed host-side tools) are qualitatively a bit different than regular "the code that implements tool X depends on the code in library Y" type of dependencies.  As SwiftPM becomes more advanced, this level of nuance is going to become more required.  I know some of us have discussed this in the past but as it relates to the showing of this warning, I wanted express this sentiment to anyone who happens to be reading as something to consider.

SDGGiesbrecht:
This is incompatible with target‐based resolution, at least in its current form. (The related test failed when I updated and tried activating the resolution switch.)

From reading [SE‐0332](https://github.com/apple/swift-evolution/blob/main/proposals/0332-swiftpm-command-plugins.md), it appears the intent to use a dependency’s command is not explicitly stated in the manifest?

Then there are several options for moving forward:

1. Revise SE‐0332 so that there is an explicit reference somewhere in the manifest.
    - Implementation: Just adjust `Manifest.targetsRequired(for:)` to append the stated command dependencies when the product filter is `.everything`.
    - Efficiency: Resolves only what is desired.
    - Pins: Commands are pinned in the same file as everything else.
    - Compatibility: Desired commands have to fit into the same dependency graph as the package itself.

2. Resolve command dependencies lazily by waiting until they are called from the command line, resolving either...
    1. ...as a distinct dependency graph.
        - Implementation: Take the plugin argument to `swift package`, search for it in the first‐level dependency manifests (which are loaded at the `Manifest` level anyway for their diagnostic value), then use that package and product as the root of a separate dependency graph, doing all the work in a separate directory.
        - Efficiency: Resolves only what is needed and only when it is needed.
        - Pins: Commands are not pinned (unless we design a separate pinning system for them).
        - Compatibility: Commands work fine even with mutually incompatible dependency graphs.
    2. ...integrated with the main dependency graph.
        - Implementation: Take the plugin argument to `swift package` and add it to the first‐level product filters just like we currently do with `swift run`.
        - Efficiency: Resolves only what is needed and only when it is needed, but will re‐resolve when switching between commands.
        - Pins: Pins may change when commands are used.
        - Compatibility: Commands have to fit into the same dependency graph as the package, but not with each other.

3. Resolve command dependencies greedily for first‐level dependencies.
    - Implementation: Add an associated value to `ProductFilter.specific(Set<String>)` as `includeCommands: Bool` and overhaul `Manifest.dependenciesRequired(for:)` accordingly. `.everything` (roots) lead to `.specific([/*...*/], includeCommands: true)` (first‐level dependencies), which in turn lead to `.specific([/*...*/], includeCommands: false)` (deep dependencies).
    - Efficiency: Resolves even undesired commands that will never be used.
    - Pins: Commands are pinned in the same file as everything else.
    - Compatibility: Even undesired commands may cause incompatibilities in the main package’s dependency graph.

P.S. I am sorry I failed to notice this while SE‐0332 was still under review.

tomerd:
@abertelrud @SDGGiesbrecht does it help that command plugins are only applicable to the top level package(s)?

SDGGiesbrecht:
Commands in the root package should be fine as‐is (roots resolve everything anyway). But that is not where this test is looking. It is expecting a command from a first‐level dependency.

Since a command is a product, legacy resolution resolves and loads it just as it does any other product (presumably even from deep dependencies unless something was done to prevent that).

But with target‐based resolution, a product is unreachable from the root node unless something references it. Thus a command in this position is neither resolved nor loaded, nor usable at the moment. The code in this pull request that silences the warning does so by checking for commands in the graph. But since none will have been loaded (ever), it assumes the dependency is not a command (even though that might have been its intent) and throws the same warning it does for normal unused dependencies. This extra warning is what causes the test to fail.

Something somewhere needs to add that missing reference in order to actually make the command available. The different ways of doing so have the various trade‐offs I listed above.

SDGGiesbrecht:
To unblock both features quickly, I will submit a PR shortly that implements option 3, which seems to be what the proposal describes.

(Then I will cross my fingers that none of my dependencies ever add any command plugins with extra dependencies...)

abertelrud:
Sorry I'm late to this conversation, and that I didn't consider the target-based dependencies when working on this PR.

Based on the intended uses of command plugins, I don't think we can require that a package specifically references them.  So I think option 3 is the only feasible one, and I think that considering each command plugin product as implicitly being referenced is conceptually right based on the design.

We can't for example do it on-demand based on what plugin is being invoked, since there are features such as being able to list the available plugins.

> (Then I will cross my fingers that none of my dependencies ever add any command plugins with extra dependencies...)

Today the plugins themselves cannot depend on other packages, though any other targets they depend on can do so.  But if any command plugin is implicitly reachable, then wouldn't any of its dependencies would also be part of the graph, and therefore considered reachable?  It seems to me that the only required special case here is to consider any plugin product to be implicitly reachable.

abertelrud:
Perhaps I'm misunderstanding something about how target-based dependencies work.  My understanding was that even with target-based dependencies, any package directly referenced by the root would be resolved and loaded, but indirect dependencies might not.  Is that incorrect, and could it be that even a directly specified package dependency might not get loaded?

In that case, yes, the possibility of command plugins could indeed cause more package loading than before, and now I see the issue you describe.  I do think option 3 is the better one in the short term.

In the long run, I'm not that fond of reusing the package dependencies as a way of depending on commands, since it mixes the build-time and runtime graphs.  So I think a future iteration may well want to separate out those dependencies or mark them in some way, which might help get back the ability to prune out first-level package dependencies.

SDGGiesbrecht:
> So I think option 3 is the only feasible one, and I think that considering each command plugin product as implicitly being referenced is conceptually right based on the design.

I have mostly finished implementing this; I just have some final clean‐up and testing yet to do.

> It seems to me that the only required special case here is to consider any plugin product to be implicitly reachable.

Yes, that is correct. That parenthesized comment was meant as a user, and has no bearing on the implementation.

What I meant was that I have been encouraging developers for years to group multiple products into the same package and not worry about it causing extra dependencies. But now I have to start training them that command plugins are an exception because they _do_ cause their dependencies to be inherited by the other products in the package. Thus it will be bad practice in my book to combine commands with other products into one package.

> My understanding was that even with target-based dependencies, any package directly referenced by the root would be resolved and loaded, but indirect dependencies might not. Is that incorrect, and could it be that even a directly specified package dependency might not get loaded?

Some cheating is being done in that an unused direct dependency is fetched anyway and its _manifest_ is loaded (as a `Manifest`). This is done for the sake of providing helpful diagnostics when the user is only halfway through hooking up a dependency. Maybe this it what you were thinking of?

But at the level of resolution and loading (`Package`, `ResolvedPackage`, etc.), such a package is _empty_, lacking any products or targets. At these levels, packages only load what they are asked to load. Unreferenced things do not exist. Indeed, if you tried to load something not reachable from the top, it could be broken, because it might depend on things that were never fetched since the top did not ask for them. (Targets not attached to any product have been handled this way since before I ever got involved.)

> In that case, yes, the possibility of command plugins could indeed cause more package loading than before, and now I see the issue you describe. I do think option 3 is the better one in the short term.
> 
> In the long run, I'm not that fond of reusing the package dependencies as a way of depending on commands, since it mixes the build-time and runtime graphs. So I think a future iteration may well want to separate out those dependencies or mark them in some way, which might help get back the ability to prune out first-level package dependencies.

I came to more‐or‐less the same conclusions about what is best for the short term and what might be better in the long run.

abertelrud:
> I have mostly finished implementing this; I just have some final clean‐up and testing yet to do.

Thank you!

> > It seems to me that the only required special case here is to consider any plugin product to be implicitly reachable.
> 
> Yes, that is correct. That parenthesized comment was meant as a user, and has no bearing on the implementation.
> 
> What I meant was that I have been encouraging developers for years to group multiple products into the same package and not worry about it causing extra dependencies. But now I have to start training them that command plugins are an exception because they _do_ cause their dependencies to be inherited by the other products in the package. Thus it will be bad practice in my book to combine commands with other products into one package.

I would expect that most packages that provide commands would be geared toward the commands they produce.  This is probably good advice though to include in the documentation that I will be adding to this repository.

> > My understanding was that even with target-based dependencies, any package directly referenced by the root would be resolved and loaded, but indirect dependencies might not. Is that incorrect, and could it be that even a directly specified package dependency might not get loaded?
> 
> Some cheating is being done in that an unused direct dependency is fetched anyway and its _manifest_ is loaded (as a `Manifest`). This is done for the sake of providing helpful diagnostics when the user is only halfway through hooking up a dependency. Maybe this it what you were thinking of?

It might be.  I don't know specifically why I had that impression.  Now that you mention the part about the manifest, that makes sense.  It might also help a solution here, since the manifest will indicate if there are any command plugins.

> But at the level of resolution and loading (`Package`, `ResolvedPackage`, etc.), such a package is _empty_, lacking any products or targets. At these levels, packages only load what they are asked to load. Unreferenced things do not exist. Indeed, if you tried to load something not reachable from the top, it could be broken, because it might depend on things that were never fetched since the top did not ask for them. (Targets not attached to any product have been handled this way since before I ever got involved.)

Thank your for this explanation as well — that makes sense.

> > In that case, yes, the possibility of command plugins could indeed cause more package loading than before, and now I see the issue you describe. I do think option 3 is the better one in the short term.
> > In the long run, I'm not that fond of reusing the package dependencies as a way of depending on commands, since it mixes the build-time and runtime graphs. So I think a future iteration may well want to separate out those dependencies or mark them in some way, which might help get back the ability to prune out first-level package dependencies.
> 
> I came to more‐or‐less the same conclusions about what is best for the short term and what might be better in the long run.

For the long term, I think there are some very good reasons to have multiple layers of dependencies, not just for plugins but also for things like host-side tools version runtime dependencies (e.g. if my source generator and an unrelated library both use the same package, they shouldn't have to use the same version as the code never coexists).  At that point some of these implicit-node changes can hopefully be unwound.

PR #4030, opened 2022-01-14
improve dependency resolution diagnostics when resolving non-versioned dependencies

tomerd:
motivation:  when resolving graphs with non-versioned depedendencies (such as branch dependencies), the dependencies get "flattened" onto the root node which looses the original association with the dependency's parent yielding incorrect diagnostics

changes:
* more accuretly track the underlying dependency when computing root dependencies from non-versioned depedendencie
* peek into the "real" cause when producing depedency based diagnostics for the root node
* TBD: use top level package as root instead of synthesized root when appropriate 
* add a bunch of tests

rdar://87486528
rdar://64226666

### DIFF
--- a/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
+++ b/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
@@ -262,13 +262,13 @@ public struct PubgrubDependencyResolver {
         var overriddenPackages: [PackageReference: (version: BoundVersion, products: ProductFilter)] = [:]
 
         // The list of version-based references reachable via local and branch-based references.
-        // These are added as top-level incompatibilities since they always need to be statisfied.
+        // These are added as top-level incompatibilities since they always need to be satisfied.
         // Some of these might be overridden as we discover local and branch-based references.
-        var versionBasedDependencies: [DependencyResolutionNode: [VersionBasedConstraint]] = [:]
+        var versionBasedDependencies = OrderedDictionary<DependencyResolutionNode, [VersionBasedConstraint]>()
 
         // Process unversioned constraints in first phase. We go through all of the unversioned packages
         // and collect them and their dependencies. This gives us the complete list of unversioned
-        // packages in the graph since unversioned packages can only be refered by other
+        // packages in the graph since unversioned packages can only be referred by other
         // unversioned packages.
         while let constraint = constraints.first(where: { $0.requirement == .unversioned }) {
             constraints.remove(constraint)
@@ -370,7 +370,7 @@ public struct PubgrubDependencyResolver {
                     case .versionSet(let req):
                         for node in dependency.nodes() {
                             let versionedBasedConstraint = VersionBasedConstraint(node: node, req: req)
-                            versionBasedDependencies[node, default: []].append(versionedBasedConstraint)
+                            versionBasedDependencies[.root(package: constraint.package), default: []].append(versionedBasedConstraint)
                         }
                     case .revision:
                         constraints.append(dependency)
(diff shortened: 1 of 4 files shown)


tomerd:
@swift-ci please smoke test

neonichu (approved):

tomerd:
ci infra issues causing build timeout?

tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

tomerd:

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
```diff
@@ -957,11 +967,6 @@ private struct DiagnosticReportBuilder {
         return !lineNumbers.keys.contains(complex)
     }
 
-    // FIXME: This is duplicated and wrong.
-    private func isFailure(_ incompatibility: Incompatibility) -> Bool {
-        return incompatibility.terms.count == 1 && incompatibility.terms.first?.node.package.identity == .plain("<synthesized-root>")
-    }
```
tomerd, line 963: unrelated cleanup

tomerd:
not able to reproduce the package compatibility failures reported, could be from a different cause

tomerd:
package compatibility failures: https://github.com/apple/swift/pull/40877

tomerd:

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
```diff
@@ -132,10 +132,18 @@ public struct PubgrubDependencyResolver {
 
     /// Execute the resolution algorithm to find a valid assignment of versions.
     public func solve(constraints: [Constraint]) -> Result<[DependencyResolver.Binding], Error> {
-        let root = DependencyResolutionNode.root(package: .root(
-            identity: .plain("<synthesized-root>"),
-            path: .root
-        ))
+        // 👀 is this the right thing to do?
```
abertelrud, line 135: This seems like an improvement for now, especially for CLI.  For IDEs such as Xcode, where there can be multiple roots and also package dependencies that don't come from packages, I think we'll want to add more information to `PackageGraphRootInput` that can be used in diagnostics.  That should be easy for the IDE to provide and would let you know where the "root pointers" into the package graph are (either a set of root packages or some other external dependency on a package, such as from an iOS app Xcode target to a package).  That could of course be done in a separate PR.

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
```diff
@@ -262,13 +270,13 @@ public struct PubgrubDependencyResolver {
         var overriddenPackages: [PackageReference: (version: BoundVersion, products: ProductFilter)] = [:]
 
         // The list of version-based references reachable via local and branch-based references.
-        // These are added as top-level incompatibilities since they always need to be statisfied.
+        // These are added as top-level incompatibilities since they always need to be satisfied.
         // Some of these might be overridden as we discover local and branch-based references.
-        var versionBasedDependencies: [DependencyResolutionNode: [VersionBasedConstraint]] = [:]
+        var versionBasedDependencies = OrderedDictionary<DependencyResolutionNode, [VersionBasedConstraint]>()
```
abertelrud, line 275: I think switching to ordered collections is great where it makes sense — it will often help understandability if we can make the order of diagnostics stable from run to run, and also, when there's no better order, to show entities in the same order that users specify them (not always possible of course, but sometimes it is).

abertelrud (approved):

tomerd:
@swift-ci please smoke test

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
```diff
@@ -149,24 +149,21 @@ public struct PubgrubDependencyResolver {
...
+                    )
                     let diagnostic = try builder.makeErrorReport(for: rootCause)
-                    error = PubgrubError.unresolvable(diagnostic)
+                    return.failure(PubgrubError.unresolvable(diagnostic))
                 } catch {
                     // failed to construct the report, will report the original error
+                    return .failure(error)
                 }
             }
-
```
tomerd, line 161: error handling cleanup ^^

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
tomerd: also discussed with @neonichu and we agreed this is an improvement

PR #4031, opened 2022-01-15
[5.6] Don't emit "package dependency is unused" warning for packages containing command plugin products

abertelrud:
5.6 nomination cherry-picked from https://github.com/apple/swift-package-manager/pull/4028

**Explanation:**  If a package depends on another package that vends a command plugin, and if it does not use any other product from that package, then SwiftPM emits a warning saying that the dependency is unused.  This warning predates the introduction of command plugins, and is no longer correct when a package dependency vends command plugins, because the reason for the dependency might be to make those plugins available when using the package.  Since those command plugins could be invoked by the user at any time, this change elides the warning when the package dependency contains command plugin products (but not when it contains only unused build tool plugins, since the warning does apply in that case).

**Scope of Issue:**  This applies to packages that depend on other packages that vend command plugins, but where the top-level package does not depend on any other library from the package.

**Reason for Nominating to 5.6:**  When authoring a command plugin, it's not unusual for that plugin to be the only thing that other packages use from the package that defines it.  Feedback based on early use of command plugins shows that this warning causes confusion when it appears, since the command plugin _could_ be used at any time, it's just that SwiftPM doesn't know what the user is going to do.

**Risk:**  Low. The largest risk if the logic is wrong would be to incorrectly hide the warning in cases where it should be shown.  This is mitigated by the presence of the unit tests that check that this warning is emitted in the absence of command plugins.

**Reviewed By:**  @tomerd

**Automated Testing:**  A new unit test checks that 1) the warning is emitted for build tool plugins that are _not_ used, but 2) that this warning is not emitted for build tool plugins that _are_ used, and 3) that it is not emitted for command plugins.  Extant unit tests check that the warning is emitted in the older uses cases that do not involve plugins.

**Dependencies:**  None

**Impact on CI:**  None

**How to Verify:**  Write a command plugin that doesn't have any other products than a command plugin, and add a dependency on it from another package.  Verify that the warning isn't shown when the top-level package is used.

rdar://86787186

### DIFF
--- a/Sources/PackageGraph/PackageGraph+Loading.swift
+++ b/Sources/PackageGraph/PackageGraph+Loading.swift
@@ -182,11 +182,27 @@ private func checkAllDependenciesAreUsed(_ rootPackages: [ResolvedPackage], obse
                 continue
             }
 
+            // Skip this check if this dependency contains a command plugin product.
+            if dependency.products.contains(where: \.isCommandPlugin) {
+                continue
+            }
+
+            // Otherwise emit a warning if none of the dependency package's products are used.
             let dependencyIsUsed = dependency.products.contains(where: productDependencies.contains)
             if !dependencyIsUsed && !observabilityScope.errorsReportedInAnyScope {
                 observabilityScope.emit(.unusedDependency(dependency.identity.description))
             }
         }
+    }
+}
+
+fileprivate extension ResolvedProduct {
+    /// Returns true if and only if the product represents a command plugin target.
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please test

tomerd (approved):

abertelrud:
@swift-ci please test

airspeedswift (approved):

abertelrud:
Thanks!

PR #4032, opened 2022-01-16
SourceControl: avoid the builtin FS monitor

compnerd:
This disables the builtin FS monitor on the checkouts of the
dependencies.  This ensures that the global configuration of
the setting does not influence the repository local behavior
as that can cause problems on Windows. Due to the filesystem
behaviour on Windows, we cannot remove a directory with open
descriptors - which the monitor implies. This should resolve
the problem where a series of `swift package` commands would
result in failures.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -105,7 +105,11 @@ public struct GitRepositoryProvider: RepositoryProvider {
         // FIXME: Ideally we should pass `--progress` here and report status regularly.  We currently don't have callbacks for that.
         //
         // NOTE: Explicitly set `core.symlinks=true` on `git clone` to ensure that symbolic links are correctly resolved.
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please smoke test

abertelrud:
Based on some web searches it seems that this is a Windows-specific feature... is that right?  If not, does this change have any effect on other platforms, and should the disabling be conditional to Windows?

tomerd (approved):

tomerd:

neonichu (approved):

abertelrud (approved):

tomerd:
@compnerd do you want to cherry-pick this into 5.6 branch?

compnerd:
Ah, yeah, that we should do.  I'll put that up in a minute.

PR #4034, opened 2022-01-20
[5.6] SourceControl: avoid the builtin FS monitor (#4032)

compnerd:
This disables the builtin FS monitor on the checkouts of the
dependencies.  This ensures that the global configuration of
the setting does not influence the repository local behavior
as that can cause problems on Windows. Due to the filesystem
behaviour on Windows, we cannot remove a directory with open
descriptors - which the monitor implies. This should resolve
the problem where a series of `swift package` commands would
result in failures.

### Motivation:

### Modifications:

### Result:


compnerd:
@swift-ci please test

tomerd (approved):

PR #4037, opened 2022-01-21
Emit output and diagnostics from build tool plugin invocations as diagnostics

abertelrud:
Emit output and diagnostics from build tool plugin invocations as diagnostics.

### Motivation:

Printed output and emitted diagnostics were not being shown in the `swift build` CLI, and this made it difficult to see what build tool plugins were doing (note that this is distinct from command plugins, which are invoked differently and where output is already emitted).

### Modifications:

- make BuildOperation emit output text and diagnostics from the build tool plugin results
- add more fields to the build tool plugin result structure
- add a unit test

### Result:

Errors and warnings from a build tool plugin are emitted by default, and general `print()` output is emitted as `.info` in this PR.  A future PR should introduce a new `.output` log level that is emitted by default, without having to specify `-v`, but this will require more discussion about the form this output should take.

### DIFF
--- a/Sources/SPMBuildCore/PluginInvocation.swift
+++ b/Sources/SPMBuildCore/PluginInvocation.swift
@@ -240,6 +240,7 @@ extension PackageGraph {
                 let delegate = PluginDelegate(delegateQueue: delegateQueue)
 
                 // Invoke the build tool plugin with the input parameters and the delegate that will collect outputs.
+                let startTime = DispatchTime.now()
                 let success = try tsc_await { pluginTarget.invoke(
                     action: .createBuildToolCommands(target: target),
                     package: package,
@@ -256,17 +257,20 @@ extension PackageGraph {
                     callbackQueue: delegateQueue,
                     delegate: delegate,
                     completion: $0) }
+                let duration = startTime.distance(to: .now())
 
                 // Add a BuildToolPluginInvocationResult to the mapping.
-                if success {
-                    buildToolPluginResults.append(.init(
-                        plugin: pluginTarget,
-                        pluginOutputDirectory: pluginOutputDir,
-                        diagnostics: delegate.diagnostics,
-                        textOutput: String(decoding: delegate.outputData, as: UTF8.self),
-                        buildCommands: delegate.buildCommands,
-                        prebuildCommands: delegate.prebuildCommands))
-                }
+                buildToolPluginResults.append(.init(
+                    plugin: pluginTarget,
+                    pluginOutputDirectory: pluginOutputDir,
+                    package: package,
+                    target: target,
+                    succeeded: success,
+                    duration: duration,
+                    diagnostics: delegate.diagnostics,
+                    textOutput: String(decoding: delegate.outputData, as: UTF8.self),
+                    buildCommands: delegate.buildCommands,
+                    prebuildCommands: delegate.prebuildCommands))
             }
 
             // Associate the list of results with the target. The list will have one entry for each plugin used by the target.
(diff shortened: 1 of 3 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
Marking as draft because I'm going to see how hard it is to add that `output` level while doing this.  It doesn't look bad, and it's more correct.

Sources/Basics/Observability.swift
```diff
@@ -321,10 +330,12 @@ public struct Diagnostic: CustomStringConvertible {
                 return 0
             case .info:
                 return 1
-            case .warning:
+            case .output:
                 return 2
-            case .error:
+            case .warning:
```
abertelrud, line 335: Note, this numbering is internal and cannot affect clients, even though it shifts here.

Sources/Basics/Observability.swift
```diff
@@ -597,6 +608,8 @@ extension TSCBasic.Diagnostic {
             self = .init(message: .error(data), location: location)
         case .warning:
             self = .init(message: .warning(data), location: location)
+        case .output:
+            self = .init(message: .note(data), location: location)
```
abertelrud, line 612: Currently mapped to `note`, just as `info` is — changing this will require TSCBasic to add this case first.

abertelrud:
@swift-ci please smoke test

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -321,6 +325,18 @@ public struct BuildToolPluginInvocationResult {
...
+    public var package: ResolvedPackage
+
+    /// The target in that package to which the plugin was applied.
+    public var target: ResolvedTarget
+
+    /// If the plugin finished successfully.
+    public var succeeded: Bool
+
+    /// Duration in seconds of the plugin invocation.
+    public var duration: Double
```
tomerd, line 338: use Dispatch's TimeInterval?

tomerd (approved):

Sources/Commands/SwiftTool.swift
```diff
@@ -1128,6 +1128,8 @@ extension Basics.Diagnostic {
             writer.write("error: ", inColor: .red, bold: true)
         case .warning:
             writer.write("warning: ", inColor: .yellow, bold: true)
+        case .output:
+            writer.write("output: ", inColor: .gray, bold: true)
```
tomerd, line 1132: I wonder if output should be prefix free

abertelrud: That's a good point — it was the same with output from command plugins, initially that we prefixed and it got annoying.

Sources/SPMBuildCore/PluginInvocation.swift
abertelrud: Good point, is that the de facto standard for this in SwiftPM?  In Foundation it was typically NSTimeInterval, which was defined as being a double in seconds.

Sources/Basics/Observability.swift
tomerd: given the proposed number, output level would be filtered out from the default user facing output stream (>= warn). is that expected or is the idea output level is always "printed"

Sources/SPMBuildCore/PluginInvocation.swift
tomerd: yes for now (there is a stdlib type coming)

tomerd (commented):
we should discuss the new level's "printing" a bit more

abertelrud:
> we should discuss the new level's "printing" a bit more

Agreed.  One of the main purposes of this PR is to add the additional fields to the build results output struct for use by libSwiftPM clients.  I can split out the `output` level as a separate PR and also convert over BuildOperation etc to use it, and then we can arrive at the best approach before landing that.  I would then make this PR emit to `.info` again, as it did earlier.  Unless you disagree, that's what I suggest so that the other parts of this can land.

abertelrud:
Removed the `output` level from this PR and will open as a separate PR where we can discuss the right output format.  That keeps this one nominatable for 5.6, while making progress in main on the unifying the text stream output and the diagnostics in a way that preserves the order.

abertelrud:
Also changed the duration field to a `DispatchTimeInterval`.

abertelrud:
@swift-ci please smoke test

tomerd (approved):

abertelrud:
apple/swift#40954

abertelrud:
apple/swift#40954

abertelrud:
https://github.com/apple/swift/pull/40954
@swift-ci please smoke test

abertelrud:

abertelrud:
@swift-ci please smoke test

PR #4038, opened 2022-01-21
Fix Command Resolution

SDGGiesbrecht:
See #4028 for the impetus, particularly starting from [this comment](https://github.com/apple/swift-package-manager/pull/4028#issuecomment-1016071469).

This implements the option labeled “3” in that thread, attempting to exactly match the API and usage described in [SE‐0332](https://github.com/apple/swift-evolution/blob/main/proposals/0332-swiftpm-command-plugins.md). It does so by implicitly inserting every command plugin found in a first‐level dependency into the dependency graph.


Sources/PackageModel/Product.swift
```diff
@@ -241,26 +246,3 @@ extension ProductType: Codable {
         }
     }
 }
-
-extension ProductFilter: Codable {
```
SDGGiesbrecht, line 245: The old `Codable` implementation here could not be added to in a backwards compatible way. I just removed it to let compiler synthesize it.

I do not know where or if this is actually used. We stopped recording filters in the pins file a while ago. Changing it did not flag on any tests. Do you think there is anywhere we need to bump a format version?

Tests/PackageGraphPerformanceTests/DependencyResolverPerfTests.swift
```diff
@@ -174,7 +174,7 @@ extension ProductFilter: JSONSerializable, JSONMappable {
         switch self {
         case .everything:
             return "all".toJSON()
-        case .specific(let products):
+        case .specific(let products, _):
```
SDGGiesbrecht, line 177: Oops. I forgot to come back to this. I will look at the performance tests tomorrow.

Tests/WorkspaceTests/SourceControlPackageContainerTests.swift
```diff
@@ -510,6 +510,18 @@ class SourceControlPackageContainerTests: XCTestCase {
             )
         }
 
+        let v5_2ProductMappingWithoutCommands: [String: ProductFilter] = [
```
SDGGiesbrecht, line 513: The differing context means the `includeCommands` need to be opposite what they were above, so they can no longer share the same variable.

Tests/WorkspaceTests/WorkspaceTests.swift
```diff
@@ -1831,8 +1831,12 @@ final class WorkspaceTests: XCTestCase {
         }
 
         try workspace.loadDependencyManifests(roots: ["Root1"]) { manifests, diagnostics in
-            // Ensure that the order of the manifests is stable.
+            // Ensure that the order of the manifests is stable from one run to the next (changes accross versions do not matter).
+            #if ENABLE_TARGET_BASED_DEPENDENCY_RESOLUTION
+            XCTAssertEqual(manifests.allDependencyManifests().map { $0.value.manifest.displayName }, ["Bam", "Baz", "Bar", "Foo"])
```
SDGGiesbrecht, line 1836: The new nodes altered the traversal order.

SDGGiesbrecht (commented):
I left some inline comments.

Tests/PackageGraphPerformanceTests/DependencyResolverPerfTests.swift
SDGGiesbrecht: These performance tests are inoperable on `main` anyway, because the JSON files are out of date.

Tests/PackageGraphPerformanceTests/DependencyResolverPerfTests.swift
```diff
@@ -174,14 +174,17 @@ extension ProductFilter: JSONSerializable, JSONMappable {
         switch self {
         case .everything:
             return "all".toJSON()
-        case .specific(let products):
+        case .specific(var products, let includeCommands):
```
SDGGiesbrecht, line 177: I patched this up to do something sane. It does not match the actual files any more than `main` does, but will be operable if someone makes a new file.

abertelrud:
Thank you very much @SDGGiesbrecht!  Including for the detailed comments.  I will start looking through it tonight, but might not wrap up until tomorrow.  I really appreciate the fixes here.

Sources/PackageGraph/DependencyResolutionNode.swift
```diff
@@ -58,23 +58,38 @@ public enum DependencyResolutionNode {
     ///   to the information needed in order to provide actionable suggestions to help the user stitch up the dependency declarations properly.
     case root(package: PackageReference)
 
+    // FIXME: If command plugins become explicit in the manifest, or are extricated from the main graph, `implicitCommands` should be removed.
```
abertelrud, line 61: 👍

Sources/PackageModel/Product.swift
```diff
@@ -131,23 +137,22 @@ public enum ProductFilter: Equatable, Hashable {
         self = self.union(other)
     }
 
-    public func contains(_ product: String) -> Bool {
+    public func contains(_ product: String, isCommandPlugin: (String) -> Bool) -> Bool {
```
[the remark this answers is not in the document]
abertelrud: That makes sense.  Because SwiftPM doesn't yet have a versioned API I suspect that any clients could adapt, but passing in the closure makes sense too.

Sources/PackageModel/Product.swift
abertelrud: I don't think this is stored anyway, though @neonichu or @tomerd would know better.

Tests/PackageGraphPerformanceTests/DependencyResolverPerfTests.swift
```diff
@@ -174,14 +174,17 @@ extension ProductFilter: JSONSerializable, JSONMappable {
...
+            if includeCommands { products.insert("_commands_") }
```
abertelrud, line 178: I know it's a nitpick, but perhaps call this "command plugins" since command is a somewhat overloaded term?

abertelrud (approved):
This looks good to me (at least to the degree I'm familiar with the package resolution).  Thanks for taking on these fixes — hopefully when target-based dependencies are enabled by default again, the unit tests will prevent any further regressions.

Tests/PackageGraphPerformanceTests/DependencyResolverPerfTests.swift
SDGGiesbrecht: Which are you referring to? The label of the enumeration’s associated value, or the JSON placeholder string?

abertelrud: Both actually, but the name of the placeholder is one that might show up in debug output (I assume?), so that's the more important one, I'd assume.

abertelrud: @swift-ci please smoke test

SDGGiesbrecht: I did find‐and‐replace from “includeCommands” to “includeCommandPlugins” and from “_commands_” to “_command_plugins_”.

abertelrud:
@swift-ci please smoke test

Tests/PackageGraphPerformanceTests/DependencyResolverPerfTests.swift
abertelrud: Thank you!  Might save someone some confusion one day...

tomerd:
thanks @SDGGiesbrecht, need to review this in more details given the added complexity - which may be necessary, but still somewhat unfortunate

SDGGiesbrecht:
Something has changed in `main` since I submitted this, such that `testTransitiveDependencySwitchWithSameIdentity()` now crashes.

The following is able to “fix” it, but seems to indicate an underlying problem somewhere else:

```swift
// Sources/Workspace/Workspace.swift
     func testTransitiveDependencySwitchWithSameIdentity() throws {
     // ...
+        imortal = workspace._workspace?.repositoryManager.lookupSemaphore
     }
+    var imortal: Any?
```
```swift
// Sources/SPMTestSupport/MockWorkspace.swift
-    private var _workspace: Workspace?
+    public var _workspace: Workspace?
```
```swift
// Sources/Workspace/Workspace.swift
-    fileprivate var repositoryManager: RepositoryManager
+    public var repositoryManager: RepositoryManager
```
```swift
// Sources/SourceControl/RepositoryManager.swift
-    private let lookupSemaphore: DispatchSemaphore
+    public let lookupSemaphore: DispatchSemaphore
```

SDGGiesbrecht:
I moved the workaround from my last comment into the umbrella PR (#3838) in order to unblock work on its branch. The issue does not seem to be directly related to anything here.

This pull request does not require that other fix. Under legacy resolution this PR makes no difference and all remains green, and under target‐based resolution, It moves the current state from several tests broken to one test broken, which is a net improvement.

SDGGiesbrecht:
The above bug simply vanished. Some recent change to `main` must have fixed it.

abertelrud:
@swift-ci please smoke test

SDGGiesbrecht:
I was just pinged on the forums for an update on this. Are you able to clarify when “the next merge window” is?

tomerd:
thank you for your patience with this @SDGGiesbrecht, our hope to take the changes marked as "next" in the few weeks

SDGGiesbrecht:
Pinging because another few weeks have passed. @tomerd, @abertelrud, @neonichu

samdeane:
Is there a reason why this is stalled? For what it's worth, the status of this change came up in [this discussion on the forums](https://forums.swift.org/t/allow-command-plugin-dependencies-to-be-conditional/58947).

SDGGiesbrecht:
Pinging again. A total of six months have now elapsed. @tomerd, @abertelrud, @neonichu

tomerd:
thanks for your patience @SDGGiesbrecht, aware this is taking very long. we are discussing scope for the next release and this is something we are looking into

SDGGiesbrecht:
Work on SwiftPM seems to be picking up again after the 5.7 release. What came of your discussions? @tomerd

malcommac:
Is there any schedule for this MR? Thanks

SDGGiesbrecht:
Closing. This and related PRs are stale and I do not have time to maintain them anymore. I will leave the related branches around for now in case anyone else wants to pick them up.

PR #4043, opened 2022-01-22
[SR-9402] Emit a warning on duplicate target dependencies

sjavora:
~Implements https://bugs.swift.org/browse/SR-9402, which calls for SwiftPM to emit a warning when a target has duplicate dependencies.~

SPM was already partially checking for this, so this updates/expands that check. I'm not in love with `duplicateDetectionHash`, but it does what it needs to.

### DIFF
--- a/Sources/PackageLoading/PackageBuilder.swift
+++ b/Sources/PackageLoading/PackageBuilder.swift
@@ -783,9 +783,8 @@ public final class PackageBuilder {
             )
         }
 
-        // Check for duplicate target dependencies by name
-        let combinedDependencyNames = dependencies.map { $0.target?.name ?? $0.product!.name }
-        combinedDependencyNames.spm_findDuplicates().forEach {
+        // Check for duplicate target dependencies
+        dependencies.spm_findDuplicateElements(by: \.nameAndType).map(\.[0].name).forEach {
             self.observabilityScope.emit(.duplicateTargetDependency(dependency: $0, target: potentialModule.name, package: self.identity.description))
         }
 
@@ -1469,5 +1468,17 @@ extension PackageBuilder {
                                    swiftVersion: try swiftVersion(),
                                    buildSettings: buildSettings)
             }
+    }
+}
+
+private extension Target.Dependency {
+    
+    var nameAndType: String {
+        switch self {
+            case .target:
+                return "target-\(name)"
+            case .product:
+                return "product-\(name)"
+        }
     }
 }
--- a/Tests/PackageLoadingTests/PackageBuilderTests.swift
+++ b/Tests/PackageLoadingTests/PackageBuilderTests.swift
@@ -2178,7 +2178,11 @@ class PackageBuilderTests: XCTestCase {
         let fs = InMemoryFileSystem(emptyFiles:
             "/Foo/Sources/Foo/foo.swift",
             "/Foo/Sources/Foo2/foo.swift",
-            "/Bar/Sources/Bar/bar.swift"
+            "/Foo/Sources/Foo3/foo.swift",
+            "/Foo/Sources/Qux/foo.swift",
+            "/Bar/Sources/Bar/bar.swift",
+            "/Bar/Sources/Bar2/bar.swift",
+            "/Bar/Sources/Qux/bar.swift"
         )
 
         let manifest = Manifest.createRootManifest(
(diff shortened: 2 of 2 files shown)


Sources/PackageLoading/ManifestLoader.swift
```diff
@@ -1027,6 +1034,10 @@ extension Basics.Diagnostic {
     static func duplicateTargetName(targetName: String) -> Self {
         .error("duplicate target named '\(targetName)'")
     }
+    
+    static func duplicateTargetDependency(dependencyName: String) -> Self {
+        .warning("duplicate target dependency named '\(dependencyName)'")
```
SDGGiesbrecht, line 1039: Including the target name in the message would help the user find it faster.

Sources/PackageModel/Manifest/TargetDescription.swift
```diff
@@ -22,7 +22,7 @@ public struct TargetDescription: Equatable, Codable {
     }
 
     /// Represents a target's dependency on another entity.
-    public enum Dependency: Equatable {
+    public enum Dependency: Equatable, Hashable {
```
SDGGiesbrecht, line 25: Relying on the synthesized `Hashable` like this will not catch the following sort of duplicate:

```swift
"a", // i.e. .byName(name: "a")
.target(name: "a")
```

On the other hand, comparing only the name will snag valid pairs, such as this:

```swift
.target(name: "a"),
.product(name: "a", package: "c")
```

Sources/PackageModel/Manifest/PackageConditionDescription.swift
```diff
@@ -9,7 +9,7 @@
 */
 
 /// Represents a manifest condition.
-public struct PackageConditionDescription: Codable, Equatable {
+public struct PackageConditionDescription: Codable, Equatable, Hashable {
```
SDGGiesbrecht, line 12: Do you think it is worth also warning of dependencies that can be combined?

For example warning about this:

```swift
.target(name: "a", condition: .when(platforms: [.macOS])),
.target(name: "a", condition: .when(platforms: [.linux]))
```

And suggesting this:

```swift
.target(name: "a", condition: .when(platforms: [.macOS, .linux]))
```

SDGGiesbrecht (commented):
I am not the primary reviewer, but I left some comments you can think about while you wait for the real review.

Sources/PackageModel/Manifest/TargetDescription.swift
infotrinerdis: Does `byName` always represent a target? If so, it might be easiest to convert one to the other for the purposes of this comparison 🤔

sjavora: Does `byName` always represent a target? If so, it might be easiest to convert one to the other for this comparison 🤔

Sources/PackageModel/Manifest/PackageConditionDescription.swift
sjavora: That seems like a stylistic choice to me. At least in Swift itself, this is not something that is enforced, e.g.:
```swift
switch foo {
    case .a:
        print("not c")
    case .b:
        print("not c")
    case .c:
        print("c!")
}
```
vs 
```swift
switch foo {
    case .a, .b:
        print("not c")
    case .c:
        print("c!")
}
```
There is no warning to combine the first two cases.

But maybe the package manager has different conventions?

Sources/PackageModel/Manifest/TargetDescription.swift
SDGGiesbrecht: No. It is short for either `.target(name: "a")` or `.product(name: "a", package: "a"`), or in pre‐5.2 tools-versions, `.product(name: "a", package: /* all */)`. Figuring it out either requires more elaborate logic, or performing the check later in the loading process after these sorts of connections have already been resolved.

Sources/PackageModel/Manifest/PackageConditionDescription.swift
SDGGiesbrecht: I have no strong opinion either way. It just came up while I was thinking through the implications of comparing by name, and I wondered what your thoughts were.

Sources/PackageModel/Manifest/TargetDescription.swift
sjavora: > either requires more elaborate logic

Would this not just duplicate the resolution process that takes place later?

> or performing the check later in the loading process

This makes more sense to me.

What is your opinion on this?

sjavora:
Well, it turns out SPM already checks for duplicate target dependencies [here](https://github.com/apple/swift-package-manager/blob/5c4bef08836f8f30437932470e281b53f69cb709/Sources/PackageLoading/PackageBuilder.swift#L785)! But it checks only by name so the valid so this point actually applies to the current implementation:

> On the other hand, comparing only the name will snag valid pairs, such as this:
> 
> ```swift
> .target(name: "a"),
> .product(name: "a", package: "c")
> ```

I updated the implementation to handle that case and also added some explicit `.target` and `.product` to the test, since it also only had dependencies "by name".

Sources/PackageLoading/PackageBuilder.swift
```diff
@@ -1470,3 +1469,20 @@ extension PackageBuilder {
             }
     }
 }
+
+private extension Target.Dependency {
+    
+    var duplicateDetectionHash: Int {
```
tomerd, line 1475: thanks @sjavora, how about returning a string instead of a hash? e.g. (may not compile):

```
var nameAndType: String {
switch self {
             case .target:
                return "target-\(self.name)"
             case .product:
                return "product-\(self.name)"
         }
}
```

sjavora: Thanks @tomerd, compiles just fine - updated 👍.

tomerd (approved):

tomerd:
@swift-ci please smoke test

tomerd:
thanks @sjavora

PR #4044, opened 2022-01-23
Fix the `Target.recursiveTargetDependencies` plugin API function

abertelrud:
The `Target.recursiveTargetDependencies` plugin API function didn't work correctly, returning an empty array.  This fixes it and adds an end-to-end unit test to check its output for various combinations of target dependencies.

### Motivation:

This is part of the plugins feature and should return correct results.

### Modifications:

- fix the incomplete implementation of the `Target.recursiveTargetDependencies` function in the PackagePlugin module
- add an end-to-end unit test from command line to printed output

rdar://87623905

### DIFF
--- a/Sources/PackagePlugin/PackageModel.swift
+++ b/Sources/PackagePlugin/PackageModel.swift
@@ -420,21 +420,19 @@ extension Target {
     /// depends on it (i.e. in "topological sort order").
     public var recursiveTargetDependencies: [Target] {
         // FIXME: We can rewrite this to use a stack instead of recursion.
-        var result = [Target]()
         var visited = Set<Target.ID>()
-        func visit(target: Target) {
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
Hmm:
```
docker build -t ${JOB_NAME} swift-docker/swift-ci/master/ubuntu/20.04
15:08:41 + docker build -t swift-package-manager-self-hosted-Linux-smoke-test swift-docker/swift-ci/master/ubuntu/20.04
15:08:43 invalid argument "swift-package-manager-self-hosted-Linux-smoke-test" for "-t, --tag" flag: invalid reference format: repository name must be lowercase
```

tomerd (approved):

PR #4047, opened 2022-01-24
[5.6] Make the `@main` entry points for plugins be `async` so it can call other `async` APIs

abertelrud:
This also adds a `-rpath` flag when compiling the plugin to run on macOS earlier than 12.0, so that the back deployment libraries in the compiler's toolchain can be found.

Cherry-pick of #3842 to 5.6.

rdar://83206272

### DIFF
--- a/Sources/SPMTestSupport/Toolchain.swift
+++ b/Sources/SPMTestSupport/Toolchain.swift
@@ -58,3 +58,38 @@ extension UserToolchain {
         }
     }
 }
+
+extension UserToolchain {
+    /// Helper function to determine if async await actually works in the current environment.
+    public func supportsSwiftConcurrency() -> Bool {
+      #if os(macOS)
+        if #available(macOS 12.0, *) {
+            // On macOS 12 and later, concurrency is assumed to work.
+            return true
+        }
+        else {
+            // On macOS 11 and earlier, we don't know if concurrency actually works because not all SDKs and toolchains have the right bits.  We could examine the SDK and the various libraries, but the most accurate test is to just try to compile and run a snippet of code that requires async/await support.  It doesn't have to actually do anything, it's enough that all the libraries can be found (but because the library reference is weak we do need the linkage reference to `_swift_task_create` and the like).
+            do {
+                try testWithTemporaryDirectory { tmpPath in
+                    let inputPath = tmpPath.appending(component: "foo.swift")
+                    try localFileSystem.writeFileContents(inputPath, string: "public func foo() async {}\nTask { await foo() }")
+                    let outputPath = tmpPath.appending(component: "foo")
+                    let toolchainPath = self.swiftCompilerPath.parentDirectory.parentDirectory
+                    let backDeploymentLibPath = toolchainPath.appending(components: "lib", "swift-5.5", "macosx")
+                    try Process.checkNonZeroExit(arguments: ["/usr/bin/xcrun", "--toolchain", toolchainPath.pathString, "swiftc", inputPath.pathString, "-Xlinker", "-rpath", "-Xlinker", backDeploymentLibPath.pathString, "-o", outputPath.pathString])
+                    try Process.checkNonZeroExit(arguments: [outputPath.pathString])
+                }
+            } catch {
+                // On any failure we assume false.
+                return false
+            }
+            // If we get this far we could compile and run a trivial executable that uses libConcurrency, so we can say that this toolchain supports concurrency on this host.
+            return true
+        }
+      #else
+        // On other platforms, concurrency is assumed to work since with new enough versions of the toolchain.
+        return true
+      #endif
+    }
+    
+}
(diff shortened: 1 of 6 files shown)


abertelrud:
@swift-ci please test

tomerd (approved):

abertelrud:
This is still a draft — it looks like the main PR is failing because the macOS SDK on the builder is old enough that it's still emitting `/usr/lib/swift/libswift_Concurrency.dylib` into the linked binary, so the extra rpath is ignored.  I'm trying to see what happens with a full build, but it could be that the macOS SDK on the builder needs to be newer (at least as new as the one in Xcode 13.2.1).  If this fails too I'll ask around on the Slack channels.  There doesn't seem to be anything wrong with the actual diffs here, and that they are putting the rpath into the right place.

abertelrud:
@swift-ci please test

abertelrud:
Looks like the same thing is happening here.  It seems that the macOS builders have a beta of Xcode:
```
xcode-select -p
+ xcode-select -p
/Applications/Xcode-beta.app/Contents/Developer
```
and I think the SDK in it has a broken rpath for the concurrency library.  Xcode 13.2.1 final release should work.

I can probably modify the tests to detect a broken SDK and skip the tests that require async in that case.

tomerd:
> I can probably modify the tests to detect a broken SDK and skip the tests that require async in that case.

short of having @shahmishal able to update the CI builders, this could be a good compromise (ie make the tests conditional)

abertelrud:
@swift-ci please test

Sources/SPMTestSupport/Toolchain.swift
```diff
@@ -58,3 +58,38 @@ extension UserToolchain {
...
+            do {
+                try testWithTemporaryDirectory { tmpPath in
+                    let inputPath = tmpPath.appending(component: "foo.swift")
+                    try localFileSystem.writeFileContents(inputPath, string: "public func foo() async {}\nTask { await foo() }")
+                    let outputPath = tmpPath.appending(component: "foo")
+                    let toolchainPath = self.swiftCompilerPath.parentDirectory.parentDirectory
+                    let backDeploymentLibPath = toolchainPath.appending(components: "lib", "swift-5.5", "macosx")
+                    try Process.checkNonZeroExit(arguments: ["/usr/bin/xcrun", "--toolchain", toolchainPath.pathString, "swiftc", inputPath.pathString, "-Xlinker", "-rpath", "-Xlinker", backDeploymentLibPath.pathString, "-o", outputPath.pathString])
+                    try Process.checkNonZeroExit(arguments: [outputPath.pathString])
```
abertelrud, line 80: This was the most accurate test I could come up with that checks whether concurrency actually works, given the various combinations of toolchain and SDK that had various bugs.  This adds about a second per skipped test to the running time, and only on macOS 11 — I couldn't find a good way to memoize this result based on the toolchain instance, since extensions can't add properties and different toolchains could have different results — I guess we could have a mapping from toolchain to result, but I'm not sure it's worth it.

abertelrud: But the check itself seems robust.

Sources/SPMTestSupport/Toolchain.swift
```diff
@@ -58,3 +58,38 @@ extension UserToolchain {
...
+        }
+        else {
+            // On macOS 11 and earlier, we don't know if concurrency actually works because not all SDKs and toolchains have the right bits.  We could examine the SDK and the various libraries, but the most accurate test is to just try to compile and run a snippet of code that requires async/await support.  It doesn't have to actually do anything, it's enough that all the libraries can be found (but because the library reference is weak we do need the linkage reference to `_swift_task_create` and the like).
+            do {
+                try testWithTemporaryDirectory { tmpPath in
+                    let inputPath = tmpPath.appending(component: "foo.swift")
+                    try localFileSystem.writeFileContents(inputPath, string: "public func foo() async {}\nTask { await foo() }")
+                    let outputPath = tmpPath.appending(component: "foo")
+                    let toolchainPath = self.swiftCompilerPath.parentDirectory.parentDirectory
+                    let backDeploymentLibPath = toolchainPath.appending(components: "lib", "swift-5.5", "macosx")
```
abertelrud, line 78: Note that this subdirectory is always 5.5, no matter what the compiler version.

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -161,6 +161,18 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
...
             }
 
+            #if os(macOS)
+            // On macOS earlier than 12, add an rpath to the directory that contains the concurrency fallback library.
+            if #available(macOS 12.0, *) {
+                // Nothing is needed; the system has everything we need.
+            }
+            else {
+                // Add an `-rpath` so the Swift 5.5 fallback libraries can be found.
+                let swiftSupportLibPath = self.toolchain.swiftCompilerPath.parentDirectory.parentDirectory.appending(components: "lib", "swift-5.5", "macosx")
```
abertelrud, line 171: Note that this directory is always 5.5, no matter the compiler version.

tomerd (approved):

abertelrud:
Will backport these test skips to main once these tests are green, so that the diff is the same.

abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test

abertelrud:
Rebased on top-of-branch after merging other PRs — local Git can see that the differences are the same and there is no actual conflict, but GitHub didn't realize that, so repushed and restarting tests.

PR #4049, opened 2022-01-25
do not use Dictionary::uniqueKeysWithValues when input is not sanitized

tomerd:
motivation: Dictionary::uniqueKeysWithValues is asserting and will crash the process on bad data. need a throwing variant

changes:
* create Dictionary::throwingUniqueKeysWithValues which is the same as uniqueKeysWithValues but throwing insteada of asserting on duplicate key
* update call sites that take user data (which is not sanitized) to use the throwing variant
* update relevant call sites to new throwing signatures
* add test

### DIFF
--- a/Sources/PackageCollections/Providers/JSONPackageCollectionProvider.swift
+++ b/Sources/PackageCollections/Providers/JSONPackageCollectionProvider.swift
@@ -191,97 +191,101 @@ struct JSONPackageCollectionProvider: PackageCollectionProvider {
     }
 
     private func makeCollection(from collection: JSONModel.Collection, source: Model.CollectionSource, signature: Model.SignatureData?) -> Result<Model.Collection, Error> {
-        if let errors = self.validator.validate(collection: collection)?.errors() {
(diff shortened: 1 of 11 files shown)


tomerd:
@swift-ci please smoke test

Sources/PackageCollections/Providers/JSONPackageCollectionProvider.swift
```diff
@@ -191,97 +191,101 @@ struct JSONPackageCollectionProvider: PackageCollectionProvider {
...
-
-        return .success(.init(source: source,
-                              name: collection.name,
-                              overview: collection.overview,
-                              keywords: collection.keywords,
-                              packages: packages,
-                              createdAt: collection.generatedAt,
-                              createdBy: collection.generatedBy.flatMap { Model.Collection.Author(name: $0.name) },
-                              signature: signature,
-                              lastProcessedAt: Date()))
```
tomerd, line 284: diff is terrible here, but its mostly formatting as the code moved into a `try` block

neonichu (approved):

abertelrud (approved):

PR #4051, opened 2022-01-25
Build: correct emitted output file map

compnerd:
This adjusts the output file map writing code to ensure that we emit the
proper native path string and correct escaping of `\` which otherwise
yields invalid JSON which cannot be deserialized subsequently.  This
will be critical to enabling the use of swift-driver with SPM.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -25,6 +25,18 @@ extension String {
     }
 }
 
+extension AbsolutePath {
+    internal func nativePathString(escaped: Bool) -> String {
+        return URL(fileURLWithPath: self.pathString).withUnsafeFileSystemRepresentation {
+            let repr = String(cString: $0!)
+            if escaped {
+                return repr.replacingOccurrences(of: "\\", with: "\\\\")
+            }
+            return repr
+        }
+    }
+}
+
 extension BuildParameters {
     /// Returns the directory to be used for module cache.
     public var moduleCache: AbsolutePath {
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

Sources/Build/BuildPlan.swift
```diff
@@ -25,6 +25,18 @@ extension String {
     }
 }
 
+extension String {
```
tomerd, line 28: fells like this should be extensions on AbsolutePath rather than on String

compnerd: As a local extension to `AbsolutePath` is what you have in mind?

tomerd: right, this basically operates on the string version of absolute path, so the call sites are `<path>.pathString.nativePathString()` and would be better as  `<path>.nativePathString()`

tomerd:
```
extension AbsolutePath {
  internal func nativePathString() -> String {
       return URL(fileURLWithPath: self.pathString).withUnsafeFileSystemRepresentation {
           String(cString: $0!)
       }
   }
}
```

compnerd: Agreed, just wanted to make sure that the location was still within this file :)

tomerd: also consider computed var vs. func to make it more like `pathString`, tho I can go both ways given the allocations involved

tomerd: one more suggestion: make it into `nativePathString(escaped: Bool)` so this way we can collapse these two into one clean API

tomerd (approved):

neonichu (approved):

PR #4053, opened 2022-01-26
Add a simple argument extractor to PackagePlugin

abertelrud:
This helps with simple cases until it is possible to use SwiftArgumentParser or other package dependencies from package plugins

### Motivation:

The initial version of plugins can't yet link against other modules, which excludes use of SwiftArgumentParser and similar libraries.  Plugins with more advanced needs will probably include an implementation of one of the other, simpler argument parser libraries available, but after discussion we've agreed that something minimal that operates on just string lists may make sense.  The future is clearly to use the proper argument parser, but this is a useful helper until we have that.

### Modifications:

- add an ArgumentExtractor type to the PackagePlugin module
- added a new dedicated unit test for this PackagePlugin API
- extended an end-to-end plugin unit test

### Result:

Simple extraction of argument options and flags from a plugin are possible without additional code.

### DIFF
--- a/Sources/PackagePlugin/ArgumentExtractor.swift
+++ b/Sources/PackagePlugin/ArgumentExtractor.swift
@@ -1,0 +1,75 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2022 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+ */
+
+/// A rudimentary helper for extracting options and flags from a string list representing command line arguments. The idea is to extract all known options and flags, leaving just the positional arguments. This does not handle well the case in which positional arguments (or option argument values) happen to have the same name as an option or a flag. It only handles the long `--<name>` form of options, but it does respect `--` as an indication that all remaining arguments are positional.
+public struct ArgumentExtractor {
+    private var args: [String]
+    private let literals: [String]
+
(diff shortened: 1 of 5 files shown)


abertelrud:
@swift-ci please smoke test

tomerd (approved):
looks good. 

missing a dedicated unit test? could come as follow up PR

abertelrud:
> looks good.
> 
> missing a dedicated unit test? could come as follow up PR

I changed one of the overall command plugin unit tests to test it, so we do have some coverage, but I see what you mean — it would be better to have one that we could debug through, etc, and that would add more test cases.  I'll add one in an upcoming PR.

abertelrud:
> > looks good.
> > missing a dedicated unit test? could come as follow up PR
> 
> I changed one of the overall command plugin unit tests to test it, so we do have some coverage, but I see what you mean — it would be better to have one that we could debug through, etc, and that would add more test cases. I'll add one in an upcoming PR.

Changed my mind on this and am adding one.  I'm waiting for the async one to land anyway, and then will have to rebase this one (probably — not sure yet) so I might as well.

abertelrud:
@swift-ci please smoke test

tomerd (approved):

Sources/PackagePlugin/ArgumentExtractor.swift
```diff
@@ -0,0 +1,75 @@
...
+            }
+            else {
+                idx += 1
+            }
+        }
+        return count
+    }
+
+    /// Returns any unextracted flags or options (based strictly on whether remaining arguments have a "--" prefix).
+    public var unextractedOptionsOrFlags: [String] {
```
abertelrud, line 67: Should this be called `remainingOptionsOrFlags`?

abertelrud:
@swift-ci please smoke test

PR #4054, opened 2022-01-26
[test] Disable PluginTests.testContrivedTestCases on Linux

Azoy:
This is currently failing on some linux bots: https://ci.swift.org/job/oss-swift-5.6-package-centos-8-aarch64/10

rdar://88080030

### DIFF
--- a/Tests/FunctionalTests/PluginTests.swift
+++ b/Tests/FunctionalTests/PluginTests.swift
@@ -129,6 +129,11 @@ class PluginTests: XCTestCase {
         try XCTSkipIf(true, "skipping because host compiler doesn't support '-entry-point-function-name'")
         #endif
         
+        // rdar://88080030
+        #if os(Linux)
+        try XCTSkipIf(true, "Disabled on linux for now as it's failing on some CI")
+        #endif
+     
         fixture(name: "Miscellaneous/Plugins") { path in
             do {
                 let (stdout, _) = try executeSwiftBuild(path.appending(component: "ContrivedTestPlugin"), configuration: .Debug, extraArgs: ["--product", "MyLocalTool"])


abertelrud:
Is there a way to disable this just for CentOS aarch64?  AFAIK this has been running fine on other Linuxes.  Or are you seeing other cases of failure now than CentOS aarch64?

abertelrud:
Perhaps conditionalize also on the arm64 architecture?

tomerd:
a little confused on why we are disabling this, unless temporary measure?

there is nothing in this test that should be architecture sensitive afaik. is it constantly failing on some architectures?

xedin:
@tomerd It might be just a blip test but it failed at last ones https://ci.swift.org/job/oss-swift-5.6-package-centos-8-aarch64/10

tomerd:
@xedin @abertelrud I'd be fine disabling it temporarily if this is a recurring issue, but I suspect it points to a el problem and not an architecture specific one

xedin:
The package is currently building, let's see if that would fail with the same issue or not...

abertelrud:
If it's a real problem I suspect it would be limited to CentOS aarch64.  I don't think we should lose all Linux coverage if it's a platform specific issue.

abertelrud:
> @xedin @abertelrud I'd be fine disabling it temporarily if this is a recurring issue, but I suspect it may point to a real problem and not an architecture specific one

This has been consistently passing on macOS and regular Linux

abertelrud:
Looks like https://ci.swift.org/job/oss-swift-5.6-package-centos-8-aarch64/12 succeeded?

xedin:
Looks like it might have been just a blip, all the subsequent builds are green, closing this for now.

PR #4055, opened 2022-01-27
Simplify the command plugin method to get all the inputs in `arguments` instead of a specific `targets` parameter

abertelrud:
As part of this we simplify the command plugin method to not have a `targets` parameter but to instead get all the inputs in `arguments`.

Any `--target` or `--product` flags from the `swift package` invocation are now prepended as `--target` and `--product` parameters in the plugin arguments, respectively, and the ArgumentExtractor helper type makes it easier to get at this information from the plugin.  Also add methods to make it easier to look up products and targets in the plugin context's package.

### Motivation:

The `targets` parameter was fairly specific and didn't account for products and other entities that the plugin might instead operate on.  Providing API inside the plugin to help look this up based on arguments provides more flexibility for the plugin.

### Modifications:

- pass on any `--target` and `--product` flags to the `swift package` command corresponding parameters in `arguments`
- remove the `targets` parameter from the `performCommand()` method
- add PackagePlugin helpers to look up targets and products
- adjust and extend unit tests

rdar://86786279

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -878,9 +878,6 @@ extension SwiftPackageTool {
         }
     }
     
-    // Experimental command to invoke user command plugins. This will probably change so that command that is not
-    // recognized as a built-in command will cause `swift-package` to search for plugin commands, instead of using
-    // a separate `plugin` subcommand for this.
     struct PluginCommand: SwiftCommand {
         static let configuration = CommandConfiguration(
             commandName: "plugin",
@@ -891,8 +888,12 @@ extension SwiftPackageTool {
         var swiftOptions: SwiftToolOptions
 
         @Option(name: .customLong("target"),
-                help: "Target(s) to which the plugin command should be applied")
+                help: "Target(s) to pass as `--target` arguments to the plugin")
         var targetNames: [String] = []
+
+        @Option(name: .customLong("product"),
+                help: "Product(s) to pass as `--product` arguments to the plugin")
+        var productNames: [String] = []
 
         @Flag(name: .customLong("list"),
               help: "List the available plugin commands")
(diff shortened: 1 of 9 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
This needs some work on the targets that are passed to the command — putting back draft for now, but plan to move this forward today.

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -1036,6 +1036,9 @@ extension SwiftPackageTool {
                     }
                 }
             }
+            
+            // Pass any targets that were specified, followed by any plugin arguments that were given on the command line.
+            let pluginArguments = targetNames.flatMap{ ["--target", $0] } + pluginCommand.dropFirst()
```
tomerd, line 1041: do products need special treatment also?

Sources/PackagePlugin/Protocols.swift
```diff
@@ -62,7 +74,35 @@ public protocol CommandPlugin: Plugin {
...
+        let targets: [Target] = try targetNames.map{ name in
+            guard let target = context.package.targets.first(where: { $0.name == name }) else {
+                throw PluginContextError.targetNotFound(packageName: context.package.displayName, targetName: name)
+            }
+            return target
+        }
+        try await self.performCommand(context: context, targets: targets, arguments: targetArgExtractor.remainingArguments)
+    }
+
+    public func performCommand(
```
tomerd, line 98: do we need to keep this around? we have not release this api yet, so not sure we need to worry about backwards compatibility

tomerd (commented):
looks good, couple of small questions

Sources/PackagePlugin/Protocols.swift
```diff
@@ -62,7 +74,35 @@ public protocol CommandPlugin: Plugin {
 }
 
 extension CommandPlugin {
-    
+    /// Compatibility implementation that calls the older version of this method.
+    public func performCommand(
```
tomerd, line 78: if we do this default implementation, then the protocol requirement will do nothing for plugins, ie they will not know which api they need to implement. i think we should just delete the one with the targets and not have a default

Sources/PackagePlugin/Protocols.swift
abertelrud: I wanted to smooth the transition for DocC, but I think the drawback of having both is a good point.

Sources/PackagePlugin/Protocols.swift
abertelrud: Yes, this is one of the problems.  I agree that we should remove it.

Sources/Commands/SwiftPackageTool.swift
abertelrud: Yes, I'll be adding that, and also removing the fallback of passing all targets.  I don't think there are other flags that it makes sense to propagate — and that is something we can add to over time based on toolsversion without breaking anyone.  I think that's the last thing to fix before putting this out of draft.

abertelrud:
@swift-ci please smoke test

tomerd (approved):
looks great, thanks for this

Sources/PackagePlugin/PackageModel.swift
```diff
@@ -414,6 +414,30 @@ public enum FileType {
     case unknown
 }
 
+extension Package {
+    /// The list of targets matching the given names. Throws an error if any of
+    /// the targets cannot be found.
+    public func targets(named targetNames: [String]) throws -> [Target] {
+        return try targetNames.map { name in
+            guard let target = self.targets.first(where: { $0.name == name }) else {
+                throw PluginContextError.targetNotFound(name: name, package: self)
```
tomerd, line 423: consider returning none instead of throwing an error, tho probably a style thing

abertelrud: Yeah, this seemed to be the common case.  There is possibly also room for a separate method that returns an optional, but it's less of a win for the common case,  and seems like something that can be added later.

abertelrud: We can also improve the error here to include all the unknown names at once, but that wouldn't affect the API.  Totally open to change the API here though.

tomerd: I think its fine

PR #4059, opened 2022-01-28
cache registry downloads

tomerd:
motivation: better performance when using registry

cached:
* introduce RegistryDownloadsManager which manages the downloads across the local and cache
* integrate RegistryDownloadsManager into Workspace, which now uses it to download from registry instead of the registry client directly
* update workspace delegate to handle callback from registry downloads, and propogate them to workspace delegates
* improve / generalize the workspace fetching delegate methods such that they are appropriate to both repositories and registry downloads
* deprecate --enable-repository-cache/--disable-repository-cache in favor of --enable-dependency-cache/--disable-dependency-cache
* fix several download and unzip edge cases which came to bear now that a cache is in place
* add and adjust tests

### DIFF
--- a/Sources/Commands/Options.swift
+++ b/Sources/Commands/Options.swift
@@ -152,9 +152,20 @@ public struct SwiftToolOptions: ParsableArguments {
     @Option(help: "Specify the shared security directory")
     var securityPath: AbsolutePath?
 
-    /// Disables repository caching.
-    @Flag(name: .customLong("repository-cache"), inversion: .prefixedEnableDisable, help: "Use a shared cache when fetching repositories")
-    var useRepositoriesCache: Bool = true
+    /// Disables package caching.
+    @Flag(name: .customLong("dependencies-cache"), inversion: .prefixedEnableDisable, help: "Use a shared cache when fetching dependencies")
+    var _useDependenciesCache: Bool?
+
+    // TODO: simplify when deprecating the older flag
+    var useDependenciesCache: Bool {
+        if let value = self._useDependenciesCache {
+            return value
+        } else if let value = self._deprecated_useRepositoriesCache {
+            return value
+        } else {
+            return true
+        }
+    }
 
     /// The custom working directory that the tool should operate in (deprecated).
     @Option(name: [.long, .customShort("C")])
@@ -367,6 +378,10 @@ public struct SwiftToolOptions: ParsableArguments {
 
     @Flag(name: .customLong("netrc-optional"), help: .hidden)
     var _deprecated_netrcOptional: Bool = false
+
+    /// Disables repository caching.
+    @Flag(name: .customLong("repository-cache"), inversion: .prefixedEnableDisable, help: .hidden)
+    var _deprecated_useRepositoriesCache: Bool?
 
     public init() {}
 }
--- a/Sources/Commands/Options.swift
+++ b/Sources/Commands/Options.swift
@@ -153,7 +153,7 @@ public struct SwiftToolOptions: ParsableArguments {
     var securityPath: AbsolutePath?
 
     /// Disables package caching.
-    @Flag(name: .customLong("dependencies-cache"), inversion: .prefixedEnableDisable, help: "Use a shared cache when fetching dependencies")
+    @Flag(name: .customLong("dependency-cache"), inversion: .prefixedEnableDisable, help: "Use a shared cache when fetching dependencies")
     var _useDependenciesCache: Bool?
 
     // TODO: simplify when deprecating the older flag
--- a/Sources/Commands/SwiftTool.swift
+++ b/Sources/Commands/SwiftTool.swift
@@ -490,7 +490,7 @@ public class SwiftTool {
         }
 
         if options._deprecated_useRepositoriesCache != nil {
-            observabilityScope.emit(warning: "'--disable-repository-cache'/'--enable-repository-cache' flags are deprecated; use '--disable-dependencies-cache'/'--enable-dependencies-cache' instead")
+            observabilityScope.emit(warning: "'--disable-repository-cache'/'--enable-repository-cache' flags are deprecated; use '--disable-dependency-cache'/'--enable-dependency-cache' instead")
         }
 
     }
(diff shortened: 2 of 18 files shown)


Sources/Commands/Options.swift
```diff
@@ -152,9 +152,20 @@ public struct SwiftToolOptions: ParsableArguments {
     @Option(help: "Specify the shared security directory")
     var securityPath: AbsolutePath?
 
-    /// Disables repository caching.
-    @Flag(name: .customLong("repository-cache"), inversion: .prefixedEnableDisable, help: "Use a shared cache when fetching repositories")
-    var useRepositoriesCache: Bool = true
+    /// Disables package caching.
+    @Flag(name: .customLong("dependencies-cache"), inversion: .prefixedEnableDisable, help: "Use a shared cache when fetching dependencies")
```
tomerd, line 156: not sure if `--disable-dependencies-cache` or `--disable-dependency-cache` is better

tomerd:
@swift-ci please smoke test

Sources/Commands/Options.swift
abertelrud: Either is correct for an adjectival noun, but it looks as if most of the others use singular (e.g. the one you're replacing as well as things like `--enable-build-manifest-caching`).  So I'd say to go with `--disable-dependency-cache`.

Sources/Commands/SwiftTool.swift
```diff
@@ -74,39 +74,55 @@ private class ToolWorkspaceDelegate: WorkspaceDelegate {
...
+                total: total ?? progress
+            )
+
+            let progress = self.fetchProgress.values.reduce(0) { $0 + $1.progress }
+            let total = self.fetchProgress.values.reduce(0) { $0 + $1.total }
+
+            self.fetchAnimation.update(
+                step: progress > Int.max ? Int.max : Int(progress),
+                total: total > Int.max ? Int.max : Int(total),
+                text: "Fetching \(package)"
```
abertelrud, line 121: For packages that do come from repositories, will there be a way to still see that URL?  Based on how this is called it looks as if this is always the string form of the package identity.

tomerd: this PR does not change the behavior for repository based dependencies, ie they still pass the url

Sources/Commands/SwiftTool.swift
```diff
@@ -74,39 +74,55 @@ private class ToolWorkspaceDelegate: WorkspaceDelegate {
...
 
-    func fetchingWillBegin(repository: String, fetchDetails: RepositoryManager.FetchDetails?) {
+    func willFetchPackage(package: String, fetchDetails: PackageFetchDetails) {
         queue.async {
-            self.outputStream <<< "Fetching \(repository)"
-            if let fetchDetails = fetchDetails {
-                if fetchDetails.fromCache {
-                    self.outputStream <<< " from cache"
-                }
+            self.outputStream <<< "Fetching \(package)"
```
abertelrud, line 79: For packages that do come from repositories, will there be a way to still see that URL? Based on how this is called it looks as if this is always the string form of the package identity.

abertelrud: Sorry, moved my comment from one place in the code right as you were replying.  I see, thanks for clarifying.

Sources/Commands/SwiftTool.swift
abertelrud: Sorry, moved the question that this is a reply of just as you were replying.  I see, thanks.

Sources/PackageRegistry/RegistryDownloadsManager.swift
```diff
@@ -0,0 +1,310 @@
...
+        version: Version,
+        observabilityScope: ObservabilityScope,
+        delegateQueue: DispatchQueue,
+        callbackQueue: DispatchQueue,
+        completion: @escaping  (Result<AbsolutePath, Error>) -> Void
+    ) {
+        // wrap the callback in the requested queue
+        let completion = { result in callbackQueue.async { completion(result) } }
+        
+        let packageRelativePath: RelativePath!
```
abertelrud, line 58: It's unfortunate that Swift can't reason about this being set in all code paths that don't return.

Sources/PackageRegistry/RegistryDownloadsManager.swift
```diff
@@ -0,0 +1,310 @@
...
+        packagePath: AbsolutePath,
+        observabilityScope: ObservabilityScope,
+        delegateQueue: DispatchQueue,
+        callbackQueue: DispatchQueue,
+        completion: @escaping  (Result<FetchDetails, Error>) -> Void
+    ) {
+        if let cachePath = self.cachePath {
+            do {
+                let relativePath = try package.downloadPath(version: version)
+                let cachedPacakgePath = cachePath.appending(relativePath)
```
abertelrud, line 145:
```suggestion
                let cachedPackagePath = cachePath.appending(relativePath)
```

Sources/PackageRegistry/RegistryDownloadsManager.swift
```diff
@@ -0,0 +1,310 @@
...
+
+                try self.initializeCacheIfNeeded(cachePath: cachePath)
+                try self.fileSystem.withLock(on: cachedPacakgePath, type: .exclusive) {
+                    // download the package into the cache unless already exists
+                    if try self.fileSystem.validPackageDirectory(cachedPacakgePath) {
+                        // extra validation to defend from racy edge cases?
+                        if self.fileSystem.exists(packagePath) {
```
abertelrud, line 152: What happens if this was partially downloaded?  We had a case recently where the remedy was to remove the preexisting (possibly partial) directory.  Does that apply here?

abertelrud: Oh never mind, I misread the code.  Marking resolved.

abertelrud (approved):

tomerd:
@swift-ci please smoke test

neonichu (approved):

PR #4060, opened 2022-01-28
Fix issues caused by having the `plugin` subcommand be the default subcommand

abertelrud:
There were a few undesirable issues caused by a combination of a `swift-argument-parser` issue and making the `plugin` subcommand be the default subcommand so that the `plugin` subcommand could be omitted:
 - when you invoke `swift package` without any arguments it starts resolving the package graph
 - when you make a typo like `swift package resove` it emits an error that it did not find any plugin (`error: No command plugins found for ‘resove’`)
 - the `--list` option of the `plugin` subcommand incorrectly applied even when `plugin` was omitted
 - `swift package plugin --help` yields no help
 - `swift package plugin` yields `error: No plugin command name specified`

This change introduces a new separate default subcommand that has behavior suitable for omitting the `plugin` subcommand:
 - if there are no parameters, it prints the help
 - if there is an unknown parameter, it prints an error message that takes into account that it might not be a command
 - it doesn't define options like `--list` that only make sense in the context of a `plugin` subcommand
 - it removes the definitions of `--target` and `--product` from `swift package plugin` since they are no longer treated specially, and this prevents them from being folded into `swift package` arguments even when they are on the right side of the plugin verb
 - add a unit test to check that invoking `swift package` without arguments prints help

This still leaves the problem of having options defined by `swift package`, such as `--verbose`, getting "stolen" even if they are on the right hand side of the plugin name.  Fixing that depends on a fix in `swift-argument-parser` that is planned to be included in an upcoming bugfix tag, but this is still incremental progress.

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2014 - 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See http://swift.org/LICENSE.txt for license information
@@ -61,9 +61,11 @@ public struct SwiftPackageTool: ParsableCommand {
             ArchiveSource.self,
             CompletionTool.self,
             PluginCommand.self,
+            
+            DefaultCommand.self,
         ]
         + (ProcessInfo.processInfo.environment["SWIFTPM_ENABLE_SNIPPETS"] == "1" ? [Learn.self] : []),
-        defaultSubcommand: PluginCommand.self,
+        defaultSubcommand: DefaultCommand.self,
         helpNames: [.short, .long, .customLong("help", withSingleDash: true)])
 
     @OptionGroup()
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -572,7 +572,7 @@ if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
         // The 'swift-argument-parser' version declared here must match that
         // used by 'swift-driver' and 'sourcekit-lsp'. Please coordinate
         // dependency version changes here with those projects.
-        .package(url: "https://github.com/apple/swift-argument-parser.git", .upToNextMinor(from: "1.0.1")),
+        .package(url: "https://github.com/apple/swift-argument-parser.git", branch: "nate/propagate_unconditional_remaining"),
```
abertelrud, line 575: This is obviously temporary.  Now I'm going to see what the failures are without this fix, and then see whether this PR should land without it.

abertelrud:
@swift-ci please smoke test

abertelrud:
I went ahead and reverted back to the official SwiftArgumentParser dependency because the changes have incremental benefit for everything except the "arguments are being stolen from the right hand side" problem.  Switching to a newer patch tag of SwiftArgumentParser would fix that in isolation.

abertelrud:
@swift-ci please smoke test

abertelrud:
@tomerd @ethan-kusters  This should fix several of the issues you reported.  I'm going to add a couple of more unit tests for some of the specific cases (e.g. `swift package --list` not listing plugins, etc) but the actual code changes should all be complete.

This will still required a swift-argument-parser fix for the stealing of arguments, but not including `--target` and `--product` as built-ins for plugins helps because those don't get stolen.  But things like `--verbose` do.  That's what requires the swift-argument-parser fix.

abertelrud:
Added some more unit tests.

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci smoke test

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #4058, opened 2022-01-28
Make PackagePlugin `Path.init()` public, clean up conformances

abertelrud:
Remove ExpressibleByStringLiteral conformance but allow initializer to be invoked, reducing surprises.  Add Hashable conformance.  Adjust diagnostics, and add unit test.

### Motivation:

These are further ergonomic changes for plugins based on usage experience.

### Modifications:

- mark `PackagePlugin.Path.init()` public
- remove ExpressibleByStringLiteral conformance
- add Hashable conformance
- add a unit test

PR #4061, opened 2022-01-28
Disable cache usage in package index tests

yim-lee:
https://github.com/apple/swift-package-manager/pull/4058#issuecomment-1023881874


tomerd (approved):
should we add an explicit configuration option? fine to be in a follow up PR if this is urgent, but an explicit argument can help future maintainers understand what is going on.

neonichu (approved):

yim-lee:
> should we add an explicit configuration option?

https://github.com/apple/swift-package-manager/pull/4062

PR #4072, opened 2022-02-01
Increment SwiftArgumentParser dependency to 1.0.3

abertelrud:
Increment SwiftArgumentParser dependency to 1.0.3 to pull in a bug fix that would cause subcommand arguments to be folded into the parent command

### Motivation:

We'd like to get the fix in https://github.com/apple/swift-argument-parser/pull/397 in order to fix issues with `swift` `run` and `swift` `plugin` arguments.  For example, `swift package plugin foo --verbose` would cause the `--verbose` to be applied to the parent command and not the plugin.

rdar://88097258


abertelrud:
@swift-ci please smoke test

abertelrud:
Coordinating the related toolchain repositories as well.

neonichu (approved):

abertelrud:
This is awaiting a coordinated merge by @shahmishal since the toolchain also needs to be updated to use the same version.

PR #4073, opened 2022-02-01
Remove unnecessary `TSCLibc` dependency

compnerd:
Remove the `TSCLibc` dependency as we do not depend on any of the
interfaces from the module.  This matters particularly on Windows
where static linking Swift content is not available, so this will
reduce the amount of dirty memory due to the dynamic library link
that cannot be removed.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/Commands/SwiftTool.swift
+++ b/Sources/Commands/SwiftTool.swift
@@ -19,13 +19,17 @@ import PackageModel
 import SourceControl
 import SPMBuildCore
 import TSCBasic
-import TSCLibc
+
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please smoke test

tomerd (approved):

compnerd:
@swift-ci please smoke test

neonichu (approved):

compnerd:
@swift-ci please test

abertelrud (approved):

PR #4076, opened 2022-02-01
[5.6] Increment SwiftArgumentParser dependency to 1.0.3

abertelrud:
Increment SwiftArgumentParser dependency to 1.0.3 to pull in a bug fix that would cause subcommand arguments to be folded into the parent command

Motivation:

We'd like to get the fix in apple/swift-argument-parser#397 in order to fix issues with swift run and swift plugin arguments. For example, swift package plugin foo --verbose would cause the --verbose to be applied to the parent command and not the plugin.

rdar://88097258


abertelrud:
@swift-ci test

tomerd (approved):

abertelrud:

PR #4082, opened 2022-02-03
move netrc parser into SwiftPM (from TSC)

tomerd:
motivation: the Netrc parser is only used by SwiftPM and can be moved into, which brings us closer to reducing TSC codebase

changes:
* migrate netrc parser from TSC into SwiftPM
* clean up the API a bit, e.g. use throwing instead of returning Result in non-async methods
* spelit netrc utility to parser and model vs. one monolithic utility
* migrate and clean up the tests
* update call sites

### DIFF
--- a/Sources/Basics/AuthorizationProvider.swift
+++ b/Sources/Basics/AuthorizationProvider.swift
@@ -15,7 +15,6 @@ import Security
 #endif
 
 import TSCBasic
-import TSCUtility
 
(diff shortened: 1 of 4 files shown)


tomerd:
@swift-ci smoke test

tomerd:
@swift-ci smoke test

abertelrud (approved):

PR #4083, opened 2022-02-03
move archiver to SwiftPM (from TSC)

tomerd:
motivation: the archiver utilities are only used by SwiftPM and can be moved into it, which brings us closer to reducing TSC codebase

changes:
* migrate archiver from TSC into SwiftPM
* clean up async handling by using Process async methods
* migrate and clean up the tests
* update call sites


tomerd:
@swift-ci smoke test

abertelrud (approved):

PR #4084, opened 2022-02-03
Mark plugin Command input parameters as CustomStringConvertible instead of String

abertelrud:
Mark the Command input parameters as CustomStringConvertible instead of String so that Path and others work without explicit wrapping.

### Motivation:

This allows build tool plugins to pass Path and other types that conform to CustomStringConvertible when creating the Commands.  This doesn't break any existing plugins since String trivially conforms to CustomStringConvertible.

### Modifications:

- change the `arguments` and values of the `environment` parameters in the Command type in PackagePlugin from String to CustomStringConvertible
- adjust a unit test

### Result:

Plugins can pass Path and other types that conform to CustomStringConvertible without having to manually convert to a string.


abertelrud:
@swift-ci please smoke test

tomerd (approved):
nice

PR #4087, opened 2022-02-03
[5.6] Mark plugin Command input parameters as CustomStringConvertible instead of String (#4084)

abertelrud:
This is the 5.6 nomination of #4084.

* Split out the utility functions into a separate source file from the package model

* Mark the Command input parameters as CustomStringConvertible instead of String so that Path and others work without explicit wrapping

(cherry picked from commit 2861bd05b9414791819707920bbc039007c551de)


abertelrud:
@swift-ci please test

tomerd (approved):

tomerd (approved):

PR #4091, opened 2022-02-03
Make the order in which targets (including plugin targets) are processed consistent

abertelrud:
The code that processes targets in a BuildPlan was traversing a Set without any particular order, result in different order of processing each time (due to collection enumeration randomization).

The new code that collects lists of plugin descriptions became dependent on this order, causing plugins to be compiled in different order, which would sometimes cause a unit test to fail.

Rather than just imposing the order later, it seems prudent to make sure that the initial code that traverses the targets in the graph runs in a well-defined order, if nothing else to make debuggability easier.  There is no advantage to randomizing this order, and so this PR sorts them by name.  Ideally SwiftPM would preserve the order in which targets were declared, on the theory that the user had some purpose for organizing them that way in the manifest, but since that information has been lost at this point in the code, traversing them in alphabetical order by name seems like the next best thing.  We should consider using ordered sets in the future, if nothing else to make the output more predictable to the user.

rdar://88453397

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -1616,7 +1616,7 @@ public class BuildPlan {
         // given to LLBuild.
         var targetMap = [ResolvedTarget: TargetBuildDescription]()
         var pluginDescriptions = [PluginDescription]()
-        for target in graph.allTargets {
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #4093, opened 2022-02-03
fix flaky test

tomerd:
motivation: stable ci

changes: leave sync group only after recording results

rdar://88453312


tomerd:
@swift-ci please smoke test

abertelrud (approved):

PR #4096, opened 2022-02-04
Add support for -X* flags when building with Xcode

sergiocampama:
This forwards the -Xlinker, -Xcc, -Xcxx and -Xswiftc flags to the
xcbuild invocation.

### Motivation:

This use case appeared when creating universal binaries by specifying
the arm64 and x86_64 and needing to specify the rpath flag. Since the
change is small, decided to add all of the BuildFlags as well.

### Modifications:

Changes in how the XcodeBuildSystem is configured using the BuildFlags already available in `buildParameters`.

### Result:

This command will work as expected when building universal binaries:

```swift
swift build --arch arm64 --arch x86_64 -Xlinker -rpath -Xlinker -/some/rpath
```

### DIFF
--- a/Sources/XCBuildSupport/XcodeBuildSystem.swift
+++ b/Sources/XCBuildSupport/XcodeBuildSystem.swift
@@ -171,9 +171,19 @@ public final class XcodeBuildSystem: SPMBuildCore.BuildSystem {
         // Always specify the path of the effective Swift compiler, which was determined in the same way as for the native build system.
         settings["SWIFT_EXEC"] = buildParameters.toolchain.swiftCompiler.pathString
         settings["LIBRARY_SEARCH_PATHS"] = "$(inherited) \(buildParameters.toolchain.toolchainLibDir.pathString)"
-        settings["OTHER_CFLAGS"] = "$(inherited) \(buildParameters.toolchain.extraCCFlags.joined(separator: " "))"
-        settings["OTHER_CPLUSPLUSFLAGS"] = "$(inherited) \(buildParameters.toolchain.extraCPPFlags.joined(separator: " "))"
-        settings["OTHER_SWIFT_FLAGS"] = "$(inherited) \(buildParameters.toolchain.extraSwiftCFlags.joined(separator: " "))"
+        settings["OTHER_CFLAGS"] = (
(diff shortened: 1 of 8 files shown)


jakepetroules (approved):

neonichu (approved):

Sources/XCBuildSupport/XcodeBuildSystem.swift
```diff
@@ -171,9 +171,19 @@ public final class XcodeBuildSystem: SPMBuildCore.BuildSystem {
         // Always specify the path of the effective Swift compiler, which was determined in the same way as for the native build system.
         settings["SWIFT_EXEC"] = buildParameters.toolchain.swiftCompiler.pathString
         settings["LIBRARY_SEARCH_PATHS"] = "$(inherited) \(buildParameters.toolchain.toolchainLibDir.pathString)"
-        settings["OTHER_CFLAGS"] = "$(inherited) \(buildParameters.toolchain.extraCCFlags.joined(separator: " "))"
-        settings["OTHER_CPLUSPLUSFLAGS"] = "$(inherited) \(buildParameters.toolchain.extraCPPFlags.joined(separator: " "))"
-        settings["OTHER_SWIFT_FLAGS"] = "$(inherited) \(buildParameters.toolchain.extraSwiftCFlags.joined(separator: " "))"
+        settings["OTHER_CFLAGS"] = (
+            ["$(inherited)"] + buildParameters.toolchain.extraCCFlags + buildParameters.flags.cCompilerFlags
+        ).joined(separator: " ")
```
abertelrud, line 176: I know this is existing code, but it would probably be good to do `.map{ $0.spm_shellEscaped() }.joined(separator: " ")` here.  Build setting quoting and shell quoting aren't exactly the same but this should be more conservative and avoid breaking on things like space characters in paths.

sergiocampama: It actually failed the tests to do shell escaping on the entire array, but I added it to just the inputs that are coming from the command line instead and it works as expected. thanks!

abertelrud: Hmm, that's odd — I'll make a note to take a look at that.  Either way, this looks good.  Thanks!

abertelrud (approved):

abertelrud:
@swift-ci please smoke test

PR #4100, opened 2022-02-05
[wip] refactor handling of streams into observability

tomerd:
motivation: consistent handling of output streams

changes:
* define new API on observability to scope to report raw output and progress
* stop passing outputstream around, instead use observability scope to emit output and progress events
* implement output and progress handlers in the SwiftTool such that output organized sequentially and safely
* update call sites


Sources/Basics/Observability.swift
```diff
@@ -123,11 +146,19 @@ public final class ObservabilityScope: DiagnosticsEmitterProtocol, CustomStringC
         self.diagnosticsHandler.handleDiagnostic(scope: self, diagnostic: diagnostic)
     }
 
-    private struct DiagnosticsHanderWrapper: DiagnosticsHandler {
-        private let underlying: DiagnosticsHandler
+    public func emit(output: String) {
+        self.outputHandler.handleOutput(scope: self, output: output)
+    }
+
+    public func emit(step: Int64, total: Int64, unit: String?, description: String?) {
```
abertelrud, line 153: Should the total be set when the scope is created?  I suppose that would make it harder to update it later.

Sources/Basics/Observability.swift
```diff
@@ -567,14 +598,25 @@ extension ObservabilitySystem {
...
+        var outputHandler: ScopeOutputHandler { self }
+        var progressHandler: ScopeProgressHandler { self }
 
         func handleDiagnostic(scope: ObservabilityScope, diagnostic: Diagnostic) {
             diagnosticEngine.emit(.init(diagnostic))
         }
+
+        #warning("FIXME: do something here?")
+        func handleOutput(scope: ObservabilityScope, output: String) {
+            // NOOP
```
abertelrud, line 614: It seems fine to leave this empty for the DiagnosticsEngine compatibility and instead implement output handers in the right places where we want output.

Sources/Basics/Observability.swift
```diff
@@ -663,3 +705,43 @@ extension ObservabilityMetadata {
...
+    func handleOutput(scope: ObservabilityScope, output: String)
+}
+
+public protocol OutputEmitterProtocol {
+    func emit(output: String)
+}
+
+extension OutputEmitterProtocol {
+    public func emit(output bytes: [UInt8]) {
+        if let output = String(bytes: bytes, encoding: .utf8) {
```
abertelrud, line 737: It might be better to use the decoding form that patches up invalid UTF-8, since otherwise output that contains invalid UTF-8 just disappears.

Also, should this split up by line, or is that done at the higher level?  If this is not by line, then we should probably keep as data until it's been split up, since we might break multi-byte UTF-8 sequences otherwise.

PR #4101, opened 2022-02-05
[5.6] Add module kind enum to the package plugin API

abertelrud:
SwfitPM 5.6 nomination of #4098.

Although the different basic kinds of targets were expressed through types, the module kind (executable, test, or regular) for a source module target was not expressed in the API.  This adds that property.

**Explanation:**  This adds a property to the `SourceModuleTarget` type in `PackagePlugin` to let plugins identify the kind of target:  executable, test, or generic.  This is useful for some plugins such as docc, which may want to skip test targets when generating documentation.

**Scope of Issue:**  This affects some package plugins that need a way to filter out different purposes of targets.

**Reason for Nominating to 5.6:**  This is a refinement requested by the .docc plugin (and useful to other plugins) in new functionality in 5.6 (command plugins).  It is not part of the evolution proposal API but is intended to be included in the next proposal for expanded functionality.

**Risk:**  Very low — this vends a property that is already present in the SwiftPM data model so that it can be used by a plugin.  The risk of getting it wrong would primarily be that this new property would be incorrect but wouldn't break existing functionality.

**Reviewed By:**  @tomerd

**Automated Testing:**  A new unit test covers this new property, and checks for all three possible values using test fixtures.

**Dependencies:**  None

**Impact on CI:**  None

**How to Verify:**  Write a command plugin and verify the value of the `kind` property on an instance of the `SourceModuleTarget`.

rdar://86786080

(cherry picked from commit b7c299105592451cde264b3965aa41f0bbc3c3fc)

### DIFF
--- a/Sources/PackagePlugin/PluginInput.swift
+++ b/Sources/PackagePlugin/PluginInput.swift
@@ -102,7 +102,7 @@ fileprivate struct PluginInputDeserializer {
         let target: Target
         switch wireTarget.info {
         
-        case let .swiftSourceModuleInfo(moduleName, sourceFiles, compilationConditions, linkedLibraries, linkedFrameworks):
+        case let .swiftSourceModuleInfo(moduleName, kind, sourceFiles, compilationConditions, linkedLibraries, linkedFrameworks):
             let sourceFiles = FileList(try sourceFiles.map {
                 let path = try self.path(for: $0.basePathId).appending($0.name)
                 let type: FileType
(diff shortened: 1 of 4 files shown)


abertelrud:
@swift-ci please test

tomerd (approved):

Sources/PackagePlugin/PluginInput.swift
```diff
@@ -121,6 +121,7 @@ fileprivate struct PluginInputDeserializer {
             target = SwiftSourceModuleTarget(
                 id: String(id),
                 name: wireTarget.name,
+                kind: (kind == .executable ? .executable : (kind == .test ? .test : .regular)),
```
tomerd, line 124: nit: could be done with an extension which would make this a bit more nicer to read / maintain

abertelrud: Agreed.  Updating with the modified changes in the PR on main.

abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test

PR #4103, opened 2022-02-06
Disable a flaky part of the unit test `PackageToolTests.testPluginCom pilationBeforeBuilding`

abertelrud:
Disable a flaky part of the unit test `PackageToolTests.testPluginCom pilationBeforeBuilding`.  Filed https://bugs.swift.org/browse/SR-15831 to investigate and fix it.

We don't want to skip the entire test since there are other good things being tested here that are not flaky, such as some of the other output as well as the exit code of the subprocess.

This also fixes a separate mistake in the test, which is not relevant to this flakiness but may cause confusion because the terminology in a fixture is wrong.


abertelrud:
@swift-ci please smoke test

elsh (approved):

PR #4104, opened 2022-02-07
Cancelling `swift package` with Control-C doesn't stop any running plugins

abertelrud:
*Motivation:*

Cancelling the plugin host (such as `swift package`) should also cancel any running plugins.

*Changes:*

Add a registry that keeps track of running plugin processes, and an API to cancel them.  We can't use `ProcessSet()` since it's tied to ToolsSupportCore's implementation of a subprocess and doesn't use Foundation.  In this case we don't need such a general solution in any case.

Added a unit test that goes to some length to try to determine that process is really gone after cancelling.

rdar://87280309

### DIFF
--- a/Sources/Commands/SwiftTool.swift
+++ b/Sources/Commands/SwiftTool.swift
@@ -360,6 +360,7 @@ public class SwiftTool {
             SwiftTool.shutdownRegistry = (processSet: processSet, buildSystemRef: buildSystemRef)
             _ = SetConsoleCtrlHandler({ _ in
                 // Terminate all processes on receiving an interrupt signal.
+                DefaultPluginScriptRunner.cancelAllRunningPlugins()
                 SwiftTool.shutdownRegistry?.processSet.terminate()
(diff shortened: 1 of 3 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

neonichu (approved):

tomerd (approved):
this PR is fine, but I think we need to create a general purpose mechanism to eventually  replace this + the ProcessSet system as there are more cases where we need to interrupt

abertelrud:
> this PR is fine, but I think we need to create a general purpose mechanism to eventually replace this + the ProcessSet system as there are more cases where we need to interrupt

Absolutely.  This is to address a particular need for SwiftPM 5.6.  I can file an SR on the general issue.

abertelrud:
https://bugs.swift.org/browse/SR-15835

PR #4105, opened 2022-02-07
[5.6] Cancelling `swift package` with Control-C doesn't stop any running plugins

abertelrud:
This is a SwiftPM 5.6 cherry-pick of https://github.com/apple/swift-package-manager/pull/4104.

**Explanation:**  This makes sure that any running package plugins are killed as part of the shutdown procedure that runs when `swift` `package` receives a Control-C.

**Scope of Issue:**  This affects long-running package command plugins which a user may want to cancel by using Control-C on `swift` `package`.  Using Control-C causes SwiftPM to do some cleanup and exit, and does not reliably cause the subprocess to exit (killing `swift` `package` directly, outside of Control-C, does seem to always cause the subprocess to exit).

**Reason for Nominating to 5.6:**  This is an issue in new functionality (command plugins) that affects the user experience with long-running plugin processes. In particular it affects the new .docc plugin, which has a preview mode in which it stays running and vends the generated documentation.

**Risk:**  Low — this affects the Control-C behavior and not the normal running of plugins.  The risk when the plugin is running is minimal (process objects are being added and removed from a locked set but no other action is taken outside of handling Control-C).

**Reviewed By:**  @tomerd

**Automated Testing:**  A new unit test covers this new property, using a never-ending test plugin and checking for the absence of the subprocess after cancelling.

**Dependencies:**  None

**Impact on CI:**  None

**How to Verify:**  Run a long-running command plugin and use Control-C to interrupt the  `swift` `package` invocation.  Verify that the plugin subprocess is no longer running.

rdar://87280309

### DIFF
--- a/Sources/Commands/SwiftTool.swift
+++ b/Sources/Commands/SwiftTool.swift
@@ -361,6 +361,7 @@ public class SwiftTool {
             SwiftTool.shutdownRegistry = (processSet: processSet, buildSystemRef: buildSystemRef)
             _ = SetConsoleCtrlHandler({ _ in
                 // Terminate all processes on receiving an interrupt signal.
+                DefaultPluginScriptRunner.cancelAllRunningPlugins()
                 SwiftTool.shutdownRegistry?.processSet.terminate()
                 SwiftTool.shutdownRegistry?.buildSystemRef.buildSystem?.cancel()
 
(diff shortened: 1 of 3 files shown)


abertelrud:
@swift-ci please test

tomerd (approved):

abertelrud:
@swift-ci please test

tomerd (approved):

airspeedswift (approved):

PR #4106, opened 2022-02-08
Conditionalize references to `Foundation.Process` to allow building on iOS, tvOS, watchOS

abertelrud:
Code added in https://github.com/apple/swift-package-manager/pull/4104 unintentionally assumed the existence of `Foundation.Process`, which isn't the case on iOS, tvOS, watchOS.  This conditionalizes the use of that type — plugins aren't available on those platforms in any case, throwing an "unavailable" error.  Cancelling all running plugins on iOS, tvOS, watchOS is a no-op, since by definition, none are running.


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #4110, opened 2022-02-09
fix warning from using deprecated TSC URL API

tomerd:
motivation: TSC URL is deprecated

changes: migrate the url scheme extraction code to the one call site that needs it

### DIFF
--- a/Sources/PackageLoading/ManifestJSONParser.swift
+++ b/Sources/PackageLoading/ManifestJSONParser.swift
@@ -13,10 +13,6 @@ import Foundation
 import PackageModel
 import TSCBasic
 
-import struct TSCUtility.URL
-
-fileprivate typealias URL = Foundation.URL
-
 enum ManifestJSONParser {
     private static let filePrefix = "file://"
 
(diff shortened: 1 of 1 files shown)


Sources/PackageLoading/ManifestJSONParser.swift
```diff
@@ -415,6 +411,33 @@ enum ManifestJSONParser {
         )
     }
 
+    /// Parses the URL type of a git repository
+    /// e.g. https://github.com/apple/swift returns "https"
+    /// e.g. git@github.com:apple/swift returns "git"
+    ///
+    /// This is *not* a generic URI scheme parser!
+    private static func parseScheme(_ location: String) -> String? {
```
tomerd, line 419: we can potentially also use Foundation API here, although not obvious which. for now migrated the code as is to preserve backwards compatibility. open to other ideas.

tomerd:
@swift-ci please smoke test

Sources/PackageLoading/ManifestJSONParser.swift
abertelrud: Probably Foundation.URLComponents could help.  It does fall down pretty badly on the second one though.  So maybe try URLComponents(...).scheme first and then special-case `git@` prefix if that fails?  Not sure how much that actually buys.

tomerd: tried URLComponents but it does not handle the git@ prefix. @millenomi ?

neonichu (approved):

PR #4111, opened 2022-02-09
cleanup use of FileSystem::readFileContents

tomerd:
motivation: use common APIs to read file content and reduce use of ByteString (which we want to deprecate)

changes:
* update call sites that convert readFileContents ByteString result to Data or String to use the overloads that return these types directly
* adjust test and other call sites


tomerd:
@swift-ci please smoke test

abertelrud (approved):

PR #4115, opened 2022-02-10
Fix warning: remove unused value

elsh:


elsh:
@swift-ci smoke test

Sources/PackageModel/ManifestSourceGeneration.swift
```diff
@@ -319,7 +319,7 @@ fileprivate extension SourceCodeFragment {
             }
             self.init(enum: "target", subnodes: params)
             
-        case .product(name: let name, package: let packageName, moduleAliases: let aliases, condition: let condition):
+        case .product(name: let name, package: let packageName, moduleAliases: _, condition: let condition):
```
tomerd, line 322: should this actually include this information in the SourceCodeFragments instead of ignoring it?

elsh: It's tracked separately rdar://88722774

tomerd (approved):

Sources/PackageModel/ManifestSourceGeneration.swift
tomerd: ok

PR #4116, opened 2022-02-10
test that downloaded binary artifacts are valid archives before computing checksum

tomerd:
motivation: a web server may return an invalid archive file (say as a result of a redirect) which should emit a nicer error message instead of checksm mismatch

changes:
* add an Archiver::validate API
* implement the validate API uzing unzip -t
* add adjust tests

rdar://73225887


tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke tes

PR #4117, opened 2022-02-10
Update module aliasing functional tests 

elsh:
Resolves rdar://88722540

### DIFF
--- a/Tests/FunctionalTests/ModuleAliasingTests.swift
+++ b/Tests/FunctionalTests/ModuleAliasingTests.swift
@@ -18,8 +18,11 @@ import XCTest
  
 class ModuleAliasingTests: XCTestCase {
 
-#if swift(>=5.6)
-    func testExternalSimple() {
+    func testExternalSimple() throws {
+        #if swift(<5.6)
+        try XCTSkipIf(true, "Module aliasing is only supported on swift 5.6+")
+        #endif
+        
         fixture(name: "Miscellaneous/ModuleAliasing/DirectDeps") { prefix in
             let app = prefix.appending(components: "AppPkg")
             XCTAssertBuilds(app)
@@ -29,9 +32,16 @@ class ModuleAliasingTests: XCTestCase {
             XCTAssertFileExists(prefix.appending(components: "AppPkg", ".build", UserToolchain.default.triple.platformBuildPathComponent(), "release", "GameUtils.swiftmodule"))
             XCTAssertFileExists(prefix.appending(components: "AppPkg", ".build", UserToolchain.default.triple.platformBuildPathComponent(), "debug", "Utils.swiftmodule"))
             XCTAssertFileExists(prefix.appending(components: "AppPkg", ".build", UserToolchain.default.triple.platformBuildPathComponent(), "release", "Utils.swiftmodule"))
+            
             let result = try SwiftPMProduct.SwiftBuild.executeProcess([], packagePath: app)
-            XCTAssertEqual(result.exitStatus, .terminated(code: 0))
+            let output = try result.utf8Output() + result.utf8stderrOutput()
+            
+            #if os(Linux)
+            // FIXME: the process crashes on CentOS 7, so print the output for further investigation
+            print(output)
+            #else
+            XCTAssertEqual(result.exitStatus, .terminated(code: 0), "output: \(output)")
+            #endif
         }
     }
-#endif
 }
--- a/Tests/FunctionalTests/ModuleAliasingTests.swift
+++ b/Tests/FunctionalTests/ModuleAliasingTests.swift
@@ -37,7 +37,8 @@ class ModuleAliasingTests: XCTestCase {
             let output = try result.utf8Output() + result.utf8stderrOutput()
             
             #if os(Linux)
-            // FIXME: the process crashes on CentOS 7, so print the output for further investigation
+            // FIXME: rdar://88722540
+            // The process from above crashes on CentOS 7, so print the output for further investigation
             print(output)
             #else
             XCTAssertEqual(result.exitStatus, .terminated(code: 0), "output: \(output)")
--- a/Tests/FunctionalTests/ModuleAliasingTests.swift
+++ b/Tests/FunctionalTests/ModuleAliasingTests.swift
@@ -40,7 +40,7 @@ class ModuleAliasingTests: XCTestCase {
                 // FIXME: rdar://88722540
                 // The process from above crashes in a certain env, so print
                 // the output for further investigation
-                try XCTSkipIf(true, output)
+                try XCTSkipIf(true, "Skipping due to an expected failure being investigated in rdar://88722540\nResult: \(result.exitStatus)\nOutput: \(output)")
             }
         }
     }
--- a/Tests/FunctionalTests/ModuleAliasingTests.swift
+++ b/Tests/FunctionalTests/ModuleAliasingTests.swift
@@ -39,7 +39,7 @@ class ModuleAliasingTests: XCTestCase {
             // FIXME: rdar://88722540
             // The process from above crashes in a certain env, so print
             // the output for further investigation
-            try XCTSkipIf(result.exitStatus != 0, "Skipping due to an expected failure being investigated in rdar://88722540\nResult: \(result.exitStatus)\nOutput: \(output)")
+            try XCTSkipIf(result.exitStatus != .terminated(code: 0), "Skipping due to an expected failure being investigated in rdar://88722540\nResult: \(result.exitStatus)\nOutput: \(output)")
         }
     }
 }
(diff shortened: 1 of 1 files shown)


elsh:
@swift-ci smoke test

elsh:
@swift-ci smoke test

Tests/FunctionalTests/ModuleAliasingTests.swift
```diff
@@ -29,9 +32,17 @@ class ModuleAliasingTests: XCTestCase {
...
-            XCTAssertEqual(result.exitStatus, .terminated(code: 0))
+            let output = try result.utf8Output() + result.utf8stderrOutput()
+            
+            #if os(Linux)
+            // FIXME: rdar://88722540
+            // The process from above crashes on CentOS 7, so print the output for further investigation
+            print(output)
+            #else
+            XCTAssertEqual(result.exitStatus, .terminated(code: 0), "output: \(output)")
+            #endif
```
tomerd, line 45: is this a temporary patch to help figure out the compiler issue?

elsh:
@swift-ci smoke test

abertelrud:
If the diagnostic output is removed, what is the plan for figuring out what the original error was?

Tests/FunctionalTests/ModuleAliasingTests.swift
elsh: According to @shahmishal it's happening on macOS too; removed the process launch line where the crash happens.

tomerd:
so this test right now does nothing, correct?

elsh:
> if we need to disable the test we should do it clearly with `XCTSkip`, removing the process invocation line is too subtle

The  xctassert lines above the process launch line are still valid checks. I had added that process launch line only to get the build status. I can create a separate test to do that after figuring out in what env it fails.

tomerd:
> The xctassert lines above the process launch line are still valid checks. I had added that process launch line only to get the build status. I can create a separate test to do that after figuring out in what env it fails.

so iiuc `XCTAssertBuilds(app)` passes which means the build passes, but the resulting program fails?

I guess the higher order bit is that disabling the test is okay as temporary measure, but how do we get to the bottom of why its failing since it means the functionality is broken

elsh:
Split the test into two; one that outputs results of the process launch so we can investigate further.

elsh:
@swift-ci smoke test

abertelrud:
> I guess the higher order bit is that disabling the test is okay as temporary measure, but how do we get to the bottom of why its failing? afaiu it means the functionality as a whole is broken so disabling defeats the goal of this test?

I think this originated in a Slack discussion about the CI failure, which seems to be a SIGSEGV in the inferior specifically on CentOS.  So the suggestion was to mark the test as skipped (not removed) only on Linux (since I don't think it's possible to single out CentOS) but to also print the output in that case so that there's some hope to get to the bottom of this.  Thus my question above that if the output is removed (before skipping the test) then how do get closer to how it failed?

elsh:
@swift-ci smoke test

elsh:
Updated to print the output for all platforms since the crash happens on macOS as well.

abertelrud:
> Updated to print the output for all platforms since the crash happens on macOS as well.

Oh I see.  That should make it a lot more convenient to repro and debug as well.

tomerd:
yea this looks like it can help us understand the crash now. @elsh please share the crash output once we have it so we can all investigate the root cause

Tests/FunctionalTests/ModuleAliasingTests.swift
```diff
@@ -29,9 +32,14 @@ class ModuleAliasingTests: XCTestCase {
...
             XCTAssertFileExists(prefix.appending(components: "AppPkg", ".build", UserToolchain.default.triple.platformBuildPathComponent(), "release", "Utils.swiftmodule"))
+            
             let result = try SwiftPMProduct.SwiftBuild.executeProcess([], packagePath: app)
-            XCTAssertEqual(result.exitStatus, .terminated(code: 0))
+            let output = try result.utf8Output() + result.utf8stderrOutput()
+           
+            // FIXME: rdar://88722540
+            // The process from above crashes in a certain env, so print
+            // the output for further investigation
+            print(output)
```
abertelrud, line 42: Would it make sense to still mark as skipped if it fails here (after emitting the output)?  Otherwise it looks as if it passed even if the inferior crashed.

I think all that's wanted here, temporarily, while investigating this was to:
- emit the output if it fails
- treat it as skipped and not failure

The current code would mark as success whether or not the inferior crashes, which seems unfortunate.

tomerd: +1 on this idea

elsh: makes sense; updated

elsh:
@swift-ci smoke test

Tests/FunctionalTests/ModuleAliasingTests.swift
```diff
@@ -29,9 +32,16 @@ class ModuleAliasingTests: XCTestCase {
...
+            
             let result = try SwiftPMProduct.SwiftBuild.executeProcess([], packagePath: app)
-            XCTAssertEqual(result.exitStatus, .terminated(code: 0))
+            let output = try result.utf8Output() + result.utf8stderrOutput()
+           
+            if result.exitStatus != 0 {
+                // FIXME: rdar://88722540
+                // The process from above crashes in a certain env, so print
+                // the output for further investigation
+                try XCTSkipIf(true, output)
```
abertelrud, line 43: Somewhat of a nitpick but for anyone else who looks at these logs, it would be great if the message was something like: `skipping due to expected failure being investigated in rdar://88722540\nresult: \(result.exitStatus)\noutput: \(output)`.  I know that's a bit verbose but then if others are looking at failures it's clear what's going on and a reference to the Radar.  Plus it tells us the exit status, in case it suddenly changes from signal 11 to something else.

abertelrud: I think you could even do `XCTSkipIf(result.exitStatus != 0, "...")` but maybe that's overthinking this.  The comment is still useful though.

abertelrud: I'm saying this partly because there have been cases where printed output from our tests that had "error:" at the start of a line was found by people grepping the logs and caused some confusion (the tests was that a failure occurred and _expected_ the "error:" message to be there).

abertelrud: So clarity is probably good here for anyone looking at logs who has never heard of this particular failure.

abertelrud (approved):
LGTM!

abertelrud:
So this only repros in CI, but on any platform?

elsh:
@swift-ci smoke test

elsh:
> So this only repros in CI, but on any platform?

So far reported failure is on macOS and CentOS 7 cc @shahmishal

abertelrud:
@elsh 
```15:34:24 
/Users/buildnode/jenkins/workspace/swift-package-manager-with-xcode-self-hosted-PR-osx/branch-main/swiftpm/Tests/FunctionalTests/ModuleAliasingTests.swift:42:45: error: referencing operator function '!=' on 'BinaryInteger' requires that 'ProcessResult.ExitStatus' conform to 'BinaryInteger'
15:34:24             try XCTSkipIf(result.exitStatus != 0, "Skipping due to an expected failure being investigated in rdar://88722540\nResult: \(result.exitStatus)\nOutput: \(output)")
15:34:24                                             ^
15:34:24 Swift.BinaryInteger:1:11: note: where 'Self' = 'ProcessResult.ExitStatus'
15:34:24 extension BinaryInteger {
15:34:24           ^
```
That should be `result.exitStatus != .terminated(code: 0)`.  Did what was there before build locally?

elsh:
@swift-ci smoke test

Tests/FunctionalTests/ModuleAliasingTests.swift
```diff
@@ -29,9 +32,14 @@ class ModuleAliasingTests: XCTestCase {
...
             XCTAssertFileExists(prefix.appending(components: "AppPkg", ".build", UserToolchain.default.triple.platformBuildPathComponent(), "release", "Utils.swiftmodule"))
+            
             let result = try SwiftPMProduct.SwiftBuild.executeProcess([], packagePath: app)
-            XCTAssertEqual(result.exitStatus, .terminated(code: 0))
+            let output = try result.utf8Output() + result.utf8stderrOutput()
+           
+            // FIXME: rdar://88722540
+            // The process from above crashes in a certain env, so print
+            // the output for further investigation
+            try XCTSkipIf(result.exitStatus != .terminated(code: 0), "Skipping due to an expected failure being investigated in rdar://88722540\nResult: \(result.exitStatus)\nOutput: \(output)")
```
tomerd, line 42: nice

abertelrud:
@elsh @shahmishal is this ready to merge?

shahmishal:
Merged to unblock CI

PR #4118, opened 2022-02-10
improve handling of redundant output on stdout for commands that emit consumable information

tomerd:
motvation: some CLI command emit consumable informaiton such as json. in those case we dont want to include any other information on stdout, so that it can be consumed by pipes

changes:

* for the following commands:
  * swift package dump-package
  * swift package dump-pif
  * swift package describe
  * swift package experimental-api-diff
  * swift package show-dependencies
  * swift test --list
* redirect stdout to stderr prior to calling other actions
* adjust result printing to always use stdout

rdar://78612028

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -173,6 +173,9 @@ extension SwiftPackageTool {
         var type: DescribeMode = .text
         
         func run(_ swiftTool: SwiftTool) throws {
+            // redirect all other output to stderr so that the output is the only thing that is printed on stdout
+            swiftTool.redirectStdoutToStderr()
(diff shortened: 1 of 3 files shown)


tomerd:
@swift-ci pleas smoke test

tomerd:
@swift-ci smoke test

neonichu (approved):

tomerd:
> It's an existing function and might just be me, but I find the name redirectStdoutToStderr very misleading. I was thinking we were doing something with the actual underlying streams when we're actually just setting a variable :) Maybe useStdErrAsOutputStream or something could be clearer.

good point. planning to make more changes in this area, including dealing with this function so will handle it then

PR #4126, opened 2022-02-11
disable module aliasing test while investigating

tomerd:
motivation: stable ci

changes: disable new module aliasing tests


tomerd:
@swift-ci please smoke test

elsh (approved):

PR #4131, opened 2022-02-12
Make it impossible to create an invalid `BuildSettingCondition`

sjavora:
Instead of a single `when` method that takes two optional values, use multiple methods so that it is impossible for both to be `nil`.

### DIFF
--- a/Sources/PackageDescription/BuildSettings.swift
+++ b/Sources/PackageDescription/BuildSettings.swift
@@ -58,20 +58,41 @@ public struct BuildSettingCondition: Encodable {
         self.config = config
     }
     
-    /// Creates a build setting condition.
-    ///
-    /// At least one parameter is mandatory.
-    ///
-    /// - Parameters:
-    ///   - platforms: The applicable platforms for this build setting condition.
-    ///   - configuration: The applicable build configuration for this build setting condition.
+    @available(_PackageDescription, deprecated: 999.0)
     public static func when(
         platforms: [Platform]? = nil,
         configuration: BuildConfiguration? = nil
     ) -> BuildSettingCondition {
-        // FIXME: This should be an error, not a precondition.
         precondition(!(platforms == nil && configuration == nil))
         return BuildSettingCondition(platforms: platforms, config: configuration)
+    }
+    
+    /// Creates a build setting condition.
+    ///
+    /// - Parameters:
+    ///   - platforms: The applicable platforms for this build setting condition.
+    ///   - configuration: The applicable build configuration for this build setting condition.
+    @available(_PackageDescription, introduced: 999.0)
+    public static func when(platforms: [Platform], configuration: BuildConfiguration) -> BuildSettingCondition {
+        BuildSettingCondition(platforms: platforms, config: configuration)
+    }
+    
+    /// Creates a build setting condition.
+    ///
+    /// - Parameters:
+    ///   - platforms: The applicable platforms for this build setting condition.
+    @available(_PackageDescription, introduced: 999.0)
+    public static func when(platforms: [Platform]) -> BuildSettingCondition {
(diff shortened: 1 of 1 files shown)


Sources/PackageDescription/BuildSettings.swift
```diff
@@ -60,18 +60,27 @@ public struct BuildSettingCondition: Encodable {
...
+    public static func when(platforms: [Platform], configuration: BuildConfiguration) -> BuildSettingCondition {
+        BuildSettingCondition(platforms: platforms, config: configuration)
+    }
+    
+    /// Creates a build setting condition.
     ///
     /// - Parameters:
     ///   - platforms: The applicable platforms for this build setting condition.
+    public static func when(platforms: [Platform]) -> BuildSettingCondition {
+        BuildSettingCondition(platforms: platforms, config: nil)
```
tomerd, line 75: nit: use `.none` instead of `nil` to make it more semantically clear

Sources/PackageDescription/BuildSettings.swift
```diff
@@ -60,18 +60,27 @@ public struct BuildSettingCondition: Encodable {
...
     ///   - configuration: The applicable build configuration for this build setting condition.
-    public static func when(
-        platforms: [Platform]? = nil,
-        configuration: BuildConfiguration? = nil
-    ) -> BuildSettingCondition {
-        // FIXME: This should be an error, not a precondition.
-        precondition(!(platforms == nil && configuration == nil))
-        return BuildSettingCondition(platforms: platforms, config: configuration)
+    public static func when(configuration: BuildConfiguration) -> BuildSettingCondition {
+        BuildSettingCondition(platforms: nil, config: configuration)
```
tomerd, line 83: nit: use `.none` instead of `nil` to make it more semantically clear

tomerd:
@swift-ci please smoke test

tomerd:
thanks @sjavora

sjavora:
@tomerd Updated to use `.none` 👍 

In the meantime I also found https://github.com/apple/swift-package-manager/blob/2ea136b5703039b5a596c8a61d090e4dc300ee98/Sources/PackageDescription/PackageDescriptionSerialization.swift#L354

...which seems like it got copy/pasted as there is just one optional parameter. Should I remove that as well? `TargetDependencyCondition` is `Encodable` so I'm not 100% sure removing the optionality won't break something...

tomerd:
thanks @sjavora 

`PackageDescriptionSerialization` is a bit more sensitive since it drives the manifest parsing. this does need to be improved, but maybe a topic for a separate PR

SDGGiesbrecht:
This is a public, source‐breaking change to the `PackageDescription` module. (For example, `.when(platforms: [.macOS], configuration: nil)` will break.) Does the old method need to be retained with tools version constraints and/or deprecation notices?

tomerd:
> This is a public, source‐breaking change to the PackageDescription module. (For example, .when(platforms: [.macOS], configuration: nil) will break.) Does the old method need to be retained with tools version constraints and/or deprecation notices?

good catch @SDGGiesbrecht did not notice this was in `PackageDescription`. @sjavora please restore the previous API as well but mark as deprecated

SDGGiesbrecht:
👍 Looks good now.

tomerd (approved):

tomerd:
@swift-ci please smoke test

shahmishal:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

shahmishal:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

PR #4132, opened 2022-02-12
Refactor nested `if`s into a switch statement

sjavora:
Also reduces some of the assignment repetitions, which makes the code even clearer (IMO).


tomerd (approved):
nice. thanks you

tomerd:
@swift-ci please smoke test

PR #215, opened 2016-03-21
Remove release-mode dead-strip

mxcl:
Refs: https://bugs.swift.org/browse/SR-521

PR #4135, opened 2022-02-13
Add -dead_strip / --gc-sections for release builds

keith:
In the past it was unsafe to pass -dead_strip because there was some
required swift metadata that was stripped. It seems that at this point all
of those cases are either fixed or marked as @llvm.used, so passing
these flags should safely reduce binary size in some cases. If there are
other cases where this isn't safe we should likely annotate them as
@llvm.used.

Related: https://bugs.swift.org/browse/SR-521 https://github.com/apple/swift-package-manager/pull/215

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ Swift 5.7
  
   Add a `--disable-testable-imports` flag to `swift test` with which tests are built without the testability feature (`import @tstable` disabled).
 
+* [#4135]
+
+  Enable linker dead stripping for all platforms. This can be disabled with `--disable-dead-strip`
+
 Swift 5.6
 -----------
 * [SE-0332]
@@ -217,3 +221,4 @@ Swift 3.0
 [#3901]: https://github.com/apple/swift-package-manager/pull/3901
 [#3942]: https://github.com/apple/swift-package-manager/pull/3942
 [#4119]: https://github.com/apple/swift-package-manager/pull/4119
+[#4135]: https://github.com/apple/swift-package-manager/pull/4135
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -1201,7 +1201,7 @@ public final class ProductBuildDescription {
     }
 
     private var deadStripArguments: [String] {
-        if buildParameters.linkerDeadStrip {
+        if !buildParameters.linkerDeadStrip {
             return []
         }
 
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -1200,6 +1200,21 @@ public final class ProductBuildDescription {
         return args.filter({ !invalidArguments.contains($0) })
     }
 
+    private var deadStripArguments: [String] {
+        switch buildParameters.configuration {
+        case .debug:
+            return []
+        case .release:
+            if buildParameters.triple.isDarwin() {
+                return ["-Xlinker", "-dead_strip"]
+            } else if buildParameters.triple.isWindows() {
+                return ["-Xlinker", "/OPT:REF"]
+            } else {
+                return ["-Xlinker", "--gc-sections"]
+            }
+        }
+    }
+
     /// The arguments to link and create this product.
     public func linkArguments() throws -> [String] {
         var args = [buildParameters.toolchain.swiftCompiler.pathString]
@@ -1246,12 +1261,14 @@ public final class ProductBuildDescription {
             case .manifest:
                 args += ["-emit-executable"]
             }
+            args += deadStripArguments
         case .library(.dynamic):
             args += ["-emit-library"]
             if buildParameters.triple.isDarwin() {
                 let relativePath = "@rpath/\(buildParameters.binaryRelativePath(for: product).pathString)"
                 args += ["-Xlinker", "-install_name", "-Xlinker", relativePath]
             }
+            args += deadStripArguments
         case .executable, .snippet:
             // Link the Swift stdlib statically, if requested.
             if buildParameters.shouldLinkStaticSwiftStdlib {
@@ -1262,6 +1279,7 @@ public final class ProductBuildDescription {
                 }
             }
             args += ["-emit-executable"]
+            args += deadStripArguments
             
             // If we're linking an executable whose main module is implemented in Swift,
             // we rename the `_<modulename>_main` entry point symbol to `_main` again.
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -1201,6 +1201,10 @@ public final class ProductBuildDescription {
     }
 
     private var deadStripArguments: [String] {
+        if buildParameters.disableDeadStrip {
+            return []
+        }
+
         switch buildParameters.configuration {
         case .debug:
             return []
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -1201,7 +1201,7 @@ public final class ProductBuildDescription {
     }
 
     private var deadStripArguments: [String] {
-        if buildParameters.disableDeadStrip {
+        if buildParameters.linkerDeadStrip {
             return []
         }
 
--- a/Sources/Commands/Options.swift
+++ b/Sources/Commands/Options.swift
@@ -375,6 +375,9 @@ public struct SwiftToolOptions: ParsableArguments {
     @Option(name: .customLong("resolver-fingerprint-checking"))
     var resolverFingerprintCheckingMode: FingerprintCheckingMode = .warn
 
+    @Flag(name: .customLong("disable-dead-strip"), help: "Disable dead code stripping by the linker")
+    var disableDeadStrip: Bool = false
+
     @Flag(name: .customLong("netrc"), help: .hidden)
     var _deprecated_netrc: Bool = false
 
--- a/Sources/Commands/Options.swift
+++ b/Sources/Commands/Options.swift
@@ -375,8 +375,11 @@ public struct SwiftToolOptions: ParsableArguments {
     @Option(name: .customLong("resolver-fingerprint-checking"))
     var resolverFingerprintCheckingMode: FingerprintCheckingMode = .warn
 
-    @Flag(name: .customLong("disable-dead-strip"), help: "Disable dead code stripping by the linker")
-    var disableDeadStrip: Bool = false
+    @Flag(
+        name: .customLong("dead-strip"),
+        inversion: .prefixedEnableDisable,
+        help: "Disable/enable dead code stripping by the linker")
+    var linkerDeadStrip: Bool = true
 
     @Flag(name: .customLong("netrc"), help: .hidden)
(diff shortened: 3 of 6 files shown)


tomerd:
thanks @Keith this makes sense based on what @kubamracek was saying on another thread. couple of questions / ideas:

1. should this be behind a flag so that its possible to disable/enable (ie enabled by default but possible to disable If causes issues as an escape hatch)
2. do we need to add a small unit test? especially if do add the flag (which I lean towards thinking we should)
3. do we know that this also works fine on Linux and Window which use different linkers afaik?

Sources/Build/BuildPlan.swift
```diff
@@ -1200,6 +1200,21 @@ public final class ProductBuildDescription {
...
+    private var deadStripArguments: [String] {
+        switch buildParameters.configuration {
+        case .debug:
+            return []
+        case .release:
+            if buildParameters.triple.isDarwin() {
+                return ["-Xlinker", "-dead_strip"]
+            } else if !buildParameters.triple.isWindows() {
+                return ["-Xlinker", "--gc-sections"]
+            } else {
```
compnerd, line 1212: Please add a case for windows, the flag should be `/OPT:REF`.

compnerd (changes requested):
CC: @kubamracek (I believe Kuba was involved in the changes for the metadata preservation).

Sources/Build/BuildPlan.swift
keith: Thanks, lmk if this update looks right!

keith:
A flag sounds fine to me, would folks prefer opt-in or opt-out? I'm not sure how much testing an opt-in flag would get at this point, but I'll defer to you all for your risk preference.

I did test this on Linux and I think `--gc-sections` is safe for all ELF linkers

tomerd:
@Keith I would go with opt out given that @kubamracek believes this is safe enough. we need @compnerd init on Windows too.

kubamracek:
I'm definitely supportive of this, I'm not aware of any problems/bugs in this area (but TBH I have mostly only worked with Darwin / Mach-O / ld64, and almost not at all with other platforms / file formats / linkers). I am even supportive in making dead stripping be the default behavior, but I want to point out that there are cases where dead-stripping will affect a program -- and mostly it's "shady" things like dlsym'ing for an otherwise unused symbol, and probably only really affect C code and not Swift code. Before we make dead-stripping the default, we should acknowledge that stuff like this might be affected. I personally think it's the right choice (and programs relying on not-dead-stripping would either need to opt out, or fix their code, there's some simple options like `__attribute__((used))`), but I wonder what the thinking of others is :)

tomerd:
its also worth mentioning that at the current state of things, such flag would have little effect on Swift code, and would mostly effect C code built with SwiftPM

keith:
Updated with a flag!

Sources/Commands/Options.swift
```diff
@@ -375,6 +375,9 @@ public struct SwiftToolOptions: ParsableArguments {
     @Option(name: .customLong("resolver-fingerprint-checking"))
     var resolverFingerprintCheckingMode: FingerprintCheckingMode = .warn
 
+    @Flag(name: .customLong("disable-dead-strip"), help: "Disable dead code stripping by the linker")
+    var disableDeadStrip: Bool = false
```
tomerd, line 379:
```suggestion
    @Flag(name: .customLong("dead-strip"), inversion: .prefixedEnableDisable, help: "Disable/enable dead code stripping by the linker. Enabled by default")
    var enableDeadStrip: Bool = true
```

keith: Sounds good, updated, I figured since it was a disable only we didn't need the positive case, but nbd to me! Also it looks like swift-argument-parser automatically adds the default to the help so I removed that piece here

compnerd:
Please do a test build on Windows just to make sure that things don't break.

tomerd:
> Please do a test build on Windows just to make sure that things don't break.

@compnerd @shahmishal is there way to trigger CI for this?

tomerd:
@swift-ci please smoke test

shahmishal:
> > Please do a test build on Windows just to make sure that things don't break.
> 
> @compnerd @shahmishal is there a way to trigger CI for Windows?

Hopefully soon we will be able to from SwiftPM, however for now you will have to do cross PR to trigger Windows testing.

tomerd:
> Hopefully soon we will be able to from SwiftPM, however for now you will have to do cross PR to trigger Windows testing.

thanks @shahmishal created https://github.com/apple/swift/pull/41374

what is the trigger phrase for Windows verification?

shahmishal:
`@swift-ci test Windows`

compnerd:
@shahmishal does that now trigger a different job?  There are two separate builds - the `@swift-ci please test Windows` builds and tests the toolchain (which is why it is quick); the full toolchain build is almost twice as long but builds a full toolchain image (and tests a bit more, though still in sufficient for this as of yet :-()

tomerd:
this is pending verification on Windows, see discussion on https://github.com/apple/swift/pull/41374

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ Swift 5.7
  
   Add a `--disable-testable-imports` flag to `swift test` with which tests are built without the testability feature (`import @tstable` disabled).
 
+* [#4135]
```
tomerd, line 10: sorry also need to add a link, e.g. https://github.com/apple/swift-package-manager/pull/4135/commits/6e3ed66afad54dcd2cee8390b58ace2164117912#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL202

tomerd:
@swift-ci please smoke test

tomerd (approved):
pending verification on Windows, see discussion on https://github.com/apple/swift/pull/41374

tomerd:
@swift-ci please smoke test

compnerd:
:-( it seems that something isn't really correct:

```
S:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swiftc.exe -Xlinker -debug -L S:\SourceCache\swift-win32\.build\x86_64-unknown-windows-msvc\release -o S:\SourceCache\swift-win32\.build\x86_64-unknown-windows-msvc\release\Calculator.exe -module-name Calculator -lSwiftCOM -lcassowary -emit-executable @S:\SourceCache\swift-win32\.build\x86_64-unknown-windows-msvc\release\Calculator.product\Objects.LinkFileList -target x86_64-unknown-windows-msvc -lUser32 -lComCtl32 -sdk S:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk -libc MD -I S:\Library\Developer\Platforms\Windows.platform\Developer\Library\XCTest-development\usr\lib\swift\windows\x86_64 -L S:\Library\Developer\Platforms\Windows.platform\Developer\Library\XCTest-development\usr\lib\swift\windows
```

Note that the `-Xlinker /OPT:REF` is missing.  That was built with `-c release`.

keith:
I inverted the boolean when we renamed from disable -> enable, fixed, sorry about that!

tomerd:
@swift-ci please smoke test

compnerd:
~~~
S:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swift-frontend.exe -modulewrap S:\SourceCache\swift-win32\.build\x86_64-unknown-windows-msvc\release\Calculator.build\Calculator.swiftmodule -target x86_64-unknown-windows-msvc -o S:\SourceCache\swift-win32\.build\x86_64-unknown-windows-msvc\release\Calculator.build\Calculator.o
S:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swiftc.exe -Xlinker -debug -L S:\SourceCache\swift-win32\.build\x86_64-unknown-windows-msvc\release -o S:\SourceCache\swift-win32\.build\x86_64-unknown-windows-msvc\release\Calculator.exe -module-name Calculator -emit-executable -Xlinker /OPT:REF @S:\SourceCache\swift-win32\.build\x86_64-unknown-windows-msvc\release\Calculator.product\Objects.LinkFileList -target x86_64-unknown-windows-msvc -lUser32 -lComCtl32 -lOle32 -lPortableDeviceGuids -sdk S:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk -libc MD -I S:\Library\Developer\Platforms\Windows.platform\Developer\Library\XCTest-development\usr\lib\swift\windows\x86_64 -L S:\Library\Developer\Platforms\Windows.platform\Developer\Library\XCTest-development\usr\lib\swift\windows
~~~

Built and ran the calculator demo, seems to be fine, and the linker was also fine with the flags.  I think that this should be fine from the Windows side at least.  And the command line indicates that the flag is now there.

compnerd (approved):

keith:
thanks for testing!

tomerd:
thanks @keith

keith:
Thanks all!

drodriguez:
@keith: do you remember which linkers and versions did you use to test in Linux ELF? With a recent LLD you have https://github.com/apple/llvm-project/commit/6d2d3bd0a61f5fc7fd9f61f48bc30e9ca77cc619 which enables GC of the start/stop symbols, which seems to be affecting the Swift metadata section symbols in our setup.

keith:
Interesting, I just tried the defaults, do you have a repro case I can debug with?

drodriguez:
Not with the upstream build system, sadly. I was just trying to compile a newer SwiftPM with a recently compiled SwiftPM that includes this (and pointing to a quite recent LLD) and started seeing these linking problems about the `start` and `stop` symbols being hidden undefined. I think it will need a fix in how the sections are declared in https://github.com/apple/swift/blob/main/stdlib/public/runtime/SwiftRT-ELF.cpp but a first idea adding `"R"` to the section declaration did fix many test in the check-swift test suite, but not all of them (IIRC there were 76 left, many of them crashes).

I will try to keep looking. For the time being I can work around it with `--disable-dead-strip` so it is not the end of the world.

keith:
I think the assumption from folks here was generally that things should be correctly referenced / non-strippable today, so if that's not the case we may have to temporarily reconsider this default unless we can come up with a fix for those

buttaface:
I just hit this when building the latest Mar. 9 trunk source snapshot of the toolchain natively on my Android AArch64 phone that uses lld, which is fixed by disabling this flag by default in the SPM source. This may be a recent lld regression, which this new flag triggers for us on Android, as [this guy reported the issue on linux with lld 13, unrelated to this SPM flag, on the forum last month](https://forums.swift.org/t/swift-and-clang-interop-broken-with-clang-13-on-linux/55203), or maybe lld or clang 13 enabled this flag by default all of a sudden on linux too.

Strangely, my Android CI, which cross-compiles an Android SDK and several packages against it, worked fine on [the first run with this new Mar. 9 SPM](https://github.com/buttaface/swift-android-sdk/actions/runs/1972820447), I'll look into that further.

There may be issues with this flag and lld on ELF platforms interacting badly with `swiftrt.o`, we'll need to investigate.

PR #3687, opened 2021-08-25
improve and relax the validation of explicit dependency names

tomerd:
motivation: follow up work to relaxing the requirement to set an explicit name on dependency, this relaxes the requirement that such name (if/when set) must match the manifest's name attribute, which we want to deprecate over time

changes:
* change the validation flow to ignore the name set in the manifest, instead manage an in-memory list of names used for target based pacakge lookup which suffices
* remove a few tests that are no longer needed given the relaxed requirment
* fix a couple of tests that were always wrong but previous validation did not catch

PR #4138, opened 2022-02-14
Fix llbuild package name for SWIFTCI_USE_LOCAL_DEPS

keith:
Without this fix when this var is set I see:

```
'SwiftPM' /path/to/oss-swift/swiftpm: error: 'swiftpm' dependency on '/path/to/oss-swift/llbuild' has an explicit name 'swift-llbuild' which does not match the name 'llbuild' set for '/path/to/oss-swift/llbuild'
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -568,10 +568,10 @@ if ProcessInfo.processInfo.environment["SWIFTPM_LLBUILD_FWK"] == nil {
         // In Swift CI, use a local path to llbuild to interoperate with tools
         // like `update-checkout`, which control the sources externally.
         package.dependencies += [
-            .package(name: "swift-llbuild", path: "../llbuild"),
+            .package(name: "llbuild", path: "../llbuild"),
         ]
     }
-    package.targets.first(where: { $0.name == "SPMLLBuild" })!.dependencies += [.product(name: "llbuildSwift", package: "swift-llbuild")]
+    package.targets.first(where: { $0.name == "SPMLLBuild" })!.dependencies += [.product(name: "llbuildSwift", package: "llbuild")]
 }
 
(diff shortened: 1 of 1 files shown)


keith:
Maybe my checkout state is invalid somehow?

tomerd (changes requested):
we normally do not use `SWIFTCI_USE_LOCAL_DEPS` mode for dependencies unless we _need_ to make changes to those as well, and even then you can always change the dep to path based instead of url based. 

in cases we do decide to use `SWIFTCI_USE_LOCAL_DEPS` we check out `swift-llbuild` into `llbuild` since this is how our CI does things (for historical reasons) see https://github.com/apple/swift-package-manager/blob/main/CONTRIBUTING.md#the-bootstrap-script

keith:
I believe I have the setup you're describing:

```
% git remote -v
origin  https://github.com/apple/swift-llbuild.git (fetch)
origin  https://github.com/apple/swift-llbuild.git (push)
% pwd
/Users/ksmiley/dev/oss-swift/llbuild
```

But I believe the problem is the mismatch with this `llbuild`, not the repo name? https://github.com/apple/swift-llbuild/blob/25333daca549aed69004c9b3d4c6b69d3f32d1d0/Package.swift#L9

tomerd:
@Keith which version of swift are you using?

keith:
Ah 5.5 aka stable Xcode. If it's fixed later we can close

tomerd:
and you see this when you build with the command line or through Xcode?

keith:
Command line. Just `swift build` with this var set. It sounds unexpected so I can debug more

tomerd:
this validation was removed in https://github.com/apple/swift-package-manager/pull/3687 so 5.6

keith:
Thanks, sorry for the noise!

PR #4146, opened 2022-02-16
make XCTest optional on macOS for operations that do not require testing

tomerd:
motivation: users should be able to use SwiftPM on macOS without Xcode installed (with the exception of testing for now)

changes:
* in UserToolchain make xctest lookup optional on macOS and only validate its exists when trying to perform tests
* consolidate two different representations of xctest path (one was for Windows and the other for macOS)

rdar://88651484


tomerd:
@swift-ci please smoke test

abertelrud (approved):

tomerd:
@swift-ci please smoke test

PR #4147, opened 2022-02-16
cleanup workspace binary artifacts tests

tomerd:
motivation: binary artifacts tests had redundant complexity that made them difficult to maintain

changes: remove extra layer of indirection when not required to the test by making the binary targets on the root package instead of on dependencies


tomerd:
@swift-ci please smoke test

abertelrud (approved):

PR #4148, opened 2022-02-17
resolve source control dependencies via the registry when available

tomerd:
motivation: support mixed dependencies graphs where some dependencies come from source control while other from registry, and encourage the transition to the registry

changes:
* introduce new CLI flag to determine the source control to registry transformation mode (disabled, identity only, swizzle)
* wrap manifest loading with a transformer that translates source control dependencies to registry ones when the dependency is found in the registry, based on the mode
* add a bunch of tests

### DIFF
--- a/Sources/Commands/Options.swift
+++ b/Sources/Commands/Options.swift
@@ -376,11 +376,22 @@ public struct SwiftToolOptions: ParsableArguments {
         static func name(for value: Self) -> NameSpecification {
             switch value {
             case .disabled:
-                return .customLong("disable-scm-to-regsitry-lookup")
+                return .customLong("disable-scm-to-registry-transformation")
             case .identity:
                 return .customLong("use-registry-identity-for-scm")
             case .swizzle:
-                return .customLong("raplace-scm-with-registry")
+                return .customLong("replace-scm-with-registry")
+            }
+        }
+
+        static func help(for value: SwiftToolOptions.SourceControlToRegistryDependencyTransformation) -> ArgumentHelp? {
+            switch value {
+            case .disabled:
+                return "disable source control to registry transformation"
+            case .identity:
+                return "look up source control dependencies in the registry and use their registry identity when possible to help deduplicate across the two origins"
+            case .swizzle:
+                return "look up source control dependencies in the registry and use the registry to retrieve them instead of source control when possible"
             }
         }
     }
--- a/Sources/Commands/Options.swift
+++ b/Sources/Commands/Options.swift
@@ -267,6 +267,9 @@ public struct SwiftToolOptions: ParsableArguments {
     @Flag(name: [.long, .customLong("disable-automatic-resolution"), .customLong("only-use-versions-from-resolved-file")], help: "Only use versions from the Package.resolved file and fail resolution if it is out-of-date")
     var forceResolvedVersions: Bool = false
 
+    @Flag(help: "Enable or disable indexing-while-building feature")
+    var indexStoreMode: StoreMode = .autoIndexStore
+
     // @Flag works best when there is a default value present
     // if true, false aren't enough and a third state is needed
     // nil should not be the goto. Instead create an enum
@@ -274,30 +277,11 @@ public struct SwiftToolOptions: ParsableArguments {
         case autoIndexStore
         case enableIndexStore
         case disableIndexStore
-
-        /// The mode to use for indexing-while-building feature.
-        var indexStoreMode: BuildParameters.IndexStoreMode {
-            switch self {
-            case .autoIndexStore:
-                return .auto
-            case .enableIndexStore:
-                return .on
-            case .disableIndexStore:
-                return .off
-            }
-        }
     }
-
-    @Flag(help: "Enable or disable indexing-while-building feature")
-    var indexStoreMode: StoreMode = .autoIndexStore
 
     /// Whether to enable generation of `.swiftinterface`s alongside `.swiftmodule`s.
     @Flag(name: .customLong("enable-parseable-module-interfaces"))
     var shouldEnableParseableModuleInterfaces: Bool = false
-
-    /// Write dependency resolver trace to a file.
-    @Flag(name: .customLong("trace-resolver"), help: .hidden)
-    var _deprecated_enableResolverTrace: Bool = false
 
     /// The number of jobs for llbuild to start (aka the number of schedulerLanes)
     @Option(name: .shortAndLong, help: "The number of jobs to spawn in parallel during the build process")
@@ -360,17 +344,17 @@ public struct SwiftToolOptions: ParsableArguments {
 
     /// Whether to use keychain for authenticating with remote servers
     /// when downloading binary artifacts or communicating with a registry.
-#if canImport(Security)
+    #if canImport(Security)
     @Flag(inversion: .prefixedEnableDisable,
           exclusivity: .exclusive,
           help: "Search credentials in macOS keychain")
     var keychain: Bool = true
-#else
+    #else
     @Flag(inversion: .prefixedEnableDisable,
           exclusivity: .exclusive,
           help: .hidden)
     var keychain: Bool = false
-#endif
+    #endif
     
     @Option(name: .customLong("resolver-fingerprint-checking"))
     var resolverFingerprintCheckingMode: FingerprintCheckingMode = .warn
@@ -381,6 +365,26 @@ public struct SwiftToolOptions: ParsableArguments {
         help: "Disable/enable dead code stripping by the linker")
     var linkerDeadStrip: Bool = true
 
+    @Flag(help: "Define automatic transformation of source control based dependencies to registry based ones")
+    var sourceControlToRegistryDependencyTransformation: SourceControlToRegistryDependencyTransformation = .disabled
+
+    enum SourceControlToRegistryDependencyTransformation: EnumerableFlag {
+        case disabled
+        case identity
+        case swizzle
+
+        static func name(for value: Self) -> NameSpecification {
+            switch value {
+            case .disabled:
+                return .customLong("disable-scm-to-regsitry-lookup")
+            case .identity:
+                return .customLong("use-registry-identity-for-scm")
+            case .swizzle:
+                return .customLong("raplace-scm-with-registry")
+            }
+        }
+    }
+
     @Flag(name: .customLong("netrc"), help: .hidden)
     var _deprecated_netrc: Bool = false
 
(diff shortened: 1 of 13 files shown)


Sources/Workspace/Workspace.swift
```diff
@@ -4185,6 +4189,169 @@ extension Workspace.Location {
     }
 }
 
+#warning("tomer: experiment")
+extension Workspace {
+    fileprivate struct RegistryAwareManifestLoader: ManifestLoaderProtocol {
```
tomerd, line 4194: this is the key change. it basically translates source control dependencies to registries ones when possible.

this is still in wip state but the concept is clear

Sources/Workspace/Workspace.swift
```diff
@@ -4185,6 +4189,219 @@ extension Workspace.Location {
...
+                        // an implementation mistake
+                        throw InternalError("unexpected non-source-control dependency")
+                    }
+                    switch settings.requirement {
+                    case .exact(let version):
+                        modifiedDependency = .registry(identity: registryIdentity, requirement: .exact(version), productFilter: settings.productFilter)
+                    case .range(let range):
+                        modifiedDependency = .registry(identity: registryIdentity, requirement: .range(range), productFilter: settings.productFilter)
+                    case .revision:
+                        // FIXME: warn?
```
yim-lee, line 4321: Deprecation warning maybe? Though feels more like it should be "info" during the initial launch.

Sources/Commands/Options.swift
```diff
@@ -267,38 +267,22 @@ public struct SwiftToolOptions: ParsableArguments {
...
-    var indexStoreMode: StoreMode = .autoIndexStore
-
     /// Whether to enable generation of `.swiftinterface`s alongside `.swiftmodule`s.
     @Flag(name: .customLong("enable-parseable-module-interfaces"))
     var shouldEnableParseableModuleInterfaces: Bool = false
 
-    /// Write dependency resolver trace to a file.
-    @Flag(name: .customLong("trace-resolver"), help: .hidden)
-    var _deprecated_enableResolverTrace: Bool = false
-
```
tomerd, line 301: small unrelated cleanup (moved code around)

Sources/Commands/Options.swift
```diff
@@ -381,6 +365,26 @@ public struct SwiftToolOptions: ParsableArguments {
         help: "Disable/enable dead code stripping by the linker")
     var linkerDeadStrip: Bool = true
 
+    @Flag(help: "Define automatic transformation of source control based dependencies to registry based ones")
+    var sourceControlToRegistryDependencyTransformation: SourceControlToRegistryDependencyTransformation = .disabled
```
tomerd, line 369: the new flag, it a mouth full!

Sources/Commands/SwiftTool.swift
```diff
@@ -1114,6 +1115,13 @@ extension ObservabilityMetadata {
     }
 }
 
+extension Workspace.ManagedDependency {
+    fileprivate var isEdited: Bool {
+        if case .edited = self.state { return true }
+        return false
+    }
+}
```
tomerd, line 1123: the code that moved around

Sources/Workspace/Workspace.swift
```diff
@@ -4185,6 +4189,219 @@ extension Workspace.Location {
...
+        }
+
+        private func mapRegistryIdentity(
+            url: URL,
+            observabilityScope: ObservabilityScope,
+            callbackQueue: DispatchQueue,
+            completion: @escaping (Result<PackageIdentity?, Error>) -> Void
+        ) {
+            self.registryClient.lookupIdentities(url: url, observabilityScope: observabilityScope, callbackQueue: callbackQueue) { result in
+                // FIXME: returns first result... need to consider how to address multuiple ones
```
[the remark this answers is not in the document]
tomerd: yes, will address in follow up PR

Sources/Workspace/Workspace.swift
```diff
@@ -4185,6 +4194,290 @@ extension Workspace.Location {
...
+                }
+            }
+        }
+
+        private func transformManifest(
+            manifest: Manifest,
+            transformations: [PackageDependency: PackageIdentity],
+            transformationMode: TransformationMode,
+            observabilityScope: ObservabilityScope
+        ) throws -> Manifest {
```
tomerd, line 4321: this performs the swizzling and the heart of this feature

Sources/Workspace/Workspace.swift
```diff
@@ -4185,6 +4189,219 @@ extension Workspace.Location {
...
+        func purgeCache() throws {
+            try self.underlying.purgeCache()
+        }
+
+        // transforms source control dependencies to registry based ones when appropriate
+        // FIXME:
+        // 1. what to do with revision / branch based deps
+        // 2. add flag to turn this look up off?
+        // 3. what about a mixed situation when some versions on the registry but some on source control? do we need a second lookup that is version based?
+        // 4. check if registry is even configured, or not configured with a default
```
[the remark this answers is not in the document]
tomerd: yes

Sources/Workspace/Workspace.swift
tomerd: found a solution in which we only replace the identity instead of full swizzle. works well!

Sources/Commands/Options.swift
```diff
@@ -381,6 +365,26 @@ public struct SwiftToolOptions: ParsableArguments {
...
+        case swizzle
+
+        static func name(for value: Self) -> NameSpecification {
+            switch value {
+            case .disabled:
+                return .customLong("disable-scm-to-regsitry-lookup")
+            case .identity:
+                return .customLong("use-registry-identity-for-scm")
+            case .swizzle:
+                return .customLong("raplace-scm-with-registry")
```
abertelrud, line 383:
```suggestion
                return .customLong("replace-scm-with-registry")
```

abertelrud: Is there a way to add help strings for each of the modes?  The longer names are useful but the distinction between the last two seems a bit nuanced.

Sources/Commands/Options.swift
```diff
@@ -381,6 +365,26 @@ public struct SwiftToolOptions: ParsableArguments {
...
+
+    enum SourceControlToRegistryDependencyTransformation: EnumerableFlag {
+        case disabled
+        case identity
+        case swizzle
+
+        static func name(for value: Self) -> NameSpecification {
+            switch value {
+            case .disabled:
+                return .customLong("disable-scm-to-regsitry-lookup")
```
abertelrud, line 379:
```suggestion
                return .customLong("disable-scm-to-registry-lookup")
```

Sources/PackageModel/Manifest/PackageDependencyDescription.swift
```diff
@@ -12,19 +12,19 @@ import Foundation
 import TSCBasic
 
 /// Represents a package dependency.
-public enum PackageDependency: Equatable {
+public enum PackageDependency: Equatable, Hashable {
```
abertelrud, line 15: Hashable implies Equatable, but it makes sense to me as a style thing to list both.  I think I've been doing just Hashable, should we switch to consistently listing both in this code base going forward?  I kind of prefer it for clarity now that I see it.

Sources/SPMTestSupport/MockPackage.swift
```diff
@@ -73,6 +73,7 @@ public struct MockPackage {
         name: String,
         platforms: [PlatformDescription] = [],
         identity: String,
+        alternativeURLs: [String]? = .none,
```
abertelrud, line 76: Is there a semantic distinction between `[]` and `.none` here?  i.e. should it be maybe be non-optional?

Sources/SPMTestSupport/MockRegistry.swift
```diff
@@ -94,13 +107,21 @@ public class MockRegistry {
     private func handleRequest(request: HTTPClient.Request) throws -> HTTPClient.Response {
         let routeComponents = request.url.absoluteString.dropFirst(Self.mockRegistryURL.absoluteString.count + 1).split(separator: "/")
         switch routeComponents.count {
+        case _ where routeComponents[0].hasPrefix("identifiers?url="):
+            guard let query = request.url.query else {
+                throw StringError("invalid url: \(request.url)")
+            }
+            guard let sourceControlURL = URL(string: String(query.dropFirst(4))) else {
```
abertelrud, line 114: Are these four characters always the `url=` mentioned above, and therefore no additional check is needed for what they are?

Sources/SPMTestSupport/MockRegistry.swift
```diff
@@ -114,21 +135,33 @@ public class MockRegistry {
...
             }
         )
 
+        /*
+         <https://github.com/mona/LinkedList>; rel="canonical",
+         <ssh://git@github.com:mona/LinkedList.git>; rel="alternate"
+         */
+        let sourceControlURLs = self.packagesSourceControlURLs[packageIdentity]
+        let links = sourceControlURLs?.map { url in
+            "<\(url.absoluteString)>; rel=alternate"
```
abertelrud, line 156: Does this `absoluteString` result need escaping if it contains characters that have special meanings?

Sources/Workspace/Workspace.swift
```diff
@@ -4185,6 +4194,290 @@ extension Workspace.Location {
...
+        ) throws -> Manifest {
+            var targetDependencyPackageNameTransformations = [String: String]()
+
+            var modifiedDependencies = [PackageDependency]()
+            for dependency in manifest.dependencies {
+                var modifiedDependency = dependency
+                if let registryIdentity = transformations[dependency] {
+                    guard case .sourceControl(let settings) = dependency, case .remote = settings.location else {
+                        // an implementation mistake
+                        throw InternalError("unexpected non-source-control dependency")
```
abertelrud, line 4330: Should the error contain the dependencies to make it easier to debug if this is every thrown?

abertelrud (approved):

Sources/Commands/Options.swift
tomerd: good idea. will do

Sources/PackageModel/Manifest/PackageDependencyDescription.swift
tomerd: its a style thing. we have a mixed bag in the code base today. probably topic for a cleanup PR

Sources/SPMTestSupport/MockPackage.swift
tomerd: its a style thing, I guess (not sure about memory implication, empty array maybe also memory efficient). I personally prefer to use optionals when the data is semantically optional

Sources/SPMTestSupport/MockRegistry.swift
tomerd: this is in the testing infrastructure so should be safe enough

Sources/SPMTestSupport/MockRegistry.swift
tomerd: this is in the testing infrastructure so should be safe enough

tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

PR #4149, opened 2022-02-17
Fix module aliasing integration test fixture

elsh:
Resolves rdar://88722540

### DIFF
--- a/Fixtures/Miscellaneous/ModuleAliasing/DirectDeps/AppPkg/Package.swift
+++ b/Fixtures/Miscellaneous/ModuleAliasing/DirectDeps/AppPkg/Package.swift
@@ -10,14 +10,14 @@ let package = Package(
         .watchOS(.v5)
     ],
     dependencies: [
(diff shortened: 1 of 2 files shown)


elsh:
@swift-ci smoke test

neonichu (approved):

Fixtures/Miscellaneous/ModuleAliasing/DirectDeps/AppPkg/Package.swift
```diff
@@ -10,14 +10,14 @@ let package = Package(
...
-        .package(url: "../GamePkg", from: "1.0.0"),
+        .package(path: "../GamePkg"),
     ],
     targets: [
         .executableTarget(
             name: "App",
             dependencies: [
                 "Utils",
-                .product(name: "Utils",
+                .product(name: "UtilsProd",
```
tomerd, line 20: @elsh its very common that folks use the same name for the target and the product. is this causing issues?

PR #4154, opened 2022-02-20
improve output stream handling in CLI

tomerd:
motivation: better handling of what neds to go to stdout vs stderr

changes:
* update all CLI commands to use stdout for the final result. this means "swift build" and "swift test" output goes to stdout
* update all other background activities to output to stderr, this includes output from build operations and pacakge resolution that take place as part of another action
* set SwiftTool default output stream to sdterr instead of stdout
* update "swift test" to emit its output to stdout
* refactor all instance creation sites of BuildOperation to use a standard factory method which uses the appropriate output stream (normally stderr), whith a customization point for "swift build" to direct the output to stdout
* add and adjust tests

### DIFF
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -88,6 +88,10 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
         fileSystem: TSCBasic.FileSystem,
         observabilityScope: ObservabilityScope
     ) {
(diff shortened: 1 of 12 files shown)


tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

abertelrud (approved):

tomerd:
@swift-ci please smoke test

neonichu (approved):

PR #4155, opened 2022-02-21
Refactor and clean up plugin communication and wire messages

abertelrud:
Motivation:

Refactor the plugin machinery for better maintainability, consolidation of wire protocol struct definitions, and groundwork for extending the plugin functionality (such as adding new capabilities or new plugin messages such as dynamic querying of capabilities).

Changes:

- factor out the Codable declarations of enums and struct for the plugin wire protocol into a PluginMessage file that can be shared between the plugin and plugin host; this gets us compile-time checking so that the host and plugin are always in sync
- separate out the functionality of DefaultPluginScriptRunner to only handle the mechanics to talking to the plugin and not be concerned with what the messages are (this is factored back down into PluginInvocation)
- hoist the different kinds of plugin action messages up to top-level messages, allowing each to have a separate set of parameters
- factor the plugin input context serializer and deserializer out into their own source files and better separating from the semantics
- better separation of the structures for the plugin request structures (for building, testing, etc) to not reference any part of the plugin API; this will allow separate representations when/if necessary
- pull the package parameter into the plugin action, to support actions that don't operate on packages (such as querying the plugin for information, etc)

No functional changes.

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -1064,8 +1064,7 @@ extension SwiftPackageTool {
             // Run the command plugin.
             let buildEnvironment = try swiftTool.buildParameters().buildEnvironment
             let _ = try tsc_await { plugin.invoke(
-                action: .performCommand(arguments: arguments),
-                package: package,
+                action: .performCommand(package: package, arguments: arguments),
                 buildEnvironment: buildEnvironment,
                 scriptRunner: pluginScriptRunner,
                 workingDirectory: swiftTool.originalWorkingDirectory,
--- a/Sources/PackagePlugin/PackageManagerProxy.swift
+++ b/Sources/PackagePlugin/PackageManagerProxy.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See http://swift.org/LICENSE.txt for license information
@@ -29,14 +29,14 @@ public struct PackageManager {
     ) throws -> BuildResult {
         // Ask the plugin host to build the specified products and targets, and wait for a response.
         // FIXME: We'll want to make this asynchronous when there is back deployment support for it.
-        return try sendMessageAndWaitForReply(.buildOperationRequest(subset: subset, parameters: parameters)) {
+        return try sendMessageAndWaitForReply(.buildOperationRequest(subset: .init(subset), parameters: .init(parameters))) {
             guard case .buildOperationResponse(let result) = $0 else { return nil }
-            return result
+            return .init(result)
         }
     }
     
     /// Specifies a subset of products and targets of a package to build.
-    public enum BuildSubset: Encodable {
+    public enum BuildSubset {
         /// Represents the subset consisting of all products and of either all
         /// targets or (if `includingTests` is false) just non-test targets.
         case all(includingTests: Bool)
@@ -49,7 +49,7 @@ public struct PackageManager {
     }
     
     /// Parameters and options to apply during a build.
-    public struct BuildParameters: Encodable {
+    public struct BuildParameters {
         /// Whether to build for debug or release.
         public var configuration: BuildConfiguration
         
@@ -76,17 +76,17 @@ public struct PackageManager {
     
     /// Represents an overall purpose of the build, which affects such things
     /// asoptimization and generation of debug symbols.
-    public enum BuildConfiguration: String, Encodable {
+    public enum BuildConfiguration: String {
         case debug, release
     }
     
     /// Represents the amount of detail in a build log.
-    public enum BuildLogVerbosity: String, Encodable {
+    public enum BuildLogVerbosity: String {
         case concise, verbose, debug
     }
     
     /// Represents the results of running a build.
-    public struct BuildResult: Decodable {
+    public struct BuildResult {
         /// Whether the build succeeded or failed.
         public var succeeded: Bool
         
@@ -98,7 +98,7 @@ public struct PackageManager {
         public var builtArtifacts: [BuiltArtifact]
         
         /// Represents a single artifact produced during a build.
-        public struct BuiltArtifact: Decodable {
+        public struct BuiltArtifact {
             /// Full path of the built artifact in the local file system.
             public var path: Path
             
@@ -108,7 +108,7 @@ public struct PackageManager {
             /// Represents the kind of artifact that was built. The specific file
             /// formats may vary from platform to platform — for example, on macOS
             /// a dynamic library may in fact be built as a framework.
-            public enum Kind: String, Decodable {
+            public enum Kind: String {
                 case executable, dynamicLibrary, staticLibrary
             }
         }
@@ -129,14 +129,14 @@ public struct PackageManager {
     ) throws -> TestResult {
         // Ask the plugin host to run the specified tests, and wait for a response.
         // FIXME: We'll want to make this asynchronous when there is back deployment support for it.
-        return try sendMessageAndWaitForReply(.testOperationRequest(subset: subset, parameters: parameters)) {
+        return try sendMessageAndWaitForReply(.testOperationRequest(subset: .init(subset), parameters: .init(parameters))) {
             guard case .testOperationResponse(let result) = $0 else { return nil }
-            return result
+            return .init(result)
         }
     }
         
     /// Specifies what tests in a package to run.
-    public enum TestSubset: Encodable {
+    public enum TestSubset {
         /// Represents all tests in the package.
         case all
 
@@ -147,7 +147,7 @@ public struct PackageManager {
     }
     
     /// Parameters that control how the tests are run.
-    public struct TestParameters: Encodable {
+    public struct TestParameters {
         /// Whether to collect code coverage information while running the tests.
         public var enableCodeCoverage: Bool
         
@@ -157,7 +157,7 @@ public struct PackageManager {
     }
     
     /// Represents the result of running unit tests.
-    public struct TestResult: Decodable {
+    public struct TestResult {
         /// Whether the test run succeeded or failed.
         public var succeeded: Bool
         
@@ -171,24 +171,24 @@ public struct PackageManager {
 
         /// Represents the results of running some or all of the tests in a
         /// single test target.
-        public struct TestTarget: Decodable {
+        public struct TestTarget {
             public var name: String
             public var testCases: [TestCase]
             
             /// Represents the results of running some or all of the tests in
             /// a single test case.
-            public struct TestCase: Decodable {
+            public struct TestCase {
                 public var name: String
                 public var tests: [Test]
 
                 /// Represents the results of running a single test.
-                public struct Test: Decodable {
+                public struct Test {
                     public var name: String
                     public var result: Result
                     public var duration: Double
 
                     /// Represents the result of running a single test.
-                    public enum Result: String, Decodable {
+                    public enum Result: String {
                         case succeeded, skipped, failed
                     }
                 }
@@ -206,19 +206,19 @@ public struct PackageManager {
     ) throws -> SymbolGraphResult {
         // Ask the plugin host for symbol graph information for the target, and wait for a response.
         // FIXME: We'll want to make this asynchronous when there is back deployment support for it.
-        return try sendMessageAndWaitForReply(.symbolGraphRequest(targetName: target.name, options: options)) {
+        return try sendMessageAndWaitForReply(.symbolGraphRequest(targetName: target.name, options: .init(options))) {
             guard case .symbolGraphResponse(let result) = $0 else { return nil }
-            return result
+            return .init(result)
         }
     }
 
     /// Represents options for symbol graph generation.
-    public struct SymbolGraphOptions: Encodable {
+    public struct SymbolGraphOptions {
         /// The symbol graph will include symbols at this access level and higher.
         public var minimumAccessLevel: AccessLevel
 
         /// Represents a Swift access level.
-        public enum AccessLevel: String, CaseIterable, Encodable {
+        public enum AccessLevel: String, CaseIterable {
             case `private`, `fileprivate`, `internal`, `public`, `open`
         }
 
@@ -236,13 +236,15 @@ public struct PackageManager {
     }
 
     /// Represents the result of symbol graph generation.
-    public struct SymbolGraphResult: Decodable {
+    public struct SymbolGraphResult {
         /// The directory that contains the symbol graph files for the target.
         public var directoryPath: Path
     }
-    
+}
+
+fileprivate extension PackageManager {
     /// Private helper function that sends a message to the host and waits for a reply. The reply handler should return nil for any reply message it doesn't recognize.
-    fileprivate func sendMessageAndWaitForReply<T>(_ message: PluginToHostMessage, replyHandler: (HostToPluginMessage) -> T?) throws -> T {
+    func sendMessageAndWaitForReply<T>(_ message: PluginToHostMessage, replyHandler: (HostToPluginMessage) -> T?) throws -> T {
         try pluginHostConnection.sendMessage(message)
         guard let reply = try pluginHostConnection.waitForNextMessage() else {
             throw PackageManagerProxyError.unspecified("internal error: unexpected lack of response message")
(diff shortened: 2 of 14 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -1061,8 +1061,7 @@ extension SwiftPackageTool {
             // Run the command plugin.
             let buildEnvironment = try swiftTool.buildParameters().buildEnvironment
             let _ = try tsc_await { plugin.invoke(
-                action: .performCommand(arguments: arguments),
-                package: package,
+                action: .performCommand(package: package, arguments: arguments),
```
abertelrud, line 1064: The package is pulled into the action to pave the way for package-less actions, e.g. for SwiftPM to query the plugin in some way (at some point to do something similar to SwiftArgParser and glean the parameters from the declaration of the plugin, though we're far from that yet — it's an illustration of the kind of thing that requires this factoring).

Sources/PackagePlugin/PackageManagerProxy.swift
```diff
@@ -29,14 +29,14 @@ public struct PackageManager {
...
-        return try sendMessageAndWaitForReply(.buildOperationRequest(subset: subset, parameters: parameters)) {
+        return try sendMessageAndWaitForReply(.buildOperationRequest(subset: .init(subset), parameters: .init(parameters))) {
             guard case .buildOperationResponse(let result) = $0 else { return nil }
-            return result
+            return .init(result)
         }
     }
     
     /// Specifies a subset of products and targets of a package to build.
-    public enum BuildSubset: Encodable {
```
abertelrud, line 39: These structs, which are part of the API, were all directly serialized to JSON.  The evolution proposal doesn't state that they should be Codable, and this was a case of implementation leaking through.  Now the serialized structure is separate from the API.

Sources/PackagePlugin/PackageManagerProxy.swift
```diff
@@ -264,3 +266,170 @@ public enum PackageManagerProxyError: Error {
     /// An unspecified other kind of error from the Package Manager proxy.
     case unspecified(_ message: String)
 }
+
```
abertelrud, line 269: This section consists of code that bridges between the API structs and the serialized structs, to decouple the messaging implementation from the API.

Sources/PackagePlugin/Plugin.swift
```diff
@@ -104,27 +104,36 @@ extension Plugin {
         }
     }
     
+    /// Handles a single message received from the plugin host.
     fileprivate static func handleMessage(_ message: HostToPluginMessage) async throws {
         switch message {
-
```
abertelrud, line 109: The section that starts here is one of the key changes, and moves the plugin action that was nested inside `performAction` to instead be up at the top level.  The simplifies the structure and has the plugin message contain the input structure from which the package context is formed.

Sources/PackagePlugin/Plugin.swift
```diff
@@ -133,65 +142,89 @@ extension Plugin {
...
-                // Check that the plugin implements the appropriate protocol
-                // for its declared capability.
-                guard let plugin = plugin as? BuildToolPlugin else {
-                    throw PluginDeserializationError.malformedInputJSON(
-                        "Plugin declared with `buildTool` capability but doesn't conform to `BuildToolPlugin` protocol")
+            // Invoke the plugin to create build commands for the target.
+            let generatedCommands = try await plugin.createBuildCommands(context: context, target: target)
+            
+            // Send each of the generated commands to the host.
+            for command in generatedCommands {
```
abertelrud, line 157: This section is mostly indentation changes due to the removal of nesting.

Sources/PackagePlugin/Plugin.swift
```diff
@@ -213,60 +246,9 @@ extension Plugin {
     }
 }
 
-/// Message channel for communicating with the plugin host.
+/// Message channel for bidirectional communication with the plugin host.
 internal fileprivate(set) var pluginHostConnection: PluginHostConnection!
 
-/// A message that the host can send to the plugin.
-enum HostToPluginMessage: Decodable {
```
abertelrud, line 220: These are moved out into `PackageMessages`, which also contains the Codable structures that are used for all communication between the plugin and the host.

Sources/PackagePlugin/PluginContextDeserializer.swift
```diff
@@ -278,190 +248,6 @@ fileprivate struct PluginInputDeserializer {
     }
 }
 
-/// The input structure received as JSON from SwiftPM, consisting of an array
```
abertelrud, line 281: This is moved out into PluginMessages, and is now shared between the plugin side and the host side.

Sources/PackagePlugin/PluginMessages.swift
```diff
@@ -0,0 +1,316 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+ */
+
```
abertelrud, line 10: This pulls together all the host-to-plugin messages and plugin-to-host messages, and the structures they referenced.  This source file is shared between the PackagePlugin module and SwiftPM code so that there is only one source of truth about the Codable protocol and we get compiler detection of mismatches.

Sources/SPMBuildCore/PluginContextSerializer.swift
```diff
@@ -0,0 +1,276 @@
...
+import PackageGraph
+import PackageLoading
+import PackageModel
+import TSCBasic
+
+typealias WireInput = HostToPluginMessage.InputContext
+
+/// Creates the serialized input structure for the plugin script based on all
+/// the input information to a plugin.
+internal struct PluginContextSerializer {
```
abertelrud, line 22: This isn't new code but is pulled out from the DefaultPluginScriptRunner, which now deals only with the mechanics of process launching and message sending.

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -64,45 +61,220 @@ extension PluginTarget {
...
+                    arguments: arguments)
+            }
+            initialMessage = try actionMessage.toData()
         }
         catch {
             return callbackQueue.async { completion(.failure(PluginEvaluationError.couldNotSerializePluginInput(underlyingError: error))) }
         }
+        
+        // Handle messages and output from the plugin.
+        class ScriptRunnerDelegate: PluginScriptRunnerDelegate {
```
abertelrud, line 122: This code is hoisted up from the DefaultPluginScriptRunner, which now deals only with the mechanics of invoking the subprocess and sending messages, and not with the format of the messages.  So there is now a PluginScriptRunnerDelegate which handles callbacks about output and messages received, and that in turn calls back to the PluginInvocationDelegate, which is what the caller sees and has callbacks specific to plugin actions.

This way the PluginScriptRunner is restored to what it originally was, which is just a way of optionally compiling and then calling a subprocess and sending messages to it on a particular platform.

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -64,45 +61,220 @@ extension PluginTarget {
...
+                }
+            }
+        }
+        let runnerDelegate = ScriptRunnerDelegate(callbackQueue: callbackQueue, delegate: delegate)
+        
         // Call the plugin script runner to actually invoke the plugin.
         scriptRunner.runPluginScript(
             sources: sources,
-            input: inputStruct,
+            initialMessage: initialMessage,
```
abertelrud, line 233: Since the PluginScriptRunner no longer does the serialization, it is now passed the initial message to pass to the plugin (any message) and not necessarily one that contains a serialized package, etc. (there might not be one, depending on the kind of message it is).

Subsequent messages are then passed back to the delegate.  In the future the intent is to expand this to allow any number of messages to be sent at any time — for now having a single initial message and then letting others be follow-ons in response to requests keeps things simple and is all that's needed.

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -506,35 +587,35 @@ public protocol PluginInvocationDelegate {
     func pluginRequestedSymbolGraph(forTarget name: String, options: PluginInvocationSymbolGraphOptions, completion: @escaping (Result<PluginInvocationSymbolGraphResult, Error>) -> Void)
 }
 
-public struct PluginInvocationSymbolGraphOptions: Decodable {
```
abertelrud, line 509: As with the PackagePlugin API, the PluginMessages file now contains all the structs that get serialized over the wire, and so these structures are only client-facing now and there is no coupling to the wire protocol.

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -631,525 +712,170 @@ public extension PluginInvocationDelegate {
     }
 }
```
abertelrud, line 714: The diffs in the following section aren't easy to follow, but essentially the redeclaration of the serializable struct is removed (since it's now shared with the PackagePlugin module via PluginMessages.swift) but is instead replaced with extensions that instantiate the client-facing structure from the Codable structure.

Sources/SPMBuildCore/PluginScriptRunner.swift
```diff
@@ -0,0 +1,115 @@
...
+import Basics
+import Foundation
+import PackageModel
+import PackageLoading
+import PackageGraph
+import TSCBasic
+import struct TSCUtility.Triple
+
+/// Implements the mechanics of running and communicating with a plugin (implemented as a set of Swift source files). In most environments this is done by compiling the code to an executable, invoking it as a sandboxed subprocess, and communicating with it using pipes. Specific implementations are free to implement things differently, however.
+public protocol PluginScriptRunner {
```
abertelrud, line 20: This factored out from PluginInvocation.swift and deals just with the contract between the plugin invocation and the entity that can compile, run, and talk to the plugin on any given platform (of which DefaultPluginScriptRunner is the stock implementation).

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -535,16 +479,8 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
                             command: command,
                             output: String(decoding: stderrData, as: UTF8.self))
                     }
-                    // Otherwise we return a result based on its exit code. If
```
abertelrud, line 538: This code is now hoisted up to PluginInvocation, which deals with the semantics, leaving the plugin script runner to only deal with the mechanics of communicating with the plugin.

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -630,93 +566,35 @@ public enum DefaultPluginScriptRunnerError: Error, CustomStringConvertible {
     }
 }
 
-/// A message that the host can send to the plugin.
-enum HostToPluginMessage: Encodable {
```
abertelrud, line 634: This is moved out into the shared PluginMessages file.

Sources/PackagePlugin/PackageManagerProxy.swift
tomerd: that is a great change, the model and API should be separate from the serialization format as it buys us an important degree of freedom (even if at the cost of some healthy duplication)

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -64,45 +63,212 @@ extension PluginTarget {
...
+            
+            initialMessage = try actionMessage.toData()
         }
         catch {
             return callbackQueue.async { completion(.failure(PluginEvaluationError.couldNotSerializePluginInput(underlyingError: error))) }
         }
+        
+        // Handle messages and output from the plugin.
+        class ScriptRunnerDelegate: PluginScriptRunnerDelegate {
+            let callbackQueue: DispatchQueue
```
tomerd, line 117: is this the delegate queue or for some other callback?

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -64,45 +63,212 @@ extension PluginTarget {
...
+                        do {
+                            switch $0 {
+                            case .success(let result):
+                                responder(try HostToPluginMessage.testOperationResponse(result: .init(result)).toData())
+                            case .failure(let error):
+                                responder(try HostToPluginMessage.errorResponse(error: String(describing: error)).toData())
+                            }
+                        }
+                        catch {
+                            print("couldn't send reply to plugin \(error)")
```
tomerd, line 200: should this use an observability scope to email instead of print?

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -64,45 +63,212 @@ extension PluginTarget {
...
             writableDirectories: writableDirectories,
             readOnlyDirectories: readOnlyDirectories,
             fileSystem: fileSystem,
             observabilityScope: observabilityScope,
             callbackQueue: callbackQueue,
-            delegate: delegate,
-            completion: completion)
+            delegate: runnerDelegate) { result in
+                dispatchPrecondition(condition: .onQueue(callbackQueue))
+                switch result {
```
tomerd, line 237: result.map or result.tryMap can make this more concise

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -64,45 +63,212 @@ extension PluginTarget {
...
+                case .failure(let error):
+                    completion(.failure(error))
+                }
+        }
+    }
+}
+
+fileprivate extension HostToPluginMessage {
+    func toData() throws -> Data {
+        let encoder = JSONEncoder()
```
tomerd, line 258: makeWithDefaults?

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -64,45 +63,212 @@ extension PluginTarget {
...
+        encoder.outputFormatting = [.prettyPrinted, .sortedKeys, .withoutEscapingSlashes]
+        return try encoder.encode(self)
     }
 }
 
+fileprivate extension PluginToHostMessage {
+    init(_ data: Data) throws {
+        let decoder = JSONDecoder()
```
tomerd, line 266: makeWithDefaults?

Sources/SPMBuildCore/PluginScriptRunner.swift
```diff
@@ -0,0 +1,115 @@
...
+        initialMessage: Data,
+        toolsVersion: ToolsVersion,
+        workingDirectory: AbsolutePath,
+        writableDirectories: [AbsolutePath],
+        readOnlyDirectories: [AbsolutePath],
+        fileSystem: FileSystem,
+        observabilityScope: ObservabilityScope,
+        callbackQueue: DispatchQueue,
+        delegate: PluginScriptRunnerDelegate,
+        completion: @escaping (Result<Int32, Error>) -> Void
```
tomerd, line 50: do we need two queue here? one for the completion handler and one for the delegate?

Tests/SPMBuildCoreTests/PluginInvocationTests.swift
```diff
@@ -149,10 +150,10 @@ class PluginInvocationTests: XCTestCase {
                         inputFiles: [],
                         outputFiles: [])
                 }
-                
+            */
```
tomerd, line 153: left overs?

tomerd (approved):
quite a big PR so not sure that I was able to follow all the nuances. the approach seems good! and of course we can address any additional adjustments in follow up PRs

Sources/PackagePlugin/PackageManagerProxy.swift
abertelrud: I thought so too.  I'm wondering if there's a way to automate some of that duplication, e.g. with .gyb or something.

Sources/SPMBuildCore/PluginInvocation.swift
abertelrud: Oh right.  Yes, I meant to fix this before marking ready.  Thanks for calling it out!

Sources/SPMBuildCore/PluginInvocation.swift
abertelrud: Ah right, I keep forgetting about that one.  :)  Thanks for pointing it out.

Sources/SPMBuildCore/PluginScriptRunner.swift
abertelrud: I'll need to double-check — I just moved this around, it doesn't change functionally from before.

Tests/SPMBuildCoreTests/PluginInvocationTests.swift
abertelrud: Hmm, I thought I pushed changes to fix this.  I'll double-check.

abertelrud:
> quite a big PR so not sure that I was able to follow all the nuances. the approach seems good! and of course we can address any additional adjustments in follow up PRs

Great, thanks for taking the time.  Sorry about the big diff — I kind of got carried away in restructuring but should probably have gone back to split things out.  I'll be a bit more mindful of how reviewable it is next time.

Mostly it's just the boilerplate of moving things around and introducing the extra wrappers.  But I appreciate the places you caught where I need to go back and see that I pushed all the last cleanup diffs.  Especially the commented code is something I know I fixed, so I want to make sure those diffs didn't go missing.

Thanks, and I'll update this in tomorrow to address the feedback!

Tests/SPMBuildCoreTests/PluginInvocationTests.swift
abertelrud: Yes, I definitely fixed this to return actual messages for round-trip testing.  I'll push the latest changes (it wasn't a ton of changes so doesn't invalidate any review, just some last cleanup).

Sources/SPMBuildCore/PluginInvocation.swift
abertelrud: It's used for both the delegate and the completion block, and it was the name used for this before this PR as well.  Possibly it should be renamed, but `delegateQueue` might suggest it is only used for the delegate.

So a typical client would pass in a serial queue and thus get a sequence of zero or more non-overlapping delegate calls and then finally the completion handler on the same queue.

I've always been a bit on the fence about whether the completion block should be a `didFinish` kind of delegate call, but it's convenient to have it separate if the delegate is optional and there are cases where the completion handler is needed but a delegate isn't.

Sources/SPMBuildCore/PluginInvocation.swift
abertelrud: Actually, I think what should happen here is that the default script handler should probably hold onto the error and go into an error state.  I'll have to look what should happen (the previous code dropped the error, so potentially we could also improve this separately).

Sources/SPMBuildCore/PluginInvocation.swift
abertelrud: BTW, the only reason the delegate has the queue as a property here is for the dispatch queue precondition check.  Perhaps we should drop that and only do that in unit tests.

neonichu (approved):

Sources/SPMBuildCore/PluginInvocation.swift
abertelrud: Removed this delegate actually, and also added comments for the purpose of the properties.

Sources/SPMBuildCore/PluginInvocation.swift
abertelrud: I keep forgetting about that one.  It really does make things a lot cleaner; thanks!

Sources/SPMBuildCore/PluginScriptRunner.swift
abertelrud: Actually, it's the same queue.  It's used for both the delegate and the completion block, and it was the name used for this before this PR as well. Possibly it should be renamed, but a name like `delegateQueue` might suggest it is only used for the delegate.

So a typical client would pass in a serial queue and thus get a sequence of zero or more non-overlapping delegate calls and then finally the completion handler on the same queue.

So there should definitely be only one queue (IMO) but the question is just what the right name is when there is both a delegate and a completion handler.  Callback seemed to be a good generic term for that.

abertelrud:
I've address the conflict and the feedback.  Thanks both!

abertelrud:
The Linux test failure is:
```
+ tar xzf /tmp/jenkins-07572e16/workspace/swift-package-manager-self-hosted-Linux-smoke-test/branch-main/latest_toolchain.tar.gz -C /tmp/jenkins-07572e16/workspace/swift-package-manager-self-hosted-Linux-smoke-test/branch-main/latest_toolchain

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```
which indicates a problem downloading the toolchain.  As the failing one isn't required and is unrelated to these changes, I will merge.

PR #4159, opened 2022-02-22
fix a couple of data races in plugin tests

tomerd:
motivation: address issues reported by tsan

changes: use thread safe data stores in plugin test delegates

### DIFF
--- a/Sources/Basics/ConcurrencyHelpers.swift
+++ b/Sources/Basics/ConcurrencyHelpers.swift
@@ -182,6 +182,20 @@ public final class ThreadSafeBox<Value> {
         return value
     }
 
+    @discardableResult
+    public func memoize(body: () throws -> Value?) rethrows -> Value? {
+        if let value = self.get() {
+            return value
+        }
+        if let value = try body() {
+            self.lock.withLock {
+                self.underlying = value
+            }
+            return value
+        }
+        return nil
+    }
+
     public func clear() {
         self.lock.withLock {
             self.underlying = nil
--- a/Tests/FunctionalTests/PluginTests.swift
+++ b/Tests/FunctionalTests/PluginTests.swift
@@ -364,7 +364,7 @@ class PluginTests: XCTestCase {
             let delegateQueue = DispatchQueue(label: "plugin-invocation")
             class PluginDelegate: PluginInvocationDelegate {
                 let delegateQueue: DispatchQueue
-                var diagnostics: [Basics.Diagnostic] = []
+                var diagnostics = ThreadSafeArrayStore<Basics.Diagnostic>()
 
                 init(delegateQueue: DispatchQueue) {
                     self.delegateQueue = delegateQueue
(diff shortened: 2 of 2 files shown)


tomerd:
@swift-ci please smoke test

tomerd:
this was in reaction to a data race reported by tsan

abertelrud:
Right, understood, I need to see exactly where and why.

abertelrud:
Hm, not able to repro a tsan violation here.  On what platform are you seeing that?

abertelrud:
Running repeated tests to see if it's timing-based.

tomerd:
`swift test --parallel --sanitize=thread --filter FunctionalTests`

abertelrud:
Thanks!

abertelrud:
I can repro this from the command line, but not Xcode.  But I think figured out why — enabling tsan in the scheme in Xcode only affected running but not testing — it has to also be enabled in the Test tab to take effect during testing.  So the tests were actually not running with testing.

So that all said, I think that the right fix here is that the `processIdentifier` of the delegate should actually be accessed from the delegate queue where it is set.  Making this property thread-safe should not be necessary and would in fact hide any other tsan issues.

The good news is that there does not seem to be any underlying bug here.

So this fixes it for me.  WDYT?

```
diff --git a/Tests/FunctionalTests/PluginTests.swift b/Tests/FunctionalTests/PluginTests.swift
index 6ca7b417..f8d8180b 100644
--- a/Tests/FunctionalTests/PluginTests.swift
+++ b/Tests/FunctionalTests/PluginTests.swift
@@ -636,7 +636,11 @@ class PluginTests: XCTestCase {
             XCTAssertEqual(result, .timedOut, "expected the plugin to time out")
             
             // At this point we should have parsed out the process identifier. But it's possible we don't always — this is being investigated in rdar://88792829.
-            guard let pid = delegate.parsedProcessIdentifier else {
+            var pid: Int? = .none
+            delegateQueue.sync {
+                pid = delegate.parsedProcessIdentifier
+            }
+            guard let pid = pid else {
                 throw XCTSkip("skipping test because no pid was received from the plugin; being investigated as rdar://88792829\n\(delegate.diagnostics.description)")
             }

tomerd:
if a delegate queue if provided by the call site, then it should be fine to rely on that. 

more generally, its not a pattern I am personally a fan of, but this will go away once we adopt async/await.

abertelrud:
> if a delegate queue if provided by the call site, then it should be fine to rely on that.

Right, the API contract here is that the caller provides a delegate and a queue on which the delegate should be invoked.  As with the other delegates in SwiftPM, the delegate queue passed by the caller would almost always be serial, so that things like `willFetch` and `didFetch` etc happen in order.

> more generally, its not a pattern I am personally a fan of, but this will go away once we adopt async/await.

Looking forward to seeing how we'll apply async/await to the delegates, and how to keep delegate calls in sequence without use of a callback queue.

PR #4160, opened 2022-02-22
Fix a bug that caused a plugin dependency on a local executable to not work properly

abertelrud:
Motivation:  This bug affects how packages that vend plugins that rely on built tools can be structured (remote tool dependencies work but local dependencies do not, without this fix).

Changes:
- pass `.product` instead of `.target` to make sure that the executable is linked and not just compiled
- add a unit test that checks remote and local tools, with both implicitly and explicitly defined products for the local case

rdar://89267060

### DIFF
--- a/Tests/FunctionalTests/PluginTests.swift
+++ b/Tests/FunctionalTests/PluginTests.swift
@@ -471,6 +471,22 @@ class PluginTests: XCTestCase {
         }
     }
 
+    func testLocalAndRemoteToolDependencies() throws {
+        // Only run the test if the environment in which we're running actually supports Swift concurrency (which the plugin APIs require).
+        try XCTSkipIf(!UserToolchain.default.supportsSwiftConcurrency(), "skipping because test environment doesn't support concurrency")
+
+        try fixture(name: "Miscellaneous/Plugins/PluginUsingLocalAndRemoteTool") { path in
+            let (stdout, stderr) = try executeSwiftPackage(path.appending(component: "MyLibrary"), configuration: .Debug, extraArgs: ["plugin", "my-plugin"])
+            XCTAssert(stderr.contains("Linking RemoteTool"), "stdout:\n\(stderr)\n\(stdout)")
+            XCTAssert(stderr.contains("Linking LocalTool"), "stdout:\n\(stderr)\n\(stdout)")
+            XCTAssert(stderr.contains("Linking ImpliedLocalTool"), "stdout:\n\(stderr)\n\(stdout)")
+            XCTAssert(stderr.contains("Build complete!"), "stdout:\n\(stderr)\n\(stdout)")
+            XCTAssert(stdout.contains("A message from the remote tool."), "stdout:\n\(stderr)\n\(stdout)")
+            XCTAssert(stdout.contains("A message from the local tool."), "stdout:\n\(stderr)\n\(stdout)")
+            XCTAssert(stdout.contains("A message from the implied local tool."), "stdout:\n\(stderr)\n\(stdout)")
+        }
+    }
+
     func testCommandPluginCancellation() throws {
         // Only run the test if the environment in which we're running actually supports Swift concurrency (which the plugin APIs require).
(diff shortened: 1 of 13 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
This will be nominated to 5.6 as soon as it passes testing.

tomerd (approved):

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -471,6 +471,28 @@ class PluginTests: XCTestCase {
         }
     }
 
+    func testLocalAndRemoteToolDependencies() throws {
+        // Only run the test if the environment in which we're running actually supports Swift concurrency (which the plugin APIs require).
+        try XCTSkipIf(!UserToolchain.default.supportsSwiftConcurrency(), "skipping because test environment doesn't support concurrency")
+
+        try fixture(name: "Miscellaneous/Plugins/PluginUsingLocalAndRemoteTool") { path in
```
tomerd, line 478: fixture already prints the error when caught so the catch here is redundant afaik

abertelrud: Great, thanks — this was based on an older test so there might need to be more cleanup needed here.  I'll fix this one new test for now at least.

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -471,6 +471,22 @@ class PluginTests: XCTestCase {
...
+            let (stdout, _) = try executeSwiftPackage(path.appending(component: "MyLibrary"), configuration: .Debug, extraArgs: ["plugin", "my-plugin"])
+            XCTAssert(stdout.contains("Linking RemoteTool"), "stdout:\n\(stdout)")
+            XCTAssert(stdout.contains("Linking LocalTool"), "stdout:\n\(stdout)")
+            XCTAssert(stdout.contains("Linking ImpliedLocalTool"), "stdout:\n\(stdout)")
+            XCTAssert(stdout.contains("Build complete!"), "stdout:\n\(stdout)")
+            XCTAssert(stdout.contains("A message from the remote tool."), "stdout:\n\(stdout)")
+            XCTAssert(stdout.contains("A message from the local tool."), "stdout:\n\(stdout)")
+            XCTAssert(stdout.contains("A message from the implied local tool."), "stdout:\n\(stdout)")
```
tomerd, line 486: note I just merged the std/stderr cleanup PR so this may be different now?

Tests/FunctionalTests/PluginTests.swift
abertelrud: Fixed it.

abertelrud:
@swift-ci please smoke test

neonichu (approved):

SDGGiesbrecht (approved):

Tests/FunctionalTests/PluginTests.swift
abertelrud: Yes, this will probably need to change.  Thanks for the heads-up.  Still not sure why this is failing on Linux — looks like the build log messages are not being emitted there — going to repo in Docker.

tomerd: the build output in such case is expected to go to stderr. this may also be why the linux build is failing

abertelrud: Thanks, that's helpful.  Odd that it doesn't seem to happen on macOS.  In any case, I'll rebase on main and see what the right fix here is (ideally that also works unmodified on 5.6).  Technically speaking the build log output doesn't matter to this test — it's that we get the output from the plugin that matters, so perhaps those asserts aren't needed here.

abertelrud:
@swift-ci please smoke test

abertelrud:
Hmm, now it's `WorkspaceTests.testResolutionMixedRegistryAndSourceControl()` that's failing on Linux.

abertelrud:
Looks like an issue common to various CI PRs:
```
Swift/StringStorage.swift:537: Fatal error
```

tomerd:

PR #4163, opened 2022-02-22
Fix a tsan issue in `PluginTests.testCommandPluginCancellation()`

abertelrud:
This fixes a reported thread sanitizer issue in the `PluginTests.testCommandPluginCancellation()` test.

**Description:** The delegate is called in sequence because the caller passes in a serial queue, but there is a race condition between the timeout of waiting for the subprocess to finish and accessing the delegate’s parsed-out process identifier.  Accessing it on the delegate queue will preserve the thread access integrity.

This is an alternate (and in my view better) fix of https://github.com/apple/swift-package-manager/pull/4159 because making the data structures thread-safe would mask further tsan issues.

**Motivation:** Fixes a reported tsan violation in unit tests.

**Changes:**
- access the parsed process identifier from the serial delegate callback queue


abertelrud:
@swift-ci please smoke test

tomerd (approved):

neonichu (approved):

PR #4164, opened 2022-02-22
[5.6] Fix a bug that caused a plugin dependency on a local executable to not work properly

abertelrud:
This is a SwiftPM 5.6 cherry-pick of https://github.com/apple/swift-package-manager/pull/4160.

**Explanation:** When a package command plugin has a dependency on an executable built from source (as opposed to a binary target) and that executable is in the same package as the plugin, the dependency ends up being on the executable target and not the product. This means that the module gets built but the final link step is never performed, so trying to use the executable from within the plugin fails.

**Scope of Issue:** This affects helper tools that are built from source and defined in the same package as the plugin that uses them.

**Reason for Nominating to 5.6:** This is a serious bug in new functionality (command plugins) that affects the user experience. It is also very safe, since the fix is in the new code that can only be executed when using this new functionality.

**Risk:** Very low — this affects only the code path taken when dealing with a target dependency of a plugin on an executable target in the same package.

**Origination:** This was a bug in the initial implementation of package command plugins.

**Reviewed By:** @tomerd

**Automated Testing:** A new unit test covers use of both implicit and explicitly defined executable products in the same package as the plugin, as well as the case of remote package products, which worked fine.

**Dependencies:** None

**Impact on CI:** None

**How to Verify:** Check out the SwiftPM source code, and use the test fixture at `Fixtures/Miscellaneous/Plugins/PluginUsingLocalAndRemoteTool` and change directory into its `MyLibrary` subdirectory.  Invoke it using `swift package my-plugin` and verify that it succeeds.

### DIFF
--- a/Fixtures/Miscellaneous/Plugins/PluginUsingLocalAndRemoteTool/MyLibrary/Package.swift
+++ b/Fixtures/Miscellaneous/Plugins/PluginUsingLocalAndRemoteTool/MyLibrary/Package.swift
@@ -1,0 +1,18 @@
+// swift-tools-version: 5.6
+import PackageDescription
+
+let package = Package(
+    name: "MyLibrary",
+    dependencies: [
+        .package(path: "../MyPlugin")
+    ],
+    targets: [
+        .target(
+            name: "MyLibrary"
+        ),
+        .testTarget(
+            name: "MyLibraryTests",
+            dependencies: ["MyLibrary"]
+        )
+    ]
+)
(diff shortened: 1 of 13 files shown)


abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please test

abertelrud:
@shahmishal  It doesn't look as if the tests are starting for this one.  Is that a known issue?  Note that this is targeting 5.6.

shahmishal:
@swift-ci please test

shahmishal:
@swift-ci smoke test

abertelrud:
Thanks @shahmishal!  Looks like maybe the linux self hosted test is unrelated to the code change in the PR?  But also, it's not one of the required ones, it seems.

abertelrud:
Merging after approval.

PR #4171, opened 2022-02-23
refactor target build settings model

tomerd:
motivation: as part of auditing the code for preconditions and assets it came to light that the target build setting model can be improved to handle preconditions via the type system

changes:
* move the build settings values into the build setting kind enum as associated values (istead of floating attribute)
* adjust call sites and test


tomerd:
@swift-ci please smoke test

tomerd:

PR #4173, opened 2022-02-24
cancellation handler

tomerd:
motivation: allow clients of SwiftPM to cancel background activities

changes:
* introduce new "cancellator" utility that is a registry for cancellation handlers
* use the "cancellator" instead of ProcessSet to manage the cancellation of active processes (eg git, tests) and build system
* add cancel method to GitRepositoryProvider protocol so that it can be interuppted
* add cancel methods to RepositoryManager, RegistryDownloadManager and HTTPClient so that it can be interrupted
* change workspace initializer to take a terminator so that CLI and other consumer of libSwiftPM can interrupt
* register interruption points mentioned above in workspace
* adjust related call sites, ie CLI signal handler now uses the workspace terminator to interrupt
* add tests

rdar://64900054
rdar://63723896

### DIFF
--- a/Sources/Basics/Archiver+Zip.swift
+++ b/Sources/Basics/Archiver+Zip.swift
@@ -12,11 +12,14 @@ import TSCBasic
 import Dispatch
 
 /// An `Archiver` that handles ZIP archives using the command-line `zip` and `unzip` tools.
-public struct ZipArchiver: Archiver {
+public struct ZipArchiver: Archiver, Cancellable {
     public var supportedExtensions: Set<String> { ["zip"] }
 
     /// The file-system implementation used for various file-system operations and checks.
     private let fileSystem: FileSystem
+
+    /// Helper for cancelling in-fligh requests
+    private let cancellator: Cancellator
 
     /// Creates a `ZipArchiver`.
     ///
@@ -24,6 +27,7 @@ public struct ZipArchiver: Archiver {
     ///   - fileSystem: The file-system to used by the `ZipArchiver`.
     public init(fileSystem: FileSystem) {
         self.fileSystem = fileSystem
+        self.cancellator = Cancellator(observabilityScope: .none)
     }
 
     public func extract(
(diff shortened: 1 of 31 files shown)


Sources/Basics/Terminator.swift
```diff
@@ -0,0 +1,112 @@
...
+            return "noop" // nothing was done
+        }
+        let key = UUID().uuidString
+        self.observabilityScope?.emit(debug: "registering '\(name)' with terminator")
+        self.registry[key] = (name: name, handler: handler)
+        return key
+    }
+
+    public func register(_ process: TSCBasic.Process) -> RegistrationKey {
+        self.register(name: "\(process.arguments.joined(separator: " "))", handler: {
```
abertelrud, line 51: Would it make sense to require the process to have been launched already and use a key that there can't be two at once of, like the pid?  We can probably make things like plugin command lines unique but there's no guarantee that they are (since plugins communicate with the host more interactively than this).

Either way it's a details, but the part that affects this is whether to require the process to already be running.

Sources/Basics/Terminator.swift
```diff
@@ -0,0 +1,112 @@
...
+}
+
+extension TSCBasic.Process {
+    fileprivate func terminate(timeout: DispatchTime) {
+        // send graceful shutdown signal
+        self.signal(SIGINT)
+
+        // start a thread to see if we need to terminate more forcibly
+        let forceKillSemaphore = DispatchSemaphore(value: 1)
+        let forceKillThread = TSCBasic.Thread {
```
abertelrud, line 96: Is the reason for using a thread and not a dispatch async that we can't know if a dispatch loop is running?

Sources/SPMTestSupport/GitRepositoryExtensions.swift
```diff
@@ -68,3 +72,13 @@ public extension GitRepository {
         try systemQuietly([Git.tool, "-C", self.path.pathString, "push", remote, branch])
     }
 }
+
+extension GitRepositoryProvider {
+    public init() {
+        self.init(terminator: .NOOP)
```
abertelrud, line 78: What was the reason for not having the terminator be an optional?

Sources/Commands/SwiftTool.swift
```diff
@@ -430,8 +424,7 @@ public class SwiftTool {
 
                 // Terminate all processes on receiving an interrupt signal.
                 DefaultPluginScriptRunner.cancelAllRunningPlugins()
```
abertelrud, line 426: Looking forward to replacing this line with the new API you're adding (in a follow-on PR)! :)

tomerd:
@swift-ci please smoke test

Sources/Basics/Cancellator.swift
```diff
@@ -0,0 +1,135 @@
...
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+ */
+
+import Dispatch
+import Foundation
+import TSCBasic
+
+public typealias CancellationHandler = (DispatchTime) throws -> Void
+
+public class Cancellator: Cancellable {
```
tomerd, line 17: Originally was named "terminator" which is cute but less accurate. Happy to take suggestions for a nicer name

tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

tomerd:
@abertelrud this is ready for review

@compnerd could you help make sure this builds / works on Windows

Sources/Commands/SwiftTool.swift
```diff
@@ -400,17 +398,13 @@ public class SwiftTool {
...
             #if os(Windows)
             // set shutdown handler to terminate sub-processes, etc
-            SwiftTool.shutdownRegistry = (processSet: processSet, buildSystemRef: buildSystemRef)
+            SwiftTool.cancellator = cancellator
             _ = SetConsoleCtrlHandler({ _ in
                 // Terminate all processes on receiving an interrupt signal.
                 DefaultPluginScriptRunner.cancelAllRunningPlugins()
-                SwiftTool.shutdownRegistry?.processSet.terminate()
-                SwiftTool.shutdownRegistry?.buildSystemRef.buildSystem?.cancel()
+                SwiftTool.cancellator?.cancel()
```
compnerd, line 407:
```suggestion
                try? SwiftTool.cancellator?.cancel(deadline: .now + seconds(30))
```

compnerd (commented):
Delta: ~30KiB
Basics.dll: 276,992 -> 297,984
Commands.dll: 5,531,648 -> 5,532,672
Workspace.dll: 1,356,800 -> 1,3564,992

I'll play around with it itself to see if there are any issues, though I suspect not.

tomerd:
@swift-ci please smoke test

Sources/Basics/Terminator.swift
tomerd: the name is not the unique key, its only for presentation

Sources/Basics/Terminator.swift
tomerd: since this is likely canceling threads that have been scheduled with Dispatch, it gives us a bit of a separation between the systems. dispatch does work to minimize how much threads it spawns so there is some risk the shutdown thread will not be spawned or prioritized if we use out global dispatch queue

Sources/Commands/SwiftTool.swift
tomerd: happy to try and do a follow up PR for that, need to learn more about how plugin cancellation could work

tomerd:
@swift-ci please smoke test

Sources/Basics/Terminator.swift
abertelrud: I see — good points, thanks for explaining.

Sources/Basics/Terminator.swift
abertelrud: Oh right — I think I saw that later down in the PR.

tomerd:
@swift-ci please smoke test

Sources/Basics/Archiver+Zip.swift
```diff
@@ -12,54 +12,99 @@ import TSCBasic
...
+                    guard processResult.exitStatus == .terminated(code: 0) else {
+                        throw try StringError(processResult.utf8stderrOutput())
+                    }
+                }))
+            }
+        } catch {
+            return completion(.failure(error))
         }
 
+        /*
```
abertelrud, line 66: left-over debugging?

tomerd: ugh yes

neonichu (approved):

compnerd:
This seemed to have dropped the Windows fixes and has broken the windows build: https://ci-external.swift.org/job/oss-swift-windows-toolchain-x86_64-vs2019/1130/console

PR #4176, opened 2022-02-25
build: explicitly select the librarian on Darwin

compnerd:
Newer CMake will attempt to use the LLVM toolchain if clang is detected.
This generates an archive that ld64 may not be able to properly process.
Explicitly specify the librarian to be `ar` so that the cctools archiver
is used.

### Motivation:

### Modifications:

### Result:


yim-lee:
rdar://87850031

neonichu (approved):

compnerd:
Turns out that we needed to alter `CMAKE_C_CREATE_ARCHIVE` which we were unable to pass from the command line :-(.  This is a workaround for an actual issue in CMake (which I've reported upstream).  Since we needed to modify the second parameter, it makes sense to just sink this into the repo :-(.

PR #4178, opened 2022-02-25
improve manfiest loader concurrency control

tomerd:
motivation: more correct and robust concurrency control for manifest loading

changes:
* use a seperate queue to schedule the evaluation tasks to ensure concurrency control is done correctly
* move concurrency control to the methods that forks the child processes so that its more targeted
* clarify what is the delegate queue and what is the callback queue
* move a handful of data structure defintions around to make code easier to read
* make concurrency tests more challanging

### DIFF
--- a/Sources/PackageLoading/ManifestLoader.swift
+++ b/Sources/PackageLoading/ManifestLoader.swift
@@ -51,7 +51,8 @@ public protocol ManifestLoaderProtocol {
     ///   - identityResolver: A helper to resolve identities based on configuration
     ///   - fileSystem: File system to load from.
     ///   - observabilityScope: Observability scope to emit diagnostics.
-    ///   - on: The dispatch queue to perform asynchronous operations on.
+    ///   - delegateQueue: The dispatch queue to call delegate handlers on.
+    ///   - callbackQueue: The dispatch queue to perform completion handler on.
     ///   - completion: The completion handler .
     func load(
         at path: AbsolutePath,
@@ -64,7 +65,8 @@ public protocol ManifestLoaderProtocol {
         identityResolver: IdentityResolver,
         fileSystem: FileSystem,
         observabilityScope: ObservabilityScope,
-        on queue: DispatchQueue,
+        delegateQueue: DispatchQueue,
+        callbackQueue: DispatchQueue,
         completion: @escaping (Result<Manifest, Error>) -> Void
     )
 
@@ -105,7 +107,10 @@ public final class ManifestLoader: ManifestLoaderProtocol {
 
     private let sdkRootCache = ThreadSafeBox<AbsolutePath>()
 
+    /// DispatchSemaphore to restrict concurrent manifest evaluations
     private let concurrencySemaphore: DispatchSemaphore
+    /// OperationQueue to park pending lookups
+    private let evaluationQueue: OperationQueue
 
     public init(
         toolchain: ToolchainConfiguration,
@@ -123,6 +128,10 @@ public final class ManifestLoader: ManifestLoaderProtocol {
 
         self.databaseCacheDir = cacheDir.map(resolveSymlinks)
 
+        // this queue and semaphore is used to limit the amount of concurrent manifest loading taking place
+        self.evaluationQueue = OperationQueue()
+        self.evaluationQueue.name = "org.swift.swiftpm.manifest-loader"
+        self.evaluationQueue.maxConcurrentOperationCount = Concurrency.maxOperations
         self.concurrencySemaphore = DispatchSemaphore(value: Concurrency.maxOperations)
     }
 
@@ -184,7 +193,8 @@ public final class ManifestLoader: ManifestLoaderProtocol {
                 identityResolver: identityResolver,
                 fileSystem: fileSystem,
                 observabilityScope: ObservabilitySystem(diagnosticEngine: diagnostics ?? DiagnosticsEngine()).topScope,
-                on: queue,
+                delegateQueue: queue,
+                callbackQueue: queue,
                 completion: completion
             )
         } catch {
@@ -203,7 +213,8 @@ public final class ManifestLoader: ManifestLoaderProtocol {
         identityResolver: IdentityResolver,
         fileSystem: FileSystem,
         observabilityScope: ObservabilityScope,
-        on queue: DispatchQueue,
+        delegateQueue: DispatchQueue,
+        callbackQueue: DispatchQueue,
         completion: @escaping (Result<Manifest, Error>) -> Void
     ) {
         do {
@@ -219,11 +230,14 @@ public final class ManifestLoader: ManifestLoaderProtocol {
                 identityResolver: identityResolver,
                 fileSystem: fileSystem,
                 observabilityScope: observabilityScope,
-                on: queue,
+                delegateQueue: delegateQueue,
+                callbackQueue: callbackQueue,
                 completion: completion
             )
         } catch {
-            return completion(.failure(error))
+            callbackQueue.async {
+                completion(.failure(error))
+            }
         }
     }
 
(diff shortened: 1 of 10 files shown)


tomerd:
@swift-ci please smoke test

Sources/PackageLoading/ManifestLoader.swift
```diff
@@ -898,108 +897,121 @@ public final class ManifestLoader: ManifestLoaderProtocol {
...
-                    cmd += ["-handle", "\(String(hFile, radix: 16))"]
-        #else
-                    cmd += ["-fileno", "\(fileno(jsonOutputFileDesc))"]
-        #endif
+        // we must not block the calling thread (for concurrency control) so nesting this in a queue
+        self.evaluationQueue.async {
+            do {
+                // park the evaluation thread based on the max concurrency allowed
+                self.concurrencySemaphore.wait()
+                // run the evaluation
```
tomerd, line 912: this diff is terrible but its basically nesting under the queue as noted in these 4 lines

tomerd:
@swift-ci please smoke test

tomerd:

tomerd:
@swift-ci please smoke test

Sources/PackageLoading/ManifestLoader.swift
```diff
@@ -1079,6 +1034,110 @@ public final class ManifestLoader: ManifestLoaderProtocol {
...
+
+        mutating func delay() -> DelayableAction {
+            let next = DelayableAction(target: target, action: action)
+            target = nil
+            action = nil
+            return next
+        }
+    }
+}
+
```
tomerd, line 1140: this data structs moved from line the code to the bottom to make code more readable. no function changes in them

Sources/Workspace/RegistryPackageContainer.swift
```diff
@@ -120,87 +120,89 @@ public class RegistryPackageContainer: PackageContainer {
...
-                        let manifests = result.manifests
-                        let fileSystem = result.fileSystem
-
-                        // first, decide the tools-version we should use
-                        let preferredToolsVersion = try self.toolsVersionLoader.load(at: .root, fileSystem: fileSystem)
-                        // validate preferred the tools version is compatible with the current toolchain
-                        try preferredToolsVersion.validateToolsVersion(
-                            self.currentToolsVersion,
-                            packageIdentity: self.package.identity
+            switch result {
```
tomerd, line 126: this is just an indention fix

Sources/Workspace/SourceControlPackageContainer.swift
```diff
@@ -397,18 +397,21 @@ internal final class SourceControlPackageContainer: PackageContainer, CustomStri
...
+                version: version,
+                revision: nil,
+                toolsVersion: toolsVersion,
+                identityResolver: self.identityResolver,
+                fileSystem: fileSystem,
+                observabilityScope: self.observabilityScope,
+                delegateQueue: .sharedConcurrent,
+                callbackQueue: .sharedConcurrent,
+                completion: $0
+            )
```
tomerd, line 414: this is just an indention fix

Sources/Workspace/Workspace.swift
```diff
@@ -2137,58 +2137,52 @@ extension Workspace {
...
-                //let manifestLoadingObservabilityScope = observabilityScope.makeChildScope(description: "Loading manifest")
-
-                let manifestLoadingDiagnostics = ThreadSafeArrayStore<Basics.Diagnostic>()
-                let manifestLoadingScope = ObservabilitySystem( { _, diagnostic in
-                    observabilityScope.emit(diagnostic)
-                    manifestLoadingDiagnostics.append(diagnostic)
-                }).topScope.makeChildScope(description: "Loading manifest") {
-                    .packageMetadata(identity: packageIdentity, kind: packageKind)
-                }
+        do {
```
tomerd, line 2140: this is just an indention fix

Tests/WorkspaceTests/ManifestSourceGenerationTests.swift
```diff
@@ -36,18 +36,21 @@ class ManifestSourceGenerationTests: XCTestCase {
...
+                    version: nil,
+                    revision: nil,
+                    toolsVersion: toolsVersion,
+                    identityResolver: identityResolver,
+                    fileSystem: fs,
+                    observabilityScope: observability.topScope,
+                    delegateQueue: .sharedConcurrent,
+                    callbackQueue: .sharedConcurrent,
+                    completion: $0
+                )
```
tomerd, line 53: this is just an indention fix

Tests/WorkspaceTests/ManifestSourceGenerationTests.swift
```diff
@@ -64,18 +67,21 @@ class ManifestSourceGenerationTests: XCTestCase {
...
```
tomerd, line 84: this is just an indention fix

Sources/PackageLoading/ManifestLoader.swift
```diff
@@ -51,7 +51,8 @@ public protocol ManifestLoaderProtocol {
     ///   - identityResolver: A helper to resolve identities based on configuration
     ///   - fileSystem: File system to load from.
     ///   - observabilityScope: Observability scope to emit diagnostics.
-    ///   - on: The dispatch queue to perform asynchronous operations on.
+    ///   - delegateQueue: The dispatch queue to call delegate handlers on.
+    ///   - callbackQueue: The dispatch queue to perform completion handler on.
```
abertelrud, line 55: Is it important for performance to use two queues here, or should this be one queue to ensure that all the delegate methods have completed before the completion closure is called?  For the plugins, that does matter, e.g. receiving all the output before the completion handler is called, but perhaps the situation is different here?  No strong opinion but wanted to check my understanding and assumptions.

I suppose a caller that cares about that can always pass the same queue...

Sources/PackageLoading/ManifestLoader.swift
```diff
@@ -123,6 +128,10 @@ public final class ManifestLoader: ManifestLoaderProtocol {
 
         self.databaseCacheDir = cacheDir.map(resolveSymlinks)
 
+        // this queue and semaphore is used to limit the amount of concurrent manifest loading taking place
+        self.evaluationQueue = OperationQueue()
+        self.evaluationQueue.name = "org.swift.swiftpm.manifest-loader"
+        self.evaluationQueue.maxConcurrentOperationCount = Concurrency.maxOperations
         self.concurrencySemaphore = DispatchSemaphore(value: Concurrency.maxOperations)
```
abertelrud, line 135: Similar question as in the HTTP PR:  if this is the same concurrency value, what does the semaphore provide that the queue limit doesn't?

Sources/PackageLoading/ManifestLoader.swift
abertelrud: If they need to be different, should the second one maybe be called "completionQueue" so it's obvious that it applies to the completion closure parameter?

Sources/PackageLoading/ManifestLoader.swift
abertelrud: Got my answer from that PR: when the operation being performed by the queue is async, the semaphore is needed in order for the operation queue limit to be effective.

abertelrud (approved):

Sources/PackageLoading/ManifestLoader.swift
tomerd: they dont need to be different and in practice they are the same, but the API contract is now clearer and allows the call site to separate between the completion handler queue and the delegate queue _if_ they need to. the API and impl previously intermixed them so the API contract was broken in a way

abertelrud: That makes sense to me.

PR #4181, opened 2022-02-27
http client to support configurable max concurrent requests

tomerd:
motivation: controlling how many concurrent requests are taking place is important for controlling bandwidth requirements, especially as we are transitioning to registry based dependencies

changes:
* add max concurrency controls to http client using a dedicated queue
* remove workspace binary dependencies concurrency control since its not longer necessary (applied at http client level)
* fix a bug in URLSeesion based HTTP client where download can be removed between delegate calls
* add and adjust tests

### DIFF
--- a/Sources/Basics/ConcurrencyHelpers.swift
+++ b/Sources/Basics/ConcurrencyHelpers.swift
@@ -12,6 +12,8 @@ import Dispatch
 import class Foundation.ProcessInfo
 import TSCBasic
 
+public typealias Lock = TSCBasic.Lock
+
 /// Thread-safe dictionary like structure
 public final class ThreadSafeKeyValueStore<Key, Value> where Key: Hashable {
     private var underlying: [Key: Value]
@@ -194,6 +196,12 @@ public final class ThreadSafeBox<Value> {
         }
     }
 
+    public func get(`default`: Value) -> Value {
+        self.lock.withLock {
+            self.underlying ?? `default`
+        }
+    }
+
     public func put(_ newValue: Value) {
         self.lock.withLock {
             self.underlying = newValue
@@ -217,6 +225,23 @@ public final class ThreadSafeBox<Value> {
                 if var value = self.underlying {
                     value[keyPath: keyPath] = newValue
                 }
+            }
+        }
+    }
+}
+
+extension ThreadSafeBox where Value == Int {
+    public func increment() {
+        self.lock.withLock {
+            if let value = self.underlying {
+                self.underlying = value + 1
+            }
+        }
+    }
+    public func decrement() {
+        self.lock.withLock {
+            if let value = self.underlying {
+                self.underlying = value - 1
             }
         }
     }
--- a/Sources/Basics/HTTPClient.swift
+++ b/Sources/Basics/HTTPClient.swift
@@ -13,6 +13,7 @@ import struct Foundation.Data
 import struct Foundation.Date
 import class Foundation.JSONDecoder
 import class Foundation.NSError
+import class Foundation.OperationQueue
 import struct Foundation.URL
 import TSCBasic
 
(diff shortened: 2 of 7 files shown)


Sources/Basics/HTTPClient.swift
```diff
@@ -114,45 +130,65 @@ public struct HTTPClient {
...
-                    // handle retry strategy
-                    if let retryDelay = self.shouldRetry(response: response, request: request, requestNumber: requestNumber) {
-                        observabilityScope?.emit(warning: "\(request.url) failed, retrying in \(retryDelay)")
-                        // TODO: dedicated retry queue?
-                        return self.configuration.callbackQueue.asyncAfter(deadline: .now() + retryDelay) {
-                            self._execute(request: request, requestNumber: requestNumber + 1, observabilityScope: observabilityScope, progress: progress, completion: completion)
+        // we must not block the calling thread (for concurrency control) so nesting this in a queue
+        self.requestsQueue.addOperation {
+            // park the request thread based on the max concurrency allowed
+            self.concurrencySemaphore.wait()
```
tomerd, line 144: this diff look terrible but its mostly just nesting under the new queue dispatch calls

Sources/Workspace/Workspace.swift
```diff
@@ -2538,7 +2535,6 @@ extension Workspace {
                 }
             }
 
-            semaphore.wait()
```
tomerd, line 2541: not needed any more as this was moved into the http client itself

Tests/WorkspaceTests/WorkspaceTests.swift
```diff
@@ -6737,14 +6737,13 @@ final class WorkspaceTests: XCTestCase {
         let sandbox = AbsolutePath("/tmp/ws/")
         let fs = InMemoryFileSystem()
 
-        // this has knowledge of the expected behavior, we can also make this configurable
-        let maxConcurrentRequests = Concurrency.maxOperations
```
tomerd, line 6741: yay!

tomerd:
@swift-ci please smoke test

Sources/Basics/ConcurrencyHelpers.swift
```diff
@@ -194,6 +196,12 @@ public final class ThreadSafeBox<Value> {
         }
     }
 
+    public func get(`default`: Value) -> Value {
```
abertelrud, line 199: Nice!

Sources/Basics/HTTPClient.swift
```diff
@@ -53,6 +59,14 @@ public struct HTTPClient {
...
         // FIXME: inject platform specific implementation here
         self.underlying = handler ?? URLSessionHTTPClient().execute
+
+        // this queue and semaphore is used to limit the amount of concurrent http requests taking place
+        // the default max number of request chosen to match Concurrency.maxOperations which is the number of active CPUs
+        let maxConcurrentRequests = configuration.maxConcurrentRequests ?? Concurrency.maxOperations
+        self.requestsQueue = OperationQueue()
+        self.requestsQueue.name = "org.swift.swiftpm.http-client"
+        self.requestsQueue.maxConcurrentOperationCount = maxConcurrentRequests
+        self.concurrencySemaphore = DispatchSemaphore(value: maxConcurrentRequests)
```
abertelrud, line 69: What is the reason the semaphore is needed If the concurrency queue's max operation count is already set?

tomerd: the underlying network operation is async, so the operation queue is not enough and the semaphore is used to park the operation

abertelrud (approved):

Sources/Basics/HTTPClient.swift
tomerd: exactly

PR #4183, opened 2022-02-27
build: clean up some unnecessary conditions

compnerd:
The dependencies on Foundation are effectively required on all
non-Darwin platforms.  Unconditionalize them and force the link against
the libraries irrespective of whether they are wired up to the build or
rely on the default linker search path.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/Commands/CMakeLists.txt
+++ b/Sources/Commands/CMakeLists.txt
@@ -43,12 +43,8 @@ target_link_libraries(Commands PUBLIC
   Workspace
   XCBuildSupport
   Xcodeproj)
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
-  if(Foundation_FOUND)
-    target_link_libraries(Commands PUBLIC
-      FoundationXML)
-  endif()
-endif()
+target_link_libraries(Commands PRIVATE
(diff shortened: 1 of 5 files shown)


compnerd:
@swift-ci please smoke test

Sources/Commands/CMakeLists.txt
```diff
@@ -43,12 +43,8 @@ target_link_libraries(Commands PUBLIC
   Workspace
   XCBuildSupport
   Xcodeproj)
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
-  if(Foundation_FOUND)
-    target_link_libraries(Commands PUBLIC
-      FoundationXML)
-  endif()
-endif()
+target_link_libraries(Commands PRIVATE
```
tomerd, line 46: public to private intentional? to be clear that change seems fine but making sure its intentional

compnerd: Yeah, I didn't hit any build failures with that change.  `PRIVATE` should be preferred - it prevents all users of this library from linking against `FoundationXML` explicitly.

tomerd (approved):

PR #4185, opened 2022-03-01
refactor how plugins call the build system when using the CLI

tomerd:
motivation: consolidate use of build operation construction which is important for cancellation support

changes:
* use the general purpose function to create a build operation
* small refactoring of how build plan is extracted after the build is complete

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -1170,32 +1170,20 @@ final class PluginDelegate: PluginInvocationDelegate {
 
         // Create a build operation. We have to disable the cache in order to get a build plan created.
         let outputStream = BufferedOutputByteStream()
-        let buildOperation = BuildOperation(
-            buildParameters: buildParameters,
+        let buildOperation = try self.swiftTool.createBuildOperation(
+            explicitProduct: explicitProduct,
             cacheBuildManifest: false,
-            packageGraphLoader: { try self.swiftTool.loadPackageGraph(explicitProduct: explicitProduct) },
-            pluginScriptRunner: try self.swiftTool.getPluginScriptRunner(),
-            pluginWorkDirectory: try self.swiftTool.getActiveWorkspace().location.pluginWorkingDirectory,
-            outputStream: outputStream,
-            logLevel: logLevel,
-            fileSystem: swiftTool.fileSystem,
-            observabilityScope: self.swiftTool.observabilityScope
+            customBuildParameters: buildParameters,
+            customOutputStream: outputStream,
+            customLogLevel: logLevel
         )
 
-        // Save the instance so it can be canceled from the interrupt handler.
-        self.swiftTool.buildSystemRef.buildSystem = buildOperation
-
-        // Get or create the build description and plan the build.
-        let _ = try buildOperation.getBuildDescription()
-        let buildPlan = buildOperation.buildPlan!
-        
         // Run the build. This doesn't return until the build is complete.
-        var success = true
-        do {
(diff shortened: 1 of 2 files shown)


tomerd:
@abertelrud this is a breakout PR for a cleanup required by https://github.com/apple/swift-package-manager/pull/4173

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -1170,32 +1170,20 @@ final class PluginDelegate: PluginInvocationDelegate {
...
-            packageGraphLoader: { try self.swiftTool.loadPackageGraph(explicitProduct: explicitProduct) },
-            pluginScriptRunner: try self.swiftTool.getPluginScriptRunner(),
-            pluginWorkDirectory: try self.swiftTool.getActiveWorkspace().location.pluginWorkingDirectory,
-            outputStream: outputStream,
-            logLevel: logLevel,
-            fileSystem: swiftTool.fileSystem,
-            observabilityScope: self.swiftTool.observabilityScope
+            customBuildParameters: buildParameters,
+            customOutputStream: outputStream,
+            customLogLevel: logLevel
```
tomerd, line 1178: this is the key change - basically use the generic swiftTool.createBuildOperation which will record things with the cancellation utility which is less error prone than doing it manually

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -1170,32 +1170,20 @@ final class PluginDelegate: PluginInvocationDelegate {
...
-        do {
-            try buildOperation.build(subset: buildSubset)
-        }
-        catch {
-            success = false
+        let success = buildOperation.buildIgnoringError(subset: buildSubset)
+
+        // Get the build plan used
+        guard let buildPlan = buildOperation.buildPlan else {
+            throw InternalError("invalid state, buildPlan is undefined")
```
tomerd, line 1186: the above is just a small cleanup

Sources/Commands/SwiftPackageTool.swift
```diff
@@ -1942,3 +1930,14 @@ private extension Basics.Diagnostic {
...
+extension BuildOperation {
+    fileprivate func buildIgnoringError(subset: BuildSubset) -> Bool {
+        do {
+            try self.build(subset: subset)
+            return true
+        } catch {
+            return false
+        }
+    }
+}
```
tomerd, line 1943: we can also use observabilityScope.trap instead of this, but this preserved the previous behavior exactly

tomerd:
@swift-ci please smoke test

Sources/Commands/SwiftPackageTool.swift
abertelrud: This is better than the `!`.  Thanks for cleaning it up!  In practice, should never happen but still.

Sources/Commands/SwiftPackageTool.swift
abertelrud: This is great.

Sources/Commands/SwiftPackageTool.swift
abertelrud: Looks like there was only one call site for this.  Do you think there will be more, or would it be better to keep this do/catch at the one callsite?

abertelrud (approved):

Sources/Commands/SwiftPackageTool.swift
tomerd: normally you want to do something with the error, but in this case the code only seems to care about the end result.

PR #4190, opened 2022-03-03
refactor CLI options

tomerd:
motivation: fix an issue when not all options are presented as expected, and make the code clearer to reason about

changes:
* rename SwiftToolOptions to GlobalOptions
* refactor SwiftToolOptions (GlobalOptions) to multiple groups per domains: locations, caching, logging, build, linker, etc
* set GlobalOptions "hiddenFromHelp" flag to false for swift-build, swift-test and swift-package
* move enableCodeCoverage from GlobalOptions to TestToolOptions, since they apply there
* refactor code that construct test build to pass in more fine grained options rather than take the entire global options model
* adjust call sites and test

rdar://82658867

### DIFF
--- a/Sources/Commands/Options.swift
+++ b/Sources/Commands/Options.swift
@@ -18,7 +18,223 @@ import Build
 import struct TSCUtility.BuildFlags
 import struct TSCUtility.Triple
 
-struct BuildFlagsGroup: ParsableArguments {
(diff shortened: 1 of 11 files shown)


tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

neonichu (approved):

PR #4195, opened 2022-03-04
improve location CLI options

tomerd:
motivation: nicer CLI experience

changes:
* rename --build-path to --scratch-space-path which better aligns with the meaning, and improve help messae
* use the "direcotry" completion option for locations since they mostly point to directories
* hide deprecate options like -C
* rename flag variables to better align with meaning
* adjust call sites

### DIFF
--- a/Sources/Commands/Options.swift
+++ b/Sources/Commands/Options.swift
@@ -66,14 +66,14 @@ struct LocationOptions: ParsableArguments {
     }
 
     /// The custom .build directory, if provided.
-    @Option(name: .customLong("working-directory-path"), help: "Specify a custom working directory path (default .build)", completion: .directory)
-    var _workingDirectory: AbsolutePath?
+    @Option(name: .customLong("scratch-path"), help: "Specify a custom scratch directory path (default .build)", completion: .directory)
+    var _scratchDirectory: AbsolutePath?
 
     @Option(name: .customLong("build-path"), help: .hidden)
     var _deprecated_buildPath: AbsolutePath?
 
-    var workingDirectory: AbsolutePath? {
-        self._workingDirectory ?? self._deprecated_buildPath
+    var scratchDirectory: AbsolutePath? {
+        self._scratchDirectory ?? self._deprecated_buildPath
     }
 
     /// The path to the file containing multiroot package data. This is currently Xcode's workspace file.
--- a/Sources/Commands/Options.swift
+++ b/Sources/Commands/Options.swift
@@ -46,33 +46,42 @@ struct GlobalOptions: ParsableArguments {
 struct LocationOptions: ParsableArguments {
     init() {}
 
-    /// The custom build directory, if provided.
-    @Option(help: "Specify build/cache directory")
-    var buildPath: AbsolutePath?
+    @Option(name: .customLong("package-path"), help: "Specify the package path to operate on (default current directory). This changes the working directory before any other operation", completion: .directory)
+    var _packageDirectory: AbsolutePath?
 
-    @Option(help: "Specify the shared cache directory")
-    var cachePath: AbsolutePath?
+    @Option(name: .customLong("cache-path"), help: "Specify the shared cache directory path", completion: .directory)
+    var cacheDirectory: AbsolutePath?
 
-    @Option(help: "Specify the shared configuration directory")
-    var configPath: AbsolutePath?
+    @Option(name: .customLong("config-path"), help: "Specify the shared configuration directory path", completion: .directory)
+    var configurationDirectory: AbsolutePath?
 
-    @Option(help: "Specify the shared security directory")
-    var securityPath: AbsolutePath?
+    @Option(name: .customLong("security-path"), help: "Specify the shared security directory path", completion: .directory)
+    var securityDirectory: AbsolutePath?
 
-    /// The custom working directory that the tool should operate in (deprecated).
-    @Option(name: [.long, .customShort("C")])
-    var chdir: AbsolutePath?
+    @Option(name: [.long, .customShort("C")], help: .hidden)
+    var _deprecated_chdir: AbsolutePath?
 
-    /// The custom working directory that the tool should operate in.
-    @Option(help: "Change working directory before any other operation")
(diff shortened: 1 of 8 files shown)


tomerd:
@swift-ci please smoke test

tomerd:

tomerd:

Sources/Commands/Options.swift
```diff
@@ -46,33 +46,43 @@ struct GlobalOptions: ParsableArguments {
 struct LocationOptions: ParsableArguments {
     init() {}
 
-    /// The custom build directory, if provided.
-    @Option(help: "Specify build/cache directory")
-    var buildPath: AbsolutePath?
+    /// The custom .build directory, if provided.
+    @Option(name: .customLong("scratch-space-path"), help: "Specify the scratch space directory (default .build)", completion: .directory)
```
abertelrud, line 50: This name seems unfortunate to me but I don't have a better suggestion right now.  I'll try to come up with one and suggest it here.  I agree that "build" is too narrow here for what it does.

Sources/Commands/Options.swift
```diff
@@ -46,33 +46,43 @@ struct GlobalOptions: ParsableArguments {
...
-    /// The custom working directory that the tool should operate in (deprecated).
-    @Option(name: [.long, .customShort("C")])
-    var chdir: AbsolutePath?
+    @Option(name: .customLong("security-path"), help: "Specify the shared security directory", completion: .directory)
+    var securityDirectory: AbsolutePath?
 
     /// The custom working directory that the tool should operate in.
-    @Option(help: "Change working directory before any other operation")
-    var packagePath: AbsolutePath?
+    @Option(name: .customLong("package-path"), help: "Change working directory before any other operation", completion: .directory)
```
abertelrud, line 70: It's actually a bit surprising that this is implemented as changing directory instead of just referencing a package.  I think I was able to mitigate that for plugins by having them start in the original working directory, e.g. for a plugin invoked from the command line it's natural for relative arguments to be relative to the current working directory.  But I thought that was just an implementation shortcut and not how it's officially documented.

abertelrud: I see now that `git -C` works this way, and perhaps it makes sense.  Interestingly, `swift run` doesn't seem to work that way, as `swift run --package-path /tmp/MyExec MyExec` will run in the original working directory (which is what I'd expect).

Sources/Commands/Options.swift
```diff
@@ -46,33 +46,43 @@ struct GlobalOptions: ParsableArguments {
...
+    }
 
     /// The path to the file containing multiroot package data. This is currently Xcode's workspace file.
-    @Option(name: .customLong("multiroot-data-file"), completion: .file())
+    @Option(name: .customLong("multiroot-data-file"), help: .hidden, completion: .directory)
     var multirootPackageDataFile: AbsolutePath?
 
     /// Path to the compilation destination describing JSON file.
-    @Option(name: .customLong("destination"), completion: .file())
+    @Option(name: .customLong("destination"), help: .hidden, completion: .directory)
```
abertelrud, line 85: What does this one even do?  Is it something that should still be supported in some fashion?

abertelrud: Not saying it should, just raising the question since it will now get hidden.  That's probably a good thing though — there seems to have been organic growth among these options (which isn't a catastrophe) but it's nice to trim it down, as you're doing.

Sources/Commands/Options.swift
tomerd: right, I am not 100% set on it. `build` is too narrow and confusion. `workspace` could work also

Sources/Commands/Options.swift
tomerd: this PR does not change this functionality - it only cleans up the way the CLI flags are organized. material for future PR?

Sources/Commands/Options.swift
tomerd: it defines where the llbuild manifest get written to. there is not user facing feature I am aware of around this so my assumption is an accidental leak of a flag designed for for debugging

tomerd:
@swift-ci please smoke test

Sources/Commands/Options.swift
abertelrud: Yes, I think that would be better than "scratch-space".

Sources/Commands/Options.swift
abertelrud: Yes, good point, this is unrelated to this PR.  I don't know whether that qualifies as a bug to be fixed or not — it just seemed surprising to me when I discovered it.

Sources/Commands/Options.swift
abertelrud: Great, thanks for explaining.

abertelrud:
Looks good to me but I do think "working-directory" would be better than "scratch-space" for the location formerly known as "build".

tomerd:
@abertelrud updated

abertelrud (approved):

abertelrud:
Looks good to me.

Sources/Commands/Options.swift
```diff
@@ -46,33 +46,42 @@ struct GlobalOptions: ParsableArguments {
...
-    /// The custom working directory that the tool should operate in.
-    @Option(help: "Change working directory before any other operation")
-    var packagePath: AbsolutePath?
+    var packageDirectory: AbsolutePath? {
+        self._packageDirectory ?? self._deprecated_chdir
+    }
+
+    /// The custom .build directory, if provided.
+    @Option(name: .customLong("working-directory-path"), help: "Specify a custom working directory path (default .build)", completion: .directory)
+    var _workingDirectory: AbsolutePath?
```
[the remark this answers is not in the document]
abertelrud: That's a great point, and from that perspective, `scratch-directory` is probably the best.

tomerd:
@swift-ci please smoke test

tomerd:

tomerd:

tomerd:
@swift-ci please smoke test

PR #4200, opened 2022-03-07
SR-15929: Package plugin dependency on executable product from different package doesn't find the tool

abertelrud:
### Motivation:

Fix a bug that caused a plugin with a dependency on an executable product from different package to not find the tool when looking it up with `context.tool(named:)`.  Dependencies on executables in the same package as the plugin did work.

### Modifications:

- add case to handle `.product()` in addition to `.target()` for dependencies built from executables
- add test fixture
- add unit test

rdar://89693333


abertelrud:
@swift-ci please smoke test

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -480,8 +481,12 @@ public extension PluginTarget {
                     // TODO: How do we determine what the executable name will be for the host platform?
                     return [.builtTool(name: target.name, path: RelativePath(target.name))]
                 }
+                else {
+                    return []
+                }
+            case .product(let product, _):
+                return [.builtTool(name: product.name, path: RelativePath(product.name))]
```
abertelrud, line 488: This is the actual fix.  For a product we need the relative path of the produced executable, which will be the same as the product name.  The plugin itself will add any platform specific suffix, such as `.exe` on Windows.

tomerd (approved):

neonichu (approved):

PR #4201, opened 2022-03-08
Make the `WorkspaceTests.testSimpleAPI` test more robust w.r.t. XCTest minimum deployment version

abertelrud:
### Motivation:

This test is dependent on the minimum deployment version of the installed XCTest.  This can result in warnings causing the test to fail.

### Modifications:

- when creating the test fixture, pass options to honor the min deployment target

### Discussion

It's possible that package creation should always set a minimum deployment target of at least what XCTest supports as minimum.  For my toolchain this is macOS 11.  But it's a more general problem, since this can vary.

It's also not clear that the minimum version if there is no declaration should remain at 10.15.4.  Perhaps tools version 5.7 should increase that to 11?

This test only sets the macOS deployment version since that's the only place I've seen this come up.  Iterating over all the platforms doesn't easily work since some of them aren't supported in the manifest with the same names.

However we do this, we should find a way to make the test not fail.  Perhaps we could also strip out the version warnings.

Putting this up partly for discussion, but we do want to avoid the warnings.

### DIFF
--- a/Tests/WorkspaceTests/WorkspaceTests.swift
+++ b/Tests/WorkspaceTests/WorkspaceTests.swift
@@ -4147,9 +4147,13 @@ final class WorkspaceTests: XCTestCase {
         try testWithTemporaryDirectory { path in
             // Create a temporary package as a test case.
             let packagePath = path.appending(component: "MyPkg")
+            let xctestMinVersions = MinimumDeploymentTarget.default.xcTestMinimumDeploymentTargets
             let initPackage = try InitPackage(
                 name: packagePath.basename,
-                packageType: .executable,
+                options: .init(
+                    packageType: .executable,
+                    platforms: xctestMinVersions[.macOS].map{ .init(platform: .macOS, version: $0) }.map{ [$0] } ?? []
+                ),
                 destinationPath: packagePath,
                 fileSystem: localFileSystem
             )


abertelrud:
@swift-ci please smoke test

SDGGiesbrecht:
I see these warnings most often with watchOS. I trained CI to ignore them, because there is no reason the library itself and its clients should be constrained just because the tests cannot run on older devices. If deployment targets could be set on a per‐target basis, I would simply set them on the test targets. But without the ability to do that, I have just learned to live with them as omnipresent warnings.

Maybe, as long as XCTest is the only supported testing platform, SwiftPM could infer minimums for test targets based on the XCTest in the associated SDK?

Tests/WorkspaceTests/WorkspaceTests.swift
```diff
@@ -4147,9 +4147,13 @@ final class WorkspaceTests: XCTestCase {
...
             // Create a temporary package as a test case.
             let packagePath = path.appending(component: "MyPkg")
+            let xctestMinVersions = MinimumDeploymentTarget.default.xcTestMinimumDeploymentTargets
             let initPackage = try InitPackage(
                 name: packagePath.basename,
-                packageType: .executable,
+                options: .init(
+                    packageType: .executable,
+                    platforms: xctestMinVersions[.macOS].map{ .init(platform: .macOS, version: $0) }.map{ [$0] } ?? []
+                ),
```
tomerd, line 4156: not sure about this change. the test is mimicking what users will get when they create a new package and if that emits warning / requires a min deployment target not to emit warning that is a not a good sign

tomerd: > Maybe, as long as XCTest is the only supported testing platform, SwiftPM could infer minimums for test targets based on the XCTest in the associated SDK?

I thought this was already the case?

SDGGiesbrecht: Possibly. I just checked and the warning seems to be gone.

abertelrud: > Maybe, as long as XCTest is the only supported testing platform, SwiftPM could infer minimums for test targets based on the XCTest in the associated SDK?

> I thought this was already the case?

Perhaps this is the real bug.  I agree this wasn't the right fix but will use this PR to get to the bottom of this when I find the time.

PR #4203, opened 2022-03-08
consolidate and improve manifest validation

tomerd:
motivation: make code easier to reason about and maintain

changes:
* create new ManfeistValidator utility
* move manifest validation code out of ManifestLoader and Workspace and into the new validation utility
* seperate out manifst validation diagnostics from manifest loading diagnostics such that a fake diagnostics scope is no longer required
* adjust call sites and test

### DIFF
--- a/Sources/PackageLoading/ManifestLoader+Validation.swift
+++ b/Sources/PackageLoading/ManifestLoader+Validation.swift
@@ -1,0 +1,327 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Basics
+import Foundation
+import PackageModel
+import TSCBasic
+
+public struct ManifestValidator {
+    static var supportedLocalBinaryDependencyExtensions: [String] {
+        ["zip"] + BinaryTarget.Kind.allCases.filter{ $0 != .unknown }.map { $0.fileExtension }
+    }
+    static var supportedRemoteBinaryDependencyExtensions: [String] {
+        ["zip", "artifactbundleindex"]
+    }
+
+    private let manifest: Manifest
+    private let sourceControlValidator: ManifestSourceControlValidator
+    private let fileSystem: FileSystem
+
+    public init(manifest: Manifest, sourceControlValidator: ManifestSourceControlValidator, fileSystem: FileSystem) {
+        self.manifest = manifest
+        self.sourceControlValidator = sourceControlValidator
+        self.fileSystem = fileSystem
+    }
+
+    /// Validate the provided manifest.
+    public func validate() -> [Basics.Diagnostic] {
+        var diagnostics = [Basics.Diagnostic]()
+
+        diagnostics += self.validateTargets()
+        diagnostics += self.validateProducts()
+        diagnostics += self.validateDependencies()
+
+        // Checks reserved for tools version 5.2 features
+        if self.manifest.toolsVersion >= .v5_2 {
+            diagnostics += self.validateTargetDependencyReferences()
+            diagnostics += self.validateBinaryTargets()
+        }
+
+        return diagnostics
+    }
+
+    private func validateTargets() -> [Basics.Diagnostic] {
+        var diagnostics = [Basics.Diagnostic]()
+
+        let duplicateTargetNames = self.manifest.targets.map({ $0.name }).spm_findDuplicates()
+        for name in duplicateTargetNames {
+            diagnostics.append(.duplicateTargetName(targetName: name))
+        }
+
+        return diagnostics
+    }
+
+    private func validateProducts()  -> [Basics.Diagnostic] {
+        var diagnostics = [Basics.Diagnostic]()
+        
+        for product in self.manifest.products {
+            // Check that the product contains targets.
+            guard !product.targets.isEmpty else {
+                diagnostics.append(.emptyProductTargets(productName: product.name))
+                continue
+            }
+
+            // Check that the product references existing targets.
+            for target in product.targets {
+                if !self.manifest.targetMap.keys.contains(target) {
+                    diagnostics.append(.productTargetNotFound(productName: product.name, targetName: target, validTargets: self.manifest.targetMap.keys.sorted()))
+                }
+            }
+
+            // Check that products that reference only binary targets don't define a type.
+            let areTargetsBinary = product.targets.allSatisfy { self.manifest.targetMap[$0]?.type == .binary }
+            if areTargetsBinary && product.type != .library(.automatic) {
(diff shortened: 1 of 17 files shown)


tomerd:
@neonichu @abertelrud this needs a bit more work to get the tests aligned, but wanted early feedback on the approach

Sources/Workspace/Workspace.swift
```diff
@@ -2172,17 +2168,21 @@ extension Workspace {
...
-                    manifestLoadingScope.trap { try self.validateManifest(manifest) }
-                    self.delegate?.didLoadManifest(packagePath: packagePath, url: packageLocation, version: version, packageKind: packageKind, manifest: manifest, diagnostics: manifestLoadingDiagnostics.get())
+                    let validationDiagnostics = ManifestValidator.validate(manifest, sourceControlValidator: self.repositoryManager, fileSystem: self.fileSystem)
+                    if !validationDiagnostics.isEmpty {
+                        result = .failure(Diagnostics.fatalError)
+                        manifestLoadingDiagnostics.append(contentsOf: validationDiagnostics)
+                    }
+                    self.delegate?.didLoadManifest(packagePath: packagePath, url: packageLocation, version: version, packageKind: packageKind, manifest: manifest, diagnostics: manifestLoadingDiagnostics)
                 }
+                manifestLoadingScope.emit(manifestLoadingDiagnostics)
```
tomerd, line 2185: here is the key change (and motivation): it allows us to get rid of the fake observability scope and special casing Diagnostics.fatalError

Tests/PackageLoadingTests/PDLoadingTests.swift
```diff
@@ -30,30 +30,30 @@ class PackageDescriptionLoadingTests: XCTestCase, ManifestLoaderDelegate {
...
+    func loadAndValidateManifest(
         _ contents: String,
         toolsVersion: ToolsVersion? = nil,
         packageKind: PackageReference.Kind? = nil,
         observabilityScope: ObservabilityScope,
         file: StaticString = #file,
         line: UInt = #line
-    ) throws -> Manifest {
-        try self.loadManifest(
+    ) throws -> (manifest: Manifest, diagnostics: [Basics.Diagnostic])  {
```
tomerd, line 40: the validator does not take an observability scope, instead it returns an array of diagnostics which the caller can decide what to with

Sources/Workspace/Workspace.swift
abertelrud: That looks like a good approach.  And this means the `didLoadManifest()` callback and the diagnostics handler will always see all the diagnostics, from the looks of it.

Does that mean that a client that just wanted to capture diagnostics at the end (in `didLoadManifest()`) could pass a no-op manifestLoadingScope and just implement the delegate callback?  In practice it seems that any given client would only want to do one or the other, right, otherwise it would get duplicate diagnostics.

abertelrud: I realize this might not be any different from before because of this PR, but it's always been somewhat unclear here what a client should ideally do to avoid duplication.

Tests/PackageLoadingTests/PDLoadingTests.swift
abertelrud: What's the advantage of doing it that way rather than a custom scope that just captures them and then passes them on?  That would allow the validation to use the normal pattern for emitting diagnostics.

Sources/PackageLoading/ManifestLoader+Validation.swift
```diff
@@ -0,0 +1,346 @@
...
+//
+//  Created by Tom Doron on 3/4/22.
+//
+
+import Basics
+import Foundation
+import PackageModel
+import TSCBasic
+
+public enum ManifestValidator {
```
abertelrud, line 13: What's the advantage of making everything static here rather than instantiating a `ManifestValidator` struct with a manifest so that the manifest doesn't have to be passed around to all the functions?  I realize there's no material difference but it seems conceptually a bit cleaner if there is every, for example, derived state that the validator might want to create and keep (e.g. lazily initialized lookup tables or whatnot).

tomerd: that's a good idea!

Sources/Workspace/Workspace.swift
tomerd: > Does that mean that a client that just wanted to capture diagnostics at the end (in didLoadManifest()) could pass a no-op manifestLoadingScope and just implement the delegate callback? In practice it seems that any given client would only want to do one or the other, right, otherwise it would get duplicate diagnostics.

who is the client in this case? this is private function, and so the observabilityScope passed to it may contain other diagnostics above and beyond manifest loading

Tests/PackageLoadingTests/PDLoadingTests.swift
tomerd: an API design preference: a validator is designed to return a list of validation issues vs. some functional api that _may_ need to emit diagnostics so takes a "scope" to do so

tomerd:
@swift-ci smoke test

tomerd:
@swift-ci smoke test

tomerd:
@abertelrud @neonichu this has been rebased and ready for review

Sources/PackageLoading/ManifestLoader+Validation.swift
```diff
@@ -0,0 +1,322 @@
+//
+//  File.swift
+//  
+//
+//  Created by Tom Doron on 3/4/22.
+//
```
abertelrud, line 6: Boilerplate header?

Sources/PackageLoading/ManifestLoader+Validation.swift
```diff
@@ -0,0 +1,322 @@
...
+        for target in self.manifest.targets {
+            for targetDependency in target.dependencies {
+                switch targetDependency {
+                case .target:
+                    // If this is a target dependency, we don't need to check anything.
+                    break
+                case .product(_, let packageName, _, _):
+                    if self.manifest.packageDependency(referencedBy: targetDependency) == nil {
+                        diagnostics.append(.unknownTargetPackageDependency(
+                            packageName: packageName ?? "unknown package name",
```
abertelrud, line 186: I thought the `nil` case of a package name meant "the local package", in which case is the unknown package name correct behaviour?

abertelrud: Oh I see now that this was existing code that was just moved around.  Still, I wonder if it's correct.

Sources/Workspace/Workspace.swift
abertelrud: Sorry, I misread this and didn't realize that it's a private function (thus the client would be the client of libSwiftPM, but that point is moot, as it's private).

I agree with the other sentiment and think we'll want to revisit the delegate here at some point.  There's a bit of a split personality between the delegate callbacks but then the diagnostics in the scope.

abertelrud:
Looks good, and the only nit now was about what looks like a boilerplate source file header (which I guess that new script could fix actually  :-).

abertelrud (approved):

tomerd:
@swift-ci smoke test

tomerd:
@swift-ci smoke test

tomerd:
@swift-ci smoke test

PR #4204, opened 2022-03-08
improve validation and error message for mismatch between binary target and artifact name

tomerd:
motivation: binary target artifact lookup is name based, but can be confusing given the file extension stripping used

changes:
* improve the diagnostics when binary target is not found to specify the expected binary artifact name
* soften the extension stripping logic to strip all extensions, not just the last one
* pass the target -> BinaryArtifact mapping to PackageBuilder so it does not need to look up by path
* udpate and add tests

rdar://89622333


tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

neonichu (approved):

abertelrud (approved):

PR #4216, opened 2022-03-15
Allow empty unsafeFlags from dependencies

keith:
Some projects like SwiftSyntax conditionally add some unsafe flags based
on environment variables. In the case they don't need them they add an
empty unsafe flags target. Previously this caused a failure, with this
change there is only a failure in the case the unsafe declaration
provides flags.

SwiftSyntax has also updated its Package.swift to not add the
empty `.unsafeFlags([])`, but this seems reasonable to be less strict
about as well.

Fixes https://bugs.swift.org/browse/SR-15989


keith:
@swift-ci please test

tomerd (approved):

tomerd:
thanks @Keith . worth back portin  into 5.6 as well?

tomerd:
@swift-ci please smoke test

keith:
Happy to submit it, seems low risk enough, I can do that after this merges!

tomerd:

tomerd:

keith:
https://github.com/apple/swift-package-manager/pull/4222

PR #4219, opened 2022-03-16
Support OCSP check via HTTP GET on non-Darwin platforms

yim-lee:
Motivation:
OCSP check can be done via HTTP POST or GET. The current code only does
POST and gives up if that fails.

Modification:
Try GET if POST fails.

rdar://90280783


yim-lee:
@swift-ci please smoke test

tomerd (approved):

tomerd:
not sure if we can also add a test?

tomerd:

yim-lee:
> not sure if we can also add a test?

There is but needs real cert to run: https://github.com/apple/swift-package-manager/blob/main/Tests/PackageCollectionsSigningTests/CertificatePolicyTests.swift#L116

neonichu (approved):

PR #4220, opened 2022-03-16
integrate plugins with new cancellation sub-system

tomerd:
motivation: consistent cancellation system to include plugins

changes:
* DefaultPluginScriptRunner is now Cancellable, with an internal copy of canncelator to maange its internal state
* dedup call sites that create DefaultPluginScriptRunner so that we can register it more easily with the SwiftTool::Cancellator instance
* Cancellator now supports Foundation::Process
* add and adjust tests

### DIFF
--- a/Sources/Commands/SwiftTool.swift
+++ b/Sources/Commands/SwiftTool.swift
@@ -403,7 +403,6 @@ public class SwiftTool {
             SwiftTool.cancellator = cancellator
             _ = SetConsoleCtrlHandler({ _ in
                 // Terminate all processes on receiving an interrupt signal.
-                DefaultPluginScriptRunner.cancelAllRunningPlugins()
                 try? SwiftTool.cancellator?.cancel(deadline: .now() + .seconds(30))
 
                 // Reset the handler.
@@ -423,7 +422,6 @@ public class SwiftTool {
                 interruptSignalSource.cancel()
 
                 // Terminate all processes on receiving an interrupt signal.
-                DefaultPluginScriptRunner.cancelAllRunningPlugins()
                 try? cancellator.cancel(deadline: .now() + .seconds(30))
 
                 #if os(macOS) || os(OpenBSD)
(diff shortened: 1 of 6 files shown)


tomerd:
@swift-ci please smoke test

tomerd:

Sources/Commands/SwiftTool.swift
```diff
@@ -403,7 +403,6 @@ public class SwiftTool {
             SwiftTool.cancellator = cancellator
             _ = SetConsoleCtrlHandler({ _ in
                 // Terminate all processes on receiving an interrupt signal.
-                DefaultPluginScriptRunner.cancelAllRunningPlugins()
```
abertelrud, line 406: Nice!

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -506,22 +508,10 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
...
-    /// Cancels all currently running plugins, resulting in an error code indicating that they were interrupted. This is intended for use when the host process is interrupted.
-    public static func cancelAllRunningPlugins() {
-#if !os(iOS) && !os(watchOS) && !os(tvOS)
-        currentlyRunningPlugins.lock.withLock {
-            currentlyRunningPlugins.processes.forEach{
-                $0.terminate()
-            }
-            currentlyRunningPlugins.processes = []
-        }
-#endif
```
abertelrud, line 519: It's great to get this out of here.  Thanks for taking on this task!

abertelrud (approved):
This looks great.  Thanks for adapting plugins to cancellation!

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -506,22 +508,10 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner {
...
+
+    public func cancel(deadline: DispatchTime) throws {
+        try self.cancellator.cancel(deadline: deadline)
```
abertelrud, line 513: If I read the underlying code correctly, this will block until the cancellation has happened (or deadline passes), right, so should always be called from a background thread in an interactive environment?

tomerd: yes this call is blocking. we can add a non-blocking variant if useful

PR #4222, opened 2022-03-16
[5.6] Allow empty unsafeFlags from dependencies (#4216)

keith:
Some projects like SwiftSyntax conditionally add some unsafe flags based
on environment variables. In the case they don't need them they add an
empty unsafe flags target. Previously this caused a failure, with this
change there is only a failure in the case the unsafe declaration
provides flags.

Theoretically SwiftSyntax could update its Package.swift to not add the
empty `.unsafeFlags([])`, but this seems reasonable to be less strict
about as well.

Fixes https://bugs.swift.org/browse/SR-15989

(cherry picked from commit 1e79329be7e3588016d43a8a96b5128429e22078)


keith:
@swift-ci please test

neonichu (approved):

tomerd:
@swift-ci please test

tomerd:
@keith looks like the 5.6 code base is a bit different - rebase on top of release/5.6?

keith:
Ah sorry, there weren't any conflicts so I assumed it was still valid, updated!

keith:
@swift-ci please smoke test

tomerd:
@swift-ci please smoke test

PR #4225, opened 2022-03-17
improve platform conditional imports

tomerd:
motivation: more reliable conditional imports

changes: use canImport(module) instead of os(OS) to drive conditional imports

### DIFF
--- a/Sources/Basics/HTTPClient.swift
+++ b/Sources/Basics/HTTPClient.swift
@@ -20,9 +20,7 @@ import TSCBasic
 
 #if canImport(Glibc)
 import Glibc
-#endif
-
-#if os(Windows)
+#elseif canImport(CRT)
 import CRT
 #endif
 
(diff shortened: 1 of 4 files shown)


tomerd:
@swift-ci please smoke test

tomerd:
@swift-ci smoke test

Sources/Basics/HTTPClient.swift
```diff
@@ -20,9 +20,7 @@ import TSCBasic
 
 #if canImport(Glibc)
 import Glibc
-#endif
-
-#if os(Windows)
+#elseif canImport(CRT)
```
abertelrud, line 23: Is there a risk of something called CRT in the search path on other platforms?

tomerd: yes, but I dont think its very realistic. we can always make it more complicated if/when we run into that

abertelrud (approved):

PR #4226, opened 2022-03-17
Re-add class name to '__allTests' for test discovery

glbrntt:
Motivation:

In https://github.com/apple/swift-package-manager/pull/3992 the class name was stripped from the static array of tests (i.e. `__allTests_CLASSNAME` became `__allTests`). The class name ensured the variable was uniquely named. Without it, in a test class which is derived from another, it is not unique and results in compile-time errors. 

This is a regression introduced into Swift 5.6. See also: https://bugs.swift.org/browse/SR-15955

Modifications:

- Reintroduce the class name
- Add tests

Result:

Derived test classes do not result in compilation errors on Linux.

### DIFF
--- a/Fixtures/Miscellaneous/TestDiscovery/Subclass/Package.swift
+++ b/Fixtures/Miscellaneous/TestDiscovery/Subclass/Package.swift
@@ -1,0 +1,10 @@
+// swift-tools-version:5.6
+import PackageDescription
+
+let package = Package(
+    name: "Subclass",
+    targets: [
+        .target(name: "Subclass"),
+        .testTarget(name: "SubclassTests", dependencies: ["Subclass"]),
+    ]
+)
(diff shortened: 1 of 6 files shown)


neonichu (approved):

tomerd:
@swift-ci smoke test

tomerd:
good catch and thanks for the fix @glbrntt, once merged could you also cherry-pick for release/5.6 so we can include in the next patch release?

tomerd:
@swift-ci smoke test linux

glbrntt:
> good catch and thanks for the fix @glbrntt, once merged could you also cherry-pick for release/5.6 so we can include in the next patch release?

No problem. Cherry picked: https://github.com/apple/swift-package-manager/pull/4234

PR #4233, opened 2022-03-18
Allow binary targets to change checksums and URLs at the same time

keith:
Previously if you had a binary target, that you replaced the checksum
and URL for at the same time, SwiftPM would reject it because the
checksum had changed. In my experience this is the common case when you
are copy pasting the inclusion instructions from documentation for an
update and then you run `swift build`.

In the case of package resolution the addedOrUpdatedPackages list is
always empty.

### DIFF
--- a/Sources/Workspace/Workspace.swift
+++ b/Sources/Workspace/Workspace.swift
@@ -2358,14 +2358,15 @@ extension Workspace {
             ]
 
             if let existingArtifact = existingArtifact {
-                if case .remote(_, let existingChecksum) = existingArtifact.source {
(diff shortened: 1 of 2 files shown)


keith:
@swift-ci please smoke test

tomerd (approved):
nice catch, thanks @Keith. this is another one that could be nice to bring into 5.6

tomerd:
@swift-ci smoke test

keith:
@swift-ci please smoke test macOS Platform

tomerd:

keith:
https://github.com/apple/swift-package-manager/pull/4238

PR #4231, opened 2022-03-18
[5.6] Fix SIGINT handling during `swift run` (Cherry-pick of #4224, #4235, #4250)

kkk669:
This is a cherry-pick PR of #4224, #4235, and #4250 for Swift 5.6.

(CC: @tomerd)

rdar://90574791

PR #4235, opened 2022-03-18
Fix a build failure on Windows (#4224)

kkk669:
This is a hot fix for Windows build broken by #4224. #4231 also needs this change.

### Motivation:

#4224 broke Windows build as @compnerd pointed out at https://github.com/apple/swift-package-manager/pull/4224#issuecomment-1072474100. The global function `exec` added in #4224 is actually unnecessary for Windows because [`signal(SIGINT, SIG_IGN)`](https://github.com/apple/swift-package-manager/blob/0e804edf19e05ddd02d05c1534d4ceea01117e5f/Sources/Commands/SwiftTool.swift#L418) is not compiled for Windows target.

```swift
            #if os(Windows)
            // set shutdown handler to terminate sub-processes, etc
            SwiftTool.cancellator = cancellator
                try? SwiftTool.cancellator?.cancel(deadline: .now() + .seconds(30))

                // Reset the handler.
                _ = SetConsoleCtrlHandler(nil, false)

                // Exit as if by signal()
                TerminateProcess(GetCurrentProcess(), 3)

                return true
            }, true)
            #else
            // trap SIGINT to terminate sub-processes, etc
            signal(SIGINT, SIG_IGN)
            ...
            #endif
```

### Modifications:

- Enclosed `exec` with `#if !(os(Windows))` and `#endif`
- Changed comments to be clearer

### Result:

Combined with this change, #4224 no longer affects Windows.

### DIFF
--- a/Sources/Commands/SwiftTool.swift
+++ b/Sources/Commands/SwiftTool.swift
@@ -288,8 +288,11 @@ extension SwiftCommand {
 
 /// A safe wrapper of TSCBasic.exec.
 func exec(path: String, args: [String]) throws -> Never {
-    // SwiftTool can't handle signals anymore, so reset the signal handler to SIG_DFL before call TSCBasic.exec()
+    #if !os(Windows)
+    // On platforms other than Windows, signal(SIGINT, SIG_IGN) is used for handling SIGINT by DispatchSourceSignal,
+    // but this process is about to be replaced by exec, so SIG_IGN must be returned to default.
     signal(SIGINT, SIG_DFL)
+    #endif
 
     try TSCBasic.exec(path: path, args: args)
 }


compnerd:
@swift-ci please smoke test

kkk669:
@compnerd 

> (Having a separate exec is likely going to cause further conflicts as things evolve ... this should handle the Windows path internally).

I initially considered alternatives such as `SwiftTool.exec`. However, there is a concern that when changes are made to add TSCBasic's `exec` in the future, the author may forget to use `SwiftTool.exec`.

In light of that, it might be better to revert and reimplement #3815 in other ways.

> this should have an analogous counterpart to Windows. 

I think that would be the most ideal too.

compnerd:
> I initially [considered alternatives such as `SwiftTool.exec`](https://github.com/apple/swift-package-manager/pull/4224#discussion_r829338932). However, there is a concern that when changes are made to add TSCBasic's `exec` in the future, the author may forget to use `SwiftTool.exec`.

I think that we could just bring over `exec` from `TSCBasic`.  I haven't gotten around to removing the last of the `TSCUtility` from swift-driver, which would then allow us to focus on removing the uses of `TSCBasic`.  Doing that more immediately is reasonable.  We should be able to entirely remove the use of `TSCBasic.exec` and use the internal `exec` only.

kkk669:
`TSCBasic` provides other types and operators as well, so it's difficult for me to determine if `TSCBasic` should be removed.

By any chance, could your concern be addressed with a code such as the following?

```swift
func exec(...) {
    #if !(os(Windows))
    // signal(SIGINT, SIG_IGN) is used for handling SIGINT by DispatchSourceSignal,
    // but this process is about to be replaced by exec, so SIG_IGN must be returned to default.
    signal(SIGINT, SIG_DFL)
    #endif

    TSCBasic.exec(...)
}

...

public class SwiftTool {
    ...

    // marked internal for testing
    internal init(outputStream: OutputByteStream, options: GlobalOptions) throws {
        ...

        do {
            ...

            #if os(Windows)
            ...
            #else
            ...
            signal(SIGINT, SIG_IGN)
            ...
            #endif

            ...
        } catch {
            ...
        }

        ...
    }

    ...
}

...
```

tomerd (approved):

tomerd:

compnerd:
Yeah, I think that would be better than the current version.

kkk669:
@tomerd I pushed that change.

@tomerd Sorry for the change after you approve. Please could you check it again?

tomerd:
@swift-ci smoke test

tomerd:
no problem @kkk669 please bring this commit over to the 5.6 PR once merged

kkk669:
Yes sure. Thank you.

tomerd:
@swift-ci smoke test linux

compnerd:

tomerd:
@swift-ci smoke test linux

tomerd:
@swift-ci smoke test linux

tomerd:
@swift-ci smoke test linux

compnerd:

PR #4236, opened 2022-03-18
The last amount of `stderr` output from a plugin subprocess could occasionally get lost

abertelrud:
This could happen because the Process termination handler only called `availableData` and not `readToEnd()`, so based on timing, there might be data that had been emitted by the plugin subprocess before it exited but that hadn't yet been received.

### Motivation:

This caused sporadic failures of the `testLocalAndRemoteToolDependencies` unit test.  And could of course sometimes cause output to be lost in actual usage too.

### Modifications:

The fix is to switch to using `readToEnd()`.  This was not a problem for messages from the plugin, because they were already being read (and not collected using `availableData`).

rdar://89490777

### DIFF
--- a/Sources/Workspace/DefaultPluginScriptRunner.swift
+++ b/Sources/Workspace/DefaultPluginScriptRunner.swift
@@ -444,14 +444,16 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
         let stderrPipe = Pipe()
         let stderrLock = Lock()
         var stderrData = Data()
-        stderrPipe.fileHandleForReading.readabilityHandler = { fileHandle in
+        let stderrHandler = { (data: Data) in
             // Pass on any available data to the delegate.
-            stderrLock.withLock {
-                let data = fileHandle.availableData
-                if data.isEmpty { return }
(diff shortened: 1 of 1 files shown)


abertelrud:
@swift-ci please smoke test

tomerd (approved):

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -444,14 +444,16 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
         let stderrPipe = Pipe()
         let stderrLock = Lock()
         var stderrData = Data()
-        stderrPipe.fileHandleForReading.readabilityHandler = { fileHandle in
+        let stderrHandler = { (data: Data) in
             // Pass on any available data to the delegate.
-            stderrLock.withLock {
```
abertelrud, line 449: Note that the lock isn't going away, it's being moved down to Line 456 since the `availableData`/`readToEnd()` has to be inside the lock and since it has to be moved out of the handler closure.

tomerd:

PR #4238, opened 2022-03-19
[5.6] Allow binary targets to change checksums and URLs at the same time

keith:
Previously if you had a binary target, that you replaced the checksum
and URL for at the same time, SwiftPM would reject it because the
checksum had changed. In my experience this is the common case when you
are copy pasting the inclusion instructions from documentation for an
update and then you run `swift build`.

https://github.com/apple/swift-package-manager/pull/4233

### DIFF
--- a/Sources/Workspace/Workspace.swift
+++ b/Sources/Workspace/Workspace.swift
@@ -2115,14 +2115,15 @@ extension Workspace {
             ]
 
             if let existingArtifact = existingArtifact {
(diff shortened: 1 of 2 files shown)


keith:
I had to make some major changes to the tests here, but otherwise the patch is the same

keith:
@swift-ci please smoke test

tomerd (approved):

PR #4241, opened 2022-03-21
Fix a warning by correcting a `deprecated:` that should have been `introduced:`

abertelrud:
### Motivation:

An incorrect annotation from a recent change causes a warning.

### Modifications:

Change `deprecated:` to `introduced:` for a new enum value.

### Result:

Fix a warning (and correctly mark the symbol as not available in 5.7 and earlier).


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #4243, opened 2022-03-21
Annotate Cancellator APIs that involve Foundation.Process to exclude platforms that don't have it

abertelrud:
### Motivation:

Fix compilation failure on iOS, tvOS, watchOS.

### Modifications:

Exclude `Foundation.Process` functionality on platforms that don't have that type.

### Result:

Fix build failure.

### DIFF
--- a/Sources/Basics/Cancellator.swift
+++ b/Sources/Basics/Cancellator.swift
@@ -54,9 +54,11 @@ public class Cancellator: Cancellable {
         self.register(name: "\(process.arguments.joined(separator: " "))", handler:  process.terminate)
     }
 
+    #if !os(iOS) && !os(watchOS) && !os(tvOS)
     public func register(_ process: Foundation.Process) -> RegistrationKey? {
         self.register(name: "\(process.description)", handler: process.terminate(timeout:))
     }
+    #endif
 
     public func deregister(_ key: RegistrationKey) {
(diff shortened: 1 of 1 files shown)


abertelrud:
@swift-ci please smoke test

Sources/Basics/Cancellator.swift
```diff
@@ -54,9 +54,11 @@ public class Cancellator: Cancellable {
         self.register(name: "\(process.arguments.joined(separator: " "))", handler:  process.terminate)
     }
 
+    #if !os(iOS) && !os(watchOS) && !os(tvOS)
```
abertelrud, line 57: I'd prefer if there were a better way to do this, but I don't think there's a `#if defined(Foundation.Process)` or anything like that.

tomerd: does `#if canImport(Foundation.Process)` work?

abertelrud: Unfortunately not: `Unexpected platform condition argument: expected identifier`.  That would have been great!

Maybe this is worth a request on the compiler.  At the least the error seems wrong — it's not a platform condition.

tomerd (approved):

neonichu (approved):

PR #4245, opened 2022-03-22
ignore empty configuration files in legacy check

tomerd:
motivation: make legacy file warning less noisy if the file is empty

changes: check to see if the configuration file in the legacy location is empty before migrating it and emitting the warning


tomerd:
@swift-ci smoke test

yim-lee (approved):

PR #4246, opened 2022-03-22
[5.6] Package plugin dependency on executable product from different package doesn't find the tool (#4200)

abertelrud:
### Motivation

Fix a bug that caused a plugin with a dependency on an executable product from different package to not find the tool when looking it up with `context.tool(named:)`. Dependencies on executables in the same package as the plugin did work.

### Modifications

- add case to handle .product() in addition to .target() for dependencies built from executables
- add test fixture
- add unit test

### Scope of Issue

This affects tools that are needed by build tool plugins and that are defined in a different package than the plugin.

### Reason for Nominating to 5.6

This is a serious bug in new functionality (build tool plugins) that affects the user experience. It is also very safe, since the fix is in the new code that can only be executed when using this new functionality.

### Risk

Very low — this affects only the code path taken when dealing with a product dependency of a plugin on an executable target in a different package.

### Origination

This was a bug in the initial implementation of package command plugins.

### Automated Testing

A new unit test covers use of a product dependency in a different package from the plugin.

### Dependencies

None

### Impact on CI

None

### How to Verify

Check out the SwiftPM source code, and build the test fixture at `Fixtures/Miscellaneous/Plugins/LibraryWithLocalBuildToolPluginUsingRemoteTool`.

rdar://90650826

(cherry picked from commit 0384173437aac8695e53402a9598f06c36bdf64b)


tomerd (approved):

abertelrud:
@swift-ci please smoke test

PR #3079, opened 2020-11-24
Update the CLanguageStandard and CXXLanguageStandard enums

benrimmington:
Generate the enum cases by preprocessing the [LangStandards.def](https://github.com/apple/llvm-project/blob/apple/main/clang/include/clang/Basic/LangStandards.def) file.

### Motivation:

Follow-up to: apple/swift-package-manager#2716

### Modifications:

* Add enum cases for C17/C18, C2x, and C++20.
* Deprecate enum cases for C++1z.
* Remove enum cases for C++2a (which aren't in [release/5.3][] yet).
* Update the raw string value for C94 (bug fix).

### Result:

* New documentation comments.
* New cases for the latest C/C++ language standards.

[release/5.3]: <https://github.com/apple/swift-package-manager/blob/release/5.3/Sources/PackageDescription/LanguageStandardSettings.swift>

PR #4247, opened 2022-03-22
[Do Not Merge] Update DocC documentation in the `PackageDescription` library.

zkiraly:
**Update**
This PR has been split into two. Please continue reviewing in those replacement PRs
Part 1 https://github.com/apple/swift-package-manager/pull/4276
Part 2 https://github.com/apple/swift-package-manager/pull/4277

Update DocC documentation in the `PackageDescription` library.

### Motivation:

The DocC documentation in `PackageDescription` was not complete, and was out of sync with Apple's documentation at developer.apple.com.

### Modifications:

Many DocC documentation comments were added or changed inside `Sources/PackageDescription`.

### Result:

There was no code change. Only source code comments were changed in `Sources/PackageDescription`, and DocC extension files were added in the documentation catalog in that directory.

### DIFF
--- a/Sources/PackageDescription/LanguageStandardSettings.swift
+++ b/Sources/PackageDescription/LanguageStandardSettings.swift
@@ -169,6 +169,6 @@ public enum SwiftVersion {
 
     /// A user-defined value for the Swift version.
     ///
-    /// The value is passed as-is to the Swift compiler’s `-swift-version` flag.
+    /// The value is passed as-is to the Swift compiler's `-swift-version` flag.
     case version(String)
 }
--- a/Sources/PackageDescription/LanguageStandardSettings.swift
+++ b/Sources/PackageDescription/LanguageStandardSettings.swift
@@ -10,166 +10,161 @@
 //
 //===----------------------------------------------------------------------===//
 
-/// The supported C language standard to use for compiling C sources in the package.
-///
-/// Aliases are available for some C language standards. For example,
-/// use `c89`, `c90`, or `iso9899_1990` for the "ISO C 1990" standard.
-/// To learn more, read the [Clang Compiler User's Manual][1].
-///
-/// [1]: <https://clang.llvm.org/docs/UsersManual.html#differences-between-various-standard-modes>
+/// The supported C language standard to use for compiling C sources in the
+/// package.
 public enum CLanguageStandard: String, Encodable {
 
-    /// ISO C 1990.
+    /// The identifier for the C89 language standard.
     case c89
 
-    /// ISO C 1990.
+    /// The identifier for the C90 language standard.
     case c90
 
-    /// ISO C 1999.
+    /// The identifier for the C99 language standard.
     case c99
 
-    /// ISO C 2011.
+    /// The identifier for the C11 language standard.
     case c11
 
-    /// ISO C 2017.
+    /// The identifier for the C17 language stadard.
     @available(_PackageDescription, introduced: 5.4)
     case c17
 
-    /// ISO C 2017.
+    /// The identifier for the C18 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case c18
 
-    /// Working Draft for ISO C2x.
+    /// The identifier for the C2x draft language standard.
     @available(_PackageDescription, introduced: 5.4)
     case c2x
 
-    /// ISO C 1990 with GNU extensions.
+    /// The identifier for the GNU89 language standard.
     case gnu89
 
-    /// ISO C 1990 with GNU extensions.
+    /// The identifier for the GNU90 language standard.
     case gnu90
 
-    /// ISO C 1999 with GNU extensions.
+    /// The identifier for the GNU99 language standard.
     case gnu99
 
-    /// ISO C 2011 with GNU extensions.
+    /// The identifier for the GNU11 language standard.
     case gnu11
 
-    /// ISO C 2017 with GNU extensions.
+    /// The identifier for the GNU17 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case gnu17
 
-    /// ISO C 2017 with GNU extensions.
+    /// The identifier for the GNU18 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case gnu18
 
-    /// Working Draft for ISO C2x with GNU extensions.
+    /// The identifier for the GNU2x draft language standard.
     @available(_PackageDescription, introduced: 5.4)
     case gnu2x
 
-    /// ISO C 1990.
+    /// The identifier for the ISO9899-1990 language standard.
     case iso9899_1990 = "iso9899:1990"
 
-    /// ISO C 1990 with amendment 1.
+    /// The identifier for the ISO9899-199409 language standard.
     case iso9899_199409 = "iso9899:199409"
 
-    /// ISO C 1999.
+    /// The identifier for the ISO9899-1999 language standard.
     case iso9899_1999 = "iso9899:1999"
 
-    /// ISO C 2011.
+    /// The identifier for the ISO9899-2011 language standard.
     case iso9899_2011 = "iso9899:2011"
 
-    /// ISO C 2017.
+    /// The identifier for the ISO9899-2017 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case iso9899_2017 = "iso9899:2017"
 
-    /// ISO C 2017.
+    /// The identifier for the ISO9899-2018 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case iso9899_2018 = "iso9899:2018"
 }
 
-/// The supported C++ language standard to use for compiling C++ sources in the package.
-///
-/// Aliases are available for some C++ language standards. For example,
-/// use `cxx98` or `cxx03` for the "ISO C++ 1998 with amendments" standard.
-/// To learn more, read the [C++ Support in Clang][1] status page.
-///
-/// [1]: <https://clang.llvm.org/cxx_status.html>
+/// The supported C++ language standard to use for compiling C++ sources in the
+/// package.
 public enum CXXLanguageStandard: String, Encodable {
 
-    /// ISO C++ 1998 with amendments.
+    /// The identifier for the C++98 language standard.
     case cxx98 = "c++98"
 
-    /// ISO C++ 1998 with amendments.
+    /// The identifier for the C++03 language standard.
     case cxx03 = "c++03"
 
-    /// ISO C++ 2011 with amendments.
+    /// The identifier for the C++11 language standard.
     case cxx11 = "c++11"
 
-    /// ISO C++ 2014 with amendments.
+    /// The identifier for the C++14 language standard.
     case cxx14 = "c++14"
 
-    /// ISO C++ 2017 with amendments.
+    /// The identifier for the ISO C++ 2017 (with amendments) language standard..
     @available(_PackageDescription, introduced: 5.4)
     case cxx17 = "c++17"
 
-    /// ISO C++ 2017 with amendments.
+    /// The identifier for the C++1z language standard.
     @available(_PackageDescription, introduced: 4, deprecated: 5.4, renamed: "cxx17")
     case cxx1z = "c++1z"
 
-    /// ISO C++ 2020 DIS.
+    /// The identifier for the ISO C++ 2020 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case cxx20 = "c++20"
 
-    /// Working draft for ISO C++ 2023 DIS.
+    /// The identifier for the ISO C++ 2023 draft language standard.
     @available(_PackageDescription, introduced: 5.6)
     case cxx2b = "c++2b"
 
-    /// ISO C++ 1998 with amendments and GNU extensions.
+    /// The identifier for the GNU++98 language standard.
     case gnucxx98 = "gnu++98"
 
-    /// ISO C++ 1998 with amendments and GNU extensions.
+    /// The identifier for the GNU++03 language standard.
     case gnucxx03 = "gnu++03"
 
-    /// ISO C++ 2011 with amendments and GNU extensions.
+    /// The identifier for the GNU++11 language standard.
     case gnucxx11 = "gnu++11"
 
-    /// ISO C++ 2014 with amendments and GNU extensions.
+    /// The identifier for the GNU++14 language standard.
     case gnucxx14 = "gnu++14"
 
-    /// ISO C++ 2017 with amendments and GNU extensions.
+    /// The identifier for the GNU++17 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case gnucxx17 = "gnu++17"
 
-    /// ISO C++ 2017 with amendments and GNU extensions.
+    /// The identifier for the GNU++1z language standard.
     @available(_PackageDescription, introduced: 4, deprecated: 5.4, renamed: "gnucxx17")
     case gnucxx1z = "gnu++1z"
 
-    /// ISO C++ 2020 DIS with GNU extensions.
+    /// The identifier for the CNU++20 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case gnucxx20 = "gnu++20"
 
-    /// Working draft for ISO C++ 2023 DIS with GNU extensions.
+    /// The identifier for the GNU++2b draft language standard.
     @available(_PackageDescription, introduced: 5.6)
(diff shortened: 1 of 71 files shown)


Sources/PackageDescription/LanguageStandardSettings.swift
```diff
@@ -10,166 +10,161 @@
...
+    /// The identifier for the GNU11 language standard.
     case gnu11
 
-    /// ISO C 2017 with GNU extensions.
+    /// The identifier for the GNU17 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case gnu17
 
-    /// ISO C 2017 with GNU extensions.
+    /// The identifier for the GNU18 language standard.
```
abertelrud, line 57: Is the change in wording based on the identifier name?  I think the thought was that the identifier name should be brief but the help text should explain the nuance about it being extensions on top of a standard.  Same comment applies to the others in this enum.

Sources/PackageDescription/LanguageStandardSettings.swift
```diff
@@ -10,166 +10,161 @@
...
 
-/// The supported C++ language standard to use for compiling C++ sources in the package.
-///
-/// Aliases are available for some C++ language standards. For example,
-/// use `cxx98` or `cxx03` for the "ISO C++ 1998 with amendments" standard.
-/// To learn more, read the [C++ Support in Clang][1] status page.
-///
-/// [1]: <https://clang.llvm.org/cxx_status.html>
+/// The supported C++ language standard to use for compiling C++ sources in the
+/// package.
```
abertelrud, line 87: These settings can be confusing so it might be beneficial to keep the middle paragraph and reference to the Clang documentation.

Sources/PackageDescription/Version.swift
```diff
@@ -12,31 +12,31 @@
...
 ///
-/// A package version is a three period-separated integer, for example `1.0.0`. It must conform to the semantic versioning standard in order to ensure
+/// A package version consists of three integers separated by periods, for example `1.0.0`. It must conform to the semantic versioning standard in order to ensure
 /// that your package behaves in a predictable manner once developers update their
 /// package dependency to a newer version. To achieve predictability, the semantic versioning specification proposes a set of rules and
 /// requirements that dictate how version numbers are assigned and incremented. To learn more about the semantic versioning specification, visit
-/// [semver.org](www.semver.org).
+/// [Semantic Versioning 2.0.0](www.semver.org).
 ///
 /// **The Major Version**
```
WowbaggersLiquidLunch, line 21: unintentional leftover heading?

Sources/PackageDescription/Version.swift
```diff
@@ -57,7 +57,7 @@ public struct Version {
     /// Initializes a version struct with the provided components of a semantic version.
     ///
     /// - Parameters:
-    ///   - major: The major version number.
+    ///   - major: The major version numner.
```
WowbaggersLiquidLunch, line 60: typo

Sources/PackageDescription/Version.swift
```diff
@@ -96,11 +96,33 @@ public struct Version {
...
+    /// Returns a Boolean value indicating whether the value of the first
+    /// argument is less than that of the second argument.
+    ///
+    /// This function is the only requirement of the `Comparable` protocol. The
+    /// remainder of the relational operator functions are implemented by the
+    /// standard library for any type that conforms to `Comparable`.
+    ///
+    /// - Parameters:
+    ///   - lhs: A value to compare.
+    ///   - rhs: Another value to compare.
```
WowbaggersLiquidLunch, line 125: I think these are unnecessary, since the documentation should automatically inherit from `Equatable`'s and `Comparable`'s. However, it might be beneficial to replace the generic explanations of `==` and `<` with ones specific to semantic versions, with notes on things that users should watch for (e.g. build metadata identifiers, numeric vs. alphanumeric identifier precedence, pre-release precedence, etc).

Sources/PackageDescription/Version+StringLiteralConvertible.swift
```diff
@@ -12,7 +12,9 @@
 
 extension Version: ExpressibleByStringLiteral {
     /// Initializes a version struct with the provided string literal.
-    /// - Parameter version: A string literal to use for creating a new version struct.
+    ///
+    /// - Parameters:
+    ///   - version: A string literal to use for creating a new version struct.
```
WowbaggersLiquidLunch, line 17: I think the current style where single paramenter uses `- Parameter parameter: description` is better, because its compactness reduces visual clatter around the code. However, it's only a matter of preference.

Sources/PackageDescription/Target.swift
```diff
@@ -14,10 +14,12 @@ import Foundation
 
 /// A target, the basic building block of a Swift package.
 ///
-/// Each target contains a set of source files that are compiled into a module or test suite.
-/// You can vend targets to other packages by defining products that include the targets.
+/// Each target contains a set of source files that Swift Package Manager compiles into a module
```
WowbaggersLiquidLunch, line 17: As far as I understand it, SwiftPM does not compile any code. Instead, it tells the driver and compiler what to compile and how to compile them.

Sources/PackageDescription/SupportedPlatforms.swift
```diff
@@ -10,8 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-/// A platform that usually corresponds to an operating system such as
-/// iOS, macOS, or Linux.
+/// A platform that corresponds to an operating system such as iOS,
+/// macOS, or Linux.
```
WowbaggersLiquidLunch, line 14: A platform is not _always_ an OS, non-OS examples listed in the type include `.driverKit` and `.wasi`.

Sources/PackageDescription/SupportedPlatforms.swift
```diff
@@ -95,23 +100,27 @@ public struct SupportedPlatform: Encodable, Equatable {
         self.version = version
     }
 
-    /// Configures the minimum deployment target version for the macOS platform.
+    /// Configure the minimum deployment target version for the macOS platform.
```
WowbaggersLiquidLunch, line 103: The existing convention is using third-person present tense for a function's description. E.g. https://developer.apple.com/documentation/swift/array/1641390-remove

Sources/PackageDescription/PackageRequirement.swift
```diff
@@ -215,18 +230,29 @@ extension Range {
...
 }
 
 @available(_PackageDescription, deprecated: 5.6)
 extension Package.Dependency.Requirement {
+    /// Returns a requirement for a version range, starting at the given minimum
+    /// version and going up to but not including the next major version. This is the recommended
+    /// version requirement.
+    ///
+    /// - Parameters:
+    ///   - version: The minimum version for the version range.
```
WowbaggersLiquidLunch, line 246: It's [recommended](https://www.swift.org/documentation/api-design-guidelines/#write-doc-comment) to begin with a summary consisting only of a single sentence fragment. I would suggest something like this:

```
/// Returns a source control requirement bounded to the given version's major version number.
///
/// Additional descriptions and explanations here.
/// - Parameter version: The given version to bound the source control requirement to.
/// - Returns: A source control requirement bounded to `version`'s major version number.
```

Sources/PackageDescription/PackageRequirement.swift
```diff
@@ -192,10 +205,12 @@ extension Package.Dependency {
     /// versions, and may require you to modify your code when they update.
     /// The version rule requires Swift packages to conform to semantic
     /// versioning. To learn more about the semantic versioning standard,
-    /// visit [semver.org](https://semver.org).
+    /// visit the [Semantic Versioning 2.0.0](https://semver.org) website.
```
WowbaggersLiquidLunch, line 208: I think it's better to either leave it as it currently is, or update the URL to `https://semver.org/spec/v2.0.0.html`

Sources/PackageDescription/PackageRequirement.swift
```diff
@@ -192,10 +205,12 @@ extension Package.Dependency {
...
     @available(_PackageDescription, introduced: 999)
     public enum RegistryRequirement {
+        /// A version based requirement.
```
WowbaggersLiquidLunch, line 211: Since the other case is also a version-based requirement, perhaps it's more accurate to highlight the exact-ness of this case.

Sources/PackageDescription/PackageDescriptionSerialization.swift
```diff
@@ -429,6 +495,16 @@ extension SwiftVersion: Encodable {
 }
 
 extension Version: Encodable {
+    /// Encodes this value into the given encoder.
+    ///
+    /// If the value fails to encode anything, `encoder` will encode an empty
+    /// keyed container in its place.
```
WowbaggersLiquidLunch, line 501: I'm not sure if this description is correct.

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -31,46 +31,56 @@ import Foundation
...
+///         .testTarget(name: "MyLibraryTests", dependencies: ["MyLibrary"])
+///     ]
+/// )
+/// ```
 ///
-/// The package manifest must begin with the string `// swift-tools-version:`,
-/// followed by a version number such as `// swift-tools-version:5.3`.
+/// The package manifest must begin with the string `// swift-tools-version:`
+/// followed by a version number specifier. The following code listing shows a
+/// few examples of valid declarations of the Swift tools version:
```
WowbaggersLiquidLunch, line 58: This has rule has been relaxed since Swift 5.4. Manifest files for Swift ≥ 5.4 no long _must_ begin with "// swift-tools-version:" followed immediately by the version string. They can contain any combination of whitespace before and after "//" and after ":". 

Also the "swift-tools-version" part has always been case-insensitive, but the convention is all-lowercase.

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -22,7 +22,7 @@ import Foundation
 
 /// The configuration of a Swift package.
 ///
-/// Pass configuration options as parameters to your package's initializer
+/// Pass configuration options as parameters to your package’s initializer
```
WowbaggersLiquidLunch, line 25: Although it's correct (page 270 of ) to use U+2019 instead of U+0027 for punctuation, it can be more difficult to edit it using many IDE and keyboard combinations. (see page 270 of [Unicode 13.0 spec](https://www.unicode.org/versions/Unicode13.0.0/ch06.pdf) which outlines the dilemma)

Sources/PackageDescription/Target.swift
abertelrud: My view is that that distinction is probably unnecessary to make for users of SwiftPM.

Sources/PackageDescription/PackageDescriptionSerialization.swift
abertelrud: I think this whole file is also an internal implementation details, i.e. manifests aren't expected to serialize anything.  If this is documented, the impression is that manifests should do that.  It's just that I don't think there is a way to hide the public codability conformance from the API.

Sources/PackageDescription/LanguageStandardSettings.swift
zkiraly: I see what you mean. This is difficult, because [GNU says](https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html#C-Dialect-Options) that both `gnu17` and `gnu18` are "GNU dialect of ISO C17." 

Anyway, I was following the pattern of the [currently published documentation](https://developer.apple.com/documentation/swift_packages/clanguagestandard), which was reviewed and approved.

Sources/PackageDescription/LanguageStandardSettings.swift
zkiraly: That's reasonable, but unfortunately that URL is dead.

zkiraly: Yes, I agree. I'll put that back.

Sources/PackageDescription/Version.swift
zkiraly: Probably... thanks for pointing this out!

Sources/PackageDescription/Version.swift
zkiraly: Thank you!

Sources/PackageDescription/Version.swift
zkiraly: Good point, definitely for the `<` function, I agree, because it's completely adapted for the purpose. I'll describe that in more detail. The others are more generic. I'll think about the `Equatable` conformance documentation, but for now, I am inclined to leave as-is.

WowbaggersLiquidLunch:
I wonder if it might be better to divide this PR into multiple smaller PRs, each focusing on just 1 or a few related files? There are a lot of lines changed in this PR. Smaller PRs can be easier (and perhaps faster) to review and prevent things from being overlooked.

Sources/PackageDescription/Version.swift
WowbaggersLiquidLunch: Some of the description in `==`'s documentation doesn't make sense. For example, this line: "This is the default implementation of the equal-to operator (`==`) for any type that conforms to `Equatable`.", since `Version` is not using the default implementation of `==`.

In general, I suggest that it's better to write the documentation for each function (not just these 2, but all functions in this PR) from scratch, instead of using the protocol's default implementation's, for the following reason:

1. If the conforming type is _not_ using the default implementation, then the documentation for the default implementation likely is incorrect for the custom implementation.

2. If the conforming type _is_ using the default implementation, then there is really no need to reuse the same documentation, because the same documentation should be automatically inherited.

Sources/PackageDescription/Target.swift
WowbaggersLiquidLunch: That's fair, and perhaps SwiftPM can be considered as the compiler frontend's frontend.

Anyway, I think the current wording is better than the proposed change, since the passive voice hides the distinction, so it's both correct and friendly to the user.

Sources/PackageDescription/LanguageStandardSettings.swift
WowbaggersLiquidLunch: This is interesting, because #3079 changed the documentation from what's shown on Apple's documentation page. So, the documentation page is at least outdated by about 16 months.

Sources/PackageDescription/Target.swift
zkiraly: Yes. This documentation is aimed at developers who use SwiftPM, and not at developers working on SwiftPM.

Sources/PackageDescription/SupportedPlatforms.swift
zkiraly: How about this: 
> A platform supported by Swift Package Manager.

Sources/PackageDescription/PackageRequirement.swift
zkiraly: I went with the second option, and I changed it in other places in `PackageDescription` also.

Sources/PackageDescription/PackageDescription.swift
zkiraly: That sounds too loosy-goosey to me. I'd rather not see people do

Sources/PackageDescription/PackageDescription.swift
zkiraly: This was due to my internal process, and I thank you for calling it out.

WowbaggersLiquidLunch: There are other locations where U+0027 is changed to U+2019. If the changes are not intended, then they should be changed back, too.

Sources/PackageDescription/PackageDescription.swift
WowbaggersLiquidLunch: We can mention the relaxation around whitespace without having to go to the extreme in recommending the format as shown in your example. One of the benefits is for the user to have more control with comments' alignment and format in their entire project. For example, I can imagine some users might prefer to have all their comments indented by some whitespace, starting with an uppercase letter, or aligned with doc comments by using 1 more spaces or a tab after "//". Providing them with this information allows them to treat the Swift version specification as a regular comment in their source code, and format it as they prefer.

Sources/PackageDescription/PackageDescription.swift
zkiraly: I changed all of them to the straight apostrophe under `PackageDescription`.

zkiraly:
The content was merged from the two split PRs. No further work on this PR. Closing.

PR #4251, opened 2022-03-23
add a test case to make sure binary artifact is deleted when download fails

tomerd:
motivation: add test coverage for a recently fixed issue

changes: add a test case for when binary dependency downlod fails and leaves behind a stray artifact which should be deleted


tomerd:
@swift-ci smoke test

tomerd:
cc @keith

abertelrud (approved):

neonichu (approved):

keith:
Thank you!

PR #4258, opened 2022-03-25
repair the build after #4255, #4256

compnerd:
Repair the Windows build after #4255 and #4256 changed things without
testing on Windows.

### Motivation:

### Modifications:

### Result:


tomerd:
cc @neonichu - lets merge this to unbar the Windows build and then find a way to make ToolchainConfiguration private again

abertelrud (approved):

PR #4260, opened 2022-03-25
Repair #4258

compnerd:
The original change was the most expedient change to get the CI
coverage restored.  After further investigation, it seems that we are
able to reconstruct the XCTest path through a public variable that has
appeared after the original Windows support was added.  This now allows
us to restore the configuration to its previous state of `private`.

### Motivation:

### Modifications:

### Result:


compnerd:
@swift-ci please smoke test

neonichu (approved):

tomerd (approved):

PR #4263, opened 2022-03-26
update example to use async await APIs

tomerd:
motivation: async / await APIs are great, use them as example!

changes:
* expose some of workspace public APIs as async/await ones (in addition to the callback ones)
* update the example to use @main and call the async / await APIs


tomerd:
@swift-ci smoke test

Examples/package-info/Package.swift
```diff
@@ -1,22 +1,25 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.5
 
 import PackageDescription
 
 let package = Package(
     name: "package-info",
     platforms: [
-        .macOS(.v10_15),
+        .macOS("10.15.4"),
```
abertelrud, line 8: Since this was an interesting minimum deployment to target (due to some Foundation API improvements), should we add this as an enum?  I know we don't usually do that for patch releases but I don't think that's policy, and this one in particular has semantic use.

abertelrud (approved):

Examples/package-info/Package.swift
tomerd: this was just to align it to the version used in SwiftPM, not  sure the genesis

tomerd: maybe we should go to latest instead?

abertelrud: Ah, I see.  I don't think it matters which one we use here for the example — latest would be fine, I think.

It just occurred to me that others wanting the Swiftier Foundation FileManager APIs would want 10.15.4 or newer, and that might justify a new enum case.  Maybe we don't yet have enough of them — with an index that would be easy to scan for.

tomerd:
@swift-ci smoke test

tomerd:
@swift-ci smoke test

Examples/package-info/Package.swift
tomerd: no strong option, but going as high as possible is good to make use of new APIs, best practices etc

neonichu (approved):

PR #4265, opened 2022-03-28
PackageModel: pass along `-sdk` property on Windows

compnerd:
If the user specifies `-sdk` always prefer that over the `SDKROOT`
environment variable.  We would previously honor that but then fail to
pass along the argument.  We never noticed this as `SDKROOT` has always
been set up until this point.  With newer SDKs, this will become less a
guarantee.

### Motivation:

### Modifications:

### Result:


neonichu (approved):

abertelrud (approved):

compnerd:
Checked with @abertelrud - good to go!

PR #4266, opened 2022-03-28
PackageModel: adjust toolchain layout for Windows

compnerd:
Add in a migration path for adjusting the layout of the toolchain.  This
is required in order to support more than one architecture SDK to be
installed simultaneously.  The problem is primarily on the XCTest side,
where the binary is not slotted to an architecture.  Having multiple
variants of XCTest is difficult with the toolchain layout.  The XCTest
installation is located at
`%DEVELOPER_DIR%\Platforms\<Platform>.platform\Library\XCTest-<version>`.
The runtime component is located under, and the DLL is always named
`XCTest.dll`.  The following options were considered:

1. renaming `XCTest.dll` to:
  - `XCTest32.dll` (x86)
  - `XCTest64.dll` (x64)
  - `XCTest32a.dll` (ARMv7)
  - `XCTest64a.dll` (ARM64)

2. renaming `bin` to:
  - `bin32` (x86)
  - `bin64` (x64)
  - `bin32a` (ARMv7)
  - `bin64a` (ARM64)

3. adding a subdirectory to `bin`

Option 2 was deemed to be the best of the terrible options.  It would
allow a migration path (as the DLL name is embedded into the import
library and subsequently the executable).

We simultaneously update the location for the import library.  Although
it is possible to build a multi-architecture import library, we would
need generate that manually before the installer package is built and
then only install the runtime.  It seems that moving the directory would
be easier and ensure that we can keep them somewhat independent.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -261,10 +261,27 @@ public final class UserToolchain: Toolchain {
                             root.appending(component: "Developer")
                                 .appending(component: "Library")
                                 .appending(component: "XCTest-\(info.defaults.xctestVersion)")
+
                             xctest = [
                                 "-I", AbsolutePath("usr/lib/swift/windows/\(triple.arch)", relativeTo: path).pathString,
-                                "-L", AbsolutePath("usr/lib/swift/windows", relativeTo: path).pathString,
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

neonichu (approved):

abertelrud (approved):

compnerd:
Checked with @abertelrud - good to go.

PR #4267, opened 2022-03-28
Time tests for display in xunit report

Austinpayne:
Time tests for display in xunit report

### Motivation:

It is useful to see simple reports on how long tests take to complete so a proper time duration has been added to the xunit report.

### Modifications:

Replaced hardcoded time of `0.0` in xunit reports with proper time duration for each test run.

### Result:

Xunit reports now report an accurate time duration for each testcase and an aggregate duration for each testsuite.

### DIFF
--- a/Sources/Commands/SwiftTestTool.swift
+++ b/Sources/Commands/SwiftTestTool.swift
@@ -13,6 +13,7 @@
 import ArgumentParser
 import Basics
 import Build
+import struct Foundation.Date
 import class Foundation.ProcessInfo
 import PackageGraph
 import PackageModel
@@ -651,6 +652,7 @@ final class ParallelTestRunner {
         var unitTest: UnitTest
         var output: String
         var success: Bool
+        var time: Double
     }
 
     /// Path to XCTest binaries.
@@ -767,11 +769,18 @@ final class ParallelTestRunner {
                         observabilityScope: self.observabilityScope
                     )
                     var output = ""
+                    let start = Date()
                     let success = testRunner.test(outputHandler: { output += $0 })
+                    let end = Date()
                     if !success {
                         self.ranSuccessfully = false
                     }
-                    self.finishedTests.enqueue(TestResult(unitTest: test, output: output, success: success))
+                    self.finishedTests.enqueue(TestResult(
+                        unitTest: test,
+                        output: output,
+                        success: success,
+                        time: end.timeIntervalSince(start)
+                    ))
                 }
             }
             thread.start()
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
Thanks @Austinpayne this is great!  I'm not that familiar with the underlying code here, but it definitely seems like the right fix.  Are there unit tests that should/could be updated as well?

Sources/Commands/SwiftTestTool.swift
```diff
@@ -767,11 +769,18 @@ final class ParallelTestRunner {
                         observabilityScope: self.observabilityScope
                     )
                     var output = ""
+                    let start = Date()
                     let success = testRunner.test(outputHandler: { output += $0 })
+                    let end = Date()
```
WowbaggersLiquidLunch, line 774: Is it possible to use the new [clock APIs](https://github.com/apple/swift-evolution/blob/main/proposals/0329-clock-instant-duration.md) instead of pulling from Foundation?

abertelrud: That looks very promising, but as far as I can tell, it's been accepted but not yet implemented. Since it doesn't affect API of packages, would it make sense to take this PR using the existing facilities but then switch to `SuspendingClock` when it's available?

abertelrud: I suggest we take this PR as coded and then switch to a more accurate timekeeping when the API is available in all versions where we want to compile SwiftPM.

abertelrud (approved):

Sources/Commands/SwiftTestTool.swift
WowbaggersLiquidLunch: It has be [implemented in the trunk](https://github.com/apple/swift/blob/541f7725e1b0f4ac2cfc1f1b71a90c5a5c813c1d/stdlib/public/Concurrency/SuspendingClock.swift).

Your suggestion makes sense.

Sources/Commands/SwiftTestTool.swift
```diff
@@ -651,6 +652,7 @@ final class ParallelTestRunner {
         var unitTest: UnitTest
         var output: String
         var success: Bool
+        var time: Double
```
tomerd, line 655: we typically use DispatchTimeInterval for this purpose

tomerd (commented):
thanks for the PR!

we typically use Dispatch time/duration data types for time measurements 

are there unit tests we should update to verify the new duration field is present?

Sources/Commands/SwiftTestTool.swift
abertelrud: That's a good point.  @Austinpayne there is some code in https://github.com/apple/swift-package-manager/blob/5f5be413e8741aa91ba52d32580d77f5916aeb79/Sources/SPMBuildCore/PluginInvocation.swift#L432 that uses DispatchTime for plugins and then computes the interval that may be useful to look at.  There are then some functions in `DispatchTimeInterval+Extensions.swift` to convert that to other types.

Austinpayne:
Thank you for the feedback @tomerd and @abertelrud! Apologies for the delay in getting in your updates. Please let me know if you would like more changes or if you would like me to squash the new commits.

abertelrud:
@swift-ci please smoke test

tomerd (approved):

tomerd:
thanks @Austinpayne

PR #4269, opened 2022-03-29
Restore the doc of `package(name:exact:)`

kateinoigakukun:
Restore the doc of package(name:exact:) to sync with its actual semantics

### Motivation:

It seems the doc comment was unintentionally taken from
`package(name:from:)`. This happend in 8dd0202d01f0d644a9d0a12fb039892d8828c240

### Modifications:

Sync the documentation with `Requirement.exact`, which has the same semantics.
https://github.com/apple/swift-package-manager/blob/8dd0202d01f0d644a9d0a12fb039892d8828c240/Sources/PackageDescription/PackageRequirement.swift#L71-L76

### Result:

### DIFF
--- a/Sources/PackageDescription/PackageDependency.swift
+++ b/Sources/PackageDescription/PackageDependency.swift
@@ -412,12 +412,10 @@ extension Package.Dependency {
 
     /// Adds a package dependency that uses the exact version requirement.
     ///
-    /// This is the recommended way to specify a remote package dependency.
-    /// It allows you to specify the minimum version you require, allows updates that include bug fixes
-    /// and backward-compatible feature updates, but requires you to explicitly update to a new major version of the dependency.
-    /// This approach provides the maximum flexibility on which version to use,
-    /// while making sure you don't update to a version with breaking changes,
-    /// and helps to prevent conflicts in your dependency graph.
+    /// Specifying exact version requirements are not recommended as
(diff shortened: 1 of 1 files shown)


kateinoigakukun:
@swift-ci Please smoke test

abertelrud:
Good catch; thank you!  It looks as if there is actually one more of these, on line 508.  Would you mind updating the PR with that one as well?  Thanks!

abertelrud:
Good catch; thank you!  It looks as if there is actually one more of these, on line 508.  Would you mind updating the PR with that one as well?  Thanks!

abertelrud:
(not sure why my comment appeared twice, probably a web glitch)

kateinoigakukun:
@abertelrud Good catch 👍

kateinoigakukun:
@swift-ci Please test

tomerd:
@swift-ci smoke test

abertelrud (approved):
Thanks for fixing this @kateinoigakukun, this looks great to me now!

abertelrud:
@zkiraly Just FYI this may conflict with https://github.com/apple/swift-package-manager/pull/4276 but is the correct thing to do as it fixes a mistake.

zkiraly:
> @zkiraly Just FYI this may conflict with #4276 but is the correct thing to do as it fixes a mistake.

Thanks for the note, I'll take this into account.

PR #4276, opened 2022-03-31
Part 1 of the PackageDescription DocC documentation update

zkiraly:
Part 1: Update DocC documentation in the PackageDescription library. Everything, except `Target`, `Product`, and symbols curated under those two.

### Motivation:

The DocC documentation in PackageDescription was not complete, and was out of sync with (and in some cases ahead of) Apple's documentation at developer.apple.com.

### Modifications:

Many DocC documentation comments were added or changed inside Sources/PackageDescription, except for `Target`, `Product`, and symbols curated under those two. The excluded symbols are in a companion PR.

### Result:

There was no code change. Only source code comments were changed in Sources/PackageDescription, and DocC extension files were added in the documentation catalog in that directory.


Sources/PackageDescription/LanguageStandardSettings.swift
```diff
@@ -10,161 +10,160 @@
...
+    /// The identifier for the ISO9899-2011 language standard.
     case iso9899_2011 = "iso9899:2011"
 
-    /// ISO C 2017.
+    /// The identifier for the ISO9899-2017 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case iso9899_2017 = "iso9899:2017"
 
-    /// ISO C 2017.
+    /// The identifier for the ISO9899-2018 language standard.
```
WowbaggersLiquidLunch, line 81: As pointed out by @abertelrud in #4247, these shouldn't be changed as proposed. The documentation published on apple.com is outdated one, so it shouldn't be the basis for changing the current documentation.

Sources/PackageDescription/LanguageStandardSettings.swift
```diff
@@ -10,161 +10,160 @@
...
     @available(_PackageDescription, introduced: 4, deprecated: 5.4, renamed: "gnucxx17")
     case gnucxx1z = "gnu++1z"
 
-    /// ISO C++ 2020 DIS with GNU extensions.
+    /// The identifier for the CNU++20 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case gnucxx20 = "gnu++20"
 
-    /// Working draft for ISO C++ 2023 DIS with GNU extensions.
+    /// The identifier for the GNU++2b draft language standard.
```
WowbaggersLiquidLunch, line 146: Similarly, these shouldn't be changed as proposed.

Sources/PackageDescription/LanguageStandardSettings.swift
```diff
@@ -10,161 +10,160 @@
...
 
+    /// The identifier for the Swift 4 language version.
     @available(_PackageDescription, introduced: 4)
     case v4
 
+    /// The identifier for the Swift 4.2 language version.
     @available(_PackageDescription, introduced: 4)
     case v4_2
 
+    /// The identifier for the Swift 5 language version.
```
WowbaggersLiquidLunch, line 166: In my opinion, it's easier to read if the documentation for each case is just "Swift 3", "Swift 4.2", etc. Most likely, a user reading it already knows that the symbols are identifiers for Swift language versions, so it's probably better to avoid information overload.

Sources/PackageDescription/PackageDependency.swift
```diff
@@ -20,24 +20,43 @@ extension Package {
...
+    /// package can use. The `Package.resolved` file records the results of the
+    /// dependency resolution and lives in the top-level directory of a Swift
+    /// package. If you add the Swift package as a package dependency to an app
+    /// for an Apple platform, you can find the `Package.resolved` file inside
+    /// your `.xcodeproj` or `.xcworkspace`.
     public class Dependency: Encodable {
+        /// The kind of dependency.
         @available(_PackageDescription, introduced: 5.6)
         public enum Kind {
+            /// A dependency based on a file system path.
```
WowbaggersLiquidLunch, line 34: Maybe "found at" or "located at" is more accurate than "based on"?

Additionally, the convention seems to prefix the word "given" to any parameters mentioned in the summary. For example, in this specific case, 

```
/// A dependency located at the given path.
/// - Parameters:
///   - name: ...
///   - path: ...
```

The same applies to the other cases.

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -27,50 +27,60 @@ import Foundation
 /// dependencies, and other configuration options.
 ///
 /// By convention, you need to define the properties of a package in a single
-/// nested initializer statement. Don’t modify it after initialization. The
+/// nested initializer statement. Don't modify it after initialization. The
```
WowbaggersLiquidLunch, line 30: As mentioned previously, U+2019 is the correct character for punctuation, but U+0027 is more convenient. I don't hold a strong opinion to either one, other than that in general, the choice should be consistent in the entire package.

Sources/PackageDescription/PackageDescription.docc/PackageDescription.md
```diff
@@ -1,46 +1,50 @@
...
+                .process("Resources/")
+            ]
+        )
+    ]
+)
+```
+
+The package manifest also allows you to define executable products, as well as plugins that Swift Package Manager can use to build other products in the manifest.
+
+For more information about adding a package dependency to your app project and creating Swift packages with Xcode, see [Adding Package Dependencies to Your App](https://developer.apple.com/documentation/xcode/adding_package_dependencies_to_your_app), [Creating a Standalone Swift Package with Xcode](https://developer.apple.com/documentation/xcode/creating_a_standalone_swift_package_with_xcode/), and [Swift Packages](https://developer.apple.com/documentation/xcode/swift_packages).
```
WowbaggersLiquidLunch, line 41: I think it's best not to rely on any website or example that's closed-source or not hosted by swift.org, unless there is no other choice.

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -27,50 +27,60 @@ import Foundation
...
+/// followed by a version number specifier. The following code listing shows a
+/// few examples of valid declarations of the Swift tools version:
 ///
-/// The Swift tools version declares:
+/// ```swift
+/// // swift-tools-version:3.0.2
+/// // swift-tools-version:3.1
+/// // swift-tools-version:4.0
+/// // swift-tools-version:5.3
+/// ```
```
WowbaggersLiquidLunch, line 65: As mentioned previously, I think we should add a sentence telling the user that the whitespace rule is relaxed since 5.4, without encouraging any extreme or impractical format. Maybe the tone should be softened a little too, so it doesn't give the user the impression that there is only one accepted format.

Additionally, it might be a good idea to include a few examples for Swift ≥ 5.4, because with the relaxed whitespace parsing rule, new generated tools version specification lines now have a space between ":" and the version specifier for Swift ≥ 5.4.

Additionally,

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -88,11 +98,11 @@ public final class Package {
...
     /// The name to use for C modules.
     ///
-    /// If present, the Swift Package Manager searches for a `<name>.pc` file
-    /// to get the required additional flags for a system target.
+    /// If present, the Swift Package Manager searches for a `<name>.pc` file to
+    /// get the required additional flags for a system target.
     public var pkgConfig: String?
 
-    /// An array of providers for a system target.
+    /// An array of providers for the system target.
```
WowbaggersLiquidLunch, line 105: I think it's correct to use the indefinite article "a" here, because there isn't a particular system target that is referred to in context.

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -309,13 +318,17 @@ public final class Package {
...
+/// [RFC5646](https://tools.ietf.org/html/rfc5646).
 public struct LanguageTag: Hashable {
 
     /// An IETF language tag.
     public let tag: String
 
     /// Creates a language tag from its IETF string representation.
+    ///
+    /// - Parameters:
+    ///   - tag: The string representation of an IETF language tag.
```
WowbaggersLiquidLunch, line 331: I think it's better to use the compact "- Parameter <parameter>: <description>" form to document single-parameter functions.

Same for the few instances below in the same file.

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -324,44 +337,68 @@ public struct LanguageTag: Hashable {
...
+    /// value, this initializer returns `nil`.
+    ///
+    /// - Parameters:
+    ///   - rawValue: The raw value to use for the new instance.
     public init?(rawValue: String) {
         tag = rawValue
     }
 }
 
+/// ExpressibleByStringLiteral implementation.
```
WowbaggersLiquidLunch, line 352: I don't think docc is able to generate documentation for this line. Alternatively, you can use something like

```
// MARK: ExpressibleByStringLiteral Conformance
```

if the intention is to help a reader find this extension quickly.

Sources/PackageDescription/PackageDescriptionSerialization.swift
```diff
@@ -76,6 +86,18 @@ extension Package.Dependency.Requirement: Encodable {
...
+    /// If the value fails to encode anything, `encoder` will encode an empty
+    /// keyed container in its place.
+    ///
+    /// This function throws an error if any values are invalid for the given
+    /// encoder's format.
+    ///
+    ///  - Note: This documentation comment was inherited from `Encodable`.
+    ///
+    /// - Parameters:
+    ///   - encoder: The encoder to write data to.
```
WowbaggersLiquidLunch, line 100: If the documentation is inherited from the default implementation of `encode`, then imo it shouldn't be included.

Sources/PackageDescription/PackageDescriptionSerialization.swift
```diff
@@ -429,6 +495,16 @@ extension SwiftVersion: Encodable {
...
+    /// Encodes this value into the given encoder.
+    ///
+    /// If the value fails to encode anything, `encoder` will encode an empty
+    /// keyed container in its place.
+    ///
+    /// This function throws an error if any values are invalid for the given
+    /// encoder's format.
+    ///
+    /// - Parameters:
+    ///   - encoder: The encoder to write data to.
```
WowbaggersLiquidLunch, line 507: I don't think this description is accurate.

Sources/PackageDescription/PackageDescriptionSerialization.swift
```diff
@@ -407,6 +463,16 @@ extension Target.PluginUsage: Encodable {
...
```
WowbaggersLiquidLunch, line 475: I don't think this description is accurate, either.

Sources/PackageDescription/PackageRequirement.swift
```diff
@@ -173,9 +182,13 @@ extension Package.Dependency {
     /// requirements before publishing a version of your package.
     @available(_PackageDescription, introduced: 5.6)
     public enum SourceControlRequirement {
+        /// A version based requirement.
```
WowbaggersLiquidLunch, line 185: As mentioned in #4247, this is not the only version-based requirement. Maybe we should emphasize the "exactness".

Sources/PackageDescription/SupportedPlatforms.swift
```diff
@@ -21,6 +20,11 @@ public struct Platform: Encodable, Equatable {
         self.name = name
     }
 
+    /// Create a custom platform.
```
WowbaggersLiquidLunch, line 23: The convention is using 3rd person present tense, so "Creates" instead of "Create".

Sources/PackageDescription/SupportedPlatforms.swift
```diff
@@ -21,6 +20,11 @@ public struct Platform: Encodable, Equatable {
...
+    ///
+    /// Use this function if none of the predefined platform names match the platform you are working with.
```
WowbaggersLiquidLunch, line 25: It might be more accurate to say "targeting" instead of "working with".

Sources/PackageDescription/SupportedPlatforms.swift
```diff
@@ -99,19 +103,23 @@ public struct SupportedPlatform: Encodable, Equatable {
...
     public static func macOS(_ version: SupportedPlatform.MacOSVersion) -> SupportedPlatform {
         return SupportedPlatform(platform: .macOS, version: version.version)
     }
 
     /// Configures the minimum deployment target version for the macOS platform
     /// using a version string.
     ///
-    /// The version string must be a series of two or three dot-separated integers, such as `10.10` or `10.10.1`.
-    ///
-    /// - Since: First available in PackageDescription 5.0
```
WowbaggersLiquidLunch, line 112: This line probably shouldn't be deleted.

Sources/PackageDescription/Version.swift
```diff
@@ -96,11 +94,43 @@ public struct Version {
...
+    ///
+    /// Equality is the inverse of inequality. For any values `a` and `b`, `a ==
+    /// b` implies that `a != b` is `false`.
+    ///
+    /// This is the default implementation of the equal-to operator (`==`) for
+    /// any type that conforms to `Equatable`.
+    ///
+    /// - Parameters:
+    ///   - lhs: A value to compare.
+    ///   - rhs: Another value to compare.
```
WowbaggersLiquidLunch, line 108: This is not the default implementation of `==`.

Sources/PackageDescription/Version.swift
```diff
@@ -96,11 +94,43 @@ public struct Version {
...
+    ///
+    /// The function determines if the inputs have any items in their
+    /// ``Version/prereleaseIdentifiers`` collections.
+    /// If only one side has prerelease identifiers, then that side will be considered
+    /// to have the lower version number. If they both have prerelease identifiers,
+    /// then the numeric prerelease identifier is considered to be lower version.
+    /// If both prerelease identifiers are numeric, then a numeric comparison is made.
+    ///
+    /// If neither of the parameters have prerelease identifiers, then the <doc://com.apple.documentation/documentation/swift/array/2853083-lexicographicallyprecedes>
+    /// function is used to perform the comparison.
```
WowbaggersLiquidLunch, line 129: This description is incorrect. Additionally, it's exposing the implementation detail to the public documentation. I think it's probably better to paraphrase the precedence rules listed on semver.org instead.

Sources/PackageDescription/LanguageStandardSettings.swift
zkiraly: We do need a sentence fragment ending in a period, as described in the [API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) under "Write a Documentation Comment." So I will leave this as is.

Sources/PackageDescription/LanguageStandardSettings.swift
zkiraly: I will use the original designations, like so

Sources/PackageDescription/LanguageStandardSettings.swift
zkiraly: Agreed. I will use the original designations.

Sources/PackageDescription/PackageDependency.swift
zkiraly: I do like "located at." Thanks for the suggestion.

Sources/PackageDescription/PackageDescription.swift
zkiraly: I had thought I changed all of those, but maybe I missed some, or I somehow reintroduced them? I'll be more vigilant.

Sources/PackageDescription/PackageDescription.docc/PackageDescription.md
zkiraly: This paragraph is about working with Xcode, which I only mention to direct the reader to the correct place where they can read up on using Swift packages with Xcode, which is the Xcode documentation. This documentation is maintained by Apple, and bug reports are appreciated if they do become outdated. Since it's not desired to teach the use of Xcode in the open source documentation, linking to those documents is unavoidable.

Sources/PackageDescription/PackageDescription.swift
zkiraly: Thanks for mentioning this again. I will figure something out, and I'd appreciate it if you gave a thumbs up or thumbs down.

Sources/PackageDescription/PackageDescription.swift
zkiraly: Let me experiment with this a bit. If DocC does not like it, then it's a bug on DocC, so I am inclined to file a bug and leave this as-is.

Sources/PackageDescription/PackageDescriptionSerialization.swift
zkiraly: The idea is to have some documentation that makes sense. If the inherited doc does not make sense, that's a problem, and I have written replacement docs in the few cases of protocol conformance docs that I found. Is that the case here too?

Sources/PackageDescription/PackageDescriptionSerialization.swift
zkiraly: I'll talk with Anders more about this. This file is really an internal implementation detail, but Anders thought it was not possible to hide it from DocC. But you are correct, the documentation should be accurate.

Sources/PackageDescription/PackageDescriptionSerialization.swift
zkiraly: Same comment as above about this source file.

Sources/PackageDescription/PackageDescription.swift
zkiraly: Oh, never mind, I misunderstood your comment. Yes, consistency is desired, and I think straight back-ticks are the way to go for this current use case.

Sources/PackageDescription/PackageDescription.swift
zkiraly: I went through, and I think I got all of them.

Sources/PackageDescription/PackageDescriptionSerialization.swift
zkiraly: Given that this extension is deprecated, and it's an internal implementation detail, I will simplify the documentation for this function.

Sources/PackageDescription/Version.swift
zkiraly: Removed reference to the default implementation.

Sources/PackageDescription/Version.swift
zkiraly: It might be best if I just wrote

> The precedence is determined according to rules described at the [Semantic Versioning 2.0.0](https://semver.org/#spec-item-11) website.

And leave it at that. I provide the exact link to the precedence rules, and the description there is easy to follow.

Sources/PackageDescription/SupportedPlatforms.swift
zkiraly: I put them back in this file. I'll go back and see if there are more in other files.

zkiraly:
@swift-ci please smoke test.

neonichu (approved):

Sources/PackageDescription/PackageDescription.docc/Curation/Extensions/LanguageTag-notEqual.md
```diff
@@ -0,0 +1,15 @@
...
+
+Returns a Boolean value indicating whether two values are not equal.
+
+- Parameters:
+  - lhs: A value to compare.
+  - rhs: Another value to compare.
+
+Inequality is the inverse of equality. For any values a and b, a != b implies that a == b is false.
+
+This is the default implementation of the not-equal-to operator (!=) for any type that conforms to Equatable.
```
WowbaggersLiquidLunch, line 15: Related to a comment i left before: I don't think it adds value to copy-past the default implementation's documentation. If there is anything unique, then we should _customize_ the documentation. If there is nothing to customize, then we should  just let the documentation automatically be inherited from the default implementations; otherwise, the copy-pasted documentation can become silently outdated when the default  documentation's changes.

WowbaggersLiquidLunch: The same opinion applies to the other PR too.

Sources/PackageDescription/SupportedPlatforms.swift
```diff
@@ -165,39 +179,47 @@ public struct SupportedPlatform: Encodable, Equatable {
...
     ///
     /// - Parameter version: The minimum deployment target that the package supports.
+    /// - Returns: A `SupportedPlatform` instance.
     public static func watchOS(_ version: SupportedPlatform.WatchOSVersion) -> SupportedPlatform {
         return SupportedPlatform(platform: .watchOS, version: version.version)
     }
-
-    /// Configures the minimum deployment target version for the watchOS platform
-    /// using a custom version string.
+
```
WowbaggersLiquidLunch, line 212: whitespace-only line

zkiraly: Looking at this in context, it seems fine.

zkiraly:
@swift-ci please smoke test.

Sources/PackageDescription/LanguageStandardSettings.swift
```diff
@@ -10,161 +10,160 @@
...
 //===----------------------------------------------------------------------===//
 
-/// The supported C language standard to use for compiling C sources in the package.
-///
-/// Aliases are available for some C language standards. For example,
-/// use `c89`, `c90`, or `iso9899_1990` for the "ISO C 1990" standard.
-/// To learn more, read the [Clang Compiler User's Manual][1].
-///
-/// [1]: <https://clang.llvm.org/docs/UsersManual.html#differences-between-various-standard-modes>
+/// The supported C language standard you use to complile C sources in the
```
abertelrud, line 13:
```suggestion
/// The supported C language standard you use to compile C sources in the
```

Sources/PackageDescription/LanguageStandardSettings.swift
```diff
@@ -10,161 +10,160 @@
...
     case iso9899_2017 = "iso9899:2017"
 
-    /// ISO C 2017.
+    /// The identifier for the ISO C 2017 language standard.
     @available(_PackageDescription, introduced: 5.4)
     case iso9899_2018 = "iso9899:2018"
 }
 
-/// The supported C++ language standard to use for compiling C++ sources in the package.
+/// The supported C++ language standard tyou use to compile C++ sources in the
```
abertelrud, line 86:
```suggestion
/// The supported C++ language standard you use to compile C++ sources in the
```

Sources/PackageDescription/PackageDependency.swift
```diff
@@ -20,24 +20,43 @@ extension Package {
...
+            /// A dependency located at the given path.
+            /// - Parameters:
+            ///    - name: The name of the dependency.
+            ///    - path: The path to the dependency.
             case fileSystem(name: String?, path: String)
+            /// A dependency based on a source control requirement.
+            ///  - Parameters:
+            ///    - name: The name of the dependency.
+            ///    - location: The Git URL of the dependency.
+            ///    - requirement: A `SourceControlRequirement` instance.
```
abertelrud, line 43: Should this also be `The version based requirement for a package` like the case below?  It's exactly the same use, and when seen in the code completion or help in Xcode, it's already apparent that the type is `SourceControlRequirement`.

abertelrud, line 43:
```suggestion
            ///    - requirement: The version based requirement for a package.
```

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -27,50 +27,61 @@ import Foundation
...
+/// package's manifest or losing access to existing packages.
 public final class Package {
     /// The name of the Swift package.
+    ///
+    /// If the name of the package is `nil`, Swift Package Manager deduces the name of the
+    /// package using its Git URL.
     public var name: String
 
-    /// The list of supported platforms with a custom deployment target.
+    /// The list of platforms supported by the package.
```
abertelrud, line 84: This isn't exactly what this property represents, unfortunately ("unfortunately" because the name suggests that it is).  What this does is specify the minimum version required for each of the platforms, e.g. macOS 10.15 for macOS, etc.  So I suggest "The list of minimum versions for supported platforms".

abertelrud, line 84:
```suggestion
    /// The list of minimum versions for platforms supported by the package.
```

abertelrud:
Looks good, thank you for doing this work!  A couple of spelling mistakes and a suggestion for an improved description.  The only significant one is that `platforms` is more about the minimum deployment targets than actually restricting supported platforms — this is a point of confusion in the API itself.  I have suggested replacement wording.

zkiraly:
@swift-ci please smoke test.

zkiraly:
@swift-ci please smoke test.

zkiraly:
@swift-ci please smoke test.

zkiraly:
@swift-ci please smoke test.

PR #4278, opened 2022-03-31
PackageModel: correct and improve XCTest location on Windows

compnerd:
This corrects a long standing issue in locating XCTest on Windows.  We
erroneously assumed that we could use `DEVELOPER_DIR` to construct a
path to XCTest.  However, this is not correct for two reasons:

  1. We may be targeting something other than Windows
  2. We may have passed `-sdk` explicitly, which would move the
     location of the SDK to something that no longer is relative
     to `DEVELOPER_DIR`.

Instead, we now construct the path relative to `SDKROOT` OR the explicit
SDK specified.  This reduces the dependency on the environment, and
makes `SDKROOT` behave as intended as well as completely drops the need
for `DEVELOPER_DIR` within SPM.

This will also restore the ability for us to have relocatable SDK
installations on Windows which eases some local testing for parallel SDK
installations.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -226,12 +226,12 @@ public final class UserToolchain: Toolchain {
                 // Windows uses a variable named SDKROOT to determine the root of
                 // the SDK.  This is not the same value as the SDKROOT parameter
                 // in Xcode, however, the value represents a similar concept.
-                if let SDKROOT = environment["SDKROOT"], let root = try? AbsolutePath(validating: SDKROOT) {
+                if let SDKROOT = environment["SDKROOT"], let sdkroot = try? AbsolutePath(validating: SDKROOT) {
                     var runtime: [String] = []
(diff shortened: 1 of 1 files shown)


abertelrud (approved):

neonichu (approved):

compnerd:
@abertelrud - is this okay to merge?  (If so, I'd likely do it over the weekend after I give this a tad bit more testing)

abertelrud:
> @abertelrud - is this okay to merge? (If so, I'd likely do it over the weekend after I give this a tad bit more testing)

Yes, this looks good to me.  Thanks @compnerd!

abertelrud:
Hi @compnerd — I just wanted to make sure you're not waiting for me to merge this.  Happy to leave it open if you're planning to do so later; no hurry, just figured I'd check.

compnerd:
There were some things that came up.  I want to wait until we have a green CI on the toolchain before merging this.

PR #4283, opened 2022-04-04
`--disable-sandbox` didn't properly disable it for build tool plugin commands (just the plugins themselves)

abertelrud:
### Motivation:

At some point `--disable-sandbox` seems to have stopped working for the commands run on behalf of plugins.  This caused one of the source compatibility suite packages (the protobuf one) to fail.

Not exactly sure when this broke, but the setting ended up not getting passed through in one of the refactorings a couple of weeks ago.  There wasn't a unit test covering this so it wasn't noticed until the source compat suite broke.  This commit adds a specific unit test for it.

### Changes:

- pass through `options.security.disableSandbox` to the corresponding `buildParameters` bit
- add a unit test

rdar://90955872

After merging this we should revert https://github.com/apple/swift-source-compat-suite/pull/652 which XFAILed the SwiftProtobuf package test.

### DIFF
--- a/Fixtures/Miscellaneous/Plugins/SandboxViolatingBuildToolPluginCommands/MyLibrary/Package.swift
+++ b/Fixtures/Miscellaneous/Plugins/SandboxViolatingBuildToolPluginCommands/MyLibrary/Package.swift
@@ -1,0 +1,16 @@
+// swift-tools-version: 5.6
+import PackageDescription
+
+let package = Package(
+    name: "MyLibrary",
+    dependencies: [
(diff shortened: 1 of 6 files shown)


abertelrud:
@swift-ci please smoke test

neonichu (approved):

abertelrud:
Linux: `failed: did not throw error`  Right, need to specialize this for macOS because we're not yet sandboxing on other platforms.

abertelrud:
@swift-ci please smoke test

PR #4286, opened 2022-04-08
Update Swift toolchain version

ahmdyasser:
Updating swift-tools-version comment that used when building the package.

### Motivation:
_[Swift 5.6 is now released and this will make SwiftPM up to date]_

### Modifications:
_[updating Swift version from 5.5 to 5.6]_

### Result:
_[This comment is used to tell the Swift Package Manager what version of the Swift toolchain to use when building our package. This will result in building packages using Swift 5.6, not Swift 5.5]_


abertelrud:
This version actually specifies the minimum toolchain version that is required in order to build SwiftPM.  For any package it should generally be set to the lowest version that will support building it — unless there is a reason why SwiftPM can no longer build using SwiftPM 5.5, we should keep it at that.

PR #4287, opened 2022-04-09
Fix ArgumentExtractor implementation bug

Kyle-Ye:
Fix ArgumentExtractor implementation bug

### Motivation:
`swift-docc-plugin` uses the PackagePlugin API and support `generate-documentation` action.

If we run
```
swift package --allow-writing-to-directory ./docs \
    generate-documentation --target SymbolKit --output-path ./docs
```
It will be fine.
However if we run
```
swift package --allow-writing-to-directory ./docs \
    generate-documentation --target=SymbolKit --output-path ./docs
```
It will loop infinitely.

And the core reason seems to be the ArgumentExtractor implementation in PackagePlugin Package.

### Modifications:

- Fix the ArgumentExtractor implementation bug
- Add testExtractOption for ArgumentExtractor

### Result:

When using `--<name>=<value>` form to specify the argument, the Plugin will no long loop infinitely.

### Other consideration

I think this should be cherry-picked into release/5.7
But should it be cherry-picked into release/5.6.2?
cc @tomerd @abertelrud

PR #4288, opened 2022-04-11
[5.7] Fix ArgumentExtractor implementation bug (#4287)

Kyle-Ye:
Fix ArgumentExtractor implementation bug

### Motivation:
`swift-docc-plugin` uses the PackagePlugin API and support `generate-documentation` action.

If we run
```
swift package --allow-writing-to-directory ./docs \
    generate-documentation --target SymbolKit --output-path ./docs
```
It will be fine.
However if we run
```
swift package --allow-writing-to-directory ./docs \
    generate-documentation --target=SymbolKit --output-path ./docs
```
It will loop infinitely.

And the core reason seems to be the ArgumentExtractor implementation in PackagePlugin Package.

### Modifications:

- Fix the ArgumentExtractor implementation bug
- Add testExtractOption for ArgumentExtractor

### Result:

When using `--<name>=<value>` form to specify the argument, the Plugin will no long loop infinitely.

PR #4289, opened 2022-04-11
[5.6] Fix ArgumentExtractor implementation bug (#4287)

Kyle-Ye:
Fix ArgumentExtractor implementation bug

### Motivation:
`swift-docc-plugin` uses the PackagePlugin API and support `generate-documentation` action.

If we run
```
swift package --allow-writing-to-directory ./docs \
    generate-documentation --target SymbolKit --output-path ./docs
```
It will be fine.
However if we run
```
swift package --allow-writing-to-directory ./docs \
    generate-documentation --target=SymbolKit --output-path ./docs
```
It will loop infinitely.

And the core reason seems to be the ArgumentExtractor implementation in PackagePlugin Package.

### Modifications:

- Fix the ArgumentExtractor implementation bug
- Add testExtractOption for ArgumentExtractor

### Result:

When using `--<name>=<value>` form to specify the argument, the Plugin will no long loop infinitely.

### Other consideration

I think this should be cherry-picked into release/5.7
But should it be cherry-picked into release/5.6.2?

### DIFF
--- a/Sources/PackagePlugin/ArgumentExtractor.swift
+++ b/Sources/PackagePlugin/ArgumentExtractor.swift
@@ -38,6 +38,7 @@ public struct ArgumentExtractor {
             else if arg.starts(with: "--\(name)=") {
                 arg.removeFirst(2 + name.count + 1)
                 values.append(arg)
+                args.remove(at: idx)
             }
             else {
                 idx += 1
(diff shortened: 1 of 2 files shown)


Kyle-Ye:
@swift-ci please smoke test

Kyle-Ye:
Hello @tomerd (Release manager for release/5.6 swift-package-manager)

Does this meet the merge bar of 5.6.2?

If not, I'll just close it and only cherry-pick this into release/5.7.

tomerd:
thanks @Kyle-Ye. will see if we can include it in the next possible window

abertelrud:
Looks like this is already in `main` as well as `release/5.7` (via https://github.com/apple/swift-package-manager/pull/4288) so the only remaining question is whether it is a candidate for 5.6.

PR #4290, opened 2022-04-12
Clean up the code and some of the API of compiling plugins, and improve persistence of cached compiler outputs

abertelrud:
### Motivation

Provide more information to clients of libSwiftPM even when a compiled plugin doesn't have to be recompiled because nothing has changed since the last time it was.  Also make the API a bit cleaner.

### Details

Clients now get the command line and compiler output even in the case of failure, and also when the plugin doesn't have to be recompiled because nothing changed (previously this information was lost).  This is done by serializing a proper struct to JSON rather than just storing the hexadecimal representation of the configuration hash.

Callers also have control over the name of the temporary directory to represent the plugin, rather than always having it be the last path component of the sources directory.

The changes to the construction of the command line to compile plugins isn't nearly as extensive as it looks — it's mostly blankspace changes due to nesting.

### Modifications:

- persist the compilation state as a JSON-coded struct rather than just a hexadecimal input hash
- let clients of PluginScriptRunner control the name of the intermediates directory (and not rely on a Sources type)
- clean up the struct members in the compilation result to have clearer names
- adjust and extend unit tests to cover cached information

### Result:

### DIFF
--- a/Sources/Workspace/DefaultPluginScriptRunner.swift
+++ b/Sources/Workspace/DefaultPluginScriptRunner.swift
@@ -366,14 +366,13 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
                 }
 
                 // Return a PluginCompilationResult for both the successful and unsuccessful cases (to convey diagnostics, etc).
-                let result = PluginCompilationResult(
+                return PluginCompilationResult(
                     succeeded: compilationState.succeeded,
                     commandLine: commandLine,
                     executableFile: execFilePath,
                     diagnosticsFile: diagFilePath,
                     compilerOutput: compilerOutput,
                     cached: false)
-                return result
             })
         }
     }
--- a/Sources/Workspace/DefaultPluginScriptRunner.swift
+++ b/Sources/Workspace/DefaultPluginScriptRunner.swift
@@ -295,7 +295,7 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
         }
         
         // Check if we already have a compiled executable and a persisted state (we only recompile if things have changed).
-        let stateFilePath = self.cacheDir.appending(component: execName + ".state")
+        let stateFilePath = self.cacheDir.appending(component: execName + "-state" + ".json")
         var compilationState: PersistedCompilationState? = .none
         if fileSystem.exists(execFilePath) && fileSystem.exists(stateFilePath) {
             do {
--- a/Sources/Workspace/DefaultPluginScriptRunner.swift
+++ b/Sources/Workspace/DefaultPluginScriptRunner.swift
@@ -237,7 +237,7 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
 
         // First try to create the output directory.
         do {
-            observabilityScope.emit(debug: "... plugin compilation output directory '\(execFilePath.parentDirectory)'")
+            observabilityScope.emit(debug: "Plugin compilation output directory '\(execFilePath.parentDirectory)'")
             try FileManager.default.createDirectory(at: execFilePath.parentDirectory.asURL, withIntermediateDirectories: true, attributes: nil)
         }
         catch {
@@ -264,7 +264,7 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
         }
         catch {
             // We couldn't compute the hash. We warn about it but proceed with the compilation (a cache miss).
-            observabilityScope.emit(debug: "... couldn't compute hash of plugin compilation inputs (\(error))")
+            observabilityScope.emit(debug: "Couldn't compute hash of plugin compilation inputs (\(error))")
             compilerInputHash = .none
         }
         
@@ -312,7 +312,7 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
             }
             catch {
                 // We couldn't read the compilation state file even though it existed. We warn about it but proceed with recompiling.
-                observabilityScope.emit(debug: "... couldn't read previous compilation state (\(error))")
+                observabilityScope.emit(debug: "Couldn't read previous compilation state (\(error))")
             }
         }
         
@@ -338,7 +338,7 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
             try fileSystem.removeFileTree(stateFilePath)
         }
         catch {
-            observabilityScope.emit(debug: "... couldn't clean up before invoking compiler (\(error))")
+            observabilityScope.emit(debug: "Couldn't clean up before invoking compiler (\(error))")
         }
         
         // Now invoke the compiler asynchronously.
@@ -362,7 +362,7 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
                 }
                 catch {
                     // We couldn't write out the `.state` file. We warn about it but proceed.
-                    observabilityScope.emit(debug: "... couldn't save plugin compilation state (\(error))")
+                    observabilityScope.emit(debug: "Couldn't save plugin compilation state (\(error))")
                 }
 
                 // Return a PluginCompilationResult for both the successful and unsuccessful cases (to convey diagnostics, etc).
--- a/Sources/Workspace/DefaultPluginScriptRunner.swift
+++ b/Sources/Workspace/DefaultPluginScriptRunner.swift
@@ -273,7 +273,6 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
             var commandLine: [String]
             var environment: [String:String]
             var inputHash: String?
-            var duration: Double
             var output: String
             var result: Result
             enum Result: Equatable, Codable {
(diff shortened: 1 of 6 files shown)


abertelrud:
@swift-ci please smoke test

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -37,43 +37,46 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
         self.cancellator = Cancellator(observabilityScope: .none)
     }
     
-    /// Public protocol function that starts compiling the plugin script to an executable. The tools version controls the availability of APIs in PackagePlugin, and should be set to the tools version of the package that defines the plugin (not of the target to which it is being applied). This function returns immediately and then calls the completion handler on the callbackq queue when compilation ends.
+    /// Public protocol function that starts compiling the plugin script to an executable. The name is used as the basename for the executable and auxiliary files.  The tools version controls the availability of APIs in PackagePlugin, and should be set to the tools version of the package that defines the plugin (not of the target to which it is being applied). This function returns immediately and then calls the completion handler on the callbackq queue after compilation ends.
```
tomerd, line 40: changes in this file are a bit hard to follow. are there specific changes you want the reviewers to focus on?

abertelrud: Ah, yes, sorry about that — it's actually a bit easier when ignoring spaces, because the nesting changed.

Maybe one useful thing would be to see what is stored in the PersistedCompilationState struct.  It doesn't yet have things like a fingerprint of the compiled plugin, but that's where it would be added.  I think that is the most interesting thing to look at that isn't just cleaning up names and structure.

This is pretty well tested by unit tests that check for whether the compilation actually happened etc so I am not too concerned about the logic itself.

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -125,204 +127,257 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
...
+                // Emit the compiler output as observable info.
+                let compilerOutput = ((try? process.utf8Output()) ?? "") + ((try? process.utf8stderrOutput()) ?? "")
+                observabilityScope.emit(info: compilerOutput)
+
+                // Save the persisted compilation state for possible reuse next time.
+                let compilationState = PersistedCompilationState(
+                    commandLine: commandLine,
+                    environment: toolchain.swiftCompilerEnvironment,
+                    inputHash: compilerInputHash,
+                    duration: Double(startTime.distance(to: .now()).milliseconds() ?? 0) / 1000.0,
```
abertelrud, line 360: Millisecond precision seems enough here.

tomerd (approved):

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -125,204 +127,257 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
...
-                    try stream <<< fileSystem.readFileContents(sourceFile).contents
+            }
+            
+            var succeeded: Bool {
+                return result == .exit(code: 0)
+            }
+        }
+        
+        // Check if we already have a compiled executable and a persisted state (we only recompile if things have changed).
+        let stateFilePath = self.cacheDir.appending(component: execName + ".state")
```
tomerd, line 299: nit: use "json" as extension, and names the file "state"

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -125,204 +127,257 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
...
-                    if (try fileSystem.readFileContents(hashFile)) == compilerInputsHash {
-                        compilationNeeded = false
-                    }
+                    try JSONEncoder.makeWithDefaults().encode(path: stateFilePath, fileSystem: self.fileSystem, compilationState)
                 }
                 catch {
-                    // We failed to read the `.inputhash` file. We warn about it but proceed with the compilation (a cache miss).
-                    observabilityScope.emit(warning: "Couldn't read previous hash of plugin compilation inputs (\(error)")
+                    // We couldn't write out the `.state` file. We warn about it but proceed.
+                    observabilityScope.emit(debug: "... couldn't save plugin compilation state (\(error))")
```
tomerd, line 368: curious about the "..." prefix in these debug messages

Sources/Workspace/DefaultPluginScriptRunner.swift
```diff
@@ -125,204 +127,257 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
...
-                            observabilityScope.emit(warning: "Couldn't write new hash of plugin compilation inputs (\(error)")
-                        }
-                        return result
-                    })
-                }
-            }
-            else {
-                // There is no need to recompile the executable, so we just call the completion handler with the results from last time.
+                // Return a PluginCompilationResult for both the successful and unsuccessful cases (to convey diagnostics, etc).
                 let result = PluginCompilationResult(
```
tomerd, line 372: nit: return here?

Tests/SPMBuildCoreTests/PluginInvocationTests.swift
```diff
@@ -89,17 +89,18 @@ class PluginInvocationTests: XCTestCase {
...
         // A fake PluginScriptRunner that just checks the input conditions and returns canned output.
         struct MockPluginScriptRunner: PluginScriptRunner {
-            
+
             var hostTriple: Triple {
                 return UserToolchain.default.triple
             }
             
-            func compilePluginScript(sources: Sources, toolsVersion: ToolsVersion, observabilityScope: ObservabilityScope) throws -> PluginCompilationResult {
+            func compilePluginScript(sourceFiles: [AbsolutePath], pluginName: String, toolsVersion: ToolsVersion, observabilityScope: ObservabilityScope) throws -> PluginCompilationResult {
```
tomerd, line 97: nit: these kind of functions that involve I/O seems like they should be async

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: Good point, though because the idea is to encode this into JSON, that would lead to additional code to encode or decode it.

I looked at adding a `fileprivate` extension to `DispatchTimeInterval` to make it adopt `Codable`, but that's not permissible — we then have to make it codable for the whole module.

Since we're currently not using the duration maybe I will just take it out for now.

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: Just the literal name "state" and not based on the name of the executable?  In that case we should maybe call the executable just "plugin".  But that makes it harder to see which plugin it is when looking for example using `top`.

I was trying to keep all three files (executable, state, and diagnostics) with the same base name derived from the plugin name.  So maybe just changing it to "json" here makes sense.

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: Probably just left over from print-style debugging (my indication for the end of an activity, whereas the beginning has a "..." suffix instead).  I'll changing it.

Sources/Workspace/DefaultPluginScriptRunner.swift
tomerd: sry I meant `execName + "-state.json"`

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: I'll fix the comment.

Sources/Workspace/DefaultPluginScriptRunner.swift
tomerd: one way to do this is use a property wrapper which implements the serialization for the duration property. this is how we often deal with date formatting

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: That makes sense.

Sources/Workspace/DefaultPluginScriptRunner.swift
abertelrud: I see.  As this is unused, I'll go ahead and remove it, but I'll keep that in mind for next time.

Tests/SPMBuildCoreTests/PluginInvocationTests.swift
abertelrud: I think the client calls that used this have been removed, so I can probably remove these.

abertelrud: Removed it.

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

Tests/SPMBuildCoreTests/PluginInvocationTests.swift
abertelrud: I pushed the `tsc_await` to all the call sites.

abertelrud:
@tomerd Thanks for your feedback, I think I've addressed all of it.

tomerd:
👍

compnerd:
This regressed the Windows build: https://ci-external.swift.org/job/oss-swift-windows-toolchain-x86_64-vs2019/509/console

PR #4292, opened 2022-04-12
Update deprecated `path.appending`

ahmdyasser:
Update deprecated method 

### Motivation:

`path.appending` is deprecated

### Modifications:
Replaced deprecated `path.appending` with  `AbsolutePath(_, relativeTo:)`

### Result:

Remove all warnings and make the source code up-to-date


tomerd:
thanks @ahmdyasser. this is a bit more complicated and discussed at length in https://github.com/apple/swift-package-manager/pull/4264

tomerd:
hi @ahmdyasser the likely fix is to un-deprecate the method from the TSC repo, which is under active discussion. In other words, I dont think there is any action you can or need to take at this point. we should probably close this PR.

tomerd:
closing as duplicate of #4264

PR #4294, opened 2022-04-13
update CommandTaskTracker to match llbuild's reporting of CustomTasks

tomerd:
motivation: adjust SwiftPM to llbuild reporting of task progress, which now includes custom tasks

changes: increment finished tasks count based on the commandStatusChanged event instead of the commandFinished event

rdar://84883732

### DIFF
--- a/Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
+++ b/Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
@@ -729,7 +729,7 @@ fileprivate struct CommandTaskTracker {
         case .isUpToDate:
             self.totalCount -= 1
         case .isComplete:
-            break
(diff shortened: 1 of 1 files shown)


tomerd:
@swift-ci smoke test

tomerd:
cc @BenchR267

abertelrud (approved):
Looks reasonable to me, though I would not want just my review on this as I'm not that familiar with the llbuild callbacks.  This seems like it's predicated on having the `.isComplete` status be true for both the `.succeeded` and `.skipped` results — if so, then this makes sense.

tomerd:
this was suggested by @BenchR267 based on changes done to llbuild

BenchR267 (approved):

PR #4295, opened 2022-04-14
restore disabled testSwiftRunSIGINT test

tomerd:
motivation: test coverage

changes: improve state machine to handle the case the output is fully buffered and enable the test

rdar://91024625

### DIFF
--- a/Tests/FunctionalTests/MiscellaneousTests.swift
+++ b/Tests/FunctionalTests/MiscellaneousTests.swift
@@ -385,9 +385,6 @@ class MiscellaneousTestCase: XCTestCase {
     }
 
     func testSwiftRunSIGINT() throws {
-        #if true
-        throw XCTSkip("skipping intermittently failing test (rdar://91024625)")
-        #else
         try fixture(name: "Miscellaneous/SwiftRun") { fixturePath in
             let mainFilePath = fixturePath.appending(component: "main.swift")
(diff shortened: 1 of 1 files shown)


tomerd:
@swift-ci smoke test

Tests/FunctionalTests/MiscellaneousTests.swift
```diff
@@ -440,23 +437,27 @@ class MiscellaneousTestCase: XCTestCase {
                 guard let output = String(bytes: bytes, encoding: .utf8) else {
                     return
                 }
+                print(output, terminator: "")
                 self.lock.withLock {
                     switch self.state {
                     case .idle:
-                        self.state = .buffering(output)
+                        self.state = processOutput(output)
```
abertelrud, line 444: Just curious if this could also have been done with a `fallthrough` here?

abertelrud: The separate function seems cleaner, arguably, mostly curious if I understand the logic here correctly.

abertelrud (approved):

Tests/FunctionalTests/MiscellaneousTests.swift
tomerd: cannot use `fallthrough ` afaik since we need to bind the variable

PR #4297, opened 2022-04-15
reduce load created by manifest loading test

tomerd:
motivation: flaky test

changes: reduce the amount of concurrent manifest loaded since its causing CI flakiness (runs for 1m+) and not material for what the test is checking - any number significantly larger than number of cores is sufficient

rdar://91272154


tomerd:
@swift-ci smoke test

abertelrud (approved):

neonichu (approved):

PR #4299, opened 2022-04-15
Fix breakage on Windows, which doesn't have `ProcessResult.ExitStatus.signalled`.

abertelrud:
This fixes a build break on Windows from #4290.

### DIFF
--- a/Sources/Workspace/DefaultPluginScriptRunner.swift
+++ b/Sources/Workspace/DefaultPluginScriptRunner.swift
@@ -241,14 +241,20 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
             var result: Result
             enum Result: Equatable, Codable {
                 case exit(code: Int32)
+                case abnormal(exception: UInt32)
                 case signal(number: Int32)
                 
                 init(_ processExitStatus: ProcessResult.ExitStatus) {
                     switch processExitStatus {
                     case .terminated(let code):
                         self = .exit(code: code)
+                    #if os(Windows)
+                    case .abnormal(let exception):
+                        self = .abnormal(exception: exception)
+                    #else
                     case .signalled(let signal):
                         self = .signal(number: signal)
+                    #endif
                     }
                 }
             }


abertelrud:
@swift-ci please smoke test

abertelrud:
Confirmed by code inspection that's the only use of `signal` in the original commit.

abertelrud:
For the future, it's also a bit suboptimal to have such platform specific APIs in `Process`, where each client has to conditionalize.  Maybe we can find a way of abstracting these concepts, even if not all kinds of results are applicable to all platforms.  I'm also thinking of cases such as maybe wanting to serialize this information, e.g. a unit test result that can end in an `.abnormal` condition on Windows but where it makes sense to reason about that in a cross-platform way.

compnerd:
> For the future, it's also a bit suboptimal to have such platform specific APIs in `Process`, where each client has to conditionalize. Maybe we can find a way of abstracting these concepts, even if not all kinds of results are applicable to all platforms. I'm also thinking of cases such as maybe wanting to serialize this information, e.g. a unit test result that can end in an `.abnormal` condition on Windows but where it makes sense to reason about that in a cross-platform way.

That would be nice.  I don't see how to unify that path immediately.  The concepts are different (there is no signal equivalent on Windows, and an abnormal exit is a crash really).

abertelrud:
> > For the future, it's also a bit suboptimal to have such platform specific APIs in `Process`, where each client has to conditionalize. Maybe we can find a way of abstracting these concepts, even if not all kinds of results are applicable to all platforms. I'm also thinking of cases such as maybe wanting to serialize this information, e.g. a unit test result that can end in an `.abnormal` condition on Windows but where it makes sense to reason about that in a cross-platform way.
> 
> That would be nice. I don't see how to unify that path immediately. The concepts are different (there is no signal equivalent on Windows, and an abnormal exit is a crash really - on unix, the signal may be a SIGTERM, SIGINT, SIGQUIT, SIGHUP none of which are a crash).

I was actually thinking that the union of all possible states could be represented on all platforms.  If it was Codable, then for example a unit test failure from a signal coded on a Unix box could be decoded on a Windows box with the concept of a Unix signal kept intact.

abertelrud:
Meanwhile, I'm preparing a cross-repo test run with a DNM branch in the Swift repo.  If we must revert quickly we can do that, then I would convert this to putting it back with the fix.

abertelrud:
https://github.com/apple/swift/pull/42405

tomerd (approved):

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

abertelrud:
Windows toolchain build at https://github.com/apple/swift/pull/42405 succeeded, @compnerd OK to merge?

compnerd (approved):

PR #4300, opened 2022-04-18
PackageModel: support the unified swift modules for XCTest

compnerd:
`XCTest` accidentally did not use the directory form of the swiftmodule,
which did not matter with a single architecture being available.
Migrate to the directory form to match the SDK setup.  With the SDK
migrating to the architecture independent directory for the modules,
follow suit and provide a compatibility path that will allow use of
either.

Co-dependents:
apple/swift#42419
apple/swift-installer-scripts#119
apple/swift-driver#1063

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -261,6 +261,16 @@ public final class UserToolchain: Toolchain {
                                         .appending(component: "XCTest-\(info.defaults.xctestVersion)")
 
                         xctest = [
+                            "-I", AbsolutePath("usr/lib/swift/windows", relativeTo: installation).pathString,
+                            // Migration Path
+                            //
+                            // Older Swift (<=5.7) installations placed the
+                            // XCTest Swift module into the architecture
+                            // specified directory.  This was in order to match
+                            // the SDK setup.  However, the toolchain finally
+                            // gained the ability to consult the architecture
+                            // indepndent directory for Swift modules, allowing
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

neonichu (approved):

tomerd (approved):

abertelrud (approved):

abertelrud:
Thanks @compnerd.  Is there a reasonable unit test that would make sure this doesn't break in the future?

compnerd:
I don't see a good way to test this - a mock setup would work, but would require checking in a full SDK and I'm not sure that is desirable (or am I wrong on that?).  Also note that this wasn't broken, this is setup for the new path while retaining the backwards compatibility.  At some point (possibly Swift 6), I'd like to consider dropping support for the migration path.

abertelrud:
> I don't see a good way to test this - a mock setup would work, but would require checking in a full SDK and I'm not sure that is desirable (or am I wrong on that?). Also note that this wasn't broken, this is setup for the new path while retaining the backwards compatibility. At some point (possibly Swift 6), I'd like to consider dropping support for the migration path.

Makes sense, but I always ask when a PR doesn't have a test. :)  In the past, in a different project, I've had success with mock SDKs that contain just enough bits to test the specific thing that's being tested.  That of course carries the risk that you're not testing the real thing.  I agree it would not be appropriate to check in a full SDKs.

Thanks!

elsh:
@compnerd feel free to merge when other PRs are ready

PR #4309, opened 2022-04-19
[5.7] Clean up the code and some of the API of compiling plugins, and improve persistence of cached compiler outputs (#4290 & #4299)

abertelrud:
This is the 5.7 nomination of https://github.com/apple/swift-package-manager/pull/4290 and the follow-on Windows fix https://github.com/apple/swift-package-manager/pull/4299.  Those PRs contain the details.

### Summary

Provide more information to clients of libSwiftPM even when a compiled plugin doesn't have to be recompiled because nothing has changed since the last time it was. Also make the API a bit cleaner.

### DIFF
--- a/Sources/Basics/JSON+Extensions.swift
+++ b/Sources/Basics/JSON+Extensions.swift
@@ -123,3 +123,10 @@ extension JSONDecoder {
         return try self.decode(kind, from: data)
     }
 }
+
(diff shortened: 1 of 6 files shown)


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #4311, opened 2022-04-20
Make 'PackageCollectionsSigningLibc' target conditional

yim-lee:
Motivation:
With https://github.com/apple/swift-crypto/pull/94, swift-crypto's
`Crypto` depends on `CCryptoBoringSSL` conditionally, on non-Apple
platforms only.

Since `PackageCollectionsSigningLibc` imports `CCryptoBoringSSL`
through `Crypto`, it fails to compile on Apple platforms.

Modifications:
Define `PackageCollectionsSigningLibc` target and associated
dependencies conditionally on platforms where it's available.

I tried using `.target(name:when:)` in combination with dynamic target
array, but got source not found error for
`PackageCollectionsSigningLibc`.

I also tried just changing target dependency on
`PackageCollectionsSigningLibc` to be conditional, but apparently
`PackageCollectionsSigningLibc` still gets compiled despite not being
used, thus causing build to fail.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -62,6 +62,50 @@ if useSwiftCryptoV2 {
     swiftSettings.append(.define("CRYPTO_v2"))
 }
 
+let packageCollectionSigningTargets: [Target]
+// swift-crypto's Crypto module depends on CCryptoBoringSSL on these platforms only
+#if os(Linux) || os(Windows) || os(Android)
+packageCollectionSigningTargets = [
+    .target(
+        /** Package collections signing C lib */
+        name: "PackageCollectionsSigningLibc",
+        dependencies: [
+            .product(name: "Crypto", package: "swift-crypto"), // for CCryptoBoringSSL
+        ],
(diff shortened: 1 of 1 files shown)


yim-lee:
@swift-ci please smoke test

yim-lee:
Looks like infra issue.

Package.swift
```diff
@@ -62,6 +62,50 @@ if useSwiftCryptoV2 {
...
+#else
+packageCollectionSigningTargets = [
+    .target(
+         /** Package collections signing */
+         name: "PackageCollectionsSigning",
+         dependencies: [
+            "Basics",
+            .product(name: "Crypto", package: "swift-crypto"),
+            "PackageCollectionsModel",
+         ],
```
compnerd, line 102: Can we always define the target but just add the dependency conditionally to avoid the duplication?

yim-lee:
@swift-ci please smoke test

tomerd (approved):

PR #4322, opened 2022-04-25
[5.7] Mark the SwiftPM 5.7 branch as a release branch

abertelrud:
As we do with every release, we need to turn off the development bit in SwiftPM so that it disallows use of 999.0 package manifest tools versions and also (somewhat less importantly) shows the version without a `-dev` suffix in response to `swift package --version`.

Specific changes:
- change `development` from true to false in SwiftVersion.swift

rdar://92289630


abertelrud:
@swift-ci please smoke test

abertelrud:
Failure is because apparently some of the snippet tests are still using 999.0.  Will fix.

tomerd:
@swift-ci please smoke test

abertelrud:
Thanks for taking over and finishing this PR, @tomerd!  Looks good to me.  Since I created the PR, I can't approve it but the diffs look good.

tomerd (approved):

PR #4324, opened 2022-04-25
Updated the links that point to using Swift packages with Xcode.

zkiraly:
I updated the links in the `PackageDescription` DocC docs to point to the correct location.

### Motivation:

The links were incorrect in that they used `_` (underscore) to separate words, while the actual URLs used `-` (dash).

### Modifications:

I modified the `Sources/PackageDescription/PackageDescription.docc/PackageDescription.md` file.

### Result:

Only the above mentioned file was modified, and it now contains the correct links.


zkiraly:
@swift-ci please smoke test.

neonichu (approved):
LGTM. 

Either URL spelling seems to work for me, though.

PR #4335, opened 2022-05-02
make local binary archives extraction more resilient

tomerd:
motivation: in some case duplicate artifacts or parallel processes may race the extraction of binary artifacts

changes:
add a file system lock around archive extraction of local archives, similar to what we do for remote archives

rdar://92622234

### DIFF
--- a/Sources/Workspace/Workspace+BinaryArtifacts.swift
+++ b/Sources/Workspace/Workspace+BinaryArtifacts.swift
@@ -165,12 +165,12 @@ extension Workspace {
 
             // finally download zip files, if any
             for artifact in zipArtifacts.get() {
-                let parentDirectory =  artifactsDirectory.appending(component: artifact.packageRef.identity.description)
-                guard observabilityScope.trap ({ try fileSystem.createDirectory(parentDirectory, recursive: true) }) else {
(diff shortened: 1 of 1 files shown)


Sources/Workspace/Workspace+BinaryArtifacts.swift
```diff
@@ -332,24 +332,26 @@ extension Workspace {
...
-                                observabilityScope.emit(debug: "no first level component stripping needed for \(tempExtractionDirectory)")
-                            }
-                            let content = try self.fileSystem.getDirectoryContents(tempExtractionDirectory)
-                            // copy from temp location to actual location
-                            for file in content {
-                                let source = tempExtractionDirectory.appending(component: file)
-                                let destination = destinationDirectory.appending(component: file)
-                                if self.fileSystem.exists(destination) {
-                                    try self.fileSystem.removeFileTree(destination)
+                            try self.fileSystem.withLock(on: destinationDirectory, type: .exclusive) {
```
tomerd, line 335: this is the key change. we already do this for remote / downloaded archives, this brings it back to local archives as well

tomerd:
@swift-ci please smoke test

abertelrud (approved):

PR #4872, opened 2022-05-04
Tests: use `FileManager` instead of `cp`

compnerd:
Rather than relying on UNIX semantics of `cp` which is inherently
platform dependent.  This is required to start experimenting with
getting the SPM test suite running on Windows.

### DIFF
--- a/IntegrationTests/Tests/IntegrationTests/Helpers.swift
+++ b/IntegrationTests/Tests/IntegrationTests/Helpers.swift
@@ -221,7 +221,11 @@ func fixture(
             if localFileSystem.isFile(fixtureDir.appending(component: "Package.swift")) {
                 // It's a single package, so copy the whole directory as-is.
                 let dstDir = tmpDirPath.appending(component: copyName)
+#if os(Windows)
+                try localFileSystem.copy(from: fixtureDir, to: dstDir)
+#else
                 try systemQuietly("cp", "-R", "-H", fixtureDir.pathString, dstDir.pathString)
+#endif
 
                 // Invoke the block, passing it the path of the copied fixture.
                 try body(dstDir)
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

tomerd:
+1 conceptually, need to get CI working. also can localFileSystem help?

compnerd:
Hmm, `localFileSystem` may be able to help.  I'll take a look at that, didn't think to look at it tbh.

compnerd:
Ugh, I _think_ that the failure here is due to the `-H` which follows symbolic links and I believe that the fixtures use those to share data.  I need to look into the behaviour of the copy implementation in `localFileSystem`, otherwise, we might have to do a bit more involved change.

tomerd:
where an API exists (like in this case), we should use `localFileSystem` instead of `FileSystem` directly. this allows us to reap the benefit from that abstraction that is used all throughout the code base (not just tests) so has to work on all platforms by definition. in practice `localFileSystem` uses `FileSystem`under the hood, but we can improve that if and when a better API comes along

compnerd:
@swift-ci please smoke test

compnerd:
The same problem remains - it wants to de-reference the symbolic links qwhen copying.  I'm not sure how to address this portably.  I suppose that we can just have two paths for the copying?

tomerd:
> I suppose that we can just have two paths for the copying?

you mean windows/non-windows? we could do that, or would add something to `Basics` that abstracts it and then use it from here. your call

compnerd:
@swift-ci please smoke test

compnerd:
@tomerd - any thing holding this up?

tomerd (approved):

tomerd:
@compnerd I think this is fine if urgent, but I would prefer to see use use a `FileSystem` based API on _all_ platforms, and do the platform specific work at the concrete implementation level, ie. `LocalFileSystem` can do the the right thing for each platform, and the call sites all use it without being aware to platform differences.

compnerd:
I think that the change would be significantly more complex.  We would need to re-implement the functionality of `-H` in `cp` which currently doesn't exist.

tomerd:
@compnerd @abertelrud @neonichu what if we added a new function to FileSystem to copy directory which will take some options including (or starting with) following symlinks? feels a bit more scalable that this

tomerd:
@compnerd, another question: how does this even work on Windows with this patch? I think Windows has adopted symlinks on some of its FSs?

compnerd:
Yes, I was able to pass the Swift PM test suite with this (and the other remaining patches).

compnerd:
I think that I'm going to merge this for now.  We are introducing more regressions in the SPM test suite on Windows, and in order to get stability we need to get the tests passing.  This allows us to reduce the number of failures so that we can start looking at the failures and getting things into a stable state.

PR #4874, opened 2022-05-04
Tests: don't try to compile un-executed code

compnerd:
This avoids compiling code which is not run as it makes assumptions
about the environment.


compnerd:
@swift-ci please smoke test

tomerd:
@compnerd this is okay, but I wonder why the code cannot compile (I assume on windows?)

PR #5446, opened 2022-05-04
Convert to `AbsolutePath(_:relativeTo:)`

compnerd:
This silences the warnings from the deprecation of the relative path
concatenation to a form which provides an anchor for the concatenation.
This is purely a mechanical change, replacing all the instances
identified by a build of swift-package-manager.

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -340,9 +340,9 @@ public final class ClangTargetBuildDescription {
 
         return sources.flatMap { (root, relativePaths) in
             relativePaths.map { source in
-                let path = root.appending(source)
-                let object = AbsolutePath("\(source.pathString).o", relativeTo: tempsPath)
-                let deps = AbsolutePath("\(source.pathString).d", relativeTo: tempsPath)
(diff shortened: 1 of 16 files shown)


tomerd (changes requested):
thanks @compnerd this is more or less the same solution as https://github.com/apple/swift-package-manager/pull/4264, but not quite what we want as its a far less intuitive API than the `AbsolutePath::appending(RelativePath)`. What we discussed in #4264 is that the real solution is to un-deprecate `AbsolutePath::appending(RelativePath)` and resolve any Windows specific issues in the implementation side in TSC

compnerd:
Yeah, I realize that the API isn't what you would like.  @abertelrud suggested putting this up to get the discussion going.  I don't think that there is a set of Windows specific issues - the design itself is the issue.  Appending a relative path doesn't make sense because it needs to be anchored.  Unlike Unix, there is no one root, it is a forest, so you need to specify the root.  This is not possible to represent in the construction of a `RelativePath` without an anchor.

compnerd:
Going to close this since we have the duplicate anyways ...

PR #5468, opened 2022-05-06
Basics: alter the definition of root for Windows

compnerd:
Rather than use a "root" for Windows, use the drive-relative path as a
"root" to terminate iteration/recursion.  Windows does not present the
file system as a unified singular rooted tree and instead has a forest
and thus the level-1 path is conceptually close to the root desired.


compnerd:
@swift-ci please test

tomerd (approved):

tomerd:
@swift-ci smoke test

PR #5469, opened 2022-05-06
Tests: correct some test cases for path separator

compnerd:
The test expectations were pre-constructed with `:` ignoring the path
separator.  Alter the expectation to construct the value at run time to
match the actual results.


compnerd:
@swift-ci please test

tomerd:
@swift-ci smoke test

tomerd (approved):

PR #5470, opened 2022-05-06
Tests: adjust BuildTests to compose paths

compnerd:
This adjusts most of the tests to use the path composition rather than
explicitly writing out the paths.  We now use the constructed context
for generating the expectations which should make it easier to change
expectations in the future should it be desired.

~This allows a significant portion of this test suite to now pass on
Windows (down to 11 failures from about 80).~

The BuildPlanTests suite fully passes on Windows now.

### DIFF
--- a/Tests/BuildTests/BuildPlanTests.swift
+++ b/Tests/BuildTests/BuildPlanTests.swift
@@ -26,6 +26,12 @@ import struct TSCUtility.BuildFlags
 import enum TSCUtility.Diagnostics
 import struct TSCUtility.Triple
 
+extension AbsolutePath {
+    fileprivate func escapedPathString() -> String {
+        return self.pathString.replacingOccurrences(of: "\\", with: "\\\\")
+    }
+}
+
 let hostTriple = UserToolchain.default.triple
 #if os(macOS)
     let defaultTargetTriple: String = hostTriple.tripleString(forPlatformVersion: "10.10")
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

Tests/BuildTests/BuildPlanTests.swift
```diff
@@ -343,8 +365,8 @@ final class BuildPlanTests: XCTestCase {
             let llbuild = LLBuildManifestBuilder(plan, fileSystem: fs, observabilityScope: observability.topScope)
             try llbuild.generateManifest(at: yaml)
             let contents: String = try fs.readFileContents(yaml)
-            XCTAssertMatch(contents, .contains("""
-                    inputs: ["/Pkg/Sources/exe/main.swift","/path/to/build/release/PkgLib.swiftmodule"]
+            XCTAssertMatch(contents.replacingOccurrences(of: "\\\\", with: "\\"), .contains("""
```
tomerd, line 368: can you elaborate on why `replacingOccurrences(of: "\\\\", with: "\\")` is needed, seems like maybe hints at an issue?

Tests/BuildTests/BuildPlanTests.swift
```diff
@@ -141,31 +141,49 @@ final class BuildPlanTests: XCTestCase {
...
+            "-L", objroot.pathString,
+            "-o", objroot.appending(components: "exe").pathString,
+            "-module-name", "exe",
+            "-static-stdlib",
+            "-emit-executable",
             "-Xlinker", "-rpath=$ORIGIN",
-            "@/path/to/build/debug/exe.product/Objects.LinkFileList",
+            "@\(objroot.appending(components: "exe.product", "Objects.LinkFileList"))",
             "-target", defaultTargetTriple,
         ]
```
tomerd, line 188: ^^ this is nice cleanup, thanks! one nit: can we call `objroot` `outputPath` or `buildPath`?

Tests/BuildTests/BuildPlanTests.swift
compnerd: It isn't entirely so clear to me if this is a real issue or not.  What it comes down to is that the string interpolation will translate the `\` to `\\` as the path is treated a string and `\` is an escape sequence otherwise.  This replacement will undo that string conversion so that the match actually matches the string.  Now perhaps we should not emit the secondary `\` in the string conversion, but that seems like a difficult change to make (and would require evolution likely).

Tests/BuildTests/BuildPlanTests.swift
compnerd: hahaha ... I think that we cross'ed paths in the middle.  I went from `build` (aka `buildPath`) to `objroot`.  I don't care enough tbh, if you feel that `buildPath` is better, I can do `s/objroot/buildPath/`.  I think that objroot is pretty common for the XCode builds and for Makefile, so its more familiar, but I'm not really invested in the name per se.

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

Tests/BuildTests/BuildPlanTests.swift
tomerd: since we both intuitively went for `buildPath`, I say we go with that

Tests/BuildTests/BuildPlanTests.swift
tomerd: @abertelrud wdyt?

compnerd:
@swift-ci please smoke test

Tests/BuildTests/BuildPlanTests.swift
abertelrud: So if I understand this correctly, this is the YAML that gets sent to LLBuild.  And it needs any reverse solidus that represents a path separator in the YAML file to be escaped?  If so then that seems expected, though I think it would be clearer to:
1. have a fileprivate `escapedForYAML` property on String
2. use that on the right side of the `XCTAssertMatch` call
That makes it clearer, I think, that we're looking for the YAML representation of a path in this file.

Does that seem reasonable?

compnerd: I believe that is mostly the case (the only bit that I am uncertain about is the contents of the file).

As to the proposed solution, I agree with it in principle, not in practice.  I'd prefer `escaped` over `escapedForYAML` simply because we use YAML and JSON for various things and both of them will end up escaping the sequence and we will end up with another set of confusion.

compnerd: For the moment, I've added the extension to the file, but it is likely that we will need to hoist it as we make progress with the rest of the changes.

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

Tests/BuildTests/BuildPlanTests.swift
```diff
@@ -26,6 +26,12 @@ import struct TSCUtility.BuildFlags
 import enum TSCUtility.Diagnostics
 import struct TSCUtility.Triple
 
+extension String {
```
tomerd, line 29: can this be a file private extension, and on `AbsolutePath` instead of on `String`? should it also be named something more descriptive than "escaped", eg "backslashesEscaped"

tomerd, line 29: can this be a file private extension, and on `AbsolutePath` instead of on `String`? should it also be named something more descriptive than "escaped", eg "backslashesEscaped"

compnerd: I'm confused on the duplication of the review, but I don't see any reason we couldn't do that on `AbsolutePath`.  Do we want to make it something like `escapedPathString` instead?

tomerd (approved):

tomerd:
@swift-ci smoke test

PR #5471, opened 2022-05-06
`swift package` should ask for plugin permissions interactively when on a TTY

abertelrud:
Command plugins can include a request for permission to modify the package directory.  In this case, `swift package` should:
a) if connected to a TTY, ask a "do you want to do this" question and wait for input
b) if not connected to a TTY, emit a message that you need to pass `--allow-writing-to-package-directory`

### Motivation:

The current message doesn't provide information about how to grant the permission the plugin wants.  Also, if the user is invoking it interactively, it makes more sense to ask permission than have them invoke it again.

### Modifications:

- if connected to a tty, ask the user
- if not, include the flag name in the error
- move down a extension to WritableOutputStream to detect whether it's connected to a TTY; this is now in Basics
- adjust unit test

### Results:

When on a TTY this now looks like:
```
Plugin ‘Reformat Code Using SwiftFormat’ wants permission to write to the package directory.
Stated reason: “This command modifies source files to apply formatting rules”.
Allow this plugin to write to the package directory? (yes/no) 
```
and then if you don't answer "yes":
```
error: Plugin was denied permission to write to the package directory.
```
When not on a TTY this now looks like:
```
error: Plugin ‘Reformat Code Using SwiftFormat’ wants permission to write to the package directory.
Stated reason: “This command modifies source files to apply formatting rules.”.
Use `--allow-writing-to-package-directory` to allow this.
```

rdar://89442616

### DIFF
--- a/Sources/Commands/SwiftPackageTool.swift
+++ b/Sources/Commands/SwiftPackageTool.swift
@@ -1028,8 +1028,26 @@ extension SwiftPackageTool {
                 // If the plugin requires write permission but it wasn't provided, we ask the user for approval.
                 if case .command(_, let permissions) = plugin.capability {
                     for case PluginPermission.writeToPackageDirectory(let reason) in permissions {
-                        // TODO: Ask for approval here if connected to TTY; only emit an error if not.
-                        throw ValidationError("Plugin ‘\(plugin.name)’ needs permission to write to the package directory (stated reason: “\(reason)”)")
+                        let problem = "Plugin ‘\(plugin.name)’ wants permission to write to the package directory."
+                        let reason = "Stated reason: “\(reason)”."
+                        if swiftTool.outputStream.isTTY {
+                            // We can ask the user directly, so we do so.
+                            let query = "Allow this plugin to write to the package directory?"
+                            swiftTool.outputStream.write("\(problem)\n\(reason)\n\(query) (yes/no) ".utf8)
+                            swiftTool.outputStream.flush()
+                            let answer = readLine(strippingNewline: true)
+                            // Throw an error if we didn't get permission.
+                            if answer?.lowercased() != "yes" {
+                                throw ValidationError("Plugin was denied permission to write to the package directory.")
+                            }
+                            // Otherwise append the directory to the list of allowed ones.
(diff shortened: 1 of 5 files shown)


Sources/Commands/SwiftPackageTool.swift
```diff
@@ -1967,3 +1983,18 @@ extension BuildOperation {
         }
     }
 }
+
+extension OutputByteStream {
```
abertelrud, line 1987: This is currently copied from the BuildOperation where it is private.  We should have only one implementation, so I'm looking for opinions on where this code should live.

Tests/CommandsTests/PackageToolTests.swift
```diff
@@ -1762,7 +1762,9 @@ final class PackageToolTests: CommandsTestCase {
                 let result = try SwiftPMProduct.SwiftPackage.executeProcess(["plugin", "PackageScribbler"], packagePath: packageDir, env: ["DECLARE_PACKAGE_WRITING_PERMISSION": "1"])
                 XCTAssertNotEqual(result.exitStatus, .terminated(code: 0))
                 XCTAssertNoMatch(try result.utf8Output(), .contains("successfully created it"))
-                XCTAssertMatch(try result.utf8stderrOutput(), .contains("error: Plugin ‘MyPlugin’ needs permission to write to the package directory (stated reason: “For testing purposes”)"))
```
abertelrud, line 1765: This tests the non-interactivity but I'm not sure how to add a test for the interactivity.  Maybe `echo yes | swift package foo` will work, I'll try that.

abertelrud:
@swift-ci please smoke test

Sources/Commands/SwiftPackageTool.swift
abertelrud: Move this down to Basics.  I think it's good actually that there is a public accessor for checking whether a stream is connected to a TTY.

abertelrud:
@swift-ci please smoke test

tomerd (approved):

abertelrud:
@swift-ci please smoke test

PR #5479, opened 2022-05-09
Tests: adjust test for real file systems

compnerd:
There is no guarantee that there is a singular unified file system view.
Use the current directory for the root of the filesystem.  This is
required for platforms like macOS (classic) and Windows where a forest
of roots exist.

### DIFF
--- a/Tests/XcodeprojTests/GenerateXcodeprojTests.swift
+++ b/Tests/XcodeprojTests/GenerateXcodeprojTests.swift
@@ -154,7 +154,7 @@ class GenerateXcodeprojTests: XCTestCase {
             XCTAssertNoDiagnostics(observability.diagnostics)
 
             _ = try xcodeProject(
-                xcodeprojPath: AbsolutePath.root.appending(component: "xcodeproj"),
+                xcodeprojPath: dstdir.appending(component: "xcodeproj"),
                 graph: graph, extraDirs: [], extraFiles: [],
                 options: XcodeprojOptions(), fileSystem: localFileSystem,
                 observabilityScope: observability.topScope
--- a/Tests/FunctionalTests/MiscellaneousTests.swift
+++ b/Tests/FunctionalTests/MiscellaneousTests.swift
@@ -421,7 +421,11 @@ class MiscellaneousTestCase: XCTestCase {
 
             // check for interrupt result
             let result = try process.waitUntilExit()
+#if os(Windows)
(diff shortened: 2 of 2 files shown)


compnerd:
@swift-ci please smoke test

Tests/XcodeprojTests/GenerateXcodeprojTests.swift
```diff
@@ -154,7 +154,7 @@ class GenerateXcodeprojTests: XCTestCase {
             XCTAssertNoDiagnostics(observability.diagnostics)
 
             _ = try xcodeProject(
-                xcodeprojPath: AbsolutePath.root.appending(component: "xcodeproj"),
+                xcodeprojPath: AbsolutePath(localFileSystem.currentWorkingDirectory!.components.first!).appending(component: "xcodeproj"),
```
tomerd, line 157: since we are using the local file system, should we change all operations to use directories inside the temp dir (`dstdir`)? otherwise this will have side effects on he user machine

tomerd: p.s. I realize this was there before the PR, but we might as well fix that together

compnerd: I think that it should be fine as is - we do clean up; but, I think that localizing the storage of course is better.  Does it make sense to do that as a follow up so that we can scrub through any other instances?

tomerd: oh, where do we clean up?

fwiw `testGenerateXcodeprojWithoutGitRepo` seems like it is doing something a better (uses the temp dir) to this end, so maybe we can copy that pattern?

compnerd: `testWithTemporaryDirectory ` has a `defer`ed cleanup embedded in it.  However, in the case of a crash or something, it would still leave content behind.  I'm not against the further clean up work though - I think that it is better to do that as leaving stuff in PWD is horrible.

tomerd: yea. that defer only cleans up things in the temp dir provided, not all side effects, so we need to make sure all files we create as part of the test are contained in the temp dir

compnerd:
@swift-ci please smoke test

tomerd (approved):

PR #5480, opened 2022-05-09
SourceControl: normalise repository basename computation

compnerd:
Prefer using the `url` member to get a URL representation of the
location, and then use the `pathComponents` property rather than
splitting on `/` which may be an invalid path separator.

### DIFF
--- a/Sources/SourceControl/Repository.swift
+++ b/Sources/SourceControl/Repository.swift
@@ -42,7 +42,7 @@ public struct RepositorySpecifier: Hashable {
 
     /// Returns the cleaned basename for the specifier.
     public var basename: String {
-        var basename = self.url.pathComponents.dropFirst(1).last(where: { !$0.isEmpty }) ??  ""
+        var basename = self.url.pathComponents.dropFirst(1).last(where: { !$0.isEmpty }) ?? ""
         if basename.hasSuffix(".git") {
             basename = String(basename.dropLast(4))
         }
--- a/Sources/SourceControl/Repository.swift
+++ b/Sources/SourceControl/Repository.swift
@@ -42,7 +42,7 @@ public struct RepositorySpecifier: Hashable {
 
     /// Returns the cleaned basename for the specifier.
     public var basename: String {
-        var basename = self.location.description.components(separatedBy: "/").last(where: { !$0.isEmpty }) ?? ""
+        var basename = self.url.pathComponents.dropFirst(1).last(where: { !$0.isEmpty }) ??  ""
         if basename.hasSuffix(".git") {
             basename = String(basename.dropLast(4))
         }


Sources/SourceControl/Repository.swift
```diff
@@ -42,7 +42,7 @@ public struct RepositorySpecifier: Hashable {
 
     /// Returns the cleaned basename for the specifier.
     public var basename: String {
-        var basename = self.location.description.components(separatedBy: "/").last(where: { !$0.isEmpty }) ?? ""
+        var basename = self.url.pathComponents.last(where: { !$0.isEmpty })!
```
tomerd, line 45: cleanup looks good, but lets not force unwarp

compnerd: I think that the empty string is wrong - a repository _should_ have a base name even if it is bare.  Is there some case that it would be empty?  Converting it to a throwing function makes sense.

compnerd: Wait, this is a computed property.  We can make it into an effectful property, but that might have a large fallout.

tomerd: maybe that is why it was defaulted to empty? perhaps one way to deal with this is to default to empty by add an assert to that at least in development time we know something bad has happened?

compnerd: Hmm I went with a slightly different approach - `guard` + `fatalError` ... though I suppose the result is roughly the same.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please smoke test

Sources/SourceControl/Repository.swift
```diff
@@ -42,7 +42,9 @@ public struct RepositorySpecifier: Hashable {
 
     /// Returns the cleaned basename for the specifier.
     public var basename: String {
-        var basename = self.location.description.components(separatedBy: "/").last(where: { !$0.isEmpty }) ?? ""
+        guard var basename = self.url.pathComponents.last(where: { !$0.isEmpty }) else {
+            fatalError("URL should have at least one non-empty component")
+        }
```
tomerd, line 47: this ie effectively the same as force unwrap (with a better error message). the issue with this approach is that it can cause crashes for processes hosting libSwiftPM like IDEs. given that this did not crash before, I worry about changing this behavior now. a more subtle change would be to default to empty string (as before) but assert on it, which only takes effect in debug builds and as such will not introduce new crashing behavior

compnerd: Ah, I see.

compnerd:
@swift-ci please smoke test

compnerd:
Haha, the failure seems to be in TSCBasic: https://github.com/apple/swift-tools-support-core/blob/main/Sources/TSCBasic/Path.swift#L669

abertelrud:
> Haha, the failure seems to be in TSCBasic: https://github.com/apple/swift-tools-support-core/blob/main/Sources/TSCBasic/Path.swift#L669

Right, it assumes that relative paths don't start with `/` on macOS and Linux.  Are we expecting relative paths to be able to start with a path separator now? — if so, that doesn't sound right.

compnerd:
No, nothing here makes that assumption AFAICT.

compnerd:

Sources/SourceControl/Repository.swift
```diff
@@ -42,7 +42,7 @@ public struct RepositorySpecifier: Hashable {
 
     /// Returns the cleaned basename for the specifier.
     public var basename: String {
-        var basename = self.location.description.components(separatedBy: "/").last(where: { !$0.isEmpty }) ?? ""
+        var basename = self.url.pathComponents.dropFirst(1).last(where: { !$0.isEmpty }) ??  ""
```
tomerd, line 45: sorry: kill extra space:

```suggestion
        var basename = self.url.pathComponents.dropFirst(1).last(where: { !$0.isEmpty }) ?? ""
```

tomerd:
@swift-ci please smoke test

compnerd:
@tomerd good to merge?

tomerd (approved):

PR #5483, opened 2022-05-09
[5.7] `swift package` should ask for plugin permissions interactively when on a TTY

abertelrud:
This is the 5.7 cherry-pick nomination of https://github.com/apple/swift-package-manager/pull/5471

Command plugins can include a request for permission to modify the package directory. In this case, `swift package` should:
a) if connected to a TTY, ask a "do you want to do this" question and wait for input
b) if not connected to a TTY, emit a message that you need to pass `--allow-writing-to-package-directory`

rdar://89442616
(cherry picked from commit e65712cf6548677fb60ce17b8917031c0c11f239)

### DIFF
--- a/Sources/Basics/CMakeLists.txt
+++ b/Sources/Basics/CMakeLists.txt
@@ -29,7 +29,8 @@ add_library(Basics
   Triple+Extensions.swift
   SwiftVersion.swift
   SQLiteBackedCache.swift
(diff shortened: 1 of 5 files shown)


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #5484, opened 2022-05-10
PackageLoading: always anchor the relative path

compnerd:
We would previously construct an absolute path by taking a relative path
and building from that.  However, without an anchor, we cannot validate
the path on Windows.  Construct the actual path with an anchor and use
the initial construction of a relative path as a validation step.

### DIFF
--- a/Sources/PackageLoading/PackageBuilder.swift
+++ b/Sources/PackageLoading/PackageBuilder.swift
@@ -481,11 +481,11 @@ public final class PackageBuilder {
                 }
 
                 // Make sure target is not referenced by absolute path
-                guard let relativeSubPath = try? RelativePath(validating: subpath) else {
+                if (try? RelativePath(validating: subpath)) == nil {
                     throw ModuleError.unsupportedTargetPath(subpath)
                 }
 
-                let path = packagePath.appending(relativeSubPath)
+                let path = AbsolutePath(subpath, relativeTo: packagePath)
                 // Make sure the target is inside the package root.
                 guard path.isDescendantOfOrEqual(to: packagePath) else {
                     throw ModuleError.targetOutsidePackage(package: self.identity.description, target: target.name)


compnerd:
@swift-ci please smoke test

Sources/PackageLoading/PackageBuilder.swift
```diff
@@ -481,11 +481,11 @@ public final class PackageBuilder {
                 }
 
                 // Make sure target is not referenced by absolute path
-                guard let relativeSubPath = try? RelativePath(validating: subpath) else {
+                if (try? RelativePath(validating: subpath)) == nil {
                     throw ModuleError.unsupportedTargetPath(subpath)
                 }
 
-                let path = packagePath.appending(relativeSubPath)
+                let path = AbsolutePath(subpath, relativeTo: packagePath)
```
tomerd, line 488: this change is fine, but its subtle and I can already see someone reverting this to use the nicer "appending" API at some future refactoring. in that sense it is related to our discussion on https://github.com/apple/swift-package-manager/pull/5446#pullrequestreview-962771235: we should really make `AbsolutePath::appending(RelativePath)` work well on Windows instead of deprecating it

compnerd: hmm, I _could_ see the implementation just being sugar for that potentially...

```
extension AbsolutePath {
  func appending(_ path: String) {
    AbsolutePath(path, relativeTo: self)
  }
}
```

tomerd: that is not a terrible idea IMO. @abertelrud ?

compnerd: Note that the type is `String` - we don't want to be able to use `RelativePath` unless we change the instances to `RelativePath(unsafeUncheckedPath:)` (since we cannot validate the path without the anchor).

tomerd: hmm, yea I think we want the `AbsolutePath::appending(RelativePath)` back, still unclear to me while we cannot use them since we are still creating an instance a couple of lines above

compnerd: The creation above only validates that the path does not start with `/` on Unix, and does not start with `[A-Z]:` or `\\` or `//`.  We could just inline that there I suppose.

tomerd:
@compnerd can this be closed given https://github.com/apple/swift-package-manager/pull/5910?

tomerd:
@compnerd can this be closed given https://github.com/apple/swift-package-manager/pull/5910?

compnerd:
I think so?  The current package loading test failures are due to path normalisation rather than relative path construction.  Thanks for the reminder!

PR #5486, opened 2022-05-10
Add `nuget` package provider

compnerd:
Introduce the `nuget` package provider, which is the only option on
Windows, but is portable to other systems as well. This is required
for enabling the test suite on Windows.

Refs: #4369

### DIFF
--- a/Sources/PackageDescription/PackageDescription.swift
+++ b/Sources/PackageDescription/PackageDescription.swift
@@ -388,7 +388,7 @@ public enum SystemPackageProvider {
     /// Packages installable by the yum package manager.
     @available(_PackageDescription, introduced: 5.3)
     case yumItem([String])
-    @available(_PackageDescription, introduced: 9999)
+    @available(_PackageDescription, introduced: 999.0)
     case nugetItem([String])
 
     /// Creates a system package provider with a list of installable packages
@@ -429,7 +429,7 @@ public enum SystemPackageProvider {
     /// - Parameter packages: The list of package names.
     ///
     /// - Returns: A package provider.
-    @available(_PackageDescription, introduced: 9999)
+    @available(_PackageDescription, introduced: 999.0)
     public static func nuget(_ packages: [String]) -> SystemPackageProvider {
         return .nugetItem(packages)
     }
--- a/Sources/PackageDescription/PackageDescription.swift
+++ b/Sources/PackageDescription/PackageDescription.swift
@@ -388,6 +388,8 @@ public enum SystemPackageProvider {
     /// Packages installable by the yum package manager.
     @available(_PackageDescription, introduced: 5.3)
     case yumItem([String])
+    @available(_PackageDescription, introduced: 9999)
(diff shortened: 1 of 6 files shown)


compnerd:
@swift-ci please smoke test

Sources/PackageLoading/Target+PkgConfig.swift
```diff
@@ -148,6 +150,13 @@ extension SystemPackageProviderDescription {
             if case .linux(.fedora) = platform {
                 return true
             }
+        case .nuget:
+            switch platform {
+            case .darwin, .windows, .linux:
```
tomerd, line 155: should this be windows only? is there formal support and practical adoption of nuget on linux and darwin?

compnerd: Yes, they are formally supported and adopted - even though the primary usage is .NET, but that is more a historical artifact.

tomerd (approved):

compnerd:
Seems that it doesn't like the 9999 version, need to figure out what version main is.

Sources/PackageLoading/Target+PkgConfig.swift
compnerd: A nice example package: https://www.nuget.org/packages/runtime.debian.8-x64.runtime.native.System.Security.Cryptography.OpenSsl/

Prebuilt .so for Debian x86_64.

tomerd:
@abertelrud should know. it may be the case that we need to make a change on `main` to enable it again

abertelrud:
> @abertelrud should know. it may be the case that we need to make a change on `main` to enable it again

I think it's just 999.

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -388,6 +388,8 @@ public enum SystemPackageProvider {
     /// Packages installable by the yum package manager.
     @available(_PackageDescription, introduced: 5.3)
     case yumItem([String])
+    @available(_PackageDescription, introduced: 9999)
```
abertelrud, line 391:
```suggestion
    @available(_PackageDescription, introduced: 999.0)
```

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -420,6 +422,17 @@ public enum SystemPackageProvider {
...
         return .yumItem(packages)
     }
+
+    /// Creates a system package provider with a list of installable packages
+    /// for users of the nuget package manager on Linux or Windows.
+    ///
+    /// - Parameter packages: The list of package names.
+    ///
+    /// - Returns: A package provider.
+    @available(_PackageDescription, introduced: 9999)
```
abertelrud, line 432:
```suggestion
    @available(_PackageDescription, introduced: 999.0)
```

abertelrud (approved):

compnerd:
@swift-ci please smoke test

PR #5487, opened 2022-05-10
Pass `-Xfrontend -disable-implicit-string-processing-module-imports` to manifest and plugin compilation

abertelrud:
### Motivation:

This is to avoid warnings of the form `warning build: unable to perform implicit import of "_StringProcessing" module: no such module found`.

### Modifications:

- pass ["-Xfrontend", "-disable-implicit-string-processing-module-import"] when compiling manifests if the compiler supports it

### Result:

Avoid warnings when the `_StringProcessing` module isn't available.

### DIFF
--- a/Sources/Commands/SwiftTool.swift
+++ b/Sources/Commands/SwiftTool.swift
@@ -918,6 +918,10 @@ public class SwiftTool {
             if SwiftTargetBuildDescription.checkSupportedFrontendFlags(flags: ["disable-implicit-concurrency-module-import"], fileSystem: self.fileSystem) {
                 extraManifestFlags += ["-Xfrontend", "-disable-implicit-concurrency-module-import"]
             }
+            // Disable the implicit string processing import if the compiler in use supports it to avoid warnings if we are building against an older SDK that does not contain a StringProcessing module.
+            if SwiftTargetBuildDescription.checkSupportedFrontendFlags(flags: ["disable-implicit-string-processing-module-import"], fileSystem: self.fileSystem) {
+                extraManifestFlags += ["-Xfrontend", "-disable-implicit-string-processing-module-import"]
+            }
 
             if self.logLevel <= .info {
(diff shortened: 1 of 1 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
As I worried about, the `-disable-implicit-string-processing-module-import` isn't supported by the compiler being used for the self-hosted tests.  Looking into the best way of getting it to be passed only when the driver supports it.

abertelrud:
@artemcm @neonichu Any ideas here? At the level where we need to decide where to pass this flag, we can't import the integrated driver code to determine if it's supported. I looked at using the extra flags in the UserToolchain for this, but it seems that we'd have a lot of callsites to change in order to be at a high enough level to be able to call `supportedFrontendFlags`.

abertelrud:
Alternatively, would it make sense to filter out all remarks?  It doesn't seem to me that compiler remarks should ever result in warnings.

artemcm:
Short of special-casing filtering this warning, we could add an entry to the compiler's features manifest and check it from SwiftPM when picking the toolchain?
https://github.com/apple/swift/blob/main/lib/Option/features.json

abertelrud:
> Short of special-casing filtering this warning, we could add an entry to the compiler's features manifest and check it from SwiftPM when picking the toolchain?
> https://github.com/apple/swift/blob/main/lib/Option/features.json

That makes sense, but I think that's already what the Driver is doing?  Or do you mean we'd have SwiftPM (and libSwiftPM) do this itself at a lower level, e.g. in the UserToolchain?

Boris also pointed out that he added something similar for the concurrency, which we can use for at least the manifest (for plugins we should probably add additional flags support anyway).

In general I do think we should probably _also_ filter out compiler remarks since it doesn't seem right to ever count them as warnings.

abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #5488, opened 2022-05-11
fix concurrent access issue when running tests in parallel

tomerd:
motivation: fix data race

changes: add lock when mutating test output as it may happen concurrently

rdar://92260631


tomerd:
@swift-ci smoke test

abertelrud (approved):

tomerd:
>  Is the output line-delimited already or would it still need to also be buffered by line per test in order to avoid producing output that might be garbled?

we want the raw output in this case - its whatever the test writes to stdout

PR #5491, opened 2022-05-11
Fix a wording error in a unit test for package plugin commands

abertelrud:
It looks like the reason this wasn't caught in CI is because these tests have to be skipped since the test machines are running an older set of tools that don't support concurrency, which plugins require.  It's concerning if too many tests are skipped in CI, so we should see about having a way to not have so many tests getting skipped in CI.


abertelrud:
@swift-ci please smoke tests

tomerd:
cc @shahmishal for updating the CI hosts machines

tomerd (approved):

neonichu (approved):

PR #5500, opened 2022-05-12
SourceControl: remove unnecessary shell usage

compnerd:
Rather than redirecting the file list from a file, pass the files on the
command line.  This avoids an unnecessary file write as well as a
subshell.  Furthermore, it makes the command more portable and enables
this to now succeed on Windows.

### DIFF
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -597,29 +597,23 @@ public final class GitRepository: Repository, WorkingCheckout {
         return try self.lock.withLock {
             let stringPaths = paths.map { $0.pathString }
 
-            return try withTemporaryFile { pathsFile in
-                try localFileSystem.writeFileContents(pathsFile.path) {
-                    for path in paths {
-                        $0 <<< path.pathString <<< "\0"
-                    }
+            let output: String
+            do {
+                output = try self.git.run(["-C", self.path.pathString, "check-ignore"] + stringPaths)
+            } catch let error as GitShellError {
+                guard error.result.exitStatus == .terminated(code: 1) else {
+                    throw GitRepositoryError(path: self.path, message: "unable to check ignored files", result: error.result)
                 }
-
-                let args = [
-                    Git.tool, "-C", self.path.pathString.spm_shellEscaped(),
-                    "check-ignore", "-z", "--stdin",
-                    "<", pathsFile.path.pathString.spm_shellEscaped(),
-                ]
-                let argsWithSh = ["sh", "-c", args.joined(separator: " ")]
-                let result = try Process.popen(arguments: argsWithSh)
-                let output = try result.output.get()
-
-                let outputs: [String] = output.split(separator: 0).map { String(decoding: $0, as: Unicode.UTF8.self) }
-
-                guard result.exitStatus == .terminated(code: 0) || result.exitStatus == .terminated(code: 1) else {
-                    throw GitInterfaceError.fatalError
-                }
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

neonichu (approved):
💯 branch name 🐢

Sources/SourceControl/GitRepository.swift
```diff
@@ -30,7 +30,7 @@ private struct GitShellHelper {
     /// Private function to invoke the Git tool with its default environment and given set of arguments.  The specified
     /// failure message is used only in case of error.  This function waits for the invocation to finish and returns the
     /// output as a string.
-    func run(_ args: [String], environment: EnvironmentVariables = Git.environment, outputRedirection: Process.OutputRedirection = .collect) throws -> String {
+    func run(_ args: [String], environment: EnvironmentVariables = Git.environment, outputRedirection: Process.OutputRedirection = .collect, permitError: Bool = false) throws -> String {
```
tomerd, line 33: this change feels redundant? cant the call site catch the `GitShellError` and extract what it needs from the underlying result?

compnerd: It cannot, because if the exit code is non-zero, an error is thrown and you lose the output.

tomerd: iirc, it throws `GitShellError` with the result embedded and the output can be read from the result?

compnerd: Oh, I see; interesting thought; yes, that would be viable.  But we would need something like this at the callee:

```swift
do {
  let output: String
  do {
    output = try self.git.run(["-C", self.path.pathString, "check-ignore"] + stringPaths, permitError: true).split(separator: "\n")
  } catch let error as GitShellError {
    guard result.exitStatus == .terminated(code: 1) else {
      throw error
    }
    output = error.result.utf8output().spm_chomp()
  }
  return output.map {
    let string = String($0).replacingOccurrences(of: "\\\\", with: "\\")
    if string.utf8.first == UInt8(ascii: "\"") {
      return String(string.dropFirst(1).dropLast(1))
    }
    return string
  }
} catch let error as GitShellError {
  throw GitRepositoryError(path: self.path, message: "unable to check ignored files", result: error.result)
}
```

I find that to be a little bit worse than the new parameter, but if you feel that is better, I can switch to that.  Or is there a better way to write that?

tomerd: In principle I think making the call site more complicated is better than making a generic function more complicated for a single known use case, tho obviously its a question of style.

more pragmatically, we can probably get rid of the double do/catch:

```swift
  let output: String
  do {
    output = try self.git.run(["-C", self.path.pathString, "check-ignore"] + stringPaths)
  } catch let error as GitShellError {
    guard result.exitStatus == .terminated(code: 1) else {
      throw GitRepositoryError(path: self.path, message: "unable to check ignored files", result: error.result)
    }
    output = error.result.utf8output().spm_chomp()
  }
  return output.split(separator: "\n").map {
    let string = String($0).replacingOccurrences(of: "\\\\", with: "\\")
    if string.utf8.first == UInt8(ascii: "\"") {
      return String(string.dropFirst(1).dropLast(1))
    }
    return string
  }
```

compnerd: This sounds reasonable to me.

compnerd:
@swift-ci please smoke test

tomerd (approved):

PR #5501, opened 2022-05-13
PackageLoading: correct relative path construction

compnerd:
We would construct a `RelativePath` without rooting it, which would then
be normalized and thus cause problems.  Alter the path to validate the
path is relative and then construct the `AbsolutePath` by rooting the
relative path.

### DIFF
--- a/Sources/PackageLoading/PackageBuilder.swift
+++ b/Sources/PackageLoading/PackageBuilder.swift
@@ -877,9 +877,10 @@ public final class PackageBuilder {
                 }
 
                 // Ensure that the search path is contained within the package.
-                let subpath = try RelativePath(validating: value)
-                guard targetRoot.appending(subpath).isDescendantOfOrEqual(to: packagePath) else {
-                    throw ModuleError.invalidHeaderSearchPath(subpath.pathString)
+                _ = try RelativePath(validating: value)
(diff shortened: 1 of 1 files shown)


Sources/PackageLoading/PackageBuilder.swift
```diff
@@ -877,9 +877,10 @@ public final class PackageBuilder {
                 }
 
                 // Ensure that the search path is contained within the package.
-                let subpath = try RelativePath(validating: value)
-                guard targetRoot.appending(subpath).isDescendantOfOrEqual(to: packagePath) else {
-                    throw ModuleError.invalidHeaderSearchPath(subpath.pathString)
+                _ = try RelativePath(validating: value)
```
tomerd, line 880: same comment as https://github.com/apple/swift-package-manager/pull/5484/files#r870728183

this change is too subtle and likely to be reverted by future maintainers as the intuitive thing to do is to note ignore the relative path returned by the constructor. we really do need to address the underlying issue first.

tomerd:
@compnerd can this be closed given https://github.com/apple/swift-package-manager/pull/5910?

compnerd:
Yeah, I think that we can close this PR, I checked the current PackageLoading test failures - they seem to be related to path normalisation rather than relative path construction.  Thanks for the reminder!

PR #5509, opened 2022-05-14
[5.7] Update conflicting alias detection

elsh:
Besides different alias values, we also need to detect if there's no alias for the same key.
Resolves rdar://92990727


elsh:
@swift-ci smoke test

tomerd (approved):

PR #4333, opened 2022-04-29
[wip] add workspace state file watcher to reload workspace state when necessary

tomerd:
motivation:  long running processes (ie IDEs) need to watch in case other SwiftPM processes (ie CLI) made changes to the state

changes:
* create generic abstration for file watching
* add file watcher to workspace state, reloading the state when it detects the file is different from last known one
* add test
* NFC: move some code around in Workspace so its more readable

TODO: 
* more testing
* reload notification (to help drive testing)
* further cleanup

PR #5510, opened 2022-05-14
[alternative 2] reload workspace state before reloading package graph

tomerd:
motivation: long running host processes (ie IDEs) need this in case other SwiftPM processes (ie CLI) made changes to the state

changes:
* call reload state before reloading the graph
* add test

### DIFF
--- a/Sources/Workspace/Workspace+State.swift
+++ b/Sources/Workspace/Workspace+State.swift
@@ -65,13 +65,21 @@ public final class WorkspaceState {
         try self.save()
     }
 
+    // marked public for testing
     public func save() throws {
         try self.storage.save(dependencies: self.dependencies, artifacts: self.artifacts)
     }
 
     /// Returns true if the state file exists on the filesystem.
-    public func stateFileExists() -> Bool {
+    func stateFileExists() -> Bool {
         return self.storage.fileExists()
+    }
+
+    /// Returns true if the state file exists on the filesystem.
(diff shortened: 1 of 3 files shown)


tomerd:
@ahoppen this is a second alternative to https://github.com/apple/swift-package-manager/pull/4333

it is more minimal and less risky, and I suspect it solves the issue just as well. could you give it a try?

ahoppen:
Just tried it and it also resolves the issue 👍 I’m fine with either this alternative or the other.

tomerd:
@swift-ci smoke test

abertelrud (approved):
Looks good to me, not only given how IDEs currently interact with libSwiftPM (watching file system themselves and reloading the package graph) but also to make sure state is loaded on first load.

Sources/Workspace/Workspace+State.swift
```diff
@@ -65,14 +65,22 @@ public final class WorkspaceState {
...
         try self.storage.save(dependencies: self.dependencies, artifacts: self.artifacts)
     }
 
     /// Returns true if the state file exists on the filesystem.
-    public func stateFileExists() -> Bool {
+    func stateFileExists() -> Bool {
         return self.storage.fileExists()
     }
+
+    /// Returns true if the state file exists on the filesystem.
```
mamouneyya, line 78: Looks like you missed updating the doc comment.

PR #5513, opened 2022-05-15
Tests: use `tar` rather than `echo` for sandbox test on Windows

compnerd:
`echo` is a `cmd` shell builtin and not available as a standalone binary
as from coretuils on Linux.  It should be noted that `echo` is available
as a shell builtin on Linux and is not required to be provided as a core
binary.  `tar` is only available on more recent Windows releases with 10
or newer.


compnerd:
@swift-ci please smoke test

neonichu (approved):

tomerd (approved):

PR #5516, opened 2022-05-16
Tests: disable SQLite test on Windows

compnerd:
Windows does not permit deleting files with open handles.  This test
appears to rely on those semantics (expecting that the deletion
operation will succeed and the file will be marked for delete on close).
Because Windows does not provide such semantics for `rm`, simply skip
the test on Windows.

### DIFF
--- a/Tests/BasicsTests/SQLiteBackedCacheTests.swift
+++ b/Tests/BasicsTests/SQLiteBackedCacheTests.swift
@@ -55,8 +55,11 @@ final class SQLiteBackedCacheTests: XCTestCase {
     }
 
     func testFileDeleted() throws {
+#if os(Windows)
+        try XCTSkipIf(true, "open file cannot be deleted on Windows")
+#endif
         try XCTSkipIf(is_tsan_enabled())
-        
+
         try testWithTemporaryDirectory { tmpPath in
             let path = tmpPath.appending(component: "test.db")
             let cache = SQLiteBackedCache<String>(tableName: "SQLiteBackedCacheTest", path: path)
--- a/Tests/PackageCollectionsTests/PackageCollectionsStorageTests.swift
+++ b/Tests/PackageCollectionsTests/PackageCollectionsStorageTests.swift
@@ -73,8 +73,11 @@ class PackageCollectionsStorageTests: XCTestCase {
     }
 
     func testFileDeleted() throws {
+#if os(Windows)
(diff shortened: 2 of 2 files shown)


compnerd:
@swift-ci please smoke test

tomerd (approved):

abertelrud:
In general Darwin allows removal of open files, in the usually Unix way.  But I think SQLite databases are special, actually, and backing the file(s) shouldn't be removed while open, so I'd be surprised if this test relies on it.  At the very least there would be console messages warning about it.

compnerd:
I wonder if I misunderstood the test then.

I believe that the problem comes from:
https://github.com/apple/swift-package-manager/blob/main/Tests/BasicsTests/SQLiteBackedCacheTests.swift#L80

That will actually end up throwing on Windows with:
```
threw error "Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission.""
```

Is that operation not expected to succeed but retain the file?

tomerd:
this is a test for resiliency of the cache which is SQLite based and can run into situation where the underlying SQLite file is deleted Ryan external process

Tests/BasicsTests/SQLiteBackedCacheTests.swift
```diff
@@ -55,8 +55,11 @@ final class SQLiteBackedCacheTests: XCTestCase {
     }
 
     func testFileDeleted() throws {
+#if os(Windows)
+        throw XCTSkip("open file cannot be deleted on Windows")
+#else
```
tomerd, line 60: why do we need the else? does the code not compile windows?

compnerd: Avoids a warning - "code after throw will never be executed"

tomerd: we have allot of tests with the following:

```
#if !os(macOS)
try XCTSkipIf(true, "test is only supported on macOS")
#endif
```

does that help?

compnerd: Sure, that seems to avoid the warning, as long as we can agree to disagree (I personally find `XCTSkipIf(true, ...` terrible).

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

PR #5517, opened 2022-05-16
PackageLoading: support non-Unix platforms for PkgConfig

compnerd:
pkg-config is a common utility on UNIX systems, but does not play very
well with Windows.  This adjusts the path separator character to match
the platform as well as uses `AbsolutePath` APIs to work with paths as
a means of getting the platform specific path separator character.  It
also introduces the ability to specify the pkgconfig tool itself which
is needed for Windows which does not support an arbitrary executable.

### DIFF
--- a/Sources/PackageLoading/PkgConfig.swift
+++ b/Sources/PackageLoading/PkgConfig.swift
@@ -123,7 +123,11 @@ public struct PkgConfig {
 
     private static var envSearchPaths: [AbsolutePath] {
         if let configPath = ProcessEnv.vars["PKG_CONFIG_PATH"] {
+#if os(Windows)
+            return configPath.split(separator: ";").map({ AbsolutePath(String($0)) })
+#else
             return configPath.split(separator: ":").map({ AbsolutePath(String($0)) })
+#endif
         }
         return []
     }
@@ -174,7 +178,7 @@ internal struct PkgConfigParser {
         variables["pcfiledir"] = pcFile.parentDirectory.pathString
 
         // Add pc_sysrootdir variable. This is the path of the sysroot directory for pc files.
-        variables["pc_sysrootdir"] = ProcessEnv.vars["PKG_CONFIG_SYSROOT_DIR"] ?? "/"
+        variables["pc_sysrootdir"] = ProcessEnv.vars["PKG_CONFIG_SYSROOT_DIR"] ?? AbsolutePath.root.pathString
 
         let fileContents: String = try fileSystem.readFileContents(pcFile)
         for line in fileContents.components(separatedBy: "\n") {
@@ -382,18 +386,24 @@ internal struct PCFileFinder {
     ///
     /// This is needed because on Linux machines, the search paths can be different
     /// from the standard locations that we are currently searching.
-    public init(brewPrefix: AbsolutePath? = .none) {
+    public init(brewPrefix: AbsolutePath? = .none, pkgConfig: AbsolutePath? = .none) {
         if PCFileFinder.pkgConfigPaths == nil {
             do {
                 let pkgConfigPath: String
-                if let brewPrefix = brewPrefix {
+                if let pkgConfig = pkgConfig {
+                    pkgConfigPath = pkgConfig.pathString
+                } else if let brewPrefix = brewPrefix {
                     pkgConfigPath = brewPrefix.appending(components: "bin", "pkg-config").pathString
                 } else {
                     pkgConfigPath = "pkg-config"
                 }
                 let searchPaths = try Process.checkNonZeroExit(
                 args: pkgConfigPath, "--variable", "pc_path", "pkg-config").spm_chomp()
+#if os(Windows)
+                PCFileFinder.pkgConfigPaths = searchPaths.split(separator: ";").map { AbsolutePath(String($0)) }
+#else
                 PCFileFinder.pkgConfigPaths = searchPaths.split(separator: ":").map({ AbsolutePath(String($0)) })
+#endif
             } catch {
                 PCFileFinder.shouldEmitPkgConfigPathsDiagnostic = true
                 PCFileFinder.pkgConfigPaths = []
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please smoke test

Sources/PackageLoading/PkgConfig.swift
```diff
@@ -171,10 +175,10 @@ internal struct PkgConfigParser {
         }
 
         // Add pcfiledir variable. This is the path of the directory containing this pc file.
-        variables["pcfiledir"] = pcFile.parentDirectory.pathString
+        variables["pcfiledir"] = AbsolutePath(pcFile.parentDirectory.pathString).pathString
```
tomerd, line 178: ?

compnerd: I ... um ... maybe I was asleep?

Tests/PackageLoadingTests/PkgConfigParserTests.swift
```diff
@@ -117,22 +117,22 @@ final class PkgConfigParserTests: XCTestCase {
             "/usr/local/opt/foo/lib/pkgconfig/foo.pc",
             "/custom/foo.pc")
         XCTAssertEqual(
-            "/custom/foo.pc",
+            AbsolutePath("/custom/foo.pc").pathString,
```
tomerd, line 120: drop the `. pathString` on both ends to compare the `AbsolutePath` instead of the string version

Tests/PackageLoadingTests/PkgConfigParserTests.swift
```diff
@@ -117,22 +117,22 @@ final class PkgConfigParserTests: XCTestCase {
...
             try PCFileFinder().locatePCFile(name: "foo", customSearchPaths: [AbsolutePath("/custom")], fileSystem: fs, observabilityScope: observability.topScope).pathString
         )
         XCTAssertEqual(
-            "/custom/foo.pc",
+            AbsolutePath("/custom/foo.pc").pathString,
```
tomerd, line 124: ditto

Tests/PackageLoadingTests/PkgConfigParserTests.swift
```diff
@@ -117,22 +117,22 @@ final class PkgConfigParserTests: XCTestCase {
...
             try PkgConfig(name: "foo", additionalSearchPaths: [AbsolutePath("/custom")], fileSystem: fs, observabilityScope: observability.topScope).pcFile.pathString
         )
         XCTAssertEqual(
-            "/usr/lib/pkgconfig/foo.pc",
+            AbsolutePath("/usr/lib/pkgconfig/foo.pc").pathString,
             try PCFileFinder().locatePCFile(name: "foo", customSearchPaths: [], fileSystem: fs, observabilityScope: observability.topScope).pathString
         )
         try withCustomEnv(["PKG_CONFIG_PATH": "/usr/local/opt/foo/lib/pkgconfig"]) {
-            XCTAssertEqual("/usr/local/opt/foo/lib/pkgconfig/foo.pc", try PkgConfig(name: "foo", fileSystem: fs, observabilityScope: observability.topScope).pcFile.pathString)
+            XCTAssertEqual(AbsolutePath("/usr/local/opt/foo/lib/pkgconfig/foo.pc").pathString, try PkgConfig(name: "foo", fileSystem: fs, observabilityScope: observability.topScope).pcFile.pathString)
```
tomerd, line 132: ditto

Tests/PackageLoadingTests/PkgConfigParserTests.swift
```diff
@@ -117,22 +117,22 @@ final class PkgConfigParserTests: XCTestCase {
...
         }
         try withCustomEnv(["PKG_CONFIG_PATH": "/usr/local/opt/foo/lib/pkgconfig:/usr/lib/pkgconfig"]) {
-            XCTAssertEqual("/usr/local/opt/foo/lib/pkgconfig/foo.pc", try PkgConfig(name: "foo", fileSystem: fs, observabilityScope: observability.topScope).pcFile.pathString)
+            XCTAssertEqual(AbsolutePath("/usr/local/opt/foo/lib/pkgconfig/foo.pc").pathString, try PkgConfig(name: "foo", fileSystem: fs, observabilityScope: observability.topScope).pcFile.pathString)
```
tomerd, line 135: ditto

Sources/PackageLoading/PkgConfig.swift
```diff
@@ -382,18 +386,24 @@ internal struct PCFileFinder {
     ///
     /// This is needed because on Linux machines, the search paths can be different
     /// from the standard locations that we are currently searching.
-    public init(brewPrefix: AbsolutePath? = .none) {
+    public init(brewPrefix: AbsolutePath? = .none, pkgConfig: AbsolutePath? = .none) {
```
tomerd, line 389: is this initializer only used in tests?

compnerd: Yes, it is only used in tests; I believe that `init(brewPrefix:)` was initially for tests only.  If not, that might be something that warrants re-visiting.

tomerd: if so lets make it internal rather than public? and maybe add a comment to say its for testing?

compnerd: Wait, maybe I'm misunderstanding something.  The type itself is already marked `internal`.  Does not make this an internal interface?  That was why I did this change this way - otherwise this is a problem as it is an ABI break and we should provide a function overload rather than a default valued parameter introduction.

compnerd: Wait, maybe I'm misunderstanding something. The type itself is already marked `internal`. Does not make this an internal interface? That was why I did this change this way - otherwise this is a problem as it is an ABI break and we should provide a function overload rather than a default valued parameter introduction.

compnerd:
@swift-ci please smoke test

Sources/PackageLoading/PkgConfig.swift
tomerd: yes +1

PR #5518, opened 2022-05-16
Tests: adopt `AbsolutePath` APIs for path construction

compnerd:
Update the test expectations to use `AbsolutePath` for the construction
of the path literals in order to support non-`/` path separator
characters.  This allows these test expectations to match on Windows.

### DIFF
--- a/Tests/PackageLoadingTests/PackageBuilderTests.swift
+++ b/Tests/PackageLoadingTests/PackageBuilderTests.swift
@@ -41,9 +41,12 @@ class PackageBuilderTests: XCTestCase {
     }
 
     func testMixedSources() throws {
+        let foo: AbsolutePath = AbsolutePath("/Sources/foo")
+
         let fs = InMemoryFileSystem(emptyFiles:
-            "/Sources/foo/main.swift",
-            "/Sources/foo/main.c")
+            foo.appending(components: "main.swift").pathString,
+            foo.appending(components: "main.c").pathString
+        )
 
         let manifest = Manifest.createRootManifest(
             name: "pkg",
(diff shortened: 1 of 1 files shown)


tomerd (approved):
thanks for doing this @compnerd 

one small follow up cleanup to this would be to change the testing APIs (eg `checkSources` , `checkResource`, `InMemoryFileSystem` initializer, etc) to take `AbsolutePath` instead or in addition to String which will make the call sites nicer - can drop the `.pathString` everywhere

compnerd:
@swift-ci please smoke test

compnerd:
@tomerd ooh, I do like that clean up idea; and I agree that it is better suited as a follow up change

Tests/PackageLoadingTests/PackageBuilderTests.swift
```diff
@@ -41,9 +41,12 @@ class PackageBuilderTests: XCTestCase {
     }
 
     func testMixedSources() throws {
+        let foo: AbsolutePath = AbsolutePath("/Sources/foo")
```
abertelrud, line 44: Quick question, was the repetition of the type needed, or just stylistic?  Just curious — in theory you just need the `=` for all of these, not the `:`'.

compnerd: Just stylistic; I tend to use explicit types for my own code, and it has become habitual.  I don't mind dropping the type annotations if you prefer.

abertelrud: In SwiftPM we've tended to not spell out types (to the point of using `.init()` instead of mentioning the type when it's obvious, which took me some time to get used to).  If you don't mind doing that it would be great, but I don't think it's a huge deal.  I think @tomerd runs swift-format on his changes, I don't know if it would catch this.

In any case, I realize this is a fairly trivial comment compared to the improvements of this PR.

tomerd: yea `let foo = AbsolutePath("/Sources/foo")` is the most idiomatic afaik

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

PR #5520, opened 2022-05-17
Tests: remove errant `print`

compnerd:
This is not related to the test and simply adds noise to the output.


compnerd:
@swift-ci please smoke test

tomerd (approved):

PR #5523, opened 2022-05-18
PackageLoading: introduce new entry points for testing

compnerd:
Windows does not have a distributed pkg-config installation and we need
to stub that out for tests.

### DIFF
--- a/Sources/PackageLoading/PkgConfig.swift
+++ b/Sources/PackageLoading/PkgConfig.swift
@@ -382,23 +382,27 @@ internal struct PCFileFinder {
     ///
     /// This is needed because on Linux machines, the search paths can be different
     /// from the standard locations that we are currently searching.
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please smoke test

Sources/PackageLoading/PkgConfig.swift
```diff
@@ -401,6 +397,14 @@ internal struct PCFileFinder {
         }
     }
 
+    public init(brewPrefix: AbsolutePath?) {
+        self.init(pkgConfigPath: brewPrefix?.pathString ?? "pkg-config")
```
tomerd, line 401: did we loose the logic in L390 (` brewPrefix.appending(components: "bin", "pkg-config").pathString`)?

tomerd (approved):

PR #5524, opened 2022-05-18
Add validation checking for having plugin targets in library or executable products

abertelrud:
### Motivation:

Plugin targets can only be in plugin products, but right now there is no validation for this.  The plugin targets just get ignored, which is confusing.

### Modifications:

- add validation for library products
- clarified the product name in the description of libraries (e.g. "dynamic library" instead of just "dynamic" for the type of product)
- make both executable and library products check that there are no plugin targets
- add a new unit test for libraries, and modify one for executables

We could extend this to other kinds of products in the future, but libraries and executables are the ones that accept targets today.

### DIFF
--- a/Sources/PackageLoading/Diagnostics.swift
+++ b/Sources/PackageLoading/Diagnostics.swift
@@ -56,6 +56,10 @@ extension Basics.Diagnostic {
         .error("system library product \(product) shouldn't have a type and contain only one target")
     }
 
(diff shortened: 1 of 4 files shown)


abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

tomerd (approved):

neonichu (approved):

PR #5526, opened 2022-05-18
Tests: adjust VFS tests for Windows

compnerd:
Windows requires that we have the destination exist before we create the
file to disambiguate between the file and directory symbolic link case.
Swap the ordering of the fixture creation to accommodate this.

The `isExecutable` check on Windows is more restrictive, identifying
only binary executables rather than scripts.  As a result, we must
disable one of the checks.

### DIFF
--- a/Tests/BasicsTests/VFSTests.swift
+++ b/Tests/BasicsTests/VFSTests.swift
@@ -16,9 +16,22 @@ import XCTest
 
 class VFSTests: XCTestCase {
     func testLocalBasics() throws {
+        // tiny PE binary from: https://archive.is/w01DO
+        let contents: [UInt8] = [
+          0x4d, 0x5a, 0x00, 0x00, 0x50, 0x45, 0x00, 0x00, 0x4c, 0x01, 0x01, 0x00,
+          0x6a, 0x2a, 0x58, 0xc3, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+          0x04, 0x00, 0x03, 0x01, 0x0b, 0x01, 0x08, 0x00, 0x04, 0x00, 0x00, 0x00,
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

Tests/BasicsTests/VFSTests.swift
```diff
@@ -70,8 +73,10 @@ class VFSTests: XCTestCase {
             // isExecutableFile
             let executablePath = AbsolutePath("/exec-foo")
             let executableSymPath = AbsolutePath("/exec-sym")
+#if !os(Windows)
             XCTAssertTrue(vfs.isExecutableFile(executablePath))
             XCTAssertTrue(vfs.isExecutableFile(executableSymPath))
+#endif
```
tomerd, line 79: this disables both checks, expected?

compnerd: Yes, we aren't generating binaries, so the test is going to fail.  Should the check change, it would suddenly change, so it is safer to just disable the two checks here on Windows I think.

tomerd: doesn't this mean that the feature would be broken on windows though?

compnerd: No, there is no concept of an "executable bit" in Windows.  `chmod +x` is meaningless, it does nothing.  The only indicator of an executable is that it is a COM executable.  This is validating that a file marked with +x is serialized as executable.  We _could_ replace the place holder files with tiny DOS executables though I suppose.  I did do this for swift-driver.

tomerd: gotcha. my meta point is that we need to make sure VFS works correctly on Window too including the `isExecutableFile` API

compnerd: That makes sense.  I can dig out the binary and embed that in the test case.  The other platforms can simply use chmod on it for the test.

compnerd:
@swift-ci please smoke test

tomerd (approved):
😎

PR #5527, opened 2022-05-19
PackageLoading: further portability changes for pkg-config

compnerd:
Windows and subsequently `pkg-config` (and `pkgconf`) use `;` as the
path delimiter rather than `:`.  Adjust the codepaths to account for
the difference in the sigil.  Additionally, because `\` is the preferred
path separator on windows, the pkg-config specification behaves slightly
differently for the treatment of `\`.  Unless we are escaping spaces or
the `\` literal, emit the `\` as a character rather than treat it as an
escape modifier.  This gives us the desired behaviour, even though it is
a slight bit more clever than the reference implementations.

### DIFF
--- a/Sources/PackageLoading/PkgConfig.swift
+++ b/Sources/PackageLoading/PkgConfig.swift
@@ -123,7 +123,11 @@ public struct PkgConfig {
 
     private static var envSearchPaths: [AbsolutePath] {
         if let configPath = ProcessEnv.vars["PKG_CONFIG_PATH"] {
+#if os(Windows)
+            return configPath.split(separator: ";").map({ AbsolutePath(String($0)) })
+#else
             return configPath.split(separator: ":").map({ AbsolutePath(String($0)) })
+#endif
         }
         return []
     }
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please smoke test

Sources/PackageLoading/PkgConfig.swift
```diff
@@ -123,7 +123,11 @@ public struct PkgConfig {
 
     private static var envSearchPaths: [AbsolutePath] {
         if let configPath = ProcessEnv.vars["PKG_CONFIG_PATH"] {
+#if os(Windows)
+            return configPath.split(separator: ";").map({ AbsolutePath(String($0)) })
```
abertelrud, line 127: Would it make sense to factor this out into shared code?  If it's guaranteed to always use the same delimiter as `PATH`, then I think there's already code in UserToolchain that does this splitting.  Looks like there's something similar also in EnvironmentVariables.

In general I'm wary of oversharing of this kind of code (especially quoting, which really needs to be tailored to its use, e.g. quoting for a module map might not follow the same rules as for a shell, etc).  But if `pkg-config` already defines this to be the same as `PATH` delimiters, then it would make sense to share the implementation here, i.e. find a way to use the one in the UserToolchain.

compnerd: I think that sharing that does make sense.  I am slightly tempted to say that this is a good follow up change and should be separate from this change as this maintains the status quo structure.

tomerd (approved):

PR #5532, opened 2022-05-20
Create helper functions for temporary directories which supports new concurrency features

Overcot:
Create helper functions for creating temporary files in order to slowly migrate to async/await based approach to concurrency

### Motivation:

There are examples such as in ManifestLoading where async/await approach would be helpful to migrate from completionHandlers to async/await

### Modifications:

As mentioned in https://github.com/apple/swift-tools-support-core/pull/321#issuecomment-1132432209 it was advised to put this code inside SwiftPM/Basics module in order to not increase amount of code in TSCBasics.

### Result:

After that i would also create helper functions for creating temporary files, and maybe start to further investigate what kind of limitations exist to adopt async/await to manifest parsing. As far i can see `Process` also have to be migrated

### DIFF
--- a/Sources/Basics/TemporaryFile.swift
+++ b/Sources/Basics/TemporaryFile.swift
@@ -35,6 +35,7 @@ import TSCLibc
 public func withTemporaryDirectory<Result>(
     dir: AbsolutePath? = nil,
     prefix: String = "TemporaryDirectory",
+    fileSystem: FileSystem = localFileSystem,
     _ body: @escaping (AbsolutePath, @escaping (AbsolutePath) -> Void) async throws -> Result
 ) throws -> Task<Result, Error> {
     let temporaryDirectory = try createTemporaryDirectory(dir: dir, prefix: prefix)
@@ -42,10 +43,10 @@ public func withTemporaryDirectory<Result>(
     let task: Task<Result, Error> = Task {
         try await withTaskCancellationHandler {
             try await body(temporaryDirectory) { path in
-                _ = try? FileManager.default.removeItem(atPath: path.pathString)
+                try? fileSystem.removeFileTree(path)
             }
         } onCancel: {
-            _ = try? FileManager.default.removeItem(atPath: temporaryDirectory.pathString)
+            try? fileSystem.removeFileTree(temporaryDirectory)
         }
         
     }
@@ -73,10 +74,11 @@ public func withTemporaryDirectory<Result>(
 public func withTemporaryDirectory<Result>(
     dir: AbsolutePath? = nil,
     prefix: String = "TemporaryDirectory",
+    fileSystem: FileSystem = localFileSystem,
     removeTreeOnDeinit: Bool = false,
     _ body: @escaping (AbsolutePath) async throws -> Result
 ) throws -> Task<Result, Error> {
-    try withTemporaryDirectory(dir: dir, prefix: prefix) { path, cleanup in
+    try withTemporaryDirectory(dir: dir, prefix: prefix, fileSystem: fileSystem) { path, cleanup in
         defer { if removeTreeOnDeinit { cleanup(path) } }
         return try await body(path)
     }
--- a/Sources/Basics/TemporaryFile.swift
+++ b/Sources/Basics/TemporaryFile.swift
@@ -33,12 +33,14 @@ import TSCLibc
 /// - Throws: `MakeDirectoryError` and rethrows all errors from `body`.
 @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
 public func withTemporaryDirectory<Result>(
-  dir: AbsolutePath? = nil, prefix: String = "TemporaryDirectory" , _ body: (AbsolutePath, @escaping (AbsolutePath) -> Void) async throws -> Result
+    dir: AbsolutePath? = nil,
+    prefix: String = "TemporaryDirectory",
+    _ body: (AbsolutePath, @escaping (AbsolutePath) -> Void) async throws -> Result
 ) async throws -> Result {
-    let template = try createTemporaryDirectoryTemplate(dir: dir, prefix: prefix)
-  return try await body(AbsolutePath(String(cString: template))) { path in
-    _ = try? FileManager.default.removeItem(atPath: path.pathString)
-  }
+    let temporaryDirectory = try createTemporaryDirectory(dir: dir, prefix: prefix)
+    return try await body(temporaryDirectory) { path in
+        _ = try? FileManager.default.removeItem(atPath: path.pathString)
+    }
 }
 
 /// Creates a temporary directory and evaluates a closure with the directory path as an argument.
@@ -59,27 +61,30 @@ public func withTemporaryDirectory<Result>(
 /// - Throws: `MakeDirectoryError` and rethrows all errors from `body`.
 @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
 public func withTemporaryDirectory<Result>(
-  dir: AbsolutePath? = nil, prefix: String = "TemporaryDirectory", removeTreeOnDeinit: Bool = false , _ body: (AbsolutePath) async throws -> Result
+    dir: AbsolutePath? = nil,
+    prefix: String = "TemporaryDirectory",
+    removeTreeOnDeinit: Bool = false ,
+    _ body: (AbsolutePath) async throws -> Result
 ) async throws -> Result {
-  try await withTemporaryDirectory(dir: dir, prefix: prefix) { path, cleanup in
-    defer { if removeTreeOnDeinit { cleanup(path) } }
-    return try await body(path)
-  }
+    try await withTemporaryDirectory(dir: dir, prefix: prefix) { path, cleanup in
+        defer { if removeTreeOnDeinit { cleanup(path) } }
+        return try await body(path)
+    }
 }
 
-private func createTemporaryDirectoryTemplate(dir: AbsolutePath?, prefix: String) throws -> [Int8] {
+private func createTemporaryDirectory(dir: AbsolutePath?, prefix: String) throws -> AbsolutePath {
     // Construct path to the temporary directory.
     let templatePath = try AbsolutePath(prefix + ".XXXXXX", relativeTo: determineTempDirectory(dir))
-
+    
     // Convert templatePath to a C style string terminating with null char to be an valid input
     // to mkdtemp method. The XXXXXX in this string will be replaced by a random string
     // which will be the actual path to the temporary directory.
-    var template = [UInt8](templatePath.pathString.utf8).map({ Int8($0) }) + [Int8(0)]
-
+    var template = templatePath.pathString.utf8CString.map { $0 }
+    
     if TSCLibc.mkdtemp(&template) == nil {
         throw MakeDirectoryError(errno: errno)
     }
-    return template
+    return AbsolutePath(String(cString: template))
 }
 
 private extension MakeDirectoryError {
--- a/Sources/Basics/TemporaryFile.swift
+++ b/Sources/Basics/TemporaryFile.swift
@@ -38,7 +38,7 @@ public func withTemporaryDirectory<Result>(
     fileSystem: FileSystem = localFileSystem,
     _ body: @escaping (AbsolutePath, @escaping (AbsolutePath) -> Void) async throws -> Result
 ) throws -> Task<Result, Error> {
-    let temporaryDirectory = try createTemporaryDirectory(dir: dir, prefix: prefix)
+    let temporaryDirectory = try createTemporaryDirectory(dir: dir, prefix: prefix, fileSystem: fileSystem)
     
     let task: Task<Result, Error> = Task {
         try await withTaskCancellationHandler {
@@ -84,19 +84,18 @@ public func withTemporaryDirectory<Result>(
     }
 }
 
-private func createTemporaryDirectory(dir: AbsolutePath?, prefix: String) throws -> AbsolutePath {
+private func createTemporaryDirectory(dir: AbsolutePath?, prefix: String, fileSystem: FileSystem) throws -> AbsolutePath {
+    // This random generation is needed so that
+    // it is more or less equal to generation using `mkdtemp` function
+    let letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+
+    let randomSuffix = String((0..<6).map { _ in letters.randomElement()! })
+    
     // Construct path to the temporary directory.
-    let templatePath = try AbsolutePath(prefix + ".XXXXXX", relativeTo: determineTempDirectory(dir))
-    
-    // Convert templatePath to a C style string terminating with null char to be an valid input
-    // to mkdtemp method. The XXXXXX in this string will be replaced by a random string
-    // which will be the actual path to the temporary directory.
-    var template = templatePath.pathString.utf8CString.map { $0 }
-    
-    if TSCLibc.mkdtemp(&template) == nil {
-        throw MakeDirectoryError(errno: errno)
-    }
-    return AbsolutePath(String(cString: template))
+    let templatePath = try AbsolutePath(prefix + ".\(randomSuffix)", relativeTo: determineTempDirectory(dir))
+
+    try fileSystem.createDirectory(templatePath, recursive: true)
+    return templatePath
 }
 
 private extension MakeDirectoryError {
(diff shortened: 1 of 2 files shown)


Overcot:
One of questions is that do library needs at some point to migrate `Process` and other required types inside SwiftPM?

tomerd:
we would like to move the process handling to SwiftPM and have it be based on Foundation's process as much as possible.

Overcot:
> we would like to move the process handling to SwiftPM and have it be based on Foundation's process as much as possible.

@tomerd so can i tackle on this transition? I thought about first slowly migrating bits of code regarding creating temporary file creation and creating series of PRs to create small incremental changes

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,106 @@
...
+}
+
+private func createTemporaryDirectoryTemplate(dir: AbsolutePath?, prefix: String) throws -> [Int8] {
+    // Construct path to the temporary directory.
+    let templatePath = try AbsolutePath(prefix + ".XXXXXX", relativeTo: determineTempDirectory(dir))
+
+    // Convert templatePath to a C style string terminating with null char to be an valid input
+    // to mkdtemp method. The XXXXXX in this string will be replaced by a random string
+    // which will be the actual path to the temporary directory.
+    var template = [UInt8](templatePath.pathString.utf8).map({ Int8($0) }) + [Int8(0)]
```
tomerd, line 77: I suspect there are better way these days to get the c representation of a Swift string. @milseman ?

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,106 @@
...
+///     - prefix: The prefix to the temporary file name.
+///     - body: A closure to execute that receives the absolute path of the directory as an argument.
+///           If `body` has a return value, that value is also used as the
+///           return value for the `withTemporaryDirectory` function.
+///           The cleanup block should be called when the temporary directory is no longer needed.
+///
+/// - Throws: `MakeDirectoryError` and rethrows all errors from `body`.
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public func withTemporaryDirectory<Result>(
+  dir: AbsolutePath? = nil, prefix: String = "TemporaryDirectory" , _ body: (AbsolutePath, @escaping (AbsolutePath) -> Void) async throws -> Result
```
tomerd, line 36: formatting: for long list of argument, we try to put each argument on its own line:

```suggestion
  dir: AbsolutePath? = nil, 
  prefix: String = "TemporaryDirectory" ,
  handler: (AbsolutePath, @escaping (AbsolutePath) -> Void) async throws -> Result
```

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,106 @@
...
+) async throws -> Result {
+    let template = try createTemporaryDirectoryTemplate(dir: dir, prefix: prefix)
+  return try await body(AbsolutePath(String(cString: template))) { path in
```
tomerd, line 39: formatting? try Xcode automatic formatting or swift-format to achieve consistent formatting with rest of code base

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,106 @@
...
+///     - prefix: The prefix to the temporary file name.
+///     - removeTreeOnDeinit: If enabled try to delete the whole directory tree otherwise remove only if its empty.
+///     - body: A closure to execute that receives the absolute path of the directory as an argument.
+///             If `body` has a return value, that value is also used as the
+///             return value for the `withTemporaryDirectory` function.
+///
+/// - Throws: `MakeDirectoryError` and rethrows all errors from `body`.
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public func withTemporaryDirectory<Result>(
+  dir: AbsolutePath? = nil, prefix: String = "TemporaryDirectory", removeTreeOnDeinit: Bool = false , _ body: (AbsolutePath) async throws -> Result
```
tomerd, line 62: feels like removeTreeOnDeinit should be true by default?

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,106 @@
...
+) async throws -> Result {
+  try await withTemporaryDirectory(dir: dir, prefix: prefix) { path, cleanup in
+    defer { if removeTreeOnDeinit { cleanup(path) } }
+    return try await body(path)
+  }
+}
+
+private func createTemporaryDirectoryTemplate(dir: AbsolutePath?, prefix: String) throws -> [Int8] {
```
tomerd, line 70: should this return a string instead of byte array? the calcite all need a swift string iiuc

tomerd:
> @tomerd so can i tackle on this transition? I thought about first slowly migrating bits of code regarding creating temporary file creation and creating series of PRs to create small incremental changes

yes that would be very helpful!

tomerd:
cc @compnerd

Sources/Basics/TemporaryFile.swift
Overcot: Basically i followed the exact same default argument setup just as in https://github.com/apple/swift-tools-support-core/blob/main/Sources/TSCBasic/TemporaryFile.swift

Sources/Basics/TemporaryFile.swift
Overcot: Yeah, sorry. Smthing went wrong when copying and pasting code. Will fix in next commit

Sources/Basics/TemporaryFile.swift
Overcot: I looked at it once again and i think it would be better to return `AbsolutePath` instead of byte array and string, because call site is actually want to create `AbsolutePath` in the end

Sources/Basics/TemporaryFile.swift
Overcot: Yeah, did find a way. Take a look in a next commit

tomerd:
@swift-ci smoke test

tomerd:
lgtm. @compnerd any feedback?

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,111 @@
...
+) async throws -> Result {
+    try await withTemporaryDirectory(dir: dir, prefix: prefix) { path, cleanup in
+        defer { if removeTreeOnDeinit { cleanup(path) } }
+        return try await body(path)
+    }
+}
+
+private func createTemporaryDirectory(dir: AbsolutePath?, prefix: String) throws -> AbsolutePath {
+    // Construct path to the temporary directory.
+    let templatePath = try AbsolutePath(prefix + ".XXXXXX", relativeTo: determineTempDirectory(dir))
```
compnerd, line 77: Where is `determineTempDirectory`?

compnerd (commented):
One thing that worries me about this is that async functions must be VERY limited on Windows x86_64.  There is a missing operation and so it is possible to exhaust the stack if it is extensive.  Can we add some sort of mechanism for limiting suspension points if this is going to be adopted?

Sources/Basics/TemporaryFile.swift
Overcot: https://github.com/apple/swift-tools-support-core/blob/503e25f60303cf6c483828512ed66be2d87e4cc7/Sources/TSCBasic/TemporaryFile.swift#L55-L61

Sources/Basics/TemporaryFile.swift
milseman: `withCString { ... }`

Overcot:
> One thing that worries me about this is that async functions must be VERY limited on Windows x86_64. There is a missing operation and so it is possible to exhaust the stack if it is extensive. Can we add some sort of mechanism for limiting suspension points if this is going to be adopted?

@compnerd What do you think about exposing Task as result, instead of Result?

The api would be something like this:
```swift
@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
public func withTemporaryDirectory<Result>(
    dir: AbsolutePath? = nil,
    prefix: String = "TemporaryDirectory",
    _ body: @escaping (AbsolutePath, @escaping (AbsolutePath) -> Void) async throws -> Result
) throws -> Task<Result, Error> {}
```
it would not be marked as async and we will receive more granular control over the executing of the body making ourself

compnerd:
Hmm, so the async-ness is encapsulated within the implementation rather than the user's control?  I think that would alleviate my concern - with tests on Windows being close, this should be something that we should be able to avoid blowing through the stack.

Overcot:
Ok, will change semantics of methods and update tests respectively.

Overcot:
So basically i changed apis to return Task instead of async Result and added a test to showcase an explicit task cancellation

Overcot:
@tomerd @compnerd Could you take a look if possible?

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,121 @@
...
+    dir: AbsolutePath? = nil,
+    prefix: String = "TemporaryDirectory",
+    _ body: @escaping (AbsolutePath, @escaping (AbsolutePath) -> Void) async throws -> Result
+) throws -> Task<Result, Error> {
+    let temporaryDirectory = try createTemporaryDirectory(dir: dir, prefix: prefix)
+    
+    let task: Task<Result, Error> = Task {
+        try await withTaskCancellationHandler {
+            try await body(temporaryDirectory) { path in
+                _ = try? FileManager.default.removeItem(atPath: path.pathString)
```
tomerd, line 45: use `localFileSystem`, or even better take a FileSystem abstraction as an argument, instead of FileManager directly

Tests/BasicsTests/TemporaryFileTests.swift
```diff
@@ -0,0 +1,106 @@
...
+                return
+            }
+            await task.value
+            
+            try localFileSystem.writeFileContents(filePath, bytes: ByteString())
+            return tempDirPath
+        }.value
+        XCTAssertTrue(localFileSystem.isDirectory(path2))
+        // Cleanup.
+        try FileManager.default.removeItem(atPath: path2.pathString)
```
tomerd, line 51: use `localFileSystem`, or even better take a FileSystem abstraction everywhere

tomerd:
@swift-ci smoke test

Sources/Basics/TemporaryFile.swift
tomerd: I think we need to add a new version of that that take `FileSystem` instead of harding to `localFileSystem`

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,111 @@
...
+private func createTemporaryDirectory(dir: AbsolutePath?, prefix: String) throws -> AbsolutePath {
+    // Construct path to the temporary directory.
+    let templatePath = try AbsolutePath(prefix + ".XXXXXX", relativeTo: determineTempDirectory(dir))
+    
+    // Convert templatePath to a C style string terminating with null char to be an valid input
+    // to mkdtemp method. The XXXXXX in this string will be replaced by a random string
+    // which will be the actual path to the temporary directory.
+    var template = templatePath.pathString.utf8CString.map { $0 }
+    
+    if TSCLibc.mkdtemp(&template) == nil {
```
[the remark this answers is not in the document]
tomerd: this is where things get real :) iiuc `TSCLibc.mkdtemp` is hard wired to use the local file system. we should do a couple of things here:
1. change `createTemporaryDirectory` to also take a `FileSystem` to operate on (and wire it up in the call sites)
2. replace TSCLibc.mkdtemp with something more robust, potentially make it new API of `FileSystem`, so that we can have flexibility around testing and mocking it. The one in `LocalFileSystem` (from TSC) can continue to use `TSCLibc.mkdtemp` as its underlying implementation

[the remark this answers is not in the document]
Overcot: I found out that we do have an API inside `FileSystem` protocol which can be used for creating directories. So i changed it to use it and also saved previous behavior with generating random suffix with length of 6 numbers.
Now it is using `FileSystem` API which can be changed if it is needed for testing and mocking

Sources/Basics/TemporaryFile.swift
Overcot: Done it!

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,122 @@
...
+
+private func createTemporaryDirectory(dir: AbsolutePath?, prefix: String, fileSystem: FileSystem) throws -> AbsolutePath {
+    // This random generation is needed so that
+    // it is more or less equal to generation using `mkdtemp` function
+    let letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+
+    let randomSuffix = String((0..<6).map { _ in letters.randomElement()! })
+    
+    // Construct path to the temporary directory.
+    let templatePath = try AbsolutePath(prefix + ".\(randomSuffix)", relativeTo: determineTempDirectory(dir))
```
tomerd, line 95: since we have the `FileSystem` available now, we can probably also get rid of `determineTempDirectory` at this point... here is what it does:

```
public func determineTempDirectory(_ dir: AbsolutePath? = nil) throws -> AbsolutePath {
    let tmpDir = dir ?? localFileSystem.tempDirectory
    guard localFileSystem.isDirectory(tmpDir) else {
        throw TempFileError.couldNotFindTmpDir(tmpDir.pathString)
    }
    return tmpDir
}
```

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,122 @@
...
+    removeTreeOnDeinit: Bool = false,
+    _ body: @escaping (AbsolutePath) async throws -> Result
+) throws -> Task<Result, Error> {
+    try withTemporaryDirectory(dir: dir, prefix: prefix, fileSystem: fileSystem) { path, cleanup in
+        defer { if removeTreeOnDeinit { cleanup(path) } }
+        return try await body(path)
+    }
+}
+
+private func createTemporaryDirectory(dir: AbsolutePath?, prefix: String, fileSystem: FileSystem) throws -> AbsolutePath {
```
tomerd, line 87: lets make `fileSystem` first argument

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,122 @@
...
+///     - body: A closure to execute that receives the absolute path of the directory as an argument.
+///             If `body` has a return value, that value is also used as the
+///             return value for the `withTemporaryDirectory` function.
+///
+/// - Throws: `MakeDirectoryError` and rethrows all errors from `body`.
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public func withTemporaryDirectory<Result>(
+    dir: AbsolutePath? = nil,
+    prefix: String = "TemporaryDirectory",
+    fileSystem: FileSystem = localFileSystem,
```
tomerd, line 77: lets make `fileSystem` first argument

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,122 @@
...
+///           If `body` has a return value, that value is also used as the
+///           return value for the `withTemporaryDirectory` function.
+///           The cleanup block should be called when the temporary directory is no longer needed.
+/// - Returns: `Task<Result, Error>` which can be used
+/// - Throws: `MakeDirectoryError` and rethrows all errors from `body`.
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public func withTemporaryDirectory<Result>(
+    dir: AbsolutePath? = nil,
+    prefix: String = "TemporaryDirectory",
+    fileSystem: FileSystem = localFileSystem,
```
tomerd, line 38: lets make `fileSystem` first argument

Sources/Basics/TemporaryFile.swift
```diff
@@ -0,0 +1,122 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import TSCBasic
+import TSCLibc
```
tomerd, line 15: after all these changes we may be able to get rid of the `TSCLibc` import which would be awesome!

tomerd:
@Overcot this is getting super close. thanks for your effort and patience!

Sources/Basics/TemporaryFile.swift
Overcot: Done it!

Sources/Basics/TemporaryFile.swift
Overcot: Done it!

Overcot:
> @Overcot this is getting super close. thanks for your effort and patience!
Resolved all your comments. Please take a look again :)

Thanks for your support as well!

tomerd:
@swift-ci smoke test

tomerd:
lgtm. @abertelrud @neonichu ?

tomerd (approved):

compnerd:

tomerd:
thanks @Overcot

PR #5534, opened 2022-05-20
PackageRegistry: use `parentDirectory` to get the parent directory

compnerd:
This uses `parentDirectory` to get the dirname of the path which will
also fix the path construction for Windows.  The construction of the
relative path is normalized and `..` cannot be normalized without an
anchor.


compnerd:
@swift-ci please smoke test

tomerd (approved):

PR #5536, opened 2022-05-20
Tests: adjust path regex to account for alternate path separator

compnerd:
The Windows platform uses `\` as a path separator, which must be quoted
by Swift strings, and the regex.  Thus the pattern suffers from
toothpickitis, but allows the test to pass on Windows.


compnerd:
@swift-ci please smoke test

tomerd (approved):
😱

PR #5538, opened 2022-05-21
Tests: normalize the repository path

compnerd:
Use `url` rather than converting the path description to a string and
processing it on a specified path separator.  This will ensure that we
properly handle the path on Windows.


compnerd:
@swift-ci please smoke test

tomerd (approved):

PR #5543, opened 2022-05-22
Tests: adjust repository test and disable on Windows

compnerd:
This adds support for Windows to the compressed repository download
test.  Unfortunately, the repository contained within the tarball
contains files which contain invalid characters for some file systems.

For the time being, disable the test, though it would be preferable to
limit the repository contents to that which can be represented on all
file systems (perhaps consider the FAT32 restrictions which likely can
cover the file systems that we are likely to run on).

### DIFF
--- a/Tests/SourceControlTests/GitRepositoryTests.swift
+++ b/Tests/SourceControlTests/GitRepositoryTests.swift
@@ -113,10 +113,17 @@ class GitRepositoryTests: XCTestCase {
     /// `Inputs`, which has known commit hashes. See the `construct.sh` script
     /// contained within it for more information.
     func testRawRepository() throws {
+#if os(Windows)
+        try XCTSkipIf(true, "test repository has non-portable file names")
+#endif
         try testWithTemporaryDirectory { path in
             // Unarchive the static test repository.
             let inputArchivePath = AbsolutePath(#file).parentDirectory.appending(components: "Inputs", "TestRepo.tgz")
+#if os(Windows)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

neonichu (approved):

Tests/SourceControlTests/GitRepositoryTests.swift
```diff
@@ -113,10 +113,21 @@ class GitRepositoryTests: XCTestCase {
...
-            try systemQuietly(["tar", "-x", "-v", "-C", path.pathString, "-f", inputArchivePath.pathString])
+            func executable(_ name: String) -> String {
+#if os(Windows)
+                if !name.hasSuffix(".exe") {
+                    return "\(name).exe"
+                }
+#endif
+                return name
+            }
+            try systemQuietly([executable("tar"), "-x", "-v", "-C", path.pathString, "-f", inputArchivePath.pathString])
```
tomerd, line 130: I would prefer we remove the executable func and instead do

```
#if os(Windows)
try systemQuietly(["tar.exe", "-x", "-v", "-C", path.pathString, "-f", inputArchivePath.pathString])
#else
try systemQuietly(["tar", "-x", "-v", "-C", path.pathString, "-f", inputArchivePath.pathString])
#endif
```

compnerd: 🤷‍♂️

compnerd:
@swift-ci please smoke test

PR #5546, opened 2022-05-23
Tests: adjust slack for concurrency tests

compnerd:
The Windows time-slice is in units of 10ms.  This along with the high
contention can cause the concurrency tests to be flakey.  Increase the
timeout some to enable more machines to pass.


compnerd:
@swift-ci please smoke test

tomerd (approved):

PR #5547, opened 2022-05-23
Build: mark Windows for supporting link against executable

compnerd:
The `-rdynamic` link path works on Windows though requires that there
are exposed symbols.  In such a case, an import library will be emitted
along with the executable to permit the linking.  This marker changes
the location of the swiftmodule emission which enables the last test to
now pass.


compnerd:
@swift-ci please smoke test

neonichu (approved):

tomerd (approved):

PR #5559, opened 2022-06-02
fix mirrors reverse mapping to be more deterministic

tomerd:
motivation: stable resolved files in complex mirroring configurations where multiple mapping exist for a singl repository, leading to non-deterministic reverse mapping

chnages:
* track all reverse index mapping instead of just the latest
* when performing reverse mapping, sort based on which mapping was "visited", then alphabetically
* add tests

rdar://93579273
rdar://94029942


tomerd:
@swift-ci smoke test

neonichu (approved):

PR #5561, opened 2022-06-02
[5.7] fix an edge case where dependency resolution can go into endless loop 

tomerd:
5.7 cherry pick of #5535


tomerd:
@shahmishal @compnerd is windows failure expected on 5.7 branch?

neonichu (approved):

PR #5566, opened 2022-06-06
Avoid emitting a `.info` diagnostic with an empty message when there is no compiler output at all

abertelrud:
### Motivation

This causes clients to have to filter diagnostics in order to avoid emitting empty lines.

### Changes

- add a conditional before emitting a diagnostic
- update a unit test to check for it

rdar://94414151


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #5572, opened 2022-06-08
Create SECURITY.md

BosskurrMH:

### Motivation:

### Modifications:

### Result:


tomerd (changes requested):
thanks @BosskurrMH the content of this file should be link to https://www.swift.org/support/security.html

PR #5575, opened 2022-06-08
Change from Lock wrapper inside TSCBasic to NSLock

Overcot:
### Motivation:

As mentioned by @tomerd in https://github.com/apple/swift-tools-support-core/pull/327#issuecomment-1150101705 in order to move Process handling inside SPM there is a need to move Lock wrapper inside SPM Basics as well
### Modifications:

Basically copy pasted Lock implementation in TSCBasic to SPM Basic. Also needed to explicitly set module in order to solve conflicts when using Lock

### Result:

After the change i will continue to migrate bits that are required by process migration


tomerd:
@swift-ci smoke test

tomerd:
@abertelrud @neonichu I tend to think maybe we dont need the wrapper at all and just make public extensions on NSLock. wdyt?

Sources/Basics/Lock.swift
```diff
@@ -0,0 +1,35 @@
...
+ Licensed under Apache License v2.0 with Runtime Library Exception
+ 
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+ */
+
+import Foundation
+
+/// A simple lock wrapper.
+public struct Lock {
```
tomerd, line 14: @Overcot wdyt about ditching the wrapper approach and instead just declare an extension on `NSLock` with the `withLock` function? then we can update the callosities to use `NSLock` instead of `LocK` from TSC

Overcot: Yeah, sure. If everyone aggrees, i would it. This seems like a better approach. Btw - is the NSLock available on Windows?

tomerd (approved):
lgtm, thank you

Sources/Basics/Lock.swift
```diff
@@ -0,0 +1,20 @@
+/*
```
tomerd, line 1: one small change: rename the file to `NSLock+Extensions.swift` and add it to the CMakeLists.txt

tomerd:
@swift-ci smoke test

Sources/Basics/Lock.swift
tomerd: > Btw - is the NSLock available on Windows?

I believe so

tomerd:
thank you @Overcot

PR #5576, opened 2022-06-09
refactor LLBuildManifest graphviz file visualization flow

tomerd:
motivation: code cleanup

changes:
* move the code that prints the LLBuildManifest graphviz visualization to stdout from the build system to the build tool, which is where printg should be done
* remove printManifestGraphviz from BuildParameters as it was only necessary to drive outputing the visulaization from the build system
* move the LLBuild manifest dserializer to the Commands  module, with the rest of the user facing output serilizers
* create Utilities directory in the Commands module and move the various utilities there to better organize the code
* rename dependencies_serializer to DependenciesSerializer


tomerd:
@swift-ci smoke test

abertelrud (approved):
LGTM

PR #5577, opened 2022-06-09
refactor stream output and progress bars (part 1)

tomerd:
motivation: continue cleanup work, fixing issues when diagnositics output ay interleave with delegates writing directly to the output or reportig progress

changes:
* change ToolWorkspaceDelegate to use provided  output and progress handlers instead of wirting to OutputStream directly
* add print and progress handlers to SwiftToolObservabilityHandler, including a serial queue to maange the sequence of outputing to the output stream
* adjust tests to wait on diagnostics when necessary given the queue setup

### DIFF
--- a/Tests/CommandsTests/SwiftToolTests.swift
+++ b/Tests/CommandsTests/SwiftToolTests.swift
@@ -31,6 +31,8 @@ final class SwiftToolTests: CommandsTestCase {
                 tool.observabilityScope.emit(info: "info")
                 tool.observabilityScope.emit(debug: "debug")
 
+                tool.waitForObservabilityEvents(timeout: .now() + .seconds(1))
+
                 XCTAssertMatch(outputStream.bytes.validDescription, .contains("error: error"))
                 XCTAssertMatch(outputStream.bytes.validDescription, .contains("warning: warning"))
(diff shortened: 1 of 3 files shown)


tomerd:
@swift-ci smoke test

tomerd:
@swift-ci smoke test

neonichu (approved):

Tests/CommandsTests/SwiftToolTests.swift
```diff
@@ -31,6 +31,8 @@ final class SwiftToolTests: CommandsTestCase {
                 tool.observabilityScope.emit(info: "info")
                 tool.observabilityScope.emit(debug: "debug")
 
+                tool.waitForObservabilityEvents(timeout: .now() + .seconds(1))
```
abertelrud, line 34: Is this requirement new or was this always a flakiness in the tests?  It makes sense conceptually to wait for the output, just curious.

abertelrud (approved):

Tests/CommandsTests/SwiftToolTests.swift
tomerd: in this specific case, there is a new queue we are introducing so the tests would be flaky without a wait.

more broadly, I believe we were always subject to loosing the last messages, and now that we have queue sync we can wait on we have the opportunity to wait on it

PR #5587, opened 2022-06-13
Doc updates to module alias overriding

elsh:
Updates to overriding method
More detailed example

### DIFF
--- a/Documentation/ModuleAliasing.md
+++ b/Documentation/ModuleAliasing.md
@@ -31,8 +31,22 @@ Package manifest `swift-game`
 }
 ```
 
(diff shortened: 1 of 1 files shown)


elsh:
@swift-ci smoke test

neonichu (approved):
LGTM, but should we mention that we are allowing duplicated product names in certain scenarios, as well?

elsh:
> LGTM, but should we mention that we are allowing duplicated product names in certain scenarios, as well?

Sg, will add

elsh:
@swift-ci smoke test

Issue #5505, opened 2022-05-13
Massive JSON blob in Stdout


philipturner:
This was a bug affecting Swift-Colab 2.0 for quite a while, and I just recently found a workaround for it. Make a two-file Swift package with the following directory structure:
```
installed1
|-- Package.swift
|-- installed1.swift
```

In `Package.swift`, put the following:
```swift
// swift-tools-version:5.6
import PackageDescription
let package = Package(
  name: "installed1",
  products: [
    .library(
      name: "installed1",
      targets: ["installed1"]
    )
  ],
  dependencies: [],
  targets: [
    .target(
      name: "installed1",
      dependencies: [],
      path: ".",
      sources: ["installed1.swift"]
    )
  ]
)
```

In `installed1.swift`, put the following:
```swift
// intentionally blank

```

Build it once by entering the `installed1` directory and running `swift build`. Then, overwrite `installed1.swift` with its exact same contents using the command shown below. Run `swift build` once more and a massive JSON blob appears in stdout. This happens on both Linux (Google Colab, x86) and macOS (ARM64).
```
echo "// intentionally blank\n" > installed1.swift
```
<details><summary>Blob:</summary>

```
3291
{
  "command_arguments" : [
    "-frontend",
    "-emit-module",
    "-experimental-skip-non-inlinable-function-bodies-without-types",
    "\/Users\/philipturner\/Desktop\/installed1\/installed1.swift",
    "-target",
    "arm64-apple-macosx10.10",
    "-Xllvm",
    "-aarch64-use-tbi",
    "-enable-objc-interop",
    "-sdk",
    "\/Applications\/Xcode.app\/Contents\/Developer\/Platforms\/MacOSX.platform\/Developer\/SDKs\/MacOSX12.3.sdk",
    "-I",
    "\/Users\/philipturner\/Desktop\/installed1\/.build\/arm64-apple-macosx\/debug",
    "-I",
    "\/Applications\/Xcode.app\/Contents\/Developer\/Platforms\/MacOSX.platform\/Developer\/usr\/lib",
    "-F",
    "\/Applications\/Xcode.app\/Contents\/Developer\/Platforms\/MacOSX.platform\/Developer\/Library\/Frameworks",
    "-color-diagnostics",
    "-enable-testing",
    "-g",
    "-module-cache-path",
    "\/Users\/philipturner\/Desktop\/installed1\/.build\/arm64-apple-macosx\/debug\/ModuleCache",
    "-swift-version",
[...]

PR #5594, opened 2022-06-15
Use llbuild's new `content-exclusion-patterns` node attribute to prevent it from descending into `.build` and `.git` directories

abertelrud:
This requires the llbuild commit a1adaff2a405c3a74925e60356431050cb9a3a9d, but it does not cause problems if the llbuild being used doesn't yet support this property.  Instead it will be ignored.

Also added a dedicated test with a flat package.

We should consider whether any customized scratch directory should also be included in the list.  This is a little tricky since the exclusion patterns are applied to all subdirectory name, not just those at the top level.  Oddly, in testing I did not repro the problem if I was using a custom scratch directory, so this may not be a issue in practice.

Note that this fixes the problem with doing spurious rebuilds (of which the spurious JSON output is just a symptom) for the case of a flat package structure, where the root directory of the package is considered to be a source directory.  We have also seen this output when switching toolchains, which seems to be a completely different problem with similar symptoms.

rdar://93982172

### DIFF
--- a/Fixtures/Miscellaneous/FlatPackage/MyExec.swift
+++ b/Fixtures/Miscellaneous/FlatPackage/MyExec.swift
@@ -1,0 +1,8 @@
+@main
+public struct MyExec {
+    public private(set) var text = "Hello, World!"
+
+    public static func main() {
+        print(MyExec().text)
+    }
+}
--- a/Fixtures/Miscellaneous/FlatPackage/MyTest.swift
+++ b/Fixtures/Miscellaneous/FlatPackage/MyTest.swift
@@ -1,0 +1,8 @@
+import XCTest
+@testable import MyExec
+
+final class MyTest: XCTestCase {
+    func testExample() throws {
+        XCTAssertEqual(MyExec().text, "Hello, World!")
+    }
+}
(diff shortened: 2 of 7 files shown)


abertelrud:
I'm actually not sure how a customized directory could be supported here — the exclusion is for directory entries and contains just names (last path components), so if the scratch directory is `package-dir/foo` then we'd need to exclude `foo`, which would also eliminate `package-dir/bar/foo` (for example).  In practice this might not actually be a big problem, but there's some more exploration to be done here.

abertelrud:
@swift-ci please smoke test

tomerd (approved):

abertelrud:
Confirmed that this fixes the problem, when the latest llbuild is used, and with the test case from https://github.com/apple/swift-package-manager/issues/5482.  Surprisingly using `--scratch-path` with something like `<proj-dir>/.foo` does not seem to cause the problem even though it in theory should still do so.  I haven't yet looked into that but I think we should take this fix as it is safe and effective.

abertelrud:
Found a way to reproduce the problem using a test fixture, so I'm adding an end-to-end test as well.

abertelrud:
@swift-ci please smoke test

neonichu (approved):

philipturner:
This seems to have fixed #5482 (https://github.com/apple/swift-package-manager/issues/5482#issuecomment-1210984152). Can we add https://github.com/apple/swift-package-manager/issues/5505 as a regression test, then close both #5482 and #5505?

PR #5595, opened 2022-06-15
flush observability on error

tomerd:
motivation: follow up on previsou changes to queue observability messages

changes: catch error an flush on those cases as well


tomerd:
@swift-ci smoke test

abertelrud (approved):

neonichu (approved):

PR #5585, opened 2022-06-13
Module aliases override updates

elsh:
- Use alias value defined upstream as a key downstream for overriding and chain them
- Add pre-chained aliases to targets when needed and show diags
Resolves rdar://93445417, rdar://93218209

PR #5600, opened 2022-06-17
[5.7] Module aliases override updates

elsh:
* Use alias value defined upstream as a key downstream for overriding and chain them
* Add pre-chained aliases to targets when needed and show diags
Resolves rdar://93445417, rdar://93218209


elsh:
Cherrypicked from https://github.com/apple/swift-package-manager/pull/5585

elsh:
@swift-ci  smoke test

tomerd (approved):

elsh:
@swift-ci smoke test

PR #5605, opened 2022-06-18
suppress warnings from remote dependencies

tomerd:
motivation: warnings from remote dependencies are not actionable and make the use of -warnings-as-errors difficult

changes:
* wire up the ResolvedPackage to SwiftTargetBuildDescription
* pass -suppress-warnings for remote dependencies
* add unit test

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -530,6 +530,8 @@ public final class ClangTargetBuildDescription {
 
 /// Target description for a Swift target.
 public final class SwiftTargetBuildDescription {
+    /// The package this target belongs to.
+    public let package: ResolvedPackage
 
     /// The target described by this target.
     public let target: ResolvedTarget
@@ -661,6 +663,7 @@ public final class SwiftTargetBuildDescription {
 
     /// Create a new target description with target and build parameters.
     init(
+        package: ResolvedPackage,
         target: ResolvedTarget,
         toolsVersion: ToolsVersion,
         additionalFileRules: [FileRuleDescription] = [],
@@ -675,6 +678,7 @@ public final class SwiftTargetBuildDescription {
         guard target.underlyingTarget is SwiftTarget else {
             throw InternalError("underlying target type mismatch \(target)")
         }
+        self.package = package
         self.target = target
         self.toolsVersion = toolsVersion
         self.buildParameters = buildParameters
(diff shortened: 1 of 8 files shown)


tomerd:
@swift-ci smoke test

Sources/Build/BuildPlan.swift
```diff
@@ -869,6 +873,11 @@ public final class SwiftTargetBuildDescription {
             args += ["-emit-module-interface-path", parseableModuleInterfaceOutputPath.pathString]
         }
 
+        // suppress warnings if the package is remote
+        if self.package.isRemote {
+            args += ["-suppress-warnings"]
+        }
+
```
ktoso, line 880: Awesome 👏

ktoso (approved):
Very nice! Thank you :-)

Sources/Build/BuildPlan.swift
```diff
@@ -2303,8 +2321,19 @@ private func generateResourceInfoPlist(
...
     }
 }
+
+extension ResolvedPackage {
+    fileprivate var isRemote: Bool {
+        switch self.underlyingPackage.manifest.packageKind {
+        case .registry, .remoteSourceControl, .localSourceControl:
+            return true
+        case .root, .fileSystem:
+            return false
```
abertelrud, line 2336: Just to confirm, this would make `path` dependencies be considered local and not remote, right?  If so, I completely agree, that's seems like the right semantics.

abertelrud (approved):

abertelrud:
This is great!

Sources/Build/BuildPlan.swift
tomerd: exactly

MPLew-is:
This is fantastic to have, but it looks like remarks still made it through:
```
{redacted}/.build/checkouts/swift-aws-lambda-runtime/Sources/AWSLambdaRuntimeCore/LambdaContext.swift:18:17: remark: '@preconcurrency' attribute on module 'NIOCore' is unused
@preconcurrency import NIOCore
~~~~~~~~~~~~~~~~^
```

This won't limit usage of `-warnings-as-errors`, but it definitely still leads to clutter in the build output that you really can't do anything about. I don't see any compiler flags to silence remarks though - do you have any thoughts as to how we might go about silencing these for remote dependencies as well?

tomerd:
@elsh @abertelrud @neonichu @artemcm thoughts?

MPLew-is:
@tomerd @elsh @abertelrud @neonichu @artemcm any other updates or thoughts on silencing remarks?

artemcm:
> @tomerd @elsh @abertelrud @neonichu @artemcm any other updates or thoughts on silencing remarks?

That sounds like a very reasonable request to me.

artemcm:
I'm a bit surprised we have remarks that are emitted without being explicitly requested with some `-R...` flag, but if we are emitting these unprompted, then being able to silence them similarly to warnings makes sense.

tomerd:
@artemcm what is the way to silence them?

artemcm:
> @artemcm what is the way to silence them?

We do not currently have a way to do that, we need to build one that's similar to `-suppress-warnings`, perhaps something like this:
https://github.com/apple/swift/pull/61762

PR #5609, opened 2022-06-20
Prepare SPM Targets for Process migration by specifying that callee should use TSCBasic.Process

Overcot:
Basically it marks all usage of Process as TSCBasic.Process

### Motivation:

It is to address a problem of disambiguation that occurs when creating `Process` inside SwiftBasics library. The motivation overall is driven by comment https://github.com/apple/swift-package-manager/pull/5532#issuecomment-1133212593

### Modifications:

Specify explicitly that current `Process` is from `TSCBasic` library. It is done in source code and tests

### Result:

The result is to make sure that adoption by targets of future `Process` inside SwiftBasics to be incremental

### DIFF
--- a/Sources/Basics/Archiver+Zip.swift
+++ b/Sources/Basics/Archiver+Zip.swift
@@ -47,9 +47,9 @@ public struct ZipArchiver: Archiver, Cancellable {
             }
 
 #if os(Windows)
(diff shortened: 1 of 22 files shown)


tomerd (approved):

tomerd:
@swift-ci smoke test

PR #5612, opened 2022-06-21
allow use of warnings-as-errors while dependencies have supressed warnings

tomerd:
motivation: warnings-as-errors and suppress-warnings are mutually exclusive

changes:
* do not pass warnings-as-errors for remote dependencies which use suppress-warnings
* add test


tomerd:
@swift-ci smoke test

neonichu (approved):

PR #5621, opened 2022-06-23
Modify driver invocation on `swift run --repl` to match new REPL behavior.

artemcm:
Behaviour modified in rdar://78280429 - by making `swift` just print a help menu, and REPL requiring `swift repl`.

Resolves rdar://95793872


artemcm:
@swift-ci smoke test

tomerd (approved):

artemcm:
> do we also need to pull into 5.7?

Yes, we really do.

Issue #5580, opened 2022-06-10
Transitive dependencies in Package.swift slowing builds down


kadanes:
### Description

I work on the Amazon Flex delivery app. We use SPM with over 100 different packages, made by different teams. We have a couple of issues with SPM related to package resolution and build times. We had a discussion with Apple engineers David and Artem during WWDC22 open hour lab. They asked us to open issues with SPM for the issues we are seeing.

### Expected behavior

Build times should remain unaffected by transitive dependencies listed in Package.swift file.

### Actual behavior

Build times increase by 40-50 mins when transitive dependencies are listed in Package.swift file. After commenting them out, the build times improve significantly. But this is a bad development practice. 

### Steps to reproduce

You will need an app with complex dependency graph to repro this issue. With such app, commenting out transitive dependencies speeds up build times significantly. 

### Swift Package Manager version/commit hash

SPM 5.6

### Swift & OS version (output of `swift --version && uname -a`)

```
swift-driver version: 1.45.2 Apple Swift version 5.6.1 (swiftlang-5.6.0.323.66 clang-1316.0.20.12)
Target: arm64-apple-macosx12.0
Darwin f4d488949cc9.ant.amazon.com 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:46:32 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T6000 arm64
```

tomerd:
@parthv21 is the slowness in the build itself, or package resolution?

parthv21:
Unfortunately the slowness is during building. A more complex graph leads to the  `Pre-Planning` stage taking longer time. So we have had to do this optimization to improve performance: 

If A depends on B and C, and B depends on C, the Package.swift files would look like this

- Package.swift (B)
```
target(
    name: "B",
    dependencies: [
        .product(name: "C", package: "C"),
]),
```
And then A would comment out C
- Package.swift (A) 

```
target(
    name: "A",
    dependencies: [
//        .product(name: "C", package: "C"), // Optimization - already included transitively
        .product(name: "B", package: "B"),
]),
```

tomerd:
thanks for the additional details @parthv21. could you clarify further if the slowness is in the build, or in the "build planning" phase? one thing that could help allot here is a sysdiganose while its "spinning". if you prefer to share that privately you can file through https://feedbackassistant.apple.com/

vadimkrutovlv:
[...]

PR #5628, opened 2022-06-25
Improve speed of finding dependency cycles

mpsnp:
Significantly improve speed of finding cycles in dependency tree

### Motivation:

At the moment, if dependencies are composed to layers, depending on graph size, it can take minutes and even hours to validate that there are no cycles. Take a look at [this issue](https://github.com/apple/swift-package-manager/issues/5580) and a [PR in swift-tools-support-core](https://github.com/apple/swift-tools-support-core/pull/332).

### Modifications:

Cache already checked graph nodes and reuse results in subsequent visits.

### Result:

This modification improved complexity to O(n) instead of O(n!). So package graph of 100 layers with 100 nodes in each layer is checked in 0.2 seconds instead of hours.

### DIFF
--- a/Sources/PackageGraph/PackageGraph+Loading.swift
+++ b/Sources/PackageGraph/PackageGraph+Loading.swift
@@ -872,6 +872,7 @@ fileprivate func findCycle(
 ) rethrows -> (path: [Manifest], cycle: [Manifest])? {
     // Ordered set to hold the current traversed path.
     var path = OrderedCollections.OrderedSet<Manifest>()
+    var validPackages = Set<Manifest>()
 
     // Function to visit nodes recursively.
     // FIXME: Convert to stack.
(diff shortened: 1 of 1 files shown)


tomerd (changes requested):
is the verbosity change intentional?

mpsnp:
@tomerd excuse me, don’t get what you mean. Which verbosity?

mpsnp:
Okay, you mean to use `validManifests` instead of `validPackages`?

mpsnp:
Actually I think I’ll replace this func by the one from swift-tools-support-core.

tomerd:
@mpsnp i mean the changes to the plugin runner. they seem accidental?

tomerd:
wrt to using tools-support-core, our goal is to deprecate that library, so leaning on the SwiftPM version is preferred

mpsnp:
Yeah, that's why i was confused :) 

Will update PR today, with unit tests, small naming fixes and checks that nothing else uses function from tools-support-core

tomerd:
looks like this was done via https://github.com/apple/swift-package-manager/pull/5839/

tomerd:
thanks @mpsnp

PR #5636, opened 2022-07-01
SwiftPM does not build all dependency executables before invoking plugin build command

abertelrud:
### Background & Motivation

Plugins can specify dependencies on executables.  Currently, those executables get built before running build tool commands returned by the plugin if those dependencies were either:
- the path of the executable of a build tool command returned by the plugin, or
- explicitly included by the plugin in the list of input dependencies to a build command returned by the plugin

This could potentially be more efficient than building all dependencies, since a plugin could depend on two separate executables and then return commands that only used one or the other of them.  This would result in SwiftPM only building the executables that were actually mentioned by build commands.

However:  This led to confusion and multiple bug reports of failures resulting from not building the executables before trying to use them.  One common case is with SwiftProtobuf, which uses `protoc` as the executable and then a SwiftPM-built executable as one of the parameters to `protoc`.  If the plugin didn't explicitly include the code generation executable in the list of input dependencies in the build tools command it returns, the build would fail and it wouldn't be clear why.

### Changes

Make the list of input dependencies of an llbuild command that runs a plugin-provided build tool include the paths of all the tools on which the plugin has declared a dependency, not just the one that appears as an executable in the build command.

This can cause more command line tools to be built but is less confusing overall since a dependency on an executable by a plugin understandably implies that the executable should be built no matter what.

Note that if the plugin doesn't return _any_ build commands, none of the dependencies need to be built.  That behavior is preserved.

### Specific Modifications:

- include the paths of the depended-upon tools in the input dependencies to the plugin-provided build tool commands
- stop special-casing the executable path of build tool commands (any tool there is now included in the list of input dependencies from tools)
- add a unit test

### Results

The executables declared by a plugin are built before _any_ build command returned by the plugin (though not if the plugin doesn't return any commands, which is a no-op).

Note also that this new behavior of building the executables up-front more closely matches the behavior in Xcode 14.

rdar://93679015

### DIFF
--- a/Sources/SPMBuildCore/PluginInvocation.swift
+++ b/Sources/SPMBuildCore/PluginInvocation.swift
@@ -353,6 +353,9 @@ extension PackageGraph {
                         dict[name] = path
                     }
                 })
+                
+                // Determine additional input dependencies for any plugin commands, based on any executables the plugin target depends on.
+                let toolPaths = toolNamesToPaths.values.sorted()
 
                 // Assign a plugin working directory based on the package, target, and plugin.
                 let pluginOutputDir = outputDir.appending(components: package.identity.description, target.name, pluginTarget.name)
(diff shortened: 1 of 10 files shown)


abertelrud:
@swift-ci please smoke test

neonichu (approved):

abertelrud:
Looks like there is at least one test to update (because it codified the previous dependency) in addition to adding the new one.

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

Sources/SPMBuildCore/PluginInvocation.swift
```diff
@@ -395,7 +400,7 @@ extension PackageGraph {
                                 arguments: arguments,
                                 environment: environment,
                                 workingDirectory: workingDirectory),
-                            inputFiles: inputFiles,
+                            inputFiles: toolPaths + inputFiles,
```
abertelrud, line 403: This is the key change.  Nothing particularly risky or anything that requires extra scrutiny, I think, but it seemed worth calling out amid all the more mundane changes.

tomerd (approved):

PR #5654, opened 2022-07-12
Mark bootstrap command as required

keith:
Otherwise this crashes if you don't pass it any args


tomerd:
@swift-ci smoke test

bnbarham:
This fails in Python 3.6 - I've opened https://github.com/apple/swift-package-manager/pull/5657 to use the more compatible API (just setting it after).

Failure in https://ci.swift.org/job/oss-swift-RA-linux-ubuntu-18_04/1483

PR #5657, opened 2022-07-13
Use < 3.7 compatible API for marking subparsers as required

bnbarham:
`required` was only added as a parameter to `add_subparsers` in Python
3.7. Explicitly set `required` on `subparsers` after instead, which is
supported on all versions.


bnbarham:
@swift-ci please smoke test

tomerd (approved):

neonichu (approved):

keith:
sorry, thanks for the fix!

PR #5660, opened 2022-07-13
address warnings from using deprecated TSCBasic::Lock

tomerd:
motivation: no warnings

changes: use NSLock instead of deprecated TSCBasic::Lock


tomerd:
@swift-ci smoke test

neonichu (approved):

PR #5666, opened 2022-07-15
Improve CompositeAuthorizationProvider

mauriciomaniglia:
Decouple CompositeAuthorizationProvider from ObservabilityScope and add unit test. 

### Motivation:

CompositeAuthorizationProvider currently uses the concrete componente ObservabilityScope to only send the message `emit(info: String, metadata: ObservabilityMetadata)` which is define in `DiagnosticsEmitterProtocol`. 

### Modifications:

Instead of using the concrete componente ObservabilityScope, I've change to use only the protocol `DiagnosticsEmitterProtocol`. 

Also added unit test for the communication between CompositeAuthorizationProvider and DiagnosticsEmitterProtocol. 

### Result:

Nothing changed

### DIFF
--- a/Sources/Basics/AuthorizationProvider.swift
+++ b/Sources/Basics/AuthorizationProvider.swift
@@ -255,15 +255,15 @@ public struct KeychainAuthorizationProvider: AuthorizationProvider {
 public struct CompositeAuthorizationProvider: AuthorizationProvider {
     // marked internal for testing
     internal let providers: [AuthorizationProvider]
-    private let observabilityScope: ObservabilityScope
+    private let diagnosticsEmitter: DiagnosticsEmitterProtocol
 
-    public init(_ providers: AuthorizationProvider..., observabilityScope: ObservabilityScope) {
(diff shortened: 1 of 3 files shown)


tomerd:
thanks @mauriciomaniglia, could you elaborate on the background for the proposed change. is there some use case where you use CompositeAuthorizationProvider with a different DiagnosticsEmitter?

mauriciomaniglia:
Hi @tomerd, I can't think another use case right now because I starting to learn the system and still don't have a broad view of how everything is put together. I've raise this proposal with the intention to decouple a dependency from its concrete implementation to facilitate testing and make CompositeAuthorizationProvider only know what it needs to know.

tomerd:
thank you @mauriciomaniglia, `ObservabilityScope` is expected to be passed around, so unless there is a strong reason to "cast it down"  to `DiagnosticsEmitterProtocol` we should keep things consistent with the rest of the code base, which is to pass `ObservabilityScope`

PR #5669, opened 2022-07-15
PackagePlugin: change the `setvbuf` usage

compnerd:
Windows is much less lenient about handling and requires that a buffer
is provided if performing line buffering.  On Windows completely
disable all buffering instead which leaves us vulnerable to partial line
emissions if more than one thread is active.

Fixes: 5668


compnerd:
@swift-ci please test

compnerd:
@swift-ci please smoke test

tomerd (approved):

PR #5670, opened 2022-07-17
improve "swift test --show-codecov-path"

tomerd:
motivation: --show-codecov-path is only relevant after running test with test coverage

changes:
* only allow the use of --show-codecov-path when running swift test with --enable-code-coverage
* remove the dedicate "swift test show-codecov-path" subcommand
* print code coverage path at the end of test coverage invocation, when flag is set

rdar://88464565


tomerd:
@swift-ci smoke test

abertelrud (approved):

PR #5678, opened 2022-07-21
improve heuristics for @main support

tomerd:
motivation: heuristics was based on file name, which is not accurate enough and did not include files like LinuxMain.swift

changes:
* change heuristics to look into the ocntent of the source file and check for the @main annotation and decide on that
* update tests


tomerd:
@swift-ci smoke test

tomerd:
cc @ktoso

neonichu (approved):

PR #5687, opened 2022-07-25
improve heuristics for @main support 

tomerd:
motivation: heuristics was based on file name, which is not accurate enough and did not include files like LinuxMain.swift

changes:
* change heuristics to look into the ocntent of the source file and check for the @main annotation and decide on that
* update tests

radar/9729773


tomerd:
@swift-ci smoke test

neonichu (approved):

ktoso:
Bad radar link; I think this was: rdar://97297732

Issue #4705, opened 2019-05-30
[SR-10799] SwiftPM should have a --no-network flag that ensures no network calls are made


neonichu:

mustiikhalil:
Hey, I'm thinking of picking this up, is there anything I need to know before starting to implement it?

ishaanbedi:
Hey @Overcot !
Are you still working on this issue or I can pick it up?

Overcot:
> Hey @Overcot !
> Are you still working on this issue or I can pick it up?

Hi @ishaanbedi!
No, unfortunately i’ve stopped working on it

ishaanbedi:
Hi everyone!
Just a quick note here, I am working on this issue, and will soon open a draft PR.
Thanks & Regards!

PR #5691, opened 2022-07-26
SR-10799: Introduce flag that can disable network requests

Overcot:
Introduce a flag that can when enabled - disables all network requests

### Motivation:

as stated https://github.com/apple/swift-package-manager/issues/4705 there is a need for a flag that can disable all network requests
### Modifications:

This is still a draft but i outlined the idea that we need to change workspace initialisation of HTTPClient based on a flag

### DIFF
--- a/Sources/Basics/HTTPClient.swift
+++ b/Sources/Basics/HTTPClient.swift
@@ -87,6 +87,9 @@ public struct HTTPClient: Cancellable {
         progress: ProgressHandler? = nil,
         completion: @escaping CompletionHandler
     ) {
+        guard configuration.isEnabled else {
+            return completion(.failure(CancellationError()))
+        }
         // merge configuration
         var request = request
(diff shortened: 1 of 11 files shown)


Sources/Basics/HTTPClient.swift
```diff
@@ -87,6 +87,9 @@ public struct HTTPClient: Cancellable {
         progress: ProgressHandler? = nil,
         completion: @escaping CompletionHandler
     ) {
+        guard configuration.isEnabled else {
+            return completion(.failure(CancellationError()))
```
abertelrud, line 91: AFAICT this is the place where errors are thrown when the network is disabled — probably a more specific error that describes the reason for the cancellation would be useful.  I realize this PR is still in Draft so maybe that was already planned, but just a thought for when this becomes ready for review.

PR #5696, opened 2022-07-29
Conditionalize the keyword in the generated test main

FranzBusch:
# Motivation
Right now the code for the test main file contains this generated code

```swift
var tests = [XCTestCaseEntry]()
```

For the rare cases where a package has no tests the generated code produces a warning since the `var tests` was never mutated. If these users are also running `warnings-as-errors` their compile will fail.

# Modification
This PR conditionalizes the keyword of the variable depending on if there is a test or not.

# Result
The code is not generating any warnings any more.


FranzBusch:
@tomerd @neonichu I couldn't find any tests that are specifically testing the generation of the code in the test main. I am happy to add tests to this if you could point me into the right direction where to add them.

abertelrud:
> @tomerd @neonichu I couldn't find any tests that are specifically testing the generation of the code in the test main. I am happy to add tests to this if you could point me into the right direction where to add them.

Since the goal of this change is to avoid the warning when building, one approach would be to add a test fixture (consisting of a package that doesn't have tests, or whatever it is that triggers this warning) and then check that the warning is not present when building the package.

@neonichu or @tomerd might heave better ideas but it looks to me as if `TestDiscoveryTests` under the `FunctionalTests` directory might be a good place, since the generation of this `XCTMain` is AFAIK part of test discovery.

tomerd:
> @neonichu or @tomerd might heave better ideas but it looks to me as if TestDiscoveryTests under the FunctionalTests directory might be a good place, since the generation of this XCTMain is AFAIK part of test discovery.

 +1

tomerd:
@FranzBusch would you like to add a test as suggested above?

FranzBusch:
@tomerd I just added a test for this that checks that the build for such a package doesn't produce any warnings.

tomerd (approved):

tomerd:
@swift-ci smoke test

tomerd:
thanks @FranzBusch for this. could you also create a cherry-pick for 5.7 as well

FranzBusch:
@tomerd here you go: https://github.com/apple/swift-package-manager/pull/5721

PR #5704, opened 2022-08-01
Add .snippet target to @main heuristics for including -parse-as-library

bitjammer:
The heuristics for deciding whether to include the `-parse-as-library` flag
were recently improved in #5687 (rdar://97297732).

### Motivation

This change allows
`.snippet` target types to benefit from those improved heuristics.

### Modifications

Added `.snippet` to the cases allowed for `SwiftTargetBuildDescription.needsToBeParsedAsLibrary`.

Testing: Changed the walk when searching for snippets to explicitly use the
`PackageBuilder.fileSystem`; the default argument uses the local filesystem,
which won't work for in-memory filesystems used in tests.

Testing: Added additional snippet files, one with `@main` and one without, to
`BuildPlanTests.testParseAsLibraryFlagForExe`.

### Results

According to the same rules, snippets with `@main` should get
`-parse-as-library`, while those assumed to have top-level code should not.

rdar://97802934

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -640,7 +640,7 @@ public final class SwiftTargetBuildDescription {
         switch self.target.type {
         case .library, .test:
             return true
-        case .executable:
+        case .executable, .snippet:
             // This deactivates heuristics in the Swift compiler that treats single-file modules and source files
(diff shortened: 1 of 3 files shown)


bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please smoke test

neonichu (approved):

Issue #5698, opened 2022-07-30
Linking with lld stopped working on linux/Android since lld 13 and this repo enabled `--gc-sections`


buttaface:
### Description

After dead-stripping was enabled in #4135, [various people reported problems with using lld](https://github.com/apple/swift-package-manager/pull/4135#issuecomment-1050485754), though it wouldn't always show up. It appears to be some change with lld 13, as lld 12 works fine, perhaps the commit @drodriguez linked (I'll try reverting that commit and report back later).

Obviously, this is really a regression in how Swift interacts with lld 13 and 14, but I'm reporting it here first as we'll probably want to amend SPM 5.7 quickly before the underlying lld issue is figured out. @keith, maybe we should disable stripping on ELF if the linker is lld for now?

@3405691582, can you reproduce on OpenBSD?

### Expected behavior

Linking to work fine with lld, as it does with gold.

### Actual behavior

Several linker errors like this show up instead:
```
ld.lld: error: undefined hidden symbol: __start_swift5_protocols
>>> referenced by SwiftRT-ELF.cpp
>>>               /home/butta/swift-DEVELOPMENT-SNAPSHOT-2022-07-25-a-ubuntu20.04/usr/lib/swift/linux/x86_64/swiftrt.o (swift_image_constructor())
```

### Steps to reproduce

1. Clone swift-argument-parser
2. Remove a file because of [some unrelated bug](https://github.com/apple/swift-argument-parser/issues/463): `rm Sources/ArgumentParserTestHelpers/TestHelpers.swift`
3. Build the package in release mode with lld: `../swift-DEVELOPMENT-SNAPSHOT-2022-07-25-a-ubuntu20.04/usr/bin/swift build -v -c release -Xswiftc -use-ld=lld`

### Swift Package Manager version/commit hash

latest Jul. 25 trunk snapshot and 5.7 Jul. 23 snapshot

### Swift & OS version (output of `swift --version && uname -a`)

Swift 5.8 and 5.7 on Ubuntu 20.04 x86_64

buttaface:
I can confirm that reverting llvm/llvm-project@6d2d3bd fixes this issue. We're going to have to adapt that lld ELF commit for Swift.

keith:
Here's a fix for this, it's not fully complete yet because the problem is the `R` directive I'm adding here wasn't supported until clang 13+, and binutils 2.36, so I think I need to conditionalize this somehow based on those versions. https://github.com/apple/swift/pull/60357

[...]

PR #5707, opened 2022-08-02
Remove --gc-sections for all targets for now

keith:
Similar to the comment here about wasm-ld, lld recently switched a
default that broke this for all elf targets. This is a low risk fix
since we weren't passing this flag before either. People who really need
this behavior can still pass something like `-Xlinker --gc-sections
-Xlinker -z -Xlinker nostart-stop-gc`

Fixes https://github.com/apple/swift-package-manager/issues/5698

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -1334,17 +1334,12 @@ public final class ProductBuildDescription {
                 return ["-Xlinker", "-dead_strip"]
             } else if buildParameters.triple.isWindows() {
                 return ["-Xlinker", "/OPT:REF"]
-            } else if buildParameters.triple.arch == .wasm32 {
-                // FIXME: wasm-ld strips data segments referenced through __start/__stop symbols
+            } else {
+                // FIXME: wasm-ld / ld.lld strips data segments referenced through __start/__stop symbols
                 // during GC, and it removes Swift metadata sections like swift5_protocols
                 // We should add support of SHF_GNU_RETAIN-like flag for __attribute__((retain))
                 // to LLVM and wasm-ld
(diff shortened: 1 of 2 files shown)


keith:
Ideally we can fix this instead with https://github.com/apple/swift/pull/60357, but if that PR isn't enough, this might be necessary in the short term.

buttaface:
Can we just apply this as a short-term fix for the 5.7 branch, or do all 5.7 fixes have to go into trunk first? Your stdlib pull will fix trunk and there's no urgency with trunk, only 5.7.

keith:
here's a 5.7 PR https://github.com/apple/swift-package-manager/pull/5708

tomerd:
@Keith should we get this merged into main first, then 5.7?

tomerd:
@swift-ci smoke test

keith:
yea there was a question of if we could only do 5.7 and try to solve the core issue without reverting on main, but whatever the process is is fine with me. we can always revert this if we can solve the core issue too

tomerd:
I rather keep them symmetrical so we dont have a later regression, then we can revert / reintroduce when ready?

keith:
yep sounds fine!

keith:

tomerd (approved):

keith:
I doubt the windows failure here is related, it seems to be timing out

tomerd:
@compnerd okay to merge even with the windows failure?

compnerd:
I think that a retest should help - the offending change was reverted.

compnerd:

compnerd (approved):
This seems safe for Windows; if the Windows CI still flakes, lets go ahead and merge this anyways.

PR #5708, opened 2022-08-03
[5.7] Remove --gc-sections for all targets for now

keith:
Similar to the comment here about wasm-ld, lld recently switched a
default that broke this for all elf targets. This is a low risk fix
since we weren't passing this flag before either. People who really need
this behavior can still pass something like `-Xlinker --gc-sections
-Xlinker -z -Xlinker nostart-stop-gc`

Fixes https://github.com/apple/swift-package-manager/issues/5698

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -1290,17 +1290,12 @@ public final class ProductBuildDescription {
                 return ["-Xlinker", "-dead_strip"]
             } else if buildParameters.triple.isWindows() {
                 return ["-Xlinker", "/OPT:REF"]
-            } else if buildParameters.triple.arch == .wasm32 {
-                // FIXME: wasm-ld strips data segments referenced through __start/__stop symbols
+            } else {
+                // FIXME: wasm-ld / ld.lld strips data segments referenced through __start/__stop symbols
                 // during GC, and it removes Swift metadata sections like swift5_protocols
                 // We should add support of SHF_GNU_RETAIN-like flag for __attribute__((retain))
                 // to LLVM and wasm-ld
-                // This workaround is required for not only WASI but also all WebAssembly archs
-                // using wasm-ld (e.g. wasm32-unknown-unknown). So this branch is conditioned by
-                // arch == .wasm32
                 return []
-            } else {
-                return ["-Xlinker", "--gc-sections"]
             }
         }
     }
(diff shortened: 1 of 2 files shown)


tomerd:
cc @airspeedswift

buttaface (approved):
I think we should get this in so that 5.7 doesn't regress for lld builds, and it should be perfectly safe since we are merely going back to how 5.6 never added this flag for ELF platforms anyway.

Sources/Build/BuildPlan.swift
```diff
@@ -1290,17 +1290,12 @@ public final class ProductBuildDescription {
                 return ["-Xlinker", "-dead_strip"]
             } else if buildParameters.triple.isWindows() {
                 return ["-Xlinker", "/OPT:REF"]
-            } else if buildParameters.triple.arch == .wasm32 {
-                // FIXME: wasm-ld strips data segments referenced through __start/__stop symbols
+            } else {
```
airspeedswift, line 1293: Should we really be resolving this by making this the unconditional `else` branch? Is there no solution here where we check a second condition to get this behavior? (Sorry about my ignorance in this area, I’m mostly just coming to this conclusion in a this context-free way)

keith: I think we'd have to check that the linker was not lld, which might be fine to do just based on the `use-ld` arg value, but also that supports paths like `-use-ld=/path/to/lld`, which just made it feel potentially more error prone.

airspeedswift: Other than this concern, I’d be fine taking it onto the release branch.

tomerd: I think this is fine? its basically adding additional [dead stripping] linker arguments for certain platforms, and none in the default case

keith: It is too bad that folks not using lld can't get the dead stripping benefit, but theoretically at least they can opt in manually 😕

tomerd: @airspeedswift thoughts?

buttaface: I agree that checking for lld is too error-prone, as linker selection is not a direct SPM flag, so you're stuck parsing the various `-Xswiftc` and `-Xlinker` flags and still wouldn't catch situations where people don't pass in SPM flags but symbolically link their system linker to lld, because they use lld by default with all toolchains.

keith:
@swift-ci please test

buttaface:
Something wrong with the CI?

keith:
@swift-ci please smoke test

tomerd (approved):

PR #5721, opened 2022-08-12
[5.7] Conditionalize the keyword in the generated test main

FranzBusch:
# Motivation
Right now the code for the test main file contains this generated code

```swift
var tests = [XCTestCaseEntry]()
```

For the rare cases where a package has no tests the generated code produces a warning since the `var tests` was never mutated. If these users are also running `warnings-as-errors` their compile will fail.

# Modification
This PR conditionalizes the keyword of the variable depending on if there is a test or not.

# Result
The code is not generating any warnings any more.


tomerd:
@swift-ci smoke test

tomerd (approved):

PR #5751, opened 2022-08-29
Transition away from deprecated protocol requirement in llbuild's Tool

jakepetroules:
createCommand now returns an Optional, since command creation may fail.

rdar://99298420


jakepetroules:
@swift-ci please smoke test

tomerd (approved):

neonichu (approved):

PR #5765, opened 2022-09-10
[WIP] Dynamically generate flag expectations in tests

### LINKED ISSUE #5474: dynamically generate flag expectations in tests
Many of the tests are currently statically generating the flags for expectations in BuildPlanTests.swift based upon `#if os` which changes behaviour based on the OS rather than the target.  This can cause subtle bugs to arise.  We should dynamically compute the arguments.

azarovalex:
Fixes #5474.
Run build tests on all platforms regardless of the host platform.

### Motivation:

As @compnerd pointed out in the initial ticket, many of the tests in are currently statically generating the flags for expectations in `BuildPlanTests.swift` based upon `#if os` which changes behavior based on the OS rather than the target. This can cause subtle bugs to arise. We should dynamically compute the arguments.

### Modifications:

Removed compile time constants from `BuildPlanTests.swift` and added tests for all supported destinations.

### Result:

Most tests in `BuildPlanTests.swift` will test Windows, Linux, and Android compilator flags even if tests are running on macOS, and vice-versa.

### DIFF
--- a/Tests/BuildTests/BuildPlanTests.swift
+++ b/Tests/BuildTests/BuildPlanTests.swift
@@ -33,94 +33,109 @@ final class BuildPlanTests: XCTestCase {
         return "-j3"
     }
 
+    private let testTriples: [TSCUtility.Triple] = [
+        hostTriple,
+        .macOS,
+        .windows,
+        .arm64Linux,
+        .arm64Android
+    ]
+
     func testBasicSwiftPackage() throws {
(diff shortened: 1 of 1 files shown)


compnerd (commented):
This seems to be going in the right direction.  I wonder if we can now share some of the platform agnostic flags, extending that set for the comparison as necessary, assuming that the differences are all suffixed rather than infixed.  If not, that is okay, but the more that can be shared, the better we are.

PR #5781, opened 2022-09-25
Swift package manifest generation should not write out absolute paths for path-based package dependencies

abertelrud:
### Motivation
Path-based package dependencies were being written out as absolute during manifest source generation, even though relative is almost always (if not always) the right thing to write out in order to avoid creating packages that only work on the local machine.

### Modifications
- add a `packageDirectory` parameter to `generateManifestFileContents()` to specify where the package containing the manifest is located
- use this directory to make path-based dependencies relative
- remove the old compatibility wrapper `generatedManifestFileContents` which no longer makes sense (clients can easily switch to the regular function)
- adjust and extend unit test that covers manifest source generation for package dependencies
- in particular, that test is now more realistic to the kinds of packages that are actually useful to write

I considered also keeping a way to write out absolute path-based dependencies, but since resulting packages are inherently non-portable there didn't seem to be a reason to keep this.  It makes the API more complicated but we can add that if it's needed.

### Result
- package manifests written by the manifest source generation have absolute paths

rdar://100367134

### DIFF
--- a/Tests/WorkspaceTests/ManifestSourceGenerationTests.swift
+++ b/Tests/WorkspaceTests/ManifestSourceGenerationTests.swift
@@ -59,6 +59,7 @@ class ManifestSourceGenerationTests: XCTestCase {
 
             // Generate source code for the loaded manifest,
             let newContents = try manifest.generateManifestFileContents(
+                packageDirectory: packageDir,
                 toolsVersionHeaderComment: toolsVersionHeaderComment,
                 additionalImportModuleNames: additionalImportModuleNames)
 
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

Tests/WorkspaceTests/ManifestSourceGenerationTests.swift
```diff
@@ -239,24 +240,25 @@ class ManifestSourceGenerationTests: XCTestCase {
...
+                   .package(url: "https://example.com/MyPkg1", from: "1.0.0"),
+                   .package(url: "https://example.com/MyPkg2", .revision("58e9de4e7b79e67c72a46e164158e3542e570ab6")),
+                   .package(url: "https://example.com/MyPkg5", .exact("1.2.3")),
+                   .package(url: "https://example.com/MyPkg6", "1.2.3"..<"2.0.0"),
+                   .package(url: "https://example.com/MyPkg7", .branch("main")),
+                   .package(url: "https://example.com/MyPkg8", .upToNextMinor(from: "1.3.4")),
+                   .package(url: "ssh://git@example.com/MyPkg9", .branch("my branch with spaces")),
+                   .package(url: "../MyPkg10", from: "0.1.0"),
+                   .package(path: "../MyPkg11"),
+                   .package(path: "packages/path/to/MyPkg12"),
```
tomerd, line 252: feels like we lost some of the test cases here, should we bring them back?

abertelrud: I removed the ones that didn't make sense, e.g. `~`.  When I added those I wasn't really thinking about what made sense to create, so I replaced them with ones that make more sense.  I can add them back if you think it makes sense.

abertelrud: I can see the case for `~` with hand-authored packages, but for the mechanically generated manifests that doesn't make much sense.  But I can add back the test.

`file://` doesn't make sense at all for a `path:` so we should remove that.  I can file an issue to do that.  I'll add back the `~` tests.

abertelrud: Actually I'll add them all back and then file an issue to remove `file://`.

abertelrud: Added them back in 0dc348858a5100cb018de45d1d72e6a45eb8a0f0

abertelrud:
@swift-ci please smoke test

neonichu (approved):

tomerd (approved):

PR #5808, opened 2022-10-12
Revert "Error on empty product names"

DougGregor:
Reverts apple/swift-package-manager#5805.

This is an experimental revert.


DougGregor:
We don't need this. Closing

PR #5816, opened 2022-10-15
Tests: remove `-lstdc++` on Windows test expectations

compnerd:
Windows uses a different C++ runtime, which we rely on the linker driver to adjust the linker invocation (due to ABI incompatibilities between debug and release builds).  This allows the BuildPlan tests to now pass on Windows.


compnerd:
@swift-ci please smoke test

neonichu (approved):

PR #5817, opened 2022-10-16
PackagePlugin: make `Path` a wrapper around `AbsolutePath`

compnerd:
This should help the path representation be properly represented on different platforms.  This is motivated by Windows which would have the drive stripped from the path, converting to a relative path the absolute path.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -161,6 +161,9 @@ let package = Package(
         // bootstrap scripts build the deployable version.
         .target(
             name: "PackagePlugin",
+            dependencies: [
+                .product(name: "TSCBasic", package: "swift-tools-support-core"),
+            ],
             exclude: ["CMakeLists.txt"],
             swiftSettings: [
                 .unsafeFlags(["-package-description-version", "999.0"]),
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please smoke test

tomerd:
@abertelrud

compnerd:
Because that still needs to be tested more heavily before it can be ready to use.  There are issues that have been fixed in t-s-c that haven't been done in System yet.  Furthermore, that adds another dependency before this is really ready to be removed.  Since this is `@_implementationOnly` we can swap it later without it impacting the user.  I'd say that this makes it no more difficult to remove the t-s-c dependency.

abertelrud:
Seems reasonable though it's a bit concerning if FilePath still isn't ready for use.  @compnerd any idea on when that might be fixed?

tomerd:
@swift-ci smoke test

MaxDesiatov:
> Seems reasonable though it's a bit concerning if FilePath still isn't ready for use. @compnerd any idea on when that might be fixed?

+1 on this, what are the issues on the System repository to track feature parity of `FilePath` with `AbsolutePath`?

PR #5818, opened 2022-10-17
Make the code that deals with the remaining data in the NSFileHandle readabilityHandler be more robust

abertelrud:
There were cases in which this code in DefaultPluginScriptRunner could crash.  This change introduces local variables that keep both the handle and handler alive long enough to not crash rather than doing it all in one expression.  Note that this isn't a race condition — the handler does the correct locking.  But the object lifetime was a bit ambiguous with the existing line of code, and the new one removes the ambiguity.

No new unit tests — 21 of the existing unit tests cover this, and that's how the ambiguity was found.

rdar://101229820


abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

neonichu (approved):

PR #5827, opened 2022-10-19
Introduce `extraCXXFlags`, deprecate `extraCPPFlags`

MaxDesiatov:
### Motivation:

Usually, CPP is used as an abbreviation for the C preprocessor, while CXX more commonly refers to the C++ compiler. We also use `cxxSettings` in `Package.swift` manifest: https://developer.apple.com/documentation/packagedescription/target/cxxsettings.

### Modifications:

For this wider consistency, `extraCXXFlags` is introduced as a replacement for `extraCPPFlags`. The latter is then deprecated and is made a computed property that just returns the value of `extraCXXFlags`.

### Result:

Users who are coming from C++ background can be more familiar with the codebase. In the next schema version of `destination.json` (introduced in a future PR) we can use the new naming as well.

### DIFF
--- a/Sources/PackageModel/Destination.swift
+++ b/Sources/PackageModel/Destination.swift
@@ -64,10 +64,34 @@ public struct Destination: Encodable, Equatable {
 
     /// Additional flags to be passed to the Swift compiler.
     public let extraSwiftCFlags: [String]
+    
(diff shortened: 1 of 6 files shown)


MaxDesiatov:

neonichu (approved):

PR #5828, opened 2022-10-19
Switch away from using a deprecated Driver 'init' with 'externalTargetModulePathMap' argument.

artemcm:
Replacing it with 'externalTargetModuleDetailsMap'.


artemcm:
@swift-ci smoke test

neonichu (approved):

PR #5837, opened 2022-10-24
PackageModel: add `struct BuildFlags`, reduce reliance on TSC

MaxDesiatov:
### Motivation:

We're relying on `BuildFlags` from TSC for passing flags in some places, but there are also cases where flags are passed as separate properties, such as `extraCFlags`, `extraCXXFlags` etc.  `BuildFlags` is quite useful, as it allows grouping of these properties instead of passing them around separately. In most if not all cases they need to be passed all at once anyway. This apparently already let to issues, where `Destination` didn't have its own linker flags property, also see https://github.com/apple/swift-package-manager/pull/5793. We also had issues with inconsistent naming of these properties. Unifying everything under `BuildFlags` could prevent such issues from happening.

### Modifications:

Adding `BuildFlags` to `PackageModel`, mostly a copy of the declaration in TSC, but with a cleaned up initializer arguments naming. Modified uses of build tools flags to use this new struct instead of passing arounds properties separately.

### Result:

Reduced reliance on TSC, unified use of `BuildFlags` across the codebase.

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -423,7 +423,7 @@ public final class ClangTargetBuildDescription {
             args += ["-include", resourceAccessorHeaderFile.pathString]
         }
 
-        args += buildParameters.toolchain.extraCCFlags
+        args += buildParameters.toolchain.extraFlags.cCompilerFlags
         // User arguments (from -Xcc and -Xcxx below) should follow generated arguments to allow user overrides
         args += buildParameters.flags.cCompilerFlags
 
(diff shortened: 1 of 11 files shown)


MaxDesiatov:
@swift-ci please smoke test

MaxDesiatov:
@swift-ci please smoke test

tomerd (approved):
nice cleanup, thanks. 

should we also deprecate the TSC structure?

MaxDesiatov:
https://github.com/apple/sourcekit-lsp/pull/659
@swift-ci please smoke test

neonichu (approved):

PR #5840, opened 2022-10-25
Allow duplicate product names: vNext to v5.8

elsh:
Allow duplicate product names: vNext to v5.8
Resolves rdar://94744134


elsh:
@swift-ci smoke test

PR #5847, opened 2022-10-27
[5.7] Avoid redundant checks in cycle detection

fcanas:
Dramatically improve speed of cycle detection for large or well-connected manifest graphs.

### Motivation:

Loading a project with a large or well-connected package graph in Xcode can be very slow. 

Xcode uses `Workspace.loadPackageGraph(rootPath:...)`, which calls into
calls into [`findCycle()`](https://github.com/apple/swift-package-manager/blob/25c671ef3ef2bd7a245403f9848cb3be3321c790/Sources/PackageGraph/PackageGraph%2BLoading.swift#L866). `findCycle` has a bug leading to slower than linear cycle detection.

### Modifications:

There is a note to convert to use an algorithm tracking traversal with a stack, which would allow for linear time cycle detection. But given the output of `findCycle` is the path to first the cycle, and the cycle itself, not all the elements in a strongly connected subgraph, it's enough to memoize fully explored nodes.

Once all of a Manifests outbound edges have been traversed for cycles, it's not possible for it to participate in a cycle. So if a fully-explored Manifest is encountered again in a cycle check, exit early.

Without such an early exit, cycle detection is slower than linear. This change makes it linear.

### Result:

I made a tool to generate and profile the loading of package graphs through the same path as Xcode [here](https://github.com/fcanas/manifest-cycle-bug).

These are the results of profiling package loading a small range of number of packages.

| Number of packages | apple/spm | memoized |
|--------------------|-----------|----------|
| 22                 | 17.9s     | 0.2s     |
| 23                 | 35.8s     | 0.2s     |
| 24                 | 88.1s     | 0.2s     |
| 25                 | 160.5s    | 0.2s     |
| 26                 | 323.4s    | 0.2s     |

### DIFF
--- a/Sources/PackageGraph/PackageGraph+Loading.swift
+++ b/Sources/PackageGraph/PackageGraph+Loading.swift
@@ -872,6 +872,8 @@ fileprivate func findCycle(
 ) rethrows -> (path: [Manifest], cycle: [Manifest])? {
     // Ordered set to hold the current traversed path.
     var path = OrderedCollections.OrderedSet<Manifest>()
+    
+    var fullyVisitedManifests = Set<Manifest>()
 
     // Function to visit nodes recursively.
     // FIXME: Convert to stack.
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci please smoke test

fcanas:
My mistake. The tests need adjustment from the main branch version.

I'll push a fix soon.

MaxDesiatov:
@swift-ci please smoke test

MaxDesiatov:
@swift-ci please smoke test

fcanas:
@tomerd Is there anything else I can do to move this change along?

tomerd:
thanks @fcanas, the 5.7 release process is carefully managed. we will merge this if / when there is an opportunity to take it forward

tomerd (approved):

PR #5877, opened 2022-11-04
Basics: fix `DiagnosticsHanderWrapper` typo

MaxDesiatov:
### Motivation:

This type is `private`, so plain renaming should be safe.

### Modifications:

Applied renaming to `DiagnosticsHandlerWrapper` within `Basics/Observability.swift`. Also fixed a typo in a comment in the same file.

### Result:

No more typo here, the type easier to find with textual/symbol search.


MaxDesiatov:
@swift-ci please smoke test

neonichu (approved):

PR #5889, opened 2022-11-10
Commands: mark `class SwiftTool` as explicitly `final`

MaxDesiatov:
### Motivation:

This class is already not marked as `open`, but making it `final` explicitly signals that it is not and shouldn't be subclassed within the `Commands` module.

### Modifications:

Add `final` modifier to `public class SwiftTool` in the `Commands` module.

### Result:

When reading the source code, it's immediately clear that `SwiftTool` isn't and shouldn't be subclassed.


MaxDesiatov:
@swift-ci please smoke test

neonichu (approved):

MaxDesiatov:
@swift-ci please smoke test

tomerd (approved):

PR #5898, opened 2022-11-15
Verify that the pinned version is still available

### LINKED ISSUE #5873: Error message for retracted version in Package.resolved is unhelpful
### Description

The SQLite.swift package recently deleted the tag for 0.14 due to a bug: https://github.com/stephencelis/SQLite.swift/blob/1235d44cd3e8596e842a47c03efbcbdb61797024/CHANGELOG.md

Unfortunately for me, I happened to create a project that used this package while 0.14 existed, so this was the version number that was stored in the Package.resolved.

When I later tried to add a new package to the project, it was unable to resolve, but the error message was not very helpful:
```
Failed to resolve dependencies 
Dependencies could not be resolved because no versions of 'sqlite.swift' match the requirement 0.9.2..<1.0.0 and root depends on 'sqlite.swift' 0.9.2..<1.0.0.
```

### Expected behavior

The error message should explain that the version in Package.resolved is no longer available.

### Actual behavior

The error message from the description.

### Steps to reproduce

1. Clone the example project: https://github.com/dtweston/ExampleApp
2. Open the project (ExampleApp/ExampleApp.xcodeproj) in Xcode (v14.1)
3. Observe the resolve packages error in the Report navigator.

Swift Package Manager - Swift 5.7.1

swift-driver version: 1.62.15 Apple Swift version 5.7.1 (swiftlang-5.7.1.135.3 clang-1400.0.29.51)
Target: x86_64-apple-macosx12.0
Darwin ip-address.redacted.com 21.6.0 Darwin Kernel Version 21.6.0: Thu Sep 29 20:12:57 PDT 2022; root:xnu-8020.240.7~1/RELEASE_X86_64 x86_64

dtweston:
Before committing to the pinned version, the resolver should make sure that it's still available. This change fixes #5873.

### Motivation:

If the resolver decides to honor a pin that is no longer available, the error message produced is `Dependencies could not be resolved because no versions of 'package' match the requirement 1.0.0..<2.0.0 and root depends on 'package' 1.0.0..<2.0.0`. However, that error message is confusing, because the requirement specified in the error message was replaced by the stricter version requirement that was pinned for the project. 

Since pinned versions are discarded if they no longer meet the requirements, it seemed reasonable to do the same if the pinned version is no longer available.

### Modifications:

The resolver first checks to see if the pinned version of the package is still available before using it as a requirement.

### Result:

After this change, the package resolution will no longer fail if a pinned version of a package is no longer available. Instead, the pinned version will be ignored, and the resolver will choose the most recent package version that meets the constraints.

### DIFF
--- a/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
+++ b/Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
@@ -1115,12 +1115,16 @@ internal final class PubGrubPackageContainer {
     /// Returns the best available version for a given term.
     func getBestAvailableVersion(for term: Term) throws -> Version? {
         assert(term.isPositive, "Expected term to be positive")
-        var versionSet = term.requirement
+        let versionSet = term.requirement
 
         // Restrict the selection to the pinned version if is allowed by the current requirements.
         if let pinnedVersion = self.pinnedVersion {
             if versionSet.contains(pinnedVersion) {
-                versionSet = .exact(pinnedVersion)
+                // Make sure the pinned version is still available
+                let version = try self.underlying.versionsDescending().first { pinnedVersion == $0 }
+                if version != nil {
+                    return version
+                }
             }
         }
 
--- a/Tests/PackageGraphTests/PubgrubTests.swift
+++ b/Tests/PackageGraphTests/PubgrubTests.swift
@@ -1426,6 +1426,34 @@ final class PubgrubTests: XCTestCase {
         ])
     }
 
+    func testMissingPin() throws {
+        // This checks that we can drop pins that are no longer available but still keep the ones
+        // which fit the constraints.
+        try builder.serve("a", at: v1, with: ["a": ["b": (.versionSet(v1Range), .specific(["b"]))]])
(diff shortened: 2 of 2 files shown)


tomerd:
@swift-ci smoke test

Sources/PackageGraph/Pubgrub/PubgrubDependencyResolver.swift
```diff
@@ -1115,12 +1115,16 @@ internal final class PubGrubPackageContainer {
...
-        var versionSet = term.requirement
+        let versionSet = term.requirement
 
         // Restrict the selection to the pinned version if is allowed by the current requirements.
         if let pinnedVersion = self.pinnedVersion {
             if versionSet.contains(pinnedVersion) {
-                versionSet = .exact(pinnedVersion)
+                // Make sure the pinned version is still available
+                let version = try self.underlying.versionsDescending().first { pinnedVersion == $0 }
+                if version != nil {
```
tomerd, line 1125: `if let version = ...`

tomerd (approved):
this seems right to me. @neonichu?

neonichu (approved):

PR #5905, opened 2022-11-16
PackageModel: rename "CC SDKs" to "destinations"

MaxDesiatov:
### Motivation:

"Cross-compilation SDK" is a bit of a mouthful to say, and we've previously denoted this as a "destination". After some consideration, I think calling "cross-compilation SDKs" as "destinations" simplifies a lot of things. Additionally, "target triple" and "target platform" are terms of art after all. Renaming them to "destination triple" and "destination platform" arguably was a mistake. Since we had no tagged releases in the meantime, let's clean up the naming before it's too late.

### Modifications:

`--experimental-cross-compilation-sdks-path` is renamed to `--experimental-destinations-path`, `sharedCCSDKDirectory` to `sharedDestinationsDirectory`. `destinationTriple` is renamed to `targetTriple`.

### Result:

More clear, widely consistent naming.

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -85,7 +85,7 @@ public struct LocationOptions: ParsableArguments {
     public var customCompileDestination: AbsolutePath?
 
     @Option(name: .customLong("experimental-destinations-path"), help: .hidden, completion: .directory)
-    var destinationsDirectory: AbsolutePath?
+    var crossCompilationDestinationsDirectory: AbsolutePath?
 }
 
 public struct CachingOptions: ParsableArguments {
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -84,8 +84,8 @@ public struct LocationOptions: ParsableArguments {
     @Option(name: .customLong("destination"), help: .hidden, completion: .directory)
     public var customCompileDestination: AbsolutePath?
 
-    @Option(name: .customLong("experimental-cross-compilation-sdks-path"), help: .hidden, completion: .directory)
-    var ccSDKsDirectory: AbsolutePath?
+    @Option(name: .customLong("experimental-destinations-path"), help: .hidden, completion: .directory)
+    var destinationsDirectory: AbsolutePath?
 }
 
 public struct CachingOptions: ParsableArguments {
--- a/Sources/CoreCommands/SwiftTool.swift
+++ b/Sources/CoreCommands/SwiftTool.swift
@@ -162,7 +162,7 @@ public final class SwiftTool {
     public let sharedConfigurationDirectory: AbsolutePath?
 
     /// Path to the cross-compilation SDK directory.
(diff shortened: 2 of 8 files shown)


MaxDesiatov:
@swift-ci please smoke test

neonichu (approved):

Sources/CoreCommands/SwiftTool.swift
```diff
@@ -162,7 +162,7 @@ public final class SwiftTool {
     public let sharedConfigurationDirectory: AbsolutePath?
 
     /// Path to the cross-compilation SDK directory.
-    let sharedCCSDKDirectory: AbsolutePath?
+    let sharedDestinationsDirectory: AbsolutePath?
```
tomerd, line 165: nit: I would keep the cross compilation name here to make it easy to reason about

Sources/CoreCommands/Options.swift
```diff
@@ -84,8 +84,8 @@ public struct LocationOptions: ParsableArguments {
     @Option(name: .customLong("destination"), help: .hidden, completion: .directory)
     public var customCompileDestination: AbsolutePath?
 
-    @Option(name: .customLong("experimental-cross-compilation-sdks-path"), help: .hidden, completion: .directory)
-    var ccSDKsDirectory: AbsolutePath?
+    @Option(name: .customLong("experimental-destinations-path"), help: .hidden, completion: .directory)
```
tomerd, line 87: 👍

Sources/CoreCommands/Options.swift
```diff
@@ -84,8 +84,8 @@ public struct LocationOptions: ParsableArguments {
...
+    var destinationsDirectory: AbsolutePath?
```
tomerd, line 88: nit: I would keep the "cross compilation" in the variable name here to make it easy to reason about

Sources/CoreCommands/SwiftTool.swift
```diff
@@ -781,16 +781,16 @@ private func getSharedCacheDirectory(options: GlobalOptions, fileSystem: FileSys
     }
 }
 
-private func getSharedCCSDKsDirectory(
+private func getSharedDestinationsDirectory(
```
tomerd, line 784: nit: I would keep the "cross compilation" in the function name here to make it easy to reason about

Sources/CoreCommands/SwiftTool.swift
MaxDesiatov: Would "CC" as an abbreviation be ok here, as in `sharedCCDestinationsDirectory`? I remember @neonichu had an objection to "CC" as an abbreviation, as it can be confused with something like "C Compiler". OTOH `sharedCrossCompilationDestinationsDirectory` seems too verbose

tomerd: I am fan of explicit variable names (even if verbose) since they make maintenance / reading code much easier

MaxDesiatov:
@swift-ci please smoke test

PR #5912, opened 2022-11-18
enable scm to regsitry substitution

tomerd:
motivation: support registry based depedendencies

changes: enabe the "swizzle" mode by default, which would replace a dependency that comes from source control with one that comes frmo the regsitry when the identity mapping is known


tomerd:
@swift-ci smoke test

neonichu (approved):

PR #5922, opened 2022-11-22
SwiftTool: introduce `--experimental-destination-selector` option

MaxDesiatov:
Depends on #5911.

### Motivation:

With support for listing cross-compilation destinations with new `swift experimental-destinations list` command, we also need a way to actually use those. New `--experimental-destination-selector` option is added, through which users can either select a destination via an ID or a target triple.

### Modifications:

Added new `crossCompilationDestinationSelector` to `BuildOptions`. Modified `SwiftTool` to select a destination from this selector string. Implemented destination selection based on a selector in `DestinationsBundle.swift`. Refactored `Workspace.Location` to create shared locations in a unified way. Added a simple test for destination selection.

### Result:

Cross-compilation destinations can be selected when building a package.


MaxDesiatov:
@swift-ci please smoke test

MaxDesiatov:
@swift-ci please smoke test

MaxDesiatov:
I think `@swift-ci` doesn't support stacked PRs yet? I don't have any other explanation for current CI failures.

Sources/PackageModel/Destination.swift
```diff
@@ -103,7 +103,7 @@ public struct Destination: Encodable, Equatable {
     }
     
     /// Additional flags to be passed to the build tools.
-    public let extraFlags: BuildFlags
+    public var extraFlags: BuildFlags
```
tomerd, line 106: does it need to be mutated?

MaxDesiatov: Yes, it happens on https://github.com/apple/swift-package-manager/pull/5922/files#diff-ba866b666cdc2795c59e6713e0817015f0076f5447541e8c477db717cbb6baf3R114

Since it's a struct, I think making this `var` is safe, as any mutations are local and don't have side effects?

Sources/PackageModel/Manifest.swift
```diff
@@ -180,14 +180,14 @@ public final class Manifest {
         }
 
         var requiredDependencies: Set<PackageIdentity> = []
-        for targetTriple in self.targetsRequired(for: products) {
-            for targetDependency in targetTriple.dependencies {
+        for linuxGNUTargetTriple in self.targetsRequired(for: products) {
```
tomerd, line 183: ?

Sources/PackageModel/Destination.swift
tomerd: 👍

Sources/PackageModel/Manifest.swift
MaxDesiatov: automated Xcode renaming is bad at this, sorry...

Sources/SPMTestSupport/Observability.swift
```diff
@@ -48,7 +48,7 @@ public struct TestingObservability {
         self.collector.hasWarnings
     }
 
-    struct Collector: ObservabilityHandlerProvider, DiagnosticsHandler, CustomStringConvertible {
+    final class Collector: ObservabilityHandlerProvider, DiagnosticsHandler, CustomStringConvertible {
```
MaxDesiatov, line 51: Making this a `class` because `diagnostics: ThreadSafeArrayStore` is a class instance, so there's no actual value semantics here even with a `struct`.

Sources/Workspace/Workspace.swift
```diff
@@ -3559,62 +3559,68 @@ extension Workspace.Location {
...
+            getOrCreateHandler: {
+                try $0.getOrCreateSwiftPMCacheDirectory()
+            },
+            warningHandler: warningHandler
+        )
+
+        try location.validate(
+            keyPath: \.sharedCrossCompilationDestinationsDirectory,
+            fileSystem: fileSystem,
+            getOrCreateHandler: {
```
tomerd, line 3592: use the function pointer?

neonichu (approved):

MaxDesiatov:
@swift-ci please smoke test

MaxDesiatov:
@swift-ci please smoke test

MaxDesiatov:
@swift-ci smoke test

PR #5929, opened 2022-11-29
Only automatically link snippets to product libraries

bitjammer:
Snippets illustrate how to use a package's products by design, so they shouldn't link to library targets that aren't exported as products.

If snippets link to all library targets, one negative result would be inadvertently linking to test support libraries, which could cause failures due to automatic linking of `@rpath/libXCTestSwiftSupport.dylib`, a library not available to snippets at runtime.

rdar://102746374


bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please smoke test

neonichu (approved):

PR #5931, opened 2022-11-29
Consider snippets in ResolvedProduct.executableTarget

bitjammer:
When computing the underlying executable target for a resolved product, consider snippet targets as well. Otherwise, this results in a fatal error when building a package that contains snippets.

rdar://102784273


bitjammer:
@swift-ci Please smoke test

neonichu (approved):

PR #5935, opened 2022-11-30
improve internal target dependencies

tomerd:
motivation: more accuratly reflect internal target dependenies

changes: PackageGraph does not depend on SourceControl or Registry, only Workspace does


tomerd:
@swift-ci smoke test

neonichu (approved):

PR #5962, opened 2022-12-08
Add macOS version to Package.swift for newly generated packages

keith:
Previously if you created a new basic Swift package and added a dependency, such as swift-syntax, that requires a minimum version greater than macOS 10.13, you would have to set the platform version to something above that in the Package.swift. With this change newly generated packages on macOS automatically generate the current macOS version as the minimum supported version, and if the developer wants a lower version they can reduce it as usual in the Package.swift.

### DIFF
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import Basics
+import Foundation
 import TSCBasic
 import PackageModel
 
@@ -31,7 +32,7 @@ public final class InitPackage {
 
         public init(
             packageType: PackageType,
-            platforms: [SupportedPlatform] = []
+            platforms: [SupportedPlatform]
         ) {
             self.packageType = packageType
             self.platforms = platforms
@@ -87,7 +88,7 @@ public final class InitPackage {
     ) throws {
         try self.init(
             name: name,
-            options: InitPackageOptions(packageType: packageType),
+            options: InitPackageOptions(packageType: packageType, platforms: getDefaultPlatforms()),
             destinationPath: destinationPath,
             fileSystem: fileSystem
         )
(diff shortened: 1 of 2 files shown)


keith:
@swift-ci please test

keith:
https://forums.swift.org/t/pitch-generate-package-swift-files-with-newer-macos-versions/61925

tomerd:
cc @bitjammer

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -317,6 +316,30 @@ class InitTests: XCTestCase {
         }
     }
 
+    func testDefaultPlatform() throws {
+        #if !os(macOS)
+        // Platform default versions are only set on macOS
+        try XCTSkipIf(true, "test is only supported on macOS")
```
tomerd, line 322: is this really needed?

Sources/Workspace/InitPackage.swift
```diff
@@ -498,3 +499,13 @@ extension SupportedPlatform {
         }
     }
 }
+
+private func getDefaultPlatforms() -> [SupportedPlatform] {
+    #if os(macOS)
```
tomerd, line 504: I am not sure about this. most swift developers develop on a fairly recent Mac regardless on what the target deployment platforms are

keith: are you commenting on the functionality in general or the `#if` here? if you mean the functionality in general should that be discussed on the pitch? https://forums.swift.org/t/pitch-generate-package-swift-files-with-newer-macos-versions/61925

I just don't want to dup the justifications here if that answers some of these

Tests/WorkspaceTests/InitTests.swift
keith: Right now it would still be empty if you ran on linux, so i think until that part changes we'd have to skip? is there another way of writing this test that you're suggesting could avoid this?

PR #5969, opened 2022-12-12
Add package metadata to observability scopes used by PackageContainers

abertelrud:
`PackageContainer` is a protocol that is implemented by various types involved in SwiftPM package resolution.  They can emit messages but those messages don't get associated with the diagnostics.

There are currently three implementors of this protocol:  `FileSystemPackageContainer`, `SourceControlPackageContainer`, and `RegistryPackageContainer`.  All three just assign a given `ObservabilityScope`, and setting a package location metadata on that scope is a good way to make sure that all the diagnostics emitted to that scope get associated with the package.

This fix makes sure that diagnostics emitted through these scopes get properly associated with the package.  This is particularly useful for manifest-related diagnostics, which inherit the scope used by the various kinds of `PackageContainer`.

### Changes
- add `diagnosticsMetadata` property in extension on `PackageRerference`, similar to the one on `Package`
- use this property to associate package metadata with the `ObservabilityScope`
- add this metadata with a couple of standalone `emit` calls in `Workspace`

rdar://103229985

### DIFF
--- a/Sources/PackageModel/PackageReference.swift
+++ b/Sources/PackageModel/PackageReference.swift
@@ -173,6 +173,12 @@ extension PackageReference: Hashable {
     }
 }
 
+extension PackageReference {
+    public var diagnosticsMetadata: ObservabilityMetadata {
+        return .packageMetadata(identity: self.identity, kind: self.kind)
+    }
+}
+
 extension PackageReference: CustomStringConvertible {
     public var description: String {
         return "\(self.identity) \(self.kind)"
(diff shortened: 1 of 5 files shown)


abertelrud:
Looking for the best unit tests to extend with additional checks.

abertelrud:
@swift-ci please smoke test

neonichu (approved):

Sources/PackageModel/PackageReference.swift
```diff
@@ -173,6 +173,12 @@ extension PackageReference: Hashable {
     }
 }
 
+extension PackageReference {
+    public var diagnosticsMetadata: ObservabilityMetadata {
+        return .packageMetadata(identity: self.identity, kind: self.kind)
+    }
+}
```
tomerd, line 180: could this be a non-public extension closer to the call site? seems fairly leaky as a public property of the model

PR #5977, opened 2022-12-15
Update old link to Swift Jira in `README.md`

MaxDesiatov:
While bugs.swift.org no longer hosts Jira and redirect's to `swift` repository issue tracker, that's still wrong. We should direct people to SwiftPM's issue tracker.


MaxDesiatov:
@swift-ci please smoke test

neonichu (approved):

PR #5981, opened 2022-12-17
PackageModel: partially address #5719

compnerd:
Handle `-use-ld=` from the configuration in the selection of the librarian.  We do not honour the `-Xmanifest` or `-Xswiftc` flags being passed on the command line rather than the plist as we cannot pre-compute the flags.

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -128,7 +128,8 @@ public final class UserToolchain: Toolchain {
         binDir: AbsolutePath,
         useXcrun: Bool,
         environment: EnvironmentVariables,
-        searchPaths: [AbsolutePath]
+        searchPaths: [AbsolutePath],
+        extraSwiftFlags: [String]
     ) throws
         -> AbsolutePath
     {
@@ -145,7 +146,12 @@ public final class UserToolchain: Toolchain {
                 {
                     return librarian.basename
                 }
-                // TODO(5719) use `lld-link` if the build requests lld.
+                // TODO(5719) handle `-Xmanifest` vs `-Xswiftc`
+                // `-use-ld=` is always joined in Swift.
+                if let ld = extraSwiftFlags.first(where: { $0.starts(with: "-use-ld=") }) {
+                    let linker = String(ld.split(separator: "=").last!)
+                    return linker == "lld" ? "lld-link" : linker
+                }
                 return "link"
             }
             // TODO(compnerd) consider defaulting to `llvm-ar` universally with
@@ -449,14 +455,6 @@ public final class UserToolchain: Toolchain {
         // Use the triple from destination or compute the host triple using swiftc.
         var triple = destination.targetTriple ?? Triple.getHostTriple(usingSwiftCompiler: swiftCompilers.compile)
 
-        self.librarianPath = try UserToolchain.determineLibrarian(
-            triple: triple,
-            binDir: binDir,
-            useXcrun: useXcrun,
-            environment: environment,
-            searchPaths: envSearchPaths
-        )
-
         // Change the triple to the specified arch if there's exactly one of them.
         // The Triple property is only looked at by the native build system currently.
         if let archs = self.architectures, archs.count == 1 {
@@ -471,6 +469,15 @@ public final class UserToolchain: Toolchain {
             triple: triple,
             destination: destination,
             environment: environment
+        )
+
+        self.librarianPath = try UserToolchain.determineLibrarian(
+            triple: triple,
+            binDir: binDir,
+            useXcrun: useXcrun,
+            environment: environment,
+            searchPaths: envSearchPaths,
+            extraSwiftFlags: self.extraFlags.swiftCompilerFlags
         )
 
         if let sdkDir = destination.sdkRootDir {
--- a/Tests/PackageModelTests/PackageModelTests.swift
+++ b/Tests/PackageModelTests/PackageModelTests.swift
@@ -39,7 +39,7 @@ class PackageModelTests: XCTestCase {
         checkCodable(.executable)
         checkCodable(.test)
     }
-    
+
     func testProductFilterCodable() throws {
         // Test ProductFilter.everything
         try {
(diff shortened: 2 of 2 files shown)


Sources/PackageModel/UserToolchain.swift
```diff
@@ -128,7 +129,12 @@ public final class UserToolchain: Toolchain {
                                              environment: environment) {
                     return librarian.basename
                 }
-                // TODO(5719) use `lld-link` if the build requests lld.
+                // TODO(5719) handle `-Xmanifest` vs `-Xswiftc`.
+                // `-use-ld=` is always joined in Swift.
+                if let ld = extraSwiftFlags.first(where: { $0.starts(with: "-use-ld=") }) {
```
neonichu, line 134: Seems a bit unfortunate to have to parse the flags like this, can/should we ask users to configure alternative linkers more explicitly?

compnerd: I agree it is unfortunate.  How would you recommend that we set that up?  The general mechanism for this is `-use-ld=...` which with CMake you would pass via `CMAKE_Swift_FLAGS` and that is honoured everywhere.

tomerd: if we have to parse, could we use argument parser to help make this more reliable?

tomerd:
can we add a test?

Sources/PackageModel/UserToolchain.swift
compnerd: I'm not sure if that is going to be any better because we will need to consider the case of `-Xswiftc` or `-Xclang-linker` where the option itself is made opaque.  We would need to process both the driver and frontend flags and possibly non-swift-tool flags.

Sources/PackageModel/UserToolchain.swift
```diff
@@ -388,6 +392,12 @@ public final class UserToolchain: Toolchain {
         self.extraFlags = BuildFlags()
 
         self.extraFlags.swiftCompilerFlags = try Self.deriveSwiftCFlags(triple: triple, destination: destination, environment: environment)
+        self.librarianPath =
+            try Self.determineLibrarian(triple: triple, binDir: binDir,
```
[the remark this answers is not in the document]
compnerd: Not sure how to format this ... if only we could `swift-format` the problem away ...

compnerd:
BTW, I think that this might be something we want to pull into 5.8

Sources/PackageModel/UserToolchain.swift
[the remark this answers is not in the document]
tomerd: this is what we normally follow

```
self.librarianPath = try Self.determineLibrarian(
  triple: triple, 
  binDir: binDir,
  useXcrun: useXcrun,
  environment: environment,
  searchPaths: envSearchPaths,
  extraSwiftFlags: self.extraFlags.swiftCompilerFlags
)
```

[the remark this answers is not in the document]
tomerd: we can (and should) run swift format against files changed! its even in the readme:  https://github.com/apple/swift-package-manager/blob/main/CONTRIBUTING.md#creating-pull-requests

we should ideally enforce that too in PR checking like we do for swift server libraries. @shahmishal is this something you can help set up?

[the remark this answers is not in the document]
compnerd: Sure, #6090 pre-formats the file so that it is easier to see the diffs here.  I'll reformat the file with `swift format`.

compnerd:

compnerd:
@swift-ci please smoke test

compnerd:
> can we add a test?

Done

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

Sources/PackageModel/UserToolchain.swift
```diff
@@ -142,7 +143,12 @@ public final class UserToolchain: Toolchain {
         {
           return librarian.basename
         }
-        // TODO(5719) use `lld-link` if the build requests lld.
+        // TODO(5719) handle `-Xmanifest` vs `-Xswiftc`
```
tomerd, line 146: is this comment relevant?

Sources/PackageModel/UserToolchain.swift
```diff
@@ -142,7 +143,12 @@ public final class UserToolchain: Toolchain {
...
+        // `-use-ld=` is always joined in Swift.
+        if let ld = extraSwiftFlags.first(where: { $0.starts(with: "-use-ld=") }) {
```
tomerd, line 148: string parsing seems fragile, is there a better way?

Sources/PackageModel/UserToolchain.swift
```diff
@@ -455,6 +457,10 @@ public final class UserToolchain: Toolchain {
     self.extraFlags.swiftCompilerFlags = try Self.deriveSwiftCFlags(
       triple: triple, destination: destination, environment: environment)
 
+    self.librarianPath = try UserToolchain.determineLibrarian(
```
tomerd, line 460: we normally format each argument in its own line

MaxDesiatov: This also seems to be using 2 spaces, while the rest of the codebase uses 4

Sources/PackageModel/UserToolchain.swift
compnerd: Yes, it is, we still need to refine it further for more ways that the flags can be specified.

Sources/PackageModel/UserToolchain.swift
compnerd: I really don't see a better way unfortunately.  The driver flags, frontend flags, and non-Swift tool flags are all intertwined and will impact this.  I think that this might be the best approach (at least for now).

Sources/PackageModel/UserToolchain.swift
compnerd: This was what `swift format` did 🤷‍♂️

compnerd: Reformatted with SwiftFormat rather than swift-format.

compnerd:
@swift-ci please smoke test

Sources/PackageModel/UserToolchain.swift
compnerd: https://github.com/apple/swift-package-manager/pull/5981/commits/d77f51d6c15d4ea63cece9793008d6838d93ee56#diff-adf7999cea9f76ad6a7d320c04fbfa7222d4c7d20bd27f81667c02cf88081897R140 BTW is where this becomes Windows specific.

tomerd:
would rather see a more idiomatic solution, but fine with this change if it only impact building on windows.

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

tomerd (approved):

PR #5984, opened 2022-12-21
Add package metadata to a few more ObservabilityScopes in workspace-wide and package graph loading operations

abertelrud:
Fix more cases in which there is a relevant package or package reference, but where the `ObservabilityScope` didn't have the metadata.

### Motivation:

Clients of libSwiftPM that present diagnostics use the package metadata that can optionally be attached to an emitted `Diagnostic` to associate it with the package in the user interface.
### Modifications:

This change addresses a few more cases in which no package metadata was associated.  Where possible, the metadata is associated with the `ObservabilityScope` at the highest place in the call stack so that any diagnostic emitted below it gets the package metadata.

- create child scopes with associated package metadata
- add the metadata directly to a couple of `emit()` calls where it wasn't clearly correct to create a new child scope
- extend a unit test

### Result:

The diagnostic's metadata will have the package identity in more of the cases where there is a relevant package.

rdar://103561534

### DIFF
--- a/Sources/PackageGraph/PackageGraph+Loading.swift
+++ b/Sources/PackageGraph/PackageGraph+Loading.swift
@@ -353,7 +353,7 @@ private func createResolvedPackages(
         )
 
         // Create target builders for each target in the package.
-        let targetBuilders = package.targets.map{ ResolvedTargetBuilder(target: $0, observabilityScope: observabilityScope) }
+        let targetBuilders = package.targets.map{ ResolvedTargetBuilder(target: $0, observabilityScope: packageObservabilityScope) }
         packageBuilder.targets = targetBuilders
 
         // Establish dependencies between the targets. A target can only depend on another target present in the same package.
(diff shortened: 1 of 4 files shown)


Sources/PackageGraph/PackageGraph+Loading.swift
```diff
@@ -353,7 +353,7 @@ private func createResolvedPackages(
         )
 
         // Create target builders for each target in the package.
-        let targetBuilders = package.targets.map{ ResolvedTargetBuilder(target: $0, observabilityScope: observabilityScope) }
+        let targetBuilders = package.targets.map{ ResolvedTargetBuilder(target: $0, observabilityScope: packageObservabilityScope) }
```
abertelrud, line 356: In this case there was actually an observability child scope with the metadata, it just wasn't being passed down.

abertelrud:
@swift-ci please smoke test

tomerd (approved):

Issue #5968, opened 2022-12-11
Unintentional but unacceptable diagnostic language


dabrahams:
### Description

Not sure this bug really belongs on SPM. I just built swift-format and the build ended with:

remark: Incremental compilation has been disabled: it is not compatible with **whoremark**: Incremental compilation has been disabled: it is not compatible with **whoremark**: Incremental compilation has been disabled: it is not compatible with **whoremark**: Incremental compilation has been disabled: it is not compatible with **whoremark**: Incremental compilation has been disabled: it is not compatible with **whoremark**: Incremental compilation has been disabled: it is not compatible with **whoremark**: Incremental compilation has been disabled: it is not compatible with who[11/16] Co[25/25] Compiling SwiftFormatTestSupport DiagnosingTestCase.swift
Build complete! (103.43s)

Something needs to be inserting a newline in the output before all remarks.  

### Steps to reproduce

Build swift-format from the CLI with swift 5.7

abertelrud:
Almost certainly the remark refers to "whole module optimization".  Seems to be two bugs:
1. the message cut off
2. it needs a newline, as you say
Taking a look.

abertelrud:
Also seems unfortunate to be getting that same message so many times; either SwiftPM should dedupe or maybe that's something in the driver.  In any case the missing newline seems to be the first thing to address here, and that is almost certainly on the SwiftPM side.

abertelrud:
So this looks like a combination of two things:
1. `swift-driver` emits both its JSON output and raw remarks to `stderr` (this seems unfortunate)
2. `swift-package` tries to interpret the JSON packets from the driver but when it can't it just emits the raw bytes, and ends up losing the newline (still looking into that one)

For example:
```
swiftc -parseable-output -whole-module-optimization -incremental -c foo.swift
remark: Incremental compilation has been disabled: it is not compatible with whole module optimization
1201
{
  "command_arguments" : [
    "-frontend",
    "-c",
    "foo.swift",
    "-target",
    "arm64-apple-macosx12.0",
    "-Xllvm",
    "-aarch64-use-tbi",
    "-enable-objc-interop",
    "-stack-check",
. . .
```

abertelrud:
Filed https://github.com/apple/swift-driver/issues/1241 on the Swift Driver issue and will see what we do in SwiftPM about the current state of affairs.  Since the output is always line-based, we should be able to do better in the JSON parser than we are currently doing.

compnerd:
[...]

PR #5987, opened 2022-12-21
SwiftPM outputs raw diagnostics from SwiftDriver without trailing newlines [5968]

abertelrud:
SwiftPM uses the Swift Driver's `-parseable-output` flag to get a sequence of length-prefixed JSON-encoded messages that it can read.  Unfortunately the Swift Driver doesn't JSON-encode its own diagnostics, leading to things like https://github.com/apple/swift-package-manager/issues/5968.

I filed https://github.com/apple/swift-driver/issues/1241 on the Swift Driver to get it to encode its JSON messages, but meanwhile, it turns out that we can fairly easily fix the fallback logic that SwiftPM uses when it emits raw output.  It was simply missing a newline.  It's safe to always add one, since the logic that parses JSON splits by newlines, so we won't find any terminating newlines.

### Modifications:

- add a newline after emitting the unparseable line of Swift Driver output
- add a unit test

### Result:

Raw output from the Swift Driver has trailing newlines.

rdar://103608636

### DIFF
--- a/Sources/Build/SwiftCompilerOutputParser.swift
+++ b/Sources/Build/SwiftCompilerOutputParser.swift
@@ -162,7 +162,7 @@ extension SwiftCompilerOutputParser: JSONMessageStreamingParserDelegate {
             return
         }
 
-        let message = SwiftCompilerMessage(name: "unknown", kind: .unparsableOutput(text))
+        let message = SwiftCompilerMessage(name: "unknown", kind: .unparsableOutput(text + "\n"))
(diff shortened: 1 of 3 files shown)


abertelrud:
@swift-ci please smoke test

tomerd (approved):

neonichu (approved):

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

abertelrud:
@swift-ci please smoke test

PR #5988, opened 2022-12-21
Add quiet flag (-q, --quiet) to LoggingOptions.

### LINKED ISSUE #4395: [SR-15037] Add --quiet option to `swift run`
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-15037      |
   |Radar             | rdar://problem/81620352         |
   |Original Reporter | @BenchR267      |
   |Type              | New Feature    |

   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 1         |
   |Component/s       | Package Manager    |
   |Labels            | New Feature        |
   |Assignee          | None      |
   |Priority          | Medium      |

   

   md5: c5f31e82419529323d967dcf872f390c

  </details>

**Issue Description:**

\`swift run\` is often used for running Swift scripts. When running those scripts from other scripts, the output is in an expected format and should not contain any output from SwiftPM itself.  
Currently, all output from SwiftPM seems to be printed on stderr so it's an option to nuke that in those situations (\`2\>\>/dev/null\`). But that doesn't work if the script itself is using stderr too.  
I propose to add a new option to \`swift run\` which suppresses all output about building the script and only prints the output of the executable itself, \`--quiet\`.  
Open for discussion is how build failures are surfaced to the user and I don't have a strong opinion on that. It could suppress that too making the command fail silently and only provide the exit status as indicator for failure or it could print those out instead since the executable will never run anyway.

clayellis:
This PR adds a quiet flag (`-q`, `--quiet`) to `swift` commands which suppresses all but error output.

### Motivation:

Closes #4395. Executable targets often depend on the output of other executable targets. When an executable target is run with `swift`, there is currently no way to suppress the build output, which makes it more difficult to pass the output of one executable as the input of another. The new quiet flag suppresses all output except errors.

### Modifications:

- Added a `@Flag` to `LoggingOptions`
- Interpret that flag as a `Diagnostic.Severity` in `LoggingOptions.logLevel`.
- Added tests for both the short and long flag.

### Result:

When a user passes the quiet flag, only errors will be logged.

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -138,6 +138,10 @@ public struct LoggingOptions: ParsableArguments {
     /// The verbosity of informational output.
     @Flag(name: [.long, .customLong("vv")], help: "Increase verbosity to include debug output")
     public var veryVerbose: Bool = false
+
+    /// Whether logging output should be limited to error.
+    @Flag(name: .shortAndLong, help: "Decrease verbosity to only include error output.")
+    public var quiet: Bool = false
 }
 
 public struct SecurityOptions: ParsableArguments {
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -139,7 +139,7 @@ public struct LoggingOptions: ParsableArguments {
     @Flag(name: [.long, .customLong("vv")], help: "Increase verbosity to include debug output")
     public var veryVerbose: Bool = false
 
-    /// Whether logging output should be limited to error.
+    /// Whether logging output should be limited to `.error`.
     @Flag(name: .shortAndLong, help: "Decrease verbosity to only include error output.")
     public var quiet: Bool = false
 }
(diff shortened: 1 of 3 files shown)


clayellis:
@tomerd I followed [your suggestions](https://github.com/apple/swift-package-manager/issues/4395#issuecomment-1361633100) in making these changes. I decided not to introduce the `log-level` option and left it for someone else as it may be useful to express the existing flags in terms of the `log-level` option, which would be a larger change.

Sources/CoreCommands/Options.swift
```diff
@@ -138,6 +138,10 @@ public struct LoggingOptions: ParsableArguments {
     /// The verbosity of informational output.
     @Flag(name: [.long, .customLong("vv")], help: "Increase verbosity to include debug output")
     public var veryVerbose: Bool = false
+
+    /// Whether logging output should be limited to error.
```
MaxDesiatov, line 142: nit: wording here is weird, unless it's meant to refer to an enum case, or however the corresponding logging level is declared.
```suggestion
    /// Whether logging output should be limited to `.error`.
```

MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

tomerd:
thanks @clayellis

neonichu (approved):

Issue #4806, opened 2018-05-31
[SR-7825] SwiftPM should consider a target with header files but no sources as a ClangTarget


aciidb0mb3r:
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-7825      |
   |Radar             | rdar://problem/34412874         |
   |Original Reporter | @aciidb0mb3r      |
   |Type              | Bug    |

   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 0         |
   |Component/s       | Package Manager    |
   |Labels            | Bug, StarterBug        |
   |Assignee          | None      |
   |Priority          | Medium      |

   

   md5: 8153b914e535cdeee67774fc66e8763e

  </details>

**Issue Description:**

Currently, swiftpm will complain that there are no source files in the target.

k-kohey:
@aciidb0mb3r 
I make a draft https://github.com/apple/swift-package-manager/pull/6006
If no one has worked on this issue, could you review the changes?

PR #6006, opened 2022-12-28
Target that does not contain source but only headers is made into a CLangTarget.

k-kohey:
This PR change allows the generation of targets containing only header files.

### Motivation:

Closes https://github.com/apple/swift-package-manager/issues/4806 .
Currently, if we attempt to create a target that does not contain source files, the Swift Package Manager returns an error and prompts us to put the source files. Therefore, it is not possible to create a CLangTarget that only contains header files.

### Modifications:

In addition to the absence of a source file, Target generation now fails when a header file does not exist.

### Result:

With the changes in this PR, a target that does not contain a source file but does contain headers is a CLangTarget.

### DIFF
--- a/Sources/PackageLoading/PackageBuilder.swift
+++ b/Sources/PackageLoading/PackageBuilder.swift
@@ -819,7 +819,7 @@ public final class PackageBuilder {
         // The name of the bundle, if one is being generated.
         let potentialBundleName = self.manifest.displayName + "_" + potentialModule.name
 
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci please smoke test

tomerd:
looks good to me. @neonichu @abertelrud ?

abertelrud (approved):

tomerd:
thanks @k-kohey

PR #6007, opened 2022-12-29
build: bring CMake build up to date

compnerd:
Add CMake based build support for the missing components as the Windows build depends on this still.  This allows us to work towards enabling the swift package collection and registry functionality.


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:

compnerd:

compnerd:

tomerd:
Thanks @compnerd I believe @neonichu was working to reduce the amount of modules that are built with CMake as part of the effort to do a minimal CMake bootstrap build, then build the rest with SwiftPM so that not all dependencies need to be included in the CMake build. This PR seems to bring back some of those dependencies. 

I know we ran into some issues making the minimal bootstrap work on Windows, but its seems like maybe better investment of our effort is to make that work, so we can scale out dependencies better. Should we rekindle that discussion / effort?

compnerd:
Sure, I think that we should do that, but in the meantime, we should get the package registry and package collections bits enabled for Windows so that we can keep parity for 5.8 (alternatively, we could remove that support from the other targets if you prefer).

tomerd:
I think the challange is that the package registry is about to get more dependencies that would also need to get CMake support so we are trying to avoid spending the time on all that, if we can spend the time on the "proper" fix. would be good to evaluate how much effort / how far we are from the "proper" fix so we can decide which is more pragmatic to do

compnerd:
@tomerd @neonichu - I believe that the last conversation we had about this was that we are far enough away that we should bring this into the tree now and then deal with removing it when we can.  Can you approve this PR (once done, I can resolve the conflicts).

tomerd:
@compnerd I believe we has agreed to bring it into 5.8 but not main is that we address this more deeply by doing the minimal-bootstrap on windows. I would much rather see the energy go there. if we cannot do this int time, we can pull this into 5.9 too but I rather not pull it into main.

compnerd:
@neonichu and I spoke about this, it turns out that we are significantly far away from what we need to do in SPM to enable the building of the artifacts on Windows (likely will miss 5.9).  But that means that in the mean time, main misses functionality, which means we cannot test/use that

tomerd:
@compnerd since we did not resolve the issue on main yet, should we create a version of this for 5.9?

compnerd:
@tomerd - yes, I believe so.  I'll try to get to this in the next day or two.

compnerd:
@tomerd I think that we might've missed this for 5.9?  And now we need to do it again for 5.10.  Can we please just merge this into main because it keeps getting dropped and there is functionality that is being dropped on Windows.  Alternatively, can we drop the functionality on macOS and Linux?

compnerd:
Mostly replaced with #8834, I think that there is a bit here that is left that I would like to get updated later.

PR #6010, opened 2023-01-01
Add support for writing the tests output in JSON format

NSCoder:
Add support for writing the result of the tests to a json file, similar to the XUnit feature

### Motivation:

I would like to be able to parse the result of the tests in other commands/programs without to have an elaborated regular expression parsing. For my needs specifically, I would love to be able to read the results using elisp and present them in a nice buffer with options such as retry.

### Modifications:

Added a very similar implementation as how XUnit file generation is implemented. I didn't want to do additional modifications in order to keep the changes as minimum as possible but in the future we could abstract this functionality to its own struct/file/module and generalize it to support mode encodings.

One particularity is that I didn't used codable for the implementation given that I noticed that we have our own custom JSON support. I imagined that this has a bigger reason (maybe not wanting to `import foundation`), so I decided to follow suit.

That said, if I were able to made additional changes I would have loved to implement a `--output` option so we could simplify the call site read like: `swift test --parallel -output json | jq"

### Result:

The `swift test` command has a new option `--json-output` that receives an absolute path where the result of the tests will be written encoded in JSON format.

I'm new to the source code and the repository and I'm very flexible to move things around, or change the implementation details. Please let me know! I would love to be able to merge this at some point.

### DIFF
--- a/Sources/Commands/SwiftTestTool.swift
+++ b/Sources/Commands/SwiftTestTool.swift
@@ -111,6 +111,10 @@ struct TestToolOptions: ParsableArguments {
             help: "Path where the xUnit xml file should be generated.")
     var xUnitOutput: AbsolutePath?
 
+    @Option(name: .customLong("json-output"),
+            help: "Path where the json output file should be generated.")
+    var jsonOutput: AbsolutePath?
+
     /// Generate LinuxMain entries and exit.
     @Flag(name: .customLong("testable-imports"), inversion: .prefixedEnableDisable, help: "Enable or disable testable imports. Enabled by default.")
     var enableTestableImports: Bool = true
@@ -305,6 +309,15 @@ public struct SwiftTestTool: SwiftCommand {
                     results: testResults
                 )
                 try generator.generate(at: xUnitOutput)
+            }
+
+            // Generate json file if requested
+            if let jsonOutput = options.jsonOutput {
+                let generator = JSONGenerator(
+                    fileSystem: swiftTool.fileSystem,
+                    results: testResults
+                )
+                try generator.generate(at: jsonOutput)
             }
 
             // process code Coverage if request
@@ -578,6 +591,16 @@ struct UnitTest {
     }
 }
 
+extension UnitTest: JSONSerializable {
+    func toJSON() -> JSON {
+        JSON([
+            "productPath": productPath,
+            "name": name,
+            "testCase": testCase
+        ])
+    }
+}
+
 /// A class to run tests on a XCTest binary.
 ///
 /// Note: Executes the XCTest with inherited environment as it is convenient to pass senstive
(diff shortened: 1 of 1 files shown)


Sources/Commands/SwiftTestTool.swift
```diff
@@ -111,6 +111,10 @@ struct TestToolOptions: ParsableArguments {
             help: "Path where the xUnit xml file should be generated.")
     var xUnitOutput: AbsolutePath?
 
+    @Option(name: .customLong("json-output"),
+            help: "Path where the json output file should be generated.")
+    var jsonOutput: AbsolutePath?
```
tomerd, line 116: this flag should be mutually exclusive with with xUnitOutput?

Sources/Commands/SwiftTestTool.swift
```diff
@@ -690,14 +713,34 @@ final class TestRunner {
...
+        }
+        return nanoseconds.toJSON()
+    }
+}
+
 /// A class to run tests in parallel.
 final class ParallelTestRunner {
     /// An enum representing result of a unit test execution.
-    struct TestResult {
+    struct TestResult: JSONSerializable {
```
tomerd, line 730: ideally all the JSONSerializable conformances (or better, Encodable) is done in an extension

MaxDesiatov: +1 to this, I'm also not sure why `Encodable` is not used here instead of adding a new dependency on TSC

tomerd:
thanks @NSCoder this is a good addition. couple of ideas / comments:

1. JSONSerializable is an older convention before Foundation was stable. At this point we should not use any more, so lets create little Encodable helpers here
2. Is there a some standard for test results in the JSON format, ie something like XUnit? if there was one, it would probably  be better than to roll our own format

NSCoder:
Thank you for taking a look @tomerd! The feedback makes sense, I will change the JSON Encoding to use codable. About the json output format. I did some quick research and it seems that there is not a JSON based standard, perhaps what we could do is just return the same XUnit structure but in JSON:

```json
{
  "name": "SwiftTest",
  "tests": 2,
  "failures": 1,
  "time": 2,
  "testsuite": [
    {
      "name": "UnitTests",
      "errors": 0,
      "failures": 0,
      "skipped": 0,
      "timestamp": "<A Timestamp>",
      "time": 1,
      "tests": 1,
      "testcase": [
        {
          "classname": "SimpleTest",
          "name": "testASimpleFunction",
          "time": 0.1
        }
      ]
    },
    {
      "name": "FailingUnitTests",
      "errors": 0,
      "failures": 1,
      "skipped": 0,
      "timestamp": "<A Timestamp>",
      "time": 1,
      "tests": 1,
      "testcase": [
        {
          "classname": "FailingTests",
          "name": "testAFailingFunction",
          "time": 0.1,
          "failure": [
            {
              "XCAssert": "1 is not equal to 2"
            }
          ]
        }
      ]
    }
  ]
}
```

(reference: https://github.com/Kesin11/ts-junit2json)

I'm not sure that we have access to this level of detail in the testing data to fully populate this format, but in the meanwhile we could return what we have.

What do you think?

Sources/Commands/SwiftTestTool.swift
NSCoder: Good questions, I don't see a reason why they have to be mutually exclusive besides modifying the options from the command.

@MaxDesiatov I was thinking about this as well, but giving that I'm not familiar with the project I decided to keep this patch it as simple as possible. Would it be alright if we try to add the output option refactor in a following up patch?

Sources/Commands/SwiftTestTool.swift
NSCoder: Sure thing, is now updated to use `Encodable`

tomerd:
@NSCoder using an XUnit JSON representation sounds very good to me.

Sources/Commands/SwiftTestTool.swift
```diff
@@ -578,6 +592,8 @@ struct UnitTest {
     }
 }
 
+extension UnitTest: Encodable {}
```
tomerd, line 595: if we go for an XUnit like format, we should probably create utility structs which will be the ones that are encoded. there will be a small cost of crating a second representation in memory, but I think it is worth it since it will give us a degree of freedom between the "real" model and the one we want to write to disk as JSON. its a technique I often use when build server APIs so I know it works well.

tomerd: fwiw, we can then also modify the XUnit generator to use these XUnit specific representations which will make that consistent across the two on-disk representations

Sources/Commands/SwiftTestTool.swift
tomerd: yes, I think it is fine for now

PR #6011, opened 2023-01-02
Basics: extract `darwinCacheDirectories` from `TSCUtility`

compnerd:
This function is used solely in SPM and is the last portion of `TSCUtility.Platform` in use.  Move the implementation into the single use site to allow us to deprecate the interface in swift-tools-support-core.

### DIFF
--- a/Sources/Basics/Sandbox.swift
+++ b/Sources/Basics/Sandbox.swift
@@ -13,10 +13,7 @@
 import Foundation
 import TSCBasic
 
-import enum TSCUtility.Platform
-
 public enum Sandbox {
-
     /// Applies a sandbox invocation to the given command line (if the platform supports it),
     /// and returns the modified command line. On platforms that don't support sandboxing, the
     /// command line is returned unmodified.
@@ -55,6 +52,29 @@ public enum Sandbox {
 // MARK: - macOS
 
 #if os(macOS)
+fileprivate let threadSafeDarwinCacheDirectories: [AbsolutePath] = {
+    func GetConfStr(_ name: CInt) -> AbsolutePath? {
+        let length: Int = confstr(name, nil, 0)
+
+        let buffer: UnsafeMutableBufferPointer<CChar> = .allocate(capacity: length)
+        defer { buffer.deallocate() }
+
+        guard confstr(name, buffer.baseAddress, length) == length else { return nil }
+
+        let value: String = String(cString: buffer.baseAddress!)
+        guard value.hasSuffix("/") else { return nil }
+
+        return try? resolveSymlinks(AbsolutePath(validating: value))
+    }
+
+    var directories: [AbsolutePath] = []
+    try? directories.append(AbsolutePath(validating: "/private/var/tmp"))
+    (try? TSCBasic.determineTempDirectory()).map { directories.append($0) }
+    GetConfStr(_CS_DARWIN_USER_TEMP_DIR).map { directories.append($0) }
+    GetConfStr(_CS_DARWIN_USER_CACHE_DIR).map { directories.append($0) }
+    return directories
+}()
+
 fileprivate func macOSSandboxProfile(
     strictness: Sandbox.Strictness,
     writableDirectories: [AbsolutePath],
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

MaxDesiatov (approved):
This is great, thanks! 👍

Sources/Basics/Sandbox.swift
```diff
@@ -55,6 +53,29 @@ public enum Sandbox {
 // MARK: - macOS
 
 #if os(macOS)
+fileprivate let threadSafeDarwinCacheDirectories: ThreadSafeArrayStore<AbsolutePath> = {
```
tomerd, line 56: the use of ThreadSafeArrayStore here was for caching and we loose the cache with this change. that may be okay, but then we dont need ThreadSafeArrayStore

@abertelrud wdyt about loosing the cache?

tomerd (changes requested):
we should address the usage of ThreadSafeArrayStore / loss of cache

Sources/Basics/Sandbox.swift
compnerd: We do retain the caching in some aspect - the value will be computed once and re-used rather than computed each time.  The idea of replacing the `ThreadSafeArrayStore<AbsolutePath>` with `[AbsolutePath]` however still stands and would also simplify this further.  I think that need to wait for @abertelrud to chime in on that.

abertelrud: I do think it's a good idea to compute this only once, but it does seem that the change looses the thread safety aspect.  Couldn't it be implemented the same way as before where the contents of ThreadSafeArrayStore are what get cached, rather than possibly having two separate ThreadSafeArrayStores get created racily?  i.e. keep the  property as static though no longer on an extension of TSCUtility.Platform.

abertelrud: Actually, are the initializers of Swift statics atomic already?  If so, do we need `ThreadSafeArrayStore` at all here?

abertelrud: I see — sorry, I read it as being inside the `public enum Sandbox {}` so a property of that enum.  In this case, it seems to me that the `ThreadSafeArrayStore` is unnecessary and we still get the compute-only-once behaviour.  @tomerd am I misunderstanding anything?  IIUC you said that this gets rid of the cache, but it doesn't sound as if it does.

abertelrud: Based on this https://docs.swift.org/swift-book/LanguageGuide/Properties.html#ID264 it seems that the initialization is inherently atomic.

tomerd: yes I agree - ThreadSafeArrayStore is unnecessary here

tomerd: also please run with TSAN to make sure thread safety is okay (which is probably why we used ThreadSafeArrayStore in the first place

abertelrud: A static property of the `Sandbox` type makes sense to me.

compnerd: I'm not sure that moving it into `Sandbox` makes it any more encapsulated, but does namespace it.  I don't see the value, but I can move it nonetheless.  I'll remove the now unnecessary use of `ThreadSafeArrayStore`.

abertelrud: I have no strong opinion about whether it's a fileprivate global or a static inside Sandbox, so whatever @tomerd thinks is fine with me.

compnerd: Actually, that is even more odd - the value is used in a single free-standing function only.  I'm leaving it currently as a global until @tomerd responds.

tomerd (approved):
thanks

tomerd:
@swift-ci smoke test

Sources/Basics/Sandbox.swift
tomerd: I prefer it inside Sandbox, but won't hold this back

PR #6013, opened 2023-01-03
[5.8] There's no need to pretty-print the JSON that a compiled manifest sent back to SwiftPM (#5995)

abertelrud:
This is a 5.8 nomination of https://github.com/apple/swift-package-manager/pull/5995.  The change is already in main.


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #6019, opened 2023-01-06
Pass through the package identity as a parameter to `willLoadManifest`/`didLoadManifest` callbacks

abertelrud:
Adds a new package identity parameter to the callbacks for loading manifests.

### Motivation:

Without this, implementors of these delegate calls that want the package identity have to re-derive it from the URL, which is lossy.  Instead the callbacks should pass through the package identity it already knows about.

### Modifications:

- add a package identity parameter to each of the will/did callbacks for the manifest

### Result:

- the client code has the original package identity


abertelrud:
@swift-ci please smoke test

tomerd (approved):

PR #6020, opened 2023-01-06
Pass along the duration in the remaining WorkspaceDelegate calls as well

abertelrud:
### Motivation:

The wall clock duration is already passed in didFetch, didUpdate, etc.  For completeness, Workspace should pass along the duration in the didLoadManifest/didCreateWorkingCopy/didCheckOut calls as well, so that libSwiftPM clients that want to use this information has access to it.

### Modifications:

- capture the start and end times, and pass them in the delegate callback

### Result:

- clients of these delegate calls have access to the duration for all of the will/did delegate call pairs

### Note:

This builds on top of https://github.com/apple/swift-package-manager/pull/6019 (because the changes touch the same lines), so only the newer of the two commits is specific to this PR.

### DIFF
--- a/Sources/SourceControl/RepositoryManager.swift
+++ b/Sources/SourceControl/RepositoryManager.swift
@@ -206,10 +206,10 @@ public class RepositoryManager: Cancellable {
                 return handle
             }
             // Update the repository when it is being looked up.
-            let start = DispatchTime.now()
             delegateQueue.async {
                 self.delegate?.willUpdate(package: package, repository: handle.repository)
             }
+            let start = DispatchTime.now()
             let repository = try handle.open()
             try repository.fetch()
             let duration = start.distance(to: .now())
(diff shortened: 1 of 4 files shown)


Sources/SourceControl/RepositoryManager.swift
```diff
@@ -206,10 +206,10 @@ public class RepositoryManager: Cancellable {
                 return handle
             }
             // Update the repository when it is being looked up.
-            let start = DispatchTime.now()
```
abertelrud, line 209: Moving this down to below the delegate call so that its duration isn't counted as part of it.  This is consistent with the other delegate method duration capturing.

abertelrud:
@swift-ci please smoke test

tomerd (approved):

abertelrud:
@swift-ci please smoke test

PR #6021, opened 2023-01-06
[NFC] PubGrub: move a few types to separate files

MaxDesiatov:
### Motivation:

`PubGrubDependencyResolver.swift` reached almost 1.5k lines, making it hard to navigate in related areas of the codebase.

### Modifications:

Moving a few types from that file to separate files. Additionally, `PubGrub` capitalization was made consistent in the naming.

### Result:

`PubGrubDependencyResolver.swift` is smaller and more readable.


MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci smoke test

tomerd (approved):
let's also use this opportunity to run swift format on the changed files

MaxDesiatov:
@tomerd could you clarify what exact formatting tool we should use? Do we rely on SwiftFormat and [this `.swfitformat` file](https://github.com/apple/swift-package-manager/blob/main/.swiftformat) for configuration, or [this formatting tool](https://github.com/apple/swift-package-manager/blob/main/Sources/Commands/PackageTools/Format.swift) for using swift-format?

tomerd:
yes: "SwiftFormat and [this .swfitformat file](https://github.com/apple/swift-package-manager/blob/main/.swiftformat) for configuration" mark sure to run it on modified only to minimize changes

MaxDesiatov:
> we could add this to the soundness script (@shahmishal this is the script we should run in CI) as we do in the server libraries, eg: https://github.com/apple/swift-log/blob/main/scripts/soundness.sh#L66 but we should make a version that checks only the modified files

Can we assume that SwiftFormat is installed on CI bots, or should we install if it's absent as a part of `soundness.sh`?

MaxDesiatov:
@swift-ci smoke test

PR #6033, opened 2023-01-09
Basics: don't point users to bugs.swift.org with `InternalError`

MaxDesiatov:
Since https://bugs.swift.org redirects to https://github.com/apple/swift/issues/, we'd like to point users to the SwiftPM repository issues instead.


MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

PR #6035, opened 2023-01-10
BuildPlan: provide user-actionable error description for `.noBuildableTarget`

MaxDesiatov:
### Motivation:

Currently, the error message `BuildPlan.Error.noBuildableTarget` doesn't specify any actions that a user could take to resolve the issue.

### Modifications:

Expanded the string literal returned from `description` of this error type to include a suggestion for users to take to resolve it.

### Result:

Users who stumble upon the error are less confused by the error message.


MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

PR #6036, opened 2023-01-10
PackagePlugin: fix typo in `PackageManagerProxyError` case

MaxDesiatov:
`unimlemented` -> `unimplemented`

Since it's a case in a `public enum` of `PackagePlugin` module, which plugins are supposed to import, this is technically a source-breaking change. OTOH, do we expect that anyone had a `switch` for every case of this `enum`? This error case isn't thrown from anywhere.

### DIFF
--- a/Sources/PackagePlugin/PackageManagerProxy.swift
+++ b/Sources/PackagePlugin/PackageManagerProxy.swift
@@ -267,7 +267,7 @@ fileprivate extension PackageManager {
 
 public enum PackageManagerProxyError: Error {
     /// Indicates that the functionality isn't implemented in the plugin host.
-    case unimlemented(_ message: String)
+    case unimplemented(_ message: String)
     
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci smoke test macOS

tomerd (approved):

abertelrud (approved):
Thanks Max.  I think the risk of this breaking anything is very low indeed.

MaxDesiatov:
Do you think this should be cherry-picked for 5.8?

MaxDesiatov:
I think it's worth landing this in 5.8 so that we reduce the chance of more people using this enum case (whether on purpose or by accident), PR submitted here https://github.com/apple/swift-package-manager/pull/6039

PR #6039, opened 2023-01-11
[5.8] PackagePlugin: fix typo in `PackageManagerProxyError` case

MaxDesiatov:
`unimlemented` -> `unimplemented`

Since it's a case in a `public enum` of `PackagePlugin` module, which plugins are supposed to import, this is technically a source-breaking change. OTOH, do we expect that anyone had a `switch` for every case of this `enum`? This error case isn't thrown from anywhere.


MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

Issue #4758, opened 2018-09-09
[SR-8720] [SwiftPM] Add --hide-build option to `swift run`


ddunbar:
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-8720      |
   |Radar             | rdar://problem/44307764         |
   |Original Reporter | @ddunbar      |
   |Type              | Bug    |

   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 4         |
   |Component/s       | Package Manager    |
   |Labels            | Bug, StarterBug        |
   |Assignee          | None      |
   |Priority          | Medium      |

   

   md5: 242446b5c75fedfcae01e6e6a2ccfd96

  </details>

**Issue Description:**

It is often useful to use \`swift run\` as a replacement for building and running.

 

However, for many tools which when run produce output, it is cumbersome to see the build log output. Perhaps \`swift run\` should have an option like \`--hide-build\` which hides all of the build output ****unless**** it fails. This makes it very convenient to use \`swift run\` as a replacement in more situations.

PR #6040, opened 2023-01-11
Add hide-build flag to swift-run

k-kohey:
A flag hide-build has been added to the swift-run command to hide the output of the build.

### Motivation:

closes https://github.com/apple/swift-package-manager/issues/4758

If the build log is mixed with the log output at runtime, it becomes cumbersome to see the log output at runtime.

### Modifications:

Add the hide-build flag and switch the standard error output to a different output when this flag is true.

### Result:

Logs that used to be sent to standard error output at build time will no longer be displayed.


tomerd:
thanks @k-kohey does https://github.com/apple/swift-package-manager/pull/5988 solve the issue?

k-kohey:
Thanks for letting me know.
I will close this PR as it has been resolved.

PR #6057, opened 2023-01-18
Pass WorkspaceConfiguration from convenience constructor

giginet:
### Motivation:

A convenience constructor of `Workspace` received some arguments. However, they are ignored. 
It may be a simple mistake.

### Modifications:

I passed some arguments to the default constructor from arguments.

### Result:


MaxDesiatov:
@swift-ci smoke test

MaxDesiatov (approved):
Seems legit, thanks!

neonichu (approved):
Thanks!

PR #6064, opened 2023-01-19
[5.8] Enable use of product ID to distinguish duplicate product names

elsh:
Enable use of product ID to distinguish duplicate product names
Resolves rdar://102138022


elsh:
@swift-ci smoke test

tomerd (approved):

PR #6065, opened 2023-01-19
Warn instead of error when non-Swift sources may be included in a target with module aliases

elsh:
Warn instead of error when non-Swift sources may be included in a target that has module aliases. This helps when a Swift target being aliased in a package is in a dependency graph of another package which contains non-Swift sources.
Resolves rdar://102137567


elsh:
@swift-ci smoke test

neonichu (approved):

PR #6075, opened 2023-01-24
[NFC] Basics: add `SendableBox` and `TokenBucket` helpers

MaxDesiatov:
### Motivation:

When working with Swift Concurrency, Dispatch-based primitives such as locks and semaphores are not suitable. We should have `async`-friendly primitives that replace those.

### Modifications:

New `SendableBox` and `TokenBucket` actor types were added. I also took this as an opportunity to split `ConcurrencyHelpers.swift` into smaller files and move relevant files into the new `Basics/Concurrency` directory.

### Result:

`SendableBox` and `TokenBucket` are available for implementation and testing of features such as the new `async`-friendly `HTTPClient`, which is coming in a subsequent PR.

Depends on https://github.com/apple/swift-package-manager/pull/6058 and is a requirement of https://github.com/apple/swift-package-manager/pull/6022 for PR stacking purposes.


Sources/Basics/Concurrency/AsyncBox.swift
```diff
@@ -0,0 +1,46 @@
...
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=5.5.2)
+
+import struct Foundation.Date
+
+/// A `Sendable` storage that allows access from concurrently running tasks in an `async` closure.
+public actor AsyncBox<Value: Sendable> {
```
tomerd, line 18: curious, why is this even needed?

Sources/Basics/Concurrency/NSLock+Extensions.swift
```diff
@@ -12,9 +12,9 @@
 
 import class Foundation.NSLock
 
-extension NSLock {
+public extension NSLock {
     /// Execute the given block while holding the lock.
-    public func withLock<T> (_ body: () throws -> T) rethrows -> T {
+    func withLock<T>(_ body: () throws -> T) rethrows -> T {
```
tomerd, line 17: we prefer the visibility modifier on the functions not the extension

Sources/Basics/Concurrency/TokenBucket.swift
```diff
@@ -0,0 +1,57 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=5.5.2)
+
+import DequeModule
+
+/// Type modeled after a "token bucket" pattern, which is similar to a semaphore, but is built with
+/// Swift Concurrency primitives.
+actor TokenBucket {
```
tomerd, line 19: does this need to public so it can be used by other modules?

tomerd (approved):

Sources/Basics/Concurrency/AsyncBox.swift
MaxDesiatov: In `HTTPClient` tests we need a counter for checking how many times a closure was invoked and we need to assert on that counter after the closure has run. Previously this was done with a lock. Now with Swift Concurrency `async` closures can't capture an arbitrary variably, but they can capture references to actors.

Sources/Basics/Concurrency/TokenBucket.swift
MaxDesiatov: Thanks, that's addressed now

neonichu (approved):

Sources/Basics/Concurrency/AsyncBox.swift
MaxDesiatov: Ah, I remember the details now. It's a direct replacement for `ThreadSafeBox`.

tomerd: okay - maybe `SendableBox` then?

MaxDesiatov:
For some reason GitHub automatically closed this PR when I merged the base branch, instead of retargeting it against `main` like it used to do. I also can't reopen it and I'm not sure why. I've re-created it as https://github.com/apple/swift-package-manager/pull/6085, but needs review approvals again unfortunately.

PR #6080, opened 2023-01-27
Fix PluginInvocationTests.testCompilationDiagnostics() to avoid throwing if no diagnostics at all are reported

abertelrud:
This is a follow-on to https://github.com/apple/swift-package-manager/pull/6076, which still contained a `XCTUnwrap()` that failed if there were no diagnostics.  See https://smooshbase.apple.com/swiftci/job/swift-pr-smoke-test/1860/console.


abertelrud:
@swift-ci please smoke test

tomerd (approved):

Issue #6081, opened 2023-01-27
`swift-test` cannot determine if output writes to a terminal


grynspan:
### Description

It's common to change the format of output (notably, adding ANSI escape codes) when `stdout`/`stderr` is a TTY/terminal. SPM uses pipes to route output from a plugin back to the main `swift` process and its `stdout`, which prevents the use of `isatty()` etc. to determine the format of output.

swift-test could, in particular, leverage this information to make failed tests more visible (red, bold, etc.)

SPM could dodge the issue by setting an environment variable on spawned test processes when it determines its own output is a TTY, and then plugins could be modified to check that variable. For example:

```swift
  var environment: [String: String] = ...
#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS) || os(Linux)
  if nil != getenv("TERM") {
    if 0 != isatty(STDOUT_FILENO) {
      environment["SPM_STDOUT_IS_TTY"] = "1"
    }
    if 0 != isatty(STDERR_FILENO) {
      environment["SPM_STDERR_IS_TTY"] = "1"
    }
  }
#elseif os(Windows)
  if let stdoutHandle = GetStdHandle(STD_OUTPUT_HANDLE) {
    var screenBufferInfo = CONSOLE_SCREEN_BUFFER_INFO()
    if GetConsoleScreenBufferInfo(stderrHandle, &screenBufferInfo) {
      environment["SPM_STDOUT_IS_TTY"] = "1"
    }
  }
  if let stderrHandle = GetStdHandle(STD_ERROR_HANDLE) {
    var screenBufferInfo = CONSOLE_SCREEN_BUFFER_INFO()
    if GetConsoleScreenBufferInfo(stderrHandle, &screenBufferInfo) {
      environment["SPM_STDERR_IS_TTY"] = "1"
    }
  }
#endif
```

### Expected behavior

I'd expect the typical behaviour for `isatty(STDOUT_FILENO)` and `isatty(STDERR_FILENO)` (and the Windows equivalents.)

### Actual behavior

These streams are pipes on the swift-test side.

### Steps to reproduce

N/A

### Swift Package Manager version/commit hash

Swift Package Manager - Swift 5.8.0-dev

### Swift & OS version (output of `swift --version && uname -a`)

swift-driver version: 1.73 Apple Swift version 5.8 (swiftlang-5.8.0.113.42 clang-1403.0.19.8)
Target: arm64-apple-macosx13.0
macOS Ventura

[...]

PR #6093, opened 2023-01-31
`swift-test` cannot determine if output writes to a terminal (#6081)

grynspan:
This change sets an environment variable before spawning swift-test if the calling process appears to support ANSI control codes in its output.

### Motivation:

XCTest does not currently consume this information, but having it will allow for improvements to CLI output during testing.

### Modifications:

This change sets an environment variable before spawning swift-test if the calling process appears to support ANSI control codes in its output. This environment variable is then propagated to swift-test/xctest.

### Result:

With this change, the environment variable `XCTEST_COLOR_DIAGNOSTICS_ENABLED` should be set to `1` in test processes if the calling process (`swift`) supports ANSI control codes (i.e. is a TTY.)

### DIFF
--- a/Sources/Commands/Utilities/TestingSupport.swift
+++ b/Sources/Commands/Utilities/TestingSupport.swift
@@ -116,6 +116,14 @@ enum TestingSupport {
     ) throws -> EnvironmentVariables {
         var env = EnvironmentVariables.process()
 
+        // If the standard output or error stream is NOT a TTY, set the NO_COLOR
+        // environment variable. This environment variable is a de facto
+        // standard used to inform downstream processes not to add ANSI escape
+        // codes to their output. SEE: https://www.no-color.org
+        if !stdoutStream.isTTY || !stderrStream.isTTY {
+            env["NO_COLOR"] = "1"
+        }
+
         // Add the code coverage related variables.
         if buildParameters.enableCodeCoverage {
             // Defines the path at which the profraw files will be written on test execution.


grynspan:
@swift-ci please test

grynspan:
@swift-ci please smoke test

neonichu (approved):
I suspect we want some more checks in here, e.g. potentially not set the env var if `--xunit-output` is being passed, but this seems like a reasonable place to start.

grynspan:
I'm not sure how to write a test here that can verify it. Any suggestions?

grynspan:
> I suspect we want some more checks in here, e.g. potentially not set the env var if `--xunit-output` is being passed, but this seems like a reasonable place to start.

At the point where I add the variable, I don't think I have access to that flag anymore, but let me do some digging and see.

grynspan:

neonichu (approved):

tomerd (changes requested):
I would rather find a more idiomatic way than a custom environment variable. what other options do we have?

grynspan:
> I would rather find a more idiomatic way than a custom environment variable. what other options do we have?

Discussing with @tomerd internally.

grynspan:
Discussed with @tomerd and he agreed that setting `NO_COLOR` would be a reasonable approach, as it's a _de facto_ standard.

neonichu (approved):

tomerd (approved):

PR #6095, opened 2023-01-31
PackageRegistryTool: repair the Windows build

compnerd:
`getpass` is an obselete function as per POSIX and is unavailable on windows.  Replicate the behaviour with `SetConsoleMode` and `ReadConsole`.  We assume that we can use the standard definition of `PASS_MAX` (255) for the value.  The usage requires that we replicate the behaviour of a leaked global buffer.


compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

neonichu (approved):

yim-lee (approved):

PR #6106, opened 2023-02-02
IntegrationTests: fix `AbsolutePath` deprecation warnings

MaxDesiatov:
For some reason these warnings were printed only in certain conditions, which may explain why these calls to a deprecated initializer of `AbsolutePath` persisted in the codebase for so long.


MaxDesiatov:
@swift-ci smoke test

neonichu (approved):

PR #6108, opened 2023-02-02
[5.8] Fix manifest source generation for `enableUpcomingFeature`/`enableExp…

cwakamo:
…erimentalFeature`.

Manifest source generation was previously transforming:

    .enableUpcomingFeature("FeatureName")

to:

    .upcomingFeatures(["FeatureName"])

because of how the build setting was stored internally.

This commit fixes the issue by switching the internal model for these build settings to match the external model so they can be properly regenerated. This also matches how other build settings (like `linkedFramework`) are modeled internally.

This addresses <rdar://problem/104718971>.

(cherry picked from commit 366e79b8f0a40252bb7a01e3ca8c0c35f2647450)


cwakamo:
@swift-ci smoke test

tomerd (approved):

PR #6117, opened 2023-02-03
[5.8] PackageModel: partially address #5719

compnerd:
Handle `-use-ld=` from the configuration in the selection of the librarian.  We do not honour the `-Xmanifest` or `-Xswiftc` flags being passed on the command line rather than the plist as we cannot pre-compute the flags.

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -113,11 +113,15 @@ public final class UserToolchain: Toolchain {
 
     // MARK: - public API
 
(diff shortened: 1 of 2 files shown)


compnerd:
@tomerd - this resolves a regression from 5.7 where you cannot use SPM outside of VsDevEnv.

compnerd:
@swift-ci please smoke test

tomerd:
okay to merge if this is only impacting windows. @compnerd please confirm?

compnerd:
Yes, this is a windows only fix.

tomerd (approved):
impacts windows only

Issue #5771, opened 2022-09-20
Detect `link` tool automatically on Windows, and fallback to `lld-link` if not found


stevapple:
### Description

In Swift 5.7, we can use SwiftPM in regular CMD or PowerShell environment, but after #5720 we're getting `error: toolchain is invalid: could not find link`. This is not desirable, and we should both try to detect `link` and fallback to `lld-link` if `link` can't be found.

### Expected behavior

SwiftPM could work without special developer environment.

### Actual behavior

It just throws `error: toolchain is invalid: could not find link`.

### Steps to reproduce

1. Open a regular CMD or PowerShell window;
2. Change to the root of any Swift package;
3. Run `swift build`.

### Swift Package Manager version/commit hash

`e1e80d7024d0a2c62dee37841d242de7560cfd75`

### Swift & OS version (output of `swift --version && uname -a`)

swift.org Swift version 5.8-dev (LLVM efc29aad7a9faa8, Swift f926816e5105fca)
Target: x86_64-unknown-windows-msvc

Windows 10 Pro Workstation 21H1
Visual Studio 17.3.4

stevapple:
** Additional Context**

- Path to MSVC toolset is discoverable by the `vswhere` tool;
- We should also consider falling back to `llvm-ar` when `ar` is missing on Linux.

tomerd:
cc @compnerd I think you worked on this specific change?

compnerd:
I'm not convinced that this is the right thing to do in general.  If this was meant specifically for static libraries, then I would say that it makes sense.  However, in that case, this is basically duplicating #5719 and #5761.  I would be slightly against the idea of the `vswhere` dependency.  If the build tools are installed, I believe that `vswhere` is not going to be installed and additional dependencies on the setup sound like a poor choice given that the current dependency set is already difficult.  I'm not sure why swift-driver is being brought up here though.

stevapple:
> If the build tools are installed, I believe that `vswhere` is not going to be installed and additional dependencies on the setup sound like a poor choice given that the current dependency set is already difficult.

That’s exactly not the case — `vswhere` is guaranteed to exist on every machines that has Visual Studio installed, so we  don’t need to vendor it.

> However, in that case, this is basically duplicating #5719 and #5761.

[...]

PR #6120, opened 2023-02-05
Use `vswhere` to find MSVC tools

stevapple:
Use [`vswhere`](https://github.com/microsoft/vswhere) tool to find MSVC tools. (Partially resolves https://github.com/apple/swift-package-manager/issues/5771)

### Motivation:

After https://github.com/apple/swift-package-manager/issues/5771, SwiftPM gains implicit dependency of `link` on Windows, but it cannot detect it from the context without opts-in like Developer command prompt.

Although the installation guide never clarified that non-developer environment is supported, SwiftPM actually works since it's ported. Some automation workflows and tools like [Swift for Visual Studio Code](https://github.com/swift-server/vscode-swift) already depends on the behavior. Breaking the compatibility makes huge pain for the Swift on Windows ecosystem.

### Modifications:

- `UserToolchain.findTool` is taught to call `vswhere` on Windows and now knows about MSVC installation.
- `UserToolchain.determineLibrarian` and `UserToolchain.determineSwiftCompilers` gets new arguments to help determine the context.
- `\UserToolchain.useXcrun` is reused for `vswhere` because they're much of serving the same purpose.

### Result:

SwiftPM will be usable (again) from non-developer shells like Windows Powershell.

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -100,7 +100,7 @@ public final class UserToolchain: Toolchain {
     return toolPath
   }
 
-  private static func findTool(_ name: String, envSearchPaths: [AbsolutePath], useXcrun: Bool)
+  private static func findTool(_ name: String, destination: Destination, envSearchPaths: [AbsolutePath], useXcrun: Bool)
     throws -> AbsolutePath
   {
     if useXcrun {
@@ -120,17 +120,27 @@ public final class UserToolchain: Toolchain {
 
     if useXcrun {
       #if os(Windows)
-        if let programFiles = TSCBasic.ProcessEnv.vars["ProgramFiles(x86)"],
-           let visualStudio = try? TSCBasic.Process.checkNonZeroExit(arguments: [
-            "\(programFiles)\\Microsoft Visual Studio\\Installer\\vswhere.exe",
-            "-latest", "-products", "*",
-            "-requires", "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",
-            "-property", "installationPath"
-           ]).spm_chomp(),
-           let vcToolsVersion = try? String(contentsOfFile: "\(visualStudio)\\VC\\Auxiliary\\Build\\Microsoft.VCToolsVersion.default.txt"),
-           let vcToolsDir = try? AbsolutePath(validating: "\(visualStudio)\\VC\\Tools\\MSVC\\\(vcToolsVersion)\\bin\\HostX64\\x64"),
-           let toolPath = try? getTool(name, binDir: vcToolsDir) {
-          return toolPath
+        func vcArchNames(triple: Triple?) -> (product: String, host: String, target: String)? {
+          switch triple?.arch {
+          case .x86_64: return ("Microsoft.VisualStudio.Component.VC.Tools.x86.x64", "HostX64", "x64")
+          case .i686: return ("Microsoft.VisualStudio.Component.VC.Tools.x86.x64", "HostX86", "x86")
+          case .arm64, .arm64e, .aarch64: return("Microsoft.VisualStudio.Component.VC.Tools.ARM64", "HostARM64", "arm64")
+          case .arm, .armv5, .armv6, .armv7: return("Microsoft.VisualStudio.Component.VC.Tools.ARM", "HostARM", "arm")
+          default: return nil
+          }
+        }
+      if let (_, vcHost, _) = vcArchNames(triple: destination.hostTriple),
+         let (vcTools, _, vcTarget) = vcArchNames(triple: destination.targetTriple),
+         let programFiles = TSCBasic.ProcessEnv.vars["ProgramFiles(x86)"] {
+          if let visualStudio = try? TSCBasic.Process.checkNonZeroExit(arguments: [
+              "\(programFiles)\\Microsoft Visual Studio\\Installer\\vswhere.exe",
+              "-latest", "-products", "*", "-requires", vcTools, "-property", "installationPath"
+             ]).spm_chomp(),
+             let vcToolsVersion = try? String(contentsOfFile: "\(visualStudio)\\VC\\Auxiliary\\Build\\Microsoft.VCToolsVersion.default.txt"),
+             let vcToolsDir = try? AbsolutePath(validating: "\(visualStudio)\\VC\\Tools\\MSVC\\\(vcToolsVersion)\\bin\\\(vcHost)\\\(vcTarget)"),
+             let toolPath = try? getTool(name, binDir: vcToolsDir) {
+            return toolPath
+          }
         }
       #endif
     }
@@ -141,7 +151,7 @@ public final class UserToolchain: Toolchain {
   // MARK: - public API
 
   public static func determineLibrarian(
-    triple: Triple, binDir: AbsolutePath,
+    triple: Triple, destination: Destination, binDir: AbsolutePath,
     useXcrun: Bool,
     environment: EnvironmentVariables,
     searchPaths: [AbsolutePath]
@@ -181,12 +191,12 @@ public final class UserToolchain: Toolchain {
     if let librarian = try? UserToolchain.getTool(tool, binDir: binDir) {
       return librarian
     }
-    return try UserToolchain.findTool(tool, envSearchPaths: searchPaths, useXcrun: useXcrun)
+    return try UserToolchain.findTool(tool, destination: destination, envSearchPaths: searchPaths, useXcrun: useXcrun)
   }
 
   /// Determines the Swift compiler paths for compilation and manifest parsing.
   public static func determineSwiftCompilers(
-    binDir: AbsolutePath, useXcrun: Bool, environment: EnvironmentVariables,
+    binDir: AbsolutePath, destination: Destination, useXcrun: Bool, environment: EnvironmentVariables,
     searchPaths: [AbsolutePath]
   ) throws -> SwiftCompilers {
     func validateCompiler(at path: AbsolutePath?) throws {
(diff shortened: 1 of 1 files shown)


Sources/PackageModel/UserToolchain.swift
```diff
@@ -117,15 +117,55 @@ public final class UserToolchain: Toolchain {
         return toolPath
       }
     }
+
+    if useXcrun {
+      #if os(Windows)
```
stevapple, line 122: Is `#if os(Windows)` check enough, or do we need to gate on host triple too?

Sources/PackageModel/UserToolchain.swift
```diff
@@ -56,7 +56,7 @@ public final class UserToolchain: Toolchain {
   /// Search paths from the PATH environment variable.
   let envSearchPaths: [AbsolutePath]
 
-  /// Only use search paths, do not fall back to `xcrun`.
+  /// Only use search paths, do not fall back to `xcrun` or `vswhere`.
   let useXcrun: Bool
```
stevapple, line 60: It's still undecided if we want to rename the variable. I didn't do this at the point because neither have I got a short and clear name (`useExternalTool` sounds verbose and confusing), nor am I going to make `vswhere` the default like `xcrun` is now. It can be prioritized in later pitches.

Sources/PackageModel/UserToolchain.swift
```diff
@@ -117,15 +117,55 @@ public final class UserToolchain: Toolchain {
         return toolPath
       }
     }
+
+    if useXcrun {
```
stevapple, line 121: I would like to introduce this approach as a fall-back from the beginning, and then makes it the default (like Clang does) if everything is working as expected.

compnerd (changes requested):
I strongly feel that this is the wrong way to handle this.  If the install location changes this breaks. If I install outside of `ProgramFiles(x86)` this breaks.  This has a bigger issue where we now are divergent from clang - clang has its own logic for selecting the VSCode version, which we must match or run the risk in mixing pieces incorrectly.

At the very least, this should be using clang to locate the tools, not `vswhere`.

stevapple:
I believe you have **very deep misunderstandings** on how `vswhere` works and how Clang finds MSVC toolsets. @compnerd 
1. `vswhere` is **more powerful and robust** than Clang’s internal selector, and LLVM is already using it in its Windows installer, instead of replicating Clang code.
2. `vswhere` is **guaranteed** to be there, in `%ProgramFiles(x86)%/Microsoft Visual Studio/Installer`, ever since Visual Studio 2017. As the path suggests, it’s a part of **the installer bundle**, not Visual Studio itself. The installer will be automatically installed at first run and the installation path cannot be changed.
3. The `vswhere` and Clang typically **gives the same result**. Clang has additional logic to handle MSVC version in triple, which is unsupported in Swift yet. It also prefers `LINK` and `LIB` environment variables, which I believe should be part of https://github.com/apple/swift-package-manager/pull/5981 instead.
4. Clang is also searching for Visual Studio installation, not the standalone MSVC toolset. If you use fancy way to install MSVC instead of `vs_installer.exe`, you should set `VCINSTALLDIR` in the environment. **Installation detection won’t work.**

Sources/PackageModel/UserToolchain.swift
```diff
@@ -117,15 +117,81 @@ public final class UserToolchain: Toolchain {
...
+    func vcArchNames(triple: Triple) -> (component: String, host: String, target: String)? {
+      guard triple.isWindows() else { return nil }
+      switch triple.arch {
+      case .x86_64:
+        return ("Microsoft.VisualStudio.Component.VC.Tools.x86.x64", "HostX64", "x64")
+      case .i686:
+        return ("Microsoft.VisualStudio.Component.VC.Tools.x86.x64", "HostX86", "x86")
+      case .arm64, .arm64e, .aarch64:
+        return ("Microsoft.VisualStudio.Component.VC.Tools.ARM64", "HostARM64", "arm64")
+      case .arm, .armv5, .armv6, .armv7:
```
compnerd, line 152: This is absolutely incorrect.  Anything but `.armv7` should be a hard error.  LLVM does not support WinCE, WinNT does not support anything prior to ARMv7, and even then requires Thumb2.

Sources/PackageModel/UserToolchain.swift
```diff
@@ -117,15 +117,81 @@ public final class UserToolchain: Toolchain {
...
+          } else {
+            return nil
+          }
+        }
+        if let hostTriple = getHostTriple(from: destination, swiftCompilerPath: swiftCompilerPath),
+           let tool = findToolInVisualStudio(name, triple: destination.targetTriple, hostTriple: hostTriple) {
+          return tool
+        }
+      #endif
+    }
```
compnerd, line 137: This makes the behaviour different when targeting Windows from Windows and Windows from another environment just making things more complicated and confusing for users.

stevapple: That's what `useXcrun` is already doing for Darwin to Darwin case, which I believe is reasonable — just like `xcrun` is only available on macOS, Visual Studio is basically a Windows thing.

Sources/PackageModel/UserToolchain.swift
stevapple: Thanks for pointing out, will fix later

stevapple:
> I'd really rather that you would stop with the personal attacks. This seems to be a repetitive thing and really does not help create a pleasant environment.

I believe I didn't use offensive words here, and if you feel offended I sincerely apologize.

For me, Swift is a general-purpose tool, and I'm always trying to lower the barrier of using Swift in general cases. Edge cases are important, but I just don't think a feature can be in a perfect stage when it first gets in the mainstream. _That's not software engineering in these days._

> Would you care to elaborate how it is "more powerful and robust"?

I'll try to explain how Clang handles MSVC detection. Since that's going to be long, see the following comment.

> Is this documented and stated as a public API?

Yes it is: https://learn.microsoft.com/en-us/visualstudio/install/tools-for-managing-visual-studio-instances?view=vs-2022

> The alternative is to use the Visual Studio COM APIs, which are documented public API (and avoids the path issues).

Basically that's how `vswhere` works. You can browse the source code at https://github.com/microsoft/vswhere.

> Sounds like you agree - _typically_. The problem case is where they do not agree, and if that happens, we get into a state that is difficult to diagnose. I think that honoring `LINK` and `LIB` is the right thing to do, and I believe that they should already be honored through `clang`. If not, that is certainly an oversight and not intentional. Would you be up to add some additional test cases to ensure that we do properly honor them?

Will be mentioned in the following comment.

> What this says to me is that the `vswhere` approach is the degenerate case then. We can support just the one way of you specifying the `VCINSTALLDIR` environment variable, but I don't think that is much better than requiring that you run from within the VSDevEnv environment.

I believe it's not. Like you already mentioned, requiring `VCINSTALLDIR` doesn't make difference from requiring VSDevEnv. `vswhere` is not in `Path` by default, but it can be called from a static path, which should work at almost any situations, including VS Code, PowerShell and other developer tools.

stevapple:
Now let's dive in how Clang detects MSVC. Visual Studio has a refreshed interface since its 2017 release, and older ones are called "Legacy Instances". Swift has never supported legacy Visual Studio — and I suppose it never will.

Clang detects MSVC from the following functions in order:
- `findVCToolChainViaCommandLine`: This function checks if Windows SDK path, VC tools path and MSVC tools version are set from command-line arguments, and searches for a toolchain accordingly. This is currently a missing piece of Swift, as it simply ignores related flags and MSVC version set by the triple. I think that's worth a bug report because, for triples with MSVC version, Swift may diverge from Clang. That's not something we can resolve before the Swift compiler supports it.
- `findVCToolChainViaEnvironment`: This function checks if `VCToolsInstallDir` or `VCINSTALLDIR` is set, and searches for a toolchain there. Clang uses `VCToolsInstallDir` to judge if a Visual Studio instance is legacy, but SwiftPM doesn't have to. Clang also searches for `Path` here — that's why I think VS detection should shadow regular env search once we prove it's fully ready.
- `findVCToolChainViaSetupConfig`: This calls the Visual Studio Setup APIs via COM, and is mostly identical to what `vswhere` does. The only difference is, Clang finds the newest Visual Studio and checks if it has MSVC tools installed; `vswhere` can find the newest Visual Studio with MSVC tools directly. That's where I think `vswhere` shines with robustness: If you have VS 2022 without MSVC and VS 2019 with it, `vswhere` can find it while Clang cannot.
- `findVCToolChainViaRegistry`: This function checks the default Visual Studio installation from registry. This is a legacy approach so we won't need it for Swift.

xwu:
@stevapple Is it possible to exhaustively enumerate all the cases where your approach would differ from Clang and demonstrate that they can be diagnosed?

stevapple:
@xwu I could only talk about the latest Clang implementation, which is already, exhaustively I believe, elaborated in https://github.com/apple/swift-package-manager/pull/6120#issuecomment-1420156180. Do you have any question?

Since Clang is evolving itself, the behavior may change, so using Clang API should be better, but I didn't know how to do that in SwiftPM easily. I'd like to first introduce a solution to make sure Swift 5.8 release won't break our clients.

stevapple:
@dafedidi I’m sorry but it seems such solution would not make things better… To use it we need to handle COM calls from Swift (which is possible though SwiftCOM but that’s another dependency), and it’s basically re-implementing a less-robust `vswhere`, which is not we want.

As I had stated, `vswhere` is not a typical “new dependency”. It’s documented and guaranteed to be there as long as you’re able to run Swift today. The risk of using `vswhere` to detect MSVC is that it introduces potential divergence with Clang.

compnerd:
@dafedidi - I think that using the COM interfaces would definitely be something I would be in favour of.  We already do that in the installer.  But that is a well documented interface that has guaranteed stability unlike executing a "random" executable and parsing the output.

PR #6134, opened 2023-02-09
improve registry client error handling

tomerd:
motivation: adjust client behavior to the spec

changes:
* try to parse server error as defined by sepc and produce a structured error with the information returned from the server
* add tests for error cases

PR #6135, opened 2023-02-09
publishing API implementation 

tomerd:
motivation: make progress on publishing from SwiftPM to a registry

changes:
* implement the publishing API in RegistryClient
* integrate the API into the publishing flow
* add tests

### DIFF
--- a/Sources/PackageRegistry/RegistryClient.swift
+++ b/Sources/PackageRegistry/RegistryClient.swift
@@ -749,115 +749,118 @@ public final class RegistryClient: Cancellable {
     }
 
     public func publish(
-            registryURL: URL,
-            packageIdentity: PackageIdentity,
-            packageVersion: Version,
-            packageArchive: AbsolutePath,
-            packageMetadata: AbsolutePath?,
-            timeout: DispatchTimeInterval? = .none,
-            fileSystem: FileSystem,
-            observabilityScope: ObservabilityScope,
-            callbackQueue: DispatchQueue,
-            completion: @escaping (Result<PublishResult, Error>) -> Void
-        ) {
-            let completion = self.makeAsync(completion, on: callbackQueue)
+        registryURL: URL,
+        packageIdentity: PackageIdentity,
+        packageVersion: Version,
+        packageArchive: AbsolutePath,
+        packageMetadata: AbsolutePath?,
+        signature: Data?,
+        timeout: DispatchTimeInterval? = .none,
+        fileSystem: FileSystem,
+        observabilityScope: ObservabilityScope,
+        callbackQueue: DispatchQueue,
+        completion: @escaping (Result<PublishResult, Error>) -> Void
+    ) {
+        let completion = self.makeAsync(completion, on: callbackQueue)
 
-            guard let scopeAndName = packageIdentity.scopeAndName else {
-                return completion(.failure(RegistryError.invalidPackage(packageIdentity)))
-            }
-            guard var components = URLComponents(url: registryURL, resolvingAgainstBaseURL: true) else {
-                return completion(.failure(RegistryError.invalidURL(registryURL)))
-            }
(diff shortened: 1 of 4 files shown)


tomerd:
https://github.com/apple/swift-package-manager/pull/6134 should go first

tomerd:
@yim-lee ptal

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -655,50 +705,160 @@ public final class RegistryClient: Cancellable {
...
+                            } else {
+                                throw RegistryError.invalidResponseStatus(expected: [200], actual: response.statusCode)
+                            }
+                        }
+                    }.mapError {
+                        RegistryError.failedPublishing($0)
+                    }
+                )
+            }
+        }
```
yim-lee, line 860: Formatting doesn't seem right

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -655,50 +705,160 @@ public final class RegistryClient: Cancellable {
...
             )
         }
     }
 
+    public func publish(
+            registryURL: URL,
+            packageIdentity: PackageIdentity,
+            packageVersion: Version,
+            packageArchive: AbsolutePath,
+            packageMetadata: AbsolutePath?,
```
yim-lee, line 756:
```suggestion
            packageMetadata: AbsolutePath?,
            signature: Data?
```

I also wonder if `packageArchive` and `packageMetadata` should be `Data` instead of path here.

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -655,50 +705,160 @@ public final class RegistryClient: Cancellable {
...
+            // archive field
+            body.append(contentsOf: """
+            --\(boundary)
+            Content-Disposition: form-data; name=\"source-archive\"
+            Content-Type: application/zip
+            Content-Transfer-Encoding: base64
+            Content-Length: \(packageArchiveContent.count)
+
+            \(packageArchiveContent.base64EncodedString())
+            """.utf8)
```
yim-lee, line 805: Why do we need `.utf8`?

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -655,50 +705,160 @@ public final class RegistryClient: Cancellable {
...
+            
+            // archive field
+            body.append(contentsOf: """
+            --\(boundary)
+            Content-Disposition: form-data; name=\"source-archive\"
+            Content-Type: application/zip
+            Content-Transfer-Encoding: base64
+            Content-Length: \(packageArchiveContent.count)
+
+            \(packageArchiveContent.base64EncodedString())
```
yim-lee, line 804: I don't know if we need `\r` at the end of each line:

```swift
        Content-Disposition: form-data; name="source-archive"\r
        Content-Type: application/zip\r
        Content-Transfer-Encoding: base64\r
        Content-Length: \(sourceArchive.count)\r
        \r
        \(sourceArchive.base64EncodedString())\r
```

I have it in my prototype. 🤔 We can leave it out for now, but if there is problem parsing the request then this might be a reason.

neonichu (approved):

Sources/PackageRegistry/RegistryClient.swift
tomerd: string -> bytes so we can append to the data

Sources/PackageRegistry/RegistryClient.swift
tomerd: my plan with the path here is that we can avoid loading the file into memory and instead stream it into the upload request. was planning to do this in a follow up PR

tomerd:
@swift-ci smoke test

tomerd:
@swift-ci smoke test

tomerd:
@swift-ci smoke test macOS

PR #6136, opened 2023-02-09
skip zip compression tests in certain environments

tomerd:
motivation: stable ci

changes: skip new zip compression tests where setting the process working environment is not supported

rdar://105175770


tomerd:
@swift-ci smoke test

MaxDesiatov (approved):

PR #6149, opened 2023-02-12
Declare Data.period property is a way compatitible with `+` operator

xedin:
Cannot use `extension Data` if `period` is going to be used with `+` operator via leading-dot syntax, for example: `Data(...) + .period` because `+` is declared as `(Self, Other) -> Self` where `Other: RangeReplaceableCollection, Other.Element == Self.Element` which means that `.period` couldn't get `Data` inferred from the first argument.

### DIFF
--- a/Sources/PackageCollectionsSigning/Utilities/Utilities.swift
+++ b/Sources/PackageCollectionsSigning/Utilities/Utilities.swift
@@ -27,7 +27,12 @@ extension UInt8 {
     }
 }
 
-extension Data {
+/// Cannot use `extension Data` if `period` is going to be used with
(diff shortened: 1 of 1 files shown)


xedin:
Using this PR to test https://github.com/apple/swift/pull/63585

xedin:
@swift-ci please test

xedin:
@swift-ci smoke test

xedin:
@tomerd Would you mind if I landed this if there are no failures? Our performance testing suite cannot handle cross repository testing.

tomerd:
@xedin 👍

tomerd (approved):

neonichu (approved):

PR #6150, opened 2023-02-13
Fix `Sendable` warning in `TokenBucket.swift`

MaxDesiatov:
`CheckedContinuation` is already marked as `Sendable` in Swift 5.6 and later, but with Swift 5.5 on CI we're still seeing warnings related to this.


MaxDesiatov:
@swift-ci smoke test

neonichu (approved):

PR #6153, opened 2023-02-14
[Registry] Make 'Content-Version' optional for archive and manifest download API

yim-lee:
Motivation:
The 'Content-Version' header is required in all registry server responses to indicate API version. However, it shouldn't be required for responses that don't/can't change, such as the download archive and fetch package manifest endpoints. This may also cause problems for registry that have these files hosted elsewhere, because they may not have full control over response headers.

Modifications:
- Modify registry API spec to make 'Content-Version' optional for the said endpoints
- Adjust `RegistryClient`

rdar://105415468

### DIFF
--- a/Documentation/Registry.md
+++ b/Documentation/Registry.md
@@ -1678,13 +1678,13 @@ components:
       schema:
         type: string
         enum:
-          - - "1"
+          - "1"
     optionalContentVersion:
       required: false
       schema:
         type: string
         enum:
-          - - "1"          
+          - "1"          
 
 ```
 
--- a/Documentation/Registry.md
+++ b/Documentation/Registry.md
@@ -183,8 +183,12 @@ Valid `Accept` header field values are described by the following rules:
     accept      = "application/vnd.swift.registry" [".v" version] ["+" mediatype]
 ```
 
-A server MUST set the `Content-Type` and `Content-Version` header fields
-with the corresponding content type and API version number of the response.
+A server MUST set the `Content-Type` header field
+with the corresponding content type of the response. 
+
+A server MUST set the `Content-Version` header field
+with the API version number of the response, unless 
+explicitly stated otherwise. 
 
 ```http
 HTTP/1.1 200 OK
@@ -594,6 +598,10 @@ set to `attachment` with a `filename` parameter equal to
 the name of the manifest file
 (for example, "Package.swift").
 
+A server MAY omit the `Content-Version` header
+since the response content (i.e., the manifest) SHOULD NOT
+change across different API versions.
+
 It is RECOMMENDED for clients and servers to support
 caching as described by [RFC 7234].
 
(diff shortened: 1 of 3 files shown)


yim-lee:
@swift-ci please test

yim-lee:
@swift-ci please smoke test

Documentation/Registry.md
```diff
@@ -1379,7 +1391,7 @@ paths:
               schema:
                 type: integer
             Content-Version:
-              $ref: "#/components/headers/contentVersion"
+              $ref: "#/components/headers/optionalContentVersion"
```
yim-lee, line 1394: AFAIK we can't override `$ref`. 

Ideally we'd want to do something like this:

```suggestion
              $ref: "#/components/headers/contentVersion"
              required: false
```

yim-lee:

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -1095,8 +1095,15 @@ extension HTTPClientResponse {
         }
     }
 
-    fileprivate func validateAPIVersion(_ expectedVersion: RegistryClient.APIVersion = .v1) throws {
-        guard self.apiVersion == expectedVersion else {
+    fileprivate func validateAPIVersion(_ expectedVersion: RegistryClient.APIVersion = .v1, isOptional: Bool = false) throws {
+        let apiVersion = self.apiVersion
```
tomerd, line 1099: nit: redundant?

tomerd (approved):
lgtm

Documentation/Registry.md
tomerd: cc @czechboy0

Sources/PackageRegistry/RegistryClient.swift
yim-lee: We use this so we don't need to call `self.apiVersion` twice.

yim-lee:

neonichu (approved):

Documentation/Registry.md
czechboy0: That's right, whether a header is required is part of the content, so to have an optional variant of the same header, we have to fully duplicate it.

Documentation/Registry.md
```diff
@@ -1667,6 +1679,12 @@ components:
         type: string
         enum:
           - - "1"
+    optionalContentVersion:
+      required: false
+      schema:
+        type: string
+        enum:
+          - - "1"
```
czechboy0, line 1687: Why the double dash here? (Both here and in `contentVersion` above), I'd expect just

```yaml
enum:
  - "1"
```

yim-lee: I am not sure. I thought it was some weird syntax that I didn't know about. Amended in https://github.com/apple/swift-package-manager/pull/6153/commits/69a4a024996b8c528e556ab905980dd7e871848b

yim-lee:
@swift-ci please smoke test

czechboy0 (approved):

yim-lee:

yim-lee:
@swift-ci please smoke test

yim-lee:

PR #6155, opened 2023-02-14
Fix typo in publish API example

yim-lee:
Motivation:
To create a package release, one should do a `PUT /{scope}/{name}/{version}` HTTP request. That's what it's said everywhere in the spec except in the example.

Modification:
Fix example request to be consistent with the spec.


yim-lee:
@swift-ci please smoke test

neonichu (approved):

PR #6158, opened 2023-02-15
Fix registry package tool

yim-lee:
- `certificateChainPaths` is an empty array by default (can't be `nil`), so even though neither `privateKeyPath` nor `certificateChainPaths` is specified, thus signing is not needed, the tool fails with missing signing inputs error.
- Fix typo in error message

### DIFF
--- a/Sources/PackageRegistryTool/PackageRegistryTool+Publish.swift
+++ b/Sources/PackageRegistryTool/PackageRegistryTool+Publish.swift
@@ -153,14 +153,8 @@ extension SwiftPackageRegistryTool {
                 )
             }
 
-            if publishConfiguration.signing.privateKeyPath == nil {
+            if publishConfiguration.signing.privateKeyPath != nil {
                 guard !publishConfiguration.signing.certificateChainPaths.isEmpty else {
-                    throw StringError(
-                        "Both 'privateKeyPath' and 'certificateChainPaths' are required when one of them is set."
-                    )
-                }
-            } else {
-                guard publishConfiguration.signing.certificateChainPaths.isEmpty else {
                     throw StringError(
                         "Both 'privateKeyPath' and 'certificateChainPaths' are required when one of them is set."
                     )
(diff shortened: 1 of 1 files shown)


yim-lee:
@swift-ci please smoke test

neonichu (approved):

Sources/PackageRegistryTool/PackageRegistryTool+Publish.swift
```diff
@@ -153,18 +153,12 @@ extension SwiftPackageRegistryTool {
...
                     throw StringError(
                         "Both 'privateKeyPath' and 'certificateChainPaths' are required when one of them is set."
                     )
                 }
-            } else {
-                guard publishConfiguration.signing.certificateChainPaths.isEmpty else {
-                    throw StringError(
-                        "Both 'privateKeyPath' and 'certificateChainPaths' are required when one of them is set."
-                    )
-                }
```
tomerd, line 167: iiuc correctly they need to be both set or both not-set? if so, a fix could be:

```
if publishConfiguration.signing.privateKeyPath != nil {
                guard !publishConfiguration.signing.certificateChainPaths.isEmpty else {
                    throw StringError(
                        "Both 'privateKeyPath' and 'certificateChainPaths' are required when one of them is set."
                    )
                }
            } else {
                guard publishConfiguration.signing.certificateChainPaths.isEmpty else {
                    throw StringError(
                        "Both 'privateKeyPath' and 'certificateChainPaths' are required when one of them is set."
                    )
                }
            }
```

yim-lee: The else guard would fail if omit `privateKeyPath` and `certificateChainPaths` in the call, right? Since `certificateChainPaths` defaults to `[]` and I don't think we can change the default to `nil`.

yim-lee:
@swift-ci please smoke test

yim-lee:

tomerd (approved):
Thank you for fixing all my bugs <3

PR #6161, opened 2023-02-16
[wip] streaming uploads

tomerd:
motivation: improve upload to stream the content of the files instead of loading them to memory

changes:
* introduce new upload requiest type
* implement stream upload hanlding with URLSession
* implement an upload stream provider for registry publishing

TODO:
* add tests


Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
```diff
@@ -337,6 +353,270 @@ private class DownloadTaskManager: NSObject, URLSessionDownloadDelegate {
...
+                        inputStream.open()
+                        if inputStream.hasBytesAvailable {
+                            var buffer = [UInt8](repeating: 0, count: Self.bufferSize)
+                            let bytesRead = inputStream.read(&buffer, maxLength: buffer.count)
+                            if bytesRead == 0 {
+                                // done with stream
+                                self.dataToWrite = .none
+                                inputStream.close()
+                            } else {
+                                let bytesWritten = self.outputStream.write(buffer, maxLength: bytesRead)
```
weissi, line 588: what happens with the bytes that didn't get written? (i.e. `bytesWritten < bytesRead`)?

tomerd: great catch. is there a way to move the reader index back? otherwise I can buffer them

Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
```diff
@@ -337,6 +353,270 @@ private class DownloadTaskManager: NSObject, URLSessionDownloadDelegate {
...
+        _ session: URLSession,
+        task: URLSessionTask,
+        needNewBodyStream completionHandler: @escaping @Sendable (InputStream?) -> Void
+    ) {
+        guard let task = self.tasks[task.taskIdentifier] else {
+            return
+        }
+        completionHandler(task.inputStream)
+        // FIXME: serious hack here
+        RunLoop.current.run(until: Date(timeIntervalSinceNow: 10000.0))
```
tomerd, line 395: @weissi any ideas on why the stream is not going notified unless I run the .current runloop manually?

Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
weissi: No, you can't unread. You'll need to buffer them and you'll need to stop reading until you're done with the previous writes.

Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
weissi: that almost certainly means that you're blocking the runloop somewhere.

PR #6163, opened 2023-02-16
do not use base64 encoding when publishing archives

tomerd:
motivation: base64 emcoding is redundant in this case

changes:
* upload the archive in binary form instead of base64 encoded
* refactor the multi-part form construction logic

### DIFF
--- a/Sources/PackageRegistry/RegistryClient.swift
+++ b/Sources/PackageRegistry/RegistryClient.swift
@@ -692,34 +692,33 @@ public final class RegistryClient: Cancellable {
 
         // TODO: add generic support for upload requests in Basics
         let boundary = UUID().uuidString
-        var parts = [String]()
+        var body = Data()
 
         // archive field
-        parts.append("""
+        body.append(contentsOf: """
+        --\(boundary)\r
         Content-Disposition: form-data; name=\"source-archive\"\r
         Content-Type: application/zip\r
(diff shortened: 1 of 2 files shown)


tomerd:
@swift-ci smoke test

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -692,34 +692,33 @@ public final class RegistryClient: Cancellable {
...
+        var body = Data()
 
         // archive field
-        parts.append("""
+        body.append(contentsOf: """
+        --\(boundary)\r
         Content-Disposition: form-data; name=\"source-archive\"\r
         Content-Type: application/zip\r
-        Content-Transfer-Encoding: base64\r
-        Content-Length: \(packageArchiveContent.count)\r
```
yim-lee, line 702: The [spec](https://github.com/apple/swift-package-manager/blob/main/Documentation/Registry.md#46-create-a-package-release) says that client must set `Content-Length`. Do we need to update the spec?

tomerd: I think so, when streaming content it is hard to specify it

tomerd:
@swift-ci smoke test

tomerd:
@swift-ci smoke test macOS

yim-lee (approved):

yim-lee:

PR #6174, opened 2023-02-22
[SwiftTestTool] Add test failure messages to xUnit report

cyberowl:
### Motivation

The xUnit test run report does not provide detailed failure information and this makes it very difficult to investigate problems without additional viewing of stderr logs on some CI systems

### Modifications

1. Added the use of regular expressions to parse the output of test failures and crashes too (i think it's a simplest approach at this moment)
2. Added line breaks to the output of parallel testing in case of regular expressions usage
3. Added test coverage for introduced changes (except "Fatal Error" case for a known reason)

### Result

Better xUnit report with test failures

Before:
```
<testsuites>
<testsuite name="TestResults" errors="0" tests="5" failures="3" time="21.489489211">
<testcase classname="ParallelTestsPkgTests.ParallelTestsFailureTests" name="testExpectationFailure" time="4.297888375">
<failure message="failed"></failure>
</testcase>
<testcase classname="ParallelTestsPkgTests.ParallelTestsFailureTests" name="testAssertionFailure" time="4.297924834">
<failure message="failed"></failure>
</testcase>
<testcase classname="ParallelTestsPkgTests.ParallelTestsTests" name="testExample1" time="4.297897584">
</testcase>
<testcase classname="ParallelTestsPkgTests.ParallelTestsTests" name="testExample2" time="4.297898459">
</testcase>
<testcase classname="ParallelTestsPkgTests.ParallelTestsFailureTests" name="testSureFailure" time="4.297879959">
<failure message="failed"></failure>
</testcase>
</testsuite>
</testsuites>
```

After:
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
<testsuite name="TestResults" errors="0" tests="5" failures="3" time="15.821316957">
<testcase classname="ParallelTestsPkgTests.ParallelTestsTests" name="testExample2" time="0.04732175">
</testcase>
<testcase classname="ParallelTestsPkgTests.ParallelTestsFailureTests" name="testExpectationFailure" time="3.29376925">
<failure message="ParallelTestsFailureTests.swift:16 Asynchronous wait failed: Exceeded timeout of 0 seconds, with unfulfilled expectations: &quot;failing expectation&quot;."></failure>
</testcase>
<testcase classname="ParallelTestsPkgTests.ParallelTestsTests" name="testExample1" time="4.159991583">
</testcase>
<testcase classname="ParallelTestsPkgTests.ParallelTestsFailureTests" name="testSureFailure" time="4.160105583">
<failure message="ParallelTestsFailureTests.swift:7 failed - Giving up is the only sure way to fail."></failure>
</testcase>
<testcase classname="ParallelTestsPkgTests.ParallelTestsFailureTests" name="testAssertionFailure" time="4.160128791">
<failure message="ParallelTestsFailureTests.swift:11 XCTAssertTrue failed - Expected assertion failure."></failure>
</testcase>
</testsuite>
</testsuites>
```

### DIFF
--- a/Sources/Commands/SwiftTestTool.swift
+++ b/Sources/Commands/SwiftTestTool.swift
@@ -816,7 +816,7 @@ final class ParallelTestRunner {
                     var output = ""
                     let outputLock = NSLock()
                     let start = DispatchTime.now()
-                    let success = testRunner.test(outputHandler: { _output in outputLock.withLock{ output += _output }})
+                    let success = testRunner.test(outputHandler: { _output in outputLock.withLock{ output += _output.appending("\n") }})
                     let duration = start.distance(to: .now())
                     if !success {
                         self.ranSuccessfully = false
@@ -1034,9 +1034,17 @@ final class XUnitGenerator {
 
             """
 
+        /// Captured groups
+        /// $1 = file:line
+        /// $2 = reason
+        let testErrorRegex = try RegEx(pattern: #".*\/(.*\d+)\:\serror:.*\s\:\s(.*)"#)
+
+        // Captured groups
+        // $1 = file:line
+        // $2 = reason
+        let fatalErrorRegex = try RegEx(pattern: #".*\/(.*\d+)\:\s(Fatal error:.*)"#)
+
         // Generate a testcase entry for each result.
-        //
-        // FIXME: This is very minimal right now. We should allow including test output etc.
         for result in results {
             let test = result.unitTest
             let duration = result.duration.timeInterval() ?? 0.0
@@ -1046,7 +1054,13 @@ final class XUnitGenerator {
                 """
 
             if !result.success {
-                stream <<< "<failure message=\"failed\"></failure>\n"
+                let message = [testErrorRegex, fatalErrorRegex].map { $0.matchGroups(in: result.output) }
+                    .first(where: { !$0.isEmpty })?
+                    .first?
+                    .joined(separator: " ")
+                    .xmlEscaped ?? "failed"
+
+                stream <<< "<failure message=\"\(message)\"></failure>\n"
             }
 
             stream <<< "</testcase>\n"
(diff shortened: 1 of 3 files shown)


Sources/Commands/SwiftTestTool.swift
```diff
@@ -1046,7 +1054,13 @@ final class XUnitGenerator {
...
 
             if !result.success {
-                stream <<< "<failure message=\"failed\"></failure>\n"
+                let message = [testErrorRegex, fatalErrorRegex].map { $0.matchGroups(in: result.output) }
+                    .first(where: { !$0.isEmpty })?
+                    .first?
+                    .joined(separator: " ")
+                    .xmlEscaped ?? "failed"
+
+                stream <<< "<failure message=\"\(message)\"></failure>\n"
```
tomerd, line 1063: this seems like an improvement over the hard coded "failed" message, but a bit fragile. @stmontgomery any advice?

stmontgomery: It really should not be necessary to scrape the textual logging output of XCTests and parse it using regex. We have a proper API for observing events and retrieving this data in structured ways: [XCTestObservation](https://developer.apple.com/documentation/xctest/xctestobservation).

However, I think SwiftPM needs to coordinate doing this for it to be effective but it doesn't do that today. So I can understand why this approach relying on more-and-more intricate regex parsing is seen as the easiest solution at the moment. I believe SwiftPM already parses the textual logging output to a certain extent (right?), so this is only adding a bit more and I'm not going to stand in the way of landing this. But I think we really should spend some time building the necessary infrastructure for SwiftPM to coordinate more closely with the runner process to gather its results in a structured way, to avoid the need for any of this.

@tomerd Do you know if we already have an Issue tracking this sort of enhancement on the SwiftPM side?

tomerd: @stmontgomery we have been pushing back on adding parsing of XCTest output and we would very much like to see a structured output from XCTest instead of the non-structured one we have now - it will help solve several issues. iirc the chalk age is that the invocation of XCTest goes through several layers that essentially shell out so there is little room for setting up delegates etc, but maybe there was a flag we can pass to the process that can assist? open to ideas on how to be at achieve this

PR #6178, opened 2023-02-23
NFC: cleanup usage of registry identifiers

tomerd:
motivation: make code easier to reason about

changes:
* deprecate scopeAndName accessor in favour of .registry one
* define RegistryIdentity struct to encapsulate such identity
* update call sites and tests


tomerd:
cc @yim-lee

tomerd:
@swift-ci smoke test

yim-lee (approved):

PR #6179, opened 2023-02-23
DNM: On ELF platforms, look for the runtime libraries in an architecture-specific directory.

buttaface:
This is needed for apple/swift#63782, which changes the Unix toolchain to look for libraries in architecture-specific directories.

### DIFF
--- a/Sources/Commands/Utilities/TestingSupport.swift
+++ b/Sources/Commands/Utilities/TestingSupport.swift
@@ -134,9 +134,10 @@ enum TestingSupport {
             env.prependPath("Path", value: location.pathString)
         }
         #elseif os(Linux)
-        var libraryPaths = ["/usr/lib/swift/linux"]
+        let arch = buildParameters.triple.arch.rawValue
+        var libraryPaths = ["/usr/lib/swift/linux/\(arch)"]
         if let path = env["PATH"], let firstPathEntry = path.components(separatedBy: ":").first {
-            libraryPaths.append("\(firstPathEntry)/../lib/swift/linux")
+            libraryPaths.append("\(firstPathEntry)/../lib/swift/linux/\(arch)")
         }
         if let originalLibraryPaths = env["LD_LIBRARY_PATH"] {
             libraryPaths.append(originalLibraryPaths)
--- a/Tests/PackageModelTests/DestinationTests.swift
+++ b/Tests/PackageModelTests/DestinationTests.swift
@@ -27,7 +27,7 @@ private let extraFlags = BuildFlags(
     cCompilerFlags: ["-fintegrated-as"],
     cxxCompilerFlags: ["-fno-exceptions"],
     swiftCompilerFlags: ["-enable-experimental-cxx-interop", "-use-ld=lld"],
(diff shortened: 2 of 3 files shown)


tomerd:
cc @MaxDesiatov @etcwilde @compnerd

tomerd:
@swift-ci smoke test

compnerd (commented):
Seems reasonable to me (other than `Utilities/bootstrap` which I didn't look at).  Not approving to give others a chance to look at it as well.

MaxDesiatov:
Could we use this as an opportunity to encode more triple components in that path, such as ABI/environment, which would help with Musl support, or this is meant to be a purely incremental change?

neonichu (approved):

etcwilde (approved):

buttaface:
@MaxDesiatov, we could do it together, what scheme do you have in mind? Alternatively, if we simply name the platform `linux-musl` instead, ie store the libraries in `lib/swift/linux-musl/aarch64/`, it would require no changes to where the compiler looks for these libraries.

MaxDesiatov:
> Alternatively, if we simply name the platform `linux-musl` instead, ie store the libraries in `lib/swift/linux-musl/aarch64/`, it would require no changes to where the compiler looks for these libraries.

I already tried that route and it doesn't seem to be practical. There is enough code in the build system that identifies Musl-based distros as plain `linux`, and that gets propagated into multiple places that reconstitute these paths from scratch, specifically ClangImporter, but I'm pretty sure there are more. I hope we can just stick to using LLVM/Clang triples instead of coming up with our own directory layout and naming.

buttaface:
> There is enough code in the build system that identifies Musl-based distros as plain linux, and that gets propagated into multiple places that reconstitute these paths from scratch, specifically ClangImporter, but I'm pretty sure there are more.

Well, we will have to change the build system either way, but it would be easier to make these changes you list here than move the Swift resource directory to a whole new triple-based naming scheme.

> I hope we can just stick to using LLVM/Clang triples instead of coming up with our own directory layout and naming.

As in we move to a whole new scheme like `lib/swift/aarch64-linux-musl/` or change the current layout to handle `lib/swift/linux/aarch64-musl` too? Any of these alternatives to the simpler change I suggested will require much _more_ changes to the build system and changes to the compiler too, the latter of which my suggestion wouldn't require.

I think you raise an important point that Swift should start supporting alternate C libraries and ABIs, particularly for linux, but the question is how best to do so.

tomerd:
@MaxDesiatov @buttaface is this still pending, or the discussion impact this PR?

buttaface:
@tomerd, this is under active testing with the linked compiler pull apple/swift#63782 still: once it passes the CI, I will ask for all 8 pulls to be merged simultaneously.

buttaface:
Alright, the CI passed and the Swift compiler pull was approved, the draft/WIP tags can be removed, @tomerd.

Please don't merge yet though, as all 11 pulls should be merged together.

buttaface:
Closing since [the Swift devs want to take a different approach](https://github.com/apple/swift/pull/63782#pullrequestreview-1353429744).

PR #6187, opened 2023-02-24
Fix path to registries configuration file in gitignore

Palleas:
Update the path to the registries configuration file in the provided `.gitiginore` that is created when a new package is initialized via `swift package init`.

### Motivation:

Running `swift package init` then `swift package-registry set ...` might cause the `registries.json` file to be checked into the repository since the ignored path is wrong.

### Modifications:

Update the `swift package init` command to generate a file with the proper path to the `registries.json` file.

### Result:

Changes in the generated .gitignore file.

### DIFF
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -265,7 +265,7 @@ public final class InitPackage {
                 /Packages
                 xcuserdata/
                 DerivedData/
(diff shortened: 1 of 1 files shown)


neonichu (approved):

tomerd (approved):

tomerd:
@swift-ci smoke test

PR #6191, opened 2023-02-27
Update registry API spec: base64 encoding not required

yim-lee:
Motivation:
The example in section 4.6 might lead reader to think that certain parts of the request body must be base64-encoded.

Modifications:
- Make it clear that base64-encoding is not required. Client might sent binary and server should be able to handle it.
- Make it clear that `boundary` can be any random string and doesn't have to be `boundary`. Server should be able to handle any boundary string.
- Make `Content-Length` optional since it might not be set if request is streamed.


yim-lee:
@swift-ci please smoke test

tomerd (approved):

neonichu (approved):

PR #6195, opened 2023-02-28
PackageModel/Toolset: fix support for relative `rootPath`

MaxDesiatov:
Currently `DecodedToolset` only decodes `AbsolutePath` directly, while it should be able to support relative paths as well. Let's fix that by decoding to `String` first and then resolving to `AbsolutePath` when creating a `Toolset` value from `DecodedToolset`. Also added a corresponding test assertion that verifies that applied fix works.

### DIFF
--- a/Sources/PackageModel/Toolset.swift
+++ b/Sources/PackageModel/Toolset.swift
@@ -76,14 +76,9 @@ extension Toolset {
             )
         }
 
-        let rootPaths: [AbsolutePath]
-        if let rootPath = decoded.rootPath {
-            rootPaths = [
-                try AbsolutePath(validating: rootPath, relativeTo: toolsetPath.parentDirectory)
-            ]
-        } else {
-            rootPaths = []
-        }
+        let rootPaths = try decoded.rootPath.map {
+            [try AbsolutePath(validating: $0, relativeTo: toolsetPath.parentDirectory)]
+        } ?? []
 
         var knownTools = [KnownTool: ToolProperties]()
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci smoke test

Sources/PackageModel/Toolset.swift
```diff
@@ -76,6 +76,15 @@ extension Toolset {
             )
         }
 
+        let rootPaths: [AbsolutePath]
+        if let rootPath = decoded.rootPath {
```
tomerd, line 80: nit: can use `map`?

tomerd (approved):

neonichu (approved):

Sources/PackageModel/Toolset.swift
MaxDesiatov: We can, not sure if this is more readable?

```swift
        let rootPaths = try decoded.rootPath.map {
            [
                try AbsolutePath(validating: $0, relativeTo: toolsetPath.parentDirectory)
            ]
        } ?? []
```

I was initially a proponent of `map` everywhere for optionals due to my FP background, but eventually came to prefer `if let` from a readability perspective, and it's not much more verbose if at all.

MaxDesiatov: I'd love to use SE-0380 (`if` expression) here eventually when that becomes available though.

MaxDesiatov:
@swift-ci smoke test

PR #6202, opened 2023-03-01
Fix `InstallDestination.run` function not invoked by ArgumentParser

MaxDesiatov:
This type uses `ParsableCommand`, not `AsyncParsableCommand`, while the `run` function itself was marked as `async`. Because of this the default implementation of `run` was always called, printing the help description even when correct arguments were passed. Removing the `async` effect from the function fixes the issue, since the function doesn't use `await` in its body anyway (due to unrelated reasons).

Also added quotes around URLs and paths in logging output so that it's easier to parse them visually, especially when those contain spaces.


MaxDesiatov:
@swift-ci smoke test

neonichu (approved):

MaxDesiatov:
@swift-ci smoke test

PR #6205, opened 2023-03-01
PackageModel: fix error reporting for invalid toolsets

MaxDesiatov:
Previously, when parsing a destination file that referenced an invalid toolset, a decoding error for the toolset was propagated as a `StringError` that mentioned the destination file, which was quite confusing. It looked as if the destination file itself couldn't be parsed, which wasn't true. Cleaning up error handling here to explicitly mention the toolset file and its path if an error is thrown during its decoding.

### DIFF
--- a/Sources/PackageModel/Destination.swift
+++ b/Sources/PackageModel/Destination.swift
@@ -338,7 +338,7 @@ extension Destination {
                 decoder: decoder,
                 observabilityScope: observabilityScope
             )
-        } catch {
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

neonichu (approved):

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci smoke test

PR #6211, opened 2023-03-02
Basics: explicitly mark `SQLite` as non-Sendable reference type

MaxDesiatov:
### Motivation:

`SQLite` should be a reference type as it stores a pointer to the database. For some reason it was declared as `struct`, but to indicate its reference semantics it would be cleaner to mark it as `final class`. Additionally, this type should never be `Sendable` as it provides no locking, thus an explicit annotation should be added to indicate this.

### Modifications:

Changed the `SQLite` type from `struct` to `final class`  and added a declaration that marks it as non-Sendable.

### Result:

It's more obvious now that `SQLite` has reference semantics. In the future we could also call `close()` automatically from `SQLite.deinit`, but that needs a separate audit of existing calls to `close()`.

### DIFF
--- a/Sources/Basics/SQLite.swift
+++ b/Sources/Basics/SQLite.swift
@@ -17,7 +17,7 @@ import TSCBasic
 @_implementationOnly import SPMSQLite3
 
 /// A minimal SQLite wrapper.
-public struct SQLite {
+public final class SQLite {
     /// The location of the database.
     public let location: Location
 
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci smoke test

neonichu (approved):

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci smoke test Windows platform

MaxDesiatov:
@swift-ci smoke test Windows platform

MaxDesiatov:
@swift-ci smoke test macOS

neonichu (approved):

PR #6220, opened 2023-03-03
Wire up SignatureValidation in RegistryClient

yim-lee:
Also change default onUnsigned and onUntrustedCert to log warning instead of prompt since the latter is not yet supported.


yim-lee:
@swift-ci please smoke test

tomerd (approved):

PR #6222, opened 2023-03-03
CrossCompilationDestinationsTool: add diagnostics for no destinations

MaxDesiatov:
Currently when no CC destinations are installed, `swift experimental-destination list` produces no output, which is confusing. Let's fix that by providing a message that clarifies that no installed destinations are available.


MaxDesiatov:
@swift-ci smoke test

PR #6223, opened 2023-03-03
[NFC] Move ArtifactsArchiveMetadata to PackageModel

MaxDesiatov:
For the new `swift experimental-destination configure` subcommand we need to add configuration storage logic, which is tightly coupled to serialized destination data, which is internal to `PackageModel`. It would be beneficial to keep those serialized model types internal, but we also need to be able to access `ArtifactsArchiveMetadata` from the configuration logic.

Since that type seems to be purely a model value type related to packages, it feels reasonable to move it to `PackageModel`. Configuration logic provided in a future PR can then stay in `PackageModel` as well, without a need to expose internal model types, only providing configuration methods as `public`.

Also using this as an opportunity to rename `DestinationsBundle` to `DestinationBundle` for consistency with other model types that don't use plural.

### DIFF
--- a/Sources/CoreCommands/SwiftTool.swift
+++ b/Sources/CoreCommands/SwiftTool.swift
@@ -776,7 +776,7 @@ public final class SwiftTool {
             {
                 destination = targetDestination
             } else if let destinationSelector = options.build.crossCompilationDestinationSelector {
-                destination = try DestinationsBundle.selectDestination(
+                destination = try DestinationBundle.selectDestination(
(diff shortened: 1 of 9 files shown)


MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test Windows

MaxDesiatov:
@swift-ci test Windows

neonichu (approved):

PR #6225, opened 2023-03-03
[NFC] PackageModel: rename `validateDestinationsBundle` for consistency

MaxDesiatov:
We already use singular form everywhere, but somehow this method ended up having a different name, let's clean that up.


MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

neonichu (approved):

PR #6231, opened 2023-03-04
serialize registry metadata and cache it so it can be exposed to clients

tomerd:
motivation: clients of SwiftPM may be interested in the registry metadata for presentation

changes: serialize the metadata when downloading registry based  packages and deserializae it when construting the graph

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -233,7 +233,8 @@ let package = Package(
             name: "PackageLoading",
             dependencies: [
                 "Basics",
-                "PackageModel"
+                "PackageModel",
+                "PackageSigning"
             ],
             exclude: ["CMakeLists.txt", "README.md"]
         ),
--- a/Package.swift
+++ b/Package.swift
@@ -233,8 +233,7 @@ let package = Package(
             name: "PackageLoading",
             dependencies: [
                 "Basics",
-                "PackageModel",
-                "PackageSigning"
+                "PackageModel"
             ],
             exclude: ["CMakeLists.txt", "README.md"]
         ),
--- a/Sources/PackageLoading/RegistryReleaseMetadataSerialization.swift
+++ b/Sources/PackageLoading/RegistryReleaseMetadataSerialization.swift
@@ -1,0 +1,173 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Basics
+@_implementationOnly import Foundation
+import PackageModel
+import PackageSigning
+import TSCBasic
+
+public enum RegistryReleaseMetadataStorage {
+    public static let fileName = ".registry-metadata"
+
+    private static let encoder = JSONEncoder.makeWithDefaults()
+    private static let decoder = JSONDecoder.makeWithDefaults()
+
+    public static func save(_ metadata: RegistryReleaseMetadata, to path: AbsolutePath, fileSystem: FileSystem) throws {
+        let codableMetadata = CodableRegistryReleaseMetadata(metadata)
+        let data = try Self.encoder.encode(codableMetadata)
+        try fileSystem.writeFileContents(path, data: data)
+    }
+
+    public static func load(from path: AbsolutePath, fileSystem: FileSystem) throws -> RegistryReleaseMetadata {
+        let codableMetadata = try Self.decoder.decode(
+            path: path,
+            fileSystem: fileSystem,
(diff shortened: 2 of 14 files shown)


Sources/PackageLoading/RegistryReleaseMetadataSerialization.swift
```diff
@@ -0,0 +1,141 @@
...
+            fileSystem: fileSystem,
+            as: CodableRegistryReleaseMetadata.self
+        )
+        return RegistryReleaseMetadata(codableMetadata)
+    }
+}
+
+private struct CodableRegistryReleaseMetadata: Codable {
+    public let registry: URL
+    public let signature: Signature?
```
yim-lee, line 43: This is more like "who generated the signature"/"signed by" than "signature".

Sources/PackageLoading/RegistryReleaseMetadataSerialization.swift
```diff
@@ -0,0 +1,141 @@
...
+                }
+            )
+        }
+        self.description = seed.metadata.description
+        self.licenseURL = seed.metadata.licenseURL
+        self.readmeURL = seed.metadata.readmeURL
+        self.scmRepositoryURLs = seed.metadata.scmRepositoryURLs
+    }
+
+    public enum Signature: Codable {
```
yim-lee, line 85: ditto ^. I would call this e.g., `Signer`, `SigningEntity`.

Sources/PackageLoading/RegistryReleaseMetadataSerialization.swift
```diff
@@ -0,0 +1,141 @@
...
+        case recognized(commonName: String?, organization: String?, identity: String?)
```
yim-lee, line 86: Probably add `type` which gets its value from `SigningEntity.type`. `identity` would be `SigningEntity.organizationalUnit` in case of type `adp`.

Sources/PackageModel/RegistryReleaseMetadata.swift
```diff
@@ -13,45 +13,88 @@
 import struct Foundation.URL
 import struct TSCUtility.Version
 
+/// Information from the signing certificate.
+public enum RegistrySignature {
```
yim-lee, line 17: Similar feedback as above

Sources/PackageLoading/RegistryReleaseMetadataSerialization.swift
yim-lee: And have it inside a `Signature` container struct if we think there will be more properties in the future.

tomerd:
@swift-ci smokte test

tomerd:
@swift-ci smoke test

Package.swift
```diff
@@ -233,7 +233,8 @@ let package = Package(
             name: "PackageLoading",
             dependencies: [
                 "Basics",
-                "PackageModel"
+                "PackageModel",
+                "PackageSigning"
```
yim-lee, line 237: What is this dependency used for?

Sources/PackageLoading/RegistryReleaseMetadataSerialization.swift
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    public struct Signature: Codable {
+        let signedBy: SigningEntity?
+        let format: String
+        let base64: String
+    }
+
+    public enum SigningEntity: Codable {
+        case recognized(type: String?, commonName: String?, organization: String?, identity: String?)
```
yim-lee, line 103: `type` can be non-optional I think

Sources/PackageModel/RegistryReleaseMetadata.swift
```diff
@@ -13,45 +13,105 @@
...
+            value: [UInt8]
+        ) {
+            self.signedBy = signedBy
+            self.format = format
+            self.value = value
+        }
     }
 
+    public enum SigningEntity: Codable {
+        case recognized(type: String?, commonName: String?, organization: String?, identity: String?)
```
yim-lee, line 109: ditto ^. `type` can be non-optional.

Package.swift
tomerd: good catch, this is leftover

Sources/PackageLoading/RegistryReleaseMetadataSerialization.swift
tomerd: these are all optional on PackageSigning.SigningEntity causing this to also need to be optional. I guess we could also change PackageSigning.SigningEntity to carry these as associated values which will help. is this what you have in mind?

tomerd:
@swift-ci smoke test

Sources/PackageLoading/RegistryReleaseMetadataSerialization.swift
tomerd: addressed for now, perhaps we can make PackageSigning.SigningEntity an enum with associated values, I think it may model things nicely

tomerd:
@yim-lee @neonichu this is ready

yim-lee (approved):

PR #6243, opened 2023-03-08
[NFC] Build: format `LLBuildManifestBuilder.swift`

MaxDesiatov:
As more changes will be made to this file, let's keep the formatting changes separate. This only contains results of applying SwiftFormat to the file.


MaxDesiatov:
@swift-ci smoke test

PR #6246, opened 2023-03-08
PackageModel: add DestinationConfigurationStore

MaxDesiatov:
### Motivation:

To allow users to configure installed cross-compilation destinations by providing custom paths we need to store these custom paths on the file system. 

### Modifications:

New `DestinationConfigurationStore` allows writing out these customized properties as JSON files in a separate directory in `~/.swiftpm/destinations` or wherever that paths is symlinked to.

### Result:

This is a non-functional change, new `DestinationConfigurationStore` is not used anywhere yet, but will be used in new experimental subcommands to be added in future PRs.


neonichu (approved):

MaxDesiatov:
@swift-ci smoke test

PR #6252, opened 2023-03-09
Add `experimental-destination configuration` subcommand

MaxDesiatov:
### Motivation:

For configuring destination bundles that can't be self-contained and have to reference absolute paths we need new to provide a CLI. This comes as a new `swift experimental-destination configuration` subcommand per the proposed changes in https://github.com/apple/swift-evolution/pull/1942.

### Modifications:

`DestinationCommand` protocol is refined to wait for the observability handler to flush all output. Existing destination-related subcommands are modified to conform to it. Additionally, it now passes a newly initialized observability scope and configuration values.

Two new subcommands are added: `configuration show` and `configuration reset`. For actually updating configuration values a third subcommand `configuration set` will be added in a separate PR.

### Result:

Users can now view and reset destination-related paths configuration.

### DIFF
--- a/Sources/CoreCommands/SwiftToolObservabilityHandler.swift
+++ b/Sources/CoreCommands/SwiftToolObservabilityHandler.swift
@@ -48,7 +48,7 @@ public struct SwiftToolObservabilityHandler: ObservabilityHandlerProvider {
         self.outputHandler.outputStream
     }
 
-    func wait(timeout: DispatchTime) {
+    public func wait(timeout: DispatchTime) {
         self.outputHandler.wait(timeout: timeout)
     }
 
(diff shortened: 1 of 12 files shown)


MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci smoke test

neonichu (approved):

Sources/CoreCommands/SwiftToolObservabilityHandler.swift
```diff
@@ -198,7 +198,8 @@ extension ObservabilitySystem {
     public static func swiftTool(
         outputStream: OutputByteStream = stdoutStream,
         logLevel: Basics.Diagnostic.Severity = .warning
-    ) -> ObservabilitySystem {
-        .init(SwiftToolObservabilityHandler(outputStream: stdoutStream, logLevel: logLevel))
+    ) -> (ObservabilitySystem, SwiftToolObservabilityHandler) {
```
MaxDesiatov, line 201: This change is needed to allow getting a reference to `SwiftToolObservabilityHandler` and calling `wait` on it.

tomerd: I think this is fine, but at some point we should consider if these new command should become aligned with the other commands so they dont need this kind of external access

MaxDesiatov: I found a better way to address that within this PR. Turns out this `static func swiftTool` was used just once in `DestinationCommand`, it made more sense to move its implementation over there, so there's no need for this tuple-returning function anymore.

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

PR #6266, opened 2023-03-11
add logging to registry flows

tomerd:
motivation: easier debugging

changes:
* add info logging for actions that may be interesting to users
* add debug logging to assist troubleshooting

### DIFF
--- a/Sources/Basics/AuthorizationProvider.swift
+++ b/Sources/Basics/AuthorizationProvider.swift
@@ -225,7 +225,7 @@ public class KeychainAuthorizationProvider: AuthorizationProvider, Authorization
         } catch {
             switch error {
             case AuthorizationProviderError.notFound:
-                self.observabilityScope.emit(info: "No credentials found for server \(server) in keychain")
+                self.observabilityScope.emit(debug: "No credentials found for server \(server) in keychain")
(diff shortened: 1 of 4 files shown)


tomerd:
@swift-ci smoke test

tomerd:
cc @yim-lee

Sources/Basics/AuthorizationProvider.swift
```diff
@@ -225,7 +225,7 @@ public class KeychainAuthorizationProvider: AuthorizationProvider, Authorization
         } catch {
             switch error {
             case AuthorizationProviderError.notFound:
-                self.observabilityScope.emit(info: "No credentials found for server \(server) in keychain")
```
yim-lee, line 228: Why delete this?

tomerd: it is very noisy in the log, and I did not think it provided huge amount of value. we can also tune down to debug, or write code that caches negative results so it's called less often. opinions?

yim-lee: I see. Maybe debug? If that's still too noisy then let's remove it.

tomerd:
@swift-ci smoke test

yim-lee (approved):

PR #6271, opened 2023-03-14
The warning "dependency '<name>' is not used by any target" should have package metadata

abertelrud:
### Motivation:

Like other diagnostics, the warning "dependency '\<name\>' is not used by any target" should be associated with a package (by specifying the package in its metadata). This allows it to show up in the right place in a package resolution log in IDEs that show such things.

### Modifications:

- set the metadata on the diagnostics scope
- extend a unit test to check this metadata

rdar://106678793


abertelrud:
@swift-ci please smoke test

tomerd (approved):

tomerd:
nice

PR #6273, opened 2023-03-15
Only write .recognized signing entity to storage

yim-lee:
Motivation:
Signing entity TOFU is really meant for identities we can "trust"--i.e., if they are recognized.

Modifications:
Strictly speaking writing only recognized identities to storage is business logic, so that logic was kept out of `PackageSigningEntityStorage` in previous implementation. However, with the added APIs for writing to storage, the best way to ensure only recognized identities get saved is having the check in the storage implementation.


yim-lee:
@swift-ci please smoke test

neonichu (approved):

PR #6276, opened 2023-03-15
Add build setting for enabling C++ interoperability

egorzhdan:
### Motivation:

When importing C language dependencies, the Swift compiler internally runs Clang in C/Objective-C mode by default.
If some of the imported clang modules are actually written in C++, that causes build errors since C++ syntax might be invalid in C/Objective-C. To resolve that, Swift allows the user to explicitly enable C++/Objective-C++ language mode (along with the rest of the Swift/C++ interoperability logic) for clang via a compiler flag. The reason we cannot enable C++/Objective-C++ mode by default is because it changes the way some existing C/Objective-C APIs are imported: for instance, a library header might provide a different definition of a symbol under `#ifdef __cplusplus`.

The user should be able to enable Swift/C++ interoperability for a given Swift target without using unsafe flags.

### Modifications:

This adds a build setting to the package manifest that enables Swift/C++ Interoperability for a given Swift target:
```
.interoperabilityMode(.Cxx, version: "swift-5.9")
```

The implementation relies on the new Swift driver flag for versioned C++ interop (see https://github.com/apple/swift/pull/64088).

### DIFF
--- a/Sources/PackageLoading/ManifestJSONParser.swift
+++ b/Sources/PackageLoading/ManifestJSONParser.swift
@@ -675,6 +675,23 @@ extension TargetBuildSettingDescription.Kind {
                 throw InternalError("invalid (empty) build settings value")
             }
             return .linkedFramework(value)
+        case "interoperabilityMode":
+            guard let rawLang = values.first else {
+                throw InternalError("invalid (empty) build settings value")
+            }
+            guard let lang = TargetBuildSettingDescription.InteroperabilityMode(rawValue: rawLang) else {
+                throw InternalError("unknown interoperability mode: \(rawLang)")
+            }
+            if values.count > 2 {
+                throw InternalError("invalid build settings value")
+            }
+            let version: String?
+            if values.count == 2 {
+                version = values[1]
+            } else {
+                version = nil
+            }
+            return .interoperabilityMode(lang, version)
         case "enableUpcomingFeature":
             guard let value = values.first else {
                 throw InternalError("invalid (empty) build settings value")
--- a/Sources/PackageLoading/PackageBuilder.swift
+++ b/Sources/PackageLoading/PackageBuilder.swift
@@ -977,6 +977,31 @@ public final class PackageBuilder {
                     decl = .LINK_FRAMEWORKS
                 }
 
+            case .interoperabilityMode(let lang, let version):
+                switch setting.tool {
+                case .c, .cxx, .linker:
+                    throw InternalError("only Swift supports interoperability")
+
+                case .swift:
+                    decl = .OTHER_SWIFT_FLAGS
+                }
+
+                if lang == .Cxx {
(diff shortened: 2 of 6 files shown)


egorzhdan:
@swift-ci please test

neonichu (approved):

Sources/PackageLoading/ManifestJSONParser.swift
```diff
@@ -675,6 +675,23 @@ extension TargetBuildSettingDescription.Kind {
                 throw InternalError("invalid (empty) build settings value")
             }
             return .linkedFramework(value)
+        case "interoperabilityMode":
+            guard let rawLang = values.first else {
+                throw InternalError("invalid (empty) build settings value")
+            }
+            guard let lang = TargetBuildSettingDescription.InteroperabilityMode(rawValue: rawLang) else {
+                throw InternalError("unknown interoperability mode")
```
MaxDesiatov, line 683: I personally would prefer the cause of the error to be surfaced in the error text for easier troubleshooting.
```suggestion
                throw InternalError("unknown interoperability mode: \(rawLang)")
```

Sources/PackageLoading/PackageBuilder.swift
```diff
@@ -977,6 +977,24 @@ public final class PackageBuilder {
...
+                case .c, .cxx, .linker:
+                    throw InternalError("only Swift supports interoperability")
+
+                case .swift:
+                    decl = .OTHER_SWIFT_FLAGS
+                }
+
+                if lang == .Cxx {
+                    guard let version else {
+                        throw InternalError("C++ interoperability requires a version")
```
MaxDesiatov, line 991: Could this error string be expanded, what version is it referring to? Is that a version of Swift or C++ standard?

Sources/PackageLoading/ManifestJSONParser.swift
egorzhdan: That makes sense, thanks!

Sources/PackageLoading/PackageBuilder.swift
MaxDesiatov: > This is the compatibility version of Swift/C++ interop, it is meant to preserve source compatibility for user projects while interop is evolving. At the moment the only supported interop version is swift-5.9 which is aligned with the version of Swift itself, but this might not always be the case in the future.

Could this paragraph be added as a doc comment or just plain comment in the code to provide context right there, which could be helpful in the future when coming back to it?

egorzhdan:
@swift-ci please smoke test

Sources/PackageLoading/PackageBuilder.swift
MaxDesiatov: Perfect, thanks!

MaxDesiatov (approved):

PR #6277, opened 2023-03-15
allow users to set a default registry without a configuration file

tomerd:
motication: make it easier to use a registry

changes:
* add new --default-registry-url that allows using a registry without a registry.json configuraiton file
* amend registry configuraiton to use the new flag if set an no default regsitry set in the configuraiton file
* add test


Sources/Workspace/Workspace+Configuration.swift
```diff
@@ -686,13 +731,16 @@ public struct WorkspaceConfiguration {
...
-    
+
     ///  Signing entity checking mode. Defaults to warn.
     public var signingEntityCheckingMode: CheckingMode
 
     ///  Attempt to transform source control based dependencies to registry ones
     public var sourceControlToRegistryDependencyTransformation: SourceControlToRegistryDependencyTransformation
 
+    /// URL of the implicitly configured, default regsitry
+    public var defaultRegistry: Registry?
```
tomerd, line 742: the actual change, the rest is formatting

neonichu (approved):

PR #6279, opened 2023-03-15
print compiler output when precompiling plugins

tomerd:
motivation: surface any plugins compiler errors to users

changes:
* always print compiler output when failing to compile plugins
* restore testPluginCompilationBeforeBuilding test

### DIFF
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -335,7 +335,7 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
             }
             func didCompilePlugin(result: PluginCompilationResult) {
                 if !result.description.isEmpty {
-                    self.buildSystemDelegate?.preparationStepHadOutput(preparationStepName, output: result.description)
+                    self.buildSystemDelegate?.preparationStepHadOutput(preparationStepName, output: result.description, verboseOnly: result.succeeded)
                 }
                 self.buildSystemDelegate?.preparationStepFinished(preparationStepName, result: (result.succeeded ? .succeeded : .failed))
             }
@@ -343,7 +343,7 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
                 // Historically we have emitted log info about cached plugins that are used. We should reconsider whether this is the right thing to do.
                 self.buildSystemDelegate?.preparationStepStarted(preparationStepName)
                 if !cachedResult.description.isEmpty {
-                    self.buildSystemDelegate?.preparationStepHadOutput(preparationStepName, output: cachedResult.description)
+                    self.buildSystemDelegate?.preparationStepHadOutput(preparationStepName, output: cachedResult.description, verboseOnly: true)
                 }
                 self.buildSystemDelegate?.preparationStepFinished(preparationStepName, result: (cachedResult.succeeded ? .succeeded : .failed))
             }
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -334,16 +334,29 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
                 self.buildSystemDelegate?.preparationStepStarted(preparationStepName)
             }
             func didCompilePlugin(result: PluginCompilationResult) {
-                if !result.description.isEmpty {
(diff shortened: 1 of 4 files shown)


Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
```diff
@@ -757,10 +757,10 @@ final class BuildOperationBuildSystemDelegateHandler: LLBuildBuildSystemDelegate
     }
 
     /// Invoked when an action taken before building emits output.
-    func preparationStepHadOutput(_ name: String, output: String) {
+    func preparationStepHadOutput(_ name: String, output: String, verboseOnly: Bool) {
```
MaxDesiatov, line 760: The purpose of this new parameter is a bit unclear at a glance from its name. Maybe it could be renamed or the doc comment on this function expanded to clarify this?

Sources/Build/BuildOperation.swift
```diff
@@ -334,16 +334,30 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
                 self.buildSystemDelegate?.preparationStepStarted(preparationStepName)
             }
             func didCompilePlugin(result: PluginCompilationResult) {
-                if !result.description.isEmpty {
-                    self.buildSystemDelegate?.preparationStepHadOutput(preparationStepName, output: result.description)
+                self.buildSystemDelegate?.preparationStepHadOutput(
+                    preparationStepName,
+                    output: result.commandLine.joined(separator: " "),
```
abertelrud, line 339: Is this what the command lines for `swift build` also show, or do they go through some code that quotes whitespace and escapes special characters, etc?  If the same, then this seems fine, otherwise this code should do the same.

Sources/Build/BuildOperation.swift
```diff
@@ -334,16 +334,30 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
...
+                    )
                 }
                 self.buildSystemDelegate?.preparationStepFinished(preparationStepName, result: (result.succeeded ? .succeeded : .failed))
             }
             func skippedCompilingPlugin(cachedResult: PluginCompilationResult) {
                 // Historically we have emitted log info about cached plugins that are used. We should reconsider whether this is the right thing to do.
                 self.buildSystemDelegate?.preparationStepStarted(preparationStepName)
-                if !cachedResult.description.isEmpty {
-                    self.buildSystemDelegate?.preparationStepHadOutput(preparationStepName, output: cachedResult.description)
+                self.buildSystemDelegate?.preparationStepHadOutput(preparationStepName, output: cachedResult.compilerOutput, verboseOnly: false)
```
abertelrud, line 354: Should this line be removed?  Otherwise this would emit the cached output twice?

abertelrud (changes requested):

Sources/Build/BuildOperation.swift
tomerd: oops that is a copy past error

Sources/Build/BuildOperation.swift
tomerd: affect the build system has a very different setup with BuildSystemCommand that parse llbuild commands. so not fully applicable here. do you have some specific utility in mind that can help?

abertelrud (approved):

tomerd:
@swift-ci smoke test

Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
```diff
@@ -757,10 +757,11 @@ final class BuildOperationBuildSystemDelegateHandler: LLBuildBuildSystemDelegate
     }
 
     /// Invoked when an action taken before building emits output.
-    func preparationStepHadOutput(_ name: String, output: String) {
+    /// when verboseOnly is set to true, the output will only be printed in verbose logging mode
```
MaxDesiatov, line 760: formatting nit, which will make it render nicely in editor previews and tools like DocC
```suggestion
    /// - Parameters:
    ///   - name: name of the preparation step
    ///   - output: output from the preparation step delegated to this handler
    ///   - verboseOnly: when set to `true`, the output will only be printed in verbose logging mode
```

MaxDesiatov (approved):

Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
tomerd: this is an internal function, so not super important?

PR #6280, opened 2023-03-15
Workspace: fix `target_link_libraries`

compnerd:
The target that was supposed to link against Foundation was Workspace. Because we changed a dependency, the transitive closure happened to make this work.


compnerd:
@swift-ci please smoke test

tomerd (approved):

MaxDesiatov (approved):

PR #6290, opened 2023-03-16
Update Swift version to 5.7 in `.swiftformat`

MaxDesiatov:
Since we have `swift-tools-version:5.7` in `Package.swift`, it makes sense to align `.swiftformat` with that version.


MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

PR #6294, opened 2023-03-16
Allow sources anywhere in ./Sources when only one target is present

bitjammer:
After getting some feedback about the added `path` argument for executable packages generated with `swift package init` (#6144), this change allows a target's sources to occupy the entire sources directory when there is only one target in the package. All package types can benefit from this.

When there is more than one target in a package, the existing requirements for target sources still apply.

This change should be compatible with existing layouts as well. If there is only a single target in a package, then sources can of course continue to exist in `./Sources/<target>`.

Amend the `executable` template's generated manifest to not include the `path` argument anymore.

rdar://106829666

### DIFF
--- a/Sources/PackageLoading/PackageBuilder.swift
+++ b/Sources/PackageLoading/PackageBuilder.swift
@@ -30,7 +30,7 @@ public enum ModuleError: Swift.Error {
     case duplicateModule(String, [String])
 
     /// The referenced target could not be found.
-    case moduleNotFound(String, TargetDescription.TargetType)
+    case moduleNotFound(String, TargetDescription.TargetType, shouldSuggestRelaxedSourceDir: Bool)
 
     /// The artifact for the binary target could not be found.
     case artifactNotFound(targetName: String, expectedArtifactName: String)
@@ -87,9 +87,14 @@ extension ModuleError: CustomStringConvertible {
         case .duplicateModule(let name, let packages):
             let packages = packages.joined(separator: "', '")
             return "multiple targets named '\(name)' in: '\(packages)'; consider using the `moduleAliases` parameter in manifest to provide unique names"
-        case .moduleNotFound(let target, let type):
+        case .moduleNotFound(let target, let type, let shouldSuggestRelaxedSourceDir):
             let folderName = (type == .test) ? "Tests" : (type == .plugin) ? "Plugins" : "Sources"
-            return "Source files for target \(target) should be located under '\(folderName)/\(target)', or a custom sources path can be set with the 'path' property in Package.swift"
+            var clauses = ["Source files for target \(target) should be located under '\(folderName)/\(target)'"]
+            if shouldSuggestRelaxedSourceDir {
+                clauses.append("'\(folderName)'")
+            }
+            clauses.append("or a custom sources path can be set with the 'path' property in Package.swift")
+            return clauses.joined(separator: ", ")
         case .artifactNotFound(let targetName, let expectedArtifactName):
             return "binary target '\(targetName)' could not be mapped to an artifact with expected name '\(expectedArtifactName)'"
         case .invalidModuleAlias(let originalName, let newName):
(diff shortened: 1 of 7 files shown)


Sources/PackageLoading/PackageBuilder.swift
```diff
@@ -526,6 +526,12 @@ public final class PackageBuilder {
                 return path
             }
 
+            // If there is only one target defined, it may be allowed to occupy the
+            // entire predefined target directory.
+            if self.manifest.targets.count == 1 {
```
neonichu, line 531: I think we should also check for the tools-version here to avoid people creating packages that are unexpectedly not backwards-compatible.

bitjammer: Sounds good.

tomerd: doesn't this change _force_ it to occupy `./Sources`? I think it should _allow_ it not _force_ it. ie. we should also check if `./Sources/<target>` exist first and only if it does not fall back to `./Sources`

bitjammer: Ah yes, this check should move below checking if `path` exists.

bitjammer: Er, wait. No, it still works. This still happens first:

```
// Return the path if the predefined directory contains it.
if predefinedDir.contents.contains(target.name) {
    return path
}
```

bitjammer: Okay, added logic for this and tests for 5.9 and the default tools version.

bitjammer:
@swift-ci Please smoke test

Sources/PackageLoading/PackageBuilder.swift
tomerd: ah nice!

tomerd (approved):
Love this change

Tests/PackageLoadingTests/PackageBuilderTests.swift
```diff
@@ -998,23 +998,155 @@ class PackageBuilderTests: XCTestCase {
...
             PackageBuilderTester(manifest, in: fs) { _, diagnostics in
-                diagnostics.check(diagnostic: .contains("Source files for target Random should be located under 'Sources/Random'"), severity: .error)
+                diagnostics.check(diagnostic: .contains("Source files for target TargetA should be located under 'Sources/TargetA'" ), severity: .error)
             }
         }
 
+    }
+
+    func testManifestTargetDeclErrors() throws {
+
```
MaxDesiatov, line 1149: formatting nit

```suggestion
```

MaxDesiatov (commented):
Should this also be mentioned in `CHANGELOG.md`?

neonichu:
Do we need a matching behavior for tests?

bitjammer:
Added to the change log.

MaxDesiatov (approved):
Thanks!

MaxDesiatov:
@swift-ci test windows

bitjammer:
@swift-ci Please smoke test

Sources/PackageLoading/PackageBuilder.swift
```diff
@@ -1391,6 +1411,18 @@ public final class PackageBuilder {
         }
         return true
     }
+
+    /// Returns the first suggested predefined source directory for a given target type.
+    public static func suggestedPredefinedSourceDirectory(type: TargetDescription.TargetType) -> String {
+        switch type {
+        case .test:
+            return predefinedTestDirectories[0]
```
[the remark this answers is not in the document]
bitjammer: Yep, static constant.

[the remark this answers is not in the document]
tomerd: add comment?

bitjammer:
@swift-ci Please smoke test

bitjammer:
@swift-ci Please test Windows platform

PR #6295, opened 2023-03-16
ListDestinations: use `print` instead of `emit(info:)` for output

MaxDesiatov:
This message is user-visible, so it shouldn't contain "info: " string prefix on each line that `emit(info:)` provides. Its initial use could've suggested that this is a debugging info message.


neonichu (approved):

MaxDesiatov:
@swift-ci smoke test

PR #6302, opened 2023-03-18
Add caching to OCSP verifier and perform expiry check

yim-lee:
- OCSP check is done through HTTP requests, add caching to reduce network calls.
- OCSP responder might not be able to provide revocation checking beyond certificate expiration, so check that leaf certificate has not expired before making OCSP requests.


yim-lee:
@swift-ci please smoke test

tomerd (approved):
awesome

PR #6304, opened 2023-03-18
build: stop exporting a couple of targets

compnerd:
Reduce the overall exported set of libraries.  These are internal to SPM and do not need to be exposed to consumers.


compnerd:
@swift-ci please smoke test

MaxDesiatov (approved):

PR #6306, opened 2023-03-19
[Macro package init] Use triple-slash documentation comments

DougGregor:
In the generated `stringify` macro declaration, use triple-slash documentation comments rather than double-slash comments, so we get better documentation.


DougGregor:
@swift-ci please smoke test

MaxDesiatov (approved):
Thanks! 👍

PR #6314, opened 2023-03-22
assocaite signature warnings with the dependency that caused them

tomerd:
motivation: better UX for warnings

changes: add metadata with the package information to the warnings emitted when checking pacakge signatures


tomerd:
@swift-ci smoke test

neonichu (approved):

abertelrud (approved):

PR #6318, opened 2023-03-23
Mark types conforming to `FileSystem` as `Sendable`

MaxDesiatov:
None of these types store internal mutable state, so this should be safe. This resolves some of the sendability warnings that we currently have.


MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci smoke test

PR #6323, opened 2023-03-24
Plugin import scanner isn't passing in the library directory, making `canImport` not work

abertelrud:
The Workspace type has a facility for scanning package plugins for imports.  But since it's not passing in the actual plugin directory provided by the caller (whether it's SwiftPM or Xcode), any modules provided by an IDE aren't picked up if they are guarded by `#if canImport()`.  This includes `XcodeProjectPlugin`, which is one of the additional plugin modules provided by Xcode.  The same would be true for other IDEs.

Changes:
- `Workspace.loadPluginImports` now also passes the path of the plugin directory to the import scanning
- extended the `testScanImportsInPluginTargets` unit test to check for this case

rdar://106387043


abertelrud:
cc @neonichu

abertelrud:
@swift-ci please smoke test

neonichu (approved):

PR #6329, opened 2023-03-25
Use new shorthand syntax for unwrapping optionals

sheikhbayazid:
Use new shorthand syntax for unwrapping optional values

### Motivation:

Using new shorthand syntax would make the code easier to read. Since we have the support to add the shorthand syntax for `guard let` and `if let`, it makes the code more readable.

### Modifications:

Instead of using `if let value = value`, used `if let value` same goes for `guard let`.

### Result:

There is no change of any logic. Just improved the better syntax of unwrapping optionals.


MaxDesiatov (approved):
Seems legit, thanks!

MaxDesiatov:
@swift-ci smoke test

PR #6339, opened 2023-03-27
cleanup mirror config command

tomerd:
motivation: the mirrors config command no takes both URLs and registry identities

changes:
* nupdate flags to use "original" and "mirror" so it is not URL specific
* update API to not use URL suffix
* update call sites and tests


tomerd:
@swift-ci smoke test

neonichu (approved):

tomerd:
@swift-ci smoke test windows

PR #6341, opened 2023-03-28
Fix compilation error in Swift 5.7.2

yim-lee:
Motivation:

```
Sources/PackageGraph/PackageGraph+Loading.swift:535:16: error: type 'Pair' does not conform to protocol 'Equatable'
        struct Pair: Hashable {
```

Modification:
Move `Pair` out of `createResolvedPackages` method and it seems happy.


yim-lee:
@swift-ci please smoke test

tomerd (approved):
thanks, we should bring to 5.9 as well

PR #6345, opened 2023-03-28
PackageLoading: anchor relative paths for Windows

compnerd:
We currently cannot use `..` as a header search path for Windows even though it may be contained within the package root.  This is due to the fact that the path itself was unanchored, and a raw `..` would be treated as the drive root which will almost always be outside of the package root.  Construct the absolute path relative to the target root rather than appending, permitting us to form a well-formed path which can be checked for ancestry.

### DIFF
--- a/Sources/PackageLoading/PackageBuilder.swift
+++ b/Sources/PackageLoading/PackageBuilder.swift
@@ -1036,9 +1036,10 @@ public final class PackageBuilder {
                 }
 
                 // Ensure that the search path is contained within the package.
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

neonichu (approved):

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:

compnerd:

PR #6348, opened 2023-03-28
PackageLoading: anchor relative paths for Windows

compnerd:
We currently cannot use `..` as a header search path for Windows even though it may be contained within the package root.  This is due to the fact that the path itself was unanchored, and a raw `..` would be treated as the drive root which will almost always be outside of the package root.  Construct the absolute path relative to the target root rather than appending, permitting us to form a well-formed path which can be checked for ancestry.


compnerd:
@swift-ci please test

compnerd:
CC: @shahmishal - https://ci-external.swift.org/job/swiftpm-PR-windows/1559/console seems that there is some pollution between the various versions?  Do we not do a clean build on Windows?

compnerd:

tomerd (approved):

MaxDesiatov (approved):

tomerd:
@swift-ci smoke test

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:

PR #6354, opened 2023-03-30
change validation to throw instead of pre-condition to avoid crashes

tomerd:
motivation: avoid crashes

changes: throw and internal error instead of pre-condition when validation fails

rdar://98522178


tomerd:
@swift-ci smoke test

MaxDesiatov (approved):

PR #6362, opened 2023-03-31
[5.9] Cross-compilation: fix bundles not unpacked on installation

MaxDesiatov:
### Motivation:

`swift experimental-destination install` subcommand works with bundle directories, but fails to unpack bundle archives.

rdar://107367895

### Modifications:

We should pass an instance of an archiver to the installation function to unpack destination bundle archives if needed. Since `withTemporaryDirectory` is `async`, `SwiftDestinationTool` had to be converted to `AsyncParsableCommand` and also gain `@main` attribute for that to work. Additionally, `@main` attribute requires `-parse-as-library` passed in CMake.


MaxDesiatov:
@swift-ci smoke test

neonichu (approved):

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

neonichu (approved):

Sources/CrossCompilationDestinationsTool/RemoveDestination.swift
```diff
@@ -17,8 +17,8 @@ import PackageModel
 
 import struct TSCBasic.AbsolutePath
 
-struct RemoveDestination: DestinationCommand {
-    static let configuration = CommandConfiguration(
+public struct RemoveDestination: DestinationCommand {
```
tomerd, line 20: curious why these need to come public?

MaxDesiatov: Clarified in the PR to the `main` branch, this is due to `SwiftDestinationTool` referencing this type moving to a different target, while `RemoveDestination` could stay where it were.

tomerd (approved):

PR #6368, opened 2023-04-02
Basics: add support for `.tar.gz` archives

MaxDesiatov:
### Motivation:

`swift experimental-destination install` subcommand fails when pointed to a destination bundle compressed with `.tar.gz` extension, compressed with `tar` and `gzip`. This archival format is vastly superior to `zip` in the context of cross-compilation destination bundles. In a typical scenario we see that `.tar.gz` archives are at least 4x better than `.zip`, since with `.zip` files are compressed individually, while with `.tar.gz` compression is applied to the whole archive at once.

On macOS `tar` is always available, and we can also assume it's also available on Linux Docker images, since `tar` is used to unpack `.tar.gz` distributions of Swift itself. Since `ZipArchiver` uses `.tar.exe` on Windows, we make the same assumption that command is available, as we did for `ZipArchiver`.

### Modifications:

Added new `TarArchiver` class, `TarArchiverTests`, and corresponding test files.

### Result:

This is technically NFC, since `TarArchiver` is not called from anywhere yet.


MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

neonichu (approved):

Sources/Basics/Archiver+Tar.swift
```diff
@@ -0,0 +1,147 @@
...
+
+import class Dispatch.DispatchQueue
+import struct Dispatch.DispatchTime
+import struct TSCBasic.AbsolutePath
+import protocol TSCBasic.FileSystem
+import struct TSCBasic.FileSystemError
+import class TSCBasic.Process
+
+/// An `Archiver` that handles Tar archives using the command-line `tar` tool.
+public final class TarArchiver: Archiver {
```
tomerd, line 21: I dont think this should be a reference type. It holds no state

Sources/Basics/Archiver+Tar.swift
```diff
@@ -0,0 +1,147 @@
...
+            guard self.fileSystem.isDirectory(destinationPath) else {
+                throw FileSystemError(.notDirectory, destinationPath)
+            }
+
+            let process = TSCBasic.Process(
+                arguments: [self.tarCommand, "xzf", archivePath.pathString, "-C", destinationPath.pathString]
+            )
+
+            guard let registrationKey = self.cancellator.register(process) else {
+                throw StringError("cancellation")
```
tomerd, line 68: nicer error message?

Sources/Basics/Archiver+Tar.swift
```diff
@@ -0,0 +1,147 @@
...
+            guard self.fileSystem.exists(archivePath) else {
+                throw FileSystemError(.noEntry, archivePath)
+            }
+
+            guard self.fileSystem.isDirectory(destinationPath) else {
+                throw FileSystemError(.notDirectory, destinationPath)
+            }
+
+            let process = TSCBasic.Process(
+                arguments: [self.tarCommand, "xzf", archivePath.pathString, "-C", destinationPath.pathString]
```
tomerd, line 64: true win all platforms?

Sources/Basics/Archiver+Tar.swift
```diff
@@ -0,0 +1,147 @@
...
+    ) {
+        do {
+            guard self.fileSystem.isDirectory(directory) else {
+                throw FileSystemError(.notDirectory, directory)
+            }
+
+            let process = TSCBasic.Process(
+                arguments: [self.tarCommand, "acf", destinationPath.pathString, directory.basename],
+                workingDirectory: directory.parentDirectory
+            )
```
tomerd, line 99: true on all platforms?

Sources/Basics/Archiver+Tar.swift
MaxDesiatov: This duplicates the error message from `ZipArchiver`, would it be ok to update the error message there as well in the same PR?

Sources/Basics/Archiver+Tar.swift
MaxDesiatov: It holds references to `Cancellator` and `FileSystem`, both of which are reference types. `Cancellator` in particular holds mutable state with a cancellation handlers registry.

Sources/Basics/Archiver+Tar.swift
tomerd: please

Sources/Basics/Archiver+Tar.swift
MaxDesiatov: Yes, as clarified above by Saleem. I also checked macOS manpages that these arguments have the same meaning. AFAIU `tar.exe` on Windows follows the convention.

Sources/Basics/Archiver+Tar.swift
MaxDesiatov: I've added a new `static func` on `CancellationError` that populates its `description` with a more detailed message, now used across both `TarArchiver` and `ZipArchiver`.

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

Sources/Basics/Cancellator.swift
```diff
@@ -112,9 +112,24 @@ public protocol Cancellable {
...
-    public init() {}
+    public init() {
+        self.init(description: "Operation cancelled")
+    }
+
+    private init(description: String) {
+        self.description = description
+    }
+
+    static func failedToRegisterProcess(_ process: TSCBasic.Process) -> Self {
```
tomerd, line 125: nice

Sources/Basics/Archiver+Tar.swift
tomerd: sure, but that does not mean it also needs to be a reference type

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

tomerd (approved):

MaxDesiatov:
@swift-ci test windows

PR #6369, opened 2023-04-03
Basics: support multiple formats with a single archiver

MaxDesiatov:
### Motivation:

In certain function signatures only a single archiver instance can be taken as an argument. We need to support multiple archive formats, such as `.zip` and `.tar.gz`, for example in cross-compilation destination bundles. Instead of updating those functions to take multiple archivers, creating a single type that can support multiple formats would be a better solution.

### Modifications:

Added `MultiFormatArchiver` that can handle multiple formats by delegating to other archivers it was initialized with. Since most uses of `Archiver` are either generic or take an existential, this allows us to pass `MultiFormatArchiver` to functions that previously supported only a single archive format.

rdar://107485048

### Results

This is technically NFC as `MultiFormatArchiver` is not used anywhere yet.


neonichu (approved):

Sources/Basics/MultiFormatArchiver.swift
```diff
@@ -0,0 +1,111 @@
...
+public final class MultiFormatArchiver: Archiver {
+    public var supportedExtensions: Set<String>
+
+    enum Error: Swift.Error {
+        case unknownFormat(String, AbsolutePath)
+        case noFileNameExtension(AbsolutePath)
+
+        var description: String {
+            switch self {
+            case .unknownFormat(let ext, let path):
```
compnerd, line 25: Why not consider using magic? https://github.com/file/file/tree/ec41083645689a787cdd00cb3b5bf578aa79e46c should be possible to use on Windows even.

MaxDesiatov: Can we assume it's installed on Windows? It doesn't seem to be installed on `swift` Docker images for Ubuntu, so adding a dependency on it will require updating `Dockerfile`s in the `swift-docker` repository. This isn't supposed to be an incredibly resilient algorithm, but something that we can still land in 5.9 I hope. So distinguishing by extension should be good enough for now, and we can consider adding support for magic format detection later.

Sources/Basics/Archiver+Zip.swift
```diff
@@ -14,7 +14,7 @@ import TSCBasic
 import Dispatch
 
 /// An `Archiver` that handles ZIP archives using the command-line `zip` and `unzip` tools.
-public struct ZipArchiver: Archiver, Cancellable {
+public final class ZipArchiver: Archiver, Cancellable {
```
tomerd, line 17: ideally this remains a struct

Sources/Basics/MultiFormatArchiver.swift
```diff
@@ -0,0 +1,111 @@
...
+    private func archiver(for archivePath: AbsolutePath) throws -> any Archiver {
+        let filename = archivePath.basename
+
+        // Calculating extension manually, since ``AbsolutePath//extension`` doesn't support multiple extensions,
+        // like `.tar.gz`.
+        guard let firstDot = filename.firstIndex(of: ".") else {
+            throw Error.noFileNameExtension(archivePath)
+        }
+
+        var extensions = String(filename[firstDot ..< filename.endIndex])
```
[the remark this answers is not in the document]
MaxDesiatov: As mentioned in a comment above on the line 53, that `AbsolutePath` API doesn't support stacked extensions, it returns just `gz` for `.tar.gz` archives.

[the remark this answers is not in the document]
MaxDesiatov: I'll expand the comment to mention the behavior explicitly.

Sources/Basics/MultiFormatArchiver.swift
```diff
@@ -0,0 +1,111 @@
...
+
+        self.formatMapping = formatMapping
+        self.supportedExtensions = supportedExtensions
+    }
+
+    private func archiver(for archivePath: AbsolutePath) throws -> any Archiver {
+        let filename = archivePath.basename
+
+        // Calculating extension manually, since ``AbsolutePath//extension`` doesn't support multiple extensions,
+        // like `.tar.gz`.
```
MaxDesiatov, line 54:
```suggestion
        // like `.tar.gz`. It returns just `gz` for `.tar.gz` archives.
```

Sources/Basics/MultiFormatArchiver.swift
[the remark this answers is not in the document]
neonichu: We should probably add an API that can return multiple extensions?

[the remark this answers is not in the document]
tomerd: exactly ^^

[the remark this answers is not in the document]
MaxDesiatov: That makes sense, added in the latest commit

Sources/Basics/MultiFormatArchiver.swift
```diff
@@ -0,0 +1,111 @@
...
+                return "unknown format with extension \(ext) at path `\(path)`"
+            case .noFileNameExtension(let path):
+                return "file at path `\(path)` has no extension to detect archival format from"
+            }
+        }
+    }
+
+    private let formatMapping: [String: any Archiver]
+
+    public init(_ archivers: [any Archiver]) {
```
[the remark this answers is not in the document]
MaxDesiatov: If the archivers used by this new archiver are an implementation detail, wouldn't an enum for archive types be a premature generalization? I've changed it to take only `any FileSystem` and `Cancellator` to initialize instances of `ZipArchiver` and `TarArchiver`, like you suggested. Would that be ok for this initial implementation?

Sources/Basics/Archiver+Zip.swift
MaxDesiatov: Addressed in the base `TarArchiver` PR, both `FileSystem` and `Cancellator` are reference types, and `Cancellator` in particular holds mutable state, which transitively makes `ZipArchiver` stateful.

Sources/Basics/Archiver/UniversalArchiver.swift
```diff
@@ -0,0 +1,106 @@
...
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import struct TSCBasic.AbsolutePath
+import protocol TSCBasic.FileSystem
+
+/// An `Archiver` that handles multiple formats by delegating to other existing archivers each dedicated to its own
+/// format.
+public final class UniversalArchiver: Archiver {
```
tomerd, line 18: this should be a value type

Sources/Basics/Archiver/UniversalArchiver.swift
```diff
@@ -0,0 +1,106 @@
...
+            for ext in archiver.supportedExtensions {
+                formatMapping[ext] = archiver
+            }
+        }
+
+        self.formatMapping = formatMapping
+        self.supportedExtensions = supportedExtensions
+    }
+
+    private func archiver(for archivePath: AbsolutePath) throws -> any Archiver {
```
tomerd, line 57: nice

tomerd (approved):

MaxDesiatov:
@swift-ci smoke test

Sources/Basics/Archiver/UniversalArchiver.swift
```diff
@@ -0,0 +1,106 @@
...
+            switch self {
+            case .unknownFormat(let ext, let path):
+                return "unknown format with extension \(ext) at path `\(path)`"
+            case .noFileNameExtension(let path):
+                return "file at path `\(path)` has no extension to detect archival format from"
+            }
+        }
+    }
+
+    /// A dictionary from a file extension string to an archiver instance that supports this extension.
```
MaxDesiatov, line 36:
```suggestion
    /// A dictionary that maps file extension strings to archiver instances that supports these extensions.
```

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci smoke test macos

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci smoke test macos

PR #6388, opened 2023-04-05
Basics: make `supportedExtensions` in `UniversalArchiver` constant

MaxDesiatov:
This was marked as a mutable property by mistake when implementing `Archiver` protocol conformance. This property is assigned an initial value in `UniversalArchiver/init` and isn't supposed to be ever changed for a given archiver.


MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

PR #6389, opened 2023-04-05
Build: fix duplicate `case` of `switch` in `BuildPlan.swift`

MaxDesiatov:
The way this `switch` is written currently triggers a warning:

```
/Build/BuildPlan.swift:770:22: warning: case is already handled by previous patterns; consider removing it
                case .macro:
```

Since this `case .macro` on line 770 is never reached due to another `case .macro` covering it above on line 737, it seems safe to remove the unused case altogether.


tomerd (approved):
seems okay. @neonichu ptal

MaxDesiatov:
@swift-ci test macos self-hosted

PR #6391, opened 2023-04-05
PackageModelTests: fix warnings in `DestinationBundleTests`

MaxDesiatov:
These tests currently have `await` before non-`async` functions. Let's clean that up to fix the warning:

```
No 'async' operations occur within 'await' expression
```


MaxDesiatov:
@swift-ci smoke test

PR #6396, opened 2023-04-05
Basics: allow reusing cancellator handlers from `SwiftTool`

MaxDesiatov:
### Motivation:

These signal handlers are useful to other tools that don't have access to `SwiftTool`. For example, `DestinationCommand` would benefit from this as it needs to set up its own cancellator for handling archiver processes.

### Modifications:

Moved signal handler setup to a new `installSignalHandlers` function on `Cancellator`. Made captures of `Cancellator` in handler closures weak so that it's not referenced indefinitely and is released when no longer needed.

### Result:

`Cancellator/installSignalHandler` can be used in more places.

### DIFF
--- a/Sources/Basics/Cancellator.swift
+++ b/Sources/Basics/Cancellator.swift
@@ -28,6 +28,7 @@ public class Cancellator: Cancellable {
     private let cancelationQueue = DispatchQueue(label: "org.swift.swiftpm.cancellator", qos: .userInteractive, attributes: .concurrent)
     private let cancelling = ThreadSafeBox<Bool>(false)
 
+    private static let signalHandlerLock = NSLock()
     private static var isSignalHandlerInstalled = false
 
     public init(observabilityScope: ObservabilityScope?) {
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
@swift-ci smoke test

tomerd:
no sure how I feel about this moving to the cancelator. maybe a new static utility?

tomerd:
if we are moving to a utility like this we should also add precondition to make sure it is not called more than once

MaxDesiatov:
> if we are moving to a utility like this we should also add precondition to make sure it is not called more than once

I've added a check for this via `private static var` on `Cancellator` on the latest commit.

> no sure how I feel about this moving to the cancellator. maybe a new static utility?

Such utility would still need to take a instance of a cancellator as an argument, and then is there much difference between taking it as a function argument, or via `self`? The latter seems more obvious at the point of use IMO.

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

tomerd (approved):

Sources/Basics/Cancellator.swift
```diff
@@ -24,10 +28,77 @@ public class Cancellator: Cancellable {
...
+                to: sigaction.__Unnamed_union___sigaction_handler.self
+            )
+            sigaction(SIGINT, &action, nil)
+            kill(getpid(), SIGINT)
+            #endif
+        }
+        interruptSignalSource.resume()
+        #endif
+
+        Self.isSignalHandlerInstalled = true
```
tomerd, line 99: non blocking but this is not thread safe

Sources/CoreCommands/SwiftTool.swift
```diff
@@ -275,6 +275,7 @@ public final class SwiftTool {
     internal init(
         outputStream: OutputByteStream,
         options: GlobalOptions,
+        shouldInstallSignalHandlers: Bool = true,
```
tomerd, line 278: we can consider adding this to global options

Tests/CommandsTests/SwiftToolTests.swift
```diff
@@ -237,6 +237,7 @@ extension SwiftTool {
         return try SwiftTool(
             outputStream: outputStream,
             options: options,
+            shouldInstallSignalHandlers: false,
```
tomerd, line 240: feels like this belongs into options

Sources/CoreCommands/SwiftTool.swift
MaxDesiatov: `GlobalOptions` is dedicated to properties using `@Option` property wrappers passed from ArgumentParser. Would this `ToolWorkspaceConfiguration` type values of which are passed as arguments to this initializer also work?

Sources/Basics/Cancellator.swift
MaxDesiatov: I've added a lock for the whole function body to make sure that there's no race condition when installing these handlers.

Tests/CommandsTests/SwiftToolTests.swift
MaxDesiatov: Moved to `ToolWorkspaceConfiguration` for now per my comment above.

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci smoke test

PR #6397, opened 2023-04-06
sort identities from url lookup to make lookup result consistent

tomerd:
motivation: url lookup results should be consistent

changes: sort identities before picking the one to use


tomerd:
@swift-ci smoke test

yim-lee (approved):

PR #6405, opened 2023-04-10
ManifestSourceGeneration incorrectly writes out `DriverKit` instead of `driverKit`

abertelrud:
The manifest source generation uses the wrong spelling for this enum.  AFAIK there haven't been any reports of this affecting anyone yet — this was found during code inspection while making another change.  The incorrect spelling is for both the platform version declaration and the spelling in the condition.

This change also adds a unit test to check the latest versions of each of the platforms, and another one to check all the supported platform conditions.

rdar://107795051


abertelrud:
@swift-ci please smoke test

neonichu (approved):

PR #6409, opened 2023-04-11
[5.9] On ELF platforms, only add runpaths as needed (#6321)

buttaface:
Cherrypick of #6321

__Explanation:__ The bootstrap script currently adds two relative runpaths for ELF platforms, one for the PackageDescription and other libraries and another for the single `swift-package` executable, but since they're applied on the command line, both runpaths are applied to all binaries. This pull separates out the runpaths and only applies each where needed. It also shifts much of the bootstrap script over to applying build flags based on the build target, not on checking the build host platform in Python.

__Scope:__ Should only affect the SPM build on ELF platforms like linux

__Issue:__ None

__Risk:__ low, as it only affects building SPM for ELF platforms

__Testing:__ Passed all CI, including a trunk toolchain build that worked well

__Reviewer:__ @neonichu

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,6 +15,17 @@
 import PackageDescription
 import class Foundation.ProcessInfo
 
+// When building the toolchain on the CI for ELF platforms, remove the CI's
+// stdlib absolute runpath and add ELF's $ORIGIN relative paths before installing.
+let swiftpmLinkSettings : [LinkerSetting]
+let packageLibraryLinkSettings : [LinkerSetting]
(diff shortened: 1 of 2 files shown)


tomerd (approved):

buttaface:
Windows CI failure when testing Foundation is unrelated.

buttaface:
Thanks for the quick review.

@tomerd, should I submit for 5.8 also?

tomerd:
sure, but it will have to wait until the next release

PR #6410, opened 2023-04-11
Improve Swift Macro template

ahoppen:
- Use 4 spaces instead of 2 to be consistent with the other templates
- Use `BasicMacroExpansionContext(` instead of `BasicMacroExpansionContext.init`
- Add two newlines in the test method to improve readability
- Depend on the latest 509.0.0 prerelease of SwiftSyntax instead of `main`.


ahoppen:
@swift-ci Please smoke test

PR #6429, opened 2023-04-13
[5.9] Use exact SwiftCrypto version to work around self-hosting failure

MaxDesiatov:
There's a regression in Swift Crypto 2.4.1 on Linux that makes our Linux self-hosted jobs to fail.


MaxDesiatov:
@swift-ci smoke test

neonichu (approved):

yim-lee (approved):
Thanks @MaxDesiatov 🙏

PR #6436, opened 2023-04-14
Basics: remove `AsyncFileSystem` as unused

MaxDesiatov:
### Motivation:

Since `FileSystem` is now `Sendable`, we no longer need to keep `AsyncFileSystem`, which served only as a workaround for non-sendability. It wasn't non-blocking under the hood anyway, and a blocking FS I/O API would have to be designed from scratch anyway, not as a wrapper for the existing blocking APIs.

### Modifications:

`AsyncFileSystem` actor declaration and its uses were removed. `FileSystem` can be used directly in our async code instead.

### Result:

Reduced the amount of redundant code, FS code is now easier to understand when `async`/`await` are used.


MaxDesiatov:
@swift-ci smoke test

shahmishal:
@swift-ci smoke test macOS self hosted

shahmishal:
@swift-ci smoke test macOS self hosted

tomerd (approved):

neonichu (approved):

yim-lee:

PR #6446, opened 2023-04-17
Newly created build tool and command plugin templates should have conditional support for XcodeProjectPlugin

abertelrud:
Package plugins can be applied to packages in Xcode without any modification.  But Xcode also has an additional `XcodeProjectPlugin` module that provides extra API for applying plugins to Xcode projects.  Since this is a common use case, it makes sense for newly created plugins to have conditional use of that extra API by default.  If there is support for applying plugins to the project formats of other IDEs in the future, it would make sense to add those to the templates as well.

This support is currently always added but we could consider having options in the API for whether to add it.

Changes:
- extended the template-generated source code
- added to the unit tests

rdar://108166582

### DIFF
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -402,27 +402,56 @@ public final class InitPackage {
             if packageType == .buildToolPlugin {
                 content += """
                 struct \(typeName): BuildToolPlugin {
+                    /// Entry point for creating build commands for targets in Swift packages.
                     func createBuildCommands(context: PluginContext, target: Target) async throws -> [Command] {
-                        // The plugin can choose what parts of the package to process.
+                        // This plugin only runs for package targets that can have source files.
                         guard let sourceFiles = target.sourceModule?.sourceFiles else { return [] }
 
                         // Find the code generator tool to run (replace this with the actual one).
                         let generatorTool = try context.tool(named: "my-code-generator")
 
                         // Construct a build command for each source file with a particular suffix.
-                        return sourceFiles.map(\\.path).compactMap { inputPath in
-                            guard inputPath.extension == "my-input-suffix" else { return .none }
-                            let inputName = inputPath.lastComponent
-                            let outputName = inputPath.stem + ".swift"
-                            let outputPath = context.pluginWorkDirectory.appending(outputName)
-                            return .buildCommand(
-                                displayName: "Generating \\(outputName) from \\(inputName)",
-                                executable: generatorTool.path,
-                                arguments: ["\\(inputPath)", "-o", "\\(outputPath)"],
-                                inputFiles: [inputPath],
-                                outputFiles: [outputPath]
-                            )
+                        return sourceFiles.map(\\.path).compactMap {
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci please smoke test

neonichu (approved):
LGTM.

I think this naturally poses the question why we don't have the simpler `func createBuildCommand(for inputPath: Path, in outputDirectoryPath: Path, with generatorToolPath: Path) -> Command?` as an API, but that shouldn't prevent us from improving the template.

ShimaaElabid:
git-describe

abertelrud:
> I think this naturally poses the question why we don't have the simpler func createBuildCommand(for inputPath: Path, in outputDirectoryPath: Path, with generatorToolPath: Path) -> Command? as an API, but that shouldn't prevent us from improving the template.

Absolutely.  There are a lot of refinements that can be made to the API, and it would still be interesting to see if there could be some kind of über-API that would be generic enough to cover most conceivable IDEs.  That approach didn't seem possible for the general case given the differences between the data models of Xcode and Swift Package Manager, but for the common case of just wanting to create a command for every file of some type, it seems a shame to have to have the Xcode specialization.

Sources/Workspace/InitPackage.swift
```diff
@@ -402,40 +402,85 @@ public final class InitPackage {
...
+                            createBuildCommand(for: $0, in: context.pluginWorkDirectory, with: generatorTool.path)
                         }
                     }
                 }
 
+                #if canImport(XcodeProjectPlugin)
+                import XcodeProjectPlugin
+
+                extension \(typeName): XcodeBuildToolPlugin {
+
```
MaxDesiatov, line 424: formatting nit: this leading declaration newline is not present for the `struct` declaration above, probably would be better to have it consistent across declaration blocks

Sources/Workspace/InitPackage.swift
```diff
@@ -402,40 +402,85 @@ public final class InitPackage {
...
+                        return target.inputFiles.map(\\.path).compactMap {
+                            createBuildCommand(for: $0, in: context.pluginWorkDirectory, with: generatorTool.path)
+                        }
+                    }
+                }
+
+                #endif
+
+                extension \(typeName) {
+
```
MaxDesiatov, line 440: ditto

Sources/Workspace/InitPackage.swift
```diff
@@ -402,40 +402,85 @@ public final class InitPackage {
...
                     func performCommand(context: PluginContext, arguments: [String]) async throws {
                         print("Hello, World!")
                     }
                 }
 
+                #if canImport(XcodeProjectPlugin)
+                import XcodeProjectPlugin
+
+                extension MyCommandPlugin: CommandPlugin {
+
```
MaxDesiatov, line 475: formatting nit, ditto

MaxDesiatov (approved):

abertelrud:
Thanks @MaxDesiatov those newlines were a mistake.  I'll fix that.

abertelrud:
@swift-ci please smoke test

abertelrud:

PR #6449, opened 2023-04-18
Update BUG_REPORT.yml

compnerd:
Prefer separating commands with `;` which is more portable (works across bash, zsh, and PowerShell.

### Motivation:
Try to make the command more platform agnostic, particularly as Windows support improves, the reports are more likely to be cross-platform.

### Modifications:

Change the command chaining to use `;` rather than `&&` which also has the benefit of ensuring that the failure of one does not impact the other.

### Result:

The command can now partially run on Windows under PowerShell.

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
@@ -34,6 +34,6 @@ body:
     required: false 
 - type: textarea
   attributes:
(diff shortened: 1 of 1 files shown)


MaxDesiatov (approved):
Thanks!

compnerd:
@swift-ci please smoke test

neonichu (approved):

PR #6450, opened 2023-04-18
Tests: adjust windows librarian test slightly

compnerd:
Use the unsuffixed version of the tool's spelling.  Although we should accept the suffixed variant (which `clang` does do!) the lookup in SPM currently does not properly handle the path lookup with that.


MaxDesiatov (approved):
👍

compnerd:
@swift-ci please smoke test

neonichu (approved):

tomerd (approved):

PR #6451, opened 2023-04-18
Supplemental fix for XcodeProjectPlugin support in package init template

abertelrud:
This is a supplemental fix for https://github.com/apple/swift-package-manager/pull/6446 that removes the `async` specifier and corrects the protocol conformance of the Xcode command plug-in.  This change should have been part of that PR.

- remove the `async` specifier since the entry point that XcodeProjectPlugin defines isn't async
- correct the protocol conformance of the Xcode command plug-in
- update the unit tests

rdar://108166582


abertelrud:
@swift-ci please smoke test

neonichu (approved):

PR #6456, opened 2023-04-19
[5.9] Newly created build tool and command plugin templates should have conditional support for XcodeProjectPlugin #6446

abertelrud:
This is a SwiftPM 5.9 nomination of #6446 and it's follow-on fix #6451.

### Motivation

Package plugins can be applied to packages in Xcode without any modification. But Xcode also has an additional XcodeProjectPlugin module that provides extra API for applying plugins to Xcode projects. Since this is a common use case, it makes sense for newly created plugins to have conditional use of that extra API by default. If there is support for applying plugins to the project formats of other IDEs in the future, it would make sense to add those to the templates as well.

This support is currently always added but we could consider having options in the API for whether to add it.

### Modifications

* extended the template-generated source code
* added to the unit tests

rdar://108166582

### DIFF
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -402,27 +402,56 @@ public final class InitPackage {
             if packageType == .buildToolPlugin {
                 content += """
                 struct \(typeName): BuildToolPlugin {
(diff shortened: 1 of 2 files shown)


abertelrud:
@swift-ci smoke test

neonichu (approved):

PR #6461, opened 2023-04-20
WorkspaceTests: simplify constant construction

compnerd:
Reuse the base path from a local variable reducing the arcs that we append to reduce the amount of duplication and avoid having multiple sites to modify for changes.


compnerd:
@swift-ci please test

compnerd:
@swift-ci please smoke test

MaxDesiatov (approved):
I hope you don't mind PR title rename, otherwise LGTM 👍

PR #6478, opened 2023-04-25
Disambiguate apple and darwin triples

rauhul:
Fixes a bug where all triples with vendor apple are considered to be os variants of darwin. This change special cases "apple-none" to not match `Triple.isDarwin()`, but all other "apple-*" triples will still be considered to be darwin variant.

Fixes some fallout of making the above change to check for the newly added `Triple.isApple()` instead of `Triple.isDarwin()` to allow the triple "apple-none" to opt into behaviors like dead stripping, using libtool, etc.

As a result of these changes spm no longer appends a macOS version number to the "armv7em-apple-none-macho" triple (e.g. "armv7em-apple-none-macho10.13") removing the work around of passing "-target armv7em-apple-none-macho" via cli args or via a toolset.json.

Additionally spm no longer passes unexpected -rpath args to the linker when building a static binary for "armv7em-apple-none-macho".

### DIFF
--- a/Sources/Basics/Triple.swift
+++ b/Sources/Basics/Triple.swift
@@ -162,12 +162,23 @@ public struct Triple: Encodable, Equatable, Sendable {
         return nil
     }
 
+    public func isApple() -> Bool {
+        vendor == .apple
+    }
+
     public func isAndroid() -> Bool {
         os == .linux && abi == .android
     }
 
     public func isDarwin() -> Bool {
-        vendor == .apple || os == .macOS || os == .darwin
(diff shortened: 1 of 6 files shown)


rauhul:
this needs to be rebased on main once #6438 lands due to the use of `.noneOS`.

tomerd (commented):
very nice. could we add a small unit test to prevent regressions?

rauhul:
@swift-ci please smoke test

rauhul:

tomerd (approved):

PR #6501, opened 2023-04-30
[5.8] Fix linking of bootstrap PackageCollectionsSigning

stephank:
This was causing our bootstrap build to fail in NixOS packaging.

Doesn't appear to be necessary for `release/5.9` or `main`.

### DIFF
--- a/Sources/PackageCollectionsSigning/CMakeLists.txt
+++ b/Sources/PackageCollectionsSigning/CMakeLists.txt
@@ -43,6 +43,6 @@ target_link_libraries(PackageCollectionsSigning PUBLIC
   PackageCollectionsModel
   TSCBasic)
 target_link_libraries(PackageCollectionsSigning PRIVATE
-  PackageCollectionsSigningLibc
-  $<$<PLATFORM_ID:Darwin>:SHELL:-Xlinker -framework -Xlinker Security>)
-
+  PackageCollectionsSigningLibc)
+target_link_options(PackageCollectionsSigning PRIVATE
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci smoke test

compnerd (commented):
Hmm, I'm not 100% certain I understand the change.  Why split up the argument to `target_link_libraries`?  The quoting also seems odd to me, why introduce the additional quoting now?  Can you share how this impacts the NixOS build?  The argument shouldn't propagate there.

stephank:
Sorry, should've explained this more.

I meant Nixpkgs where I said NixOS. This is actually macOS packaging using Nix.

The quoting was copied from here: https://github.com/apple/swift-package-manager/blob/release/5.8/Sources/Basics/CMakeLists.txt#L49-L50

IIRC, adding flags to `target_link_libraries` causes prefixing with `-l`. The quoting and `SHELL:` prefix are necessary for [option de-duplication](https://cmake.org/cmake/help/latest/command/target_link_options.html#option-de-duplication).

tomerd:
@compnerd @stephank what is the next step on this one?

compnerd (approved):
Oh, its splitting out into `options` ... yeah, that seems correct.  Thanks!

compnerd (approved):
Oh, its splitting out into `options` ... yeah, that seems correct.  Thanks!

compnerd:

compnerd:

neonichu (approved):

PR #6502, opened 2023-04-30
Tests: prefer to use `first` rather than indexing

compnerd:
In the case of a failure in the test, we would previously continue to perform the slice operation which will fail in the case of an Asserts runtime due to an out of bounds access.  The use of `first?` here would unwrap or fail, allowing tests to continue execution.


compnerd:
@swift-ci please smoke test

compnerd:
Not sure about the windows failure here:
```
CMake Error: The source "C:/Users/swift-ci/jenkins/workspace/swiftpm-PR-windows/swift-asn1/CMakeLists.txt" does not match the source "C:/Users/swift-ci/jenkins/workspace/swiftpm-PR-windows/swiftpm/CMakeLists.txt" used to generate cache.  Re-run cmake with a different source directory.
```

compnerd:

neonichu (approved):

PR #6503, opened 2023-05-01
Tests: prefer to use `first` rather than indexing

compnerd:
In the case of a failure in the test, we would previously continue to perform the slice operation which will fail in the case of an Asserts runtime due to an out of bounds access.  The use of `first?` here would unwrap or fail, allowing tests to continue execution.

### DIFF
--- a/Tests/PackageLoadingTests/PD_4_2_LoadingTests.swift
+++ b/Tests/PackageLoadingTests/PD_4_2_LoadingTests.swift
@@ -596,17 +596,17 @@ class PackageDescription4_2LoadingTests: PackageDescriptionLoadingTests {
 
             let manifestLoader = ManifestLoader(toolchain: try UserToolchain.default, cacheDir: path, delegate: delegate)
 
-            func check(loader: ManifestLoader, expectCached: Bool) {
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please smoke test

Tests/PackageLoadingTests/PD_4_2_LoadingTests.swift
```diff
@@ -596,17 +596,17 @@ class PackageDescription4_2LoadingTests: PackageDescriptionLoadingTests {
 
             let manifestLoader = ManifestLoader(toolchain: try UserToolchain.default, cacheDir: path, delegate: delegate)
 
-            func check(loader: ManifestLoader, expectCached: Bool) {
+            func check(loader: ManifestLoader, expectCached: Bool) throws {
                 delegate.clear()
                 delegate.prepare(expectParsing: !expectCached)
 
-                let manifest = try! loader.load(
+                let manifest = try XCTUnwrap(loader.load(
```
tomerd, line 603: probably we don't even need XCTUnwrap?

tomerd (approved):

PR #6510, opened 2023-05-02
PackageLoadingTests: ensure that we use the correct path spelling

compnerd:
Tests here verify the paths, which need the native representation.  This ensures that we get the correct spelling of the paths in these tests.

The memory leak on the non-Darwin platforms exists due to the path string not being pushed into an autorelease pool.  The use of `withUnsafeFileSystemRepresentation` ends up mismatching the arc separator resulting in a test failure.


compnerd:
@swift-ci please smoke test

tomerd (approved):

compnerd:
@swift-ci please smoke test

compnerd:

neonichu (approved):

PR #6521, opened 2023-05-09
[5.9] sandbox: UInt8 is too small to hold a TCP or UDP port

euanh:
Cherry pick of #6518

### Motivation:

TCP and UDP port numbers are unsigned 16-bit integers, but SandboxNetworkPermission uses UInt8s so trying to specify anything but a small subset of low port numbers causes an error:

    permissions: [
        .allowNetworkConnections(
            scope: .all(ports: [443, 8080]),
            reason: "This command accesses the network"
        )
    ]

leads to

    error: integer literal '443' overflows when stored into 'Array<UInt8>.ArrayLiteralElement' (aka 'UInt8')
    error: integer literal '8080' overflows when stored into 'Array<UInt8>.ArrayLiteralElement' (aka 'UInt8')

### Modifications:

This commit updates the type definitions and adds some higher-numbered ports to the .allowNetworkConnections test cases.

### Result:

It is possible to specify that a plugin needs access to higher-numbered ports.

### DIFF
--- a/Sources/Basics/Sandbox.swift
+++ b/Sources/Basics/Sandbox.swift
@@ -15,8 +15,8 @@ import TSCBasic
 
 public enum SandboxNetworkPermission: Equatable {
     case none
-    case local(ports: [UInt8])
-    case all(ports: [UInt8])
+    case local(ports: [Int])
+    case all(ports: [Int])
     case docker
     case unixDomainSocket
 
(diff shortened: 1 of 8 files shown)


MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

PR #6544, opened 2023-05-11
Fix `Triple` formatting in error messages

MaxDesiatov:
Currently `Triple` is not `CustomStringConvertible`, which leads to issues when using it in string interpolations. It uses default implementation that prints all of the triple components separately, which isn't user-readable.

For example, when using `swift experimental-sdk configuration show` subcommand on a bundle that has wrong triples, an error message is printed which is badly formatted and likely does not show the root cause of the issue.

rdar://107882041

### Motivation:

Added `CustomStringConvertible` conformance on `Triple`, which uses underlying `tripleString` as a return value.

### Result:

`Triple` values are more human-readable when displayed in user-visible strings.

### DIFF
--- a/Sources/Basics/Triple.swift
+++ b/Sources/Basics/Triple.swift
@@ -18,7 +18,7 @@ import class TSCBasic.Process
 
 /// Triple - Helper class for working with Destination.target values
 ///
-/// Used for parsing values such as x86_64-apple-macosx10.10 into
+/// Used for parsing values such as `x86_64-apple-macosx10.10` into
 /// set of enums. For os/arch/abi based conditions in build plan.
 ///
 /// @see Destination.target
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci smoke test

neonichu (approved):

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

tomerd (approved):

tomerd:
@swift-ci test windows@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

PR #6549, opened 2023-05-12
SourceControl: rename `useBuiltinFSMonitor` to `fsmonitor`

compnerd:
Git 2.36.0 renamed `useBuiltinFSMonitor` to `fsmonitor`.  This configuration setting results in a warning.  Given that Git 2.36.0 is now over a year old, rename the option we use to silence the warning. In the case of an installation of git predating the configuration renaming, we could run into issues with git opening the files preventing subsequent git operations, but in the general case, this should be equivalent.


compnerd:
@swift-ci please test

neonichu (approved):

MaxDesiatov:
@swift-ci smoke test

PR #6553, opened 2023-05-13
SourceControl: attempt to enable long path support for Windows

compnerd:
Windows paths are limited to 261 characters by default.  Git provides the `core.longpaths` option to use the extended path support.  This should make SPM more reliable on Windows.


compnerd:
@swift-ci please smoke test

tomerd (approved):

PR #6555, opened 2023-05-15
Documentation fixes for `PluginNetworkPermissionScope`.

zkiraly:
Documentation fixes for `PluginNetworkPermissionScope`.

### Motivation:

There were two issues with the documentation for `PluginNetworkPermissionScope`:
- It was not curated, and this caused the symbol to show up at the top level of the `PackageDescription` documentation. This was not desired.
- The abstract for the symbol was not on its own line in the DocC comments. This caused the abstract to have multiple sentences, which goes against style.

### Modifications:

- The `PluginNetworkPermissionScope` symbol was curated under the `PluginPermission` symbol.
- The DocC documentation for `PluginNetworkPermissionScope` was changed to provide a single sentence abstract on its own line, and one additional sentence on a separate line.

### Result:

The `PluginNetworkPermissionScope` documentation is now curated in the correct place, and the abstract looks correct.

### DIFF
--- a/Sources/PackageDescription/PackageDescription.docc/Curation/PluginPermission.md
+++ b/Sources/PackageDescription/PackageDescription.docc/Curation/PluginPermission.md
@@ -6,3 +6,7 @@
 
 - ``allowNetworkConnections(scope:reason:)``
 - ``writeToPackageDirectory(reason:)``
+
+### Allow network connection
+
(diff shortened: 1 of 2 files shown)


zkiraly:
@swift-ci please smoke test.

neonichu (approved):

PR #6569, opened 2023-05-17
Delete macOS triple hardcode

aleksproger:
On macOS hostTriple hardcoded as `x86_64-apple-macosx` which may not reflect the real host triple. As far as I understand the latest Xcode version are shipped with `-print-target-info` compatible compiler which allows to get rid of this `FIXME`

### Motivation:
Xcode ships with necessary `-print-target-info` option
Cleanup of small FIXME and misleading hardcoded value

### Modifications:
Make a call to swift compiler on macOS to retrieve the triple information

### Result:
macOS triple now reflects real value received from swift compiler

### DIFF
--- a/Sources/Basics/Triple.swift
+++ b/Sources/Basics/Triple.swift
@@ -207,8 +207,6 @@ public struct Triple: Encodable, Equatable, Sendable {
         return String(self.tripleString.dropLast(self.osVersion?.count ?? 0)) + version
     }
 
-    public static let macOS = try! Triple("x86_64-apple-macosx")
-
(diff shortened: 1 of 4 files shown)


neonichu (approved):

MaxDesiatov (approved):

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

aleksproger:
@MaxDesiatov seems like you may merge it now

PR #6579, opened 2023-05-18
SwiftSDKBundle.swift: update strings to use "Swift SDK" naming

MaxDesiatov:
Some of user-visible strings refer to "cross-compilation destinations", while those are called "Swift SDKs" per the updated SE proposal. Let's update the strings to match the proposal.


MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

PR #6581, opened 2023-05-18
BuildPlan: infer `-lc++` based on run-time triple

MaxDesiatov:
Currently, when cross-compiling from Darwin to platforms that don't expect `-lc++` by default (e.g. most Linux distributions), the build will fail because `-lc++` is always passed. When should check the destination triple and not the host triple in the build plan code paths.

This leads to confusing build errors when cross-compiling more complex projects like Vapor.

rdar://107487090

### Modifications:

Modified `BuildPlan.swift` to check the value of the `triple` property instead of `hostTriple`. Added a test that verifies this cross-compilation case.

### Result:

Projects that contain C++ can now be cross-compiled from macOS to Linux.

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -632,9 +632,9 @@ public class BuildPlan: SPMBuildCore.BuildPlan {
         // Note: This will come from build settings in future.
         for target in dependencies.staticTargets {
             if case let target as ClangTarget = target.underlyingTarget, target.isCXX {
-                if buildParameters.hostTriple.isDarwin() {
+                if buildParameters.triple.isDarwin() {
                     buildProduct.additionalFlags += ["-lc++"]
-                } else if buildParameters.hostTriple.isWindows() {
+                } else if buildParameters.triple.isWindows() {
                     // Don't link any C++ library.
                 } else {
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
@swift-ci smoke test

compnerd (commented):
Honestly, I think that we should drop this entirely.  The swift-driver has changed the behaviour for Darwin - it now uses clang as the linker driver.  This is now something which we can control properly by using the driver.  As long as we pass `-enable-cxx-interop` to the swift driver, it will correctly handle this - using `clang++` to drive the linker which will add the C++ standard library to the linker invocation at the appropriate time.  It also means that `-Xclang-linker -stdlib=` will do the proper thing.

MaxDesiatov:
> As long as we pass `-enable-cxx-interop` to the swift driver, it will correctly handle this

This is unrelated to `-enable-cxx-interop` support though. There are plenty of projects that don't use C++ interop, but fail to cross-compile because `-lc++` is always passed when building on Darwin platforms. I'm happy to discuss the removal of this logic outside of the scope of this PR, but within it I'm only fixing the existing cross-compilation behavior without changing it.

neonichu (approved):

PR #6627, opened 2023-05-30
Use swift-driver triple

rauhul:
Removes and replaces the custom LLVM Triple type from the Basics module with the implementation from SwiftDriver. This change allows SwiftPM to understand all triples that swift-driver already does and resolves some issues surrounding incorrect triple parsing. For example the Basics.Triple was not able to parse 'armv7em-apple-none-eabihf-macho' where as SwiftDriver.Triple can.


Sources/PackageModel/UserToolchain.swift
```diff
@@ -382,11 +382,11 @@ public final class UserToolchain: Toolchain {
                             // the merged swiftmodules.  XCTest followed suit.
                             "-I",
                             AbsolutePath(
-                                validating: "usr/lib/swift/windows/\(triple.arch)",
+                                validating: "usr/lib/swift/windows/\(triple.arch?.rawValue ?? "unknown")",
```
compnerd, line 385: I'm somewhat uncomfortable with this.  The `unknown` here is confusing.  Would we be able to replace this with something like `<<NULL>>` or something that is explicitly bad and obviously so?

Tests/BasicsTests/TripleTests.swift
```diff
@@ -76,9 +76,51 @@ final class TripleTests: XCTestCase {
         XCTAssertTriple("i686-pc-windows-gnu", isApple: false, isDarwin: false)
         XCTAssertTriple("i686-pc-windows-cygnus", isApple: false, isDarwin: false)
     }
-
+
```
compnerd, line 79: Nit: bleeding whitespace

MaxDesiatov:
Is there a reason we're not using Driver's `Triple` type instead of copying it here?

rauhul:
@swift-ci please smoke test

MaxDesiatov:
I've previously copied this `Triple` implementation from TSC since we wanted to deprecate the TSC version, but it seems like the duplication in both TSC and Driver is historic. As SwiftPM already uses Driver as a dependency, I don't see why we can't use Driver's `Triple` type directly.

MaxDesiatov:
@swift-ci test windows

Tests/BasicsTests/TripleTests.swift
```diff
@@ -81,4 +81,46 @@ final class TripleTests: XCTestCase {
...
+        XCTAssertTriple("i686-apple-macos10.12.0", forPlatformVersion: "", is: "i686-apple-macos")
+        XCTAssertTriple("i686-apple-macos10.12.0", forPlatformVersion: "13.0", is: "i686-apple-macos13.0")
+
+        XCTAssertTriple("riscv64-apple-darwin", forPlatformVersion: "", is: "riscv64-apple-darwin")
+        XCTAssertTriple("riscv64-apple-darwin", forPlatformVersion: "22", is: "riscv64-apple-darwin22")
+
+        XCTAssertTriple("mips-apple-darwin19", forPlatformVersion: "", is: "mips-apple-darwin")
+        XCTAssertTriple("mips-apple-darwin19", forPlatformVersion: "22", is: "mips-apple-darwin22")
+    }
```
MaxDesiatov, line 125: Could we also test for something like `armv7em-apple-none-eabihf-macho` you mentioned in the PR description, or for whatever other triple we'd like to support with more than three components?

rauhul: done!

rauhul:
@swift-ci please smoke test

rauhul:

Sources/Basics/Triple+Basics.swift
```diff
@@ -0,0 +1,198 @@
...
+//
+// Copyright (c) 2014-2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import enum TSCBasic.JSON
```
MaxDesiatov, line 13: One last request, but otherwise this LGTM: can we have a comment in either of these files or in README.md of this directory that these files are vendored from Swift Driver, maybe also with an explanation why? Without it I don't think we make it obvious why the directory is called `Vendor` in the first place without opening this PR and reading this discussion to get the context.

rauhul: Yeah for sure, I'll push that diff later tonight/tomorrow

Sources/PackageModel/UserToolchain.swift
tomerd: should this throw in case it is null?

Sources/PackageModel/UserToolchain.swift
```diff
@@ -716,32 +716,32 @@ public final class UserToolchain: Toolchain {
                 // (~5.7), we always had a singular installed SDK.  Prefer the
                 // new variant which has an architecture subdirectory in `bin`
                 // if available.
-                switch triple.arch {
-                case .x86_64, .x86_64h:
+                switch triple.archName {
```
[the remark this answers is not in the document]
tomerd: switching on the string feels like a regression from a type safety pov. is it necessary?

Sources/SPMBuildCore/BinaryTarget+Extensions.swift
```diff
@@ -43,8 +43,8 @@ extension BinaryTarget {
         // Filter the libraries that are relevant to the triple.
         // FIXME: this filter needs to become more sophisticated
         guard let library = metadata.libraries.first(where: {
-            $0.platform == triple.os.asXCFrameworkPlatformString &&
-                $0.architectures.contains(triple.arch.rawValue)
+            $0.platform == triple.os?.asXCFrameworkPlatformString &&
+                $0.architectures.contains(triple.archName)
```
tomerd, line 47: can architectures be made more type safe instead of string array?

Sources/Basics/Triple.swift
```diff
@@ -10,173 +10,36 @@
...
-    fileprivate static func parseVersion(_ string: String) -> String? {
-        let candidate = String(string.drop(while: { $0.isLetter }))
-        if candidate != string && !candidate.isEmpty {
-            return candidate
-        }
-
-        return nil
-    }
+extension Triple {
+    public static let macOS = try! Self("x86_64-apple-macosx")
```
tomerd: does Triple need a non-throwing constructer that takes a static string (can precondition)

Sources/Basics/Triple.swift
```diff
@@ -10,173 +10,36 @@
...
-        self.abiVersion = abiVersion
-    }
-
-    fileprivate static func parseOS(_ string: String) -> OS? {
-        var candidates = OS.allCases.map { (name: $0.rawValue, value: $0) }
-        // LLVM target triples support this alternate spelling as well.
-        candidates.append((name: "macos", value: .macOS))
-        return candidates.first(where: { string.hasPrefix($0.name) })?.value
+extension Triple {
+    public init(_ description: String) throws {
```
tomerd: where is this used? worried about the public extension

Sources/Basics/Triple.swift
```diff
@@ -10,173 +10,36 @@
 //
 //===----------------------------------------------------------------------===//
 
-import protocol Foundation.CustomNSError
-import var Foundation.NSLocalizedDescriptionKey
-
 import enum TSCBasic.JSON
 import class TSCBasic.Process
+import struct SwiftDriver.Triple
```
tomerd: This file should probably be renamed to Triple+Extension to match style of repo

tomerd: seems like some of these imports can be removed?

Package.swift
```diff
@@ -152,6 +152,7 @@ let package = Package(
                 .product(name: "DequeModule", package: "swift-collections"),
                 .product(name: "OrderedCollections", package: "swift-collections"),
                 .product(name: "SwiftToolsSupport-auto", package: "swift-tools-support-core"),
+                .product(name: "SwiftDriver", package: "swift-driver"),
```
tomerd: hmm I dont love adding this to Basics. I think we are going that only for the extension which are a bit questionable imho. Maybe we can limit the dependency to the modules that actually use it and find a better home (closer to call site) for the extensions

WowbaggersLiquidLunch:
> Is there a reason we're not using Driver's `Triple` type directly instead of copying it here?

> We don't want to introduce a dependency on Driver in the "data model" portion of SwiftPM since that will cause issues for the Xcode integration.

What if `Triple` is abstracted into its own module?

tomerd:
I think creating a new library that has some of these very basic models (eg semantic version, triple) is called for. TSC was suppose to be that but became a grab bag of "stuff" that is impossible to version properly and as such lost utility and will be deprecated. while we are working to deprecate TSC, this specific solution seems like an okay compromise

Sources/PackageModel/UserToolchain.swift
compnerd: I think that having it throw would be a much safer option.  We need to know the architecture to be able to build.

Sources/SPMBuildCore/BinaryTarget+Extensions.swift
rauhul: I have absolutely no idea how this XCFramework code works and I'd like to avoid touching it if possible in this PR.
```
struct XCFrameworkMetadata.Library {
    let libraryIdentifier: String
    let libraryPath: String
    let headersPath: String?
    let platform: String
    let architectures: [String]
}
```

Sources/Basics/Triple.swift
rauhul: I think this is stale

Sources/PackageModel/UserToolchain.swift
rauhul: Resolved this by using `archName` which is non-optional

Sources/PackageModel/UserToolchain.swift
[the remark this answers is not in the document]
rauhul: Changed this back to switch on arch, the coverage of the included cases is increased as a result though.

Sources/Basics/Triple.swift
rauhul: Probably, but I want to avoid additional changes API changes in this PR. (The existing triple uses the above API shape)

Sources/Basics/Triple.swift
rauhul: I added this to match the current Basics.Triple API

Sources/Basics/Triple.swift
rauhul: Fixed!

tomerd:
@rauhul is this ready to be merged?

rauhul:
@tomerd I just pushed fixes to the review feedback, do you taking another look?

Sources/Basics/Triple+Basics.swift
rauhul: done!

rauhul:
Likely in a follow on PR `Sources/Basics/Triple+Basics.swift` should be slimmed down as much as possible, upstreaming necessary content to swift-driver

rauhul:
@swift-ci please smoke test

rauhul:

rauhul:
@swift-ci please test windows

rauhul:
@swift-ci please smoke test

rauhul:
@swift-ci please test windows

rauhul:
@swift-ci test Windows

rauhul:
@swift-ci test windows

PR #6630, opened 2023-06-01
Added availability info to the DocC catalog.

zkiraly:
Added availability info to the DocC catalog.

### Motivation:

The availability info was missing from the DocC catalog.

### Modifications:

With this change, when documentation is generated, the initial availability of `PackageDescription` will be listed on the documentation entry page.

### Result:

The `Info.plist` file in the `PackageDescription` DocC catalog now has the initial availability information.


zkiraly:
@swift-ci please smoke test.

neonichu (changes requested):
Don't think this is correct, `PackageDescription` is not part of the OS SDKs, so it doesn't have any availability aligned with them.

tomerd:
@zkiraly should we close this?

PR #6634, opened 2023-06-03
[docs] Requiring System Libraries with .systemLibrary targets

### LINKED ISSUE #6121: Docs talk about making system modules, but doing so results in deprecation warnings
### Description

The [docs](https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#requiring-system-libraries) talk about how to create a `Clibgit2` example, but if you follow that, you'll get a warning like this:

```
warning: 'clibmagic': system packages are deprecated; use system library targets instead
```

However, the docs don’t mention that, and don’t talk about “system libraries” except in the context of system modules.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

_No response_

_No response_

$ swift --version && uname -a
swift-driver version: 1.62.15 Apple Swift version 5.7.2 (swiftlang-5.7.2.135.5 clang-1400.0.29.51)
Darwin EclipseM1.local 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:03:51 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T6000 arm64

natikgadzhi:
Updates the [usage documentation](Documentation/Usage.md) around requiring system libraries.

### Motivation:

I was getting started with Swift package manager, and looking to get a bit more experience using it — and stumbled on #6121, thought I'd play around and try to clean it up. Fixes #6121.

The existing documentation on `main` worked, but following it emitted a build-time warning about wrapping system libraries in a target, instead of it's own package.

## Modifications

I've tried to make things work without warnings, tested the `.systemLibrary` approach, and documented my steps.

- [x] `clibgit` example
- [x] `cjpeg` example

### Result:

No functional changes to the package manager behavior — just the documentation change. After the change, one should be able to follow the docs step by step, and it should work without any build warnings.

### DIFF
--- a/Documentation/Usage.md
+++ b/Documentation/Usage.md
@@ -146,7 +146,7 @@ following files from the libgit2 system-package are of interest:
     /opt/homebrew/lib/libgit2.dylib      # .so on Linux
     /opt/homebrew/include/git2.h
 
-Note that the system library may be located elsewhere on your system, such as
+**Note:** the system library may be located elsewhere on your system, such as
 `/usr/local/`, or `/usr/`. The paths above are valid for Homebrew on macOS.
 
 Now let's define `Clibgit` dependency in your `Package.swift` in the example project:
@@ -187,7 +187,13 @@ let package = Package(
 ```
 
 The `pkgConfig` parameter helps SwiftPM in figuring out the include and library
-search paths for the system library. Note: If you don't want to use the `pkgConfig`
+search paths for the system library. You can look up pkg-config for a package
+by running the following:
+
+    example$ pkg-config --cflags libgit2
+    -I/opt/homebrew/Cellar/libgit2/1.6.4/include
+
+**Note:** If you don't want to use the `pkgConfig`
 parameter you can pass the path of a directory containing the library using the
 `-L` flag in commandline when building your app:
 
@@ -227,6 +233,8 @@ executable:
     git_repository_init_options(version: 0, flags: 0, mode: 0, workdir_path: nil, description: nil, template_path: nil, initial_head: nil, origin_url: nil)
     example$
 
+### Requiring a System Library Without pkg-config
+
 Let’s see another example of using [IJG’s JPEG library](http://www.ijg.org)
 from an executable, which has some caveats.
 
@@ -246,22 +254,16 @@ let jpegData = jpeg_common_struct()
 print(jpegData)
 ```
 
-Install JPEG library using a system packager, e.g, `$ brew install jpeg`
+Install JPEG library using a system packager, e.g, `$ brew install jpeg`.
+`jpeg` is a keg-only formula, meaning it won't be linked to `/opt/homebrew/lib`,
+and you'll have to link it manually at build time.
 
-Create a directory called `CJPEG` next to the `example` directory and
-initialize it as a package that builds a system module:
-
-    example$ cd ..
-    $ mkdir CJPEG
-    $ cd CJPEG
-    CJPEG$ swift package init --type empty
-
-This creates `Package.swift` file in the directory.
-Create a `module.modulemap` file so it consists of the following:
+Just like in the previous example, `mkdir Sources/CJPEG`, and add the
+following `module.modulemap`:
 
     module CJPEG [system] {
         header "shim.h"
-        header "/usr/local/include/jpeglib.h"
+        header "/opt/homebrew/opt/jpeg/include/jpeglib.h"
         link "jpeg"
         export *
     }
@@ -275,23 +277,30 @@ This is because `jpeglib.h` is not a correct module, that is, it does not contai
 the required line `#include <stdio.h>`. Alternatively, you can add `#include <stdio.h>`
 to the top of jpeglib.h to avoid creating the `shim.h` file.
 
-Create a Git repository and tag it:
-
-    CJPEG$ git init
-    CJPEG$ git add .
-    CJPEG$ git commit -m "Initial Commit"
-    CJPEG$ git tag 1.0.0
-
 Now to use the CJPEG package we must declare our dependency in our example
 app’s `Package.swift`:
 
 ```swift
+// swift-tools-version: 5.8
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
 import PackageDescription
 
 let package = Package(
     name: "example",
-    dependencies: [
-        .package(url: "../CJPEG", from: "1.0.0")
+    targets: [
+        // Targets are the basic building blocks of a package, defining a module or a test suite.
+        // Targets can depend on other targets in this package and products from dependencies.
+        .executableTarget(
+            name: "example",
+            dependencies: ["CJPEG"],
+            path: "Sources"
+            ),
+        .systemLibrary(
+            name: "CJPEG",
+            providers: [
+                .brew(["jpeg"])
+            ])
     ]
 )
 ```
@@ -299,7 +308,7 @@ let package = Package(
 Now if we type `swift build` in our example app directory we will create an
 executable:
 
-    example$ swift build -Xlinker -L/usr/local/lib/
+    example$ swift build -Xlinker -L/opt/homebrew/opt/jpeg/lib
     …
     example$ .build/debug/example
     jpeg_common_struct(err: nil, mem: nil, progress: nil, client_data: nil, is_decompressor: 0, global_state: 0)
--- a/Documentation/Usage.md
+++ b/Documentation/Usage.md
@@ -116,9 +116,9 @@ https://github.com/apple/example-package-fisheryates
 
 ## Requiring System Libraries
 
-You can link against system libraries using the package manager. To do so, there
-needs to be a special package for each system library that contains a modulemap
-for that library. Such a wrapper package does not contain any code of its own.
+You can link against system libraries using the package manager. To do so, you'll
+need to add a special `target` of type `.systemLibrary`, and a `module.modulemap`
+for each system library you're using.
 
 Let's see an example of using [libgit2](https://libgit2.github.com) from an
 executable.
@@ -143,35 +143,47 @@ To `import Clibgit`, the package manager requires that the libgit2 library has
 been installed by a system packager (eg. `apt`, `brew`, `yum`, etc.). The
 following files from the libgit2 system-package are of interest:
 
-    /usr/local/lib/libgit2.dylib      # .so on Linux
-    /usr/local/include/git2.h
-
-Swift packages that provide module maps for system libraries are handled
-differently from regular Swift packages.
+    /opt/homebrew/lib/libgit2.dylib      # .so on Linux
+    /opt/homebrew/include/git2.h
 
 Note that the system library may be located elsewhere on your system, such as
-`/usr/` rather than `/usr/local/`.
+`/usr/local/`, or `/usr/`. The paths above are valid for Homebrew on macOS.
 
-Create a directory called `Clibgit` next to the `example` directory and
-initialize it as a package that builds a system module:
-
-    example$ cd ..
-    $ mkdir Clibgit
-    $ cd Clibgit
-    Clibgit$ swift package init --type empty
-
-This creates a `Package.swift` file in the directory.
-Edit `Package.swift` and add `pkgConfig` parameter:
+Now let's define `Clibgit` dependency in your `Package.swift` in the example project:
 
 ```swift
-// swift-tools-version:5.3
+// swift-tools-version: 5.8
+// The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
 
 let package = Package(
-    name: "Clibgit",
-    pkgConfig: "libgit2"
+    name: "example",
+    targets: [
+        // Targets are the basic building blocks of a package, defining a module or a test suite.
+        // Targets can depend on other targets in this package and products from dependencies.
+        .executableTarget(
+            name: "example",
+
+            // example executable requires "Clibgit" target as it's dependency.
+            // It's a systemLibrary target defined below.
+            dependencies: ["Clibgit"],
+            path: "Sources"
+        ),
+
+        // systemLibrary is a special type of build target that wraps a system library
+        // in a target that other targets can require as their depencency.
+        .systemLibrary(
+            name: "Clibgit",
+            pkgConfig: "libgit2",
+            providers: [
+                .brew(["libgit2"]),
+                .apt(["libgit2-dev"])
+            ]
+        )
+    ]
 )
+
 ```
 
 The `pkgConfig` parameter helps SwiftPM in figuring out the include and library
@@ -179,48 +191,32 @@ search paths for the system library. Note: If you don't want to use the `pkgConf
 parameter you can pass the path of a directory containing the library using the
 `-L` flag in commandline when building your app:
 
-    example$ swift build -Xlinker -L/usr/local/lib/
+    example$ swift build -Xlinker -L/opt/homebrew/lib/
 
-Create a `module.modulemap` file so it consists of the following:
+Next, we'll create a directory `Sources/Clibgit` in your `example` projcet, and
+add a `module.modulemap` to it:
 
     module Clibgit [system] {
-      header "/usr/local/include/git2.h"
+      header "/opt/homebrew/include/git2.h"
       link "git2"
       export *
     }
+
+Note that the include path in the module map is absolute, more on that below.
 
 > The convention we hope the community will adopt is to prefix such modules
 > with `C` and to camelcase the modules as per Swift module name conventions.
 > Then the community is free to name another module simply `libgit` which
 > contains more “Swifty” function wrappers around the raw C interface.
 
-Packages are Git repositories, tagged with semantic versions, containing a
-`Package.swift` file at their root. Initializing the package created a
-`Package.swift` file, but to make it a usable package we need to initialize a
-Git repository with at least one version tag:
+The `example` directory structure should look like this now:
 
-    Clibgit$ git init
-    Clibgit$ git add .
-    Clibgit$ git commit -m "Initial Commit"
-    Clibgit$ git tag 1.0.0
-
-Now to use the Clibgit package we must declare our dependency in our example
-app’s `Package.swift`:
-
-```swift
-import PackageDescription
-
-let package = Package(
-    name: "example",
-    dependencies: [
-        .package(url: "../Clibgit", from: "1.0.0")
-    ]
-)
-```
-
-Here we used a relative URL to speed up initial development. If you push your
-module map package to a public repository you must change the above URL
-reference so that it is a full, qualified Git URL.
+    .
+    ├── Package.swift
+    └── Sources
+        ├── Clibgit
+        │   └── module.modulemap
+        └── main.swift
 
 Now if we type `swift build` in our example app directory we will create an
 executable:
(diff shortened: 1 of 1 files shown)


Documentation/Usage.md
```diff
@@ -471,9 +476,9 @@ In case the current Swift version doesn't match any version-specific manifest,
 the package manager will pick the manifest with the most compatible tools
 version. For example, if there are three manifests:
 
-`Package.swift` (tools version 3.0)
```
natikgadzhi, line 474: These three lines are non-intentional, my formatter got trigger-happy. I think that's okay?

Documentation/Usage.md
```diff
@@ -143,84 +143,86 @@ To `import Clibgit`, the package manager requires that the libgit2 library has
 been installed by a system packager (eg. `apt`, `brew`, `yum`, etc.). The
 following files from the libgit2 system-package are of interest:
 
-    /usr/local/lib/libgit2.dylib      # .so on Linux
-    /usr/local/include/git2.h
+    /opt/homebrew/lib/libgit2.dylib      # .so on Linux
+    /opt/homebrew/include/git2.h
```
tomerd, line 147: I would probably leave this as `/usr/local` since not everyone is using homebrew, and change the note to say it may be in `/opt/homebrew` etc

Documentation/Usage.md
```diff
@@ -143,84 +143,86 @@ To `import Clibgit`, the package manager requires that the libgit2 library has
...
 import PackageDescription
 
 let package = Package(
-    name: "Clibgit",
-    pkgConfig: "libgit2"
+    name: "example",
+    targets: [
+        // Targets are the basic building blocks of a package, defining a module or a test suite.
+        // Targets can depend on other targets in this package and products from dependencies.
+        .executableTarget(
```
tomerd, line 165: it may be easier to understand if we do this in two steps, first introduce the "Clibgit" target, then show how to use it in the executable target. in other words, I suggest we split this example into two snippets, and build the knowledge in layers

Documentation/Usage.md
```diff
@@ -143,84 +143,86 @@ To `import Clibgit`, the package manager requires that the libgit2 library has
...
-    example$ swift build -Xlinker -L/usr/local/lib/
+    example$ swift build -Xlinker -L/opt/homebrew/lib/
 
-Create a `module.modulemap` file so it consists of the following:
+Next, we'll create a directory `Sources/Clibgit` in your `example` projcet, and
+add a `module.modulemap` to it:
 
     module Clibgit [system] {
-      header "/usr/local/include/git2.h"
+      header "/opt/homebrew/include/git2.h"
```
tomerd, line 206: ditto re. `/usr/local` with note that homebrew users this may be `/opt/homebrew/`

Documentation/Usage.md
```diff
@@ -143,84 +143,86 @@ To `import Clibgit`, the package manager requires that the libgit2 library has
...
+
 ```
 
 The `pkgConfig` parameter helps SwiftPM in figuring out the include and library
-search paths for the system library. Note: If you don't want to use the `pkgConfig`
+search paths for the system library. You can look up pkg-config for a package
+by running the following:
+
+    example$ pkg-config --cflags libgit2
+    -I/opt/homebrew/Cellar/libgit2/1.6.4/include
```
tomerd, line 194: I am not sure what this means. can you elaborate what this is about?

Documentation/Usage.md
```diff
@@ -250,22 +254,16 @@ let jpegData = jpeg_common_struct()
...
-
-This creates `Package.swift` file in the directory.
-Create a `module.modulemap` file so it consists of the following:
+Just like in the previous example, `mkdir Sources/CJPEG`, and add the
+following `module.modulemap`:
 
     module CJPEG [system] {
         header "shim.h"
-        header "/usr/local/include/jpeglib.h"
+        header "/opt/homebrew/opt/jpeg/include/jpeglib.h"
```
tomerd, line 266: ditto re. `/usr/local` with note that homebrew users this may be `/opt/homebrew/`

Documentation/Usage.md
```diff
@@ -279,31 +277,38 @@ This is because `jpeglib.h` is not a correct module, that is, it does not contai
...
+            ])
     ]
 )
 ```
 
 Now if we type `swift build` in our example app directory we will create an
 executable:
 
-    example$ swift build -Xlinker -L/usr/local/lib/
+    example$ swift build -Xlinker -L/opt/homebrew/opt/jpeg/lib
```
tomerd, line 311: ditto re. `/usr/local` with note that homebrew users this may be `/opt/homebrew/`

Documentation/Usage.md
```diff
@@ -279,31 +277,38 @@ This is because `jpeglib.h` is not a correct module, that is, it does not contai
...
-    CJPEG$ git add .
-    CJPEG$ git commit -m "Initial Commit"
-    CJPEG$ git tag 1.0.0
-
 Now to use the CJPEG package we must declare our dependency in our example
 app’s `Package.swift`:
 
 ```swift
+// swift-tools-version: 5.8
+// The swift-tools-version declares the minimum version of Swift required to build this package.
```
tomerd, line 285: this line is redundant

Documentation/Usage.md
```diff
@@ -279,31 +277,38 @@ This is because `jpeglib.h` is not a correct module, that is, it does not contai
...
+
 import PackageDescription
 
 let package = Package(
     name: "example",
-    dependencies: [
-        .package(url: "../CJPEG", from: "1.0.0")
+    targets: [
+        // Targets are the basic building blocks of a package, defining a module or a test suite.
+        // Targets can depend on other targets in this package and products from dependencies.
```
tomerd, line 293: the comment lines are redundant

Documentation/Usage.md
```diff
@@ -279,31 +277,38 @@ This is because `jpeglib.h` is not a correct module, that is, it does not contai
...
+        .executableTarget(
+            name: "example",
+            dependencies: ["CJPEG"],
+            path: "Sources"
+            ),
+        .systemLibrary(
```
tomerd, line 299: ditto re. splitting the example into two steps

tomerd:
thanks so much  for helping improve this under-documented area @natikgadzhi, left some comments. 

cc @MaxDesiatov @neonichu

natikgadzhi:
Hey @tomerd, thank you for the review! I'll clean these up later tonight. 

As the particular examples get lengthy and contain snippets of code and example commands, I wonder if after this PR, it would make sense to repackage common usage scenarios as docc tutorials? There's [some documentation in docc format here,](https://developer.apple.com/documentation/packagedescription/package) it seems. If there's interest, I would take on that work, and learn a bit more about SwiftPM in the process.

tomerd:
hi @natikgadzhi I think that is indeed worth considering, but we currently do not have the automation in place to publish such. we are definitely planning to address this at some point in the future tho.

Documentation/Usage.md
natikgadzhi: Moving this into the note about system paths, where it makes more sense. My point was that on most systems, for most libraries, running `pkg-config --cflags {library-name}` will get you the `inclide` and `lib` paths.

Documentation/Usage.md
natikgadzhi: Cleaned this up in this and other spots.

Documentation/Usage.md
natikgadzhi: It did feel a little bit overwhelming when I tried to reproduce the steps, so I agree. Rewritten that in two steps, and cleaned up the `pkg-config` mention as well.

Documentation/Usage.md
natikgadzhi: After splitting the first one, I'm not so sure if we should split this one as well. It would add another blob of code, but since the second example builds up on the first one, I think the concept should be clear even without splitting. 

But, happy to split if you'd prefer it in two steps here as well!

natikgadzhi:
@tomerd, took another pass on the feedback, and addressed most of it, commented on the rest. Thank you for a thorough read-through and review! How does it look to you now?

I'll play around with the idea of tutorials and see how far I can get. If anything, that should help me understand SwiftPM and docc better. If that ends up being useful to someone else — great, if not — no worries, I'm doing that to learn.

MaxDesiatov:
@swift-ci smoke test

Documentation/Usage.md
```diff
@@ -116,14 +116,14 @@ https://github.com/apple/example-package-fisheryates
...
 
-You can link against system libraries using the package manager. To do so, there
-needs to be a special package for each system library that contains a modulemap
-for that library. Such a wrapper package does not contain any code of its own.
+You can link against system libraries using the package manager. To do so, you'll
+need to add a special `target` of type `.systemLibrary`, and a `module.modulemap`
+for each system library you're using.
 
 Let's see an example of using [libgit2](https://libgit2.github.com) from an
 executable.
```
MaxDesiatov, line 124:
```suggestion
Let's see an example of adding [libgit2](https://libgit2.github.com) as a
dependency to an executable target.
```

Documentation/Usage.md
```diff
@@ -146,81 +146,106 @@ following files from the libgit2 system-package are of interest:
     /usr/local/lib/libgit2.dylib      # .so on Linux
     /usr/local/include/git2.h
 
-Swift packages that provide module maps for system libraries are handled
-differently from regular Swift packages.
+**Note:** the system library may be located elsewhere on your system, such as
+`/usr/`, or `/opt/homebrew/` if you're using Homebrew on an Apple silicon Mac.
```
MaxDesiatov, line 150:
```suggestion
`/usr/`, or `/opt/homebrew/` if you're using Homebrew on an Apple Silicon Mac.
```

Documentation/Usage.md
```diff
@@ -146,81 +146,106 @@ following files from the libgit2 system-package are of interest:
...
-parameter you can pass the path of a directory containing the library using the
-`-L` flag in commandline when building your app:
+**Note:** If you don't want to use the `pkgConfig` parameter you can pass the
+path of a directory containing the library using the `-L` flag in command line
+when building your app instead:
 
     example$ swift build -Xlinker -L/usr/local/lib/
 
-Create a `module.modulemap` file so it consists of the following:
+Next, we'll create a directory `Sources/Clibgit` in your `example` projcet, and
```
MaxDesiatov, line 190: Nit: typo, I'd also avoid the use of "we" here to stay consistent with using "you" in other paragraphs.
```suggestion
Next, create a directory `Sources/Clibgit` in your `example` project, and
```

Documentation/Usage.md
```diff
@@ -231,6 +256,8 @@ executable:
     git_repository_init_options(version: 0, flags: 0, mode: 0, workdir_path: nil, description: nil, template_path: nil, initial_head: nil, origin_url: nil)
     example$
 
+### Requiring a System Library Without pkg-config
```
MaxDesiatov, line 259:
```suggestion
### Requiring a System Library Without `pkg-config`
```

Documentation/Usage.md
```diff
@@ -250,22 +277,16 @@ let jpegData = jpeg_common_struct()
...
-Install JPEG library using a system packager, e.g, `$ brew install jpeg`
-
-Create a directory called `CJPEG` next to the `example` directory and
-initialize it as a package that builds a system module:
-
-    example$ cd ..
-    $ mkdir CJPEG
-    $ cd CJPEG
-    CJPEG$ swift package init --type empty
+Install JPEG library using a system packager, e.g, `$ brew install jpeg`.
```
MaxDesiatov, line 280:
```suggestion
Install JPEG library using a Homebrew package manager with `brew install jpeg`.
```

Documentation/Usage.md
natikgadzhi: 🤦🏼 yikes, sorry about that.

natikgadzhi:
@MaxDesiatov, thank you for reviewing the whole thing, and for your feedback. Agreed, I was getting a bit lost in the narration style (we vs you). Ready for another round / merge.

Documentation/Usage.md
```diff
@@ -250,22 +277,16 @@ let jpegData = jpeg_common_struct()
...
-Install JPEG library using a system packager, e.g, `$ brew install jpeg`
-
-Create a directory called `CJPEG` next to the `example` directory and
-initialize it as a package that builds a system module:
-
-    example$ cd ..
-    $ mkdir CJPEG
-    $ cd CJPEG
-    CJPEG$ swift package init --type empty
+Install JPEG library, on macOS you can use Homebrew package manager: `brew install jpeg`.
```
MaxDesiatov, line 280:
```suggestion
Install the JPEG library, on macOS you can use Homebrew package manager: `brew install jpeg`.
```

Documentation/Usage.md
```diff
@@ -250,22 +277,16 @@ let jpegData = jpeg_common_struct()
...
+`jpeg` is a keg-only formula, meaning it won't be linked to `/usr/local/lib`,
+and you'll have to link it manually at build time.
 
-This creates `Package.swift` file in the directory.
-Create a `module.modulemap` file so it consists of the following:
+Just like in the previous example, `mkdir Sources/CJPEG`, and add the
```
MaxDesiatov, line 284:
```suggestion
Just like in the previous example, run `mkdir Sources/CJPEG` and add the
```

MaxDesiatov (commented):
Overall LGTM, just a couple more nits

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci smoke test windows

Documentation/Usage.md
```diff
@@ -146,81 +146,106 @@ following files from the libgit2 system-package are of interest:
...
+when building your app instead:
 
     example$ swift build -Xlinker -L/usr/local/lib/
 
-Create a `module.modulemap` file so it consists of the following:
+Next, create a directory `Sources/Clibgit` in your `example` project, and
+add a `module.modulemap` to it:
 
     module Clibgit [system] {
       header "/usr/local/include/git2.h"
```
tomerd, line 194: there is a somewhat better way to do this, namely to create you own header file, which then include the required one with `<header>` so that no path hardcoding is required.  eg: https://github.com/apple/swift-package-manager/tree/main/Sources/SPMSQLite3

natikgadzhi: Sorry, long flight — back in PST now. 

Ah, that's a TIL for me. So it looks like:
- You can, and probably should, add a header file in Sources/{LibName}/{LibName.h} that just includes {LibName.h} and/or other required headers. 
- The `#include` will resolve with the compiler environment, so it should pick up the headers correctly, and thus you don't have to write the absolute header path in the `module.modulemap`. 

Right?

tomerd: yup afaik. @neonichu to keep me honest here

natikgadzhi:
@neonichu @tomerd alright, done: https://github.com/apple/swift-package-manager/pull/6634/commits/f71f92eaf5dd96751b130f45926f3560039d0c37

I've double-checked that the example project builds this way in SwiftPM 5.8. Haven't extensively tested other platforms, though.

Considered adjusting the `CJPEG` example as well, but it's a bit different. On macOS, Homebrew does not link the headers (and libraries) for it, so just adding `#include <jpeglib.h>` would not work, and you'd still have to use an absolute path. So, I left the path in `module.modulemap` as an example.

Mergeable?

natikgadzhi:
@swift-ci please smoke test

natikgadzhi:
@swift-ci test windows

tomerd:
@swift-ci smoke test

Documentation/Usage.md
```diff
@@ -146,81 +146,116 @@ following files from the libgit2 system-package are of interest:
     /usr/local/lib/libgit2.dylib      # .so on Linux
     /usr/local/include/git2.h
 
-Swift packages that provide module maps for system libraries are handled
-differently from regular Swift packages.
+**Note:** the system library may be located elsewhere on your system, such as
```
compnerd, line 149: I think that we should clear up some of the text here.  Windows is a support platform, and these paths are not relevant there.  Additionally, `pkg-config` is not a commonly available utility on Windows.

Documentation/Usage.md
```diff
@@ -146,81 +146,116 @@ following files from the libgit2 system-package are of interest:
...
+    name: "example",
+    targets: [
+        // systemLibrary is a special type of build target that wraps a system library
+        // in a target that other targets can require as their depencency.
+        .systemLibrary(
+            name: "Clibgit",
+            pkgConfig: "libgit2",
+            providers: [
+                .brew(["libgit2"]),
+                .apt(["libgit2-dev"])
```
compnerd, line 176: Perhaps add a `.nuget` entry as well?

Documentation/Usage.md
```diff
@@ -146,81 +146,116 @@ following files from the libgit2 system-package are of interest:
...
+    ]
 )
+
 ```
 
-The `pkgConfig` parameter helps SwiftPM in figuring out the include and library
-search paths for the system library. Note: If you don't want to use the `pkgConfig`
-parameter you can pass the path of a directory containing the library using the
-`-L` flag in commandline when building your app:
+**Note:** If you don't want to use the `pkgConfig` parameter you can pass the
```
compnerd, line 184: `pkgConfig` should be elided for Windows only packages as `pkg-config` is not expected to be available.

Documentation/Usage.md
```diff
@@ -146,81 +146,116 @@ following files from the libgit2 system-package are of interest:
...
+      header "git2.h"
       link "git2"
       export *
     }
 
+The header file should look like this:
+
+```c
+# git2.h
+#pragma once
```
compnerd, line 203: Should we recommend include guards instead?

Documentation/Usage.md
```diff
@@ -250,22 +287,16 @@ let jpegData = jpeg_common_struct()
...
 
-This creates `Package.swift` file in the directory.
-Create a `module.modulemap` file so it consists of the following:
+Just like in the previous example, run `mkdir Sources/CJPEG` and add the
+following `module.modulemap`:
 
     module CJPEG [system] {
         header "shim.h"
-        header "/usr/local/include/jpeglib.h"
+        header "/usr/local/opt/jpeg/include/jpeglib.h"
```
compnerd, line 299: Should we recommend something with a VFS example?  That would make the majority of the code more cross-platform friendly.

natikgadzhi: For this and all your other comments, thank you! I'll catch up and clean things up a bit later today (PST) and ping you when it's all updated.

Documentation/Usage.md
natikgadzhi: Hmmm. Should we suggest [`LibGit2Sharp`](https://www.nuget.org/packages/LibGit2Sharp)? Swift doesn't need the .Net wrapper to build it's package, but there doesn't seem to be another minimal nuget package for it.

Documentation/Usage.md
natikgadzhi: I honestly don't have enough experience with C, so I don't know if using `#pragma once` is perfectly fine, or a bad taste. My understanding is that using guards is guaranteed to work anywhere, and using `#pragma once` is shorter, but might not work in certain environments?

Documentation/Usage.md
natikgadzhi: Added that to the note.

Documentation/Usage.md
natikgadzhi: Added an entry on Windows paths.

natikgadzhi:
@compnerd, I cleaned up some text and recommendations around Windows builds. But I don't feel confident re: VFS paths example and guards vs pragma once. 

I don't have a strong opinion that we should not recommend guards — I just don't have enough experience to have an opinion yet.

Same with VFS — I don't have enough experience to suggest where / when to use them, and what a good example might look like.

I want to see this little piece of work through. Can you please help me polish those areas? I'm open to suggestions, or perhaps we should merge this pull request, and I should get myself a Windows VM, and try and work through all the documentation on Windows, see what works, what does not, and make another pull request with any cleanup necessary?

tomerd:
I think this is a big improvement the previous version and we should merge it. 

@compnerd the previous version did not address windows at all, so I think we can tackle that a separate PR

tomerd:
@swift-ci smoke test

tomerd:
@swift-ci smoke test windows

PR #6636, opened 2023-06-06
Add PackageDescription platform constants for the 2023 Apple OS versions

abertelrud:
Add PackageDescription platform constants for the 2023 Apple OS versions.

### Motivation

This is an update that happens every year at Apple's annual WWDC conference, where the Apple platforms usually have their versions incremented.  This allows packages that specify the latest tools version (in this case 5.9) to declare that they require the latest platform versions, which may be appropriate in some cases.  Note that most packages should continue supporting the oldest platform tools version that they can (based on the APIs they use), so that they will be usable to the maximum number of client packages and projects.

### Modifications

- added new platform version number constants for macOS 14, iOS 17, tvOS 17, watchOS 10, macCatalyst 17, and DriverKit 23 (which has its own version number, though it isn't a separate platform)
  - these are available when the package manifest specifies tools version 5.9
- include a unit test that checks the 5.9 tools version spelling of these new platform versions
- update a test that checks the 5.9 tools version package manifest generation

### Result

Packages that specify tools version 5.9 can use these new platform constants.  Older packages can continue to spell these new version numbers using the free-form variant of each platform enum that takes a string.

### Remarks

Note that the just-announced visionOS is purposefully excluded from this PR, as it has not yet been released and as no version number has been announced.

rdar://107794704

### DIFF
--- a/Sources/PackageDescription/SupportedPlatforms.swift
+++ b/Sources/PackageDescription/SupportedPlatforms.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift open source project
 //
-// Copyright (c) 2018-2022 Apple Inc. and the Swift project authors
+// Copyright (c) 2018-2023 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See http://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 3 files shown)


abertelrud:
@swift-ci please test

abertelrud:
@swift-ci please smoke test

MaxDesiatov (approved):

neonichu (approved):

PR #6647, opened 2023-06-13
PackagePlugin: make `Path` a wrapper around `AbsolutePath`

compnerd:
This should help the path representation be properly represented on different platforms.  This is motivated by Windows which would have the drive stripped from the path, converting to a relative path the absolute path.


compnerd:
@swift-ci please test

PR #6653, opened 2023-06-15
Tests: attempt to handle arc separators in tests

compnerd:
The arc separator on Windows is `\` rather than `/`.  The path tests hard coded the expectation that the unix path separator is always used which is not a valid assumption.  Add a branch for windows in a large number of cases.  Although we do not fully pass the tests, this gets us closer.

### DIFF
--- a/Tests/BasicsTests/FileSystem/PathTests.swift
+++ b/Tests/BasicsTests/FileSystem/PathTests.swift
@@ -12,14 +12,20 @@ import Basics
 import Foundation
 import XCTest
 
+#if os(Windows)
+private var windows: Bool { true }
+#else
+private var windows: Bool { false }
+#endif
+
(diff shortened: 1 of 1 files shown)


tomerd (approved):
would have been nice if we could express these as something like:

```
XCTAssertEqual(AbsolutePath("/").pathString, windows ? "\" : "/")
```

would make maintenance / scaling this easier

compnerd:
I like that!  I'll take a pass and see if I can get that done.

neonichu (approved):

compnerd:
@swift-ci please smoke test

tomerd (approved):
nice!

compnerd:

compnerd:

PR #6657, opened 2023-06-19
Update PackageRegistryUsage.md (fix typo)

spqw:
Fix typo `iff` -> `if`

### Modifications:
`iff` -> `if`


MaxDesiatov:
This isn't a typo, `iff` is used in [the meaning of "if and only if"](https://en.wikipedia.org/wiki/If_and_only_if#Origin_of_iff_and_pronunciation) here.

PR #6662, opened 2023-06-21
Add `visionOS` as a platform alongside `iOS`, `tvOS`, `watchOS`, and the others

abertelrud:
Add new enum constants for visionOS in all the places where it is appropriate.

This includes the `platforms` parameter in the manifest, whose purpose is to list the minimum deployment targets that a package requires. Since there is initially only visionOS 1.0, and because absence of a platform in the `platforms` parameter means that the default version is supported (not that the platform is unsupported), all packages implicitly support visionOS 1.0 whether or not they list it in the `platforms` parameter. But having the enum constant alongside those for the other Apple platforms makes everything consistent, and will become important for any future versions.

This change also makes the `visionOS` platform available for use in conditional target dependencies.

### Motivation:

Xcode Beta 2, released earlier today, includes support for the new `visionOS` platform, and this change aligns SwiftPM with that.  Note that this does _not_ mean you can develop visionOS apps with SwiftPM CLI — support is at the same level as for iOS, tvOS, etc, which is that the platform can be mentioned in the manifest, and that when using the experimental `xcbuild` build system, building libraries for visionOS will be supported at the same level as iOS, tvOS, etc provided that a suitable Xcode is installed.

### Modifications:

- add `visionOS` wherever iOS, tvOS, watchOS, etc are defined
- add manifest source generation support
- adjust `PackageDescription.docc/Curation/SupportedPlatforms.md` (courtesy of @zkiraly)
- adjust unit tests

rdar://107064954

### DIFF
--- a/Sources/PackageCollections/Providers/JSONPackageCollectionProvider.swift
+++ b/Sources/PackageCollections/Providers/JSONPackageCollectionProvider.swift
@@ -577,6 +577,8 @@ extension PackageModel.Platform {
             self = PackageModel.Platform.tvOS
         case let name where name.contains("watchos"):
             self = PackageModel.Platform.watchOS
+        case let name where name.contains("visionos"):
+            self = PackageModel.Platform.visionOS
         case let name where name.contains("driverkit"):
             self = PackageModel.Platform.driverKit
         case let name where name.contains("linux"):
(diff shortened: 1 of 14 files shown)


abertelrud:
@swift-ci please smoke test

neonichu (approved):

MaxDesiatov (approved):

zkiraly (commented):
LGTM

PR #6665, opened 2023-06-22
Build: centralise the debug information flag handling

compnerd:
Rather than spreading out the flag addition across the various different targets, we can control the flags at the SwiftTool level, driving the necessary flag addition at a single site.  This would allow easier management of the flags as we can see the additions uniformly applied.

The test changes alter the expectations for ordering and additional flags being applied for various compilation cases.

### DIFF
--- a/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
@@ -220,10 +220,7 @@ public final class ClangTargetBuildDescription {
             args += ["-fobjc-arc"]
         }
         args += try buildParameters.targetTripleArgs(for: target)
-        args += ["-g"]
-        if buildParameters.triple.isWindows() {
-            args += ["-gcodeview"]
-        }
+
         args += optimizationArguments
         args += activeCompilationConditions
         args += ["-fblocks"]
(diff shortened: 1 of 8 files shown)


MaxDesiatov (approved):
I assume changes in tests only adjust the order of `-g` expectations? Otherwise LGTM.

compnerd:
@MaxDesiatov yeah, largely ... a couple of them do *add* additional cases of `-g` being passed, and there is one weird case where there is `-g -g` ... but, yes, the changes in the test expectations are due to the ordering changes.  I'm not sure how many of them actually care about the flags vs are just listing every flag rather than checking explicitly what they are interested in.

neonichu (changes requested):
I think I am kind of regretting the suggestion to put this logic into `SwiftTool` 😆 

I think at a minimum we have to preserve passing `-g` if someone uses SwiftPM as a library, but I am wondering now whether we should centralize all the logic into the `Build` module and have the debug format be an option in `BuildParameters`.

Sorry about flip-flopping here from my previous suggestion, but I think it's important that we preserve the existing functionality for anyone using SwiftPM as a library. In principle it seems ok for `SwiftTool` to offer additional functionality, but something basic like passing `-g` when building the debug config should come out of the box for clients of the library.

compnerd:
No need to be sorry, I should have pushed back harder - this is why my initial thought was to put this into the build.  I'm more than happy to do the migration into the right location, but, I don't seem to follow how to get the target information :-(.

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please smoke test

neonichu (approved):

compnerd:

PR #6666, opened 2023-06-22
Add support for Musl libc

MaxDesiatov:
Since Musl is sufficiently different from Glibc (see https://wiki.musl-libc.org/functional-differences-from-glibc.html), it requires a different import, which now should be applied to files that have `import Glibc` in them.


neonichu (approved):
I kinda hate that we have to do this kind of dance for very basic operations like exiting a process (!) or signal handling. This feels like something swift-system should provide abstractions for tbh.

MaxDesiatov:
> This feels like something swift-system should provide abstractions for tbh.

100%, but until then we have to import platform-specific libcs explicitly.

Sources/Basics/Cancellator.swift
```diff
@@ -77,27 +77,22 @@ public final class Cancellator: Cancellable {
...
                 #else
-                var action = sigaction()
                 action.__sigaction_handler = unsafeBitCast(
                     SIG_DFL,
                     to: sigaction.__Unnamed_union___sigaction_handler.self
                 )
+                #endif
                 sigaction(SIGINT, &action, nil)
                 kill(getpid(), SIGINT)
-                #endif
```
tomerd, line 100: this diff looks like it is doing more than just Musl support?

al45tair: I think this is just code rearrangement; every case did

```
sigaction(SIGINT, &action, nil)
kill(getpid(), SIGINT)
```

and I think Max just moved it outside of the `#if` conditions.

MaxDesiatov: Yes, those lines are common for all cases, they only differ in handler setup.

tomerd: ah I see the `#endif` moved 👍

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

finagolfin:
> 100%, but until then we have to import platform-specific libcs explicitly.

[This has already been pitched](https://forums.swift.org/t/pitch-the-cstdlib-module/51373). Would you take that over, @MaxDesiatov, since you are adding Musl, including adding Bionic and the BSD libcs?

I'll be happy to help split off Bionic, I just don't want to push all this through the Evolution process.

PR #6670, opened 2023-06-27
Allow opt-out of PII in CONTRIBUTORS.txt

therealbnut:
Update `generate_contributors_list.sh` to allow email removal to protect against PII visibility from web searches. This mirrors an already merged change in `swift-nio` (https://github.com/apple/swift-nio/pull/2453).

### Motivation:

`CONTRIBUTORS.txt` is leaking PII of contributors and making them easily searchable on sites like Google. The PII remains in the commit history, but that does not seem to be indexed in a similar way. This change adds an opt-out way to have these emails removed.

### Modifications:

This adds a new file `.mailfilter` which has a similar purpose to `.mailmap`, but is used to filter instead of map. The contents of this file are lines with a sha1 hash of each opted-out contributor's name email pair. The hash is intended to obfuscate its entries from search indexing, but does not need to be secure otherwise as the information is still in the git log.

If the git log format ever changes then this would need to be regenerated. Regeneration could easily be done by re-running the script from the same commit it was last generated from, and hashing the diff additions in `CONTRIBUTORS.txt`.

The script `generate_contributors_list.sh` has also been updated to use the filter, although `mailmap` doesn't appear to be used there.

### Result:

`CONTRIBUTORS.txt` can now be anonymised.

```diff
-- Someone's name <the-old-email>
```


tomerd:
@swift-ci smoke test

tomerd (approved):

PR #6682, opened 2023-07-07
Tests: adjust the BuildPlanTests for Windows

compnerd:
The Windows toolchain does not find the default linker by default as
that requires the sourcing of Visual Studio's scripts to setup the
environment.  Adjust the tests to explicitly specify the linker to
ensure that the toolchain is usable on Windows without constraint on the
environment.  While in the area, adjust the expectations to account for
platform path spellings.  This allows us to pass these tests on Windows
now.

Adjust the path representation to account for Windows to allow the
matching to succeed.

### DIFF
--- a/Tests/BuildTests/BuildPlanTests.swift
+++ b/Tests/BuildTests/BuildPlanTests.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import Basics
+@testable import Basics
 @testable import Build
 @_implementationOnly import DriverSupport
 import PackageLoading
@@ -43,12 +43,14 @@ final class BuildPlanTests: XCTestCase {
             "/barPkg/Sources/BarLogging/file.swift"
         )
         let observability = ObservabilitySystem.makeForTesting()
+        let fooPkg: AbsolutePath = "/fooPkg"
+        let barPkg: AbsolutePath = "/barPkg"
         XCTAssertThrowsError(try loadPackageGraph(
             fileSystem: fs,
             manifests: [
                 Manifest.createFileSystemManifest(
                     displayName: "fooPkg",
-                    path: "/fooPkg",
+                    path: fooPkg,
                     products: [
                         ProductDescription(name: "Logging", type: .library(.dynamic), targets: ["FooLogging"]),
                     ],
@@ -57,7 +59,7 @@ final class BuildPlanTests: XCTestCase {
                     ]),
                 Manifest.createFileSystemManifest(
                     displayName: "barPkg",
-                    path: "/barPkg",
+                    path: barPkg,
                     products: [
                         ProductDescription(name: "Logging", type: .library(.static), targets: ["BarLogging"]),
                     ],
@@ -69,24 +71,21 @@ final class BuildPlanTests: XCTestCase {
                     path: "/thisPkg",
                     toolsVersion: .v5_8,
                     dependencies: [
-                        .localSourceControl(path: "/fooPkg", requirement: .upToNextMajor(from: "1.0.0")),
-                        .localSourceControl(path: "/barPkg", requirement: .upToNextMajor(from: "2.0.0")),
+                        .localSourceControl(path: fooPkg, requirement: .upToNextMajor(from: "1.0.0")),
+                        .localSourceControl(path: barPkg, requirement: .upToNextMajor(from: "2.0.0")),
                     ],
                     targets: [
                         TargetDescription(name: "exe",
-                                          dependencies: [.product(name: "Logging",
-                                                                  package: "fooPkg"
-                                                                 ),
-                                                         .product(name: "Logging",
-                                                                  package: "barPkg"
-                                                                 ),
+                                          dependencies: [.product(name: "Logging", package: "fooPkg"),
+                                                         .product(name: "Logging", package: "barPkg"),
                                           ],
                                           type: .executable),
                     ]),
             ],
             observabilityScope: observability.topScope
         )) { error in
-            XCTAssertEqual((error as? PackageGraphError)?.description, "multiple products named 'Logging' in: 'barpkg' (at '/barPkg'), 'foopkg' (at '/fooPkg')")
+            XCTAssertEqual((error as? PackageGraphError)?.description,
+                           "multiple products named 'Logging' in: 'barpkg' (at '\(barPkg)'), 'foopkg' (at '\(fooPkg)')")
         }
     }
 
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:

Tests/BuildTests/BuildPlanTests.swift
```diff
@@ -3536,7 +3536,7 @@ final class BuildPlanTests: XCTestCase {
             toolset: .init(
                 knownTools: [
                     .cCompiler: .init(extraCLIOptions: ["-I/fake/sdk/sysroot", "-clang-flag-from-json"]),
-                    .swiftCompiler: .init(extraCLIOptions: ["-swift-flag-from-json"])
+                    .swiftCompiler: .init(extraCLIOptions: ["-swift-flag-from-json", "-use-ld=lld"]),
```
tomerd, line 3539: is this a windows only change, or impacting on all platforms?

compnerd: It will be used on all the platforms.  The issue is that we were previously relying on leaking `ld` on the host and relying on that existing on Linux and macOS.

tomerd: but `lld` does not exists everywhere either?

compnerd: Ah, using `-use-ld=lld` will change the linker that is picked up and use `lld` which is part of the toolchain.  Let me see if I can instead add `.linker: .init(path: "/fake/toolchain/not-a-linker")`

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

tomerd (approved):

compnerd:

Tests/BuildTests/BuildPlanTests.swift
```diff
@@ -3770,17 +3787,17 @@ final class BuildPlanTests: XCTestCase {
...
         // Compile Swift Target
         let exeCompileArguments = try result.target(for: "exe").swiftTarget().compileArguments()
-        let exeCompileArgumentsPattern: [StringPattern] = ["-I", "\(sdkIncludeSearchPath)"]
+        let exeCompileArgumentsPattern: [StringPattern] = ["-I", "\(AbsolutePath(sdkIncludeSearchPath)._nativePathString(escaped: false))"]
         XCTAssertMatch(exeCompileArguments, exeCompileArgumentsPattern)
 
         // Link Product
         let exeLinkArguments = try result.buildProduct(for: "exe").linkArguments()
-        let exeLinkArgumentsPattern: [StringPattern] = ["-L", "\(sdkIncludeSearchPath)"]
+        let exeLinkArgumentsPattern: [StringPattern] = ["-L", "\(AbsolutePath(sdkIncludeSearchPath)._nativePathString(escaped: false))"]
```
MaxDesiatov, line 3800: @tomerd WDYT of adjusting `CustomStringConvertible` and `description` of `AbsolutePath` to use `_nativePathString` under the hood so that we don't have to spell it out explicitly everywhere?

compnerd: @MaxDesiatov - and then add an `escaped` member to escape it?  Or would we do that via a standalone function or would that be an member function?

tomerd: @compnerd - @MaxDesiatov @neonichu @abertelrud and myself discussed this some more and believe the solution should be to change `AbsolutePath::pathString` to return the native path string unescaped then escape where necessary. @MaxDesiatov will follow up with how implement and test this, so that we can get rid of `_nativePathString` and make that behavior the default string representation

MaxDesiatov: Partially addressed in https://github.com/apple/swift-package-manager/pull/6730. If you think that's a correct approach I'll expand it to clean up the remaining uses of `_nativePathString` in our codebase.

compnerd:
@swift-ci please smoke test

Tests/BuildTests/BuildPlanTests.swift
```diff
@@ -4740,9 +4751,8 @@ final class BuildPlanTests: XCTestCase {
 
         let yamlContents: String = try fs.readFileContents(yaml)
         XCTAssertMatch(yamlContents, .contains("""
-            inputs: ["/Pkg/Snippets/ASnippet.swift","/Pkg/.build/debug/Lib.swiftmodule"
+            inputs: ["\(AbsolutePath("/Pkg/Snippets/ASnippet.swift")._nativePathString(escaped: true))","\(AbsolutePath("/Pkg/.build/debug/Lib.swiftmodule")._nativePathString(escaped: true))"
```
MaxDesiatov, line 4754: I think with `https://github.com/apple/swift-package-manager/pull/6825` we can declare these strings outside of `XCTAssertMatch` as `SerializedJSON` and that will interpolate them correctly with escaping. That also should be semantically correct, since JSON is a subset of YAML.

MaxDesiatov: Turns out that PR is not needed, this should just work:
```suggestion
            inputs: ["\(AbsolutePath("/Pkg/Snippets/ASnippet.swift"))","\(AbsolutePath("/Pkg/.build/debug/Lib.swiftmodule"))"
```

compnerd: No, the escaping here is necessary as the strings themselves are escaped in the JSON.  This is correcting the expectation.

MaxDesiatov: Ah, I missed `SerializedJSON` wrapping in my suggestion, and for some reason GitHub doesn't allow me to posit an easily committable multi-line suggestion here. I think it should look like this:

```swift
        let yamlContents: String = try fs.readFileContents(yaml)
        let inputs: SerializedJSON = """
            inputs: ["/Pkg/Snippets/ASnippet.swift","/Pkg/.build/debug/Lib.swiftmodule"
            inputs: ["\(AbsolutePath("/Pkg/Snippets/ASnippet.swift"))","\(AbsolutePath("/Pkg/.build/debug/Lib.swiftmodule"))"
        """
        XCTAssertMatch(yamlContents, .contains(inputs.underlying))
```

compnerd:
@swift-ci please smoke test

MaxDesiatov (approved):
Perfect, thank you! 👍

compnerd:
@swift-ci please smoke test

MaxDesiatov:
@swift-ci test windows

PR #6684, opened 2023-07-08
Remove extra paths added to LD_LIBRARY_PATH on Linux when running tests

bnbarham:
These break running tests when there's already a swift on PATH that isn't the swift currently being run.

They were originally added as a workaround to (https://github.com/apple/swift-package-manager/pull/5952), but the RUNPATH looks correct (includes the library of the compiler that built the test and "$ORIGIN") and tests are working without it.

### DIFF
--- a/Sources/Commands/Utilities/TestingSupport.swift
+++ b/Sources/Commands/Utilities/TestingSupport.swift
@@ -157,16 +157,6 @@ enum TestingSupport {
         if let location = toolchain.xctestPath {
             env.prependPath("Path", value: location.pathString)
         }
-        #elseif os(Linux)
-        var libraryPaths = ["/usr/lib/swift/linux"]
-        if let path = env["PATH"], let firstPathEntry = path.components(separatedBy: ":").first {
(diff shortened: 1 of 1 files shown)


bnbarham:
@swift-ci please test

DougGregor (approved):

DougGregor:
@swift-ci smoke test

DougGregor:
@swift-ci please bootstrap

DougGregor:
@swift-ci please test Windows

DougGregor:

DougGregor:
@swift-ci test Windows

bnbarham:
@swift test windows platform

bnbarham:

bnbarham:

tomerd:
removing workarounds is great 👍🏼. let’s make sure @neonichu is comfortable as he has the most background on why that was even needed in the first place

bnbarham:
So looking through the history, it looks like this was actually about the RUNPATH of `libXCTest.so`, which is indeed still missing the `$ORIGIN`. I'm not sure why this now passes, but then by the looks of it we also weren't sure why it started failing in the first place.

PR #6686, opened 2023-07-08
Tests: adjust PackageModelTests for Windows

compnerd:
Adjust the path spelling for the native path arc separator and escaping
for the encoding in the JSON.  Windows uses `\` as the path arc
separator which needs to be escaped for ensuring that the JSON is valid.

### DIFF
--- a/Tests/PackageModelTests/SwiftSDKTests.swift
+++ b/Tests/PackageModelTests/SwiftSDKTests.swift
@@ -408,12 +408,13 @@ final class DestinationTests: XCTestCase {
             fileSystem: fs,
             observabilityScope: observability
         )) {
+            let toolsetDefinition: AbsolutePath = "/tools/asdf.json"
             XCTAssertEqual(
                 $0 as? StringError,
                 StringError(
                     """
-                    Couldn't parse toolset configuration at `/tools/asdf.json`: /tools/asdf.json doesn't exist in file \
-                    system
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please smoke test

compnerd:

tomerd:
all these use of `_nativePathString(escaped: true)` makes me feel bit uncomfortable and maybe pointing to something more broadly missing

compnerd:
> all these use of `_nativePathString(escaped: true)` makes me feel bit uncomfortable and maybe pointing to something more broadly missing

I don't feel that is the case.  We need to ensure that the JSON that we are hardcoding here is valid - the paths have `\` as the arc separator, we need to escape them.

compnerd:

tomerd:
> > all these use of _nativePathString(escaped: true) makes me feel bit uncomfortable and maybe pointing to something more broadly missing
> I don't feel that is the case. We need to ensure that the JSON that we are hardcoding here is valid - the paths have \ as the arc separator, we need to escape them.

right, I understand why it was done, the point is that if need to litter too many of our tests with `_nativePathString(escaped: true) ` it points to needing a different approach potentially

compnerd:
@swift-ci please smoke test

MaxDesiatov (approved):
Nice, thank you!

PR #6687, opened 2023-07-09
Build: use target properties to define the target

compnerd:
Use target specific build parameters rather than the global ones to define the build target.  This will simplify building multiple variants of the target.

### DIFF
--- a/Sources/Build/LLBuildManifestBuilder.swift
+++ b/Sources/Build/LLBuildManifestBuilder.swift
@@ -619,18 +619,18 @@ extension LLBuildManifestBuilder {
             name: cmdName,
             inputs: inputs + [Node.file(target.sourcesFileListPath)],
             outputs: cmdOutputs,
-            executable: self.buildParameters.toolchain.swiftCompilerPath,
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please smoke test

Sources/Build/LLBuildManifestBuilder.swift
```diff
@@ -619,18 +619,18 @@ extension LLBuildManifestBuilder {
             name: cmdName,
             inputs: inputs + [Node.file(target.sourcesFileListPath)],
             outputs: cmdOutputs,
-            executable: self.buildParameters.toolchain.swiftCompilerPath,
```
rauhul, line 622: JW: what are "self"'s build parameters in the context of a build? Some set of default target agnostic params? something else?

compnerd: Yeah, they would be the global build parameters.  In order to support libraries, we need to have target specific build parameters as they need to be compiled with different flags based upon whether you are statically or dynamically linking to the library.

neonichu (approved):

tomerd:
nice

PR #6688, opened 2023-07-10
PackageModel: honour `.librarian` for toolsets

compnerd:
When defining a custom toolset, a specified librarian shall be given precedence over the platform's librarian.  Use this to repair the toolset test on Windows where `ar` is unavailable.  This test used to succeed due to the leaking of the host tools.


compnerd:
@swift-ci please smoke test

neonichu (approved):

PR #6692, opened 2023-07-10
[5.8] Remove extra paths added to LD_LIBRARY_PATH on Linux when running tests

bnbarham:
* **Testing**: Toolchain build and test on all Linux platforms.

### DIFF
--- a/Sources/Commands/Utilities/TestingSupport.swift
+++ b/Sources/Commands/Utilities/TestingSupport.swift
@@ -133,16 +133,6 @@ enum TestingSupport {
         if let location = toolchain.xctestPath {
             env.prependPath("Path", value: location.pathString)
         }
-        #elseif os(Linux)
-        var libraryPaths = ["/usr/lib/swift/linux"]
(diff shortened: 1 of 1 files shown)


bnbarham:
I suspect this may fail, building/testing toolchains now.

tomerd:
@bnbarham does this need to be merged?

bnbarham:
Same comment as the other one. Yes, but we need to check it first.

bnbarham:
Same results as 5.9 - the test executable has eg. `/home/build-user/ws/tmp/pmtest58/usr/lib/swift/linux:$ORIGIN` (where pmtest58 is the toolchain downloaded from the linked PR) in its RPATH and works fine without the extra LD_LIBRARY_PATH.

bnbarham:
@swift-ci please test

DougGregor (approved):

MaxDesiatov:
@swift-ci smoke test

neonichu (approved):

PR #6701, opened 2023-07-12
build: silence a pkg-config warning on Windows

compnerd:
Windows does not have pkg-config by default.  Since the nuget support is still incomplete, prefer to simply disregard the pkg-config spec on Windows.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -61,6 +61,12 @@ let swiftPMProduct = (
     ]
 )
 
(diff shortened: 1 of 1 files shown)


tomerd (approved):

MaxDesiatov (approved):

compnerd:
@swift-ci please smoke test

PR #6711, opened 2023-07-14
Preserve port when computing the login URL

Palleas:
When using a registry URL that contains a port, ensure the port is preserved when computing the login URL.

### Motivation:

Currently if you run the following
```
swift package-registry set https://packages.example.com:8081
swift package-registry login --token XYZ
```

You might get something like 
```
Error: Unable to connect to server: https://packages.example.com
```

That's because the login url is built like this and ignores the port: `URL(string: "https://\(host)\(loginAPIPath ?? "/login")")`

### Modifications:

- Use `URLComponents` to build the login URL (which maintains the port)
- Add a testing target for `SwiftPackageRegistryTool`

### Result:

SPM can use registry URLs with ports.

I do have a question though, the registry authentication configuration dingus currently uses the host as the key:
https://github.com/apple/swift-package-manager/blob/main/Sources/PackageRegistryTool/PackageRegistryTool%2BAuth.swift#L234-L237

I suspect it's pretty unlikely that someone would use multiple registries available at the same host on different ports, but should I update that, as well?


yim-lee:
@swift-ci please test

Tests/PackageRegistryToolTests/PackageRegistryToolTests.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+import XCTest
+@testable import PackageRegistryTool
+
+final class PackageRegistryToolTests: XCTestCase {
```
yim-lee, line 5: These new tests can go into `CommandsTests/PackageRegistryToolTests.swift`.

yim-lee:
@swift-ci please smoke test

yim-lee:
> I suspect it's pretty unlikely that someone would use multiple registries available at the same host on different ports, but should I update that, as well?

I agree that it's unlikely to have multiple registries on the same host. 

If we were to support/fix this however, I don't know if using `host:port` instead of just `host` is good enough. What if instead of different ports, different paths (e.g., `/staging-1`, `/staging-2`) on the same host are used for multiple registries? Using (registry) `URL`s as keys could make things tricky in [`RegistryClient`](https://github.com/apple/swift-package-manager/blob/main/Sources/PackageRegistry/RegistryClient.swift#L78-L96). But then I think distinguishing registries by paths is even less likely to happen than ports.

I think changing from `host` to `host:port` is pretty straight-forward--the `:port` portion should only be added if `registryURL.port` is non-`nil`, so I am open to us making that change. I don't think the change is a MUST though.

Tests/PackageRegistryToolTests/PackageRegistryToolTests.swift
Palleas: Roger that, done!

Palleas:
> I think changing from host to host:port is pretty straight-forward--the :port portion should only be added if registryURL.port is non-nil, so I am open to us making that change. I don't think the change is a MUST though.

Done!

Sources/PackageRegistryTool/PackageRegistryTool+Auth.swift
```diff
@@ -90,6 +90,28 @@ private func readpassword(_ prompt: String) throws -> String {
 
 extension SwiftPackageRegistryTool {
     struct Login: SwiftCommand {
+
+        static func authenticationStorageKey(for registryURL: URL) throws -> String {
```
yim-lee, line 94: We need to update the "read" too: https://github.com/apple/swift-package-manager/blob/main/Sources/PackageRegistry/RegistryConfiguration.swift#L69-L72

yim-lee:
@swift-ci please smoke test

yim-lee:

Sources/PackageRegistryTool/PackageRegistryTool+Auth.swift
```diff
@@ -226,15 +236,12 @@ extension SwiftPackageRegistryTool {
                 loginAPIPath = registryURL.path
             }
 
-            // Login URL must be HTTPS
-            guard let loginURL = URL(string: "https://\(host)\(loginAPIPath ?? "/login")") else {
-                throw ValidationError.invalidURL(registryURL)
-            }
+            let loginURL = try SwiftPackageRegistryTool.Login.loginURL(from: registryURL, loginAPIPath: loginAPIPath)
```
yim-lee, line 239: Does `Self` work?

```suggestion
            let loginURL = try Self.loginURL(from: registryURL, loginAPIPath: loginAPIPath)
```

Tests/CommandsTests/PackageRegistryToolTests.swift
```diff
@@ -903,6 +903,23 @@ final class PackageRegistryToolTests: CommandsTestCase {
...
+    func testCreateLoginURL() {
+        let registryURL = URL(string: "https://packages.example.com")!
+
+        XCTAssertEqual(try SwiftPackageRegistryTool.Login.loginURL(from: registryURL, loginAPIPath: nil).absoluteString, "https://packages.example.com/login")
+
+        XCTAssertEqual(try SwiftPackageRegistryTool.Login.loginURL(from: registryURL, loginAPIPath: "/secret-sign-in").absoluteString, "https://packages.example.com/secret-sign-in")
+
+    }
+
+    func testCreateLoginURLMaintainsPort() {
```
yim-lee, line 915: <3

yim-lee (approved):

yim-lee:

Sources/PackageRegistryTool/PackageRegistryTool+Auth.swift
Palleas: I don't see why not 🤔

yim-lee:
@swift-ci please smoke test

yim-lee:

yim-lee:

tomerd (approved):

yim-lee:

yim-lee:
@Palleas This looks good to merge! Can you please create a PR for the `release/5.9` branch also?

Palleas:
> @Palleas This looks good to merge! Can you please create a PR for the `release/5.9` branch also?

Here you go! https://github.com/apple/swift-package-manager/pull/6715

yim-lee:
@Palleas Thank you!

PR #6715, opened 2023-07-18
[5.9] Preserve port when computing the login URL

Palleas:
(Same as #6711 but for 5.9.)

When using a registry URL that contains a port, ensure the port is preserved when computing the login URL.

### Motivation:

Currently if you run the following
```
swift package-registry set https://packages.example.com:8081
swift package-registry login --token XYZ
```

You might get something like 
```
Error: Unable to connect to server: https://packages.example.com
```

That's because the login url is built like this and ignores the port: `URL(string: "https://\(host)\(loginAPIPath ?? "/login")")`

### Modifications:

- Use `URLComponents` to build the login URL (which maintains the port)
- Update RegistryConfiguration to compute the storage key

### Result:

SPM can use registry URLs with ports.

### DIFF
--- a/Sources/PackageRegistry/RegistryClient.swift
+++ b/Sources/PackageRegistry/RegistryClient.swift
@@ -75,7 +75,7 @@ public final class RegistryClient: Cancellable {
 
         if let authorizationProvider {
             self.authorizationProvider = { url in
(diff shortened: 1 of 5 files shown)


Palleas:
@swift-ci please test

yim-lee:
@swift-ci please smoke test

PR #6717, opened 2023-07-18
Swift SDKs: implement `swiftResourcesPath`

kabiroberai:
This PR implements `swiftResourcesPath` and `swiftStaticResourcesPath` from the `swift-sdk.json` spec.

### Motivation:

SwiftPM previously knew how to parse these fields from the SDK manifest, but didn't actually do anything with them.

### Modifications:

- Pass `-resource-dir` to swiftc for compile jobs
- Pass `-resource-dir` to both swiftc and clang for link jobs

### Result:

Toolchains that override `swiftResourcesPath` and/or `swiftStaticResourcesPath` should now behave as expected.


Sources/PackageModel/UserToolchain.swift
```diff
@@ -40,6 +40,14 @@ public final class UserToolchain: Toolchain {
     /// An array of paths to search for libraries at link time.
     public let librarySearchPaths: [AbsolutePath]
 
+    public var swiftResourcesPath: AbsolutePath? {
```
MaxDesiatov, line 43: These new properties need doc comments.

MaxDesiatov (approved):
Thanks!

kabiroberai:
@MaxDesiatov mind kicking off CI again? Had to rebase on top of 0e50b63.

Sources/Build/BuildDescription/ProductBuildDescription.swift
```diff
@@ -218,6 +219,7 @@ public final class ProductBuildDescription: SPMBuildCore.ProductBuildDescription
                     self.observabilityScope.emit(.swiftBackDeployError)
                 } else if self.buildParameters.targetTriple.isSupportingStaticStdlib {
                     args += ["-static-stdlib"]
+                    hasStaticStdlib = true
```
rauhul, line 222: Shouldn't shouldLinkStaticSwiftStdlib when !isSupportingStaticStdlib be an error?

MaxDesiatov:
> @MaxDesiatov mind kicking off CI again? Had to rebase on top of [0e50b63](https://github.com/apple/swift-package-manager/commit/0e50b63ccdf5881f14b8bd1a998804411c33b36b).

We usually squash and merge here, so there's no need to rebase unless you really prefer that workflow. A lot of times a merge commit is easier to resolve conflicts with, and it won't show up after squash and merge in the target branch anyway.

Sources/Build/BuildDescription/ProductBuildDescription.swift
kabiroberai: It seems a bit harder to adopt that approach here because the inclusion of `-static-stdlib` in the linker invocation also depends on the product type. If I understand your point correctly, I agree that there should be better validation around `shouldLinkStaticSwiftStdlib` (e.g. emitting an error if the product is a dynamic library) but imo that would be better suited to a follow-up PR since it's orthogonal to this change — what do you think?

kabiroberai:
Gotcha, will keep that in mind — wasn't sure about the convention since it seems they do merge commits over on apple/swift.

Sources/Build/BuildDescription/ProductBuildDescription.swift
MaxDesiatov: A follow-up PR is fine by me, but in that case would you mind adding a TODO or FIXME note in code comments? I hope this would be addressed sooner rather than later.

Sources/Build/BuildDescription/ProductBuildDescription.swift
```diff
@@ -243,6 +246,16 @@ public final class ProductBuildDescription: SPMBuildCore.ProductBuildDescription
             throw InternalError("unexpectedly asked to generate linker arguments for a plugin product")
         }
 
+        if let resourcesPath = self.buildParameters.toolchain.swiftResourcesPath(isStatic: isLinkingStaticStdlib) {
+            args += ["-resource-dir", "\(resourcesPath)"]
```
kabiroberai, line 250: @MaxDesiatov is it correct to rely on interpolation here instead of `pathString`?

tomerd: should this just move into the `if self.buildParameters.shouldLinkStaticSwiftStdlib` branch? its used only there it seems

MaxDesiatov: Interpolation will render it properly on Windows: https://github.com/apple/swift-package-manager/blob/14ff299060cfd7f1bab3a4fc9065f77d427f8093/Sources/Basics/FileSystem/NativePathExtensions.swift#L16

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test

MaxDesiatov:
> This seems to have broken SPM on Windows, I'm going to revert this for now, please make sure to run the tests on Windows locally in the future.

@compnerd Would you be able to provide the exact error message? This would help us fix the issue with no readily available local access to Windows and no testing set up on CI either. +1 to Tom's question why wasn't this caught on CI in the first place?

compnerd:
@MaxDesiatov I _think_ that it was actually the change after this (XCTest integration), at least after a bit more digging. This wasn't caught on CI due to the test suite not running in CI, only building SPM.

PR #6721, opened 2023-07-18
Fix CMake build to explicitly pick up ar and ranlib from toolchain.

al45tair:
CMake under various conditions will attempt to use `llvm-ar` and `llvm-ranlib` when it spots that we're building using `clang`, and while that works fine on Linux, it doesn't work so well on Darwin, leaving us with a link failure with a perplexing error message.

The rest of the Swift build is resilient to this problem already, but the `bootstrap` script here doesn't pass the relevant options to CMake, so it's possible to end up with the build failing when trying to link `lib/libCYaml.a` after the build has mangled the static library with `llvm-ranlib`.

rdar://112478653

### DIFF
--- a/Utilities/bootstrap
+++ b/Utilities/bootstrap
@@ -104,6 +104,14 @@ def add_build_args(parser):
         metavar='PATH',
         help='path to the ninja binary to use for building with CMake')
     parser.add_argument(
+        '--ar-path',
+        metavar='PATH',
+        help='path to the ar binary to use for building with CMake')
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


al45tair:
@swift-ci Please test

MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

tomerd:
@al45tair please create 5.9 cherry pick once this lands on main

etcwilde (approved):
Seems reasonable to me.

PR #6728, opened 2023-07-19
Remove -incremental when using -whole-module-optimization

keith:
Passing both results in noisy remarks when building from the command line like:

```
remark: Incremental compilation has been disabled: it is not compatible with whole module optimization
```

https://forums.swift.org/t/incremental-compilation-has-been-disabled-it-is-not-compatible-with-whole-module-optimization/66092

### DIFF
--- a/Sources/LLBuildManifest/Tools.swift
+++ b/Sources/LLBuildManifest/Tools.swift
@@ -316,7 +316,6 @@ public struct SwiftCompilerTool: ToolProtocol {
             }
         }
         arguments += [
-            "-incremental",
             "-emit-dependencies",
             "-emit-module",
             "-emit-module-path", moduleOutputPath.pathString,
@@ -327,6 +326,8 @@ public struct SwiftCompilerTool: ToolProtocol {
         }
         if wholeModuleOptimization {
             arguments += ["-whole-module-optimization", "-num-threads", "\(Self.numThreads)"]
+        } else {
+            arguments += ["-incremental"]
         }
         arguments += ["-c", "@\(self.fileList.pathString)"]
         arguments += ["-I", importPath.pathString]


keith:
@swift-ci please test

finagolfin (approved):

MaxDesiatov:
@swift-ci smoke test

artemcm (approved):
Thanks.

neonichu (approved):

keith:
hrm seems like this isn't a problem in the release branch https://github.com/apple/swift-package-manager/blob/6ed7215e66c50176f21342e3135d3afccea1c7d1/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift#L543

keith:
ah it looks like llbuild needs a similar change, I will submit PRs there

keith:
https://github.com/apple/swift-llbuild/pull/884

keith:
so at HEAD this change is enough but on 5.9 this change isn't needed and the llbuild change is enough. I submitted the llbuild change in both places either way

neonichu:
Yah, we switched away from using llbuild's built-in Swift command in `main`

doozMen:
When I run with swift 5.9 but in my package.swift I have 5.7 it still outputs remark: Incremental compilation has been disabled: it is not compatible with whole module optimization. Does this mean it is not fixed in 5.9?

finagolfin:
> Does this mean it is not fixed in 5.9?

It sounds like you're picking up an old Swift in your path somewhere. Try running with `-v` verbose mode and see which Swift compiler it's using. If you're sure it's 5.9 or later, let us know.

kaishin:
> It sounds like you're picking up an old Swift in your path somewhere. Try running with -v verbose mode and see which Swift compiler it's using. If you're sure it's 5.9 or later, let us know.

I am still seeing this with Swift 5.9.

a594508257:
image

Apple Swift version 5.9.2 (swiftlang-5.9.2.2.56 clang-1500.1.0.2.5)
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-help intro

PR #6730, opened 2023-07-19
Add `SerializedJSON` type as string literal escaping helper

MaxDesiatov:
### Motivation:

Currently we have to call `_nativePathString` manually in a lot of places where `AbsolutePath` is interpolated into strings. We also need to take into account whether `_nativePathString` should be called with `escaped: true` or `escaped: false` arguments, which is prone to errors. The serialization format should carry the knowledge of its correct escaping algorithm, not the path representation itself.

We can make this much cleaner if string interpolations call `_nativePathString` automatically where needed, and serialization formats add proper escaping on top of that.

Let's mplements this for JSON, specifically where it is used in Swift SDKs.

### Modifications:

Added `SerializedJSON` type as string literal escaping helper.

Also added a custom `AbsolutePath` interpolation to the new type, which removes the need to call `_nativePathString` when paths are interpolated in JSON strings.

### Result:

We don't have to call `_nativePathString` manually in as many places. In future PRs `_nativePathString` should be declared `private` or at least `internal` so that it's hidden as an implementation detail of appropriate string interpolations.

### DIFF
--- a/Tests/PackageModelTests/SwiftSDKTests.swift
+++ b/Tests/PackageModelTests/SwiftSDKTests.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import Basics
+@testable import Basics
 @testable import PackageModel
 @testable import SPMBuildCore
 import XCTest
@@ -31,7 +31,7 @@ private let extraFlags = BuildFlags(
 )
 
 private let destinationV1 = (
-    path: "\(bundleRootPath)/destinationV1.json",
+    path: bundleRootPath.appending(component: "destinationV1.json"),
     json: #"""
     {
         "version": 1,
@@ -42,11 +42,11 @@ private let destinationV1 = (
         "extra-swiftc-flags": \#(extraFlags.swiftCompilerFlags),
         "extra-cpp-flags": \#(extraFlags.cxxCompilerFlags)
     }
-    """#
+    """# as SerializedJSON
 )
 
 private let destinationV2 = (
-    path: "\(bundleRootPath)/destinationV2.json",
+    path: bundleRootPath.appending(component: "destinationV2.json"),
     json: #"""
     {
         "version": 2,
@@ -59,11 +59,11 @@ private let destinationV2 = (
         "extraCXXFlags": \#(extraFlags.cxxCompilerFlags),
         "extraLinkerFlags": \#(extraFlags.linkerFlags)
     }
-    """#
+    """# as SerializedJSON
 )
 
 private let toolsetNoRootDestinationV3 = (
-    path: "\(bundleRootPath)/toolsetNoRootDestinationV3.json",
+    path: bundleRootPath.appending(component: "toolsetNoRootDestinationV3.json"),
     json: #"""
     {
         "runTimeTriples": {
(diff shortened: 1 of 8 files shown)


MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

Tests/PackageModelTests/SwiftSDKTests.swift
```diff
@@ -348,15 +348,15 @@ final class DestinationTests: XCTestCase {
             otherToolsNoRoot,
             someToolsWithRoot,
             invalidToolset,
-        ] {
-            try fs.writeFileContents(AbsolutePath(validating: testFile.path), string: testFile.json)
+        ] as [(path: AbsolutePath, json: SerializedJSON)] {
```
compnerd, line 351: Why not move this into the variable declaration?

```
for testFile: [(AbsolutePath, SerializedJSON)] in [
  ...
] {
  try fs.writeFileContents($0.path, string: $0.json.underlying)
}

**Review by compnerd — APPROVED** [review]

File: Tests/PackageModelTests/SwiftSDKTests.swift (hunk 1 above)
Comment (MaxDesiatov) [reply Tests/PackageModelTests/SwiftSDKTests.swift:351]: Separated out into a file-local `private let` declaration in the latest commit.

**Comment by MaxDesiatov** [comment]
@swift-ci smoke test

**Comment by MaxDesiatov** [comment]
@swift-ci test windows

File: Tests/PackageModelTests/SwiftSDKTests.swift [hunk 2]
```diff
@@ -31,7 +31,7 @@ private let extraFlags = BuildFlags(
 )
 
 private let destinationV1 = (
-    path: "\(bundleRootPath)/destinationV1.json",
+    path: bundleRootPath.appending(component: "destinationV1.json"),
```
tomerd, line 34: `component: ` is redundant

Tests/PackageModelTests/SwiftSDKTests.swift
```diff
@@ -59,11 +59,11 @@ private let destinationV2 = (
         "extraCXXFlags": \#(extraFlags.cxxCompilerFlags),
         "extraLinkerFlags": \#(extraFlags.linkerFlags)
     }
-    """#
+    """# as SerializedJSON
```
tomerd, line 62: make SerializedJSON ExpressibleByStringLiteral so this case can go away

Tests/PackageModelTests/SwiftSDKTests.swift
```diff
@@ -209,29 +209,29 @@ private let invalidToolsetSwiftSDKv4 = (
...
 )
 
 private let usrBinTools = Dictionary(uniqueKeysWithValues: Toolset.KnownTool.allCases.map {
     ($0, "/usr/bin/\($0.rawValue)")
 })
 
 private let otherToolsNoRoot = (
-    path: "/tools/otherToolsNoRoot.json",
+    path: try! AbsolutePath(validating: "/tools/otherToolsNoRoot.json"),
```
tomerd, line 220: `try validating:` is not required in tests (there is an extension that does the force try)

Tests/PackageModelTests/SwiftSDKTests.swift
tomerd: oh, this is tuple... maybe use a type to avoid the cast then

PR #6734, opened 2023-07-20
PackageLoading,Basics: properly handle cross-root manifests

compnerd:
The path representation uses a relative path (`/relative/paths`) as an absolute path due to limitations of Unix systems.  If the source is located in one root and the manifest is built into another root, the path remains a relative path which will fail to match the VFS entry as the relative path is made absolute by normalizing the path to the current drive as per Windows rules.  This then results in the VFS entry not being honoured and the path fails to be found.  More aggressively perform the path normalization when loading the manifest to resolve this issue.

### DIFF
--- a/Sources/Basics/FileSystem/NativePathExtensions.swift
+++ b/Sources/Basics/FileSystem/NativePathExtensions.swift
@@ -14,6 +14,8 @@ import Foundation
 import struct TSCBasic.AbsolutePath
 
 extension AbsolutePath {
+    /// Returns the File System Representation of the `AbsolutePath`'s
+    /// `pathString` property converted into a `URL`.
     public func _nativePathString(escaped: Bool) -> String {
         return URL(fileURLWithPath: self.pathString).withUnsafeFileSystemRepresentation {
             let repr = String(cString: $0!)
--- a/Sources/PackageLoading/ManifestLoader.swift
+++ b/Sources/PackageLoading/ManifestLoader.swift
@@ -24,6 +24,30 @@ import struct TSCBasic.ProcessResult
 import enum TSCUtility.Diagnostics
 import struct TSCUtility.Version
 
+#if os(Windows)
+import WinSDK
+#endif
+
+extension AbsolutePath {
+    /// Returns the `pathString` on non-Windows platforms.  On Windows
+    /// platforms, this provides the path string normalized as per the Windows
+    /// path normalization rules.  In the case that the path is a long path, the
+    /// path will use the extended path syntax (UNC style, NT Path).
+    internal var _normalized: String {
+#if os(Windows)
+        return self.pathString.withCString(encodedAs: UTF16.self) { pwszPath in
+            let dwLength = GetFullPathNameW(pwszPath, 0, nil, nil)
+            return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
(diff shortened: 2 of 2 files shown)


compnerd:
@swift-ci please smoke test

Sources/Basics/FileSystem/NativePathExtensions.swift
```diff
@@ -23,6 +26,20 @@ extension AbsolutePath {
             return repr
         }
     }
+
+    public var _native: String {
```
MaxDesiatov, line 30: What's the difference between `_nativePathString` and `_native`? When would we use one and not the other? If we have to maintain both versions, would be better to rename them to make the distinction more obvious?

tomerd: also i would strongly prefer such extensions be on path and not on string

compnerd: @tomerd they are on `AbsolutePath`, not on `String`, or did I screw up the rebase?

@MaxDesiatov The `_native` might be better named `_normalized`, it is the path normalized as per the Windows path normalization rules, rather than the FSR of the path.  I'll rename it.

compnerd:
@swift-ci please smoke test

neonichu (approved):

Sources/Basics/FileSystem/NativePathExtensions.swift
MaxDesiatov: What's FSR of the path? How's it defined? Can these details be added as doc comments to this file so that we can reference it later when working on it?

compnerd: Its the file system representation, it doesn't guarantee the normalization per se, only that it will convert the URL representation to what the file system expects.  I don't think we should be defining that, but mentioning that the function just returns the FSR vs the normalized path seems reasonable.

compnerd:
@swift-ci please smoke test

compnerd:

Sources/Basics/FileSystem/NativePathExtensions.swift
```diff
@@ -23,6 +28,24 @@ extension AbsolutePath {
             return repr
         }
     }
+
+    /// Returns the `pathString` on non-Windows platforms.  On Windows
+    /// platforms, this provides the path string normalized as per the Windows
+    /// path normalization rules.  In the case that the path is a long path, the
+    /// path will use the extended path syntax (UNC style, NT Path).
+    public var _normalized: String {
```
tomerd, line 36: should this be the default behavior of `pathString` on windows?

compnerd: I've been thinking about it, but I want to fix up the test suite first and then test things before making such a change.  That will allow us to ensure that we don't accidentally regress something.  In the mean time, this would repair the use of SPM.

tomerd: in that case I think you should move `_normalized` closer to the single call site, and if we have a second one we make it the default behavior. wdyt?

compnerd: Sure, so you want to move it into that file?  I'm totally fine with that.

compnerd:
@swift-ci please smoke test

tomerd (approved):

compnerd:

PR #6742, opened 2023-07-21
Build: filter `-use-ld=lld` from `swift-symbolgraph-extract`

compnerd:
This option is not parsed by the symbol graph extractor, nor does it materially make a difference to the tool.  Filter out this flag on the SPM side to ensure that we are able to properly run the tool.  This partially enables the use of `swift package dump-symbol-graph` on Windows.

### DIFF
--- a/Sources/Build/BuildPlan.swift
+++ b/Sources/Build/BuildPlan.swift
@@ -916,7 +916,11 @@ public class BuildPlan: SPMBuildCore.BuildPlan {
         let buildPath = buildParameters.buildPath.pathString
         var arguments = ["-I", buildPath]
 
-        var extraSwiftCFlags = buildParameters.toolchain.extraFlags.swiftCompilerFlags
+        // swift-symbolgraph-extract does not support parsing `-use-ld=lld` and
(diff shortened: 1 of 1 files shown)


Sources/Build/BuildPlan.swift
```diff
@@ -916,7 +916,7 @@ public class BuildPlan: SPMBuildCore.BuildPlan {
         let buildPath = buildParameters.buildPath.pathString
         var arguments = ["-I", buildPath]
 
-        var extraSwiftCFlags = buildParameters.toolchain.extraFlags.swiftCompilerFlags
+        var extraSwiftCFlags = buildParameters.toolchain.extraFlags.swiftCompilerFlags.filter { !$0.starts(with: "-use-ld=") }
```
MaxDesiatov, line 919: Please add a code comment above this line so that the context for this use of `filter` is clear when coming back to this code later.

compnerd:
@swift-ci please smoke test

MaxDesiatov (approved):
Thanks!

neonichu (approved):

PR #6756, opened 2023-07-29
Add MacPorts package manager support

### LINKED ISSUE #6716: Please add support for MacPorts in SystemPackageProvider.
### Description

MacPorts provides official mirrors in Mainland China and is extremely stable as a package manager. 

GitHub is severely DNS-polluted here, and the entire Brew depends on GitHub.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

_No response_

_No response_

_No response_

Kyle-Ye:
Add MacPorts package manager support

### Motivation:

Close #6716 

### Modifications:

Add port item after brew item

### Result:

Provide MacPorts package manager as a SystemPackageProvider

### TODO:

- Add test case for MacPorts
- Test it with a demo package
[Clibgit.zip](https://github.com/apple/swift-package-manager/files/12206286/Clibgit.zip)


Kyle-Ye:
@swift-ci please test

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -395,6 +398,16 @@ public enum SystemPackageProvider {
...
         return .brewItem(packages)
     }
+    
+    /// Creates a system package provider with a list of installable packages
+    /// for people who use the MacPorts package manager on macOS.
+    ///
+    /// - Parameter packages: The list of package names.
+    ///
+    /// - Returns: A package provider.
+    public static func port(_ packages: [String]) -> SystemPackageProvider {
```
MaxDesiatov, line 408: Let's make the name more obvious, applicable to other places where `port` was used as well.
```suggestion
    public static func macPorts(_ packages: [String]) -> SystemPackageProvider {
```

MaxDesiatov (changes requested):

Sources/PackageDescription/PackageDescription.swift
Kyle-Ye: 1. We should consider the consistency of API.
For `HomeBrew` package manager (`brew install xx`), we use `brew` here.
For `apt-get` package manager (`apt install xx`), we use `apt` here.

IMO for `MacPorts` package manager (`port install xx`), we should use `port` here.

2. Swift have a namespace system.
So I'm not expecting it to appear in the wrong place. 
We can only use `.port(["xx"])` if the callee accepts a `SystemPackageProvider` params.

MaxDesiatov: Sticking to the name of a CLI command, e.g. `brew` or `apt` there was a mistake. There are enough package managers CLI named `port` or `pkg` on BSD systems for example. By squatting `port` here we'll create a mess for possible future BSD support. I'm not sure how exactly limited means of namespacing in Swift would help.

MaxDesiatov: Additionally, using plain `port` creates confusion for a reader. I'm skimming through a package and see `.port` there without any disambiguation. Is that a Mac port, FreeBSD port, or OpenBSD port? Brevity and consistency is good, but not at the cost of reduced readability and increased chance for confusion.

Kyle-Ye: > Sticking to the name of a CLI command, e.g. `brew` or `apt` there was a mistake. There are enough package manager CLI commands named `port` or `pkg` on BSD systems for example. By squatting `port` here we'll create a mess for possible future BSD support. I'm not sure how exactly limited means of namespacing in Swift would help.

Got it. I'll update the feedback to the forum post.

Kyle-Ye: > Sticking to the name of a CLI command, e.g. `brew` or `apt` there was a mistake. There are enough package manager CLI commands named `port` or `pkg` on BSD systems for example. By squatting `port` here we'll create a mess for possible future BSD support. I'm not sure how exactly limited means of namespacing in Swift would help.

I'm convinced by the logic here and happy to make such change. but I'd like to know how other reviewers and the community thinking of this before making a new commit.

MaxDesiatov: @tomerd WDYT? Does deprecating old names and introducing a new one require an SE pitch, proposal, and review? Or should we just introduce a new `macPorts` here without an SE process, and discuss deprecation of the old ones separately?

Kyle-Ye: Any progress on this? Can we merge it with `@available(_PackageDescription, introduced: 999.0)` the same as #5486 ?

Kyle-Ye:
@swift-ci please test

ShikiSuen:
How's this PR going?

Kyle-Ye:
> How's this PR going?

@ShikiSuen We need more code owner to review and push this.

See https://github.com/apple/swift-package-manager/pull/6756#discussion_r1360667392

Sources/PackageDescription/PackageDescription.swift
ShikiSuen: @Kyle-Ye I prefer the `.macPorts(` since MacPort is specific to macOS and is not 100% equal to the `port` in other systems (e.g. FreeBSD ports, etc.).

ShikiSuen: @MaxDesiatov Please do consider respond to Kyle-Ye's response above. This PR solves the issue that in Mainland China one can hardly use Brew without taking legal risks to use a cross-GFW virtual private network. MacPorts is accessible since its mirrors are available inside the Mainland China.

(Also, again I suggest treating `MacPorts` and non-mac `Ports` differently with individual APIs since the availability of a package might differ between them.)

ShikiSuen:
@Kyle-Ye If convenient, please consider synchronizing all new changes from the main branch to this PR.

This PR is really important and I can't understand why it gets hindered for years just due to the lack of reviews from the maintainers.

Kyle-Ye:
> @Kyle-Ye If convenient, please consider synchronizing all new changes from the main branch to this PR.
> 
> This PR is really important and I can't understand why it gets hindered for years just due to the lack of reviews from the maintainers.

Thanks for your focus on the PR. I'll update the PR/solve the conflicts this weekend and try to bring it to more eyes by the new Contributor Experience Workgroup to see if other reviewer can help give it a look to speed the merge process.

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
Update to use `MacPorts` as API name as review has suggested.

Any other suggestion or blocking items otherwise?

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
The nightly toolchain CI failure seems unrelated with this PR. Will rerun it later.

```
    error: link command failed with exit code 1 (use -v to see invocation)
    ld: multiple errors: could not parse bitcode object file /private/var/folders/bb/hcrjxg1s0b96pfst0ymhmp240000gn/T/swiftLib-eaff33.o: 'Unknown attribute kind (86) (Producer: 'LLVM17.0.0' Reader: 'LLVM APPLE_1_1500.1.0.2.4_0')', using libLTO version 'LLVM version 15.0.0' in '/private/var/folders/bb/hcrjxg1s0b96pfst0ymhmp240000gn/T/swiftLib-eaff33.o'; could not parse bitcode object file /private/var/folders/bb/hcrjxg1s0b96pfst0ymhmp240000gn/T/main-cf0ec5.o: 'Unknown attribute kind (86) (Producer: 'LLVM17.0.0' Reader: 'LLVM APPLE_1_1500.1.0.2.4_0')', using libLTO version 'LLVM version 15.0.0' in '/private/var/folders/bb/hcrjxg1s0b96pfst0ymhmp240000gn/T/main-cf0ec5.o'
    clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Sources/PackageLoading/Target+PkgConfig.swift
```diff
@@ -209,6 +215,8 @@ extension SystemPackageProviderDescription {
                 }
             }
             return try packages.map({ try AbsolutePath(validating: brewPrefix).appending(components: "opt", $0, "lib", "pkgconfig") })
+        case .macPorts:
+            return [try AbsolutePath(validating: "/opt/local/lib/pkgconfig")]
```
jakepetroules, line 219: MacPorts can be installed places other than /opt/local, right? Should we have a similar `macPortsPrefixOverride` that gets threaded through, similar to the one for Homebrew?

jakepetroules (commented):
Just had one comment. Overall, the change looks solid!

Sources/PackageLoading/Target+PkgConfig.swift
Kyle-Ye: 1. It appears that the brewPrefix is primarily used in the unit test context. Correct me if I'm wrong.
2. Regarding the Homebrew prefix, the prefix (Most because different default path for different arch: /opt/homebrew for Apple Silicon, /usr/local for macOS Intel) is handled by `brew --prefix`, and I haven’t found an equivalent straightforward command for MacPorts to get the prefix.

> Occasionally a MacPorts developer may wish to install more than one MacPorts instance on the same host. Only one copy of MacPorts may use the default prefix /opt/local, so for additional installations, the --prefix option can be used, as shown below.

Additionally, to use a custom prefix for MacPorts, the user must install it from source, which is a rare case for most SwiftPM users.

Kyle-Ye: I'm open to continue adding support for custom prefix MacPorts installation.

But I think we can land the basic support first and if there is downstream user actually needs this, we can add it later.

jakepetroules: Fair enough.

Kyle-Ye:
@swift-ci please test windows

Kyle-Ye:
The Windows and macOS nightly toolchain CI seems to be broken at the moment.

I'll merge the latest commits in main and try the pipelines later when a new PR pass all pipelines and has got merged.

By the way, if there is no other issue can I get an approve for this PR? cc @MaxDesiatov @jakepetroules

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
@swift-ci please test Windows

Kyle-Ye:
@swift-ci please smoke test

wwalexander:
What needs to happen to get this merged? Would really love to not need Homebrew to use `.systemLibrary`.

Kyle-Ye:
According to the forum discussion, it is suggested that I need to raise a formal Swift proposal before landing this change.

PR #6758, opened 2023-07-30
WIP: executable products installation PoC

MaxDesiatov:
### Motivation:

Users building executable products would likely want to install such products to a location in their `PATH` so that they could use such products in their workflows.

### Modifications:

Introduced a new `swift package experimental-install` subcommand that builds and copies executable products to `~/.swiftpm/installed-packages/bin`, which then can be added to `PATH` by the user if needed.

#### Limitations:

1. No version tracking (yet). For a proper implementation we should install to subdirectories that contain versions of corresponding packages, i.e. `~/.swiftpm/installed-packages/artifacts/swift-openapi-generator/0.1.5/bin/swift-openapi-generator` and then symlinked into `~/.swiftpm/installed-packages/bin/swift-openapi-generator`
2. Products installed via this command are fragile on Linux as they aren't linked statically and we're unlikely to achieve full static linking until `import FoundationNetworking` and `import FoundationXML` can be linked completely statically with all of their transitive dependencies.
3. We should (but currently don't) check if any libraries in the graph for these products are either system libraries or dynamically linked and bail out with an error if that's the case. While dynamic libraries potentially could be installed, I don't think that's something we should aim to implement in this prototype. I also don't think we can do much about system libraries. Since SwiftPM can't prevent users from updating or removing system libraries, packages should package such dependencies with SwiftPM to be able to install products that depend on them.
4. Not added to `CMakeLists.txt` files as Windows support is untested and will likely require a somewhat different approach.

### Result:

A barebones proof of concept implementation is available to experiment with. As it hasn't gone through Swift Evolution, it has an `experimental-` prefix and isn't displayed in `--help` output.

### DIFF
--- a/Sources/Basics/FileSystem/FileSystem+Extensions.swift
+++ b/Sources/Basics/FileSystem/FileSystem+Extensions.swift
@@ -492,6 +492,67 @@ extension FileSystem {
     }
 }
 
+// MARK: - Installed Packages
+
+private let installedPackagesDirectoryName = "installed-packages"
+
+extension FileSystem {
+    /// Path to the installed packages directory (if exists)
+    public var installedPackagesDirectory: AbsolutePath {
+        get throws {
+            if let path = try idiomaticSwiftPMDirectory {
+                return path.appending(component: installedPackagesDirectoryName)
+            } else {
+                return try dotSwiftPMSwiftInstalledPackagesDirectory
+            }
+        }
+    }
+
+    private var dotSwiftPMSwiftInstalledPackagesDirectory: AbsolutePath {
+        get throws {
+            try dotSwiftPM.appending(component: installedPackagesDirectoryName)
+        }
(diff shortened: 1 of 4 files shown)


MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci smoke test linux

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
Closing in favor of https://github.com/apple/swift-package-manager/pull/6768

PR #6780, opened 2023-08-05
canonicalize repository lcoation for storage path calculation

tomerd:
motivation: there is some built-in non-determinism where the graph resolution races and sometime one location variant wins in the location -> identity mapping, this can lead to looking up a different storage path since that does not use the normalized form of the location

changes:
* normalize (canonicalize) the repository location when computing the storage path
* add and adjust tests

rdar://112134161

### DIFF
--- a/Tests/WorkspaceTests/WorkspaceTests.swift
+++ b/Tests/WorkspaceTests/WorkspaceTests.swift
@@ -3808,7 +3808,7 @@ final class WorkspaceTests: XCTestCase {
                     products: [
                         MockProduct(name: "Foo", targets: ["Foo"]),
                     ],
-                    versions: ["1.0.0"],
+                    versions: ["1.0.0", "1.1.0"],
                     revisionProvider: { version in version } // stable revisions
                 ),
                 MockPackage(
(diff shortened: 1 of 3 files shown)


tomerd:
@swift-ci smoke test

tomerd:
@swift-ci smoke test

Tests/WorkspaceTests/WorkspaceTests.swift
```diff
@@ -3808,7 +3808,7 @@ final class WorkspaceTests: XCTestCase {
                     products: [
                         MockProduct(name: "Foo", targets: ["Foo"]),
                     ],
-                    versions: ["1.0.0"],
+                    versions: ["1.0.0", "1.1.0"],
```
MaxDesiatov, line 3811: Is this change related?

MaxDesiatov (approved):

Tests/WorkspaceTests/WorkspaceTests.swift
tomerd: yea it demonstrated the bug unintentionally

neonichu (approved):

PR #6784, opened 2023-08-08
Fingerprint filename for the same Git repository is not consistent

yim-lee:
Motivation:
The fingerprint filename for package coming from a Git repository should take the form of `{PACKAGE_NAME}-{REPOSITORY_URL_HASH}.json`. The `REPOSITORY_URL_HASH` part is to distinguish repository URLs with the same last path component (and therefore same package name). For example: https://github.com/apple/swift-nio
https://github.com/yim-lee/swift-nio

`REPOSITORY_URL_HASH` is computed by converting `hashValue` property of the URL string to hexadecimal, but turns out `hashValue` doesn't yield consistent value across program executions (because random seed is used), and as a result multiple fingerprint files get created for the same (package name, repository URL) pair.

Modification:
Generate consistent hash value to be used for fingerprint filename.

rdar://113540625

### DIFF
--- a/Sources/PackageFingerprint/FilePackageFingerprintStorage.swift
+++ b/Sources/PackageFingerprint/FilePackageFingerprintStorage.swift
@@ -398,8 +398,21 @@ extension PackageReference: FingerprintReference {
         }
 
         let canonicalLocation = CanonicalPackageLocation(sourceControlURL.absoluteString)
-        let locationHash = String(format: "%02x", canonicalLocation.description.hashValue)
+        // Cannot use hashValue because it is not consistent across executions
+        let locationHash = String(format: "%02x", canonicalLocation.description.stableHash)
         return "\(self.identity.description)-\(locationHash).json"
+    }
+}
+
+extension String {
+    // From: https://stackoverflow.com/questions/35882103/hash-value-of-string-that-would-be-stable-across-ios-releases/43149500#43149500
+    fileprivate var stableHash: UInt64 {
+        var result = UInt64(5381)
+        let buf = [UInt8](self.utf8)
+        for b in buf {
+            result = 127 * (result & 0x00FF_FFFF_FFFF_FFFF) + UInt64(b)
+        }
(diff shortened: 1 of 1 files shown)


yim-lee:
@swift-ci please smoke test

Sources/PackageFingerprint/FilePackageFingerprintStorage.swift
```diff
@@ -398,11 +398,24 @@ extension PackageReference: FingerprintReference {
         }
 
         let canonicalLocation = CanonicalPackageLocation(sourceControlURL.absoluteString)
-        let locationHash = String(format: "%02x", canonicalLocation.description.hashValue)
+        // Cannot use hashValue because it is not consistent across executions
+        let locationHash = String(format: "%02x", canonicalLocation.description.stableHash)
```
tomerd, line 402: could we use the same hash function on string as we do for repository storage location?

tomerd (approved):

Sources/PackageFingerprint/FilePackageFingerprintStorage.swift
yim-lee: can you point me to that code please?

tomerd: https://github.com/apple/swift-package-manager/pull/6780/files#diff-989b9f25e2078b5e7709d63f4d57027c8263c49ac74e2200413ee6f6dc878db5R571

yim-lee: thanks! amended https://github.com/apple/swift-package-manager/pull/6784/commits/aac13e7f8905c88a7b2e670a824eb1762a996c7e

yim-lee:
@swift-ci please smoke test

yim-lee:

tomerd (approved):

neonichu (approved):

PR #6799, opened 2023-08-11
[5.9] Build command plugin dependencies for the host, not the target

MaxDesiatov:
When cross-compiling, SwiftPM plugins and their dependencies run on the host and so must be compiled for the host OS and architecture, not the cross-compiled target OS and architecture. SwiftPM will compile a command plugin for the host but if the plugin depends on an executable it will cross-compile the executable for the target, so the plugin will not be able to run it:

    error: Error Domain=NSPOSIXErrorDomain Code=8 "Exec format error"

Command plugin dependencies are already handled specially in PluginCommand.run; this commit makes that special build step use the host toolchain instead of the target toolchain.

https://github.com/apple/swift-package-manager/pull/6060 handled the equivalent problem for build tool plugins.

(cherry picked from commit 1daaa1c0339584e50d42f62b5a7794a2e97b5782)
```
# Conflicts:
#	Sources/CoreCommands/SwiftTool.swift
```


MaxDesiatov:
@swift-ci smoke test

tomerd (approved):

PR #6803, opened 2023-08-12
Generate docc bundle and contents on library init

iThinker:
Generate docc bundle and contents on library init.

### Motivation:
Xcode Framework template generates docc documentation bundle automatically. Which is a great way to promote documentation.
Generating new Swift package does not do this as of now. Which seems inconsistent.

### Modifications:

- Generate `%moduleName%.md` and `Resources` folder under `Sources/%moduleName%.docc`. This follows same pattern as Xcode Framework template.
- Only `library` module for now. But might be worth expanding it to all packages that generate sources.
- Update tests.

### Result:

When creating a new library package `Sources/%moduleName%.docc` with `%moduleName%.md` and `Resources` folder will be generated.

### Discussion
Not sure if this falls into category of "materially change the user facing semantics of Swift Package Manager" or not. IMHO it does not but someone might hold different opinion.
On one hand new user artifacts are generated.
On the other hand the artifacts are minor, and don't break anything.
https://www.swift.org/getting-started/#using-the-package-manager mentions `README.md` is supposed to be generated. But it is not the case as of now. So this is kinda a more modern version of `README.md`.

### DIFF
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -613,14 +613,14 @@ public final class InitPackage {
         try makeDirectories(tests)
         try writeTestFileStubs(testsPath: tests)
     }
-    
+
     private func writeLibraryDocumentation(_ path: AbsolutePath) throws {
         let docc = path.appending("\(moduleName).docc")
         guard fileSystem.exists(docc) == false else {
             return
         }
         try makeDirectories(docc)
-        
+
         let content = ####"""
         # ``\####(moduleName)``
 
@@ -636,12 +636,12 @@ public final class InitPackage {
 
         - <!--@START_MENU_TOKEN@-->``Symbol``<!--@END_MENU_TOKEN@-->
         """####
-        
+
         let mdFile = docc.appending("\(moduleName).md")
         try writePackageFile(mdFile) { stream in
             stream.write(content)
         }
-        
+
         let resources = docc.appending("Resources")
         try makeDirectories(resources)
     }
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -595,6 +595,9 @@ public final class InitPackage {
           try writeMacroPluginSources(sources.appending("\(pkgname)Macros"))
           try writeMacroClientSources(sources.appending("\(pkgname)Client"))
         }
+        if packageType == .library {
+            try writeLibraryDocumentation(sources)
+        }
     }
 
     private func writeTests() throws {
@@ -609,6 +612,38 @@ public final class InitPackage {
         progressReporter?("Creating \(tests.relative(to: destinationPath))/")
         try makeDirectories(tests)
         try writeTestFileStubs(testsPath: tests)
+    }
+    
+    private func writeLibraryDocumentation(_ path: AbsolutePath) throws {
+        let docc = path.appending("\(moduleName).docc")
+        guard fileSystem.exists(docc) == false else {
+            return
+        }
+        try makeDirectories(docc)
+        
+        let content = ####"""
+        # ``\####(moduleName)``
+
+        <!--@START_MENU_TOKEN@-->Summary<!--@END_MENU_TOKEN@-->
+
+        ## Overview
+
+        <!--@START_MENU_TOKEN@-->Text<!--@END_MENU_TOKEN@-->
+
+        ## Topics
+
+        ### <!--@START_MENU_TOKEN@-->Group<!--@END_MENU_TOKEN@-->
+
+        - <!--@START_MENU_TOKEN@-->``Symbol``<!--@END_MENU_TOKEN@-->
+        """####
+        
+        let mdFile = docc.appending("\(moduleName).md")
+        try writePackageFile(mdFile) { stream in
+            stream.write(content)
+        }
+        
+        let resources = docc.appending("Resources")
+        try makeDirectories(resources)
     }
 
     private func writeLibraryTestsFile(_ path: AbsolutePath) throws {
--- a/Tests/WorkspaceTests/InitTests.swift
+++ b/Tests/WorkspaceTests/InitTests.swift
@@ -141,7 +141,9 @@ class InitTests: XCTestCase {
                           \(testFileContents)
                           """)
             XCTAssertMatch(testFileContents, .contains("func testExample() throws"))
-
+            
+            XCTAssertEqual(try fs.getDirectoryContents(path.appending("Sources").appending("Foo.docc")), ["Foo.md", "Resources"])
+            
             // Try building it
             XCTAssertBuilds(path)
             let triple = try UserToolchain.default.targetTriple


SDGGiesbrecht:
The DocC plug‐in currently [looks for the documentation catalog in the target’s source directory](https://github.com/apple/swift-docc-plugin/blob/3aad4aabc42835f09f600bb17070f6a11bc84180/Plugins/SharedPackagePluginExtensions/SourceModuleTarget%2BdoccCatalogPath.swift#L14-L17), i.e. `Sources/%moduleName%/%moduleName%.docc`.

Several forward‐looking considerations that would be good to know:
- Can DocC handle more than one target in a package yet?
- Once it does, will it do so by accepting one catalog per package or one catalog per target?

iThinker:
@SDGGiesbrecht thanks for catching that. Those rules are unobvious, and I got lost in the subfolders. Will update to move it down.
Not sure how to answer those forward-looking questions. Probably need to test it. Is there a command in SwiftPM to add a target?
My hunch is that it should be able to handle one catalog per target.

iThinker:
Did a quick local test.
Documentation is generated per target and included in the package documentation bundle. The only condition is for the target to be included in any product.

This should answer the above questions.
Can DocC handle more than one target in a package yet? -  **Yes.**
Once it does, will it do so by accepting one catalog per package or one catalog per target? - **One catalog per target.**

SDGGiesbrecht:
To verify we are talking about the same thing: How many sites does it produce? And can you navigate the menus from one target’s page to another?

I ask because DocC’s documentation [still says](https://apple.github.io/swift-docc-plugin/documentation/swiftdoccplugin/publishing-to-github-pages/) it can only build for one target at a time. Some time ago, when I last checked, the plug‐in appeared to work, having constructed multiple page collections (each a directory under `documentation`) that shared the web resources in neighbouring directories (`css`, `js`, etc.), but which all needed, paradoxically, to be at the hosting base path. It was just an accident from repeatedly overwriting the same output directory.

At that time, if you had multiple targets, you needed to manually do one at a time to put them in fully separate sites in order for all of them to work. Then if you wanted them cross‐linked, you needed to do it manually in the `.md` files. (The end result of all that manual work was a file structure [like this](https://github.com/SDGGiesbrecht/SDGCornerstone/tree/gh-pages).) That is not what I call support, and it was not how the team wanted it to eventually work.

So did the team make more progress than I am aware of and their documentation is out of date, or is the apparently unified output you saw inherently broken?

SDGGiesbrecht:
If anyone wants more information on the development of multi‐target documentation, see these two threads:

https://forums.swift.org/t/use-cases-for-combined-documentation-of-multiple-targets-in-swift-docc/59319
https://forums.swift.org/t/combined-documentation-of-multiple-targets/59579

SDGGiesbrecht:
@d-ronnqvist (if I have the right one),

In the view of DocC itself, what is the standard arrangement for package documentation both (a) as current workaround and (b) according to the most‐likely final design? (See the preceding conversation.)

iThinker:
Screenshot 2023-08-13 at 14 16 11
Screenshot 2023-08-13 at 14 16 24
Screenshot 2023-08-13 at 14 18 30

Here is what I test and get:
- 3 targets in 2 products.
- Xcode generates individual target documentation under workspace documentation.
- SwiftPM Documentation Plugin generates 3 separate archives. One per target.

@SDGGiesbrecht not sure how cross linking should look like so cannot test it.

Given my limited knowledge on the DocC techical vision, my take is that current schema is docc-per-target that needs to be supported. Current SwiftPM PR is done in line with it. If there are any changes to this schema, like introduction of a package level DocC, then `Init` would need to be updated accordingly. If any such change is imminent, then it probably makes sense to hold off current PR. Otherwise this PR is IMHO a good addition.

neonichu:
cc @bitjammer

natikgadzhi:
If this moves forward, it would be good to also document this behavior so that folks don't end up with dummy empty docc bundles that they don't need. And perhaps we could make this optional, with `--skip-documenation` flag to opt out?

Sources/Workspace/InitPackage.swift
```diff
@@ -611,6 +614,38 @@ public final class InitPackage {
...
+        ## Overview
+
+        <!--@START_MENU_TOKEN@-->Text<!--@END_MENU_TOKEN@-->
+
+        ## Topics
+
+        ### <!--@START_MENU_TOKEN@-->Group<!--@END_MENU_TOKEN@-->
+
+        - <!--@START_MENU_TOKEN@-->``Symbol``<!--@END_MENU_TOKEN@-->
+        """####
```
bitjammer, line 638: I think it would be better to not put anticipatory content in here, and it's especially important to not use Xcode-specific template content.

```suggestion
        let content = "# ``\(moduleName)``\n\n"
```

bitjammer (changes requested):
Thanks for opening this PR! I've been discussing this off and on with some folks so this is good timing.

This is a worthy goal. I'd like to start this as opt-in, as we currently just moved to simplifying all of the init templates. My concrete suggestion is to add a `--with-docs` flag to the `init` command that will conditionally add this content. Be sure to only allow the flag when `--type library` is active for the time being. We've actually just recently moved to emitting less for all of the templates and we should continue that trajectory.

On that note, please don't put anticipatory content in the generated files themselves. We just moved away from that and are actually emitting only the bare minimum. It wouldn't hurt to add the module name header, though!

Hope this helps-thanks again!

iThinker:
@bitjammer thanks for the feedback. I will address it a bit later.
My initial thought is that adding optional flags will significantly reduce discoverability of the feature. But I totally get the idea of reducing bloat.
IMHO interactive mode for init (and mb other tools) would help a lot. But that's outside of this PR scope.

d-ronnqvist:
> @d-ronnqvist (if I have the right one),
> 
> In the view of DocC itself, what is the standard arrangement for package documentation both (a) as current workaround and (b) according to the most‐likely final design? (See the preceding conversation.)

Yes, that's me. 

From the DocC perspective, each target is documented separately with its own .docc catalog of markup and assets for documentation about that target. 

Our designs for the future support for combined documentation for multiple targets will continue to use per target .docc catalogs for target-specific documentation and use an additional .docc catalog of markup and assets for documentation 
that isn't specific to either of the targets in the combined archive.

As such, adding a target specific .docc catalog is a correct approach from the DocC perspective.

iThinker:
OK. I am finally back from Path Of Exile.
@bitjammer any suggestions how to make new flag (`--with-docs`) only available when init type is library? It looks like Argument Parser does not have anything for that. So I will do custom validation either in `Init` or in `InitPackage`. I am leaning towards former since it is easier. Let me know if some other approach is preferred.

sofiaromorales:
@iThinker I'm doing some work on Swift-DocC for initializing catalogs, and considering integration with the DocC plugin as the next step. I believe it's a great idea to pursue this effort while contemplating integration with Swift-DocC to leverage the same implementation and templating work. You probably want to take a look at this https://forums.swift.org/t/pitch-docc-init-command-catalog-templates/67648

iThinker:
@sofiaromorales it kinda becomes chicken and egg problem.
Docc plugin is needed to generate documentation, but it is unavailable on initial package creation.
If there was a setup wizard it could ask user if documentation should be added, and then include docc plugin.
But in the current model not sure there are good options.
If this PR is ignored, and instead we have some easy way to add documentation that is fine too.
We need some sort of meta-init. So that when user wants to add a plugin, it can be done via a command, that could also execute plugin init. E.g. `swiftpm add plugin swift-docc %optional-version%`.

PR #6811, opened 2023-08-14
[netrc] Use the most recent match for auth

yim-lee:
Motivation:
We update `netrc` by appending new entry to the end of the file (we don't delete or modify existing entries). This means when we look for entries matching a given host in the `netrc` file, we should use the _last_ match for the most recent entry.

Modification:
Check `netrc` lookup to use `last` instead of `first` match.


yim-lee:
@swift-ci please smoke test

neonichu (approved):

PR #6812, opened 2023-08-15
CLI commands should not require workspace to update user-level registries config

yim-lee:
Motivation:
Certain `package-registry` subcommands, such as `set --global` and `login`, work with user-level `registries.json` only and therefore should not require workspace to run.

Modifications:
In those subcommands, do not read local `registries.json`.

rdar://113875867


yim-lee:
@swift-ci please smoke test

MaxDesiatov (approved):

PR #6815, opened 2023-08-15
Commands: add missing dependency on LLBuildManifest

compnerd:
This corrects the missing dependency which happens to work currently due to overlinking.  While trying to improve the support for static linking on Windows, this missing dependency was exposed.


compnerd:
@swift-ci please smoke test

PR #6822, opened 2023-08-16
[5.9.0] Remove extra paths added to LD_LIBRARY_PATH on Linux when running tests

bnbarham:
* **Testing**: Toolchain build and test on all Linux platforms. Manually tested that `swift test` works on a SwiftPM project using a toolchain with this change.

### DIFF
--- a/Sources/Commands/Utilities/TestingSupport.swift
+++ b/Sources/Commands/Utilities/TestingSupport.swift
@@ -141,16 +141,6 @@ enum TestingSupport {
         if let location = toolchain.xctestPath {
             env.prependPath("Path", value: location.pathString)
         }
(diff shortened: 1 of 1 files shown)


bnbarham:
> Risk: Low, our tests would fail if this was still needed.

Note that in the 5.9 cherry-pick @neonichu mentioned:
> I'm not sure that is actually true. When I introduced the workaround, the shipped nightly toolchain was broken which I think implies that the tests were passing despite that being the case.

So I did additionally test that the toolchain with this change is still able to `swift test`.

bnbarham:
@swift-ci please smoke test

DougGregor (approved):

neonichu (approved):

PR #6858, opened 2023-08-24
increase number of concurrent git operations allowed

tomerd:
motivation: better performance

changes: increase the max number of concurrent git operations to 3/4 of active CPUs


tomerd:
@swift-ci smoke test

neonichu (approved):

PR #6876, opened 2023-09-07
Fix missing `from` arguments in PackageDependency's comments

bc-lee:
`.upToNextMajor` and `.upToNextMinor` need a `from` argument, but it was missing in the comments.

### Motivation:

While I was modifying a `Package.swift` file, I noticed that the comments for `.upToNextMajor` and `.upToNextMinor` were missing the `from` argument. So, I'd like to fix that.

### Modifications:

### Result:


MaxDesiatov (approved):
Great catch, thanks!

MaxDesiatov:
@swift-ci test

neonichu (approved):

PR #6901, opened 2023-09-14
Attempt to clarify build and prebuild command documentation

dabrahams:
It's not clear if I've exactly captured the intent of the existing docs.  In particular, the existing docs seem to imply that prebuild commands can only generate `.swift` files and that no prebuild output files will be interpreted as resources.  My proposed changes contradict that implication.

### DIFF
--- a/Documentation/Plugins.md
+++ b/Documentation/Plugins.md
@@ -27,7 +27,7 @@ To get access to a plugin defined in another package, add a package dependency o
 
 ### Making use of a build tool plugin
 
-To make use of a build tool plugin, list its name in each target to which it should apply:
+Add the plugin to the `plugins:` parameter of each target to which it applies:
 
 ```swift
 // swift-tools-version: 5.6
@@ -170,11 +170,15 @@ struct MyPlugin: BuildToolPlugin {
 
 The plugin script can import *Foundation* and other standard libraries, but in the current version of SwiftPM, it cannot import other libraries.
 
+##### Build Commands
+
 In this example, the returned command is of the type `buildCommand`, so it will be incorporated into the build system's command graph and will run if any of the output files are missing or if the contents of any of the input files have changed since the last time the command ran.
 
-Note that build tool plugins are always applied to a target, which is passed in the parameter to the entry point.  Only source module targets have source files, so a plugin that iterates over source files will commonly test that the target it was given conforms to `SourceModuleTarget`.
+The target to which the plugin applies is passed as the `target` parameter.  Only source module targets have source files, so a plugin that iterates over source files will commonly test that the target it was given conforms to `SourceModuleTarget`.
 
-A build tool plugin can also return commands of the type `prebuildCommand`, which run before the build starts and can populate a directory with output files whose names are not known until the command runs:
+##### Prebuild Commands
+
+A build tool plugin can return a combination of build commands and prebuild commands.  A `prebuildCommand` runs after the build tool plugin but before the build starts.  This one populates a `GeneratedFiles/` directory:
 
 ```swift
 import PackagePlugin
@@ -203,13 +207,9 @@ struct MyBuildToolPlugin: BuildToolPlugin {
 }
 ```
 
-In the case of prebuild commands, any dependencies must be binary targets, since these commands run before the build starts.
+A prebuild command has no inputs, so it will never be re-run due to changes in source files. The only trigger for re-running a prebuild command is a change to a declared dependendencies, which can only be (prebuilt) binary targets, since these commands run before any other targets have been built.
 
-Note that a build tool plugin can return a combination of build tool commands and prebuild commands.  After the plugin runs, any build commands are incorporated into the build graph, which may result in changes that require commands to run during the subsequent build.
-
-Any prebuild commands are run after the plugin runs but before the build starts, and any files that are in the prebuild command's declared `outputFilesDirectory` will be evaluated as if they had been source files in the target.  The prebuild command should add or remove files in this directory to reflect the results of having run the command.
-
-The current version of the Swift Package Manager supports generated Swift source files and resources as outputs, but it does not yet support non-Swift source files.  Any generated resources are processed as if they had been declared in the manifest with the `.process()` rule.  The intent is to eventually support any type of file that could have been included as a source file in the target, and to let the plugin provide greater controls over the downstream processing of generated files.
+Any `.swift` files that are outputs of build commands or prebuild commands will be compiled into the target using the plugin. Currently, any other output files are treated as resources and processed as if they had been declared in the manifest with the `.process()` rule.  The intent is to eventually support any type of file that could have been included as a source file in the target, and to let the plugin provide greater controls over the downstream processing of generated files.
 
 ### Command plugins
 


Documentation/Plugins.md
```diff
@@ -203,13 +207,9 @@ struct MyBuildToolPlugin: BuildToolPlugin {
 }
 ```
 
-In the case of prebuild commands, any dependencies must be binary targets, since these commands run before the build starts.
-
-Note that a build tool plugin can return a combination of build tool commands and prebuild commands.  After the plugin runs, any build commands are incorporated into the build graph, which may result in changes that require commands to run during the subsequent build.
-
-Any prebuild commands are run after the plugin runs but before the build starts, and any files that are in the prebuild command's declared `outputFilesDirectory` will be evaluated as if they had been source files in the target.  The prebuild command should add or remove files in this directory to reflect the results of having run the command.
+A prebuild command has no inputs, so it will never be re-run due to changes in source files. The only trigger for re-running a prebuild command is a change to a declared dependendencies, which can only be (prebuilt) binary targets, since these commands run before any other targets have been built.
```
MaxDesiatov, line 210: @tomerd would we ever want to change this behavior? As currently specified it has consequences for https://github.com/apple/swift-package-manager/pull/7053, where we'll need to rerun these prebuild commands when target triple changes. We get the target triple when the build process is initiated and earlier, certainly not during resolution time as

> a change to a declared dependendencies

makes me think.

MaxDesiatov, line 210: typo nit
```suggestion
A prebuild command has no inputs, so it will never be re-run due to changes in source files. The only trigger for re-running a prebuild command is a change to declared dependencies, which can only be (prebuilt) binary targets, since these commands run before any other targets have been built.
```

dabrahams: IIUC this is just a question for @tomerd and doesn't have a bearing on the acceptability of this PR?

Documentation/Plugins.md
```diff
@@ -170,11 +170,15 @@ struct MyPlugin: BuildToolPlugin {
 
 The plugin script can import *Foundation* and other standard libraries, but in the current version of SwiftPM, it cannot import other libraries.
 
+##### Build Commands
+
 In this example, the returned command is of the type `buildCommand`, so it will be incorporated into the build system's command graph and will run if any of the output files are missing or if the contents of any of the input files have changed since the last time the command ran.
 
-Note that build tool plugins are always applied to a target, which is passed in the parameter to the entry point.  Only source module targets have source files, so a plugin that iterates over source files will commonly test that the target it was given conforms to `SourceModuleTarget`.
+The target to which the plugin applies is passed as the `target` parameter.  Only source module targets have source files, so a plugin that iterates over source files will commonly test that the target it was given conforms to `SourceModuleTarget`.
```
tomerd, line 177:
```suggestion
Build tool plugins are always applied to a specific target. The target to which the plugin applies is passed as the `target` parameter.  Only source module targets have source files, so a plugin that iterates over source files will commonly test that the target it was given conforms to `SourceModuleTarget`.
```

Documentation/Plugins.md
tomerd: I think that is def worth considering changing, such that there are more conditions which trigger re-running rebuild commands.

MaxDesiatov: IDK, specifying that "The only trigger for re-running a prebuild command is a change to declared dependencies" feels too strong to me if we have to change this behavior in the process of my work on https://github.com/apple/swift-package-manager/pull/7053

Documentation/Plugins.md
```diff
@@ -27,7 +27,7 @@ To get access to a plugin defined in another package, add a package dependency o
 
 ### Making use of a build tool plugin
 
-To make use of a build tool plugin, list its name in each target to which it should apply:
+Add the plugin to the `plugins:` parameter of each target to which it applies:
```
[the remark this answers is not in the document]
dabrahams: Why is that an improvement?  It is wordier at the least and to me the extra words only seem to make room for misunderstanding.

Documentation/Plugins.md
dabrahams: I don't understand.  If you want to change the behavior of the system you can change the documentation appropriately.  Why the reluctance to precisely document the current behavior?

Documentation/Plugins.md
```diff
@@ -170,11 +170,15 @@ struct MyPlugin: BuildToolPlugin {
...
 
-Note that build tool plugins are always applied to a target, which is passed in the parameter to the entry point.  Only source module targets have source files, so a plugin that iterates over source files will commonly test that the target it was given conforms to `SourceModuleTarget`.
+The target to which the plugin applies is passed as the `target` parameter.  Only source module targets have source files, so a plugin that iterates over source files will commonly test that the target it was given conforms to `SourceModuleTarget`.
+
+##### Prebuild Commands
 
-A build tool plugin can also return commands of the type `prebuildCommand`, which run before the build starts and can populate a directory with output files whose names are not known until the command runs:
+A build tool plugin can return a combination of build commands and prebuild commands.  A `prebuildCommand` runs after the build tool plugin but before the build starts.  This one populates a `GeneratedFiles/` directory:
```
tomerd, line 181: the original text said

```
Any prebuild commands are run after the plugin runs but before the build starts, and any files that are in the prebuild command's declared `outputFilesDirectory` will be evaluated as if they had been source files in the target. The prebuild command should add or remove files in this directory to reflect the results of having run the command
```
@neonichu @abertelrud which one is true?

Documentation/Plugins.md
```diff
@@ -203,13 +207,9 @@ struct MyBuildToolPlugin: BuildToolPlugin {
...
 
-The current version of the Swift Package Manager supports generated Swift source files and resources as outputs, but it does not yet support non-Swift source files.  Any generated resources are processed as if they had been declared in the manifest with the `.process()` rule.  The intent is to eventually support any type of file that could have been included as a source file in the target, and to let the plugin provide greater controls over the downstream processing of generated files.
+Any `.swift` files that are outputs of build commands or prebuild commands will be compiled into the target using the plugin. Currently, any other output files are treated as resources and processed as if they had been declared in the manifest with the `.process()` rule.  The intent is to eventually support any type of file that could have been included as a source file in the target, and to let the plugin provide greater controls over the downstream processing of generated files.
```
tomerd, line 212:
```suggestion
The current version of the Swift Package Manager supports generated Swift source files and resources as outputs, but it does not yet support non-Swift source files. Any `.swift` files that are outputs of build commands or prebuild commands will be compiled into the target using the plugin. Currently, any other output files (including non-Swift source files) are treated as resources and processed as if they had been declared in the manifest with the `.process()` rule.  The intent is to eventually support any type of file that could have been included as a source file in the target, and to let the plugin provide greater controls over the downstream processing of generated files.
```

tomerd:
thanks @dabrahams added some comments, one factual question regarding the output directory of prebuilt command  we need input on

Documentation/Plugins.md
tomerd: agree with Dave the doc should reflect the state of things and then updated when behavior changes

Documentation/Plugins.md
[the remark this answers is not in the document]
tomerd: this is of course subjective, but I think it is clearer.

Documentation/Plugins.md
tomerd: this is of course subjective, but I think it is clearer.

Documentation/Plugins.md
dabrahams: Is GitHub lying to me?  I do remember the words you're quoting, but the diff shows that this was the original text:

> A build tool plugin can also return commands of the type `prebuildCommand`, which run before the build starts and can populate a directory with output files whose names are not known until the command runs.

Regardless, any files whose extensions don't match known source file extensions are in fact evaluated as though they were listed as resources, which makes the text you cited wrong.

tomerd: mostly curious about `outputFilesDirectory` vs `This one populates a GeneratedFiles/`

the quote is from a paragraph down below that was deleted as part of the suggested changes

Documentation/Plugins.md
dabrahams: Using the phrase “non-Swift source files” or “Swift source files” concerns me, because the actual issue is not what programming language is used in the file, but the file's extension.  What I wrote was absolutely unambiguous, but if you put C source in a `.swift` file, it will not in fact be treated as a resource, and if you put Swift source in a `.foo` file, it will, and that conflicts with the language you've added.

Documentation/Plugins.md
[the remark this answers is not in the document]
dabrahams: Please explain how and why. 

I can explain how it hurts:

- “list a name in a target” is not a thing, and "list a name in a target…using the plugins parameter” is a very roundabout way of saying “pass it as an element of the plugins parameter.”  It requires more thinking from a reader to comprehend.
- “should apply” begs the question of what “should” means.

Documentation/Plugins.md
tomerd: IMO this additional sentence sets the context clearly to the rest of the paragraph.

Documentation/Plugins.md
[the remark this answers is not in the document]
tomerd: not a strong opinion about this one. I personally found the proposed sentence "Add the plugin to the `plugins:` parameter of each target to which it applies:" too terse, and was attempting to make it less so by reusing some of the original language. Maybe a good path forward could be:

"Build tool plugin operate on targets. Add the plugin to the `plugins:` parameter of each target to which it applies"

Documentation/Plugins.md
tomerd: How about 

The current version of the Swift Package Manager supports generated .swift files (which will be treated as Swift source) and resources as outputs, but it does not yet support other types of source files.

Documentation/Plugins.md
dabrahams: I understand why you're saying that.  However making that statement by itself raises the questions, “as opposed to what?  why are you telling me this? am I supposed to imagine what it might mean to have a build tool that doesn't apply to a target so I can rule out some misconception I may be harboring?”  It raises more questions than it answers, and the reader doesn't need this dangling fact in order to understand the phrase "The target to which the plugin applies."

**If** you're going to say this, it belongs in a section about _what a build tool is_, not in this section which is about _how to use a build tool_.

Documentation/Plugins.md
abertelrud: I agree as well — the documentation should match the implementation, and change when the behaviour changes.

Is it true today that today's implementation doesn't run prebuild commands at the start of every build?  That sounds different from the original intent.  IIRC, the only reason prebuild commands exist in the first place is because the two primary build systems for Swift Packages (`swift build` and `xcbuild`) don't support having a command generate an arbitrary set of output files based on the contents of its inputs.  So we needed to have some command that can always run at the start of every build to generate arbitrary output files, using caches to avoid unnecessary work.

Documentation/Plugins.md
abertelrud: Whatever directory the plugin passes in the `outputFilesDirectory` parameter is the one that will be checked.  There is nothing special about the name "GeneratedFiles" — that's just what the example happens to use as the name of its output directory.

abertelrud:
> thanks @dabrahams added some comments, one factual question regarding the output directory of prebuilt command we need input on

Not sure I understand the conflict ("which one is true").  There is no assumption about the name of the directory, so whatever the plugin passes in the `outputFilesDirectory:` parameter when creating the `prebuild` command is used.  The example in the documentation happens to use one called "GeneratedFiles", but there's not conflict there; that's just an example.

Documentation/Plugins.md
tomerd: okay, then I think the  suggested documentation should point out usage of outputFilesDirectory (as done originally) and highlight that  GeneratedFiles is just an example.

Documentation/Plugins.md
tomerd: sure. you are obviously welcome to come up with a different way to make this + the other sentence above a little less terse - that is my main feedback about both of them

dabrahams: Then there are two issues we need to sort out:

1. I remember now why I removed the original wording, aside from the fact that they are extra: if you read them as though they must be adding information, they are actually wrong.  As phrased, it makes it sound as though each build tool plugin applies to one specific target.  In fact, a build tool plugin can “apply to” (a better phrase would be “be used to build”) any number of targets, including zero.   If you want me to say something about that, I can do it in the appropriate place (the section about what a build tool is), although I think any framing of “what a build tool is” is going to make that patently obvious.  The bigger problem is that it won't do anything to make *this sentence* less terse, which is what you say you want.
2. You and I seem to have utterly different value systems w.r.t. concision in documentation.  For me, concision is a virtue, provided no necessary information is omitted, out of order, or ambiguously presented.  As a user, every word I have to process and sort out to understand the meaning of documentation makes my job harder.  Therefore, I have no understanding of how to make the docs less terse without making them actively worse.  For me, and I'm not exaggerating, it would be better for me to add words like “@tomerd thought this was too terse so please insert some words here that don't change the meaning,” because anything else leads the reader to wonder what meaning they are supposed to extract from the additional words.  What I suggest, if you don't want to simply reject this PR, is that you accept some phrasing you consider to be accurate but “too terse” and then you make an additional change yourself so that you're comfortable with the style.

Documentation/Plugins.md
dabrahams: Let me see if I can understand what that implies for this PR… back soon with a proposed update.

Documentation/Plugins.md
dabrahams: It's OK by itself, but then when you put it with the next sentence it's redundant and verbose.  It seems like the only thing you're trying to add is an _explicit_ mention that compiling output files in languages other than Swift isn't supported.  Does this accomplish what you want?
 
```suggestion
Any `.swift` files that are outputs of build commands or prebuild commands will be treated as Swift source files and compiled into the target being build by the plugin. Currently, compilation of output files in other source langauges isn't supported, so any other output files are treated as resources and processed as if they had been declared in the manifest with the `.process()` rule.  The intent is to eventually support any type of file that could have been included as a source file in the target, and to let the plugin provide greater controls over the downstream processing of generated files.
```

BTW, I don't understand what “and to let the plugin provide greater controls over the downstream processing of generated files” might actually mean, so if I were you I would consider striking or clarifying it.  The only reason I didn't do that is because in this context, where you're just talking about a future SPM that doesn't exist yet, I can easily know the information isn't relevant to me.  I would also suggest you consider moving the whole sentence into a footnote, because it really is irrelevant to almost anyone who is trying to understand what a build tool is and what it can do.

tomerd: > Does this accomplish what you want?

yes

Documentation/Plugins.md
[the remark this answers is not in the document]
dabrahams: That doesn't make the sentence less terse. It does make the paragraph less terse, but again this information is misplaced.  The relationship between targets and build tool plugins doesn't belong in the instructions about how to use them; it belongs in a discussion of what build tool plugins are.

Documentation/Plugins.md
dabrahams: The original pointing out of `outputFilesDirectory` was in an inaccurate context, and disconnected from the language on line 212 that covered the same topic.  The new language that covers that information is in [this paragraph](https://github.com/apple/swift-package-manager/pull/6901/commits/370b26cd797139e8944777d56d42cc66bd9ec232).  It doesn't mention `outputFilesDirectory` but instead mentions the output files themselves.  I would expect that if the build tool left files in `outputFilesDirectory` directory that were not in the output files list it would be an error (otherwise, what's the point of having an output files list at all?)  Are you sure this needs to change?  If so, what change do you want?

Documentation/Plugins.md
dabrahams: If that's the case I should obviously change the wording to reflect it… but that said I see no reason for them to run before every build if the only goal is to “generate files whose names can't be predicted from inputs.”  You can do that once and record a marker that it's been done, then use those files over and over until they are cleaned or declared dependencies of the prebuild command change.  If running before every build is necessary, there must be some other, currently unstated motivation for that requirement.  In fact, I extrapolated the behavior I documented from what I had been told about a prebuild command's reason for existence and the most efficient way to achieve that.

So that I document the right thing: are they run before *every* build of *every* target, or are they only run before builds of targets that are using the plugin?

Documentation/Plugins.md
tomerd: happy to hear other opinions, I am not sure if I can add a new signal beyond my subjective opinion that I have already expressed

Documentation/Plugins.md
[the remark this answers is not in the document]
tomerd: https://github.com/apple/swift-package-manager/pull/6901#discussion_r1387283074

Documentation/Plugins.md
dabrahams: And I respect that.  If you are unwilling to solve the problem by doing what I suggest in №2 above, then I need a specific suggestion that makes the sentence less terse and isn't actively wrong, because as I wrote, I don't understand how to satisfy your desire.

Documentation/Plugins.md
tomerd: my original point was that the proposed text talks about the concrete example below (GeneratedFiles) and not about the general case (whatever is specified in outputFilesDirectory). IMO the general text should describe the functionality more generally, and not just the example. does that make sense?

maybe something like the following (combination of the new + old text):

> A build tool plugin can return a combination of build commands and prebuild commands. Any prebuild commands are run after the build tool plugin but before the build. Any files that are generate in the prebuild command's declared `outputFilesDirectory` will be evaluated as if they had been source files in the target. 
>
> In the following example, the `prebuildCommand` populates a `GeneratedFiles/` directory:

Documentation/Plugins.md
tomerd: generally, I would like to avoid rejecting PRs and instead work with the authors to fine tune the proposed changes such that everyone involved is satisfied.

perhaps a path forward could be to ask feedback from a few other folks, maybe they can help us find the right words that will make it less terse but still not over-wordy.

cc @abertelrud @TimTr @bitjammer

tomerd: cc @amartini51

TimTr: I tend to agree that the additional wording "_Build tool plugins are always applied to a specific target_" felt like it was trying to document things beyond this parameter, which is out of scope. Documenting "bigger" stuff than the current scope can get out of hand quickly, and could be done for almost any API or parameter. In this case, the use feels pretty self-explanatory to me.

Perhaps it is useful to add an adjective like "mandatory" in this case, as perhaps that solves the suggester's concern? Seems like there is a desire for emphasis more than additional, external documentation.

tomerd: @TimTr please note this is not API docs - this is a user guide

tomerd: @dabrahams I read the entire section again starting from "Implementing the build tool plugin script" and then the two "Build Commands" and "Prebuild Commands" sub-section. The latter seem to be focused on explaining the examples for build and pre-build while the paragraphs above give a more broad overview of how build tool plugins work. In the light, I would argue the whole paragraph "The target to which the plugin applies is passed as the target parameter. Only source module targets have source files, so a plugin that iterates over source files will commonly test that the target it was given conforms to SourceModuleTarget" should be hoisted to the top part (above  "Build Commands" sub-section) as it is not specific to the example. At that point we can decide if we want to give additional context such as "SwiftPM will provide context information when invoking the plugin, through the context and target parameters..."

Issue #6865, opened 2023-08-28
Specify registry specific .netrc location in ~/.swiftpm/configuration/registries.json


dafangc-amazon:
### Description

Currently we need to specify the `--netrc-file` option in every swift command to use the .netrc file on a custom path. It would be much more convenient if we only had to specify `--netrc-file` in the `swift package-registry login` command and Swift would automatically store the custom path in registries.json for reference. 

It would benefit users who don't want to store every credential in ~/.netrc, because they could use custom .netrc more easily without having to specify the path whenever resolving dependencies or building the project.

### Expected behavior

Swift would add new field in the json similar to below:
```
{
  "authentication" : {
    "https://example.com" : {
      "loginAPIPath" : "/registry-1/login",
      "type" : "token"
      "netrcPath": "/custom/path/.netrc"
    }
  },
  "registries" : {
    "default" : {
      "url" : "hhttps://example.com/registry-1"
    }
  },
  "version" : 1
}

### Actual behavior

_No response_

### Steps to reproduce

_No response_

### Swift Package Manager version/commit hash

_No response_

### Swift & OS version (output of `swift --version && uname -a`)

```
Swift version 5.8.1 (swift-5.8.1-RELEASE)
Target: x86_64-unknown-linux-gnu
```
```
5.4.250-173.369.amzn2int.x86_64
```

yim-lee:
Another approach that we might consider is for SwiftPM to support custom auth provider that can be plugged in and used instead of `netrc` and Keychain.

yim-lee:
Discussed with @dafangc-amazon/team. This is not needed for now, so closing this. Please reopen if there is a need for it in the future.

PR #6903, opened 2023-09-15
[WIP] Specify registry specific credentials file in registries.json

yim-lee:
Motivation:
Current supported credential stores (namely Keychain and netrc file) are intended for long-lasting credentials (weeks/months vs. minutes/hours). For registries that use short-lived access tokens, frequent updates are made to the credential store. Using netrc file then becomes challenging, because SwiftPM doesn't do in-place edit of an existing entry but always appends to the end of file and uses the last match. The netrc file would become bloated very quickly.

https://github.com/apple/swift-package-manager/issues/6865 rdar://114607754

Modifications:

- Add new `swiftpmrc` format which is similar to `netrc` but it's in JSON.

```
        {
          "machines": [
            {
              "name": "example.com",
              "login": "anonymous",
              "password": "qwerty"
            },
            {
              "name": "example.com:8080",
              "password": "secret-token"
            }
          ],
          "version": 1
        }
```

 - Add option `swiftpmrcPath` to `authentication` block in `registries.json`. Short-lived tokens can be written to a `swiftpmrc` file and overwritten as needed without worrying that it's shared across apps like `netrc` does.
 - Update `RegistryClient` to read from `swiftpmrc` file if specified
 - Update `package-registry login` subcommand to support `--swiftpmrc-file` option

### DIFF
--- a/Sources/Basics/CMakeLists.txt
+++ b/Sources/Basics/CMakeLists.txt
@@ -1,6 +1,6 @@
 # This source file is part of the Swift open source project
 #
-# Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
+# Copyright (c) 2014 - 2023 Apple Inc. and the Swift project authors
 # Licensed under Apache License v2.0 with Runtime Library Exception
 #
 # See http://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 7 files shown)


tomerd (approved):
nice!

yim-lee:
This is not needed for now. Closing.

PR #6912, opened 2023-09-17
Modified some functions to computed properties for API style

jaesung-0o0:
I modified some functions to computed properties for better API design.

### Motivation:

These methods returns a state or a simple value rather than performing an action such as `isDarwin()`, `isApple()` or `withoutVersion`
I modified some methods to the computed properties for better API design.
However, these changes depend on your style.

### Modifications:

**DispatchTimeInterval**

| As-is | To-be |
| --- | --- |
| timeInterval() | timeInterval |
| nanoseconds() | nanoseconds |
| milliseconds() | milliseconds |
| seconds() | seconds |

**Tripple extensions**
| As-is | To-be |
| --- | --- |
| isApple() | isApple |
| isAndroid() | isAndroid |
| isDarwin() | isDarwin |
| isLinux() | isLinux |
| isWindows() | isWindows |
| isWASI() | isWASI |
| isOpenBSD() | isOpenBSD |
| withoutVersion() | withoutVersion |

### DIFF
--- a/Sources/Basics/Cancellator.swift
+++ b/Sources/Basics/Cancellator.swift
@@ -151,7 +151,7 @@ public final class Cancellator: Cancellable {
 
         let deadline = deadline ?? .now() + .seconds(30)
         // deadline for individual handlers set slightly before overall deadline
-        let delta: DispatchTimeInterval = .nanoseconds(abs(deadline.distance(to: .now()).nanoseconds() ?? 0) / 5)
+        let delta: DispatchTimeInterval = .nanoseconds(abs(deadline.distance(to: .now()).nanoseconds ?? 0) / 5)
         let handlersDeadline = deadline - delta
 
         let cancellationHandlers = self.registry.get()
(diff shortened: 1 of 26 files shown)


rauhul:
I would expect computed properties to behave near identically to simple functions, with the only difference coming from changes in inlining. Have you measured any performance improvement from this change?

jaesung-0o0:
> I would expect computed properties to behave near identically to simple functions, with the only difference coming from changes in inlining. Have you measured any performance improvement from this change?

Oh you're right. There's technically no difference in performance in this PR. My initial description had incorrect information on the modification.

The modification is related to the API style, not the performance.

Some methods return a state or a simple value rather than performing an action. Similar to `isEmpty` in collection types, using computed properties feels like a better API style.

tomerd:
personally I prefer the computed property style for these

cc @neonichu

rauhul:
I do as well, but I would recommend we figure out a better story for where to put a swift-triple package/library instead of maintaining a duplicate codebase with the swift-driver before making the triple changes (imo)

MaxDesiatov:
Additionally, I don't think that changing a function to a property with the same name is source-compatible, and that would break users of libSwiftPM if not done in a graceful manner.

jaesung-0o0:
Even if the previous are not officially offered, since they are public, 
is it unacceptable? or requires more discussion?

PR #6925, opened 2023-09-19
[5.10] Fix backward compatibility in host triple checks

MaxDesiatov:
Cherry-pick of https://github.com/apple/swift-package-manager/pull/6819.

When setting host triple in Swift SDKs to `arm64-apple-macosx13.0` to allow cross-compiling from an older version of macOS, this triple is not recognized as directly matching `arm64-apple-macosx14.0` on a newer version of macOS.

We should support backward compatibility with Swift SDKs that were built for older version of macOS.

Resolves rdar://113967401.

```
# Conflicts:
#	Sources/PackageModel/SwiftSDKBundle.swift
```

### DIFF
--- a/Sources/Basics/Triple+Basics.swift
+++ b/Sources/Basics/Triple+Basics.swift
@@ -183,6 +183,19 @@ extension Triple {
             return ".resources"
         }
     }
+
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
@swift-ci test

tomerd (approved):

PR #6926, opened 2023-09-19
[NFC] Update SwiftPM-Package.xcscheme

MaxDesiatov:
Since this file hasn't been added to `.gitignore`, we have to update it after the list of targets is updated.


MaxDesiatov:
@swift-ci test windows

PR #6918, opened 2023-09-18
Fix remaining mentions of "cross-compilation destinations"

MaxDesiatov:
Not all of the help strings were updated to use "Swift SDK" instead of "cross-compilation destination" after SE-0387 was accepted.

rdar://115680101

PR #6927, opened 2023-09-19
Rename "destination" to "Swift SDK" in comments and code

MaxDesiatov:
While user-visible strings were cleaned up separately in https://github.com/apple/swift-package-manager/pull/6918, there are still uses of "destinations" left in variable/function naming and comments. Let's clean those up too.


neonichu (approved):

rauhul:
thank you for this, the mix of names made the code much more confusing to try to understand

PR #6928, opened 2023-09-19
[5.10] Rename "destination" to "Swift SDK" in comments and code

MaxDesiatov:
Cherry-pick of #6927.

While user-visible strings were cleaned up separately in https://github.com/apple/swift-package-manager/pull/6918, there are still uses of "destinations" left in variable/function naming and comments. Let's clean those up too.


MaxDesiatov:
@swift-ci test

tomerd (approved):

PR #6955, opened 2023-09-29
do not emit "exhausted attempts to resolve the dependencies graph" when previous errors occurred

tomerd:
motivation: better diagnostics

changes:
* check for previous errors when updating dependencies and terminate early if core update/checkout dependencies error occurred
* return state changes so that dry run actually works


neonichu (approved):
👍

tomerd:
@swift-ci smoke test

tomerd:
@swift-ci test

PR #6960, opened 2023-09-29
improve validation of existing checkouts

tomerd:
motivation: fix edge cases where checkouts point to the wrong repository

changes:
* update WorkingCheckout::isAlternateObjectStoreValid to take an expected repository path
* create an API on RepositoryManager to make sure the checkout aligns with the locall repository

rdar://112134161

### DIFF
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -743,7 +743,7 @@ public final class GitRepository: Repository, WorkingCheckout {
     }
 
     /// Returns true if there is an alternative object store in the repository and it is valid.
-    public func isAlternateObjectStoreValid() -> Bool {
+    public func isAlternateObjectStoreValid(expected: AbsolutePath) -> Bool {
         let objectStoreFile = self.path.appending(components: ".git", "objects", "info", "alternates")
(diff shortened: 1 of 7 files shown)


tomerd:
@swift-ci test

Sources/SourceControl/GitRepository.swift
```diff
@@ -752,7 +752,11 @@ public final class GitRepository: Repository, WorkingCheckout {
         guard let firstLine = ByteString(split[0]).validDescription else {
             return false
         }
-        return (try? localFileSystem.isDirectory(AbsolutePath(validating: firstLine))) == true
+        guard let objectsPath = try? AbsolutePath(validating: firstLine), localFileSystem.isDirectory(objectsPath) else {
+            return false
+        }
+        let repositoryPath = objectsPath.parentDirectory
+        return expected == repositoryPath
```
tomerd, line 759: this is the key change. isAlternateObjectStoreValid now also checks that the "pointer" the repo is valid not just exists. before you could easily run into a situation where the checkout would point to a valid path but outside the source tree and not the one we would run "update" on

neonichu (approved):

PR #6971, opened 2023-10-04
[NFC] Build/Commands: fix `triple` property deprecation warnings

MaxDesiatov:
As `triple` was deprecated with the recommendation to use explicit `targetTriple` property instead, we should update all of our code to get rid of the remaining deprecation warnings.


MaxDesiatov:
@swift-ci test

neonichu (approved):

PR #6975, opened 2023-10-05
[NFC] LLBuildManifest: add `WriteAuxiliary.EntitlementPlist`

MaxDesiatov:
### Motivation:

For enabling backtraces on macOS we need to grant a certain entitlement on executables built with SwiftPM. For that we need to write out a plist file with this entitlement setting.

This new auxiliary file type allows writing an entitlements plist file during the build process.

Related to rdar://112065568.

### Modifications:

Added `WriteAuxiliary.EntitlementPlist` with a corresponding test, moved `LLBuildManifestTests` into its own test target that depends only on the `LLBuildManifest` target and few required targets such as `Basics` and test support target.

### Result:

A corresponding build step can be added to build description and build planning in a subsequent PR.

### DIFF
--- a/.swiftpm/xcode/xcshareddata/xcschemes/SwiftPM-Package.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/SwiftPM-Package.xcscheme
@@ -829,6 +829,16 @@
                ReferencedContainer = "container:">
             </BuildableReference>
          </TestableReference>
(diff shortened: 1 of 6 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test linux

neonichu (approved):

PR #6979, opened 2023-10-06
Add readFileInChunks

spqw:
Add function to read archive in chunks when publishing archive for registry

### Motivation:

This helps with memory efficiency and performance for large files. 
This is implementing a TODO.

### Modifications:

Before: The file is read with readFileContents, loading the entire file in memory
Modification: A readFileInChunks function was added, using an InputStream.

### Result:

The archive is read in chunks


spqw:
@swift-ci please test

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -2441,3 +2441,35 @@ extension URLComponents {
         path += (path.last == "/" ? "" : "/") + components.joined(separator: "/")
     }
 }
+
+extension FileSystem {
+    /// Get the contents of a file without loading all of it in memory
+    ///
+    /// - Returns: The file contents as Data, or nil if missing
+    func readFileInChunks(path: AbsolutePath, bufferSize: Int = 1024) -> Data? {
```
MaxDesiatov, line 2449: Sorry, I'm not sure I understand what this is trying to achieve. How exactly would this prevent reading the whole file into memory if the resulting `Data` value contains the whole file in memory by definition?

MaxDesiatov:
Hi @spqw, would you be interested in addressing the PR review provided above?

Sources/PackageRegistry/RegistryClient.swift
spqw: You're totally right. 
My bad, this is actually far from being ready for review. 
Marking it as a draft for now if that's ok.

PR #6984, opened 2023-10-10
[NFC] SPMBuildCore: make `BuildSystemProvider.Provider` a protocol

MaxDesiatov:
### Motivation:

When expressed as a closure, it confuses the type checker when working on any related code. Multi-line closure type inference has been known to be suboptimal in Swift, and converting it to a protocol localizes "Type of expression is ambiguous without a type annotation" to a specific line that causes the error instead of attaching it to the whole closure.

### Modifications:

Created a new `protocol BuildSystemFactory` with a `makeBuildSystem` method that replaces the previous `BuildSystemProvider.Provider` closure typealias. Added new `private struct NativeBuildSystemFactory` and `private struct XcodeBuildSystemFactory` in `BuildSystemSupport.swift`, which conform to the new protocol and are used in the implementation of `defaultBuildSystemProvider` computed property.

### Result:

It's much easier to work with `BuildSystemFactory` protocol: it no longer confuses the type checker when changes to the closure body are made or need to be made. Additionally, the new protocol method can use named function arguments instead of unnamed ones, since closures don't support the former.

### DIFF
--- a/Sources/CoreCommands/BuildSystemSupport.swift
+++ b/Sources/CoreCommands/BuildSystemSupport.swift
@@ -19,41 +19,67 @@ import struct PackageGraph.PackageGraph
 import struct PackageLoading.FileRuleDescription
 import protocol TSCBasic.OutputByteStream
 
+private struct NativeBuildSystemFactory: BuildSystemFactory {
+    let swiftTool: SwiftTool
+
+    func makeBuildSystem(
+        explicitProduct: String?,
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

neonichu (approved):

PR #7009, opened 2023-10-16
Avoid using `temp_await` on `loadRootPackage`

MaxDesiatov:
### Motivation:

Using locks to convert asynchronous functions to synchronous is dangerous, since this can deadlock the Swift Concurrency thread pool. Unfortunately, `temp_await` (AKA `tsc_await`) function utilizes exactly this approach. 

As SourceKit-LSP starts adopting Swift Concurrency (https://github.com/apple/sourcekit-lsp/pull/850), we should avoid using `temp_await` in our codebase and make the callers of corresponding `async` or callback-taking functions themselves `async`. Otherwise there's a risk of deadlocks in SourceKit-LSP itself, since their Swift Concurrency tasks may be blocked by our uses of `temp_await` under the hood. We have ~60 uses of `temp_await` in our codebase, carefully removing them in small portions will reduce the potential disruption.

### Modifications:

`loadRootPackage` already has an `async` overload in `Workspace`. Converted callers in SwiftPM commands to use `AsyncSwiftCommand` protocol instead of `SwiftCommand`. Also, test cases were made `async` too.

### Result:

Reduced the number of `temp_await` uses in `Source`.


Tests/WorkspaceTests/WorkspaceTests.swift
```diff
@@ -8803,13 +8800,15 @@ final class WorkspaceTests: XCTestCase {
 
         let observability = ObservabilitySystem.makeForTesting()
         let wks = try workspace.getOrCreateWorkspace()
-        XCTAssertNoThrow(try temp_await {
```
MaxDesiatov, line 8806: `XCTAssertNoThrow` doesn't support `async` 🥲

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@compnerd would you be able to confirm that this change is fine from the perspective of Swift Concurrency support on Windows?

compnerd:
> @compnerd would you be able to confirm that this change is fine from the perspective of Swift Concurrency support on Windows?

I think that as long as we restrict this to 5.10/5.11 it _should_ be okay.  Of course we should run through the test suite as well.

MaxDesiatov:
There are no plans to backport it to 5.9. How far are we from enabling the test suite on Windows CI? Until then is the only way to test it on Windows to build latest 5.10 locally and to run `swift test` locally too?

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/SPMTestSupport/MockWorkspace.swift
```diff
@@ -510,13 +510,16 @@ public final class MockWorkspace {
         public let diagnostics: [Basics.Diagnostic]
     }
 
-    public func checkPrecomputeResolution(_ check: (ResolutionPrecomputationResult) -> Void) throws {
```
AndrewHoos, line 513: This method should be full async. It makes more sense to have it just return the `ResolutionPrecomputationResult` result and let the user do the checking inline

AndrewHoos (changes requested):
It all looks great except for `checkPrecomputeResolution`. I don't see why the `checkPrecomputeResolution` method needs to have a closure. Closures normally are for async callbacks, context managers, or optional extra  validation (e.g. assert throws). It seems like that API should get reworked to by natively async

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

AndrewHoos (approved):

Sources/Commands/PackageTools/Format.swift
```diff
@@ -71,7 +69,7 @@ extension SwiftPackageTool {
             let args = [swiftFormat.pathString] + formatOptions + [packagePath.pathString] + paths
             print("Running:", args.map{ $0.spm_shellEscaped() }.joined(separator: " "))
 
-            let result = try TSCBasic.Process.popen(arguments: args)
+            let result = try await TSCBasic.Process.popen(arguments: args)
```
AndrewHoos, line 72: I haven't looked, but I wonder if this is actually non-blocking

AndrewHoos (approved):

Sources/Commands/PackageTools/Format.swift
MaxDesiatov: Good catch, I'm cleaning that up on TSC side now.

MaxDesiatov:
@swift-ci test windows

Sources/Commands/PackageTools/Format.swift
MaxDesiatov: Wrapped that with a concurrent queue in https://github.com/apple/swift-tools-support-core/pull/450

MaxDesiatov:
@swift-ci test windows

Sources/Commands/PackageTools/Format.swift
AndrewHoos: 🤣 ask me if I have deadlocked the swift concurrency pool before...

AndrewHoos: Do you want to use the safe_async wrapper until that lands?

MaxDesiatov: I've enabled auto-merge on that one, so I don't think there will be a significant amount of time when this PR is merged, but one in TSC isn't.

PR #7011, opened 2023-10-16
Tests: mitigate progressivity issue in LegacyHTTPClientTests

kateinoigakukun:
The legacy client can block several threads by semaphore and the test case itself blocks a thread per request. The test case waits until all requests go in the middle of the process, but some of requests won't be started due to a lack of available threads on a machine with a few cores. Reducing the number of concurrent requests does not solve the root issue (it ultimately requires lock/wait free implementation and I think it's not worth it just for legacy one), but should be enough to repair CI.


kateinoigakukun:
@swift-ci Please test

tomerd:
probably should be based on activeCoreCount instead of hard coded value

kateinoigakukun:
Sounds better to me :+1: Updated using physical core count while limiting up to 10

kateinoigakukun:
@swift-ci Please test

tomerd (approved):

PR #7012, opened 2023-10-17
Remove outdated check for Ubuntu 14.04 and Clang 3.6

MaxDesiatov:
Swift no longer supports Ubuntu 14.04, the latest supported version is 18.04, so this check should no longer be needed.


MaxDesiatov:
@swift-ci test

neonichu (approved):

PR #7018, opened 2023-10-18
SwiftSyntax version in new macro template has trailing period in dependency version, causing error reading manifest

stmontgomery:
This fixes a regression I believe was introduced in #6774 where the new macro package template has its dependency on swift-syntax expressed as `from: "509.0.0."`, including an unncessary trailing period character. This causes an error reading the manifest.

I have not tested this, but I experienced the bug and from source inspection believe this is likely the issue.

Resolves rdar://117132800

### DIFF
--- a/Sources/PackageModel/InstalledSwiftPMConfiguration.swift
+++ b/Sources/PackageModel/InstalledSwiftPMConfiguration.swift
@@ -25,7 +25,7 @@ public struct InstalledSwiftPMConfiguration: Codable {
         }
 
         public var description: String {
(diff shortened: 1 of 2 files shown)


MaxDesiatov (changes requested):
Would you mind adding a test for this in `WorkspaceTests/InitTests.swift`?

stmontgomery:
> Would you mind adding a test for this in `WorkspaceTests/InitTests.swift`?

It's not precisely what you asked for, but I just added a few very targeted unit tests as a start.

neonichu (approved):

PR #7020, opened 2023-10-18
llbuild: Support `is-mutated`/`is-command-timestamp` nodes

MaxDesiatov:
Dependency of https://github.com/apple/swift-package-manager/pull/7010.

Added per the llbuild documentation at https://github.com/apple/swift-llbuild/blob/a3b8d34be319f1b3ba8de53c434386419dec1f4f/docs/buildsystem.rst#L387 and a corresponding test at https://github.com/apple/swift-llbuild/blob/a3b8d34be319f1b3ba8de53c434386419dec1f4f/tests/BuildSystem/Build/mutable-outputs.llbuild#L66.

Newly introduced attributes are not used anywhere yet in this PR, so technically it is NFC.

I also took the opportunity to get rid of deprecated TSC dependencies in `ManifestWriter.swift`, replacing it with string interpolation instead.

### DIFF
--- a/Sources/LLBuildManifest/ManifestWriter.swift
+++ b/Sources/LLBuildManifest/ManifestWriter.swift
@@ -23,14 +23,8 @@ public struct ManifestWriter {
 
     """
 
-    init(manifest: BuildManifest, buffer: String = """
-    client:
-      name: basic
-    tools: {}
-
-    """) {
+    init(manifest: BuildManifest) {
         self.manifest = manifest
-        self.buffer = buffer
 
         self.render(targets: manifest.targets)
 
--- a/Sources/LLBuildManifest/ManifestWriter.swift
+++ b/Sources/LLBuildManifest/ManifestWriter.swift
@@ -11,63 +11,108 @@
 //===----------------------------------------------------------------------===//
 
 import Basics
-import protocol TSCBasic.ByteStreamable
-import struct TSCBasic.Format
-import protocol TSCBasic.OutputByteStream
-import class TSCBasic.BufferedOutputByteStream
+
+private let namesToExclude = [".git", ".build"]
 
 public struct ManifestWriter {
-    let fileSystem: FileSystem
+    private let manifest: BuildManifest
+    private var buffer = """
+    client:
+      name: basic
+    tools: {}
 
-    public init(fileSystem: FileSystem) {
-        self.fileSystem = fileSystem
+    """
+
+    init(manifest: BuildManifest, buffer: String = """
+    client:
+      name: basic
+    tools: {}
+
+    """) {
+        self.manifest = manifest
+        self.buffer = buffer
+
+        self.render(targets: manifest.targets)
+
+        self.buffer += "default: \(manifest.defaultTarget.asJSON)\n"
+
+        self.render(nodes: manifest.commands.values.flatMap { $0.tool.inputs + $0.tool.outputs })
+
+        self.render(commands: manifest.commands)
     }
 
-    public func write(
-        _ manifest: BuildManifest,
-        at path: AbsolutePath
-    ) throws {
-        let stream = BufferedOutputByteStream()
-        stream.send(
-            """
-            client:
-              name: basic
-            tools: {}
-            targets:\n
-            """
-        )
+    public static func write(_ manifest: BuildManifest, at path: AbsolutePath, _ fileSystem: FileSystem) throws {
+        let writer = ManifestWriter(manifest: manifest)
 
-        for (_, target) in manifest.targets.sorted(by: { $0.key < $1.key }) {
-            stream.send("  ").send(Format.asJSON(target.name))
-            stream.send(": ").send(Format.asJSON(target.nodes.map(\.name).sorted())).send("\n")
+        try fileSystem.writeFileContents(path, string: writer.buffer)
+    }
+
+    private mutating func render(targets: [BuildManifest.TargetName: Target]) {
+        self.buffer += "targets:\n"
+        for (_, target) in targets.sorted(by: { $0.key < $1.key }) {
+            self.buffer += "  \(target.name.asJSON): \(target.nodes.map(\.name).sorted().asJSON)\n"
+        }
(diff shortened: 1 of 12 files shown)


Sources/LLBuildManifest/ManifestWriter.swift
```diff
@@ -11,63 +11,108 @@
...
+    private mutating func render(targets: [BuildManifest.TargetName: Target]) {
+        self.buffer += "targets:\n"
+        for (_, target) in targets.sorted(by: { $0.key < $1.key }) {
+            self.buffer += "  \(target.name.asJSON): \(target.nodes.map(\.name).sorted().asJSON)\n"
         }
+    }
+
+    private mutating func render(nodes: [Node]) {
+        // We need to explicitly configure certain kinds of nodes.
+        let directoryStructureNodes = Set(nodes.filter { $0.kind == .directoryStructure })
```
owenv, line 59: It's probably fine to separate all these for now, but it is valid for a node to be, for example, both a directory structure and mutating node

owenv (approved):

Sources/LLBuildManifest/ManifestWriter.swift
```diff
@@ -179,3 +216,74 @@ extension Node {
         }
     }
 }
+
+extension String {
+    fileprivate var asJSON: String {
+        "\"\(self.jsonEscaped)\""
+    }
+
+    private var jsonEscaped: String {
```
MaxDesiatov, line 225: Directly lifted from TSC, suitable just for this use case. When we fully deprecate the `JSON` type, we can deprecate this in TSC too.

Sources/LLBuildManifest/Node.swift
```diff
@@ -37,15 +45,27 @@ public struct Node: Hashable, Codable {
...
+            kind: .virtual,
+            attributes: isCommandTimestamp ? .init(isCommandTimestamp: isCommandTimestamp) : nil
+        )
     }
 
     public static func file(_ name: AbsolutePath) -> Node {
         Node(name: name.pathString, kind: .file)
     }
 
+    public static func file(_ name: AbsolutePath, isMutated: Bool) -> Node {
```
MaxDesiatov, line 61: Couldn't create a default parameter for this one without breaking `.map(Node.file)` constructs that we have elsewhere in our code.

Sources/LLBuildManifest/Tools.swift
```diff
@@ -27,13 +27,13 @@ public protocol ToolProtocol: Codable {
     var outputs: [Node] { get }
 
     /// Write a description of the tool to the given output `stream`.
-    func write(to stream: ManifestToolStream)
+    func write(to stream: inout ManifestToolStream)
```
MaxDesiatov, line 30: Required after converting `ManifestToolStream` to a `struct` with a `String` buffer instead of using TSC streams.

Sources/LLBuildManifest/ManifestWriter.swift
```diff
@@ -11,63 +11,102 @@
...
-        let stream = BufferedOutputByteStream()
-        stream.send(
-            """
-            client:
-              name: basic
-            tools: {}
-            targets:\n
-            """
-        )
+    public static func write(_ manifest: BuildManifest, at path: AbsolutePath, _ fileSystem: FileSystem) throws {
```
tomerd, line 38: nit: use labeled parameter for `fileSystem`

MaxDesiatov:
@swift-ci test windows

Sources/LLBuildManifest/ManifestWriter.swift
tomerd: is there nothing in Foundation that we can use instead of inventing our own? further why not use Codable?

Tests/LLBuildManifestTests/LLBuildManifestTests.swift
```diff
@@ -62,7 +62,7 @@ final class LLBuildManifestTests: XCTestCase {
         manifest.addNode(.virtual("Foo"), toTarget: "main")
 
         let fs = InMemoryFileSystem()
-        try ManifestWriter(fileSystem: fs).write(manifest, at: "/manifest.yaml")
+        try ManifestWriter.write(manifest, at: "/manifest.yaml", fs)
```
tomerd, line 65: while we are here, should we call this something different than "Manifest"? the current name is pretty confusing with the package manifest

tomerd (approved):
she comments / ideas

Sources/LLBuildManifest/ManifestWriter.swift
MaxDesiatov: Nothing that they publicly expose. The new Swift Foundation has an internal `serializedForJSON` method on `String`, I'm not sure it's worth vendoring that instead of the TSC version? Additionally, we'd have to write our own encoder, since this is strictly neither JSON, nor YAML, it only uses JSON-escaping for strings. If we use `JSONEncoder` for some bits, without refactoring this much more we'd have to instantiate `JSONEncoder` on every call property call, which would be quite expensive and probably a significant slowdown for large projects.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test

Sources/LLBuildManifest/ManifestWriter.swift
MaxDesiatov: Hm, my bad, my impression was that it was somewhat custom, but I see it does mention YAML. Would you be opposed just refactoring this to render the build manifest with `Codable` and `JSONEncoder` in a future PR?

MaxDesiatov: Since we don't need to parse these manifest, I think rendering to JSON would mean we don't have to pull in additional dependencies for YAML support.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

PR #7034, opened 2023-10-25
use resolved location when managed dependencies and resolved graph do not agree

tomerd:
motivation: avoid removing entries from resolved file when managed dependencies has an older location (URL) of a dependency

changes:
* when managed dependencies have an older / different URL from the one resolved, use the resolved one for the resolved file
* add test

rdar://117381924


tomerd:
@swift-ci test

tomerd:
replaced by https://github.com/apple/swift-package-manager/pull/7039

PR #7053, opened 2023-11-06
Add `hostTriple` and `targetTriples` to `PluginContext`

MaxDesiatov:
### Motivation:

Per a request from @phausler:
> SwiftPM should support target architecture information from `BuildToolPlugin` or `PluginContext`.

### Modifications:

Added new `Triple` type to `PackagePlugin`, which for now is a trivial `struct` with a single `string` property. Added `hostTriple` and `targetTriples` properties on `PluginContext` that plugins now can use to apply different logic based on the combination of triples.

### Result:

Resolves rdar://111050554.

### DIFF
--- a/Sources/PackagePlugin/Triple.swift
+++ b/Sources/PackagePlugin/Triple.swift
@@ -10,6 +10,9 @@
 //
 //===----------------------------------------------------------------------===//
 
-public struct Triple {
-    public let string: String
+/// A type of machine on which code is build and/or executed. Usually includes information such as CPU architecture,
+/// vendor name, operating system, ABI, and object file format among other possible features. Swift and SwiftPM use
+/// Clang conventions for triple components and their naming.
+public struct Triple: CustomStringConvertible {
+    public let description: String
 }
--- a/Sources/PackagePlugin/Plugin.swift
+++ b/Sources/PackagePlugin/Plugin.swift
@@ -151,8 +151,8 @@ extension Plugin {
                 context = PluginContext(
                     package: package,
                     pluginWorkDirectory: pluginWorkDirectory,
-                    hostTriple: Triple(string: wireInput.hostTriple),
-                    targetTriples: wireInput.targetTriples.map(Triple.init(string:)),
+                    hostTriple: Triple(description: wireInput.hostTriple),
+                    targetTriples: wireInput.targetTriples.map(Triple.init(description:)),
                     accessibleTools: accessibleTools,
                     toolSearchDirectories: toolSearchDirectories)
                 target = try deserializer.target(for: targetId)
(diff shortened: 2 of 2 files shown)


phausler:
This looks great, it should be just barely enough to be able to determine (out of the supported arch's if the GYB files should be 32 bit or 64 bit emissions). This will let a stdlib build happen outside of the regular toolchain build process directly from a swift package.

phausler (approved):
lgtm

Sources/PackagePlugin/Triple.swift
```diff
@@ -0,0 +1,15 @@
...
+// Copyright (c) 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+public struct Triple {
+    public let string: String
```
rauhul, line 14: maybe: `description: String` instead? At some point this might want to be strongly typed

MaxDesiatov:
Converting to draft, as we need to clarify if necessary changes can be implemented in all users of libSwiftPM.

Sources/PackagePlugin/Triple.swift
MaxDesiatov: I think we'd have to replicate the whole `Triple` type we already have in SwiftPM implementation itself, and I'd like to avoid it if possible. If we want to land this in the near term, we'll have to keep it minimal, unblocking plugins to parse the triple as needed on their side.

MaxDesiatov: Evolving the `Triple` type, providing a public API for its parsing and other helpers like bit-width should come in future iterations, as Swift Evolution pitches/proposals I'd assume.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
Converted `targetTriple` to a `targetTriples: [Triple]` array to account for the fact that it can be empty (in command plugins), or contain multiple triples if the build graph happens to be shaped that way (e.g. with the Xcode build system)

Sources/PackagePlugin/PluginContext.swift
```diff
@@ -33,6 +33,16 @@ public struct PluginContext {
     /// it or the command will need.
     public let pluginWorkDirectory: Path
 
+    /// Triple of the machine that this plugin is running on.
+    public let hostTriple: Triple
+
+    /// Array of triples of machines that compiled code from this package will be running on. The plugin should be
+    /// prepared to handle case when either no target triples passed, or multiple triples passed at the same time. This
```
MaxDesiatov, line 40:
```suggestion
    /// Array of triples that compiled code from this package will be running on. The plugin should be
    /// prepared to handle cases when either no target triples passed, or multiple triples passed at the same time. This
```

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

tomerd:
looks good to me, modulo figuring out clients ability to provide the data

PR #7055, opened 2023-11-06
refactor dependency mapping

tomerd:
motivation: address broken validation when using relative path in remote depednencies

changes:
* simply to have a single API that encapsulated all the mapping use cases
* introduce intermediate MappablePackageDependency model to converge the various implementations
* simplify the mapping logic to better address realtive paths
* add and adjust tests

rdar://100823575


tomerd:
@swift-ci smoke test

tomerd:
@swift-ci test

Sources/PackageModel/DependencyMapper.swift
```diff
@@ -29,143 +28,220 @@ public struct DefaultDependencyMapper: DependencyMapper {
...
+                return try AbsolutePath(validating: location).pathString
+            // if the location has a scheme, assume a URL and nothing to normalize
+            case .some(_):
+                return .none
             }
         }
+    }
+}
+
+// trivial representation for mapping
```
MaxDesiatov, line 127: Can this be converted to a doc comment with more details describing what exactly is being mapped?

Sources/PackageModel/DependencyMapper.swift
```diff
@@ -29,143 +28,220 @@ public struct DefaultDependencyMapper: DependencyMapper {
...
+        case .registry:
+            return .none
+        }
+    }
+
+    fileprivate var sourceControlRequirement: PackageDependency.SourceControl.Requirement {
+        get throws {
+            switch self.kind {
+            case .fileSystem:
+                throw DependencyMappingError.invalidMapping("invalid mapping of file system to source control, requirement information mismatch")
```
MaxDesiatov, line 226: Can we make this error message actionable for the user? Maybe provide expected/actual values in the error message to clarify what exactly is not matching?

Sources/PackageModel/DependencyMapper.swift
```diff
@@ -29,143 +28,220 @@ public struct DefaultDependencyMapper: DependencyMapper {
...
-                // nothing to "fix"
-                return dependencyLocation
+        }
+    }
+
+    fileprivate var registryRequirement: PackageDependency.Registry.Requirement {
+        get throws {
+            switch self.kind {
+            case .fileSystem:
+                throw DependencyMappingError.invalidMapping("invalid mapping of file system to registry, requirement information mismatch")
```
MaxDesiatov, line 239: Ditto: can we provide more details on the mismatch in the error message to make it actionable?

Sources/PackageModel/DependencyMapper.swift
tomerd: what this means is that you cannot map file system dependency to source control one as you are missing the version requirement information. do you have a suggestion as to how better say that? maybe "version requirement information not available"?

Sources/PackageModel/DependencyMapper.swift
tomerd: this is an internal utility for applying mirrors, tbh I think this is pretty clear from the name and code what this does so I dont think adding docs would add value

MaxDesiatov: Having a quick look at the code I didn't get that it has a connection to mirrors. I hope this could mention that it's related to mirrors at least. While it's internal, there's still huge value in such documentation, as it makes it easier to review and easier to circle back to it later when needed.

Sources/PackageModel/DependencyMapper.swift
MaxDesiatov: I hope the error message could become more verbose. "Version requirement information not available" sounds better, but maybe it can provide more details where exactly it's unavailable, where it's expected to be, and how it can be provided?

tomerd: this is about mirroring depednecies. this error will be presented if someone tries to mirror file system dependency `dependency(path:)` to a source control one since there isn't version requirement on file system dependency ones. there isn't a way for the user to fix this since its invalid by definition, they need to remove it

MaxDesiatov: > perhaps something like:`mapping a file system dependency (<path>) to a source control one (<url>) is invalid`

This SGTM 👍

tomerd:
@swift-ci test

Sources/PackageModel/DependencyMapper.swift
tomerd: added comment to the main utility

tomerd:
@swift-ci test

MaxDesiatov (approved):
Thank you!

MaxDesiatov:
@swift-ci test windows

PR #7058, opened 2023-11-07
Warn if specified `--toolchain` path doesn't exist

MaxDesiatov:
The least we can do is issue a warning if a given path is non-existent.


MaxDesiatov:
@swift-ci test

neonichu (approved):

PR #7072, opened 2023-11-10
remove graph loading on package update

tomerd:
motivation: consistent behavior between pacakge update and package resolve

changes: do not attempt to load the dependencies graph (which performs additional validations), this means validation is deferred until build is attempted

rdar://113413552


tomerd:
@swift-ci test

neonichu (approved):

PR #7074, opened 2023-11-10
Allow tar.xz archives

swift-user326:
Allow tar.xz archives

### Motivation:

Better compression, [important](https://code.videolan.org/videolan/VLCKit/-/merge_requests/127#:~:text=I%27m%20not%20super%20happy%20that%20SPM%20does%20seem%20to%20support%20.zip%20archives%20only%20as%20this%20noticeably%20increases%20the%20download%20size%20compared%20to%20.xz.) for projects such as VLC

### Modifications:

All I had to do was add to list of allowed extensions, so system tar can handle it.

### Result:

Tar.xz are allowed

### DIFF
--- a/Sources/Basics/Archiver/TarArchiver.swift
+++ b/Sources/Basics/Archiver/TarArchiver.swift
@@ -17,7 +17,7 @@ import class TSCBasic.Process
 
 /// An `Archiver` that handles Tar archives using the command-line `tar` tool.
 public struct TarArchiver: Archiver {
-    public let supportedExtensions: Set<String> = ["tar", "tar.gz"]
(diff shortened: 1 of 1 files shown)


MaxDesiatov (commented):
How exactly does this handle systems that don't have XZ support available? At what stage and what exact error message is displayed then?

MaxDesiatov:
To elaborate, this change in isolation is fine, but I don't think that `.xz` support in `tar` is as widespread as `.gz`. At a minimum it might require installation of additional XZ-support dependencies by the user. We have to ensure that the error message in case of dependencies absence guides our users to a correct solution.

MaxDesiatov:
Hi @swift-user326, would you be interested in addressing the PR feedback and getting this PR over the finish line?

PR #7085, opened 2023-11-13
Enable strict concurrency checks for all targets

MaxDesiatov:
This requires bumping `// swift-tools-version` to 5.8.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
Marking as draft as this requires macOS CI to support Swift 5.8 and we need to fix actual build errors surfaced by these new checks.

Issue #7081, opened 2023-11-12
Internal error. Please file a bug at https://github.com/apple/swift-package-manager/issues with this info. symlinks not supported


compnerd:
### Description

If a package contains a symlink for the manifest, the package build seems to fail.  Example: https://github.com/compnerd/swift-firebase/actions/runs/6836319759/job/18591096135?pr=27#logs

### Expected behavior

Package should be possible to build.

### Actual behavior

Package manifest fails to build.

compnerd:
@MaxDesiatov is this really Windows specific?  It seems to be a generic code path, I don't see how this would be Windows.

PR #7092, opened 2023-11-14
SourceControl: add basic support for symlinks

compnerd:
The contents of a symlink is the path to which it points.  We can simply read the contents as a new path, and use `getEntry` to get the contents of the symbolic link.  This implementation should also fully traverse any path that is a series of symbolic link.  This does allow for a DoS attack by creating a recursive symbolic link cycle (e.g. `a` -> `b` -> `a`).

Fixes: #7081

### DIFF
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -917,8 +917,10 @@ public final class GitRepository: Repository, WorkingCheckout {
 
     /// Read a symbolic link.
     func readLink(hash: Hash) throws -> String {
-        return try callGit("cat-file", "-p", String(describing: hash.bytes),
-                           failureMessage: "Couldn't read '\(String(describing: hash.bytes))'")
+        return try callGit(
+            "cat-file", "-p", String(describing: hash.bytes),
+            failureMessage: "Couldn't read '\(String(describing: hash.bytes))'"
+        )
     }
 }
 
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Sources/SourceControl/GitRepository.swift
```diff
@@ -914,6 +914,12 @@ public final class GitRepository: Repository, WorkingCheckout {
             }
         }
     }
+
+    /// Read a symbolic link.
+    func readLink(hash: Hash) throws -> String {
+        return try callGit("cat-file", "-p", String(describing: hash.bytes),
+                           failureMessage: "Couldn't read '\(String(describing: hash.bytes))'")
```
MaxDesiatov, line 921: Formatting nit: this is what we have codified in the `.swiftformat` config.

```suggestion
        return try callGit(
            "cat-file", "-p", String(describing: hash.bytes),
            failureMessage: "Couldn't read '\(String(describing: hash.bytes))'"
        )
```

compnerd:
@swift-ci please test

MaxDesiatov (approved):
Thanks!

Sources/SourceControl/GitRepository.swift
```diff
@@ -1074,13 +1082,18 @@ private class GitFileSystemView: FileSystem {
...
         }
-        return try self.repository.readBlob(hash: hash)
+        switch entry.type {
+        case .symlink:
+            let path = try repository.readLink(hash: hash)
+            return try readFileContents(TSCAbsolutePath(validating: path))
+        case .blob:
+            return try self.repository.readBlob(hash: hash)
+        default:
+            fatalError()
```
tomerd, line 1095: https://github.com/apple/swift-package-manager/pull/7096

PR #7093, opened 2023-11-14
Ensure Xcode SDK paths are set when running test targets.

grynspan:
Currently, some of our code paths that run test targets don't correctly set the `DYLD_FRAMEWORK_PATH` and `DYLD_LIBRARY_PATH` arguments on macOS. This hasn't been a problem in practice because the XCTest helper tool is able to sort it out for us, but my plan with swift-testing is to run an executable directly (like we do for corelibs-xctest.) That means we'll have a code path that could potentially rely on XCTest or other Apple SDK bits, but which is not calling through the XCTest helper tool.

This change hoists our setting of those arguments to a position where they are always set (on macOS only!) before we run a test target, regardless of build style.

This change was previously landed as #7040 but was reverted (#7054) while we investigated a toolchain regression. We do not believe #7040 is the root cause of that issue, so this PR reapplies it.

### DIFF
--- a/Sources/Commands/Utilities/TestingSupport.swift
+++ b/Sources/Commands/Utilities/TestingSupport.swift
@@ -108,7 +108,7 @@ enum TestingSupport {
         #if os(macOS)
         let data: String = try withTemporaryFile { tempFile in
             args = [try Self.xctestHelperPath(swiftTool: swiftTool).pathString, path.pathString, tempFile.path.pathString]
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

neonichu (approved):

PR #7094, opened 2023-11-14
Make sure `swift test --show-code-cov-path` works when swift-testing is enabled.

grynspan:
This PR ensures that the aforementioned option behaves correctly when `--enable-experimental-swift-testing` is passed.
    
As it stands, `--show-code-cov-path` is implemented as a subcommand but behaves like an option; this PR removes the subcommand so that it "just works" without regard for other options.
    
A longer-term solution could be to make it into a proper subcommand (e.g. `swift test show-path code-coverage`.) That would ensure that we would get appropriate output from Swift Argument Parser when mixing it with other (nonsensical) options.


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

MaxDesiatov (approved):

PR #7096, opened 2023-11-14
small cleanup after 7092

tomerd:
motivation: do not use fatalError (which will crash the program) with no additional information

changes:
* throw InternalError instead of fatalError
* use correct version of AbsolutePath

### DIFF
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -1092,7 +1092,7 @@ private class GitFileSystemView: FileSystem {
         case .blob:
             return try self.repository.readBlob(hash: hash)
         default:
-            throw InternalError("unsupported git entry type \(entry.type)")
+            throw InternalError("unsupported git entry type \(entry.type) at path \(path)")
         }
     }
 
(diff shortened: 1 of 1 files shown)


tomerd:
@swift-ci test

neonichu (approved):

tomerd:
@swift-ci test

Sources/SourceControl/GitRepository.swift
```diff
@@ -1088,11 +1088,11 @@ private class GitFileSystemView: FileSystem {
         switch entry.type {
         case .symlink:
             let path = try repository.readLink(hash: hash)
-            return try readFileContents(TSCAbsolutePath(validating: path))
+            return try readFileContents(AbsolutePath(validating: path))
         case .blob:
             return try self.repository.readBlob(hash: hash)
         default:
-            fatalError()
+            throw InternalError("unsupported git entry type \(entry.type)")
```
MaxDesiatov, line 1095:
```suggestion
            throw InternalError("unsupported git entry type \(entry.type) at path \(path)")
```

MaxDesiatov (dismissed):
IMO the error string should contain the path so that it's obvious to the user what exactly to fix. I should've highlighted it in the original PR.

tomerd:
@swift-ci test

tomerd:
@swift-ci test windows

PR #7101, opened 2023-11-16
cache binary artifact globally

tomerd:
motivation: like other dependencies, binary artifacts are good candidates for user level caching such that they do not need to be re-downloaded

changes:
* update BinaryArtifactsManager to take cache path and use the cache to store binary artifacts when downloading them
* update test infra to enable/disable artifacts caching
* update workspaace call sites
* update workspace delegate to indicate when using cached binary artifact
* update and add tests

rdar://111774147

### DIFF
--- a/Sources/Commands/ToolWorkspaceDelegate.swift
+++ b/Sources/Commands/ToolWorkspaceDelegate.swift
@@ -135,12 +135,16 @@ class ToolWorkspaceDelegate: WorkspaceDelegate {
         self.outputHandler("Computed \(location) at \(version) (\(duration.descriptionInSeconds))", false)
     }
 
-    func willDownloadBinaryArtifact(from url: String) {
-        self.outputHandler("Downloading binary artifact \(url)", false)
+    func willDownloadBinaryArtifact(from url: String, fromCache: Bool) {
+        if fromCache {
+            self.outputHandler("Fetching binary artifact \(url) from cache", false)
+        } else {
+            self.outputHandler("Downloading binary artifact \(url)", false)
+        }
     }
 
-    func didDownloadBinaryArtifact(from url: String, result: Result<AbsolutePath, Error>, duration: DispatchTimeInterval) {
-        guard case .success = result, !self.observabilityScope.errorsReported else {
+    func didDownloadBinaryArtifact(from url: String, result: Result<(path: AbsolutePath, fromCache: Bool), Error>, duration: DispatchTimeInterval) {
+        guard case .success(let fetchDetails) = result, !self.observabilityScope.errorsReported else {
             return
         }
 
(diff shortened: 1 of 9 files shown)


tomerd:
@swift-ci test

Sources/Commands/ToolWorkspaceDelegate.swift
```diff
@@ -155,7 +159,11 @@ class ToolWorkspaceDelegate: WorkspaceDelegate {
             }
         }
 
-        self.outputHandler("Downloaded \(url) (\(duration.descriptionInSeconds))", false)
+        if fetchDetails.fromCache {
+            self.outputHandler("Fetched \(url) (\(duration.descriptionInSeconds))", false)
```
MaxDesiatov, line 163: IMO without mentioning the cache, "fetched" on its own isn't clear enough that it actually did fetch from cache

```suggestion
            self.outputHandler("Fetched \(url) from cache (\(duration.descriptionInSeconds))", false)
```

Sources/Commands/ToolWorkspaceDelegate.swift
```diff
@@ -135,12 +135,16 @@ class ToolWorkspaceDelegate: WorkspaceDelegate {
...
+        if fromCache {
+            self.outputHandler("Fetching binary artifact \(url) from cache", false)
+        } else {
+            self.outputHandler("Downloading binary artifact \(url)", false)
+        }
     }
 
-    func didDownloadBinaryArtifact(from url: String, result: Result<AbsolutePath, Error>, duration: DispatchTimeInterval) {
-        guard case .success = result, !self.observabilityScope.errorsReported else {
+    func didDownloadBinaryArtifact(from url: String, result: Result<(path: AbsolutePath, fromCache: Bool), Error>, duration: DispatchTimeInterval) {
```
MaxDesiatov, line 146: I'm unsure about the use of `Result` here in the first place. Why would `didDownloadBinaryArtifact` called in the `failure` case of `Result`? Wouldn't it be better to have a separate `didFailBinaryArtifactDownload` method for that? And then the use of tuple here exacerbates the problem. If it were two separate delegate methods you wouldn't need a tuple.

Sources/Workspace/Workspace+BinaryArtifacts.swift
```diff
@@ -563,23 +559,124 @@ extension Workspace {
...
-        result: Result<AbsolutePath, Error>,
+        result: Result<(path: AbsolutePath, fromCache: Bool), Error>,
         duration: DispatchTimeInterval
     )
     /// The workspace is downloading a binary artifact.
     func downloadingBinaryArtifact(from url: String, bytesDownloaded: Int64, totalBytesToDownload: Int64?)
     /// The workspace finished downloading all binary artifacts.
     func didDownloadAllBinaryArtifacts()
+
+
```
MaxDesiatov, line 679: formatting nit

```suggestion
```

Sources/Commands/ToolWorkspaceDelegate.swift
tomerd: mostly I was trying to keep the set of changes more minimal, but I dont see huge value in a result here too. do you think we should expand the scope of the PR, or a separate one?

tomerd:
@swift-ci test

tomerd:
@swift-ci test windows

Sources/Commands/ToolWorkspaceDelegate.swift
MaxDesiatov: I've been meaning to starting converting delegates to `AsyncSequence` APIs as a part of our concurrency changes. Then probably not worth touching it here or in a separate PR if it's going to be fixed in an imminent overhaul

MaxDesiatov (approved):

PR #7109, opened 2023-11-20
PackagePlugin: silence a deprecation warning on Windows

compnerd:
Prefer to use the microsoft recommended function for errno string description conversion. This silences a warning and relies on "safer" alternatives (at least ones which will not use a global shared buffer).


compnerd:
@swift-ci please test

MaxDesiatov (approved):

PR #7111, opened 2023-11-20
Tests: clean up path representation for manifest checks

compnerd:
The path representation on Windows is not Unix-like.  Use `AbsolutePath` to ensure that we correctly represent the path in expectations.  This fixes the test on Windows.


compnerd:
@swift-ci please test

PR #7119, opened 2023-11-24
build: simplify the build rules with newer CMake

compnerd:
Swift now requires CMake 3.19.6 to build.  Update the CMake requirement to 3.19 to allow us to simplify the build logic by removing the workarounds for older releases.


compnerd:
@swift-ci please test

MaxDesiatov (approved):

PR #7127, opened 2023-11-27
[NFC] Remove redundant `DependencyResolver` protocol

MaxDesiatov:
### Motivation:

This protocol doesn't define any requirements and was used for nemaspacing a single tuple typealias. Instead of typealiasing this fairly complex and pervasively used tuple, we should define a separate struct for it. Then the protocol itself can be removed as completely unused.

### Modifications:

Defined new `struct DependencyResolverBinding` with cleanly named fields, which now avoids expressions that look like `binding.binding` and re-declaration of this tuple in multiple places.

One other `typealias` in  `protocol DependencyResolver` is unused, thus the whole protocol is unused and now can be removed.

Also used this as an opportunity to clean up file naming and directory layout in the `PackageGraph` module.

### Result:

This confusing protocol is removed from the codebase. Readability is improved with explicit struct field names instead of ad-hoc tuple use. Directory layout makes it more obvious the purpose of specific files for potential new contributors and for existing contributors when reviewing code retrospectively.


MaxDesiatov:
@swift-ci test

tomerd (approved):

PR #7128, opened 2023-11-28
[NFC] Clean up package graph caching code

MaxDesiatov:
### Motivation:

Depends on https://github.com/apple/swift-package-manager/pull/7127.

`Resolved*` types in the `PackageGraph` module were confusingly named as `Resolved*Builder`, while they had no relation to [the established builder pattern](https://www.swiftbysundell.com/articles/using-the-builder-pattern-in-swift/) in Swift. In fact, these types are doing only caching, although proxying a few properties to initializer calls. The impact of this proxying remains to be benchmarked, as I have suspicion that these types specifically don't bring a significant speed up, but only introduce a redundant level of indirection.

### Modifications:

I've renamed these types to explicitly mention caching and moved them out to a separate directory.

### Result:

It's easier to understand that caching is being used. Relevant caching types are easier to find.


tomerd:
👍 this is a good change! also agree these types add complexity and are is not necessarily adding value.

one thought, should we consider calling them "CachedXxx" instead of "MemoizedXxx"?

tomerd (approved):

MaxDesiatov:
Closing as superseded https://github.com/apple/swift-package-manager/pull/7160. After refactoring this code I found that in a convoluted way these actually are "builder" types, but they rely on multiple side effects of other reference types in `PackageGraph` and `PackageModel` modules. We should clean that up and just remove these types in the future.

PR #7137, opened 2023-11-29
Rename the swift-testing entry point.

grynspan:
This PR renames the entry point function used by SwiftPM to call into swift-testing such that it does not need to be SPI. By marking it SPI, we unintentionally made it very difficult to use with a toolchain that does not include private module information.

swift-testing will continue to export the old symbol name as used by SwiftPM until this change has firmly landed in toolchains, after which we will remove the old function.

See: https://github.com/apple/swift-testing/pull/141


grynspan:
@swift-ci please test

neonichu (approved):

stmontgomery (approved):

tomerd (approved):

PR #7150, opened 2023-11-30
wip: add an API of plugins to retrieve authentication information

tomerd:
motivation: some plugins need authentication information, having an API to retrieve it from SwiftPM is helpful to abstract different authentication providers

changes: 
* create a new API for plugins to request authentication information from SwiftPM
* implement the IPC encoding/decoding for the API 

TODO:
[] tests
[] encrypt IPC so that auth info cannot be man-in-the-middle
[] add plugin privilege to signal a plugin would access auth information 

rdar://116226961


MaxDesiatov:
cc @euanh

PR #7154, opened 2023-12-01
[NFC] Remove unused `fileSystem` from `GraphLoadingNode`

MaxDesiatov:
### Motivation:

`GraphLoadingNode`s in the same graph always use same file system instance, there's no need to store references to it in each node.

### Modifications:

Removed `fileSystem` property on `GraphLoadingNode`, as it also simplifies `Equatable` and `Hashable` implementations on this type.

### Result:

`GraphLoadingNode` is now a true value type, with no need to define `Equatable` and `Hashable` manually on it.


MaxDesiatov:
@swift-ci test

tomerd (approved):

PR #7159, opened 2023-12-02
Fix ignored symbol graph extraction exit code

MaxDesiatov:
### Motivation:

Doesn't seem likely that symbol graph extraction process exit code was ignored on purpose, especially since `extractSymbolGraph` started returning process execution result recently.

### Modifications:

Explicitly checked whether process exit code is 0.

### Result:

Non-zero exit code from symbol graph extraction process launches are no longer ignored.


MaxDesiatov:
@swift-ci test

PR #7165, opened 2023-12-06
[cxx-interop] Make test discovery compatible with C++ interop

egorzhdan:
### Motivation:

If a Swift module is built with C++ interop enabled, all of its dependencies must also be built with C++ interop enabled.

Swift packages that use test discovery get a synthesized "PackageDiscoveredTests" target which is built by SwiftPM when someone runs `swift test` command. This module is currently always built without C++ interop, which is causing issues for projects that rely on it.
### Modifications:

This patch makes sure that the `-cxx-interoperability-mode=` command line flag gets propagated to the test discovery target.

### Result:

Packages that use both C++ interop and test discovery can now be built successfully by SwiftPM.

rdar://117078320 / resolves https://github.com/apple/swift-package-manager/issues/6990


Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
```diff
@@ -535,6 +535,22 @@ public final class SwiftTargetBuildDescription {
             args += ["-color-diagnostics"]
         }
 
+        // If this is a generated test discovery target, it might import a test
+        // target that is built with C++ interop enabled. In that case, the test
+        // discovery target must enable C++ interop as well
+        switch testTargetRole {
+        case .discovery:
+            for dependency in try self.target.recursiveTargetDependencies() {
```
egorzhdan, line 543: This is somewhat ugly, but it should go away once we have a better story for mixing C-interop and C++-interop in a single project.

neonichu (approved):

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
```diff
@@ -535,6 +535,22 @@ public final class SwiftTargetBuildDescription {
...
+                let dependencyScope = self.buildParameters.createScope(for: dependency)
+                let dependencySwiftFlags = dependencyScope.evaluate(.OTHER_SWIFT_FLAGS)
+                if let interopModeFlag = dependencySwiftFlags.first(where: { $0.hasPrefix("-cxx-interoperability-mode=") }) {
```
tomerd, line 546: are we doing these kind of checks elsewhere in the code? should it be encapsulated behind something like `dependencySwiftFlags.cxxInteropMode` returning an optional?

MaxDesiatov (commented):
Overall the change makes sense, but would be great to avoid the hackiness of adding this logic directly in `SwiftTargetBuildDescription`. Also, please cherry-pick it against `release/5.10` after it's merged.

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
MaxDesiatov: Generalizing `extraSwiftFlags.first(where:)` makes sense to me, the other fine seems fine as is.

egorzhdan: Hmm, I just realized that the other usage of `extraSwiftFlags.first` operates on `[String]`, not on flags/scopes. So I don't really see a way to simplify this.

egorzhdan:
@swift-ci please test

MaxDesiatov:
`BuildParameters` come from `SwiftTool` and are specified on the command line or via Swift SDKs and toolsets. I haven't worked with `BuildSettings` yet, but that seems like something manifest-specific, which is more appropriate here, isn't it? @neonichu can correct me if I'm wrong.

egorzhdan:
@swift-ci please test

PR #7178, opened 2023-12-11
[5.10] make it possible to specify permitted network port access at runtime

MaxDesiatov:
Cherry-pick of https://github.com/apple/swift-package-manager/pull/7151

motivation: The 'network' plugin permission in SwiftPM allows a plugin to request external network access to all or to a set of ports, but the command line flag does not allow to specify ports forcing users to allow all

changes:
* support passing ports to `local` and `all` --allow-network-connections flag
* write parsing logic to extract ports from `local:port,port,port` and `all:port,port,port`
* update remedy description to accurately specify the usage
* update tests

rdar://116241125

PR #7179, opened 2023-12-11
[5.10] make it possible to specify permitted network port access at runtime 

tomerd:
5.10 cherry-pick of #7151


tomerd:
@swift-ci test

tomerd:
ah looks like @MaxDesiatov already created #7178

PR #7182, opened 2023-12-11
Adopt `async`/`await` in more tests

AndrewHoos:
Adopt async/await in more tests

### Motivation:

Same as before. async/await good

### Modifications:

Moved multiple test targets to async/await

### Result:

Only three uses of `temp_await` remain in test targets after this PR.


MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

AndrewHoos:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

PR #7188, opened 2023-12-12
[cxx-interop] Do not pass the C++ standard if C++ interop is not enabled

egorzhdan:
This fixes the Swift PR testing failures.

### Motivation:

If C++ interop is not enabled, SwiftPM should not pass `-Xcc -std=c++17` flags.

### Result:

Only pass these flags if C++ interop is enabled.


egorzhdan:
@swift-ci please test

MaxDesiatov (approved):

PR #7204, opened 2023-12-16
Add option to allow HTTP (not HTTPS) registries

### LINKED ISSUE #7170: Support for non-HTTPS localhost registries
### Description

My work involves using an [SwiftPM Registry](https://github.com/apple/swift-package-manager/blob/main/Documentation/PackageRegistry/Registry.md) service hosted locally on my developer machine. This is an important requirement for our development workflow and internal policies.

Up until [SE-0392](https://github.com/apple/swift-evolution/blob/main/proposals/0391-package-registry-publish.md) was implemented, we had to write our own tooling for publishing SwiftPM packages to our registry. Now that the `swift` CLI has a command for publishing a package to a registry, we would like to be able to deprecate our homegrown tooling in favor of the built-in `swift package-registry publish` command.

However, we cannot start this migration because [`swift package-registry publish` validates](https://github.com/apple/swift-package-manager/blob/main/Sources/PackageRegistryTool/PackageRegistryTool%2BPublish.swift#L109) that [the given registry starts with HTTPS](https://github.com/apple/swift-package-manager/blob/main/Sources/PackageRegistryTool/PackageRegistryTool.swift#L165-L169), and our localhost SwiftPM registry can only support HTTP.

It is not tractable for us to add HTTPS support to our localhost registry, unless using self-signed certificates. I suspect (but was unable to confirm) that the HTTP client SwiftPM uses would reject self-signed certificates, and require TLS verification (If self-signed certificates would be supported, I'm happy to resolve this as a non-issue). Setting up a proper Certificate Authority and key distribution mechanism to be used on our developer machines to pass TLS verification would be extremely costly and does not significantly improve security posture.

While I applaud that SwiftPM has taken a "secure by default" approach for registries in this regard, we should also recognize the risks of not being able to reuse SwiftPM's own tooling for publishing for our use case. If we're not able to use the `swift package-registry publish` command, we have to continue having our own implementation which has already fallen behind with respect to things like support for [package signing](https://github.com/apple/swift-evolution/blob/main/proposals/0391-package-registry-publish.md#package-signing). Bespoke implementations may also have flaws that impact the security or correctness of publishing.

I would like to contribute a change to SwiftPM that would allow us to adopt the publishing command in the future, and I am looking for feedback regarding two potential paths to enable my use case:

1. Don't enforce the HTTPS requirement if the registry host URL component is `localhost`, `127.0.0.1`, or `::1`. This is the most simple, but has the downside of skirting [the principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) depending if you have the localhost registry use case or not.
1. Expose an option for users to allow HTTP registries. There's a few ways this could work such as environment variables or CLI flags, but I think a per-registry configuration in the existing `registries.json` would have the least risk of being used unintentionally.

Footpad:
Resolves #7170: this adds an option to the `package-registry` `set` and `publish` commands to allow using an HTTP registry endpoint, bypassing the existing requirement that package registries use HTTPS.

### Motivation:

In short, the motivation is to allow publishing to a registry that is hosted on a local machine. It is difficult to host a fully TLS-compliant HTTPS endpoint on developer machines at scale. See more detail in #7170

### Modifications:

* Adds a new `allowInsecureHTTP` flag to the Registry model.
* Adds a new `--allow-insecure-http` flag to the `package-registry` `set` and `publish` commands.
* Adds tests that validate the behavior of the model and command additions.
* Adds tests that validate the requested constraints in #7170 regarding HTTP endpoints, such as not being able to publish to an HTTP endpoint if there is authentication configured for it, and `login` command requiring HTTPS.

### Result:

We will be able to use the `swift` CLI to publish to our HTTP registries locally hosted on our development machines. As a result, we will be able to migrate to the official tooling instead of needing to continue using home-rolled publish tooling.


yim-lee:
@swift-ci please test

Sources/PackageModel/Registry.swift
```diff
@@ -16,10 +16,12 @@ import struct Foundation.URL
 public struct Registry: Hashable, CustomStringConvertible, Sendable {
     public var url: URL
     public var supportsAvailability: Bool
+    public var allowInsecureHTTP: Bool
```
yim-lee, line 19: I think the change to `validateRegistryURL` is sufficient so we don't need to change the registry model/configuration.

Tests/CommandsTests/PackageRegistryToolTests.swift
```diff
@@ -920,6 +1107,18 @@ final class PackageRegistryToolTests: CommandsTestCase {
...
+        try URL(string: "http://packages.example.com")!.validateRegistryURL(allowHTTP: true)
+
+        // Invalid
+        XCTAssertThrowsError(try URL(string: "http://packages.example.com")!.validateRegistryURL())
+        XCTAssertThrowsError(try URL(string: "http://packages.example.com")!.validateRegistryURL(allowHTTP: false))
+        XCTAssertThrowsError(try URL(string: "ssh://packages.example.com")!.validateRegistryURL())
+        XCTAssertThrowsError(try URL(string: "ftp://packages.example.com")!.validateRegistryURL(allowHTTP: true))
+    }
+
     private func testRoots() throws -> [[UInt8]] {
```
yim-lee, line 1122: Nice tests! ⭐

MaxDesiatov:
Hi @Footpad, thank you for your contribution! Would you be willing to address feedback on this PR and get it over the finish line?

Sources/PackageRegistryTool/PackageRegistryTool.swift
```diff
@@ -55,6 +55,9 @@ public struct SwiftPackageRegistryTool: AsyncParsableCommand {
         @Option(help: "Associate the registry with a given scope")
         var scope: String?
 
+        @Flag(name: .customLong("allow-insecure-http"), help: "Allows using a non-HTTPS registry URL")
```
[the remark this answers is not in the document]
Footpad: Will update

Sources/PackageRegistryTool/PackageRegistryTool+Publish.swift
```diff
@@ -82,6 +82,9 @@ extension SwiftPackageRegistryTool {
         )
         var certificateChainPaths: [AbsolutePath] = []
 
+        @Flag(name: .customLong("allow-insecure-http"), help: "Allows using a non-HTTPS registry URL")
```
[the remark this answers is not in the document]
Footpad: Will update

Sources/PackageModel/Registry.swift
Footpad: I'm realizing that the PR doesn't actually do this, but I was thinking that if an HTTP endpoint was already set by `swift package-registry set`, a following `swift package-registry publish` without a URL specified should not need the `--allow-insecure-http` flag.

The motivation for this would be to prevent the need to have something like the `--allow-insecure-http` option present in other Swift client tooling (e.g. Xcode) or other commands that might validate the URL in the future. I'd like to be able to add this option once to SPM and then not have to worry about tools like Xcode needing to also support it separately. 

Abstractly I agree it is not strictly necessary for there to be a change to the Registry model to add this flag: if it passed validation during `package-registry set`, then consumers of that model *may* take the stance that it doesn't need to be validated again. The only reason I could see to keep the configuration flag is if SPM wants to perform validation of URLs _after_ they have already been `package-registry set` - validation which may be desired in the fullness of time. I don't think I have a good sense of which direction we want to head.

I see a few possible paths, let me know what you'd like to see or if you have other ideas:

1.  We only revert the changes to the registry model.
    * UX: Developers who both pull and push to an HTTP registry are forced to specify `--allow-insecure-http` for both `package-registry set` for each HTTP registry and every time they `package-registry publish`. 
    * Impact on other tooling: Other SwiftPM tools like Xcode may need to have parallel options for being able to publish to HTTP registries. 
2. We revert the changes to the registry model _and_ don't require `--allow-insecure-http` if an HTTP URL has already been configured for the given package identifier or provided registry `--url`.
    * UX: Developers don't need to specify `--allow-insecure-http` to `publish` if they already did for that URL in `package-registry set`.
    * Impact on other tooling: Establishes precedence that the a configured URL need not be validated after it has been set with `package-registry set`.
4. We keep the changes to the registry model, and update `publish` to not require `--allow-insecure-http` if the Registry configuration's `allowInsecureHTTP` is true for that registry. In retrospect this was what I had originally intended, but did not actually check the registry configuration to have it short-circuit a check on `--allow-insecure-http`.
    * UX: Developers don't need to specify `--allow-insecure-http` to `publish` if they already did for that URL in `package-registry set`.
    * Impact on other tooling: Explicit modeling of the flag removes the ambiguity about how to handle HTTP registry URLs.

Footpad:
My apologies for the delay in checking back in on this PR - it has been quite busy since the holidays!

There's an outstanding feedback that I'd like to get more input on from the team in terms of which direction we should take the feedback. Please take a look and let me know how we should proceed. Thanks!

Sources/PackageModel/Registry.swift
yim-lee: Thanks for sharing your ideas and thought process.

`package-registry set` and `publish` work off of the `scopedRegistries` dictionary in `RegistryConfiguration`:

```swift
public var scopedRegistries: [PackageIdentity.Scope: Registry]
```

`set` updates the dictionary, while `publish` uses it to compute the registry URL given the package scope.

It _is_ possible for a registry to show up more than once in this dictionary, so a potential complication with having `Registry.allowInsecureHTTP` is that we have to make sure we update all instances of the same registry in the dictionary. 

Another scenario to consider is if a user edits the file manually and makes a mistake, such that different entries for the same registry are now in conflict.

It would also be easy for user to copy-paste an existing `registries.json` and miss updating `allowInsecureHTTP` to fit their usage, and unintentionally lead to a security hole.

Given these, I would pick option 1.

Footpad: Gotcha, given the additional complexity of the configuration data model think it makes sense then to keep this option isolated to the control plane of the tool. I do hope that Xcode won't have a similar assertion once it has publish tooling, but if it does then unfortunately we'll need to follow up with that as well. Have pushed a commit going with option 1.

Footpad:
Updated PR incorporating feedback

MaxDesiatov:
@swift-ci test

yim-lee (approved):
LGTM. Thanks for your work @Footpad!

MaxDesiatov:
@swift-ci test windows

Footpad:
Thanks folks for the help in getting this merged! Appreciate your patience as well 😄

PR #7211, opened 2023-12-20
Use `topologicalSort` with `Identifiable` on `ResolvedTarget`

MaxDesiatov:
### Motivation:

A serious performance regression was introduced for deep target dependency graphs in https://github.com/apple/swift-package-manager/pull/7160. After converting `ResolvedTarget` to a value type, its synthesized `Hashable` conformance traversed the whole dependency tree multiple times when using `topologicalSort`. This made package resolution seemingly hang for packages that had a lot of nested target graphs.

### Modifications:

We already have an implementation of `topologicalSort` on `Identifiable`, so we're using that now instead. I've also added a performance test for this in `PackageGraphPerfTests` to prevent future regressions in this area.

### Result:

Resolved rdar://119807385.

### DIFF
--- a/Sources/SPMTestSupport/ResolvedTarget+Mock.swift
+++ b/Sources/SPMTestSupport/ResolvedTarget+Mock.swift
@@ -1,0 +1,42 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2014-2017 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import PackageGraph
+import PackageModel
+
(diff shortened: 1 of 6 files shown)


MaxDesiatov:
@swift-ci test

Sources/SPMTestSupport/ResolvedTarget+Mock.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import PackageGraph
+import PackageModel
+
+extension ResolvedTarget {
```
MaxDesiatov, line 16: Moved from `ResolvedTargetTests` to reuse in perf tests.

neonichu (approved):

tomerd (approved):

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

PR #7215, opened 2023-12-24
[wip] refactor AbsolutePath based on URL and RelativePath based on [String]

tomerd:
motivation: better support for Windows style paths

changes:
* refactor AbsolutePath based on URL
* refactor RelativePath based on [String]
* remove redundant API surface area from RelativePath (anything not used outside tests)
* remove RelativePath tests that covered redundant APIs


compnerd:
From a quick look, we still have some API shape changes that need to be made.  The concept of a root is not portable.  The Windows model is a forest, not a tree.  Furthermore, there are technically an infinite set of roots, so we cannot really enumerate them, but can check if a given path is a root of a tree.

rauhul:
> From a quick look, we still have some API shape changes that need to be made. The concept of a root is not portable. The Windows model is a forest, not a tree. Furthermore, there are technically an infinite set of roots, so we cannot really enumerate them, but can check if a given path is a root of a tree.

Does system's FilePath type handle this?

MaxDesiatov:
> Does system's FilePath type handle this?

AFAIU it unfortunately does not, and the lack of Windows CI on swift-system prevents us from checking that regressions in that are don't happen.

tomerd:
> From a quick look, we still have some API shape changes that need to be made. 

@compnerd agreed, and your help is needed to define how to make it so that it works better for Windows style paths

PR #7216, opened 2023-12-25
Only output when tests fail if --quiet flag is present

clayellis:
Only output when tests fail if the `--quiet` flag is present.

### Motivation:

Resolves the latest conversation on #4395 concerning how the `--quiet` flag doesn't do what it advertises when used with `swift test` and `swift build`.

### Modifications:

- When not testing in parallel, and `--quiet` is present, collect output until tests finish and print output to stdout conditionally.
- Pass `LoggingOptions` into `ParallelTestRunner`.
- Determine if any of the parallel tests failed and, if so, output all test results (regardless of passing status) if `--quiet` is present.
- Adds three new test fixtures (`TestQuietPass`, `TestQuietPassFail`, `TestQuietFail`) for testing the quiet flag
- Adds tests for the `--quiet` flag when all tests pass, when some tests pass/fail, when all tests fail.

### Result:

If tests pass successfully and the `--quiet` flag is present, nothing will be output. Output as normal when tests fail. The only notable change to output in the failure case is that when running tests in parallel (`--parallel`), the progress animation is not updated.

### DIFF
--- a/Sources/Commands/SwiftTestTool.swift
+++ b/Sources/Commands/SwiftTestTool.swift
@@ -21,6 +21,7 @@ import SPMBuildCore
 import func TSCLibc.exit
 import Workspace
 
+import class TSCBasic.BufferedOutputByteStream
 import struct TSCBasic.ByteString
 import enum TSCBasic.JSON
 import class TSCBasic.Process
@@ -259,6 +260,7 @@ public struct SwiftTestTool: SwiftCommand {
                 numJobs: options.numberOfWorkers ?? ProcessInfo.processInfo.activeProcessorCount,
                 buildOptions: globalOptions.build,
                 buildParameters: buildParameters,
+                loggingOptions: swiftTool.options.logging,
                 shouldOutputSuccess: swiftTool.logLevel <= .info,
                 observabilityScope: swiftTool.observabilityScope
             )
@@ -392,13 +394,31 @@ public struct SwiftTestTool: SwiftCommand {
             library: library
         )
 
+        // If the quiet flag is present, only output on failure.
+        let runQuietly = swiftTool.options.logging.quiet
+
+        // No need to close this stream.
+        // Buffered output's close method is a no-op.
+        var outputStream = BufferedOutputByteStream()
+
         // Finally, run the tests.
         let ranSuccessfully = runner.test(outputHandler: {
-            // command's result output goes on stdout
-            // ie "swift test" should output to stdout
-            print($0, terminator: "")
+            if runQuietly {
+                print($0, terminator: "", to: &outputStream)
+            } else {
+                // command's result output goes on stdout
+                // ie "swift test" should output to stdout
+                print($0, terminator: "")
+            }
         })
+
         if !ranSuccessfully {
+            if runQuietly {
+                let bytes = outputStream.bytes
+                // command's result output goes on stdout
+                // ie "swift test" should output to stdout
+                print(bytes.validDescription ?? bytes.description, terminator: "")
+            }
             swiftTool.executionStatus = .failure
         }
 
@@ -852,6 +872,7 @@ final class ParallelTestRunner {
 
     private let buildOptions: BuildOptions
     private let buildParameters: BuildParameters
+    private let loggingOptions: LoggingOptions
 
     /// Number of tests to execute in parallel.
     private let numJobs: Int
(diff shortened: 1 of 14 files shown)


clayellis:
@swift-ci smoke test

clayellis:
This is far enough along now that I think it's ready for review. I'd like some feedback on the approaches taken, whether they're in the right direction.

I'd still like to add a few more tests for `swift build --quiet` and to resolve the discrepancies between a failed build's output with and without the quiet flag. (See [this comment](https://github.com/apple/swift-package-manager/issues/4395#issuecomment-1869271982).)

MaxDesiatov:
@swift-ci test

Sources/Commands/SwiftTestTool.swift
```diff
@@ -392,13 +394,31 @@ public struct SwiftTestTool: SwiftCommand {
             library: library
         )
 
+        // If the quiet flag is present, only output on failure.
+        let runQuietly = swiftTool.options.logging.quiet
+
+        // No need to close this stream.
+        // Buffered output's close method is a no-op.
+        var outputStream = BufferedOutputByteStream()
```
MaxDesiatov, line 402: I'd advise against using `BufferedOutputByteStream`. It's just a mutable string under the hood, with an added amount of confusing indirection, and (as the TSC library itself) will be deprecated in the future. Simplest alternative is plain `var stringAccumulator = ""` instead. You also won't need to convert to bytes and back to string again on L417-L421.

Sources/Commands/SwiftTestTool.swift
```diff
@@ -983,9 +1009,12 @@ final class ParallelTestRunner {
...
         progressAnimation.complete(success: processedTests.get().contains(where: { !$0.success }))
 
+        let testResults = processedTests.get()
+        let atLeastOneTestFailed = testResults.contains(where: { !$0.success })
+
         // Print test results.
-        for test in processedTests.get() {
-            if (!test.success || shouldOutputSuccess) && !buildParameters.testingParameters.experimentalTestOutput {
+        for test in testResults {
+            if (!test.success || shouldOutputSuccess || (atLeastOneTestFailed && loggingOptions.quiet)) && !buildParameters.testingParameters.experimentalTestOutput {
```
MaxDesiatov, line 1017: nit: line length, in newly written or updated code we're sticking to 120 characters per [the formatting config](https://github.com/apple/swift-package-manager/blob/ace3488c153f2d99732611109408d02186b75c20/.swiftformat#L15)

Sources/SPMTestSupport/XCTAssertHelpers.swift
```diff
@@ -131,7 +132,7 @@ public func XCTAssertBuildFails(
     line: UInt = #line
 ) -> CommandExecutionError? {
     var failure: CommandExecutionError? = nil
-    XCTAssertThrowsCommandExecutionError(try executeSwiftBuild(path, Xcc: Xcc, Xld: Xld, Xswiftc: Xswiftc), file: file, line: line) { error in
+    XCTAssertThrowsCommandExecutionError(try executeSwiftBuild(path, extraArgs: extraArgs, Xcc: Xcc, Xld: Xld, Xswiftc: Xswiftc), file: file, line: line) { error in
```
MaxDesiatov, line 135: nit: line length. If you're editing in Xcode, for simple function invocations like this one, most of the time Ctrl-M hotkey formats in the style we want.

MaxDesiatov (changes requested):
Overall this makes sense, but IMO it should avoid new uses of `BufferedOutputByteStream` and resolve a few line length formatting nits.

MaxDesiatov:
@swift-ci test

grynspan:
Does this change take swift-testing into account? If not, are there changes needed in swift-testing to support it?

Sources/Commands/SwiftTestTool.swift
clayellis: Good to know. Thanks for the insight @MaxDesiatov

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
The newly added tests need to be fixed on Linux for both `swift build` and `./Utilities/build-using-self`, and on macOS only for the latter.

MaxDesiatov (changes requested):
The newly added tests need to be fixed on Linux for both `swift build` and `./Utilities/build-using-self`, and on macOS only for the latter.

MaxDesiatov:
Hi @clayellis, would you be able to resolve conflicts and address review feedback on this PR? Thanks!

clayellis:
Yes, I'll aim to complete this PR by the end of the week (hopefully earlier.)

PR #7221, opened 2024-01-03
Fix CLI linker path arguments

krzyzanowskim:
Fixes `swiftCompilerFlags` value for linker path. In particular this fixes case when the linker path contains a space character


krzyzanowskim:
the actual problem is something else https://github.com/apple/swift-package-manager/issues/7222

PR #7231, opened 2024-01-05
Add `swift test --no-parallel`

grynspan:
This PR makes the `--parallel` flag on the swift-test subtool invertible with a "no" prefix.

We want swift-testing to default to enabling parallelization, but we don't want to break XCTest (which for historical reasons strongly prefers serial testing.) Deprecating `--parallel` and replacing it with `--{enable,disable}-parallelism` or similar seems a bit heavy-handed. Swift Argument Parser gives us the option to use a "no" prefix instead which means that the existing `--parallel` option will continue to work as intended.

Support for `--no-parallel` in swift-testing is here: https://github.com/apple/swift-testing/pull/174


grynspan:
@swift-ci please test

MaxDesiatov (approved):

PR #7235, opened 2024-01-08
Automatically enable swift-testing support when building swift-testing

grynspan:
Follow-up to #7220. swift-testing itself should get swift-testing support enabled by default, but it doesn't list itself as a dependency. This change extends the code in that PR to also check the identities of the packages being built (rather than just their dependencies.) If the package being built is named "swift-testing", then support is enabled.


grynspan:
@swift-ci please test

neonichu (approved):

PR #7239, opened 2024-01-10
Optional xcframework support for Linux

hassila:
Provide experimental support for XCFrameworks for Linux for well-defined, identical, deployments.

### Motivation:

Providing informal, experimental support for (very) heterogenous cluster deployments on Linux as discussed previously in https://github.com/apple/swift-package-manager/issues/5714

It allows for deployments which use XCFrameworks and library evolution on identically patches / toolchains on Linux as an interim escape hatch until wider support might be available. 

### Modifications:

Enable XCFrameworks for Linux if the `_SWIFTPM_EXPERIMENTAL_LINUX_XCFRAMEWORK` environment variable is set.

### Result:

Users may experiment with library evolution and XCFrameworks on Linux without having to roll their own toolchains while waiting for wider support for shared libraries / evolution on Linux in general.

The default behavior remains unchanged and there should be zero impact for users who doesn't set this environment variable.

### DIFF
--- a/Sources/SPMBuildCore/BinaryTarget+Extensions.swift
+++ b/Sources/SPMBuildCore/BinaryTarget+Extensions.swift
@@ -96,7 +96,7 @@ extension Triple.OS {
     /// Returns a representation of the receiver that can be compared with platform strings declared in an XCFramework.
     fileprivate var asXCFrameworkPlatformString: String? {
         switch self {
-        case .darwin, .linux, .wasi, .win32, .openbsd, .noneOS:
+        case .darwin, .wasi, .win32, .openbsd, .noneOS:
             return nil // XCFrameworks do not support any of these platforms today.
         case .macosx:
             return "macos"
@@ -106,6 +106,8 @@ extension Triple.OS {
             return "tvos"
         case .watchos:
             return "watchos"
+        case .linux:
+            return ProcessInfo.processInfo.environment["_SWIFTPM_EXPERIMENTAL_LINUX_XCFRAMEWORK"] == "1" ? "linux" : nil
         default:
             return nil // XCFrameworks do not support any of these platforms today.
         }


MaxDesiatov:
To manage expectations on this, I personally don't think XCFrameworks should be ever supported anywhere outside of builds on macOS with the Xcode build system. It's an Xcode-specific format that's not meant to be cross-platform outside of Xcode use cases isolated to Darwin. The proper format for true cross-platform binary libraries is `.artifactbundle`, which was designed in the open specifically for that purpose.

hassila:
Thanks @MaxDesiatov, we'll have a look at that, appreciate the pointers.

hassila:
@dschaefer2 ping from Swift in server

ktoso (commented):
Thanks for the chat @hassila @dschaefer2 ! Hope we can get this in :-)

Sources/SPMBuildCore/BinaryTarget+Extensions.swift
```diff
@@ -133,6 +133,8 @@ extension Triple.OS {
             return "tvos"
         case .watchos:
             return "watchos"
+        case .linux:
```
jakepetroules, line 134: This would need to return nil for Android triples. We should move the extension from `Triple.OS` to `Triple` and check both the os and environment fields; ensuring the environment field is gnu or musl, or at least not android/androideabi.

dschaefer2:
As we discussed in London, I think this is an important short term stop gap until we can the SwiftPM Vision doc together and cover all the use cases we see for binary artifacts. We learned a lot from prebuilts to support many platforms and compiler versions that I want to formalize and bring forward.

Sources/SPMBuildCore/BinaryTarget+Extensions.swift
```diff
@@ -133,6 +133,8 @@ extension Triple.OS {
...
+            return ProcessInfo.processInfo.environment["_SWIFTPM_EXPERIMENTAL_LINUX_XCFRAMEWORK"] == "1" ? "linux" : nil
```
dschaefer2: I prefer this being an experimental flag that we can make hidden if necessary. This is how we did the prebuilts.

dschaefer2: I'm just actively trying to make sure SwiftPM doesn't necessitate the need for Environment changes. It's just too many degrees of separation between where they're set and where they're used.

Sources/SPMBuildCore/BinaryTarget+Extensions.swift
vsarunas: Hi @jakepetroules ,

Moved and added an android check

Sources/SPMBuildCore/BinaryTarget+Extensions.swift
vsarunas: Hi @dschaefer2 ,

Do you mean to define a hidden boolean flag under [GlobalOptions.BuildOptions in Options.swift](https://github.com/swiftlang/swift-package-manager/blob/swift-DEVELOPMENT-SNAPSHOT-2025-10-02-a/Sources/CoreCommands/Options.swift#L456) and then to pass this to [BuildParameters](https://github.com/swiftlang/swift-package-manager/blob/swift-DEVELOPMENT-SNAPSHOT-2025-10-02-a/Sources/SPMBuildCore/BuildParameters/BuildParameters.swift) via  [SwiftCommandState](https://github.com/swiftlang/swift-package-manager/blob/swift-DEVELOPMENT-SNAPSHOT-2025-10-02-a/Sources/CoreCommands/SwiftCommandState.swift)  and making use of this in:

1. [BuildPlan+Swift.swift](https://github.com/swiftlang/swift-package-manager/blob/swift-DEVELOPMENT-SNAPSHOT-2025-10-02-a/Sources/Build/BuildPlan/BuildPlan%2BSwift.swift#L64)
2. [BuildPlan+Product.swift](https://github.com/swiftlang/swift-package-manager/blob/swift-DEVELOPMENT-SNAPSHOT-2025-10-02-a/Sources/Build/BuildPlan/BuildPlan%2BProduct.swift#L299)
3. [BuildPlan+Clang.swift](https://github.com/swiftlang/swift-package-manager/blob/swift-DEVELOPMENT-SNAPSHOT-2025-10-02-a/Sources/Build/BuildPlan/BuildPlan%2BClang.swift#L68)

Or is there more locations where changes are required?

dschaefer2: Yeah, the environment variable is pretty deep in the call tree here. That whole getter has a pretty heavy side effect relying on the name being nil to disallow a platform. Not a choice I would have made :). But let's not change too much.

We should really check at the BuildPlan level wherever we're calling parseXCFramework and only do linux if the flag is enabled.

vsarunas: Ok, any preference on what the flag should be named ?

```
/// Hidden option to allow XCFrameworks on Linux
@Flag(
    name: .customLong("experimental-xcframeworks-on-linux"),
    help: .hidden
)
public var enableXCFrameworksOnLinux: Bool = false
```

dschaefer2: That looks fine.

vsarunas: Hi @dschaefer2,

I have removed the env variable and added `swift-build --experimental-xcframeworks-on-linux`, which compiles our product successfully.

dschaefer2: Excellent. Thanks! I'm wondering if there's a test we can add to make sure it stays working as we make changes in there.

vsarunas: Do you mean something like [testXCFrameworkBinaryTargets()](https://github.com/swiftlang/swift-package-manager/blob/swift-DEVELOPMENT-SNAPSHOT-2025-10-02-a/Tests/BuildTests/BuildPlanTests.swift#L6681), to create a new test function for Linux ?

vsarunas: Hi @dschaefer2 ,

I have added a test on BuildPlan for the optional flag validation and also a functional test which compiles a library with Library Evolution; packages this into XCF; builds and runs a test binary on Linux.

hassila:
Thanks for the feedback and thanking @vsarunas for addressing it - @dschaefer2 anything more you'd need from us to move this further?

vsarunas:
@swift-ci please test

vsarunas:
@dschaefer2 - could you please trigger CI test again?

dschaefer2:
:).

vsarunas:
@dschaefer2 - Thanks! But unfortunately one more attempt for Windows is needed.

dschaefer2:
@swift-ci please test windows

vsarunas:
Tests have passed, but I have merged with main in the morning.

Does that mean the tests need to be run again?

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

vsarunas:
@dschaefer2 - is it possible to merge this PR?

dschaefer2:
Soon. :). I'm tied up with something else at the moment. Hopefully by the end of the week.

Tests/BuildTests/BuildPlanTests.swift
```diff
@@ -6807,6 +6807,114 @@ class BuildPlanTestCase: BuildSystemProviderTestCase {
...
+    func testXCFrameworkBinaryTargetsLinux(platform: String = "linux", arch: String, targetTriple: Basics.Triple) async throws {
+        let Pkg: AbsolutePath = "/Pkg"
+
+        let fs = InMemoryFileSystem(
+            emptyFiles:
+            Pkg.appending(components: "Sources", "exe", "main.swift").pathString,
+            Pkg.appending(components: "Sources", "Library", "Library.swift").pathString
+        )
+
+        try! fs.createDirectory("/Pkg/Framework.xcframework", recursive: true)
```
jakepetroules: This should be a regular try, not a force try.

Tests/BuildTests/BuildPlanTests.swift
```diff
@@ -6807,6 +6807,114 @@ class BuildPlanTestCase: BuildSystemProviderTestCase {
...
+        try! fs.writeFileContents(
```
jakepetroules: same here

Tests/FunctionalTests/LibraryEvolutionXCFLinuxTests.swift
```diff
@@ -0,0 +1,113 @@
...
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import _InternalTestSupport
+import Basics
+import Testing
+
+#if os(Linux)
```
jakepetroules: Cross compilation: this should not be behind #if os(Linux). The requireHostOS declaration is sufficient.

Tests/FunctionalTests/LibraryEvolutionXCFLinuxTests.swift
```diff
@@ -0,0 +1,113 @@
...
+                    buildSystem: .native
+                )
+
+                #if arch(arm64)
+                let arch = "arm64"
+                #elseif arch(x86_64)
+                let arch = "x86_64"
+                #endif
+
+                #if os(macOS)
```
jakepetroules: This test is requireHostOS linux so the if-else here isn't necessary

Tests/FunctionalTests/LibraryEvolutionXCFLinuxTests.swift
vsarunas: For macOS and Linux, it is sufficient ; but Windows build failing otherwise.

My browser history shows https://ci.swift.org/job/swift-package-manager-with-xcode-self-hosted-PR-osx-release-toolchain/2516 , but unfortunately the logs have been cleared.

Tests/FunctionalTests/LibraryEvolutionXCFLinuxTests.swift
vsarunas: I can remove, but I used it for development while being on macOS.

Tests/BuildTests/BuildPlanTests.swift
vsarunas: Removed.

Force try was from the existing test [testXCFrameworkBinaryTargets()](https://github.com/swiftlang/swift-package-manager/blob/swift-DEVELOPMENT-SNAPSHOT-2025-10-16-a/Tests/BuildTests/BuildPlanTests.swift#L6651-L6652)

Tests/BuildTests/BuildPlanTests.swift
vsarunas: Removed.

Tests/FunctionalTests/LibraryEvolutionXCFLinuxTests.swift
vsarunas: The failure was with Swift 6.1 on Windows - I added `requireSwift6_2` trait to try to avoid that failure.

@jakepetroules / @dschaefer2 - could you please trigger a CI check again?

Tests/FunctionalTests/LibraryEvolutionXCFLinuxTests.swift
vsarunas: Removed.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

hassila:
Hi @dschaefer2, friendly ping - let us know if there is anything else we can do to help make this land - getting close to branch time for 6.3, would be fantastic if we could get this in.

dschaefer2:
Don't worry about branching. There is a lot left for SwiftPM for 6.3 that will happen after the branch.

dschaefer2 (approved):
That said, approved :).

dschaefer2:
We'll also need to make sure this works with `--build-system swiftbuild` as well since we have promised parity (as much as possible) in 6.3.

dschaefer2:
Hey guys, we've had to revert this. It broke the toolchain builds for aarch64. (Apparently our CI only tests x86_64). Could you please make sure this works there as well. My first guess is you're using arm64 for the arch, but should be aarch64. Maybe?

vsarunas:
Hi @dschaefer2 , 

Indeed. Fixed in https://github.com/swiftlang/swift-package-manager/pull/9375

PR #7244, opened 2024-01-11
Bump swift-tools-version to 5.9

MaxDesiatov:
This change only bumps tools version to 5.9 without making any other changes, which allows us to isolate CI errors caused by this if such do happen.


MaxDesiatov:
@swift-ci test

neonichu (approved):

PR #7248, opened 2024-01-12
Fix `PackageGraph` traversal performance regressions

MaxDesiatov:
Depends on https://github.com/apple/swift-package-manager/pull/7258

This restores performance of traversal of deep package graphs previously available when `ResolvedTarget`, `ResolvedProduct`, and `ResolvedPackage` were reference types. Now instead of recursively applying `Hashable` and `Equatable` conformances to all properties in graph's node, we only hash and compare for equality their identities defined by `Identifiable`.

Uses of `Set` on these types are now replaced with the new `IdentifiableSet` type, which uses a dictionary of `[Element.ID: Element]` type as its storage.

The main benefit of this approach is that build triple of `ResolvedProduct` and `ResolvedTarget` is now a part of their identity. Thus we can still use value types, while the same `Target` from `PackageModel` instantiated as two different `ResolvedTarget`s with different `BuildTriples` will have different identities.

Resolves rdar://120861833
Resolves rdar://120711211


neonichu (approved):

MaxDesiatov:
@swift-ci test

MaxDesiatov:
The actual fix is also verified by `PackageGraphPerfTests`, which with the regression takes minutes to complete, but now with the fix passes in seconds. Not sure though why the regression doesn't trigger baseline checks on CI, maybe this test is somehow skipped? I previously did see it catching a more severe performance regression on CI, but only once. 🤔

MaxDesiatov:
Currently blocked due to https://github.com/apple/swift-package-manager/pull/7249

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
https://github.com/apple/sourcekit-lsp/pull/1025

@swift-ci test

MaxDesiatov:
https://github.com/apple/sourcekit-lsp/pull/1025

@swift-ci test windows

PR #7254, opened 2024-01-15
Tests: Exercise output and diagnostic messages from command plugins

euanh:
This test exercises command plugin console output under different verbosity settings:

### Motivation:

There doesn't seem to be a test which checks the handling of diagnostic messages from command plugins in a once place.   This test will help to avoid regressions in this area.

### Modifications:

* A new stub plugin prints to standard output or sends diagnostic output, depending on its arguments
* A new test uses the stub to generate outputs and checks what `swift package` prints to the console at different verbosity levels

### Result:

This test verifies existing functionality.    No change in Swift Package Manager.


MaxDesiatov (approved):
Thanks!

MaxDesiatov:
@swift-ci test

PR #7256, opened 2024-01-16
command plugins: Add a 'progress' diagnostic message at default verbosity

euanh:
This change adds a 'progress' message to the diagnostics which a plugin can send back to SwiftPM.   This message is printed to standard error at the default verbosity level.

 ### Motivation:

Currently, a plugin can write output to standard output or send diagnostics which SwiftPM writes to standard error.   If the plugin spawns long-running operations we might want to print progress messages to let the user know that it is still running.   The existing options all have compromises:

* Anything the plugin prints to its standard output or standard error is echoed to SwiftPM's standard output, but we may want to keep progress information separate from other plugin outputs - the plugin might be called as part of a shell pipeline where a downstream process will consume its output, for example.
* The existing diagnostic messages - remark (info), warning and error - are all suppressed at the default verbosity level.
* Increasing the level with `-v` causes SwiftPM to print lots of extra logging information from the build, which swamps the plugin info messages.

This commit adds a 'progress' message which SwiftPM will print to standard error at the default verbosity level, allowing progress to be shown without polluting standard output or pulling in additional build logging which might not be relevant to the user.

 ### Modifications:

* `Diagnostics.progress` takes a message as a string and passes it to SwiftPM, which prints it to standard error.

From the plugin programmer's point of view `Diagnostics.progress` is logically a diagnostics message, however it is a new type in the SwiftPM <-> plugin protocol because extending the existing diagnostics enum would require extensive changes to add a the new message and in many existing cases it would not be relevant.

 ### Result:

A plugin can show the user that it is making progress by printing messages which SwiftPM will echoed to stderr at the default verbosity level.

Example output (`diagnostics-stub` is the name of the plugin in `Package.swift`):
```
% swift-package print-diagnostics progress
[diagnostics-stub] command plugin: Diagnostics.progress
```

### DIFF
--- a/Sources/SPMBuildCore/Plugins/PluginInvocation.swift
+++ b/Sources/SPMBuildCore/Plugins/PluginInvocation.swift
@@ -240,7 +240,10 @@ extension PluginTarget {
                         diagnostic = .info(message, metadata: metadata)
                     }
                     self.invocationDelegate.pluginEmittedDiagnostic(diagnostic)
-                    
+
+                case .emitProgress(let message):
+                    self.invocationDelegate.pluginEmittedProgress(message)
+
                 case .defineBuildCommand(let config, let inputFiles, let outputFiles):
                     if config.version != 2 {
                         throw PluginEvaluationError.pluginUsesIncompatibleVersion(expected: 2, actual: config.version)
@@ -485,7 +488,9 @@ extension PackageGraph {
                         dispatchPrecondition(condition: .onQueue(delegateQueue))
                         outputData.append(contentsOf: data)
                     }
-                    
+
+                    func pluginEmittedProgress(_ message: String) {}
+
                     func pluginEmittedDiagnostic(_ diagnostic: Basics.Diagnostic) {
                         dispatchPrecondition(condition: .onQueue(delegateQueue))
                         diagnostics.append(diagnostic)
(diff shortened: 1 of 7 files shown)


Sources/SPMBuildCore/Plugins/PluginInvocation.swift
```diff
@@ -485,7 +488,9 @@ extension PackageGraph {
                         dispatchPrecondition(condition: .onQueue(delegateQueue))
                         outputData.append(contentsOf: data)
                     }
-                    
+
+                    func pluginEmittedProgress(_ message: String) {}
```
euanh, line 492: Not clear what is the best option here.   Could append to `outputData` or add a `progress` array field similar to `diagnostics` below.

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -767,6 +769,9 @@ class PluginTests: XCTestCase {
                     dispatchPrecondition(condition: .onQueue(delegateQueue))
                     diagnostics.append(diagnostic)
                 }
+
+                func pluginEmittedProgress(_ message: String) {}
+
```
MaxDesiatov, line 774: formatting nit
```suggestion
```

MaxDesiatov (approved):
Overall LGTM, but this needs @neonichu's approval before merging.

MaxDesiatov:
@swift-ci test

euanh:
Rebased on top of #7262

MaxDesiatov:
@swift-ci test

euanh:
@swift-ci test windows

PR #7260, opened 2024-01-16
Fix `./Utilities/soundness.sh` regressions

MaxDesiatov:
Since we don't run this script on CI right now, we have to fix regressions with manual runs of this script and separate PRs created for the fixes in the meantime.


MaxDesiatov:
@swift-ci test

neonichu (approved):

PR #7265, opened 2024-01-17
Replace uses of `package` with `@_spi(SwiftPMInternal)`

MaxDesiatov:
There are certain build configurations that currently fail due to the use of `package`. We can use `@_spi(SwiftPMInternal)` in the meantime.


MaxDesiatov:
@swift-ci test

bnbarham (approved):

PR #7266, opened 2024-01-17
Lower `swift-tools-version` to 5.8 to prevent use of `package`

MaxDesiatov:
Since use of `package` can introduce regressions in our CI, we should explicitly prevent it by lowering `swift-tools-version` to 5.8, which will make PRs using `package` fail. This is a temporary measure until CI issues are resolved.


MaxDesiatov:
@swift-ci test

bnbarham (approved):

neonichu (approved):

MaxDesiatov:
@swift-ci test

PR #7272, opened 2024-01-18
Add `SSH_AUTH_SOCK` to `EnvironmentVariables.nonCachableKeys`

MaxDesiatov:
Yet another environment variable that shouldn't trigger full package rebuilds.

Resolves rdar://107029374

### DIFF
--- a/Sources/Basics/EnvironmentVariables.swift
+++ b/Sources/Basics/EnvironmentVariables.swift
@@ -64,7 +64,7 @@ extension EnvironmentVariables {
 // rdar://107029374
 extension EnvironmentVariables {
     // internal for testing
-    internal static let nonCachableKeys: Set<String> = [
+    static let nonCachableKeys: Set<String> = [
         "TERM",
         "TERM_PROGRAM",
         "TERM_PROGRAM_VERSION",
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

Sources/Basics/EnvironmentVariables.swift
```diff
@@ -64,7 +64,7 @@ extension EnvironmentVariables {
 // rdar://107029374
 extension EnvironmentVariables {
     // internal for testing
-    internal static let nonCachableKeys: Set<String> = [
+    static let nonCachableKeys: Set<String> = [
```
bnbarham, line 67: Would it make sense to have an list that we know *should* cause rebuilds rather than the other way around?

bnbarham (approved):

Sources/Basics/EnvironmentVariables.swift
MaxDesiatov: Not until we support something like package traits to customise packages through predefined configuration values. Implementing that requires major changes to package graph resolution and manifest evaluation code.

Right now the only way for users to configure packages at build time is to pass an environment variable to `swift build` and read that variable in `Package.swift`. We can't know upfront which environment variables will be used for customisation by an arbitrary package, and changes in such variables have to cause rebuilds for this customisations to work in the first place.

PR #7273, opened 2024-01-18
Fix overlapping plugins/products build database access

MaxDesiatov:
This regression was introduced in https://github.com/apple/swift-package-manager/pull/7164/files#diff-7d5950f89d75bc14a591deb8270ebd32308c18248892c7abcb9d0ee3c6b14fd5L201.

We can restore the previous behavior in a hacky way by making sure that build parameters passed to plugin builds and invocations have the database path set in the same way as previously. This somewhat defeats the point of tools/products build parameters isolation, but we're able to isolate it only to a few lines in a single function.

I've added an explicit test case for this in new `PluginsBuildPlanTests.swift` file, so we can be sure it doesn't regress again.

Resolves rdar://120560817


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test linux

MaxDesiatov:
https://github.com/apple/swift/pull/70996

@swift-ci test linux

bnbarham (approved):

PR #7280, opened 2024-01-22
command plugins: Build command plugin dependencies for the host, not the target

euanh:
Always build command line plugin dependencies for the host triple.

### Motivation:

Since #7164, dependencies of command plugins are once again being built for the _target_ rather than the host.   This causes problem when cross compiling because the host needs to be able to run the plugin dependencies, but finds target binaries instead.

This problem was fixed before in #6791 by forcing command plugin dependencies to be built for the host by overriding the default build parameters in swiftTool.createBuildSystem().  The same solution still works in this commit, but a better long-term option would be to rework BuildOperation.plan() to handle command plugin dependencies specially, as it already does for build plugin dependencies.

### Modifications:

At present, BuildOperation.plan calls graph.invokeBuildToolPlugins to process sources.  invokeBuildToolPlugins finds all build tool dependecies and builds them separately, using a specially-created BuildOperation instance:

    https://github.com/apple/swift-package-manager/blob/34efc0bfe9d40d9a019644ac8fcd0b852c491dfe/Sources/SPMBuildCore/Plugins/PluginInvocation.swift#L409

There is no equivalent step for command plugin dependencies, so they are built for the host architecture.   Ideally we should rework BuildOperation.plan to build command and build plugin dependencies in the same way.   This commit forces all plugin dependencies to be built for the host - this is similar to what was done in #6791 and #7273.

### Result:

Command plugins can be used again when cross-compiling.


euanh:
@swift-ci test

euanh:
It's surprisingly tricky to write an automated test for this.  On macOS, a Swift installation can typically build for either `x86_64` or `arm64`, so the test can target the 'opposite' architecture and make sure that the plugin dependencies are build for the host.   On Linux, without cross-compilation SDKs installed, it's typically only possible to build for the host triple.  Trying to target an alternative triple will result in a build failure because SwiftPM still tries to build the dependency for host and target architectures, but doesn't have the standard library available.

MaxDesiatov:
@swift-ci test windows

MaxDesiatov (changes requested):
Would you be able to come up with a test that prevents future regressions in this area?

euanh:
> Would you be able to come up with a test that prevents possible future regressions in this area?

As I mentioned in the comment above (https://github.com/apple/swift-package-manager/pull/7280#issuecomment-1903824537), I think it's possible to have a macOS-only integration test, but I can't see a way to do the same for Linux.

MaxDesiatov:
Ah, sorry I missed that. Yes, macOS-only test is fine. There are at least a few tests that are macOS-only already.

neonichu (approved):

euanh:
@swift-ci test windows

MaxDesiatov (approved):
Thanks!

PR #7297, opened 2024-01-24
 Inherit host Swift SDK's toolset paths to fallback to host tools

kateinoigakukun:
Fallback to host SDK tools if target SDK doesn't have tools.

### Motivation:

Currently, Swift SDK fallbacks to system `PATH` search path for tools that are not found in the SDK. Usually, host compiler tool is capable to cross compile but just lacks some of artifacts like platform libraries.
In that case, we don't need to have a separate compiler tools in target-specific Swift SDK, so we can share the host compiler tools as long as the host Swift and target Swift are the same version.

### Modifications:

This patch makes Swift SDK fallback to host Swift SDK tools if the target Swift SDK doesn't have such tools. 

### Result:

This unlocks eliminating compiler tools from target-specific Swift SDKs, and will result in a smaller SDK size. Also makes it easier to build host platform independent Swift SDK.

### DIFF
--- a/Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
+++ b/Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
@@ -613,6 +613,7 @@ public struct SwiftSDK: Equatable {
       fileSystem: FileSystem
     ) throws -> SwiftSDK {
         var swiftSDK: SwiftSDK
+        var isBasedOnHostSDK: Bool = false
         // Create custom toolchain if present.
         if let customDestination = customCompileDestination {
             let swiftSDKs = try SwiftSDK.decode(
@@ -639,6 +640,7 @@ public struct SwiftSDK: Equatable {
         } else {
             // Otherwise use the host toolchain.
             swiftSDK = hostSwiftSDK
+            isBasedOnHostSDK = true
         }
         // Apply any manual overrides.
         if let triple = customCompileTriple {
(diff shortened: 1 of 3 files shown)


Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
```diff
@@ -598,6 +598,82 @@ public struct SwiftSDK: Equatable {
         return nil
     }
 
+    /// Computes the target Swift SDK for the given options.
+    public static func deriveTargetSwiftSDK(
```
MaxDesiatov, line 602: Let's avoid plain `public` to avoid making SwiftPM API too broad (and we can't use `package` yet due to unresolved CI issues)

```suggestion
    @_spi(SwiftPMInternal)
    public static func deriveTargetSwiftSDK(
```

kateinoigakukun: Make sense :+1: Updated patch

kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):
Thanks!

MaxDesiatov:
@swift-ci test

MaxDesiatov:
Linux smoke test failure seems unrelated

MaxDesiatov:
@swift-ci test linux

MaxDesiatov:
@swift-ci test macos

PR #7312, opened 2024-02-02
Fix non-tarballed SDK installation with remote URL

kateinoigakukun:
### Motivation:

4714ea92a introduced a regression where non-tarball SDKs could not be installed from a remote URL due to the wrong assumption that the downloaded file would always be a tarball. 

This issue exists in 5.10 release branch, so will cherry-pick after merging this PR.

### Modifications:

This commit fixes the issue by checking the file extension part of the URL for every supported archive format, then falling back to the tarball format if no supported extension is found.

### Result:

Non-tarball SDK artifact bundles can be installed from remote URL.


kateinoigakukun:
@swift-ci Please smoke test

kateinoigakukun:
@swift-ci Please test

MaxDesiatov (approved):
Great catch, thanks!

PR #7314, opened 2024-02-02
Move `swiftpm-project-settings.el` settings to `.dir-locals.el`

judemille:
Directory-local variables have been moved so that they are set directly
in `.dir-locals.el`.

### Motivation:

The variables set in `swiftpm-project-settings.el` can be set directly
in `dir-locals.el`. There is no reason to move them to a separate file,
and in fact doing that will automatically make Emacs flag the dir-locals
as unsafe, since it will be being requested to evaluate a block of code,
instead of setting known-safe variables.

### Modifications:

`Utilities/Emacs/swiftpm-project-settings.el` has been removed. All variables
set in it have been moved to `.dir-locals.el` itself. This allows the evaluation
of all directory locals to be considered safe.

Additionally, `swift-basic-offset` was changed to `swift-mode:basic-offset`.
That appears to be the proper variable for setting indent width for Swift in Emacs.

### Result:

Emacs users can edit the project more peacefully.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,19 +1,5 @@
-;;; Directory Local Variables
+;;; Directory Local Variables            -*- no-byte-compile: t -*-
 ;;; For more information see (info "(emacs) Directory Variables")
 
-
-((nil
-  (eval .
-        ;; Auto-load the swiftpm project settings.
-        (unless (featurep 'swiftpm-project-settings)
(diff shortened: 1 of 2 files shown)


MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

PR #7318, opened 2024-02-04
Fix typos in comments and symbols naming

rex4539:


Sources/Workspace/Diagnostics.swift
```diff
@@ -36,13 +36,13 @@ public enum WorkspaceDiagnostics {
     // MARK: - Errors
 
     /// The diagnostic triggered when an operation fails because its completion
-    /// would lose the uncommited changes in a repository.
-    public struct UncommitedChanges: Error, CustomStringConvertible {
+    /// would lose the uncommitted changes in a repository.
+    public struct UncommittedChanges: Error, CustomStringConvertible {
```
MaxDesiatov, line 40: This may be a breaking change for SwiftPM clients, since this type is `public`.

MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):
Verified that no known clients rely on renamed public symbols. Thanks!

PR #7326, opened 2024-02-07
Optimize for type checker in `embedInCode` with `[UInt8](arrayLiteral:)`

eonil:
Produces faster compiled code for embedded file contents.

### Motivation:
Old code produces typical array literal code that requires complicated compiler type inferences, and it slowed down compilation greatly.
For example, 

### Modifications:
Updated to use a certain syntax that is known to be faster.
Details are well explained in [this thread](https://forums.swift.org/t/why-is-swift-so-slow-timeout-in-compiling-this-code/61382/23).

For the generated code for a 300KB file on M1 machine,
- With old array syntax, took 20 seconds.
- With new array syntax, took 2-3 seconds.

### DIFF
--- a/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
@@ -355,7 +355,12 @@ public final class SwiftTargetBuildDescription {
             let variableName = $0.path.basename.spm_mangledToC99ExtendedIdentifier()
             let fileContent = try Data(contentsOf: URL(fileURLWithPath: $0.path.pathString)).map { String($0) }.joined(separator: ",")
 
-            content += "static let \(variableName): [UInt8] = [\(fileContent)]\n"
+            /// Typical array literal requires complicated type inference, therefore slow to parse.
+            /// This slows down compilation greatly if source file size is > 100KB.
+            /// - With typical array syntax, compilation of 300KB file on M1 machine took 20 seconds.
+            /// - With this certain syntax, compilation of 300KB file on M1 machine took 2-3 seconds.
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
Hi @eonil, thank you for the PR. Would you mind updating the description? It seems to be cut off in the "Motivation" section after the words "For example,".

Additionally, would you be able to add a fixture for this in the [`Fixtures/Resources`](https://github.com/apple/swift-package-manager/tree/main/Fixtures/Resources) directory that we could expose in benchmarks later to prevent regressions in this area?

MaxDesiatov:
@swift-ci test

grynspan:
Do you get the same performance speedup if you change the line to:

```swift
            content += "static let \(variableName) = [\(fileContent)] as [UInt8]\n"
```

? Or is that also pessimized for you? I ask because generally, you're not expected to directly invoke the `init(arrayLiteral:)` initializer.

eonil:
@MaxDesiatov 
Oh sorry for late reply.
I'll check and add a fixture.
Do you have any convention or manual to follow here? Can you provide a link if there's any?

@grynspan
- As far as I remember, your proposed notation didn't show significant performance improvement with Swift compiler shipped with Xcode 15.2. 
- Only literal initializer showed significant speedup. 
- I think this symptom can be fixed by improving compiler in a long term, and we can keep this as a short-term solution until we get a better compiler.
- I agree that this literal initializer is not supposed to be used from regular code. 
- I tested with a minimum Swift program with several different syntaxes. 
- I'll test again and attach the result.

eonil:
@MaxDesiatov Thanks.

BTW, I tested array-literal syntax on another machine and discovered generated code is getting compiled very slowly. I'll check this again and update PR.

eonil:
Closing as this doesn't look promising and I cannot proceed on this anymore.

PR #7329, opened 2024-02-10
Prevent build tasks from decreasing

rauhul:
Currently SwiftPM will reduce the total number of jobs in the CLI output if llbuild determines a task does not need to be run. This results in confusing output to the user where as the build progresses the number of completed tasks strictly increases but the total can fluctuate. This commit updates the task counting logic to consider "up to date" build operations as complete instead of reducing the total count leading to clearer progress output.

### DIFF
--- a/Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
+++ b/Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
@@ -1006,7 +1006,7 @@ private struct CommandTaskTracker {
         case .isScanning:
             self.totalCount += 1
         case .isUpToDate:
-            self.totalCount -= 1
+            self.finishedCount += 1
         case .isComplete:
             self.finishedCount += 1
         @unknown default:


rauhul:
@swift-ci test

MaxDesiatov (dismissed):
Makes sense, thanks!

MaxDesiatov (commented):
On a second look, can we come up with a test case for this?

MaxDesiatov:
@swift-ci test linux

compnerd:
Tree pruning is pretty common in build systems (e.g. make, ninja). I think that the more confusing (and frustrating) aspect is the increasing target counts.

rauhul:
> Tree pruning is pretty common in build systems (e.g. make, ninja). I think that the more confusing (and frustrating) aspect is the increasing target counts.

Thats fine, but the fact that the total count first increases _then_ decreases is pretty user experience. we could add a "skipped" count (which I actually plan on doing in a future PR) but that would not be "ninja" style output

rauhul:
> Tree pruning is pretty common in build systems (e.g. make, ninja). I think that the more confusing (and frustrating) aspect is the increasing target counts.

Thats fine, but the fact that the total count first increases _then_ decreases is pretty poor/confusing user experience. We could add a "skipped" count (which I actually plan on doing in a future PR) but that would not be "ninja" style output

rauhul:
@swift-ci test linux

MaxDesiatov:
@rauhul would you mind adding a test for this? One way could be to match stdout from e2e builds on fixture packages in `SwiftCommandStateTests`.

PR #7331, opened 2024-02-11
Implementation of same-package product dependencies

stackotter:
The implementation accompanying [an evolution proposal](https://github.com/apple/swift-evolution/blob/40afeb9eed259eb2b54dc46f18265946489a2409/proposals/NNNN-swiftpm-targets-same-package-product-dependencies.md) that would allow targets to depend on products within the same package.

### Motivation:

These changes allows dynamic linking between sibling targets, alleviating code-size and type-casting issues in certain situations. They also allow for executables to implement plugin systems without resorting to nested packages. See the proposal for motivating examples.

### Modifications:

- Created the `Target.Dependency.product(name:conditions:)` overload
- Added suitable representations for same-package products (a.k.a. `innerProduct`s) throughout SwiftPM's serialization, package model, and package builder code
- Updated `PackageBuilder.swift` and `PackageGraph+Loading.swift` to correctly construct and build package graphs containing same-package product dependencies

### Result:

Targets can now depend on products within the same package by using the `Target.Dependency.product(name: ..., conditions: ...)` static method when creating a target's dependencies.

### Tasks

- [x] Allow targets to depend on products within the same package
- [ ] Check for cycles
- [ ] Add tests
- [ ] Ensure that plugins can't depend on same-package products when it wouldn't make sense (will have to make that more rigorous)

### DIFF
--- a/Sources/Build/BuildPlan/BuildPlan+Product.swift
+++ b/Sources/Build/BuildPlan/BuildPlan+Product.swift
@@ -135,6 +135,9 @@ extension BuildPlan {
                 switch $0 {
                 case .product:
                     return nil
+                case .innerProduct:
+                    // TODO: Does this just mean that we can't @testable inner products? (seems fair enough?)
+                    return nil
                 case .target(let target, _):
                     return target
                 }
--- a/Sources/Commands/Utilities/MermaidPackageSerializer.swift
+++ b/Sources/Commands/Utilities/MermaidPackageSerializer.swift
@@ -115,6 +115,14 @@ extension MermaidPackageSerializer.Node {
                 border: .hexagon,
                 subgraph: product.package
             )
+        case let .innerProduct(product, _):
+            // TODO: Do we need a subgraph here?
+            self.init(
+                id: product.name,
+                title: product.name,
+                border: .hexagon,
+                subgraph: nil
+            )
         case let .target(target, _):
             self.init(target: target)
         }
--- a/Sources/PackageDescription/PackageDescriptionSerialization.swift
+++ b/Sources/PackageDescription/PackageDescriptionSerialization.swift
@@ -157,6 +157,7 @@ enum Serialization {
 
         case target(name: String, condition: Condition?)
         case product(name: String, package: String?, moduleAliases: [String: String]?, condition: Condition?)
+        case innerProduct(name: String, condition: Condition?)
         case byName(name: String, condition: Condition?)
     }
 
--- a/Sources/PackageDescription/PackageDescriptionSerializationConversion.swift
+++ b/Sources/PackageDescription/PackageDescriptionSerializationConversion.swift
@@ -204,6 +204,11 @@ extension Serialization.TargetDependency {
                 moduleAliases: moduleAliases,
                 condition: condition.map { .init($0) }
             )
+        case .innerProductItem(let name, let condition):
+            self = .innerProduct(
+                name: name,
+                condition: condition.map { .init($0) }
+            )
         case .byNameItem(let name, let condition):
             self = .byName(name: name, condition: condition.map { .init($0) })
         }
--- a/Sources/PackageDescription/Target.swift
+++ b/Sources/PackageDescription/Target.swift
@@ -57,6 +57,13 @@ public final class Target {
         ///    - moduleAlias: The module aliases for targets in the product.
         ///    - condition: A condition that limits the application of the target dependency. For example, only apply a dependency for a specific platform.
         case productItem(name: String, package: String?, moduleAliases: [String: String]?, condition: TargetDependencyCondition?)
+        /// A dependency on a product in the current package.
+        ///
+        /// - Parameters:
+        ///    - name: The name of the product.
+        ///    - moduleAlias: The module aliases for targets in the product.
+        ///    - condition: A condition that limits the application of the target dependency. For example, only apply a dependency for a specific platform.
+        case innerProductItem(name: String, condition: TargetDependencyCondition?)
         /// A by-name dependency on either a target or a product.
         ///
         /// - Parameters:
@@ -1261,6 +1268,20 @@ extension Target.Dependency {
 @available(_PackageDescription, obsoleted: 5.2, message: "the 'package' argument is mandatory as of tools version 5.2")
     public static func product(name: String, package: String? = nil) -> Target.Dependency {
         return .productItem(name: name, package: package, moduleAliases: nil, condition: nil)
+    }
+
+    /// Creates a dependency on a product from the same package.
+    ///
+    /// - Parameters:
+    ///   - name: The name of the product.
+    ///   - condition: A condition that limits the application of the target dependency. For example, only apply a
+    ///       dependency for a specific platform.
+    /// - Returns: A `Target.Dependency` instance.
+    public static func product(
+        name: String,
+        condition: TargetDependencyCondition? = nil
+    ) -> Target.Dependency {
+        return .innerProductItem(name: name, condition: condition)
     }
 
     /// Creates a dependency that resolves to either a target or a product with the specified name.
--- a/Sources/PackageGraph/PackageGraph+Loading.swift
+++ b/Sources/PackageGraph/PackageGraph+Loading.swift
@@ -393,6 +393,8 @@ private func createResolvedPackages(
                     return .target(targetBuilder, conditions: conditions)
                 case .product:
                     return nil
+                case .innerProduct:
+                    return nil
                 }
             }
             targetBuilder.defaultLocalization = packageBuilder.defaultLocalization
@@ -400,13 +402,26 @@ private func createResolvedPackages(
         }
 
         // Create product builders for each product in the package. A product can only contain a target present in the same package.
-        packageBuilder.products = try package.products.map{
-            try ResolvedProductBuilder(product: $0, packageBuilder: packageBuilder, targets: $0.targets.map {
+        packageBuilder.products = try package.products.map { product in
+            let productBuilder = try ResolvedProductBuilder(product: product, packageBuilder: packageBuilder, targets: product.targets.map {
                 guard let target = targetMap[$0] else {
                     throw InternalError("unknown target \($0)")
                 }
                 return target
             })
+            for targetBuilder in targetBuilders {
+                var conditions: [PackageCondition]?
+                for dependency in targetBuilder.target.dependencies {
+                    if case let .innerProduct(innerProduct, dependencyConditions) = dependency, innerProduct.name == product.name {
+                        conditions = dependencyConditions
+                    }
+                }
+                if let conditions = conditions {
+                    // TODO: Check for cycles
+                    targetBuilder.dependencies.append(.product(productBuilder, conditions: conditions))
+                }
+            }
+            return productBuilder
         }
 
         // add registry metadata if available
@@ -968,6 +983,7 @@ extension Target {
         }
     }
 }
+
 /// Builder for resolved package.
 private final class ResolvedPackageBuilder: ResolvedBuilder<ResolvedPackage> {
     /// The package reference.
--- a/Sources/PackageLoading/ManifestJSONParser.swift
+++ b/Sources/PackageLoading/ManifestJSONParser.swift
@@ -338,6 +338,8 @@ extension TargetDescription.Dependency {
                 package = try identityResolver.mappedIdentity(for: .plain(packageName)).description
             }
             self = .product(name: name, package: package, moduleAliases: moduleAliases, condition: condition.map { .init($0) })
+        case .innerProduct(let name, let condition):
+            self = .innerProduct(name: name, condition: condition.map { .init($0) })
         case .byName(let name, let condition):
             self = .byName(name: name, condition: condition.map { .init($0) })
         }
--- a/Sources/PackageLoading/ManifestLoader+Validation.swift
+++ b/Sources/PackageLoading/ManifestLoader+Validation.swift
@@ -196,6 +196,9 @@ public struct ManifestValidator {
                             validPackages: self.manifest.dependencies
                         ))
                     }
+                case .innerProduct:
+                    // TODO: Diagnose any possible issues here
+                    break
                 case .byName(let name, _):
                     // Don't diagnose root manifests so we can emit a better diagnostic during package loading.
                     if !self.manifest.packageKind.isRoot &&
--- a/Sources/PackageLoading/PackageBuilder.swift
+++ b/Sources/PackageLoading/PackageBuilder.swift
@@ -653,18 +653,23 @@ public final class PackageBuilder {
             // No reference of this target in manifest, i.e. it has no dependencies.
             guard let target = self.manifest.targetMap[$0.name] else { return [] }
             // Collect the successors from declared dependencies.
-            var successors: [PotentialModule] = target.dependencies.compactMap {
+            var successors: [PotentialModule] = target.dependencies.flatMap {
                 switch $0 {
                 case .target(let name, _):
                     // Since we already checked above that all referenced targets
                     // has to present, we always expect this target to be present in
                     // potentialModules dictionary.
-                    return potentialModuleMap[name]!
+                    return [potentialModuleMap[name]!]
                 case .product:
-                    return nil
+                    return []
+                case .innerProduct(let product, _):
+                    guard let product = self.manifest.products.first(where: { $0.name == product }) else {
+                        return []
+                    }
+                    return product.targets.compactMap { potentialModuleMap[$0] }
                 case .byName(let name, _):
                     // By name dependency may or may not be a target dependency.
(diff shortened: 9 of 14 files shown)


MaxDesiatov:
I don't think this is the right way to solve the problem. If the goal is to allow controlling how targets and products are linked, it should be done directly at the place of dependency declaration, and not in a roundabout way that allows depending on products in the same package.

Let's say I want to link a private target dynamically, is my only option then to expose it in my package as a product?

Another problem with this approach is that it constrains users of a package to specific linking kind without any ability to control it at the place of dependency declaration.

A slight terminology nit: documents like the one linked from PR description are usually not called evolution proposals until they successfully pass the pitch phase and are scheduled for review by the Swift team.

I advise gathering some feedback on Swift Forums and getting to a consensus on a pitch first, before proceeding with the implementation.

stackotter:
Hey Max thanks for the feedback,

I only created the implementation last night (even though it hasn't been pitched since the initial forum thread) so that I could test it out on my own projects and see if it solved my issues in practice.

I have thought about the alternative solution of allowing targets to control the linking mode when depending on another target, however that alone wouldn't solve the issues that this pitch is trying to address;

```swift
let package = Package(
    name: "Library",
    products: [
        .library(name: "LibA", type: .dynamic, targets: ["A"]),
        .library(name: "LibB", type: .dynamic, targets: ["B1", "B2]),
    ],
    targets: [
        .target(name: "A", dependencies: [
            .dynamic("B1") // Ignore the syntax, just for demonstration
        ]),
        .target(name: "B1"),
        .target(name: "B2"),
    ]
)
```

In the above example, `LibA` is a dynamic library `liblibA.dylib` linked to `libB1.dylib`, while `LibB` is a dynamic library `liblibB.dylib` containing both `B1` and `B2` (statically). The code of `B1` would still be duplicated between `libB1.dylib` and `liblibB.dylib`, causing the type-casting and code-size issues that this pitch is trying to address. It could potentially be made to work if you also allowed the linking mode used by products when linking against targets, however that starts getting a bit tedious. I'll be interested to see what people think about that option (I'll add it as an alternative considered).

The reason that allowing same-package product dependencies works to solve all of the cases that motivated this pitch (of which private target linking wasn't one, but more on that next), is that these issues only arise when your package is a dependency of another package, and someone is depending on more than one product from your package where one of them is a dynamic library and both of them contain some code in common. That's why in my eyes the solution will likely involve products in some way shape or form.

Controlling the linking of private targets is an important oversight, thanks for bringing that up. Since a private target wouldn't be directly included in any products, you would likely be able to avoid the limitation with target-target linking that the manifest above demonstrates, so it could be addressed by additionally introducing a way to control linking between targets. Now that I think about it, having both solutions (target-target linking control & target-product same-package dependencies) would provide even more linking flexibility and could provide the tools to solve almost any linking problems that one might face.

Apologies if all that would've been better suited to a pitch thread, I'll create one once I've updated the pitch with the points discussed above.

stackotter:
I realised that the link to the original pitch in the proposal draft is probably easily missed. Here's the original forum pitch thread that sparked this proposal draft: https://forums.swift.org/t/pitch-swiftpm-allow-targets-to-depend-on-products-in-the-same-package/57717

MaxDesiatov:
> ```swift
> let package = Package(
>     name: "Library",
>     products: [
>         .library(name: "LibA", type: .dynamic, targets: ["A"]),
>         .library(name: "LibB", type: .dynamic, targets: ["B1", "B2]),
>     ],
>     targets: [
>         .target(name: "A", dependencies: [
>             .dynamic("B1") // Ignore the syntax, just for demonstration
>         ]),
>         .target(name: "B1"),
>         .target(name: "B2"),
>     ]
> )
> ```
> 
> In the above example, `LibA` is a dynamic library `liblibA.dylib` linked to `libB1.dylib`, while `LibB` is a dynamic library `liblibB.dylib` containing both `B1` and `B2` (statically). The code of `B1` would still be duplicated between `libB1.dylib` and `liblibB.dylib`, causing the type-casting and code-size issues that this pitch is trying to address.

My main question here would be then: why do `LibA` and `LibB` require to be declared as `.dynamic` products in the first place? I don't see a clear use case for that in the presence of specifying linking kind at the place of dependency declaration. Products should just stick to the default `.auto` declaration instead:

```swift
let package = Package(
    name: "Library",
    products: [
        .library(name: "LibA", targets: ["A"]),
        .library(name: "LibB", targets: ["B1", "B2]),
    ],
    targets: [
        .target(name: "A", dependencies: [
            .dynamic("B1") // Ignore the syntax, just for demonstration
        ]),
        .target(name: "B1"),
        .target(name: "B2"),
    ]
)
```
As specified this way, we can infer the right way to link `B1` based on a computed products and targets graph in each case. `swift build --product LibB` would build a statically linked `B1` and `B2`, `swift build --product LibA` would dynamically link `B1`, and plain `swift build` without products specified would dynamically link `B1` into both `LibA` and `LibB`, unless we provide a way to override that as well.

But the main question is: why do you need to link dynamically in the first place? Dynamic linking is ok for system libraries that ship in an SDK that multiple apps can reuse at the same time. If you're building a single app with specific versions of your packages that can't be shared with other apps (and this is always the cases for apps on the iOS App Store), just link everything statically. That will not only avoid the duplication, but will also be beneficial for final binary size and app startup time. And that's the default for `.auto` anyway.

stackotter:
In my specific use case, the reason is that my app has a plugin system, and plugins written in Swift need to link against the same copy of the plugin API as the app itself (otherwise objects passed between the two just don't work correctly whatsoever, e.g. a plugin wouldn't be able to use an `any Event` passed to it by the app since type-casting would just fail for every concrete `Event` type), so dynamic linking between the app and its plugin API is required (static linking to the plugin API would give the plugin and the app their own separate copies of the plugin API). I wouldn't be using dynamic linking if I didn't have to, I generally dislike it a lot.

If it was only a thin plugin API that had to be in a separate repo/package, none of this would be an issue for me, but the dylib with the plugin API isn't just the plugin API, it necessarily includes basically the whole implementation of the app except for the UI (plugins can interact with a lot), and many changes that I make span across both the top-level package and the nested package (making splitting them across repos a non-option for maintainability).

Additionally, there are multiple other people in the original pitch thread who claim that they've wanted this for quite a while and that they want a way to fix their dynamic linking issues without resorting to multiple nested Package.swift's (which then requires `@_exported` if the common code is intended to be used by consumers) or splitting their project into separate repos (which often makes development/contributing tedious, and doesn't make sense). However, none of them were super specific about their use-cases so I can't tell you exactly why they're using dynamic linking in the first place.

MaxDesiatov:
I understand there is a need for dynamic linking in certain cases, but I'll reiterate that this should be solved by providing additional linking kind controls, which is orthogonal to the way that dependencies are specified within a package.

Expressing a dependency on products within the same package would have a similar effect only because of the limited way linking controls can be specified right now. We should lift those limitations on linking controls first of all. Not sure it's worth spending time instead on workarounds that bring their own limitations, like the aforementioned inability to keep dynamically linked targets private.

stackotter:
I agree that a more flexible solution is better. If people aren't opposed to a more overarching set of changes regarding linking behaviour then I'd be happy to explore that avenue instead. I think the original idea was to keep the proposal simple and self-contained (I wasn't the original pitcher so I can't say for sure).

If the same-package products idea gets scrapped we could instead (as you've mentioned) introduce controls for configuring linking modes of targets depending on targets (and possibly products depending on targets). I think you'd often end up with more dynamic libraries than strictly required (e.g. in the example above with `A`, `B1`, and `B2`, you'd end up with `B1` and `B2` as separate dylibs), but that may not be a big issue in practice.

### Aside

The following isn't necessary to solve the code size and type-casting issues (as long as you're happy with `B1` and `B2` being in separate dylibs), but to be able to express the linking setup that same-package product dependencies achieve in the `A`, `B1`, `B2` example above, you'd probably end up needing a way to depend on bundles of targets like so,

```swift
.target(
    name: "A",
    dependencies: [
        .dynamicBundle(["B1", "B2"]) // ignore naming for now
    ]
)
```

which ends up looking a lot like depending on an anonymous product. But I'll drop that idea completely for now.

## Exploring a new solution

Here's a small example setup where I've explored a new solution and annotated some potential pain points.

```swift
// The app's Package.swift
let package = Package(
    name: "App",
    products: [
        .executable(name: "App", targets: ["App"])
        // If `AppCore` and `AppRenderer` were to be part of the same product,
        // SwiftPM would somehow need to get told that `AppCore` and `AppRenderer`
        // must be separate dylibs when compiling the composite product for
        // dynamic linking. That's only required because `App` depends on both
        // targets as separate dylibs and the library is intended for plugins
        // to use (which need to link the same way as `App`), it isn't a rule in
        // general so it can't and shouldn't be automatically inferred. I haven't
        // yet thought of a nice API for specifying when building the product
        // as a dynamic library, that certain targets should be separate dylibs. We could
        // go with an API such as `[.dynamic("AppCore"), .dynamic("AppRenderer")]`,
        // but that implies that they're dynamically linked even under static linking,
        // which isn't desirable.
        //
        // Requiring that the issue is worked around by splitting the product into two
        // products like so could be considered an acceptable compromise for a
        // first set of changes (or forever if it doesn't bother enough people).
        .library(name: "AppCore", dependencies: ["AppCore"]),
        .library(name: "AppRenderer", dependencies: ["AppRenderer"]),
    ],
    targets: [
        // Both of App's dependencies must always be dynamically linked to,
        // so no need to design a flexible auto linking API for this part.
        .executableTarget(
            name: "App",
            dependencies: [
                .dynamic("AppCore"),
                .dynamic("AppRenderer")
            ]
        ),
        // This CLI doesn't need the renderer, so rendering is split out into
        // a separate target from `AppCore` (`AppRenderer`).
        .executableTarget(
            name: "CLI",
            dependencies: ["AppCore"]
        ),
        // In this example it's possible to infer that `AppCore` must be dynamically
        // linked to `AppRenderer` when building `App` (or building the `AppRenderer`
        // product as a dylib), but in more complicated cases I would likely have to
        // resort to using `.dynamic("AppCore")`, which would enforce dynamic linking
        // even when not necessary (not ideal).
        .target(
            name: "AppRenderer",
            dependences: ["AppCore"]
        ),
        .target(name: "AppCore")
    ]
)

// A plugin's Package.swift
let package = Package(
    name: "Plugin",
    products: [
        // In this case it makes sense to enforce dynamic linking since the plugin
        // is only ever built to be loaded by the app, and is as such always compiled
        // as a dylib. But perhaps some SwiftPM command line flags like so could be
        // useful in other use cases (the existing swiftc -emit-library flag seems
        // unreliable with SwiftPM from some initial Googling);
        // $ swift build --product Plugin --linking-mode dynamic
        .library(name: "Plugin", .dynamic, targets: ["Plugin"])
    ],
    dependencies: [
        .package(path: "../App")
    ],
    targets: [
        // Targets can specify their own linking modes for dependencies, as long as the
        // products they depend on don't explicitly specify their own linking mode.
        .target(
            name: "Plugin",
            dependencies: [
                .product(name: "AppCore", package: "App", .dynamic),
                .product(name: "AppRenderer", package: "App", .dynamic)
            ]
        )
    ]
)
```

furby-tm:
plus one on needing something here in SwiftPM, having all symbols duplicated in all libraries is a heavy setback in the current design.

furby-tm:
> But the main question is: why do you need to link dynamically in the first place? Dynamic linking is ok for system libraries that ship in an SDK that multiple apps can reuse at the same time. If you're building a single app with specific versions of your packages that can't be shared with other apps (and this is always the cases for apps on the iOS App Store), just link everything statically. That will not only avoid the duplication, but will also be beneficial for final binary size and app startup time. And that's the default for .auto anyway.

Building everything statically is **not** at all orthogonal to the way that many **C**/**ObjC**/**C++** libraries are built, it feels like development trends in **SwiftPM** are not properly evolving alongside Swift, one of Swift's biggest accomplishments (as of **5.9**) is [**Swift**/**C++** interoperability](https://www.google.com/search?client=safari&rls=en&q=swfit%2Fc%2B%2B+interop&ie=UTF-8&oe=UTF-8), and **SwiftPM** has yet to release any quality of life improvements to support this feature, it feels like we are going backwards here.

We will work to revise both the implementation and a plan, but we are experiencing a lot of friction, from simply being able to make full and proper use of **Swift's** existing feature-set.

furby-tm:
Interop efforts from @zoecarver have been instrumental to the evolution of Swift, I'm curious on her thoughts of these noticeable lacking features in SwiftPM, considering the impossibility to reliably replicate the functionality of existing C/C++ projects... purely because of the package manager limitations (see [**this comment**](https://github.com/swiftlang/swift-package-manager/issues/7413#issuecomment-2349503409)).

dschaefer2:
I wonder if we can just wind this back to the original proposal, allowing to add a dependency from a target on a local product. But then allow products to be declared private.

As I muck through the build graph, I am starting to appreciate the difference between products and targets. Targets end up being modules and bags of object files. Products package up those object files into a format for consumption of other packages. What it feels like we need is to allow packaging up of the objects for internal use as well, especially as DLLs which are a much different beast than the bag of objects.

This especially comes to a head on Windows where we are very limited on the number of symbols a PE object (exe or dll) can export. The Windows way to solve this is to break your exe's into multiple DLLs. We need packages to be friendlier to those environments.

stackotter:
> I wonder if we can just wind this back to the original proposal, allowing to add a dependency from a target on a local product. But then allow products to be declared private.
> 
> As I muck through the build graph, I am starting to appreciate the difference between products and targets. Targets end up being modules and bags of object files. Products package up those object files into a format for consumption of other packages. What it feels like we need is to allow packaging up of the objects for internal use as well, especially as DLLs which are a much different beast than the bag of objects.
> 
> This especially comes to a head on Windows where we are very limited on the number of symbols a PE object (exe or dll) can export. The Windows way to solve this is to break your exe's into multiple DLLs. We need packages to be friendlier to those environments.

Thanks for weighing in Doug! After running up against these limitations a number of times, I fully agree; there is a clear distinction between the roles of targets and products, and linking definitely seems much more related to products than targets.

---

If we were to go back to the original proposal, the main improvement I'd suggest would be to account for a `.sameAsParentProduct` linking mode or something along those lines (`.inherit` could be a better name?). I often end up only needing dynamic linking under certain circumstances, but when I do need dynamic linking I need it to propagate through multiple layers of products (which there is no existing workaround for other than building up a web of nested packages with separate `dynamic` and `static` versions of each product). This is demonstrated in the discussion of `DeltaClient` in the next few paragraphs, particularly when I get to `DeltaRenderer`.

[Delta Client](https://github.com/stackotter/delta-client) has an overarching executable product `DeltaClient` and a library product `DeltaCore`. To support plugins, `DeltaClient` must dynamically link to `DeltaCore`. Otherwise plugins all contain their own separate copies of `DeltaCore` (in addition to `DeltaClient`'s copy), which leads to runtime issues (e.g. type casting breaks).

To ~solve~ ***work around*** this, I moved `DeltaCore` into a separate package in a subdirectory of the `DeltaClient` repo (it'd be terrible for maintainability if `DeltaCore` was in a separate repo), and created two products; `DeltaCore` (dynamic), and `StaticDeltaCore`. The latter is useful because it's silly to have `DeltaCore` as a separate dylib if you're just using `DeltaCore` to build a Minecraft-related CLI (where plugins aren't useful) or using Delta Client on a platform where plugins aren't supported, such as iOS.

In addition to this, I've got a target called `DeltaRenderer` which would ideally be a separate product but can't be at the moment because otherwise it'd end up with its own copy of `DeltaCore` as well (even if I made it a dynamic library product). It's currently bundled in as part of the `DeltaCore` products, which isn't great cause the renderer is only available on certain platforms, and there are use-cases for `DeltaCore` that don't require a full-on Minecraft renderer.

A (not immediately obvious) downside of the separate packages approach is that people can't depend on the `delta-client` package using stable version requirements because one of its dependencies (`delta-core`) is a local path dependency.

---

Another improvement, which may help with Max's concerns, would be to introduce a mechanism for declaring internal products. Perhaps by adding an `internalProducts` parameter to the existing `Package` initializer?

Motivating this past 'we should do this for completeness' would probably require an example use case, and at the moment the only reasons for linking against same-package products that I can think of arise from the requirement to publicly expose a given product in the first place.

---

I believe that users of SwiftPM should be able to concoct any linking configurations they desire without splitting their project into multiple separate packages. This is especially true now that SwiftPM is being used to build large C++ projects which often have very specific linking requirements (which @furby-tm encountered while trying to write USD plugins in Swift).

I feel like this sentiment is something that most Swift devs can agree on or would at least be impartial to?

---

Regarding the next steps for this proposal, should we re-pitch on the Swift Forums, or would the existing forum thread be considered enough community consultation to move to the proposal phase after updating [my current proposal document](https://github.com/stackotter/swift-evolution/blob/wip-swiftpm-same-package-product-dependencies/proposals/NNNN-swiftpm-targets-same-package-product-dependencies.md)?

PR #7333, opened 2024-02-12
Ensure new packages have the correct minimum OS targets when using swift-testing.

grynspan:
swift-testing has a minimum deployment target set higher than the default for packages. We need to ensure that it is respected by `swift package init --enable-experimental-swift-testing`.


grynspan:
@swift-ci please test

neonichu (approved):

Sources/Workspace/InitPackage.swift
```diff
@@ -186,8 +186,8 @@ public final class InitPackage {
 
             var platforms = options.platforms
 
-            // Macros require macOS 10.15, iOS 13, etc.
-            if packageType == .macro {
+            // Macros and swift-testing require macOS 10.15, iOS 13, etc.
+            if packageType == .macro || options.supportedTestingLibraries.contains(.swiftTesting) {
```
MaxDesiatov, line 190: There doesn't seem to be an existing test that covers this check? IMO we shouldn't merge changes without corresponding test coverage, otherwise they're prone to regressions.

grynspan: Added test coverage for swift-testing. I did not write a new test for a macros package because that's a bit outside of my wheelhouse. I filed #7345 to track adding one.

grynspan:
@swift-ci please test

MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test windows

PR #7341, opened 2024-02-14
Fix test runner build on WASI

kateinoigakukun:
### Motivation:

The test runner build was failing on WASI after the introduction of the experimental summary output.

### Modifications:

This PR fixes several minor issues to pass `--build-tests` build on WASI:

* Missing `WASILibc` import
* The use of `flock` which is not available on WASI.
* Signature incompatibility of `XCTMain` on WASI.

### Result:

`swift build --build-tests --triple wasm32-unknown-wasi` will pass when compiler supports the target

### DIFF
--- a/Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
+++ b/Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
@@ -265,7 +265,14 @@ final class TestEntryPointCommand: CustomLLBuildCommand, TestBuildCommand {
                 struct Runner {
                     static func main() {
                         \#(testObservabilitySetup)
+                        #if os(WASI)
+                        // FIXME: On WASI, XCTest uses `Task` based waiting not to block the whole process, so
+                        // the `XCTMain` call can return the control and the process will exit by `exit(0)` later.
+                        // This is a workaround until we have WASI threads or swift-testing, which does not block threads.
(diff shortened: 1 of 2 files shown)


kateinoigakukun:
@swift-ci Please smoke test

MaxDesiatov:
@bnbarham actually there are more tests here I forgot about, but surfaced here.
@kateinoigakukun are these tests built only on SwiftWasm CI currently?

kateinoigakukun:
@MaxDesiatov What do you mean "tests" here? The test runner can be built on any developer's machine not limited to our CI

MaxDesiatov:
Ah, ok, sorry about the confusion. This is about `swift test` invocations, right? What do we need to do to make sure this doesn't regress in the future?

kateinoigakukun:
Ensuring that the runner can be built for WASI requires actual Swift SDK, so i think we don't have good way to have a regression test in this repository unfortunately 😢

kateinoigakukun:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

kateinoigakukun:
@swift-ci test windows

MaxDesiatov (approved):

PR #7349, opened 2024-02-19
Revert "Support macros when cross-compiling"

DougGregor:
Reverts apple/swift-package-manager#7118. It is causing the swift-syntax Examples to fail to build. See https://ci.swift.org/job/oss-swift-incremental-RA-macos/4247/

PR #7350, opened 2024-02-19
Build direct dependencies of tests for target triple

MaxDesiatov:
`assertMacroExpansion` requires macros to be compiled for target platform and evaluated in-process. We can't handle the case of indirect macro dependencies cleanly, those will be still compiled for the host. Supersedes https://github.com/apple/swift-package-manager/pull/7349.

This doesn't contain tests to expedite the fix and unblock broken CI jobs, but I've verified locally that test products of the `Examples` package in `apple/swift-syntax` repo can be built successfully and these tests all pass.

### DIFF
--- a/Sources/PackageGraph/Resolution/ResolvedPackage.swift
+++ b/Sources/PackageGraph/Resolution/ResolvedPackage.swift
@@ -34,7 +34,7 @@ public struct ResolvedPackage {
     public let underlying: Package
 
     /// The targets contained in the package.
-    public let targets: [ResolvedTarget]
+    public let targets: IdentifiableSet<ResolvedTarget>
 
     /// The products produced by the package.
     public let products: [ResolvedProduct]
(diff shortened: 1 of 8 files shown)


MaxDesiatov:
@swift-ci test

Sources/PackageGraph/Resolution/ResolvedPackage.swift
```diff
@@ -64,8 +64,41 @@ public struct ResolvedPackage {
...
+            if product.type == .test {
+                var targets = IdentifiableSet<ResolvedTarget>()
+                for var target in product.targets {
+                    var dependencies = [ResolvedTarget.Dependency]()
+                    for dependency in target.dependencies {
+                        switch dependency {
+                        case .target(var target, let conditions) where target.type == .macro:
+                            target.buildTriple = .destination
+                            dependencies.append(.target(target, conditions: conditions))
+                            processedTargets.insert(target)
```
ahoppen, line 83: Just to clarify: If the destination is the same as the host, we end up with a duplicate dependency declaration, right? Is that or could that be an issue?

ahoppen (approved):
Seems reasonable to me

Sources/PackageGraph/Resolution/ResolvedPackage.swift
MaxDesiatov: It's not necessarily a complete duplicate, as in the future it allows us to build tools (plugins and macros) with different parameters than destination products. For example, we could have macros and SwiftSyntax used in macros built in release mode, but end products (and for example another instance of SwiftSyntax used in end products) built in debug mode. If build parameters are exactly the same, we could apply an optimization and build only once, but as agreed with Boris previously, this optimization can be implemented in a separate PR.

ahoppen: OK, just wanted to make sure this wouldn’t cause any issues.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

PR #7355, opened 2024-02-20
Fix `.wasm` product paths for `wasm32-unknown-none` triple

MaxDesiatov:
Absence of this extension on produced binaries makes it inconvenient to work with triple, as you have to add that extension manually after the fact. For examples, HTTP servers frequently don't infer correct mime-type for these binaries, which confuses browsers that try to load these binaries.


Tests/BuildTests/BuildPlanTests.swift
```diff
@@ -3431,111 +3431,6 @@ final class BuildPlanTests: XCTestCase {
         XCTAssertMatch(executablePathExtension, "exe")
     }
 
-    func testWASITarget() throws {
```
MaxDesiatov, line 3434: Moved to `Tests/BuildTests/CrossCompilationBuildPlanTests.swift` to make `Tests/BuildTests/BuildPlanTests.swift` smaller and increase parallelization during testing, as separate `XCTestCase` types can be parallelized, while multiple functions in `XCTestCase` can't.

MaxDesiatov:
@swift-ci test

Sources/Basics/Triple+Basics.swift
```diff
@@ -148,6 +148,10 @@ extension Triple {
     }
 
     public var executableExtension: String {
+        guard ![.wasm32, .wasm64].contains(self.arch) else {
```
kateinoigakukun, line 151: How about adding `isWasm` like llvm's one?: https://github.com/llvm/llvm-project/blob/031f9f331723e6bebc405ffdee4b8a87a5fc0472/llvm/include/llvm/TargetParser/Triple.h#L1005

I think `[.wasm32, .wasm64].contains(self.triple.arch)` appears repeatedly, so it's worth adding memory space size independent predicate method.

MaxDesiatov: Good point, I've added it on an extension in `Basics/Triple+Basics.swift`

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

kateinoigakukun (approved):

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

PR #7358, opened 2024-02-22
Exclude `XCBuildSupport` when building with CMake

MaxDesiatov:
XCBuildSupport is not available on non-Darwin platforms, especially Windows, where we have to have full CMake support at the moment. Maintaining XCBuildSupport in these unsupported configurations adds unnecessary overhead, especially in cases like https://github.com/apple/swift-package-manager/pull/7258, where we have to add new dependencies only when `XCBuildSupport` is available. We should exclude from CMake builds to reduce this maintenance overhead.


Sources/swift-bootstrap/main.swift
```diff
@@ -330,6 +334,9 @@ struct SwiftBootstrapBuildTool: ParsableCommand {
                     fileSystem: self.fileSystem,
                     observabilityScope: self.observabilityScope
                 )
+                #else
+                fatalError("SwiftPM was built without XCBuild support")
```
MaxDesiatov, line 338: This line is technically unreachable, since `case .xcode` is never reachable on non-Darwin platforms, and we don't call `swift-bootstrap --build-system xcode` anywhere during the build process.

Sources/Commands/CMakeLists.txt
```diff
@@ -59,8 +59,11 @@ target_link_libraries(Commands PUBLIC
   SourceControl
   TSCBasic
   TSCUtility
-  Workspace
-  XCBuildSupport)
+  Workspace)
+
+target_compile_definitions(Commands
+  PRIVATE SKIP_XCBUILD_SUPPORT)
```
compnerd, line 65: I think that `DISABLE_XCBUILD_SUPPORT` would be better ... its less "skipping".

compnerd (approved):
Thanks! This seems like a good idea to me.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

bnbarham:
> We should exclude from CMake builds to reduce this maintenance overhead.

Don't we bootstrap with CMake today? The toolchain builds are blocked because we end up hitting the error in `Sources/swift-bootstrap/main.swift` 😅.

Sources/swift-bootstrap/main.swift
bnbarham: > This line is technically unreachable

Heh, famous last words:
https://ci.swift.org/job/oss-swift-package-macos/2753

```
--- bootstrap: note: Building SwiftPM (with a freshly built swift-bootstrap)
main/main.swift:345: Fatal error: SwiftPM was built without XCBuild support
```

bnbarham:
```
bootstrap/bin/swift-bootstrap ... --arch x86_64 --arch arm64 ...
```

And:
```
    private var buildSystem: BuildSystemProvider.Kind {
        #if os(macOS)
        // Force the Xcode build system if we want to build more than one arch.
        return self.architectures.count > 1 ? .xcode : .native
        #else
        // Force building with the native build system on other platforms than macOS.
        return .native
        #endif
    }
```

PR #7360, opened 2024-02-22
Allow reading `Package.swift` with the executable bit set

bnbarham:
786c5131ed5ec823d3be90ad31923123fcaccd56 added support for symlinks, but didn't handle `executableBlob` in the added `switch`. `executableBlob` is the returned type for a file with the executable bit set, so needs to be handled in the same way as a regular `blob`.


bnbarham:
@swift-ci please test

MaxDesiatov (approved):

PR #7364, opened 2024-02-23
Fix some warnings with `ProcessEnvironmentBlock`

MaxDesiatov:
Previous implementation didn't account for case insensitivity on Windows. Cleaning up these warnings fixes that.

This doesn't include APIs exposed by Swift Driver yet, as those are harder to adapt for `ProcessEnvironmentBlock` with significant sources compatibility breakage.

### DIFF
--- a/Sources/Basics/EnvironmentVariables.swift
+++ b/Sources/Basics/EnvironmentVariables.swift
@@ -11,8 +11,11 @@
 //===----------------------------------------------------------------------===//
 
 import class Foundation.ProcessInfo
+import typealias TSCBasic.ProcessEnvironmentBlock
+import struct TSCBasic.ProcessEnvironmentKey
+import enum TSCBasic.ProcessEnv
 
-public typealias EnvironmentVariables = [String: String]
+public typealias EnvironmentVariables = ProcessEnvironmentBlock
 
 extension EnvironmentVariables {
     public static func empty() -> EnvironmentVariables {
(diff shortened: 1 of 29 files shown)


MaxDesiatov:
https://github.com/apple/swift-tools-support-core/pull/470
@swift-ci test

Sources/Basics/EnvironmentVariables.swift
```diff
@@ -11,35 +11,38 @@
 //===----------------------------------------------------------------------===//
 
 import class Foundation.ProcessInfo
+import typealias TSCBasic.ProcessEnvironmentBlock
+import struct TSCBasic.ProcessEnvironmentKey
+import enum TSCBasic.ProcessEnv
 
-public typealias EnvironmentVariables = [String: String]
+public typealias EnvironmentVariables = ProcessEnvironmentBlock
```
compnerd, line 18: Do we need to worry about this accidentally causing build issues?

compnerd (approved):

Sources/Basics/EnvironmentVariables.swift
MaxDesiatov: I don't think we expose it anywhere outside of SwiftPM to any clients, at least on purpose. IIUC corresponding TSC APIs were widely adopted instead. As soon as `package` support is merged in my other PR, I'll try using `package` visibility here to prevent any unintended adoption of this typealias in the future.

bnbarham: Could use SPI in the meantime, would also make it easy to remember all the places we want `package` 😅

MaxDesiatov:
@swift-ci test

Sources/Basics/EnvironmentVariables.swift
bnbarham: And can now use `package`.

MaxDesiatov:
@swift-ci test

PR #7367, opened 2024-02-24
Add `AsyncBuildOperation` alternative to `BuildOperation`

MaxDesiatov:
Depends on https://github.com/apple/swift-package-manager/pull/7365.

For now this is mostly a copy of `BuildOperation`, but with `async` entrypoints. This can be expanded to use more structured concurrency features in the future as we migrate more code. It is disabled by default, for testing purposes I'm exposing it as `swift build --build-system experimental-async`, which creates and uses `AsyncBuildOperation` instead of blocking synchronous `BuildOperation`.

PR #7369, opened 2024-02-24
Allow `swift-build` to have `async` entrypoint

MaxDesiatov:
Due to a bug in Swift 5.9, we can't have multiple executables with `async` entrypoints linked into test products. This was the reason why we weren't able to make `swift-build` entrypoint `async`. Since we have a single `swift-package-manager` executable anyway, we can link that one into tests instead, and pass "simulated" executable name via an environment variable. This allows us to work around the bug and start using structured concurrency from `swift-build` entrypoint.

Also converted `SwiftRunCommand` entrypoint to `async` to reliably reproduce possible regressions highlighted by `RunCommandTests.testSwiftRunSIGINT`. These regressions were handled by resetting signals masks and closing file descriptors before a subsequent call to `execv`.

This change is a prerequisite for https://github.com/apple/swift-package-manager/pull/7367.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci clean test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

Tests/CommandsTests/RunCommandTests.swift
```diff
@@ -102,6 +103,12 @@ final class RunCommandTests: CommandsTestCase {
     }
 
     func testSwiftRunSIGINT() throws {
+        // FIXME: not sure how this was supposed to work previously, since `swift-run` uses `execv` that doesn't inherit
+        // signal handlers, and the spawned process doesn't install signal handlers on its own. `async` effect on
+        // `@main` arguably makes it behave correctly?
```
kateinoigakukun, line 108: It seems this issue is not only a problem in tests but causes a real problem. I could reproduce this uninterruptable swift-run when invoking just built `swift-package-manager` through symlink named `swift-run`.

Based on my quick research, it seems like [libdispatch blocks `SIGINT` by `sigmask`](https://github.com/apple/swift-corelibs-libdispatch/blob/5529d351e304eb2312b7b3bfb073f7d8285473e0/src/init.c#L97) and [spawns a dedicated thread to catch those signals](https://github.com/apple/swift-corelibs-libdispatch/blob/5529d351e304eb2312b7b3bfb073f7d8285473e0/src/queue.c#L7003).

So the problematic code can be minimized as follows.

```swift
import Dispatch
import Foundation

DispatchQueue.main.async {
  let bin = "/usr/bin/sleep"
  let args = ["sleep", "10"]

  let cArgs = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: args.count + 1)
  for (i, arg) in args.enumerated() {
    cArgs[i] = strdup(arg)
  }
  cArgs[args.count] = nil

  execv(bin, cArgs)
  perror("execv")
}

dispatchMain()
```

I confirmed we can unblock signals as follows, but I'm not 100% sure it's a safe way.

```swift
var mask = sigset_t()
sigemptyset(&mask)
pthread_sigmask(SIG_SETMASK, &mask, nil)
execv(bin, cArgs)
```

kateinoigakukun (changes requested):

Tests/CommandsTests/RunCommandTests.swift
MaxDesiatov: IIUC Dispatch blocks delivery of `SIGINT` whenever some code is executed in `DispatchQueue.main.async`, is that correct? Isn't it a bug in Dispatch then and could be fixed on their side?

kateinoigakukun: No, libdispatch blocks `SIGINT` handler invocations during the whole program execution, but it creates a dedicated thread to listen to signals before blocking the signals in the main thread. `SIGINT` is usually handled by the dedicated thread and I guess the actual signal handling works are scheduled on some dispatch queues. So this is not a bug in libdispatch at all, but they don't provide a graceful way to reset signal mask states. Therefore I suggested resetting masks manually just before execv in SwiftPM as one of the options.

MaxDesiatov:
@swift-ci clean test windows

Sources/Commands/SwiftRunCommand.swift
```diff
@@ -284,9 +284,20 @@ public struct SwiftRunCommand: SwiftCommand {
...
+        for i in 1..<NSIG {
+            signal(i, SIG_DFL)
+        }
+        var sig_set_all = sigset_t()
+        sigfillset(&sig_set_all)
+        sigprocmask(SIG_UNBLOCK, &sig_set_all, nil)
+        
+        // 2. close all file descriptors.
+        for i in 3..<getdtablesize() {
+            close(i)
```
kateinoigakukun, line 299: Do we have any open fds managed by SwiftPM at this point? My concern is that this might corrupt the state of underlying libraries like libc and cause other problems for the case when `exec` fails for some reason. I think closing fds is not a part of this function's responsibilities but should be handled by those fds owners using `O_CLOEXEC`.

MaxDesiatov: > Do we have any open fds managed by SwiftPM at this point?
I don't think we should, but I don't have any certainty our `swift-tools-support-core` or `swift-corelibs-foundation` APIs don't leak anything.

MaxDesiatov:
@swift-ci test windows

Sources/Commands/SwiftRunCommand.swift
kateinoigakukun: I think the discussion point here is whether to prioritize potential state corruption after the exec fails or potential fd leak during the exec'ed program execution. IMO the former issue is more critical than the latter since we don't have a large number of open fd's here and emitting proper error diagnostics when a problem occurs is essential for further investigation.

MaxDesiatov: IIUC the argument that Johannes is making is that `bash` is doing it this way and it works fine for them, whatever state underlying libraries have, it'll be gone with `execv` call anyway https://github.com/bminor/bash/blob/f3b6bd19457e260b65d11f2712ec3da56cef463f/execute_cmd.c#L6162C1-L6173

weissi: Yes, that's what UNIX programs are expected to do if they exec: close the file descriptors that they're not intending to pass on. LGTM

kateinoigakukun: Okay, that makes sense to me 👍

kateinoigakukun (approved):

PR #7377, opened 2024-02-28
Add swift-testing support to `swift build --build-tests`

### LINKED ISSUE #7375: `swift build --build-tests` doesn't build swift-testing tests

- [X] Confirmed reproduction steps with SwiftPM CLI.

### Description

When I run `swift build --build-tests` it doesn't build the swift-testing tests.

### Expected behavior

`--build-tests` builds swift-testing tests.

### Actual behavior

Doesn't build swift-testing tests.

```
error: posix_spawn error: No such file or directory (2), `["/tmp/package-name/.build/arm64-apple-macosx/debug/package-namePackageTests.swift-testing", "--list-tests"]`
error: ExitCode(rawValue: 1)
[0/1] Planning builderror: no tests found; create a target in the 'Tests' directory
```

### Steps to reproduce

1. Get hold of a package that uses swift-testing
2. `rm -rf .build  # to make sure this is clean`
3. `swift build --build-tests`
4. `swift test list --skip-build`

_No response_

macOS

grynspan:
This PR generalizes the enable/disable XCTest/swift-testing options from `swift test` and `swift package init` and adds them to `swift build --build-tests` so that it is possible to build all test content without actually running the tests. If neither flag is specified, the default test content is built (always XCTest, and swift-testing if it's a dependency, same as `swift test`.)

I don't have unit tests for this work because SwiftPM cannot currently reference or link to swift-testing in CI. I'm open to ideas here.

Resolves #7375.

### DIFF
--- a/Sources/Commands/PackageCommands/Init.swift
+++ b/Sources/Commands/PackageCommands/Init.swift
@@ -44,7 +44,7 @@ extension SwiftPackageCommand {
 
         /// Which testing libraries to use (and any related options.)
         @OptionGroup()
-        var testingLibraryOptions: TestLibraryOptions
+        var testLibraryOptions: TestLibraryOptions
 
         @Option(name: .customLong("name"), help: "Provide custom package name")
         var packageName: String?
@@ -54,13 +54,13 @@ extension SwiftPackageCommand {
                 throw InternalError("Could not find the current working directory")
             }
 
-            // NOTE: Do not use testingLibraryOptions.enabledTestingLibraries(swiftCommandState:) here
+            // NOTE: Do not use testLibraryOptions.enabledTestingLibraries(swiftCommandState:) here
             // because the package doesn't exist yet, so there are no dependencies for it to query.
             var testingLibraries: Set<BuildParameters.Testing.Library> = []
-            if testingLibraryOptions.enableXCTestSupport {
+            if testLibraryOptions.enableXCTestSupport {
                 testingLibraries.insert(.xctest)
             }
-            if testingLibraryOptions.explicitlyEnableSwiftTestingLibrarySupport == true {
+            if testLibraryOptions.explicitlyEnableSwiftTestingLibrarySupport == true {
                 testingLibraries.insert(.swiftTesting)
             }
             let packageName = self.packageName ?? cwd.basename
--- a/Sources/Commands/PackageCommands/Init.swift
+++ b/Sources/Commands/PackageCommands/Init.swift
@@ -42,17 +42,9 @@ extension SwiftPackageCommand {
                 """))
         var initMode: InitPackage.PackageType = .library
 
-        /// Whether to enable support for XCTest.
-        @Flag(name: .customLong("xctest"),
-              inversion: .prefixedEnableDisable,
-              help: "Enable support for XCTest")
-        var enableXCTestSupport: Bool = true
-
-        /// Whether to enable support for swift-testing.
-        @Flag(name: .customLong("experimental-swift-testing"),
-              inversion: .prefixedEnableDisable,
-              help: "Enable experimental support for swift-testing")
-        var enableSwiftTestingLibrarySupport: Bool = false
+        /// Which testing libraries to use (and any related options.)
+        @OptionGroup()
+        var testingLibraryOptions: TestLibraryOptions
 
         @Option(name: .customLong("name"), help: "Provide custom package name")
         var packageName: String?
@@ -62,11 +54,13 @@ extension SwiftPackageCommand {
                 throw InternalError("Could not find the current working directory")
             }
 
+            // NOTE: Do not use testingLibraryOptions.enabledTestingLibraries(swiftCommandState:) here
+            // because the package doesn't exist yet, so there are no dependencies for it to query.
             var testingLibraries: Set<BuildParameters.Testing.Library> = []
-            if enableXCTestSupport {
+            if testingLibraryOptions.enableXCTestSupport {
                 testingLibraries.insert(.xctest)
             }
-            if enableSwiftTestingLibrarySupport {
+            if testingLibraryOptions.explicitlyEnableSwiftTestingLibrarySupport == true {
                 testingLibraries.insert(.swiftTesting)
             }
             let packageName = self.packageName ?? cwd.basename
--- a/Sources/Commands/SwiftTestCommand.swift
+++ b/Sources/Commands/SwiftTestCommand.swift
@@ -150,7 +150,9 @@ struct TestLibraryOptions: ParsableArguments {
     }
 
     /// Get the set of enabled testing libraries.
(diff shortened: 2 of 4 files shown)


weissi (approved):
I can't claim to fully understand the code but it does look good to me. Thank you!!

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -132,6 +148,20 @@ struct SharedOptions: ParsableArguments {
         // Default to disabled since swift-testing is experimental (opt-in.)
         return false
     }
+
+    /// Get the set of enabled testing libraries.
+    func enabledTestingLibraries(swiftCommandState: SwiftCommandState) throws -> Set<BuildParameters.Testing.Library> {
```
MaxDesiatov, line 153: Formatting nit: we avoid long lines for better readability in diff views and smaller screen sizes.
```suggestion
    func enabledTestingLibraries(
        swiftCommandState: SwiftCommandState
    ) throws -> Set<BuildParameters.Testing.Library> {
```

grynspan:
@swift-ci please test

Sources/Commands/SwiftTestCommand.swift
grynspan: What's the cutoff you generally prefer? I can set my gutter to that.

MaxDesiatov: We have all of the formatting rules specified in the formatter config https://github.com/apple/swift-package-manager/blob/main/.swiftformat. Just run `./Utilities/soundness.sh` before committing and that will reformat changed files for you with that config.

MaxDesiatov (commented):
Would you mind adding a test case to prevent future regressions?

grynspan:
> Would you mind adding a test case to prevent future regressions?

See the PR description. How would you implement it?

Sources/Commands/SwiftTestCommand.swift
grynspan: That script did not produce any diagnostics nor make any changes to my branch.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

Sources/Commands/PackageCommands/Init.swift
```diff
@@ -42,17 +42,9 @@ extension SwiftPackageCommand {
...
-        var enableXCTestSupport: Bool = true
-
-        /// Whether to enable support for swift-testing.
-        @Flag(name: .customLong("experimental-swift-testing"),
-              inversion: .prefixedEnableDisable,
-              help: "Enable experimental support for swift-testing")
-        var enableSwiftTestingLibrarySupport: Bool = false
+        /// Which testing libraries to use (and any related options.)
+        @OptionGroup()
+        var testingLibraryOptions: TestLibraryOptions
```
MaxDesiatov, line 47: there's a naming inconsistency here, which can make it harder to use textual search when looking for relevant properties or types, one way to resolve it could be this

```suggestion
        var testLibraryOptions: TestLibraryOptions
```

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -75,23 +75,39 @@ struct SharedOptions: ParsableArguments {
     /// to choose from (usually in multiroot packages).
     @Option(help: .hidden)
     var testProduct: String?
+}
 
-    /// Whether to enable support for XCTest.
+/// Which testing libraries to use (and any related options.)
+struct TestLibraryOptions: ParsableArguments {
```
MaxDesiatov, line 81: This type belongs to `CoreCommands/Options.swift` or (ideally) a separate file. `SwiftTestCommand.swift` is already too big and hard to navigate, we should avoid adding more types or extensions to it.

MaxDesiatov:
> See the PR description. How would you implement it?

I think we need a new `BuildSystemProvider` that can create mock build systems. Then we can initialize `SwiftCommandState` with desired options in the test and verify that the mock build system is passed correct build parameters for correct build subsets.

Sources/Commands/PackageCommands/Init.swift
grynspan: I will rename it.

Sources/Commands/SwiftTestCommand.swift
grynspan: I will move it.

grynspan:
> > See the PR description. How would you implement it?
> 
> I think we need a new `BuildSystemProvider` that can create a mock build system. Then we can initialize `SwiftCommandState` with desired options in the test and verify that the mock build system is passed correct build parameters for correct build subsets.

It looks like `--build-tests` doesn't really have any test coverage right now. I filed #7378 for test coverage for that option.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

MaxDesiatov:
> It looks like `--build-tests` doesn't really have any test coverage right now.

I understand, this shouldn't be the reason to reduce our overall test coverage though.

grynspan:
> > It looks like `--build-tests` doesn't really have any test coverage right now.
> 
> I understand, this shouldn't be the reason to reduce our overall test coverage though.

I don't believe I'm reducing it as this feature _already_ does not have any test coverage, but maybe I'm misunderstanding? 🙂

weissi:
> > It looks like `--build-tests` doesn't really have any test coverage right now.
> 
> I understand, this shouldn't be the reason to reduce our overall test coverage though.

I'd argue that right now the feature is untested (bad) but also not working (worse). Why don't we make the situation better by fixing it? Once this is merged (and therefore fixed) we can maybe team up to write at least some basic tests?

grynspan:
> > > It looks like `--build-tests` doesn't really have any test coverage right now.
> > 
> > 
> > I understand, this shouldn't be the reason to reduce our overall test coverage though.
> 
> I'd argue that right now the feature is untested (bad) but also not working (worse). Why don't we make the situation better by fixing it? Once this is merged (and therefore fixed) we can maybe team up to write at least some basic tests?

That sounds like an excellent plan!

finagolfin:
It appears the lack of tests hurt us here: this pull broke `swift build --build-tests` registering the `--enable-test-discovery` flag also. [My Android CI builds `swift-numerics` trunk daily with the latest Swift trunk snapshot builds and those flags](https://github.com/finagolfin/swift-android-sdk/blob/c5828e445d9c61489b77049f03a57d95bda15d75/.github/workflows/sdks.yml#L196): [it now errors with the March 1 snapshot](https://github.com/finagolfin/swift-android-sdk/actions/runs/8130039289/job/22222723739#step:3:6675).

I can reproduce the crash on linux x86_64 with the following commands, which work with the last Feb. 29 trunk snapshot before this pull:
```
> ../swift-DEVELOPMENT-SNAPSHOT-2024-03-01-a-ubi9/usr/bin/swift build --build-tests --enable-test-discovery
> .build/x86_64-unknown-linux-gnu/debug/swift-numericsPackageTests.xctest
```

PR #7381, opened 2024-02-29
Hide CLI commands API with `package` access control

MaxDesiatov:
Known clients of SwiftPM have not adopted these CLI APIs that were previously marked as `public`. To prevent accidental adoption by clients in the future, we should use `package` access control instead. 

Also replaced remaining usages of `@_spi(SwiftPMInternal)` with `package` across all of our codebase.


MaxDesiatov:
@swift-ci test

bnbarham (approved):

PR #7382, opened 2024-02-29
Remove uses of `temp_await` from `SwiftTestCommand`

MaxDesiatov:
This prevents possible deadlocks that can be caused by blocking Swift Concurrency threads with semaphores.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/PackageGraph/Resolution/PubGrub/PubGrubDependencyResolver.swift
```diff
@@ -749,6 +750,15 @@ public struct PubGrubDependencyResolver {
             }
         }
     }
+
+    func makeDecision(
+        state: State,
+        availableLibraries: [LibraryMetadata] = []
+    ) async throws -> DependencyResolutionNode? {
+        try await withCheckedThrowingContinuation {
+            makeDecision(state: state, availableLibraries: availableLibraries, completion: $0.resume)
```
bnbarham, line 759: I notice we have a `safe_async`, are we worried about `makeDecision` possibly blocking indefinitely at all?

bnbarham (approved):

MaxDesiatov:
@swift-ci test windows

Sources/PackageGraph/Resolution/PubGrub/PubGrubDependencyResolver.swift
MaxDesiatov: In an abundance of caution I've removed those changes from the diff 😅

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/PackageGraph/Resolution/PubGrub/PubGrubDependencyResolver.swift
bnbarham: To be clear, I didn't any real issue with this. It's just that I noticed the `safe_async` and wondered if it was added because we had been bitten in the past 😅

MaxDesiatov: It was added by @AndrewHoos and reasoning for that seems sensible to me. I guess in the future we should use `safe_async` for code like this if we can't come up with a more `async`-friendly way to do it.

PR #7384, opened 2024-02-29
Make ignore-lock a flag

rauhul:
This allows users to ignore package locking via `--ignore-lock` instead of `--ignore-lock true`.


rauhul:
@swift-ci test

MaxDesiatov (commented):
Is `--ignore-lock true` still allowed? If not, I wonder how many workflows we would break or if there's any way to retain compatibility?

rauhul:
With this change the old form is not allowed, I'm happy to also not make this change actually...

rauhul:
> --ignore-lock is brand new, I wouldn't worry about breaking existing workflows (#7338). Thanks for noticing @rauhul (as an aside, did you need this?)!

Sadly yes/no;
- I have a test `MMIOFileCheckTests` in `swift-mmio` that must only be run against the `MMIO` library built in release.
- My test calls into `swift build -c release` to ensure the proper dependency is available at test-run-time because I can't express this in `Package.swift`.
- I re-use the same scratch build area because the `swift-syntax` rebuild time is too painful to do a full separate build.

MaxDesiatov (approved):

rauhul:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

PR #7386, opened 2024-03-03
Add XDG Base Directory Specification support

### LINKED ISSUE #6204: ~/.swiftpm config, cache to follow the XDG base dir spec
### Description

Conform to the [XDG Base Directory spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) with no symlinks from the "$HOME" directory. 

I assume the purpose of these symlinks is discoverability.

An alternative solution would be to provide an option to not create the symlinks in the "$HOME" directory e.g. env variable

### Expected behavior

No symlinks "~/.swiftpm" in the home directory

### Actual behavior

_No response_

### Steps to reproduce

_No response_

_No response_

macOS 
Linux - all distributions
windows

lordzsolt:
### Motivation:

As discussed in #6204, users would like to have a clean home directory.
And in the discussions following the merge of #3430, it seems there is a willingness to adhere to the XDG spec, and support `~/.config/swiftpm`.

### Modifications:

If the `XDG_CONFIG_HOME` environmental variable is defined, use `$XDG_CONFIG_HOME/swiftpm` as the root `dotSwiftPM` directory.

### Result:

The symlinks that were previously stored in `~/.swiftpm` are now stored in `$XDG_CONFIG_HOME/swiftpm` when this variable is defined, therefore not cluttering the home directory of users.

Closes #6204

### DIFF
--- a/Sources/Basics/FileSystem/FileSystem+Extensions.swift
+++ b/Sources/Basics/FileSystem/FileSystem+Extensions.swift
@@ -215,7 +215,7 @@ extension FileSystem {
     /// or under $XDG_CONFIG_HOME/swiftpm if the environmental variable is defined
     public var dotSwiftPM: AbsolutePath {
         get throws {
-            if let configurationDirectory = ProcessInfo.processInfo.environment["XDG_CONFIG_HOME"] {
+            if let configurationDirectory = EnvironmentVariables.process()["XDG_CONFIG_HOME"] {
                 return try AbsolutePath(validating: configurationDirectory).appending("swiftpm")
             } else {
                 return try self.homeDirectory.appending(".swiftpm")
--- a/Sources/Basics/FileSystem/FileSystem+Extensions.swift
+++ b/Sources/Basics/FileSystem/FileSystem+Extensions.swift
@@ -12,7 +12,6 @@
 
 import struct Foundation.Data
 import class Foundation.FileManager
-import class Foundation.ProcessInfo
 import struct Foundation.UUID
 import SystemPackage
 
--- a/Sources/Basics/FileSystem/FileSystem+Extensions.swift
+++ b/Sources/Basics/FileSystem/FileSystem+Extensions.swift
@@ -12,6 +12,7 @@
 
 import struct Foundation.Data
 import class Foundation.FileManager
+import class Foundation.ProcessInfo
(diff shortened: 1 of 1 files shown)


Sources/Basics/FileSystem/FileSystem+Extensions.swift
```diff
@@ -211,9 +212,14 @@ extension FileSystem {
 
 extension FileSystem {
     /// SwiftPM directory under user's home directory (~/.swiftpm)
+    /// or under $XDG_CONFIG_HOME/swiftpm if the environmental variable is defined
     public var dotSwiftPM: AbsolutePath {
         get throws {
-            try self.homeDirectory.appending(".swiftpm")
+            if let configurationDirectory = ProcessInfo.processInfo.environment["XDG_CONFIG_HOME"] {
```
MaxDesiatov, line 218: Let's use [`EnvironmentVariables`](https://github.com/apple/swift-package-manager/blob/main/Sources/Basics/EnvironmentVariables.swift) from this same `Basics` module. With https://github.com/apple/swift-package-manager/pull/7364 `EnvironmentVariables` can also handle case-insensitivity of environment variable keys on Windows. Not sure how applicable XDG is on Windows, but since you're not excluding this logic on that platform, let's handle it properly on all platforms for now.

```suggestion
            if let configurationDirectory = EnvironmentVariables.process()["XDG_CONFIG_HOME"] {
```

MaxDesiatov (changes requested):
We either need to use `EnvironmentVariables.process()` instead of `ProcessInfo.processInfo.environment` for improved Windows handling in the future, or exclude this logic on Windows altogether.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

alichtman:
Thank you!!

bartekpacia:
Thank you for working on this!

Will this be released in Swift 6.0, or maybe earlier?

MaxDesiatov:
5.10 has been released already. Swift 6.0 is [branching off `main` tomorrow](https://forums.swift.org/t/swift-6-0-release-process/70220).

PR #7387, opened 2024-03-03
Hide more llbuild-specific APIs

MaxDesiatov:
These APIs should not be marked `public` to prevent SwiftPM clients from adopting them by accident.

### DIFF
--- a/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
@@ -28,19 +28,19 @@ import DriverSupport
 import struct TSCBasic.ByteString
 
 /// Target description for a Swift target.
-public final class SwiftTargetBuildDescription {
+package final class SwiftTargetBuildDescription {
     /// The package this target belongs to.
-    public let package: ResolvedPackage
+    package let package: ResolvedPackage
 
     /// The target described by this target.
-    public let target: ResolvedTarget
+    package let target: ResolvedTarget
 
     private let swiftTarget: SwiftTarget
 
     /// The tools version of the package that declared the target.  This can
     /// can be used to conditionalize semantically significant changes in how
     /// a target is built.
-    public let toolsVersion: ToolsVersion
+    package let toolsVersion: ToolsVersion
 
     /// The build parameters.
     let buildParameters: BuildParameters
@@ -77,22 +77,22 @@ public final class SwiftTargetBuildDescription {
     }
 
     /// The list of all source files in the target, including the derived ones.
-    public var sources: [AbsolutePath] {
+    package var sources: [AbsolutePath] {
         self.target.sources.paths + self.derivedSources.paths + self.pluginDerivedSources.paths
     }
 
-    public var sourcesFileListPath: AbsolutePath {
+    package var sourcesFileListPath: AbsolutePath {
         self.tempsPath.appending(component: "sources")
     }
 
     /// The list of all resource files in the target, including the derived ones.
-    public var resources: [Resource] {
+    package var resources: [Resource] {
         self.target.underlying.resources + self.pluginDerivedResources
     }
 
     /// The objects in this target, containing either machine code or bitcode
     /// depending on the build parameters used.
-    public var objects: [AbsolutePath] {
+    package var objects: [AbsolutePath] {
         get throws {
             let relativeSources = self.target.sources.relativePaths
                 + self.derivedSources.relativePaths
@@ -112,7 +112,7 @@ public final class SwiftTargetBuildDescription {
     }
 
     /// The path to the swiftmodule file after compilation.
-    public var moduleOutputPath: AbsolutePath { // note: needs to be public because of sourcekit-lsp
+    package var moduleOutputPath: AbsolutePath { // note: needs to be package because of sourcekit-lsp
         // If we're an executable and we're not allowing test targets to link against us, we hide the module.
         let triple = buildParameters.triple
         let allowLinkingAgainstExecutables = (triple.isDarwin() || triple.isLinux() || triple.isWindows()) && self.toolsVersion >= .v5_5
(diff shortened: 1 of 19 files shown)


MaxDesiatov:
@swift-ci test

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
```diff
@@ -112,7 +112,7 @@ public final class SwiftTargetBuildDescription {
     }
 
     /// The path to the swiftmodule file after compilation.
-    public var moduleOutputPath: AbsolutePath { // note: needs to be public because of sourcekit-lsp
+    package var moduleOutputPath: AbsolutePath { // note: needs to be package because of sourcekit-lsp
```
MaxDesiatov, line 115:
```suggestion
    public var moduleOutputPath: AbsolutePath { // note: needs to be `public` because of sourcekit-lsp
```

Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
```diff
@@ -166,7 +166,7 @@ final class TestDiscoveryCommand: CustomLLBuildCommand, TestBuildCommand {
                 import XCTest
 
                 @available(*, deprecated, message: "Not actually deprecated. Marked as deprecated to allow inclusion of deprecated tests (which test deprecated functionality) without warnings")
-                public func __allDiscoveredTests() -> [XCTestCaseEntry] {
+                package func __allDiscoveredTests() -> [XCTestCaseEntry] {
```
MaxDesiatov, line 169:
```suggestion
                public func __allDiscoveredTests() -> [XCTestCaseEntry] {
```

Sources/Build/TestObservation.swift
```diff
@@ -22,8 +22,8 @@ public func generateTestObservationCode(buildParameters: BuildParameters) -> Str
         import Foundation
         import XCTest
 
-        public final class SwiftPMXCTestObserver: NSObject {
-            public override init() {
+        package final class SwiftPMXCTestObserver: NSObject {
+            package override init() {
```
MaxDesiatov, line 26:
```suggestion
        public final class SwiftPMXCTestObserver: NSObject {
            public override init() {
```

MaxDesiatov:
@swift-ci test

rauhul:
I'd just like to say thanks for going through the work of minifying the SwiftPM api surface, this makes it much easier to make changes without breaking clients

Sources/Build/BuildPlan/BuildPlan.swift
```diff
@@ -246,7 +246,7 @@ public class BuildPlan: SPMBuildCore.BuildPlan {
     let observabilityScope: ObservabilityScope
 
     @available(*, deprecated, renamed: "init(productsBuildParameters:toolsBuildParameters:graph:)")
-    public convenience init(
+    package convenience init(
```
MaxDesiatov, line 249:
```suggestion
    public convenience init(
```

Sources/Build/BuildPlan/BuildPlan.swift
```diff
@@ -268,7 +268,7 @@ public class BuildPlan: SPMBuildCore.BuildPlan {
     }
 
     /// Create a build plan with a package graph and explicitly distinct build parameters for products and tools.
-    public init(
+    package init(
```
MaxDesiatov, line 271:
```suggestion
    public init(
```

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

bnbarham (approved):

krzyzanowskim:
@MaxDesiatov What is the replacement API to make builds now? I use `BuildOperation` just like `SwiftBuildCommand` does.

Now, only SPM can use that API and that doesn't seem fair to me. what am I missing here?

MaxDesiatov:
There is no replacement API outside of `PackagePlugin`, this wasn't supposed to be an API in the first place and was exposed because `package` access control was unavailable at the time.

Build system internals is an unversioned implementation detail, which can and will be changed without notice. Removing `public` from these symbols allows us to evolve it and migrate to structured concurrency or overhaul how it works altogether without constraints of potentially breaking external adopters.

Keeping them `public` effectively freezes them and require us to maintain that API shape indefinitely, which is not something that can be provided at the moment.

krzyzanowskim:
> Keeping them public effectively freezes them and require us to maintain that API shape indefinitely, which is not something that can be provided at the moment.

that is not true. It was never "freezed" by any means. It is actively used by the CLI client and could be used by third-party clients. Now that is no longer possible 😢

MaxDesiatov:
> I don't see a clear reason to shut down the API completly for other clients.

I think I've provided enough reasons in this PR and the linked issue, I don't see a point in going in circles and rehashing those again.

> if API is good enough for CLI, it's good enough for third-party clients

I don't see how this conjecture is made. Needs, requirements, and expectations of CLI maintained in the same package as other SwiftPM internals, as opposed to third-party clients, are significantly different. An internal SPI good enough for CLI maintained in the same package is almost always _not_ good enough for third-party clients, as the latter require stability, versioning, and a commitment of continued maintenance of those public APIs.

> Maybe swift SPI mechanism is better suited for that after all.

This specific point in isolation is correct. I'm considering what still remains `public` outside of `PackageDescription` and `PackagePlugin` modules to be hidden with `@_spi(SwiftPMInternal)` to additionally indicate that symbols outside of those two modules should not be adopted and relied on by 3rd-party clients.

krzyzanowskim:
> An internal SPI good enough for CLI maintained in the same package is almost always not good enough for third-party clients, as the latter require stability, versioning, and a commitment of continued maintenance of those public APIs.

disagree. sourcekit-lsp relies on that unstable API, and stability is not a requirement. that's my whole point. API does not have to be stable unless declared as such, which then sets an expectation. I don't think the remaining public API is stable, too anyway.

MaxDesiatov:
I'll reiterate again that SourceKit-LSP case is different as it is always built in lockstep with SwiftPM. For it API stability is not a requirement because it can't be broken for it, we literally can't merge a SwiftPM PR that breaks SourceKit-LSP, required CI jobs won't pass in that case.

This is not the case for other clients.

krzyzanowskim:
> This is not the case for other clients.

I'll reiterate again that this is an artificial requirement you put on that. As an "other clients" representative, I can assure you that's not an issue for me, and I accept the consequences of an unstable API

Issue #7389, opened 2024-03-04
`swift build --build-tests` doesn't work with `--enable-test-discovery` flag


MaxDesiatov:
It appears the lack of tests hurt us here: this pull broke `swift build --build-tests` registering the `--enable-test-discovery` flag also. [My Android CI builds `swift-numerics` trunk daily with the latest Swift trunk snapshot builds and those flags](https://github.com/finagolfin/swift-android-sdk/blob/c5828e445d9c61489b77049f03a57d95bda15d75/.github/workflows/sdks.yml#L196): [it now errors with the March 1 snapshot](https://github.com/finagolfin/swift-android-sdk/actions/runs/8130039289/job/22222723739#step:3:6675).

I can reproduce the crash on linux x86_64 with the following commands, which work with the last Feb. 29 trunk snapshot before this pull:
```
> ../swift-DEVELOPMENT-SNAPSHOT-2024-03-01-a-ubi9/usr/bin/swift build --build-tests --enable-test-discovery
> .build/x86_64-unknown-linux-gnu/debug/swift-numericsPackageTests.xctest
```

_Originally posted by @finagolfin in https://github.com/apple/swift-package-manager/issues/7377#issuecomment-1975324279_

grynspan:
`--enable-test-discovery` is deprecated and has no effect anymore; it will be removed in a future update. We should move this over to swift-numerics to remove the LinuxMain.swift file. I'm guessing WindowsMain.swift can be updated too.

@MaxDesiatov I don't seem to be able to move the issue over myself (presumably a permissions issue.)

grynspan:
Reopening while I make sure that `--enable-test-discovery` doesn't outright break the build. It'll either end up being a one-line fix or a complete removal of the option (save for a vestigial deprecation warning.)

MaxDesiatov:
With https://github.com/apple/swift-package-manager/pull/7391 under review, I don't think this issue should stay open.

grynspan:
Sure thing. I didn't realize you'd opened that PR! We can forward-dup this issue to that PR then.

grynspan:
Duplicate of #7391

finagolfin:
> --enable-test-discovery is deprecated and has no effect anymore

The very fact disabling the flag in your pull broke running a package's tests shows it had an effect prior to your pull.

[...]

PR #7391, opened 2024-03-04
Remove previously deprecated `--enable-test-discovery`

MaxDesiatov:
This flag was deprecated for 4 years since Swift 5.4, Swift 6.0 is a good opportunity to remove it.

Closes https://github.com/apple/swift-package-manager/issues/7389.

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -464,10 +464,6 @@ package struct BuildOptions: ParsableArguments {
         #endif
     }
 
-    /// Whether to enable test discovery on platforms without Objective-C runtime.
-    @Flag(help: .hidden)
-    package var enableTestDiscovery: Bool = false
-
     /// Path of test entry point file to use, instead of synthesizing one or using `XCTMain.swift` in the package (if
     /// present).
     /// This implies `--enable-test-discovery`
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -466,7 +466,6 @@ package struct BuildOptions: ParsableArguments {
 
     /// Path of test entry point file to use, instead of synthesizing one or using `XCTMain.swift` in the package (if
     /// present).
-    /// This implies `--enable-test-discovery`
     @Option(
         name: .customLong("experimental-test-entry-point-path"),
         help: .hidden
--- a/Sources/CoreCommands/SwiftCommandState.swift
+++ b/Sources/CoreCommands/SwiftCommandState.swift
@@ -371,16 +371,6 @@ package final class SwiftCommandState {
             observabilityScope.emit(.mutuallyExclusiveArgumentsError(arguments: ["--arch", "--triple"]))
         }
 
-        // --enable-test-discovery should never be called on darwin based platforms
-        #if canImport(Darwin)
-        if options.build.enableTestDiscovery {
-            observabilityScope
-                .emit(
-                    warning: "'--enable-test-discovery' option is deprecated; tests are automatically discovered on all platforms"
-                )
-        }
-        #endif
-
         if options.caching.shouldDisableManifestCaching {
             observabilityScope
                 .emit(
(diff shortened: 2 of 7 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/CoreCommands/Options.swift
```diff
@@ -464,10 +464,6 @@ package struct BuildOptions: ParsableArguments {
         #endif
     }
 
-    /// Whether to enable test discovery on platforms without Objective-C runtime.
```
grynspan, line 467: We may want to keep this flag in place and just not use it for anything except generating a diagnostic. At least for a little bit longer, to give affected packages like swift-numerics a chance to migrate.

Sources/CoreCommands/SwiftCommandState.swift
```diff
@@ -371,16 +371,6 @@ package final class SwiftCommandState {
             observabilityScope.emit(.mutuallyExclusiveArgumentsError(arguments: ["--arch", "--triple"]))
         }
 
-        // --enable-test-discovery should never be called on darwin based platforms
-        #if canImport(Darwin)
```
grynspan, line 375: Ditto.

Sources/SPMBuildCore/BuildParameters/BuildParameters+Testing.swift
```diff
@@ -75,9 +72,9 @@ extension BuildParameters {
             switch self {
             case .loadableBundle:
                 try container.encode(DiscriminatorKeys.loadableBundle, forKey: ._case)
-            case .entryPointExecutable(let explicitlyEnabledDiscovery, let explicitlySpecifiedPath):
+            case .entryPointExecutable(let explicitlySpecifiedPath):
                 try container.encode(DiscriminatorKeys.entryPointExecutable, forKey: ._case)
-                try container.encode(explicitlyEnabledDiscovery, forKey: .explicitlyEnabledDiscovery)
+                try container.encode(false, forKey: .explicitlyEnabledDiscovery)
```
grynspan, line 77: A comment here explaining why this remains in place might be helpful for future maintainers.

MaxDesiatov: I think this will be gone in a future commit in this same PR, we don't expect build parameters serialization to be compatible between versions.

grynspan: Okay! I'm fine with that.

stmontgomery (approved):

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@yim-lee @ktoso I see a few repositories you worked on still have `LinuxMain.swift`. Especially if any of those had `fatalError` in them requiring users to pass `--enable-test-discovery`, it's best to delete those files at this point and use plain `swift test` when Swift 6.0 is released.

ktoso:
Sounds good, we only support swift 5.7+ nowadays in most telos so we can drop it I think.

Swift log I think supports ancient versions but we can maybe stop testing on them or something... I'll think about it

yim-lee:
Thanks for the heads-up @MaxDesiatov.

> I see a few repositories you worked on still have `LinuxMain.swift`. 

Is `LinuxMain.swift` ignored in newer Swift versions?

> Especially if any of those had `fatalError` in them requiring users to pass `--enable-test-discovery`

We have repos that pass `--enable-test-discovery` even in newer versions, but I don't think any of them does `fatalError`. Nevertheless, we will adjust the build params as needed once this change gets merged.

@ktoso I guess we should consider bumping up the Swift version requirement if we run into trouble removing `--enable-test-discovery`/`LinuxMain.swift`. At some point it doesn't make much sense to support ancient Swift versions.

yim-lee (approved):

yim-lee:
Got it. Thanks @MaxDesiatov.

MaxDesiatov:
> Don't we want to _stop_ using it?

IDK, if we do we'd have to go through a deprecation cycle with those for at least a few releases first.

PR #7396, opened 2024-03-06
Pass all arguments when initing testing parameters in `swift build`

### LINKED ISSUE #7389: `swift build --build-tests` doesn't work with `--enable-test-discovery` flag

grynspan:
Ensures that a fully-initialized `TestingParameters` structure is used with `swift build --build-tests` instead of using any default parameters (which will tend us toward the wrong output with packages that use `--enable-test-discovery` and still have LinuxMain.swift files.)

Tested with swift-numerics on Ubuntu 22.04 aarch64; before the change, we'd hit the `fatalError()` call in that package's LinuxMain.swift file. After the change, we correctly run XCTest-based tests.

Resolves #7389.

### DIFF
--- a/Fixtures/Miscellaneous/TestDiscovery/IgnoresLinuxMain/Package.swift
+++ b/Fixtures/Miscellaneous/TestDiscovery/IgnoresLinuxMain/Package.swift
@@ -1,0 +1,9 @@
+// swift-tools-version:5.10
+import PackageDescription
+
+let package = Package(
+    name: "IgnoresLinuxMain",
+    targets: [
+        .testTarget(name: "IgnoresLinuxMainTests"),
+    ]
+)
--- a/Fixtures/Miscellaneous/TestDiscovery/IgnoresLinuxMain/Tests/IgnoresLinuxMainTests/SomeTest.swift
+++ b/Fixtures/Miscellaneous/TestDiscovery/IgnoresLinuxMain/Tests/IgnoresLinuxMainTests/SomeTest.swift
@@ -1,0 +1,5 @@
+import XCTest
+
+final class SomeTests: XCTestCase {
+    func testSomething() {}
+}
--- a/Fixtures/Miscellaneous/TestDiscovery/IgnoresLinuxMain/Tests/LinuxMain.swift
+++ b/Fixtures/Miscellaneous/TestDiscovery/IgnoresLinuxMain/Tests/LinuxMain.swift
@@ -1,0 +1,1 @@
+fatalError("Should not use the contents of LinuxMain.swift")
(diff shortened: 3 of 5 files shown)


grynspan:
@swift-ci please test

rauhul:
Can you change the visibility of `--enable-test-discovery` to be `.hidden` so it doesn't show up in help? Would it also be appropriate to emit a warning if this flag is used in 6.0?

grynspan:
@rauhul We should take changes for `--enable-test-discovery` in a separate PR once we decide exactly how we want to proceed with its deprecation/removal. It'd be out of scope for this PR which is primarily concerned with fixing a regression. 🥲

rauhul:
> @rauhul We should take changes for `--enable-test-discovery` in a separate PR once we decide exactly how we want to proceed with its deprecation/removal. It'd be out of scope for this PR which is primarily concerned with fixing a regression. 🥲

gotcha, makes sense

MaxDesiatov:
> Can you change the visibility of `--enable-test-discovery` to be `.hidden` so it doesn't show up in help? Would it also be appropriate to emit a warning if this flag is used in 6.0?

This is already marked as `@Flag(help: .hidden)`, and it was emitting warnings when used since Swift 5.4 when it was deprecated.

bnbarham:
> and it was emitting warnings when used since Swift 5.4 when it was deprecated.

It's not quite that simple. We only emit a warning when the entry point file doesn't exist (which makes sense, since if the flag was removed we would then start using the entry point file). If the entry point file is deprecated I think we need something new for that, or we should make the experimental entry point file flag not experimental and then remove the defaults (but we still need some migration path for that).

PR #7400, opened 2024-03-10
Fix XCTest entrypoint for WASI by making it async

kateinoigakukun:
This change makes the XCTest entrypoint async on WASI, as it is required to not block the main thread. The entrypoint has been changed during the upstreaming work for WebAssembly support in XCTest.

This change should be merged with https://github.com/apple/swift-corelibs-xctest/pull/478


MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

PR #7402, opened 2024-03-13
Replace getdtablesize() on Android

finagolfin:
I've had to make this change on my Android CI since #7369 was merged a couple weeks ago, finagolfin/swift-android-sdk@c5828e4, as Bionic doesn't have this function. I found that [Chromium had switched to this years ago](https://codereview.chromium.org/349323003), can replace `getdtablesize()` altogether instead if wanted.

@MaxDesiatov, let me know what you think.


MaxDesiatov:
@swift-ci test

PR #7403, opened 2024-03-15
Opt in to keep existing config of module contents for SwiftPM targets

elsh:
A .swiftmodule can be configured to only contain exportable decls. Targets in SwiftPM, however, are mostly built from source and contain all contents besides exportable decls unless specified. To continue with the existing configuration, pass a feature flag (AllowNonResilientAccessInPackage) that allows it by default.

rdar://124650423

### DIFF
--- a/Sources/Build/BuildDescription/ProductBuildDescription.swift
+++ b/Sources/Build/BuildDescription/ProductBuildDescription.swift
@@ -278,6 +278,10 @@ package final class ProductBuildDescription: SPMBuildCore.ProductBuildDescriptio
                 args += ["-enable-experimental-feature", feature]
             }
 
+            if self.package.manifest.allowNonResilientAccessInPackage {
+                args += ["-enable-experimental-feature", "AllowNonResilientAccessInPackage"]
+            }
+
             // Embed the swift stdlib library path inside tests and executables on Darwin.
             let useStdlibRpath: Bool
             switch self.product.type {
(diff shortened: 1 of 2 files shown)


elsh:
@swift-ci smoke test

Sources/Build/BuildDescription/ProductBuildDescription.swift
```diff
@@ -278,6 +278,10 @@ package final class ProductBuildDescription: SPMBuildCore.ProductBuildDescriptio
                 args += ["-enable-experimental-feature", feature]
             }
 
+            if self.package.manifest.allowNonResilientAccessInPackage {
+                args += ["-enable-experimental-feature", "AllowNonResilientAccessInPackage"]
```
elsh, line 282: Should we actually enable this by default? cc @nkcsgexi 
@tshortli  How soon do we expect skipping non-exportable decls to be the default config? If we think allowing a faster build time takes a higher priority then maybe this should just be left as opt-in.

elsh: After thinking some more, I'm going to leave this as is, i.e. require explicit opt-in. To get package optimization, a client module in the same package would also have to opt in anyways, so at the time users decide to enable this optimization, they can opt in for both (client and imported module) at the same time.

elsh:
Closing as opt-in seems to be the right behavior (see the comments above).

Sources/Build/BuildDescription/ProductBuildDescription.swift
tshortli: Sorry I forgot to follow up here yesterday; skipping non-exportable decls will eventually become the default for library evolution enabled modules, I hope within about a year. If we do want to enable non-resilient access in package in some configurations, then I think libraries with library evolution enabled should be excluded by default but still allow opt-in. Does that make sense?

PR #7406, opened 2024-03-19
Write output-file-map.json atomically

ahoppen:
There are two subtle changes in behavior here:
- This call will fail if the file system doesn’t support atomic operations. I’m not sure if we need to support file systems that don’t allow atomic operations here.

rdar://124727242

### DIFF
--- a/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
@@ -779,7 +779,8 @@ package final class SwiftTargetBuildDescription {
 
         content += "}\n"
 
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

MaxDesiatov:
> * This call will fail if the file system doesn’t support atomic operations. I’m not sure if we need to support file systems that don’t allow atomic operations here.

Does this apply to Windows? That needs to be verified before we can merge this as we shouldn't regress it.

MaxDesiatov (approved):

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

bnbarham:

PR #7408, opened 2024-03-19
Fix a memory leak in `DownloadTaskManager` and `DataTaskManager`

ahoppen:
`DownloadTaskManager` had a strong reference to `URLSession`, which retained its delegate, which was the `DownloadTaskManager`. This formed a retain cycle.

Make the reference from `URLSession` to the delegate weak and only keep `DownloadTaskManager` alive as long as `DownloadTask`s need it. This causes the `DownloadTaskManager` to be deallocated once nobody holds a reference to it anymore and all the tasks it manages are finished.

Finally, we need to call `invalidate` on the `URLSession` to tell it that we’re done and that it should release its delegate (which is now the `WeakDownloadTaskManager`wrapper).

This retain cycle was causing a leak in sourcekit-lsp. I verified that the leak no longer exists with this patch.

rdar://125012584

### DIFF
--- a/Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
+++ b/Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
@@ -87,7 +87,60 @@ final class URLSessionHTTPClient {
     }
 }
 
-private class DataTaskManager: NSObject, URLSessionDataDelegate {
+/// A weak wrapper around `DataTaskManager` that conforms to `URLSessionDataDelegate`.
+///
+/// This ensures that we don't get a retain cycle between `DataTaskManager.session` -> `URLSession.delegate` -> `DataTaskManager`.
+///
+/// The `DataTaskManager` is being kept alive by a reference from all `DataTask`s that it manages. Once all the
+/// `DataTasks` have finished and are deallocated, `DataTaskManager` will get deinitialized, which invalidates the
+/// session, which then lets go of `WeakDataTaskManager`.
+private class WeakDataTaskManager: NSObject, URLSessionDataDelegate {
+    private weak var dataTaskManager: DataTaskManager?
+
+    init(_ dataTaskManager: DataTaskManager? = nil) {
+        self.dataTaskManager = dataTaskManager
+    }
+
+    func urlSession(
+        _ session: URLSession,
+        dataTask: URLSessionDataTask,
+        didReceive response: URLResponse,
+        completionHandler: @escaping (URLSession.ResponseDisposition) -> Void
+    ) {
+        dataTaskManager?.urlSession(
+            session,
+            dataTask: dataTask,
+            didReceive: response,
+            completionHandler: completionHandler
+        )
+    }
+
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
+        dataTaskManager?.urlSession(session, dataTask: dataTask, didReceive: data)
+    }
+
+    func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
+        dataTaskManager?.urlSession(session, task: task, didCompleteWithError: error)
+    }
+
+    func urlSession(
+        _ session: URLSession,
+        task: URLSessionTask,
+        willPerformHTTPRedirection response: HTTPURLResponse,
+        newRequest request: URLRequest,
+        completionHandler: @escaping (URLRequest?) -> Void
+    ) {
+        dataTaskManager?.urlSession(
+            session,
+            task: task,
+            willPerformHTTPRedirection: response,
+            newRequest: request,
+            completionHandler: completionHandler
+        )
+    }
+}
+
+private class DataTaskManager {
     private var tasks = ThreadSafeKeyValueStore<Int, DataTask>()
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
```diff
@@ -87,7 +87,35 @@ final class URLSessionHTTPClient {
...
+    ) {
+        dataTaskManager?.urlSession(session, dataTask: dataTask, didReceive: response, completionHandler: completionHandler)
+    }
+
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
+        dataTaskManager?.urlSession(session, dataTask: dataTask, didReceive: data)
+    }
+}
+
+private class DataTaskManager: NSObject {
```
MaxDesiatov, line 118: Does this still need to inherit from `NSObject` if it no longer conforms to `URLSessionDataDelegate`?

Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
```diff
@@ -202,18 +240,49 @@ private class DataTaskManager: NSObject, URLSessionDataDelegate {
...
+            session,
+            downloadTask: downloadTask,
+            didWriteData: bytesWritten,
+            totalBytesWritten: totalBytesWritten,
+            totalBytesExpectedToWrite: totalBytesExpectedToWrite
+        )
+    }
+}
+
+private class DownloadTaskManager: NSObject {
```
MaxDesiatov, line 285: Ditto, I hope `NSObject` inheritance is redundant here.

Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
```diff
@@ -223,7 +292,11 @@ private class DownloadTaskManager: NSObject, URLSessionDownloadDelegate {
         self.delegateQueue.name = "org.swift.swiftpm.urlsession-http-client-download-delegate"
         self.delegateQueue.maxConcurrentOperationCount = 1
         super.init()
-        self.session = URLSession(configuration: configuration, delegate: self, delegateQueue: self.delegateQueue)
+        self.session = URLSession(configuration: configuration, delegate: WeakDownloadTaskManager(self), delegateQueue: self.delegateQueue)
+    }
+
+    deinit {
+      session.finishTasksAndInvalidate()
```
MaxDesiatov, line 299: Formatting seems inconsistent here with the rest of the file.

Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
ahoppen: Oh, nice catch 👍🏽

ahoppen:
@swift-ci Please test

MaxDesiatov:
@swift-ci test windows

Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
```diff
@@ -202,18 +254,61 @@ private class DataTaskManager: NSObject, URLSessionDataDelegate {
...
+
+    func urlSession(
+        _ session: URLSession,
+        downloadTask: URLSessionDownloadTask,
+        didFinishDownloadingTo location: URL
+    ) {
+        downloadTaskManager?.urlSession(session, downloadTask: downloadTask, didFinishDownloadingTo: location)
+    }
+
+    public func urlSession(
```
MaxDesiatov, line 302: `public` function in a `private` type

Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
```diff
@@ -87,7 +87,50 @@ final class URLSessionHTTPClient {
...
+        _ session: URLSession,
+        task: URLSessionTask,
+        willPerformHTTPRedirection response: HTTPURLResponse,
+        newRequest request: URLRequest,
+        completionHandler: @escaping (URLRequest?) -> Void
+    ) {
+        dataTaskManager?.urlSession(session, task: task, willPerformHTTPRedirection: response, newRequest: request, completionHandler: completionHandler)
```
MaxDesiatov, line 129: We need to keep line length under 120 columns per the `.swiftformat` config.
```suggestion
        dataTaskManager?.urlSession(session, task: task, willPerformHTTPRedirection: response, newRequest: request, completionHandler: completionHandler)
```

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

MaxDesiatov (approved):
Thanks!

Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
bnbarham: FWIW I see nothing wrong with `public` visibility inside a `private` type, ie. visibility inside the type is describing the intent of the API and then visibility on the type for how accessible you want it to be. It makes moving (and re-using) that type later much simpler. One argument against that is it's quite likely a private type hasn't had much thought to its API, which could be true 🤷.

PR #7414, opened 2024-03-20
[6.0] Suppress redundant linkage warnings

pusukuri:
**Explanation**: Improve user experience by suppressing redundant linkage warnings due to dependency linkage propagation
**Scope**: We should not see warnings such as "ld: warning: ignoring duplicate libraries: '-lsqlite3'"
**Risk**: No risks
**Testing**: Tests continue passing. 
**Issue**: rdar://112671586
**Reviewer**: @MaxDesiatov and @neonichu 
**Co-authored-by**: Max Desiatov <m_desiatov@apple.com>
(cherry picked from commit e73cb5b5efa1204024fa84e101dc57a665e5a3c8)

Original PR https://github.com/apple/swift-package-manager/pull/7404


pusukuri:
@swift-ci test

MaxDesiatov:
This PR is identical to the original one, was it submitted twice by mistake?

MaxDesiatov (changes requested):
If this is supposed to land in 6.0, it needs the base branch set accordingly instead of `main`.

pusukuri:
@MaxDesiatov it is done. Please take a look. Thanks

Tests/BuildTests/BuildSystemDelegateTests.swift
```diff
@@ -25,7 +25,7 @@ final class BuildSystemDelegateTests: XCTestCase {
             try XCTSkipIf(true, "test is only supported on macOS")
             #endif
             let (fullLog, _) = try executeSwiftBuild(fixturePath)
-            XCTAssertTrue(fullLog.contains("ld: warning: ignoring duplicate libraries: '-lz'"), "log didn't contain expected linker diagnostics")
+            XCTAssertTrue(fullLog.contains("ld: warning: search path 'foobar' not found"), "log didn't contain expected linker diagnostics")
```
bnbarham, line 28: I'm confused, isn't this now not testing what it was before? ie. with the change to the fixture, we wouldn't expect the `ignoring duplicate libraries` warning with or without your actual change.

MaxDesiatov: The change is supposed to silence the warning we were previously testing for filtering linker diagnostics. Since that warning is always suppressed, we can no longer test for it and need to trigger a different warning, hence the changes to the fixture.

bnbarham: I apparently got distracted from editing this. I saw the comment in the original PR, but do we have another test to check that duplicate libraries are now suppressed? I assume the argument would be that the command line tests are enough and that the purpose of this test was just to make sure we didn't filter linker diagnostics out? If so, sure 👍

bnbarham (approved):

PR #7416, opened 2024-03-20
[6.0] Fix a memory leak in `DownloadTaskManager` and `DataTaskManager`

ahoppen:
* **Explanation**: `DownloadTaskManager` had a strong reference to `URLSession`, which retained its delegate, which was the `DownloadTaskManager`. This formed a retain cycle.
Make the reference from `URLSession` to the delegate weak and only keep `DownloadTaskManager` alive as long as `DownloadTask`s need it. This causes the `DownloadTaskManager` to be deallocated once nobody holds a reference to it anymore and all the tasks it manages are finished.
Finally, we need to call `invalidate` on the `URLSession` to tell it that we’re done and that it should release its delegate (which is now the `WeakDownloadTaskManager` wrapper).
This retain cycle was causing a leak in sourcekit-lsp. I verified that the leak no longer exists with this patch.
* **Scope**: The change should be transparent to all users of `DataTaskManager` and `DownloadTaskManager` and only prevent the memory leak
* **Risk**: I don’t see a risk with this change since the `{Data|Download}TaskManager` is kept alive as long as there is anything left that might call it.
* **Testing**: I verified that the leak no longer exists with this patch.
* **Issue**: rdar://125012584
* **Reviewer**:  @MaxDesiatov on https://github.com/apple/swift-package-manager/pull/7408

### DIFF
--- a/Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
+++ b/Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
@@ -87,7 +87,60 @@ final class URLSessionHTTPClient {
     }
 }
 
-private class DataTaskManager: NSObject, URLSessionDataDelegate {
+/// A weak wrapper around `DataTaskManager` that conforms to `URLSessionDataDelegate`.
+///
+/// This ensures that we don't get a retain cycle between `DataTaskManager.session` -> `URLSession.delegate` -> `DataTaskManager`.
+///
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

bnbarham (approved):

MaxDesiatov (approved):

ahoppen:
@swift-ci Please test Linux

PR #7417, opened 2024-03-21
Use `.so` as the default dynamic library extension on unknown OSes

kateinoigakukun:
Use `.so` as the default dynamic library extension on unknown OSes

### Motivation:

`fatalError` here can lead to a crash while loading a package manifest containing `.library` products with none-OS targets even though the product won't be used in the build.

### Modifications:

Use `.so` as the default dynamic library extension on unknown OSes
LLVM defaults to `.so` for unknown OSes, so it makes sense to do the same in SwiftPM.

### Result:

Dynamic libraries for Embedded targets will be suffixed with `.so`

### DIFF
--- a/Sources/Basics/Triple+Basics.swift
+++ b/Sources/Basics/Triple+Basics.swift
@@ -147,7 +147,7 @@ extension Triple {
         case .wasi:
             return ".wasm"
         default:
-            fatalError("Cannot create dynamic libraries for os \"\(os)\".")
+            return ".so"
         }
     }
 


rauhul:
Out of curiosity what OS were you using which hit this code path

Sources/Basics/Triple+Basics.swift
```diff
@@ -147,7 +147,7 @@ extension Triple {
         case .wasi:
             return ".wasm"
         default:
-            fatalError("Cannot create dynamic libraries for os \"\(os)\".")
```
MaxDesiatov, line 150: Can we emit a diagnostic message as a warning here instead?

kateinoigakukun: Sounds resonable to me

MaxDesiatov: If it's an unknown OS, an actionable fix is to submit a feature request on SwiftPM to add support for such OS. Otherwise it seems wrong that we're silently using a file extension that can be incorrect without giving users any clues as to what could be wrong.

MaxDesiatov: If you don't want to see this warning for `armv7em-none-none-eabi`, I think we should distinguish here `none` and triples that we just don't know how to handle.

kateinoigakukun: > If it's an unknown OS, an actionable fix is to submit a feature request on SwiftPM to add support for such OS.

Ideally, that's right, but it does not immediately suppress the warning and they need to wait for the next release including the fix.

If we go emitting diagnostics like this, I think we should have an escape hatch for their immediate fix and emit messages recommending them to add support for such OS. Otherwise, such warnings would be too annoying, IMO.

MaxDesiatov: @bnbarham WDYT?

bnbarham: Any thoughts on what a possible escape hatch could be?

kateinoigakukun: I was thinking of providing ways to explicitly configure their expected file extension as an escape hatch via environment variable or configuration file

bnbarham: Full triple instead of platform because we expect it may change between the specified OS?

I'm not sure this is really the right place for the warning anyway though. If a platform is unsupported, then it seems likely we'd fail elsewhere. IMO we should either decide that we should have a warning for that case in general (outside of this particular call) with the escape hatch for known-customizable cases, or not diagnose.

kateinoigakukun: > Full triple instead of platform because we expect it may change between the specified OS?

Exactly. e.g. only `none` OS information does not help decide the extension.

> IMO we should either decide that we should have a warning for that case in general (outside of this particular call) with the escape hatch for known-customizable cases, or not diagnose.

Yeah, totally agree. Adding such escape hatch sounds like the right way to go because such flexibility would be quite beneficial for Embedded scenarios. (BTW Rust toolchain allows developers to have [flexible target specification](https://github.com/rust-lang/rfcs/blob/master/text/0131-target-specification.md) and we can learn from them)

MaxDesiatov: Fair enough, then this can go in as is unless Ben has any objections, and it needs a test.

rauhul: The isDarwin check should be changed to isApple. `<arch>-apple-<os>` should be almost always be a dylib

kubamracek:
IMHO we should merge this

Issue #7398, opened 2024-03-10
Unclear missing product error depends on ordering of package names


MaxDesiatov:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [X] Confirmed reproduction steps with SwiftPM CLI.

### Description

When a product that a target depends on is missing, sometimes a diagnostic message is produced:
```
error: 'repro': product 'ArgumentParser' required by package 'repro' target 'repro' not found.
```

This could look up the name of the product in packages that this package depends on. Turns out, SwiftPM does do that, but only if package identity is sorted in a particular way. When names of packages in the graph are sorted alphabetically, the name of root package needs to precede the name of the package dependency that contains a candidate product for this error to appear:

```
error: 'zzz': product 'ArgumentParser' required by package 'zzz' target 'repro' not found. Did you mean 'ArgumentParser'?
```

In no way this error message is an improvement, as it suggests exactly the same spelling that caused the error in the first place. It should say "Did you mean `.product(name: "ArgumentParser", package: "swift-argument-parser")`" instead.

### Expected behavior

The diagnostic error message is helpful, actionable, and doesn't depend on ordering of names of packages participating in the dependencies graph.

### Actual behavior

The error message is not helpful and also unexpectedly depends on ordering of package names in the graph.

### Steps to reproduce

1. `mkdir repro`
2. `cd repro`
3. `swift package init --type executable`
4. Open `Package.swift` and make sure it has this content:

```swift
// swift-tools-version: 5.10

import PackageDescription

let package = Package(
    name: "repro",
    dependencies: [
        .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.3.0")
    ],
    targets: [
        .executableTarget(
            name: "repro",
            dependencies: ["ArgumentParser"]
        )
    ]
)
```

5. Build with `swift build` and observe this error:
```
error: 'repro': product 'ArgumentParser' required by package 'repro' target 'repro' not found.
```
6. `cd ..`
7. `mv repro zzz`
8. `cd zzz`
9. Run `swift build` and observe a confusing diagnostic message:
```
error: 'zzz': product 'ArgumentParser' required by package 'zzz' target 'repro' not found. Did you mean 'ArgumentParser'?
```

### Swift & OS version (output of `swift --version ; uname -a`)

Xcode 15.3

[...]

PR #7419, opened 2024-03-24
Improve error message about the PackageGraphError.productDependencyNotFound 

k-kohey:
Fixed a bug in the productDependencyNotFound error message

### Motivation:

fix https://github.com/apple/swift-package-manager/issues/7398

The above issue suggests the following two defects.

1. package dependency resolution changes depending on the order of packages (alphabetical order)
2. the phrase "Did you meen..." in the error message in the error message is not on target.

### Modifications:

The first problem is as described in the issue, if users rename the directory containing Package.swift from repro to zzz, the "Did you meen..." will appear. will appear. Essentially, the error minutes should be displayed in full, regardless of the name of the directory. This is due to the fact that the loop used to resolve dependencies depends on the alphabetical order of the directories and the side effect on allTargetName inside the loop. Therefore, the side effect for allTargetName inside the loop has been moved to the outside of the loop.

The second problem is that when dependency A of a package is not found, a strange suggestion "Did you meen `A`? ", which is a strange suggestion. This is because when a dependency is not found, the message "Did you meen `<target name>`" is printed if there is a dependency with a similar name (even the exact same name). Thus, if the names are the same, "Did you meen `.product(name: ... , package: "swift-argugument")"`" instead of the found target name.

### Result:

Command to execute: 
`swift build`

Result of command:.
`error: 'repro': product 'ArgumentParser' required by package 'repro' target 'repro' not found. Did you mean '.product(name: "ArgumentParser", package: "swift-argument-parser")'?`

Condition:
The following steps were taken to create the environment.

1. mkdir repro
1. cd repro
1. swift package init --type executable
1. Open Package.swift and make sure it has this content:

```swift
// swift-tools-version: 5.10

import PackageDescription

let package = Package(
    name: "repro",
    dependencies: [
        .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.3.0")
    ],
    targets: [
        .executableTarget(
            name: "repro",
            dependencies: ["ArgumentParser"]
        )
    ]
)
```

### DIFF
--- a/Sources/PackageGraph/ModulesGraph+Loading.swift
+++ b/Sources/PackageGraph/ModulesGraph+Loading.swift
@@ -525,13 +525,22 @@ private func createResolvedPackages(
                         } else {
                             // Find a product name from the available product dependencies that is most similar to the required product name.
                             let bestMatchedProductName = bestMatch(for: productRef.name, from: Array(allTargetNames))
+                            var packageContainingBestMatchedProduct: String?
+                            if let bestMatchedProductName, productRef.name == bestMatchedProductName {
+                                let dependentPackages = packageBuilder.dependencies.map(\.package)
+                                for p in dependentPackages where p.targets.contains(where: { $0.name == bestMatchedProductName }) {
+                                    packageContainingBestMatchedProduct = p.identity.description
+                                    break
+                                }
+                            }
                             let error = PackageGraphError.productDependencyNotFound(
                                 package: package.identity.description,
                                 targetName: targetBuilder.target.name,
                                 dependencyProductName: productRef.name,
                                 dependencyPackageName: productRef.package,
                                 dependencyProductInDecl: !declProductsAsDependency.isEmpty,
-                                similarProductName: bestMatchedProductName
+                                similarProductName: bestMatchedProductName, 
+                                packageContainingSimilarProduct: packageContainingBestMatchedProduct
                             )
                             packageObservabilityScope.emit(error)
                         }
--- a/Sources/PackageGraph/ModulesGraph+Loading.swift
+++ b/Sources/PackageGraph/ModulesGraph+Loading.swift
@@ -428,7 +428,7 @@ private func createResolvedPackages(
     )
     try dupProductsChecker.run(lookupByProductIDs: moduleAliasingUsed, observabilityScope: observabilityScope)
 
-    // The set of all target names.
+    // The array of all target names.
     let allTargetNames = packageBuilders.reduce([]) { $0 + $1.targets }.map(\.target.name)
 
     // Track if multiple targets are found with the same name.
--- a/Sources/PackageGraph/ModulesGraph+Loading.swift
+++ b/Sources/PackageGraph/ModulesGraph+Loading.swift
@@ -429,7 +429,7 @@ private func createResolvedPackages(
     try dupProductsChecker.run(lookupByProductIDs: moduleAliasingUsed, observabilityScope: observabilityScope)
 
     // The set of all target names.
-    var allTargetNames = Set<String>()
+    let allTargetNames = packageBuilders.reduce([]) { $0 + $1.targets }.map(\.target.name)
 
     // Track if multiple targets are found with the same name.
     var foundDuplicateTarget = false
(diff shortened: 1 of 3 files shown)


Tests/PackageGraphTests/ModulesGraphTests.swift
```diff
@@ -2676,6 +2676,57 @@ final class ModulesGraphTests: XCTestCase {
...
+                            name: "zzz"
+                        )
+                    ])
+            ],
+            observabilityScope: observability.topScope
+        )
+
+        testDiagnostics(observability.diagnostics) { result in
+            result.check(
+                diagnostic: "product 'zzy' required by package 'aaa' target 'aaa' not found. Did you mean 'zzz'?",
```
k-kohey, line 2725: This PR fix will cause the error message to include "Did you mean..."

Sources/PackageGraph/ModulesGraph.swift
```diff
@@ -230,12 +230,14 @@ extension PackageGraphError: CustomStringConvertible {
...
 
-        case .productDependencyNotFound(let package, let targetName, let dependencyProductName, let dependencyPackageName, let dependencyProductInDecl, let similarProductName):
+        case .productDependencyNotFound(let package, let targetName, let dependencyProductName, let dependencyPackageName, let dependencyProductInDecl, let similarProductName, let packageContainingSimilarProduct):
             if dependencyProductInDecl {
                 return "product '\(dependencyProductName)' is declared in the same package '\(package)' and can't be used as a dependency for target '\(targetName)'."
             } else {
                 var description = "product '\(dependencyProductName)' required by package '\(package)' target '\(targetName)' \(dependencyPackageName.map{ "not found in package '\($0)'" } ?? "not found")."
-                if let similarProductName {
+                if let similarProductName, let packageContainingSimilarProduct {
+                    description += " Did you mean '.product(name: \"\(similarProductName)\", package: \"\(packageContainingSimilarProduct)\")'?"
```
k-kohey, line 239: I wrote the following test code to test the output of this message.

```swift
  func test_() throws {
      let fs = InMemoryFileSystem(emptyFiles:
          "/zzz/Sources/zzz/main.swift",
          "/aaa/Sources/aaa/source.swift"
      )

      let g = try loadModulesGraph(
          fileSystem: fs,
          manifests: [
              Manifest.createRootManifest(
                  displayName: "zzz",
                  path: "/zzz",
                  dependencies: [
                      .remoteSourceControl(
                          url: "https://example.com/example.git",
                          requirement: .exact("1.3.0")
                      )
                  ],
                  products: [],
                  targets: [
                      TargetDescription(
                          name: "zzz",
                          dependencies: ["aaa"],
                          type: .executable
                      )
                  ]),
              Manifest.createRemoteSourceControlManifest(
                  displayName: "aaa",
                  url: "https://example.com/example.git",
                  path: "/aaa",
                  version: "1.3.0",
                  products: [
                      ProductDescription(
                          name: "aaa",
                          type: .library(.automatic),
                          targets: ["aaa"]
                      )
                  ],
                  targets: [
                      TargetDescription(name: "aaa")
                  ]
              )
          ],
          observabilityScope: observability.topScope
      )

      testDiagnostics(observability.diagnostics) { result in
          result.check(
              diagnostic: "error: 'zzz': product 'aaa' required by package 'zzz' target 'repro' not found. Did you mean '.product(name: \"aaa\", package: \"aaa\")'?",
              severity: .error
          )
      }
```

However, this test fails with the message "test_(): failed - No diagnostics left to check".
I would like to know how to solve this problem.

Sources/PackageGraph/ModulesGraph+Loading.swift
```diff
@@ -428,8 +428,8 @@ private func createResolvedPackages(
     )
     try dupProductsChecker.run(lookupByProductIDs: moduleAliasingUsed, observabilityScope: observabilityScope)
 
-    // The set of all target names.
-    var allTargetNames = Set<String>()
+    // The array of all target names.
+    let allTargetNames = packageBuilders.reduce([]) { $0 + $1.targets }.map(\.target.name)
 
     // Track if multiple targets are found with the same name.
     var foundDuplicateTarget = false
```
kateinoigakukun, line 435: How about pre-computing `foundDuplicateTarget` here? Then we can avoid `filter {}.count` O(N) operation inside hot loop.

```suggestion
    // The set of all target names.
    var allTargetNames = Set<String>()

    // Track if multiple targets are found with the same name.
    var foundDuplicateTarget = false

    for packageBuilder in packageBuilders {
        for target in packageBuilder.targets {
            // Record if we see a duplicate target.
            foundDuplicateTarget = foundDuplicateTarget || !allTargetNames.insert(target.target.name).inserted
        }
    }
```

kateinoigakukun:
@swift-ci test

kateinoigakukun (approved):

kateinoigakukun:
Thank you for your contribution!

PR #7420, opened 2024-03-25
[chore] Update Package.swift in Examples/package-info

coffmark:
Update Package.swift in Examples/package-info.

### Motivation:

I couldn't build `Examples/package-info` in Venture(v13.6.2).
This is because the minimum supported version of `Examples/package-info` is lower than the minimum supported version of `swift-package-manager`.

### Modifications:

I have corrected the minimum supported version of `Examples/package-info` to be the same as the minimum supported version of `swift-package-manager`. ([`Package.swift`](https://github.com/apple/swift-package-manager/blob/main/Package.swift#L91-L94) in swift-pacakge-manger)

### Result:

`Examples/package-info` can be built.

### DIFF
--- a/Examples/package-info/Package.swift
+++ b/Examples/package-info/Package.swift
@@ -1,18 +1,18 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.9
 
 import PackageDescription
 
 let package = Package(
     name: "package-info",
     platforms: [
-        .macOS(.v12),
-        .iOS(.v13)
+        .macOS(.v13),
+        .iOS(.v16)
     ],
     dependencies: [
         // This just points to the SwiftPM at the root of this repository.
         .package(name: "swift-package-manager", path: "../../"),
         // You will want to depend on a stable semantic version instead:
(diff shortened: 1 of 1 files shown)


coffmark:
@swift-ci test

kateinoigakukun:
@swift-ci test

kateinoigakukun:

MaxDesiatov:
This example package is outdated and was created back when SwiftPM was considering semver for its own package versioning. This is no longer the case, we're not committing to a stable API outside of `PackageDescription` and `PackagePlugin` modules. IMO this example package should be deleted to avoid potential confusion, otherwise it's misleading people to believe there is a stable versioned API.

@bnbarham WDYT?

bnbarham:
> IMO this example package should be deleted to avoid potential confusion, otherwise it's misleading people to believe there is a stable versioned API.

That or changing the comment to specifically point out that the API is unstable. Either works for me.

MaxDesiatov:
Closing this as superseded by https://github.com/apple/swift-package-manager/pull/7434.

PR #7422, opened 2024-03-25
Use `arguments` instead of `argumentList` in the macro template

ahoppen:
`argumentList` has been deprecated. We should use `arguments` instead.


ahoppen:
@swift-ci Please test

MaxDesiatov (approved):

PR #7423, opened 2024-03-25
Suppress warnings for remote Clang targets

xedin:
This mirrors behavior of remote swift targets which suppress warnings as well.

### DIFF
--- a/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
@@ -323,6 +323,15 @@ package final class ClangTargetBuildDescription {
             args += ["-I", toolchainResourcesPath.pathString]
         }
 
+        // suppress warnings if the package is remote
+        if self.package.isRemote {
+            args += ["-w"]
+            // `-w` (suppress warnings) and `-Werror` (warnings as errors) flags are mutually exclusive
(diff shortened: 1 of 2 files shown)


xedin:
@MaxDesiatov what would be the best way to test this? I couldn't find any tests with `-suppress-warnings` flag in the suite :/

xedin:
This is actually a draft, "package" doesn't exist for clang targets yet, I'm about open a PR to introduce that.

xedin:
I think I figured it out - I can extend ClangTargetBuildDescriptionTests to check that by forming a remote package dependency.

MaxDesiatov:
> @MaxDesiatov what would be the best way to test this? I couldn't find any tests with `-suppress-warnings` flag in the suite :/

This looks like something that belongs to `BuildPlanTests` or maybe worth creating a new `ClangTargetBuildPlanTests` as the existing `BuildPlanTests` is too big.

MaxDesiatov:
> maybe worth creating a new `ClangTargetBuildPlanTests` as the existing `BuildPlanTests` is too big.

Turns out we already have `ClangTargetBuildDescriptionTests.swift`.

xedin:
@swift-ci please test

xedin:

MaxDesiatov:
@swift-ci test windows

PR #7424, opened 2024-03-25
Use `arguments` instead of `argumentList` in the macro template

ahoppen:
* **Explanation**: `argumentList` has been deprecated. The Swift macro template should use `arguments` instead.
* **Scope**: The Swift macro template
* **Risk**: Very low
* **Testing**: Made the same modification in a macro package that was freshly generated from the template and verified that it builds without warnings
* **Issue**: n/a
* **Reviewer**: @MaxDesiatov on https://github.com/apple/swift-package-manager/pull/7422


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #7425, opened 2024-03-25
Add a `swift-corelibs-foundation` specific workaround

xedin:
CoreFoundation depends on dispatch, this dependency is implicit
and dispatch itself is bundled with swift toolchains. Let's add
a package specific workaround for Linux/Android targets to add
toolchain resources directory to search paths of clang build targets.

This is also a prerequisite for https://github.com/apple/swift-package-manager/pull/7423

### DIFF
--- a/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
@@ -311,6 +311,18 @@ package final class ClangTargetBuildDescription {
             args += ["-I", includeSearchPath.pathString]
         }
 
+        // FIXME: Remove this once it becomes possible to express this dependency in a package manifest.
+        //
+        // On Linux/Android swift-corelibs-foundation depends on dispatch library which is
+        // currently shipped with the Swift toolchain.
+        if (triple.isLinux() || triple.isAndroid()) && self.package.id == .plain("swift-corelibs-foundation") {
+            let swiftCompilerPath = self.buildParameters.toolchain.swiftCompilerPath
+            let toolchainResourcesPath = swiftCompilerPath.parentDirectory
+                                                          .parentDirectory
+                                                          .appending(components: ["lib", "swift"])
+            args += ["-I", toolchainResourcesPath.pathString]
+        }
+
         return args
     }
 
--- a/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import Basics
+import PackageGraph
 import PackageLoading
 import PackageModel
 import struct PackageGraph.ModulesGraph
@@ -23,6 +24,9 @@ import enum TSCBasic.ProcessEnv
 
 /// Target description for a Clang target i.e. C language family target.
 package final class ClangTargetBuildDescription {
+    /// The package this target belongs to.
+    package let package: ResolvedPackage
+
     /// The target described by this target.
     package let target: ResolvedTarget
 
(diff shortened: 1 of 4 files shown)


Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
```diff
@@ -305,6 +311,16 @@ package final class ClangTargetBuildDescription {
             args += ["-I", includeSearchPath.pathString]
         }
 
+        // FIXME: Remove this once it becomes possible to express this dependency in a package manifest.
+        //
+        // On Linux/Android swift-corelibs-foundation depends on dispatch library which is
+        // currently shipped with the Swift toolchain.
+        if (triple.isLinux() || triple.isAndroid()) && self.package.id == .plain("swift-corelibs-foundation") {
```
MaxDesiatov, line 318: I don't remember off the top of my head if we have `isDarwin()` on triples, but I'd expect it tool like this:
```suggestion
        if !triple.isDarwin() && self.package.id == .plain("swift-corelibs-foundation") {
```

xedin: I re-created the check exactly how it’s in the package file which covers only Linux and Android.

Tests/BuildTests/ClangTargetBuildDescriptionTests.swift
```diff
@@ -49,8 +49,33 @@ final class ClangTargetBuildDescriptionTests: XCTestCase {
...
-            target: try makeResolvedTarget(),
+            package: .init(underlying: package,
+                           defaultLocalization: nil,
+                           supportedPlatforms: [],
+                           dependencies: [],
+                           targets: [target],
+                           products: [],
+                           registryMetadata: nil,
+                           platformVersionProvider: .init(implementation: .minimumDeploymentTargetDefault)),
+            target: target,
```
MaxDesiatov, line 78: Can a new test or something on top of existing tests be added to verify that `-I` is actually passed when targeting Linux or Android?

xedin: It might be tricky because I don’t know if we have `toolchain` setup in tests, but I will give it a shot!

MaxDesiatov: I don't think we need a full toolchain for build plan tests, passing build parameters is usually sufficient, and there's a `mockBuildParameters` helper function for that.

Tests/BuildTests/ClangTargetBuildDescriptionTests.swift
```diff
@@ -14,15 +14,40 @@ import Basics
...
 final class ClangTargetBuildDescriptionTests: XCTestCase {
     func testClangIndexStorePath() throws {
-        let targetDescription = try makeTargetBuildDescription()
+        let targetDescription = try makeTargetBuildDescription("test")
         XCTAssertTrue(try targetDescription.basicArguments().contains("-index-store-path"))
     }
 
+    func testSwiftCorelibsFoundationIncludeWorkaround() throws {
+        let macosParameters = mockBuildParameters(
+            toolchain: try UserToolchain.default,
```
MaxDesiatov, line 29: Do we need to use the real toolchain here? I wonder if the default `MockToolchain()` argument of `mockBuildParameters` would be sufficient?

MaxDesiatov (approved):
A few nits WRT to use of `UserToolchain.default` in these tests, but otherwise LGTM. Thanks!

Tests/BuildTests/ClangTargetBuildDescriptionTests.swift
xedin: Fixed!

xedin:
@swift-ci please test

xedin:

PR #7426, opened 2024-03-25
On Darwin, allow XCTest to be missing if we're only building swift-testing tests.

grynspan:
This PR removes the constraint on Darwin that XCTest.framework must be present in order to build tests using swift-testing. On Darwin, XCTest is included as a framework inside Xcode, but if a developer installs the Xcode Command Line Tools instead of the full IDE, XCTest is not included. They then get a diagnostic of the form:

> error: XCTest not available: terminated(1): /usr/bin/xcrun --sdk macosx --show-sdk-platform-path output:
>    xcrun: error: unable to lookup item 'PlatformPath' from command line tools installation
>    xcrun: error: unable to lookup item 'PlatformPath' in SDK '/Library/Developer/CommandLineTools/SDKs/MacOSX14.0.sdk'

Which is a poor experience if they aren't even using XCTest.

This change, as a (positive) side effect, suppresses the same diagnostic when running commands that are not usually dependent on the presence of XCTest such as `swift build`.

Note that swift-corelibs-xctest is not supported on Darwin, so installing the Xcode Command Line Tools and adding an explicit dependency on swift-corelibs-xctest will not produce a functional test target bundle. Supporting swift-corelibs-xctest on Darwin is a potential future direction.

Automated testing for this change is difficult because it relies on a build environment that is not supported in CI (namely the presence of the CL tools but not Xcode nor XCTest.framework.) I have manually tested the change against swift-testing's own test target.

A separate PR will be necessary in swift-testing to remove some remaining XCTest dependencies. Those changes are not covered by this PR.

Resolves rdar://125372431.

### DIFF
--- a/Sources/Commands/SwiftTestCommand.swift
+++ b/Sources/Commands/SwiftTestCommand.swift
@@ -40,7 +40,8 @@ private enum TestError: Swift.Error {
     case testProductNotFound(productName: String)
     case productIsNotTest(productName: String)
     case multipleTestProducts([String])
-    case xctestNotAvailable(reason: String? = nil)
+    case xctestNotAvailable(reason: String)
+    case xcodeNotInstalled
 }
 
 extension TestError: CustomStringConvertible {
@@ -58,11 +59,9 @@ extension TestError: CustomStringConvertible {
         case .multipleTestProducts(let products):
             return "found multiple test products: \(products.joined(separator: ", ")); use --test-product to select one"
         case let .xctestNotAvailable(reason):
-            if let reason {
-                return "XCTest not available: \(reason)"
-            } else {
-                return "XCTest not available"
-            }
+            return "XCTest not available: \(reason)"
+        case .xcodeNotInstalled:
+            return "XCTest not available; download and install Xcode to use XCTest on this platform"
         }
     }
 }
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci please test

briancroom (approved):

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -814,7 +819,7 @@ final class TestRunner {
         #if os(macOS)
         if library == .xctest {
             guard let xctestPath = self.toolchain.xctestPath else {
-                throw TestError.xctestNotAvailable
+                throw TestError.xctestNotAvailable()
```
MaxDesiatov, line 822: Why not provide a reason here for users to have an actionable error message?

Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
```diff
@@ -147,7 +147,29 @@ public struct SwiftSDK: Equatable {
...
+    @available(*, deprecated, message: "Use `xctestSupport` instead")
+    public var supportsTesting: Bool {
+        if case .supported = xctestSupport {
+            return true
+        }
+        return false
+    }
+
+    /// Whether or not the receiver supports testing using XCTest.
+    public enum XCTestSupport: Sendable, Equatable {
```
MaxDesiatov, line 159:
```suggestion
    package enum XCTestSupport: Sendable, Equatable {
```

Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
```diff
@@ -417,19 +439,44 @@ public struct SwiftSDK: Equatable {
...
+            hostTriple: hostTriple,
+            targetTriple: targetTriple,
+            toolset: toolset,
+            pathsConfiguration: pathsConfiguration,
+            xctestSupport: xctestSupport
+        )
+    }
+
     /// Creates a Swift SDK with the specified properties.
     public init(
```
MaxDesiatov, line 468:
```suggestion
    package init(
```

MaxDesiatov: Let's keep the new initializer hidden for now. We need to understand the exact use cases for exposing `SwiftSDK` as `public` in the first place before encouraging further adoption of it, especially as its API is not stable.

Sources/Commands/SwiftTestCommand.swift
grynspan: Done.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

MaxDesiatov (approved):
Thanks!

PR #7429, opened 2024-03-28
🍒[cxx-interop] Propagate interop flag to the test entry point target

egorzhdan:
**Explanation**: Projects that use C++ interop for some of the Swift targets were failing to build in test mode (`swift test`) because the generated test entry point target wasn't being compiled with C++ interop enabled.
**Scope**: Flags that are being passed to the entry point test target.
**Risk**: Low, this only affects projects that explicitly enable C++ interop in the package manifest on non-Darwin platforms only.
**Testing**: Added a build plan test.
**Issue**: rdar://125498997
**Reviewer**: @MaxDesiatov approved https://github.com/apple/swift-package-manager/pull/7428.

(cherry picked from commit f529daef40418e143ed501ff245109e061db190a)

### DIFF
--- a/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
@@ -555,11 +555,11 @@ package final class SwiftTargetBuildDescription {
             args += ["-color-diagnostics"]
         }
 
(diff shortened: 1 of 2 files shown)


egorzhdan:
@swift-ci please test

bnbarham (approved):

PR #7430, opened 2024-03-28
[6.0] On Darwin, allow XCTest to be missing if we're only building swift-testing tests.

grynspan:
**Explanation:** This PR removes the constraint on Darwin that XCTest.framework must be present in order to build tests using swift-testing. On Darwin, XCTest is included as a framework inside Xcode, but if a developer installs the Xcode Command Line Tools instead of the full IDE, XCTest is not included.

**Scope:** Improves swift-testing functionality by allowing developers to build test targets that use swift-testing without an implicit XCTest dependency. No source- or language-breaking changes.

**Issue:** N/A

**Original PR:** https://github.com/apple/swift-package-manager/pull/7426

**Risk:** Low; only impacts developers using Xcode CLTools on Darwin.

**Testing:** At-desk testing as well as CI runs. It is not possible to test the environment where the issue manifests in our CI because it would need to be configured with CLTools but no IDE.

**Reviewer:** @briancroom @MaxDesiatov

### DIFF
--- a/Sources/Commands/SwiftTestCommand.swift
+++ b/Sources/Commands/SwiftTestCommand.swift
@@ -40,7 +40,8 @@ private enum TestError: Swift.Error {
     case testProductNotFound(productName: String)
     case productIsNotTest(productName: String)
     case multipleTestProducts([String])
-    case xctestNotAvailable
+    case xctestNotAvailable(reason: String)
+    case xcodeNotInstalled
 }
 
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci please test

bnbarham (approved):

grynspan:
@swift-ci please test Windows

briancroom (approved):

PR #7431, opened 2024-03-28
Allow `swift-corelibs-foundation` to use unsafeFlags when used as a d…

xedin:
…ependency

Tagged dependencies don't currently allow use of `unsafeFlags` but we need to temporarily
lift that requirement for `swift-corelibs-foundation` that depends on C flags. In the future this
would be removed in favor of a more general mechanism for "safe" flags.

### DIFF
--- a/Sources/Workspace/Workspace+Manifests.swift
+++ b/Sources/Workspace/Workspace+Manifests.swift
@@ -141,9 +141,16 @@ extension Workspace {
                 let dependency = dependency.dependency
                 switch dependency.state {
                 case .sourceControlCheckout(let checkout):
-                    if checkout.isBranchOrRevisionBased {
-                        result.insert(dependency.packageRef)
+                    let packageRef = dependency.packageRef
+
+                    if checkout.isBranchOrRevisionBased
+                      // FIXME: Remove this once we have a general mechanism
+                      //        for passing "safe" flags.
+                      || packageRef.identity == .plain("swift-corelibs-foundation")
+                    {
+                      result.insert(packageRef)
                     }
+
                 case .registryDownload, .edited, .custom:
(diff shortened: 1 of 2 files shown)


xedin:
@swift-ci please test

Sources/Workspace/Workspace+Manifests.swift
```diff
@@ -141,9 +141,15 @@ extension Workspace {
...
                 switch dependency.state {
                 case .sourceControlCheckout(let checkout):
-                    if checkout.isBranchOrRevisionBased {
-                        result.insert(dependency.packageRef)
+                    let packageRef = dependency.packageRef
+
+                    if checkout.isBranchOrRevisionBased ||
+                       // FIXME: Remove this once we have a general mechanism
+                       //        for passing "safe" flags.
+                       packageRef.identity == .plain("swift-corelibs-foundation") {
```
MaxDesiatov, line 149: Formatting looks off for this snippet, unless GitHub is mangling the diff? We stick to 4 spaces, you can also run [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) (`brew install swiftformat`) on this file to automate that.

MaxDesiatov (approved):
LGTM, just a formatting nit

Sources/Workspace/Workspace+Manifests.swift
xedin: I keep forgetting to ask about this, thank you!

xedin:
@swift-ci please test

Sources/Workspace/Workspace+Manifests.swift
xedin: ✅

PR #7432, opened 2024-03-29
Make `supportedTriples` optional in artifactbundle metadata

kateinoigakukun:
Make `supportedTriples` optional in artifactbundle metadata

### Motivation:

We don't have a good way to express a Swift SDK that can be used on any host platform except for listing up all possible platforms.

### Modifications:

This change allows `supportedTriples` field to be null or missing to express that the SDK is universally usable.
Also this fixes a minor issue around `swift experimental-sdk configuration`, that did not take care of runtime compatibility of OS version in host triples.

### Result:

SwiftPM now accepts host-platform independent Swift SDK


kateinoigakukun:
@swift-ci test

kateinoigakukun:

PR #7434, opened 2024-04-02
Delete `Examples/package-info` directory as outdated

MaxDesiatov:
This example package was created many years ago during an initial attempt of versioning SwiftPM with semver. That approach has been abandoned since then, and SwiftPM doesn't provide any stable API outside of `PackageDescription` and `PackagePlugin` modules.

This example code can mislead users of SwiftPM to believe that it can be consumed as a library with a stable API in other modules, which is currently not the case.


MaxDesiatov:
@swift-ci test

bnbarham (approved):

PR #7441, opened 2024-04-08
Fix sendability warnings in `URLSessionHTTPClient`

MaxDesiatov:
Converted `DownloadTask` and `DataTask` to value types and added more `Sendable` annotations where needed.

### DIFF
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -12,8 +12,6 @@
 
 import Basics
 
-import Build
-
 import LLBuildManifest
 import PackageGraph
 import PackageLoading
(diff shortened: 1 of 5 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

Sources/Build/BuildOperation.swift
```diff
@@ -12,8 +12,6 @@
 
 import Basics
 
-import Build
```
MaxDesiatov, line 15: Unrelated warning, but probably not worth a separate one-liner PR, so chucking that in here instead.

MaxDesiatov:
@swift-ci test windows

PR #7446, opened 2024-04-09
NFC: Fix `@Sendable @escaping` order, symbol name typo

MaxDesiatov:
Sticking to `@escaping @Sendable` as a preferred order of attributes.

Also fixed `dependecies` -> `dependencies` typo.


MaxDesiatov:
@swift-ci test

xedin (approved):
Thanks!

PR #7448, opened 2024-04-10
Add VisionOS platform support to `PIF.PlatformFilter`

Ryu0118:
#### Motivation:
In the process of generating PIF, I identified a missing case for VisionOS in the implementation of the `[PackageConditionProtocol].toPlatformFilters()`, which led to assertionFailure when Package's platforms included VisionOS. 

#### Modifications:
- Added a new case `.visionOS` in the switch statement within the `toPlatformFilters()` extension for `[PackageCondition]`. 
- Added a new static property `visionOSFilters` within `PIF.PlatformFilter` to define filters specific to VisionOS.

#### Result:
With these changes, PIFBuilder can now generate PIF without errors for packages that include VisionOS as a target

### DIFF
--- a/Sources/XCBuildSupport/PIFBuilder.swift
+++ b/Sources/XCBuildSupport/PIFBuilder.swift
@@ -1601,6 +1601,9 @@ extension [PackageCondition] {
             case .openbsd:
                 result += PIF.PlatformFilter.openBSDFilters
 
+            case .visionOS:
+                result += PIF.PlatformFilter.visionOSFilters
+
             default:
                 assertionFailure("Unhandled platform condition: \(condition)")
                 break
@@ -1670,6 +1673,12 @@ extension PIF.PlatformFilter {
     /// WebAssembly platform filters.
     public static let webAssemblyFilters: [PIF.PlatformFilter] = [
         .init(platform: "wasi"),
+    ]
+
+    /// VisionOS platform filters.
+    public static let visionOSFilters: [PIF.PlatformFilter] = [
+        .init(platform: "visionos"),
+        .init(platform: "visionos", environment: "simulator")
     ]
 }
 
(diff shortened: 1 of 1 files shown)


Sources/XCBuildSupport/PIFBuilder.swift
```diff
@@ -1601,6 +1601,9 @@ extension [PackageCondition] {
             case .openbsd:
                 result += PIF.PlatformFilter.openBSDFilters
 
+            case .visionOS:
```
jakepetroules, line 1604: Can you move this case up below watchOS?

Sources/XCBuildSupport/PIFBuilder.swift
```diff
@@ -1671,6 +1674,12 @@ extension PIF.PlatformFilter {
     public static let webAssemblyFilters: [PIF.PlatformFilter] = [
         .init(platform: "wasi"),
     ]
+
+    /// VisionOS platform filters.
+    public static let visionOSFilters: [PIF.PlatformFilter] = [
+        .init(platform: "visionos"),
```
jakepetroules, line 1680: This needs to add "xros" or Xcode won't pick up the filters correctly. For example:

```swift
public static let visionOSFilters: [PIF.PlatformFilter] = [
    .init(platform: "xros"),
    .init(platform: "xros", environment: "simulator"),
    .init(platform: "visionos"),
    .init(platform: "visionos", environment: "simulator")
]
```

jakepetroules (changes requested):
Thank you for your contribution! I left comments requesting a couple of small changes that need to be addressed.

MaxDesiatov:
@swift-ci test

jakepetroules (approved):
Looks great, thanks!

Issue #7453, opened 2024-04-11
package-registry login has empty token


pwallrich:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [X] Confirmed reproduction steps with SwiftPM CLI.

### Description

When executing `swift package-registry login` it asks for a token but fails to retrieve it properly.

 Screenshot from mitm proxy of the request:
<img width="571" alt="Screenshot 2024-04-11 at 21 21 26" src="https://github.com/apple/swift-package-manager/assets/13999931/429c978f-51ce-42ae-8aa6-3b6b7707bba4">

I've checked the code a bit and it only happens on release builds. When spm is built with debug config it doesn't happen.

The issue seems to be related to `readpassphrase` inside PackageRegistryCommand+Auth.swift. Somehow the buffer and passwordPtr aren't valid anymore, when the string is created.

I also already fixed it locally and will submit a PR soon (not sure, if it's the best solution).

### Expected behavior

It should get the token from the cli and use it for authentication

<img width="563" alt="Screenshot 2024-04-11 at 21 26 59" src="https://github.com/apple/swift-package-manager/assets/13999931/887e2134-21e8-490f-861b-016337259f23">

### Actual behavior

It uses an empty string instead of the password. 

I assume because of some memory management issues related to `readpassphrase`

### Steps to reproduce

1. Build release version of SPM 5.10 or latest version from main
2. swift package-registry login {url}
3. authentication fails

### Swift Package Manager version/commit hash

SPM 5.10 and latest from main

### Swift & OS version (output of `swift --version ; uname -a`)

swift-driver version: 1.90.11.1 Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4)
Target: arm64-apple-macosx14.0

BlaiseMuhirwa:
Was this fixed? I happen to encounter the same problem now.

PR #7454, opened 2024-04-11
Fix empty token on swift package-registry login (Issue #7453)

pwallrich:
Fix empty token when adding a new package-registry and adding token interactively on release builds

### Motivation:

When executing swift package-registry login it asks for a token but fails to retrieve it properly. It sends an empty string to the registry server and login therefore fails. (see https://github.com/apple/swift-package-manager/issues/7453)

### Modifications:

I've debugged the code a bit and found out, that somehow `buffer` and `passwordPtr` don't seem to be holding correct values after the `readpassphrase` call.

- An easy quick fix is to make sure that `buffer` doesn't get deallocated by adding smth like `_ = buffer` after the String init. This works but is not nice.
- My first try was to use `buffer` instead of  `passwordPtr`  to create the string. These works, but I remember reading somewhere that `&` can be quite nasty sometimes.
- I also tried `buffer.withUnsafeMutablePointer`. However `readpassphrase` seems to not only change the content of `buffer` but move it. This leads to a runtime failure `Fatal error: Array withUnsafeMutableBufferPointer: replacing the buffer is not allowed` 

### Result:
The buffer is retained, the token is properly parsed and sent to the server.

### DIFF
--- a/Sources/PackageRegistryCommand/PackageRegistryCommand+Auth.swift
+++ b/Sources/PackageRegistryCommand/PackageRegistryCommand+Auth.swift
@@ -71,12 +71,13 @@ private func readpassword(_ prompt: String) throws -> String {
 
     #if canImport(Darwin)
     var buffer = [CChar](repeating: 0, count: PackageRegistryCommand.Login.passwordBufferSize)
+    password = try withExtendedLifetime(buffer) {
+        guard let passwordPtr = readpassphrase(prompt, &buffer, buffer.count, 0) else {
+            throw StringError("unable to read input")
+        }
 
-    guard let passwordPtr = readpassphrase(prompt, &buffer, buffer.count, 0) else {
-        throw StringError("unable to read input")
+        return String(cString: passwordPtr)
     }
-
-    password = String(cString: passwordPtr)
     #else
     // GNU C implementation of getpass has no limit on the password length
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
Hi @pwallrich, thank you for the contribution? Would you be able to add a test for this to our test suite to prevent possible regressions in the future?

pwallrich:
Hey @MaxDesiatov. Could be quite hard, since it only happens in release builds. But I'll try to find a way to trigger it reliably in tests. 

Do you know if there already exist tests for things like that so I wouldn't have to reinvent the wheel?

pwallrich:
Ok I've managed to write a test for it, but it's actually a bit nasty since I have to change the stdin.
Did that based on https://forums.swift.org/t/how-to-read-from-standard-input-during-unit-tests/43237/.

I also had to create a separate testTarget since the test needs to be built with optimisation and therefore can't use @testable. I don't know, how I would add to the CI.

The more I think about this bug the more I question whether reverting the change introducing `readpassphrase` would make sense. So use `getpass` again and add an additional error for users who use tokens >128 characters telling them to either use `--token` or `--token-file`. But I guess that would need a bigger discussion.

Do you have any thoughts about that?

yim-lee (approved):

MaxDesiatov:
@swift-ci test

PR #7459, opened 2024-04-15
Rename `ResolvedTarget` to `ResolvedModule`

MaxDesiatov:
### Motivation:

With host/target triples separation in the SwiftPM codebase, it gets very confusing whether at a given moment "target" refers to a module, a triple, or a low level build system target.

### Modifications:

Renamed `ResolvedTarget` to `ResolvedModule`. Added a deprecated `typealias ResolvedTarget = ResolvedModule` to allow graceful migration for users of this type.

### Result:

Confusion between target triples and package targets is reduced.

This has no impact on how these concepts are named in user-visible APIs like `PackageDescription` and `PackagePlugin`, target there can stay as "target" for as long as needed.

### DIFF
--- a/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
@@ -15,7 +15,7 @@ import PackageGraph
 import PackageLoading
 import PackageModel
 import struct PackageGraph.ModulesGraph
-import struct PackageGraph.ResolvedTarget
+import struct PackageGraph.ResolvedModule
 import struct SPMBuildCore.BuildParameters
 import struct SPMBuildCore.BuildToolPluginInvocationResult
 import struct SPMBuildCore.PrebuildCommandResult
@@ -28,7 +28,7 @@ package final class ClangTargetBuildDescription {
     package let package: ResolvedPackage
 
     /// The target described by this target.
-    package let target: ResolvedTarget
+    package let target: ResolvedModule
 
     /// The underlying clang target.
     package let clangTarget: ClangTarget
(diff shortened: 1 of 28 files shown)


MaxDesiatov:
@swift-ci test

Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
```diff
@@ -28,7 +28,7 @@ package final class ClangTargetBuildDescription {
     package let package: ResolvedPackage
 
     /// The target described by this target.
-    package let target: ResolvedTarget
+    package let target: ResolvedModule
```
xedin, line 31: Nit: looks like the comment needs updating as well :)

Sources/Build/BuildDescription/ProductBuildDescription.swift
```diff
@@ -49,7 +49,7 @@ package final class ProductBuildDescription: SPMBuildCore.ProductBuildDescriptio
     var additionalFlags: [String] = []
 
     /// The list of targets that are going to be linked statically in this product.
-    var staticTargets: [ResolvedTarget] = []
+    var staticTargets: [ResolvedModule] = []
```
xedin, line 52: Would this be updated in a follow-up?

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
```diff
@@ -33,7 +33,7 @@ package final class SwiftTargetBuildDescription {
     package let package: ResolvedPackage
 
     /// The target described by this target.
-    package let target: ResolvedTarget
+    package let target: ResolvedModule
```
xedin, line 36: Nit: Same comment issue here.

xedin (approved):
LGTM but I think it would be as important to follow-up with a PR that changes all the labels related to ResolvedModule from `target` to `module` as well as variables and other places where the type is used.

PR #7460, opened 2024-04-15
Add `UndirectedGraph` and `DirectedGraph` data structures

MaxDesiatov:
`DirectedGraph` can be used to represent the existing modules graph in the `PackageGraph` module hopefully with better performance due to reduced number of allocations. `UndirectedGraph` could represent a new linkage graph concept, which would help with breaking dependency cycles for certain packages with products that don't have a linkage cycle.

### DIFF
--- a/Sources/Basics/Graph/AdjacencyMatrix.swift
+++ b/Sources/Basics/Graph/AdjacencyMatrix.swift
@@ -1,0 +1,60 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// A matrix storing bits of `true`/`false` state for a given combination of row and column indices. Used as
+/// a square matrix indicating edges in graphs, where rows and columns are indices in a storage of graph's nodes.
+///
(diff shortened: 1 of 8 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/Basics/Graph/AdjacencyMatrix.swift
```diff
@@ -0,0 +1,61 @@
...
+
+/// A matrix storing bits of `true`/`false` state for a given combination of row and column indices. Used as
+/// a square matrix indicating edges in graphs, where rows and columns are indices in a storage of graph's nodes.
+///
+/// For example, in a graph that contains 3 nodes `matrix[row: 1, column: 2]` evaluating to `true` means an edge
+/// between nodes with indices `1` and `2` exists. `matrix[row: 1, column: 2]` evaluating to `false` means that no
+/// edge exists.
+///
+/// See https://en.wikipedia.org/wiki/Adjacency_matrix for more details.
+struct AdjacencyMatrix {
```
bnbarham, line 21: Why the difference in adjacency list vs matrix between the directed and undirected graphs?

MaxDesiatov: I started with an adjacency list in a directed graph implementation, but then when proceeded to an undirected graph adjacency matrix seemed obviously more performant for directed graphs when adding edges: just flipping two bits in the matrix instead of potential allocations with the list.

I'll add benchmarks to get a confirmation for that. I'm unsure though if either is a clear win in all cases, adjacency matrix may be fast at adding edges, but not the fastest for different kinds of graphs analysis and lookups. Keeping both implementations for reference seemed like a sensible idea given that it's an NFC PR anyway.

bnbarham (approved):

Issue #7409, opened 2024-03-19
SwiftPM fails to resolve system libraries that are present in the target SDK (`--experimental-swift-sdk`)


t089:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [X] Confirmed reproduction steps with SwiftPM CLI.

### Description

Currently, when cross-compiling packages with system dependencies, SwiftPM will parse `.pc` files from the SDK but then take those paths _as-is_ without considering the path of the SDK itself.

For example, `glib-2.0` on Ubuntu Jammy defines the following `.pc` file:

```
prefix=/usr
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

bindir=${prefix}/bin
glib_genmarshal=${bindir}/glib-genmarshal
gobject_query=${bindir}/gobject-query
glib_mkenums=${bindir}/glib-mkenums

Name: GLib
Description: C Utility Library
Version: 2.72.4
Requires.private: libpcre >=  8.31
Libs: -L${libdir} -lglib-2.0
Libs.private: -pthread -lm
Cflags: -I${includedir}/glib-2.0 -I${libdir}/glib-2.0/include
```

Building a SwiftPM package using a Swift SDK that contains this file will result in flags like `-I /usr/lib/x86_64-linux-gnu/glib-2.0`. This path is not correct, the expected behaviour is to use a path inside the SDK root:

```-I $SDK_ROOT/usr/lib/x86_64-linux-gnu/glib-2.0```

One possible solution could be to mirror the `pkg-config` behaviour and respect the `PKG_CONFIG_SYSROOT_DIR` env variable. When set, this variable should be prepended to all resolved paths from the `pc` file.

But maybe, by default, the SDK sys root should be prepended when using a Swift SDK.

### Expected behavior

Building Swift packages that rely on system libraries present in a Swift SDK should be possible. SwiftPM should pass flags that reference the files inside the SDK.

### Actual behavior

SwiftPM passes flags that reference files in the host filesystem and not inside the SDK and thus the build fails or is incorrect.

### Steps to reproduce

1. Clone the example repo: [t089/swift-glib-test](https://github.com/t089/swift-glib-test)
2. Generate a Swift SDK based on ubuntu-jammy with `libglib2.0-dev` installed (see this issue on Swift SDK Generator: https://github.com/apple/swift-sdk-generator/issues/89)
3. Try to build the example package using the generated SDK

[...]

PR #7461, opened 2024-04-16
pkgconfig: Apply PKG_CONFIG_SYSROOTDIR when generating paths

euanh:
pkgconfig: Apply PKG_CONFIG_SYSROOTDIR when generating paths

### Motivation:

SwiftPM's pkg-config implementation sets the 'pc_sysrootdir' variable,
but most .pc files do not use this variable directly.  Instead, they rely on
the pkg-config tool rewriting the generated paths to include the sysroot
prefix when necessary.   SwiftPM does not do this, so it does not generate
the correct compiler flags to use libraries from a sysroot.

https://github.com/apple/swift-package-manager/issues/7409

### Modifications:

There are two major pkg-config implementations which handle sysroot
differently:

  *  `pkg-config` (the original freedesktop.org implementation) prepends
     sysroot after variable expansion, when in creates the compiler flag
     lists

  *  `pkgconf` (a newer implementation) prepends sysroot to variables
     when they are defined, so sysroot is included when they are
     expanded

pkg-config's method skips single character compiler flags, such as '-I'
and '-L', and has special cases for longer options. It does not handle
spaces between the flags and their values properly, and prepends sysroot
multiple times in some cases, such as when the .pc file uses the
sysroot_dir variable directly or has been rewritten to hard-code the
sysroot prefix.

pkgconf's method handles spaces correctly, although it also makes
extra checks to ensure that sysroot is not applied more than once.

In 2024 pkg-config is the more popular option according to Homebrew
installation statistics, but the major Linux distributions
 have generally switched to pkgconf.

We will use pkgconf's method here as it seems more robust than
pkg-config's, and pkgconf's greater popularity on Linux means libraries
developed there may depend on the specific way it handles .pc files.

### Result:

SwiftPM will now apply the sysroot prefix to compiler flags, such as 
include (`-I`) and library (`-L`) search paths.

sysroot is only applied when the `PKG_CONFIG_SYSROOT_DIR` is set;
a future commit could apply it automatically when `--experimental-swift-sdk`
is used.

### DIFF
--- a/Sources/PackageLoading/PkgConfig.swift
+++ b/Sources/PackageLoading/PkgConfig.swift
@@ -177,6 +177,84 @@ internal struct PkgConfigParser {
         self.sysrootDir = sysrootDir
     }
 
+    // Compress repeated path separators to one.
+    private func compressPathSeparators(_ value: String) -> String {
+        let components = value.components(separatedBy: "/").filter { !$0.isEmpty }.joined(separator: "/")
+        if value.hasPrefix("/") {
+            return "/" + components
+        } else {
+            return components
+        }
+    }
+
+    // Trim duplicate sysroot prefixes, matching the approach of pkgconf
+    private func trimDuplicateSysroot(_ value: String) -> String {
+        // If sysroot has been applied more than once, remove the first instance.
+        // pkgconf makes this check after variable expansion to handle rare .pc
+        // files which expand ${pc_sysrootdir} directly:
+        //    https://github.com/pkgconf/pkgconf/issues/123
+        //
+        // For example:
+        //       /sysroot/sysroot/remainder -> /sysroot/remainder
+        //
+        // However, pkgconf's algorithm searches for an additional sysrootdir anywhere in
+        // the string after the initial prefix, rather than looking for two sysrootdir prefixes
+        // directly next to each other:
+        //
+        //     /sysroot/filler/sysroot/remainder -> /filler/sysroot/remainder
+        //
+        // It might seem more logical not to strip sysroot in this case, as it is not a double
+        // prefix, but for compatibility trimDuplicateSysroot is faithful to pkgconf's approach
+        // in the functions `pkgconf_tuple_parse` and `should_rewrite_sysroot`.
+
+        // Only trim if sysroot is defined with a meaningful value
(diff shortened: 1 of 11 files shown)


MaxDesiatov (approved):
Thanks!

euanh:
`sdkRootDir` is plumbed all the way through to where we call `PkgConfig.init`, so passing it through should be straightforward.   We also need to include the SDK path in the search paths used by `PCFileFinder` and check for corner cases.

Sources/PackageLoading/PkgConfig.swift
```diff
@@ -162,13 +166,93 @@ internal struct PkgConfigParser {
...
+    private func compressPathSeparators(_ value: String) -> String {
+        let components = value.components(separatedBy: "/").filter { !$0.isEmpty }.joined(separator: "/")
+        if value.hasPrefix("/") {
+            return "/" + components
+        } else {
+            return components
+        }
+    }
+
+    // Trim duplicate sysroot prefixes, matching the approach of pkgconf
```
MaxDesiatov, line 190: doc comment nit
```suggestion
    /// Trim duplicate sysroot prefixes, matching the approach of pkgconf
```

Sources/PackageLoading/PkgConfig.swift
```diff
@@ -162,13 +166,93 @@ internal struct PkgConfigParser {
...
+        // of the value, mirroring pkgconf's logic
+        let pathSuffix = value.dropFirst(sysrootDir.count)
+        if pathSuffix.contains(sysrootDir) {
+            return String(pathSuffix)
+        } else {
+            return value
+        }
+    }
+
+    // Apply sysroot to generated paths, matching the approach of pkgconf
```
MaxDesiatov, line 231: doc comment nit
```suggestion
    /// Apply sysroot to generated paths, matching the approach of pkgconf
```

MaxDesiatov:
@euanh feel free to cherry-pick this to `release/6.0.
Do you think it's worth mentioning this fix in the changelog? If so, would you mind adding an entry to `CHANGELOG.md` in this and the cherry-pick PR?

MaxDesiatov:
Ah, merged somewhat early, sorry. If it is changelog-worthy after all, that can be addressed in a follow-up PR.

PR #7463, opened 2024-04-16
Remove outdated doc comment in `Workspace+State.swift`

MaxDesiatov:
This "Returns true if the state file exists on the filesystem" doc comment was copy-pasted from a function above and is clearly wrong: `reload()` function doesn't return any values.


MaxDesiatov:
@swift-ci test

bnbarham (approved):

PR #7465, opened 2024-04-17
Add simple synthetic modules graph benchmarks

MaxDesiatov:
New `SyntheticModulesGraph` benchmark is introduced, which calls `loadModulesGraph` function that we already run against test fixtures in `ModulesGraphTests` and `BuildPlanTests`. The function under benchmark almost immediately delegates to `ModulesGraph.load`, which is used in real-world modules graph resolution.

Thresholds are now split into platform-specific directories so that thresholds recorded on x86_64 don't interfere with thresholds for arm64, same for the OS family.

### DIFF
--- a/Sources/Basics/Graph/DirectedGraph.swift
+++ b/Sources/Basics/Graph/DirectedGraph.swift
@@ -13,16 +13,17 @@
 import struct DequeModule.Deque
 
 /// Directed graph that stores edges in [adjacency lists](https://en.wikipedia.org/wiki/Adjacency_list).
-struct DirectedGraph<Node> {
-    init(nodes: [Node]) {
+@_spi(DontAdoptOutsideOfSwiftPMExposedForBenchmarksAndTestsOnly)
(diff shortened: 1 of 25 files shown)


MaxDesiatov:
@swift-ci test

Sources/Basics/Graph/DirectedGraph.swift
```diff
@@ -13,16 +13,17 @@
 import struct DequeModule.Deque
 
 /// Directed graph that stores edges in [adjacency lists](https://en.wikipedia.org/wiki/Adjacency_list).
-struct DirectedGraph<Node> {
-    init(nodes: [Node]) {
+@_spi(DontAdoptOutsideOfSwiftPMExposedForBenchmarksAndTestsOnly)
```
bnbarham, line 16: I think `@_spi(Testing)` or something similar is clear enough :)

bnbarham (approved):

PR #7467, opened 2024-04-17
[SE-0301] Updating a project and its manifest programmatically and from the command line

DougGregor:
Package manifest files are Swift source code, so editing them means working with the source code directly. Introduce a new library `PackageModelSyntax` based on swift-syntax that allows us to perform targeted manipulations of the manifest file programmatically, making suggested edits to the file without having to understand everything in it.

Introduce two editing commands:
* `addPackageDependency`: adds a new package dependency to the manifest file, e.g.,
    ```swift
     .package(url: "https://github.com/apple/swift-syntax.git", from: "510.0.1"),
    ```
* `addTarget`: adds a new target to the manifest file, e.g.,
    ```swift
     .executableTarget(
      name: "MyProgram",
      dependencies: [ "MyLib" ]
    )
    ```
In both cases, this is handled as a limited transform on the swift-syntax tree that then produces a series of edits that update the package manifest. We don't make any attempts to reason about the whole package, so this doesn't require any round-tripping with a semantic model. The editing commands can also introduce new files into the package. For example, adding target will create a new source file `Sources/\(targetName)/\(targetName).swift` with contents appropriate for the target type (struct for a library, `@main` for an executable, macro implementation struct for a macro target, etc.).

These editing operations are surfaced by implementing two of the commands from [SE-0301 "Package Editing Commands"](https://github.com/apple/swift-evolution/blob/main/proposals/0301-package-editing-commands.md), `swift package add-dependency` and `swift package add-target`. Adding a dependency looks a bit like this:

    swift package add-dependency --from 510.0.1 https://github.com/apple/swift-syntax.git

and will produce the `.package(...)` example above and insert it into the appropriate place in the manifest.

Adding a target looks like this:

    swift package add-target MyProgram --type executable --dependencies MyLib OtherLib

and adds this target to the manifest

    .executableTarget(
        name: "MyProgram",
        dependencies: [
            "MyLib",
            "OtherLib"
        ]
    ),

as well as `Sources/MyProgram/MyProgram.swift`:

    import MyLib
    import OtherLib

    @main
    struct MyProgramMain {
        static func main() {
            print("Hello, world")
        }
    }


MaxDesiatov:
@swift-ci test

Sources/PackageModel/ManifestSourceGeneration.swift
```diff
@@ -633,7 +633,7 @@ public struct SourceCodeFragment {
         self.subnodes = subnodes
     }
     
-    func generateSourceCode(indent: String = "") -> String {
+    public func generateSourceCode(indent: String = "") -> String {
```
MaxDesiatov, line 636: Would you mind changing all functions that were made `public` in this PR to `package` instead?

Tests/PackageModelSyntaxTests/ManifestEditTests.swift
```diff
@@ -0,0 +1,312 @@
...
+///   - actual: The actual string.
+///   - expected: The expected string.
+///   - message: An optional description of the failure.
+///   - additionalInfo: Additional information about the failed test case that will be printed after the diff
+///   - file: The file in which failure occurred. Defaults to the file name of the test case in
+///     which this function was called.
+///   - line: The line number on which failure occurred. Defaults to the line number on which this
+///     function was called.
+public func assertStringsEqualWithDiff(
+  _ actual: String,
```
MaxDesiatov, line 184: These functions use formatting inconsistent with 4 spaces in the rest of the file.

Sources/PackageModel/ManifestSourceGeneration.swift
DougGregor: Ah, sure! The only exception is likely to be the `AddPackageDependency` type itself, because I also want to use that from SourceKit-LSP.

Tests/PackageModelSyntaxTests/ManifestEditTests.swift
DougGregor: 'cause I copy-pasted it, thanks!

Sources/PackageModelSyntax/AddPackageDependency.swift
```diff
@@ -0,0 +1,111 @@
...
+                    leadingTrivia: .space,
+                    elements: [
+                        ArrayElementSyntax(
+                            leadingTrivia: innerTrivia,
+                            expression: dependencySyntax,
+                            trailingComma: .commaToken()
+                        )
+                    ],
+                    rightSquare: .rightSquareToken()
+                        .with(\.leadingTrivia, leadingTrivia)
```
ahoppen, line 90:
```suggestion
                    rightSquare: .rightSquareToken(leadingTrivia: leadingTrivia)
```

Sources/PackageModelSyntax/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,225 @@
...
+//===----------------------------------------------------------------------===//
+
+import Basics
+import PackageModel
+import SwiftSyntax
+import SwiftParser
+
+/// Default indent when we have to introduce indentation but have no context
+/// to get it right.
+let defaultIndent = TriviaPiece.spaces(4)
```
ahoppen, line 20: We do actually have a utility to infer indentation in swift-syntax: https://github.com/apple/swift-syntax/blob/728e2f6d5f9fd4a8e45eab6e52e86ad1519a7e2a/Sources/SwiftBasicFormat/InferIndentation.swift#L19-L27

Sources/PackageModelSyntax/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,225 @@
...
+extension Trivia {
+    /// Determine whether this trivia has newlines or not.
+    var hasNewlines: Bool {
+        contains { piece in
+            if case .newlines = piece {
+                return true
+            } else {
+                return false
+            }
+        }
```
ahoppen, line 31:
```suggestion
        contains(where: \.isNewline)
```

This should also cover `\r\n`

Sources/PackageModelSyntax/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,225 @@
...
+        return .visitChildren
+    }
+}
+
+extension SyntaxProtocol {
+    /// Find the first node of the Self type that matches the given predicate.
+    static func findFirst(
+        in node: some SyntaxProtocol,
+        matching predicate: (Self) -> Bool
+    ) -> Self? {
```
ahoppen, line 65: Wouldn’t it make more sense for this to be an instance function on `SyntaxProtocol` instead of a static function?

Sources/PackageModelSyntax/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,225 @@
...
+extension ArrayExprSyntax {
+    /// Produce a new array literal expression that appends the given
+    /// element, while trying to maintain similar indentation.
+    func appending(
+        element: ExprSyntax,
+        outerLeadingTrivia: Trivia
+    ) -> ArrayExprSyntax {
+        var elements = self.elements
+
+        let commaToken = TokenSyntax(.comma, presence: .present)
```
ahoppen, line 180:
```suggestion
        let commaToken = TokenSyntax.commaToken()
```

Sources/PackageModelSyntax/SyntaxEditUtils.swift
```diff
@@ -0,0 +1,225 @@
...
+        if let last = elements.last {
+            // The leading trivia of the new element should match that of the
+            // last element.
+            leadingTrivia = last.leadingTrivia
+
+            // Add a trailing comma to the last element if it isn't already
+            // there.
+            if last.trailingComma == nil {
+                var newElements = Array(elements)
+                newElements[newElements.count-1] = last.with(\.trailingComma, commaToken)
```
ahoppen, line 195:
```suggestion
                newElements[newElements.count-1]?.trailingComma = commaToken
```

Tests/PackageModelSyntaxTests/ManifestEditTests.swift
```diff
@@ -0,0 +1,312 @@
...
+            )
+            """) { manifest in
+            AddPackageDependency.addPackageDependency(
+                Self.swiftSystemPackageDependency,
+                to: manifest,
+                manifestDirectory: try! AbsolutePath(validating: "/")
+            )
+        }
+    }
+}
```
ahoppen, line 167: I think it would be good to also have a test case where there is an existing dependency on a single line array literal. Ie add a dependency to the following

```swift
let package = Package(
    name: "packages",
    dependencies: [.package(url: "https://github.com/apple/swift-syntax.git", from: "510.0.1")]
)
```

Sources/PackageModelSyntax/SyntaxEditUtils.swift
DougGregor: Oh, huh. Should I be worried that this is processing the entire file? I suppose I should thread through an optional indentation value if we're going to do that.

Sources/PackageModelSyntax/SyntaxEditUtils.swift
DougGregor: No, because you are looking for something of the `Self` type.

Tests/PackageModelSyntaxTests/ManifestEditTests.swift
DougGregor: ... and it uncovered an issue with trivia, thanks!

DougGregor:
This is in decent shape *except* that the bootstrap script is totally broken.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test macOS

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test macOS

DougGregor:
@swift-ci please test

owenv (approved):

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

Sources/PackageModelSyntax/SyntaxEditUtils.swift
ahoppen: I was thinking if it is a cleaner API to have it be like

```swift
tree.findFirst { (node: FunctionCallSyntax) -> Bool in
  doChecks(for: node)
}
```

Instead of 
```swift
FunctionCallSyntax.findFirst(in: tree) { node in
  doChecks(for: node)
}
```
which reads a little backwards to me. But maybe it’s a matter of taste.

Sources/PackageModelSyntax/SyntaxEditUtils.swift
ahoppen: I don’t think it should be a huge deal if the user actually invokes `swift package add` and better than just using 4 spaces which does tend to look stupid if the rest of the file is indented with 2 spaces.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

MaxDesiatov (approved):

Sources/PackageModelSyntax/AddPackageDependency.swift
```diff
@@ -0,0 +1,120 @@
...
+import SwiftParser
+import SwiftSyntax
+import SwiftSyntaxBuilder
+
+/// Add a package dependency to a manifest's source code.
+public struct AddPackageDependency {
+    /// The set of argument labels that can occur after the "dependencies"
+    /// argument in the Package initializers.
+    ///
+    /// TODO: Could we generate this from the the PackageDescription module, so
```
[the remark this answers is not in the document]
DougGregor: If we're going to look there, we might as well just parse the PackageDescription module sources, since we have swift-syntax around.

Sources/PackageModelSyntax/SyntaxEditUtils.swift
DougGregor: I hate requiring type annotations on closures... but I can create an `ofType` argument or similar to make this cleaner.

PR #7471, opened 2024-04-19
Place build.db under the scratch directory

kateinoigakukun:
### Motivation:

The build database should be shared between different triple builds to re-generate the build manifest correctly.

With the current implementation, the following build command sequence fails:
```
# 1st iteration: They both do not trigger "PackageStructure" build
$ swift build --experimental-swift-sdk wasm32-unknown-wasi
$ swift build

# 2nd iteration: They both trigger "PackageStructure" build because there is description.json and debug.yaml
$ swift build --experimental-swift-sdk wasm32-unknown-wasi
$ swift build

# 3rd iteration: Manifest cache entry in .build/wasm32-unknown-wasi/build.db created by 2nd iteration hits,
# so do not update debug.yaml. <--- Incorrect
$ swift build --experimental-swift-sdk wasm32-unknown-wasi
```

### Modifications:

This changes the llbuild build database to be placed under `.build/build.db` instead of `.build/<product triple>/build.db`.
Also this change splits llbuild target names for each triple to avoid cache invalidation when switching triple.

### Result:

`build.db` will be shared across product target triples, and SwiftPM will keep consistent cache state when switching triples.

### DIFF
--- a/Sources/Build/BuildManifest/LLBuildManifestBuilder.swift
+++ b/Sources/Build/BuildManifest/LLBuildManifestBuilder.swift
@@ -326,24 +326,25 @@ extension ResolvedModule {
     }
 
     public func getLLBuildResourcesCmdName(buildParameters: BuildParameters) -> String {
-        "\(self.name)-\(buildParameters.buildConfig)\(buildParameters.suffix(triple: self.buildTriple)).module-resources"
+        "\(self.name)-\(buildParameters.triple.tripleString)-\(buildParameters.buildConfig)\(buildParameters.suffix(triple: self.buildTriple)).module-resources"
     }
 }
 
 extension ResolvedProduct {
     public func getLLBuildTargetName(buildParameters: BuildParameters) throws -> String {
+        let triple = buildParameters.triple.tripleString
         let config = buildParameters.buildConfig
         let suffix = buildParameters.suffix(triple: self.buildTriple)
-        let potentialExecutableTargetName = "\(name)-\(config)\(suffix).exe"
-        let potentialLibraryTargetName = "\(name)-\(config)\(suffix).dylib"
+        let potentialExecutableTargetName = "\(name)-\(triple)-\(config)\(suffix).exe"
+        let potentialLibraryTargetName = "\(name)-\(triple)-\(config)\(suffix).dylib"
 
         switch type {
         case .library(.dynamic):
             return potentialLibraryTargetName
         case .test:
-            return "\(name)-\(config)\(suffix).test"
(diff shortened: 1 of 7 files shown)


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

Tests/BuildTests/BuildOperationTests.swift
```diff
@@ -14,34 +14,60 @@
...
+import PackageGraph
+import SPMBuildCore
 import SPMTestSupport
 import XCTest
 
 import class TSCBasic.BufferedOutputByteStream
 import class TSCBasic.InMemoryFileSystem
 
+private func mockBuildOperation(
+    buildParameters: BuildParameters,
```
xedin, line 28: Could you please split the parameters here into tools and destination? I am adding a destination to the parameters in another PR related to cross-compilation which makes this separation functional.

Sources/Build/BuildManifest/LLBuildManifestBuilder.swift
```diff
@@ -322,7 +322,7 @@ extension ResolvedModule {
     }
 
     public func getLLBuildTargetName(buildParameters: BuildParameters) -> String {
-        "\(self.name)-\(buildParameters.buildConfig)\(buildParameters.suffix(triple: self.buildTriple)).module"
+        "\(self.name)-\(buildParameters.triple.tripleString)-\(buildParameters.buildConfig)\(buildParameters.suffix(triple: self.buildTriple)).module"
```
xedin, line 325: What about other names below, do we not need exe and libs associated with triples as well?

domciiiiix: 👍🏻

kateinoigakukun: Yes, we need it to leverage build cache more when switching triples

Tests/BuildTests/BuildOperationTests.swift
kateinoigakukun: Ok, make sense to me

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test Windows

domciiiiix (approved):

xedin (approved):
LGTM, thank you! I started a source compatibility build with these changes @ https://github.com/apple/swift/pull/73908

kateinoigakukun:
Ok, source compat tests passed. At least this change does not break initial clean build 😌

PR #7472, opened 2024-04-19
[6.0] pkgconfig: Apply PKG_CONFIG_SYSROOTDIR when generating paths

euanh:
Cherry pick of #7461 

SwiftPM's pkg-config implementation sets the `pc_sysrootdir` variable, but most `.pc` files do not use this variable directly. Instead, they rely on
the pkg-config tool rewriting the generated paths to include the sysroot prefix when necessary. SwiftPM does not do this, so it does not generate the correct compiler flags to use libraries from a sysroot.

This problem was reported in issue
https://github.com/apple/swift-package-manager/issues/7409

There are two major pkg-config implementations which handle sysroot differently:

  *  `pkg-config` (the original https://pkg-config.freedesktop.org implementation) prepends sysroot after variable expansion, when it creates the compiler flag lists

  *  `pkgconf` (the newer http://pkgconf.org implementation) prepends sysroot to variables when they are defined, so sysroot is included when they are expanded

`pkg-config`'s method skips single character compiler flags, such as `-I`
and `-L`, and has special cases for longer options. It does not handle spaces between the flags and their values properly, and prepends sysroot multiple times in some cases, such as when the .pc file uses the `sysroot_dir` variable directly or has been rewritten to hard-code the sysroot prefix.

`pkgconf`'s method handles spaces correctly, although it also makes extra checks to ensure that sysroot is not applied more than once.

In 2024 `pkg-config` is the more popular option according to Homebrew installation statistics, but the major Linux distributions have generally
switched to `pkgconf`.

We will use `pkgconf`'s method here as it seems more robust than `pkg-config`'s, and `pkgconf`'s greater popularity on Linux means libraries
developed there may depend on the specific way it handles `.pc` files.

SwiftPM will now apply the sysroot prefix to compiler flags, such as include (`-I`) and library (`-L`) search paths.

This is a partial fix for
https://github.com/apple/swift-package-manager/issues/7409.

The sysroot prefix is only applied when the `PKG_CONFIG_SYSROOT_DIR` environment variable is set. A future commit could apply an appropriate sysroot automatically when the `--experimental-swift-sdk` flag is used.

**Scope**: Limited to packages relying on system libraries that utilize pkg-config.
**Risk**: Low, changes are isolated and the scope is limited to a small fraction of packages.
**Testing**: Automated with new test cases.
**Issues**: https://github.com/apple/swift-package-manager/issues/7409
**Reviewer**: @MaxDesiatov.

### DIFF
--- a/Sources/PackageLoading/PkgConfig.swift
+++ b/Sources/PackageLoading/PkgConfig.swift
@@ -47,6 +47,7 @@ public struct PkgConfig {
         name: String,
         additionalSearchPaths: [AbsolutePath]? = .none,
         brewPrefix: AbsolutePath? = .none,
+        sysrootDir: AbsolutePath? = .none,
         fileSystem: FileSystem,
         observabilityScope: ObservabilityScope
     ) throws {
@@ -54,6 +55,7 @@ public struct PkgConfig {
             name: name,
             additionalSearchPaths: additionalSearchPaths ?? [],
             brewPrefix: brewPrefix,
+            sysrootDir: sysrootDir,
             loadingContext: LoadingContext(),
             fileSystem: fileSystem,
             observabilityScope: observabilityScope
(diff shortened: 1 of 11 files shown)


euanh:
@swift-ci test

bnbarham (approved):

Issue #7478, opened 2024-04-22
AddTarget incorrectly assumes an XCTest dependency.


grynspan:
The new AddTarget code should not always assume that XCTest is a dependency for a new test target. Rather, behaviour should be configured based on the `--enable-xctest`/`--disable-xctest` and `--enable-experimental-swift-testing`/`--disable-experimental-swift-testing` flags, with XCTest on and swift-testing off by default. Compare the `swift package init` command.

PR #7481, opened 2024-04-23
Teach swift package add-target --type test about swift-testing 

DougGregor:
Introduce a command-line argument `--testing-library` to the add-target
command to specify which test library to generate the test for. This
can be 'xctest' (the prior XCTest behavior), 'swift-testing' (to use the
new swift-testing library), or 'none' (for no test library at all).

For the new swift-testing generation, also add the appropriate package
and test target dependency, along with a stub testsuite to start from.

Fixes https://github.com/apple/swift-package-manager/issues/7478

### DIFF
--- a/Sources/Commands/PackageCommands/AddTarget.swift
+++ b/Sources/Commands/PackageCommands/AddTarget.swift
@@ -21,6 +21,8 @@ import TSCBasic
 import TSCUtility
 import Workspace
 
+extension AddTarget.TestHarness: ExpressibleByArgument { }
+
 extension SwiftPackageCommand {
     struct AddTarget: SwiftCommand {
         /// The type of target that can be specified on the command line.
@@ -57,6 +59,9 @@ extension SwiftPackageCommand {
 
         @Option(help: "The checksum for a remote binary target")
         var checksum: String?
+
+        @Option(help: "The testing library to use when generating test targets, which can be one of 'xctest', 'swift-testing', or 'none'")
+        var testingLibrary: PackageModelSyntax.AddTarget.TestHarness = .default
 
         func run(_ swiftCommandState: SwiftCommandState) throws {
             let workspace = try swiftCommandState.getActiveWorkspace()
@@ -110,6 +115,7 @@ extension SwiftPackageCommand {
             let editResult = try PackageModelSyntax.AddTarget.addTarget(
                 target,
                 to: manifestSyntax,
+                configuration: .init(testHarness: testingLibrary),
                 installedSwiftPMConfiguration: swiftCommandState
                   .getHostToolchain()
                   .installedSwiftPMConfiguration
--- a/Sources/PackageModel/InstalledSwiftPMConfiguration.swift
+++ b/Sources/PackageModel/InstalledSwiftPMConfiguration.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-public struct InstalledSwiftPMConfiguration: Codable {
+public struct InstalledSwiftPMConfiguration {
     public struct Version: Codable, CustomStringConvertible {
         let major: Int
         let minor: Int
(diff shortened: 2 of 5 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

Sources/Commands/PackageCommands/AddTarget.swift
```diff
@@ -60,6 +60,9 @@ extension SwiftPackageCommand {
         @Option(help: "The checksum for a remote binary target")
         var checksum: String?
 
+        @Option(help: "The testing library to use when generating test targets, which can be one of 'xctest', 'swift-testing', or 'none'")
+        var testingLibrary: PackageModelSyntax.AddTarget.TestHarness = .default
```
grynspan, line 64: Please break this up into two flags, `--enable/disable-xctest` and `--enable/disable-experimental-swift-testing`:

1. It would be consistent with the flags we've added to other commands;
2. swift-testing is experimental only;
3. The default behaviour if neither is specified should be to add support for XCTest only;
4. In the future, we can always change the default if needed; and
5. It is valid to specify _both_ libraries as dependencies for a test target.

grynspan (changes requested):
Thanks for opening this! I understand your motivation in wanting to use a single option here, however splitting it into two flags is important for a few reasons I listed in comments.

Thanks again!

Sources/Commands/PackageCommands/AddTarget.swift
DougGregor: What you are proposing is a confusing design for functionality that is adding a brand-new test target. Roughly nobody is going to create a new test target that uses both of these testing libraries: they'll pick one and stick with it.

Other commands need to deal with one or both libraries being present, because existing test targets likely use XCTest and will want to pick up Swift Testing. This functionality is not for those use cases.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

MaxDesiatov (approved):

Sources/PackageModel/InstalledSwiftPMConfiguration.swift
```diff
@@ -40,7 +41,56 @@ public struct InstalledSwiftPMConfiguration: Codable {
                 minor: 0,
                 patch: 0,
                 prereleaseIdentifier: "latest"
-            )
+            ),
+            swiftTestingVersionForTestTemplate: defaultSwiftTestingVersionForTestTemplate
         )
     }
+
+    private static var defaultSwiftTestingVersionForTestTemplate: Version {
```
bnbarham, line 49: I'd rather not have this at all and solely use config.json, `swiftSyntaxVersionForMacroTemplate` should also be removed but we can at least not add another.

Tests/PackageModelSyntaxTests/ManifestEditTests.swift
```diff
@@ -564,6 +564,53 @@ class ManifestEditTests: XCTestCase {
...
 
+    func testAddSwiftTestingTestTarget() throws {
+        try assertManifestRefactor("""
+            // swift-tools-version: 5.5
+            let package = Package(
+                name: "packages"
+            )
+            """,
+            expectedManifest: """
+            // swift-tools-version: 5.5
```
bnbarham, line 575: Where does 5.5 actually come from 🤔?

Sources/PackageModel/InstalledSwiftPMConfiguration.swift
DougGregor: Yeah, this was driven by a failure in Linux CI in prior iterations of this patch, where we were getting a `config.json` from an older toolchain. It seemed like a reasonable policy that we would support at least slightly-older toolchains, so the default bridges that gap. If we want to be strict about this, we'll need to track down the configuration that produced this failure.

Tests/PackageModelSyntaxTests/ManifestEditTests.swift
DougGregor: 5.5 introduced some tweaks to the package manifest format (e.g., some things moved to labels), such as introducing new `package` dependencies and deprecating some old ones. We can either write a bunch of code to deal with all of the differences, or we can limit the functionality to 5.5 or newer. The latter was easier.

grynspan: swift-testing is going to need a manifest version of 6.0 or later at some point in the future.

grynspan: (I'm not sure if that affects its _clients_—I assume not?)

DougGregor: Yeah, if we end up changing the manifest format in some way for swift-testing, we'll need make changes here as well. We can enforce a 6.0 manifest or help with updating the manifest.

Sources/PackageModelSyntax/AddTarget.swift
```diff
@@ -184,14 +245,35 @@ public struct AddTarget {
...
+                }
+                """
+
+            case .swiftTesting:
+                """
+                \(imports)
+                @Suite
+                struct \(raw: target.name)Tests {
+                    @Test("\(raw: target.name) tests")
+                    func tests() {
```
stmontgomery, line 271: Stylistic suggestion: I recommend naming the func something not prefixed by the word "test", like `example()`. The `@Test` attribute already marks it as a test so having that prefix can feel redundant. (I realize this was just adapted from the prior XCTest template, of course!)

Utilities/config.json
```diff
@@ -1 +1,3 @@
-{"version":1,"swiftSyntaxVersionForMacroTemplate":{"major":600,"minor":0,"patch":0, "prereleaseIdentifier":"latest"}}
+{"version":1,
+  "swiftSyntaxVersionForMacroTemplate":{"major":600,"minor":0,"patch":0, "prereleaseIdentifier":"latest"},
+  "swiftTestingVersionForTestTemplate":{"major":0,"minor":7,"patch":0}}
```
stmontgomery, line 3: We recently released tag 0.8.0 so you can bump this, as well as another place I saw in this PR

stmontgomery (approved):
Left a couple comments but this largely looks good to me from the perspective of a swift-testing maintainer! Thank you

Sources/PackageModelSyntax/AddTarget.swift
DougGregor: Sure, I'll drop the test prefix for swift-testing.

Utilities/config.json
DougGregor: I'll bump in both places, thanks!

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test Linux

DougGregor:
https://github.com/apple/swift/pull/73507

@swift-ci please test Linux

PR #7482, opened 2024-04-23
Make `SwiftSDK.init` available via `@_spi`

MaxDesiatov:
This initializer is an internal implementation detail that should only be available via `@_spi`.


MaxDesiatov:
@swift-ci test

bnbarham (approved):

PR #7483, opened 2024-04-23
Cover `–-toolchain` argument in `SwiftCommandStateTests`

MaxDesiatov:
We should prevent possible regressions in how this argument is handled.

### DIFF
--- a/Sources/Basics/FileSystem/InMemoryFileSystem.swift
+++ b/Sources/Basics/FileSystem/InMemoryFileSystem.swift
@@ -1,0 +1,496 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+ */
+
+import class Foundation.NSLock
+import class Dispatch.DispatchQueue
+import struct TSCBasic.AbsolutePath
+import struct TSCBasic.ByteString
+import class TSCBasic.FileLock
+import enum TSCBasic.FileMode
+import struct TSCBasic.FileSystemError
+
+/// Concrete FileSystem implementation which simulates an empty disk.
+public final class InMemoryFileSystem: FileSystem {
+    /// Private internal representation of a file system node.
+    /// Not thread-safe.
+    private class Node {
+        /// The actual node data.
+        let contents: NodeContents
+        
+        /// Whether the node has executable bit enabled.
+        var isExecutable: Bool
+
+        init(_ contents: NodeContents, isExecutable: Bool = false) {
+            self.contents = contents
+            self.isExecutable = isExecutable
+        }
+
+        /// Creates deep copy of the object.
+        func copy() -> Node {
+            return Node(contents.copy())
+        }
+    }
+
+    /// Private internal representation the contents of a file system node.
+    /// Not thread-safe.
+    private enum NodeContents {
+        case file(ByteString)
+        case directory(DirectoryContents)
+        case symlink(String)
+
+        /// Creates deep copy of the object.
+        func copy() -> NodeContents {
+            switch self {
+            case .file(let bytes):
+                return .file(bytes)
+            case .directory(let contents):
+                return .directory(contents.copy())
+            case .symlink(let path):
+                return .symlink(path)
+            }
+        }
+    }
+
+    /// Private internal representation the contents of a directory.
+    /// Not thread-safe.
+    private final class DirectoryContents {
+        var entries: [String: Node]
+
+        init(entries: [String: Node] = [:]) {
+            self.entries = entries
+        }
+
+        /// Creates deep copy of the object.
+        func copy() -> DirectoryContents {
+            let contents = DirectoryContents()
+            for (key, node) in entries {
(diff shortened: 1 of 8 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

Sources/Basics/FileSystem/InMemoryFileSystem.swift
MaxDesiatov, line 1: Vendored from TSC to reduce our dependency on TSC.

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
```diff
@@ -738,7 +730,7 @@ public struct SwiftSDK: Equatable {
     /// Appends a path to the array of toolset root paths.
```
xedin, line 730: Nit: This comment needs to be changed to reflect the behavior change.

xedin: There is also https://github.com/apple/swift-package-manager/blob/main/Sources/PackageModel/SwiftSDKs/SwiftSDK.swift#L717-L719 that needs to be fixed.

xedin:
@swift-ci please test

bnbarham (approved):

xedin:
For posterity: I changed the logic here slightly, `--toolchain` gets prepended (to override existing entries) but otherwise we keep the ordering of the entries in `rootPaths` the same which means that we are going to select a tool from the first matching root in order they were provided and the fact that we weren't doing that before seems unintentional.

xedin:
@swift-ci please clean test Linux platform

xedin:
@swift-ci please clean test Linux platform

xedin:
@swift-ci please clean test Linux platform

xedin:
@swift-ci please clean test Linux platform

xedin:
For posterity there are currently multiple issues here

- `_hostToolchain` doesn't use `environment` and `fileSystem` provided instead if fallback to `.process()` and `localFileSystem` which cannot accommodate overrides that are only available in-memory
- Changing only `environment` runs into issues with `SWIFTPM_CUSTOM_BIN_DIR` 
- Changing both `environment` and `fileSystem` results in a few problems on Linux:
   - CI specifies SWIFT_EXEC* which wouldn't be available in in-memory filesystem
   - Dropping that crashes in UserToolchain.init when it tries to determine `targetTriple` for host SDK by running `/swiftpm/bin/swiftc -print-target-info`

MaxDesiatov:
>    - Dropping that crashes in UserToolchain.init when it tries to determine `targetTriple` for host SDK by running `/swiftpm/bin/swiftc -print-target-info`

I worked on this one specifically in https://github.com/apple/swift-package-manager/commit/ce7058225954acb4a1bb65dd3938209c43b55667, when specifying a fake environment with in-memory FS one should provide a predetermined host triple to avoid trying to spawn that process at a fake path.

xedin:
@MaxDesiatov This cannot be done at the moment for SwiftCommandState which is under test, I should have mentioned explicitly that the only failing test on Linux is the new one in SwiftCommandStateTests. 

We need to figure out whether we want to mock a host toolchain there, provide a host triple somehow or do something else…

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test linux

MaxDesiatov:
@swift-ci test macos

MaxDesiatov:
@swift-ci test windows

Sources/CoreCommands/SwiftCommandState.swift
```diff
@@ -872,11 +876,14 @@ public final class SwiftCommandState {
...
-            try UserToolchain(
-                swiftSDK: SwiftSDK.hostSwiftSDK(
-                    environment: self.environment,
-                    observabilityScope: self.observabilityScope
-                ),
+            var hostSwiftSDK = try SwiftSDK.hostSwiftSDK(
+                environment: self.environment,
+                observabilityScope: self.observabilityScope
+            )
+            hostSwiftSDK.targetTriple = self.hostTriple
```
xedin, line 883: Not for this PR - maybe we should rename this to `triple`?

MaxDesiatov: On `SwiftSDK` or on `SwiftCommandState`?

xedin: SwiftSDK

MaxDesiatov: SwiftSDKs have both triples specified on purpose, as one can have a Swift SDK installed supporting certain host triples that don't include the triple of a machine it's installed on.

PR #7492, opened 2024-04-24
Report whether a target is part of the root package in the sourcekit-lsp API

ahoppen:
This is used so we don’t search for tests in targets of package dependencies.

Companion of https://github.com/apple/sourcekit-lsp/pull/1201

rdar://126965614

### DIFF
--- a/Sources/SourceKitLSPAPI/BuildDescription.swift
+++ b/Sources/SourceKitLSPAPI/BuildDescription.swift
@@ -22,21 +22,33 @@ import class Build.BuildPlan
 import class Build.ClangTargetBuildDescription
 import class Build.SwiftTargetBuildDescription
 import struct PackageGraph.ResolvedModule
+import struct PackageGraph.ModulesGraph
 
 public protocol BuildTarget {
     var sources: [URL] { get }
 
-    func compileArguments(for fileURL: URL) throws -> [String]
- }
+    /// Whether the target is part of the root package that the user opened or if it's part of a package dependency.
+    var isPartOfRootPackage: Bool { get }
 
-extension ClangTargetBuildDescription: BuildTarget {
+    func compileArguments(for fileURL: URL) throws -> [String]
+}
+
+private struct WrappedClangTargetBuildDescription: BuildTarget {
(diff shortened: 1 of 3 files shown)


Sources/SourceKitLSPAPI/BuildDescription.swift
```diff
@@ -22,31 +22,45 @@ import class Build.BuildPlan
 import class Build.ClangTargetBuildDescription
 import class Build.SwiftTargetBuildDescription
 import struct PackageGraph.ResolvedModule
+import struct PackageGraph.ModulesGraph
```
MaxDesiatov, line 25: Is there a way to implement this with `private import`? All of the types above without `private import` should not leak outside of `SourceKitLSPAPI` and we should work on making this list smaller instead of adding more to it.

ahoppen: I viewed it similarly to how we use `ResolvedModule` in the API as well. 

I agree that we should not expose these types in SourceKitLSPAPI, but completely wrapping `ModuleGraph` will be a fairly significant undertaking (we use it in other places in SourceKit-LSP as well) and I think that should be done when we switch all of SourceKitLSP to only use the API module. Doing that for 6.0 is too much work and risk though, so I'd prefer to keep this as-is for now and schedule time for 6.1 to extend the SourceKitLSPAPI module so that we don't need to import any other modules. WDYT?

MaxDesiatov: That's fair, although I would like `ModulesGraph` to go through a significant refactoring for performance improvements in the near future, which very likely is going to be a breaking change. I'm not sure of the timeline yet, it might need to be included in 6.0 if we see performance regressing too much due to other changes.

MaxDesiatov:
@swift-ci test

ahoppen:
https://github.com/apple/sourcekit-lsp/pull/1201

@swift-ci Please test

MaxDesiatov (approved):

ahoppen:
https://github.com/apple/sourcekit-lsp/pull/1201

@swift-ci Please test

PR #7495, opened 2024-04-25
Set environment variable to disable building swift-syntax in Swift 6 mode

ahoppen:
This allows us to work around the following issue in CI: The self-hosted SwiftPM job has Xcode 15.3 (Swift 5.10) installed and builds a Swift 6 SwiftPM from source. It then tries to build itself as a fat binary using the just-built Swift 6 SwiftPM, which uses xcbuild from Xcode as the build system. But the xcbuild in the installed Xcode is too old and doesn't know about Swift 6 mode, so it fails with: SWIFT_VERSION '6' is unsupported, supported versions are: 4.0, 4.2, 5.0 (rdar://126952308). This is fixed by setting `SWIFTSYNTAX_DISABLE_SWIFT_6_MODE` in `build-using-self`.

The source compat suite is seeing the same issue, just by using `bootstrap`, so set it there as well.

Accompanies https://github.com/apple/swift-syntax/pull/2621


ahoppen:
https://github.com/apple/swift-syntax/pull/2621

@swift-ci Please test

MaxDesiatov (approved):

PR #7496, opened 2024-04-25
Improvements to prebuilt or provided library handling

xedin:
### Motivation:

The idea is to enable provided libraries to be a part of a toolchain which means that instead of completely eliding them from a package graph they need to be handled post-resolution in a special way and reflected in a build plan for swift targets (as compiler and linker arguments).

### Modifications:

- Instead of removing packages that are backed by a prebuilt library, dependency resolution marks `BoundVersion` associated with them with a `ProvidedLibrary` (this struct carries information about library location and its metadata).

- `PackageChangeState` is adjusted to reflect a state transition to use a library (unless the package is edited).

- New `.providedLibrary` managed dependency is added to handle loading of provided library manifests. Products and targets of the manifests are formed from `.swiftmodules` found in the library location.  New `TargetDescription.Kind` - `.providedLibrary` is added to model that (requires only a name and a absolute path to the library root).
  - I attempted to use existing `BinaryTarget` but the structure there was proven too rigid, maybe in the future we could package these libraries as frameworks and it would make things easier.

- `BuildPlan` is adjusted to inject `-I` to compiler arguments for each library together with `-L` and `-l` for each used target injected into product link arguments.

### Result:

Provided libraries can now be shipped with a toolchain in `share/pm/<name>` location. `<name>` is a `productName` is `provided-libraries.json` config.


Sources/Workspace/Workspace+Dependencies.swift
```diff
@@ -966,6 +961,9 @@ extension Workspace {
         /// The package is updated.
         case updated(State)
 
+        /// The package is replaced with a prebuilt library
+        case usesLibrary(ProvidedLibrary)
```
xedin, line 965: @MaxDesiatov I was going back and forth on this - it would be possible to adjust `unchanged` to optionally carry this and then if version has a provided library the state would always be `unchanged`. WDYT?

MaxDesiatov: I don't have a strong opinion, but for some reason I slightly prefer the non-optional case like it is now.

MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

xedin:
@swift-ci please test

xedin:

xedin:
https://github.com/apple/sourcekit-lsp/pull/1207
@swift-ci please test

PR #7502, opened 2024-04-25
[Build] Fix up the cmake build after swift-syntax was added

bnbarham:
Doesn't FetchContent when it isn't available right now, but we don't fetch any of the other required dependencies either. Can fix that when we fix them all.


bnbarham:
https://github.com/apple/swift/pull/73267

@swift-ci please test

DougGregor (approved):
This definitely looks simpler

bnbarham:
https://github.com/apple/swift/pull/73267

@swift-ci please test

bnbarham:
I forgot we bootstrap with CMake even outside of Windows :(

PR #7506, opened 2024-04-26
plugins: Pass correct dependency origin information to plugins

euanh:
### Motivation:

SwiftPM sends information about the origin of each package dependency to the plugin process, but the `PluginContext` passed to the plugin function incorrectly shows all origins as `root`.

### Modifications:

The necessary information is included correctly in the serialized structure sent to the plugin by SwiftPM.    The startup wrapper code in the plugin process correctly deserializes it but does not copy it into the final `PluginContext` struct which is passed to the plugin function; instead, `origin` is hard-coded to `.root`.

This change copies the dependency information into the final context struct.

### Result:

Plugins receive correct information about package dependencies - `local`, with a local path, or `repository` with a URL, version number and revision hash.

### DIFF
--- a/Sources/PackagePlugin/PluginContextDeserializer.swift
+++ b/Sources/PackagePlugin/PluginContextDeserializer.swift
@@ -251,7 +251,7 @@ internal struct PluginContextDeserializer {
         }
         let products = try wirePackage.productIds.map { try self.product(for: $0) }
         let targets = try wirePackage.targetIds.map { try self.target(for: $0) }
-        let origin : PackageOrigin = switch wirePackage.origin {
+        let origin: PackageOrigin = switch wirePackage.origin {
             case .root:
                 .root
             case .local(let pathId):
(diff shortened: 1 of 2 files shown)


Sources/PackagePlugin/PluginContextDeserializer.swift
```diff
@@ -251,12 +251,22 @@ internal struct PluginContextDeserializer {
         }
         let products = try wirePackage.productIds.map { try self.product(for: $0) }
         let targets = try wirePackage.targetIds.map { try self.target(for: $0) }
+        let origin : PackageOrigin = switch wirePackage.origin {
```
MaxDesiatov, line 254: formatting nit
```suggestion
        let origin: PackageOrigin = switch wirePackage.origin {
```

MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):
Thanks! 🙏

MaxDesiatov:
@swift-ci test windows

PR #7507, opened 2024-04-26
Remove `experimental` prefix from Swift SDK command

MaxDesiatov:
Since [SE-0387](https://github.com/apple/swift-evolution/blob/main/proposals/0387-cross-compilation-destinations.md) was accepted and available in Swift 5.9 and Swift 5.10, it's time to remove `experimental` prefix.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

bnbarham (approved):

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@bnbarham @euanh @kateinoigakukun do you think it's worth keeping old subcommand and argument and make them show a deprecation warning for backwards compatibility?

bnbarham:
Yeah I think that'd be nice as long as we hide it from help in general.

kateinoigakukun:
Also, it might be a good timing to refine CLI option consistency with others. 
e.g. 
- `swift package plugin --list` / `swift sdk list`
- `swift package config` / `swift sdk configuration`

MaxDesiatov:
Not sure about that one, might require a proposal amendment as we've [specified commands and their behavior explicitly in the proposal text](https://github.com/apple/swift-evolution/blob/main/proposals/0387-cross-compilation-destinations.md#swift-sdk-installation-and-configuration).

kateinoigakukun:
@MaxDesiatov Ok, I created a proposal amendment https://github.com/apple/swift-evolution/pull/2424

MaxDesiatov:
IIUC amendments go through the same process as the rest of Swift Evolution. What I mean is that we'd probably have to run this as a pitch first, schedule this for review with the Swift Evolution stakeholders if there's enough positive feedback on the pitch, then run a formal review and follow on the review decision after that.

kateinoigakukun:
Given that this is a non-functional and just a naming change, I feel going through the evolution process is too much for this change TBH... I'd like to hear @bnbarham's opinion here.

bnbarham:
> Given that this is a non-functional and just a naming change, I feel going through the evolution process is too much for this change TBH... But I can follow the process if it's really necessary. I'd like to hear @bnbarham's opinion here.

It'd be worth posting in `Evolution | Discussions` and seeing if the core team feels a review is required. Interestingly the proposal actually has `configure` and not `configuration`, so either way we probably want to do *something* here.

PR #7512, opened 2024-04-29
Bring back `experimental-sdk` as deprecated for compatibility

MaxDesiatov:
There are enough users of this command and corresponding options on `swift build` in the wild. These should be deprecated first before removing.


MaxDesiatov:
@swift-ci test

Package.swift
```diff
@@ -533,6 +533,12 @@ let package = Package(
             dependencies: ["Commands", "SwiftSDKCommand"],
             exclude: ["CMakeLists.txt"]
         ),
+        .executableTarget(
```
bnbarham, line 536: Unfortunate we need to keep the executable here :(

bnbarham (approved):

PR #7519, opened 2024-05-01
Fix incorrect paths for library products with explicit linkage

MaxDesiatov:
Library products declared with `.static` or `.dynamic` linkage didn't have a `-tools` suffix in their file names when built as dependencies of macros. This change resolves the inconsistency and adds corresponding build plan tests.

The regression in question broke the compatibility test suite, which had [an old version of `swift-power-assert`](https://github.com/kishikawakatsumi/swift-power-assert/blob/a60cb50/Package.swift) depending on an old version of `swift-syntax` with explicit `.static` linkage for its products.

rdar://127170225

### DIFF
--- a/Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
+++ b/Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
@@ -243,7 +243,7 @@ public struct BuildParameters: Encodable {
 
     /// Returns the path to the dynamic library of a product for the current build parameters.
     func potentialDynamicLibraryPath(for product: ResolvedProduct) throws -> RelativePath {
-        try RelativePath(validating: "\(self.triple.dynamicLibraryPrefix)\(product.name)\(self.triple.dynamicLibraryExtension)")
+        try RelativePath(validating: "\(self.triple.dynamicLibraryPrefix)\(product.name)\(self.suffix(triple: product.buildTriple))\(self.triple.dynamicLibraryExtension)")
     }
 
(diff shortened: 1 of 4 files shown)


MaxDesiatov:
@swift-ci test

Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
```diff
@@ -243,7 +243,7 @@ public struct BuildParameters: Encodable {
 
     /// Returns the path to the dynamic library of a product for the current build parameters.
     func potentialDynamicLibraryPath(for product: ResolvedProduct) throws -> RelativePath {
-        try RelativePath(validating: "\(self.triple.dynamicLibraryPrefix)\(product.name)\(self.triple.dynamicLibraryExtension)")
+        try RelativePath(validating: "\(self.triple.dynamicLibraryPrefix)\(product.name)\(self.suffix(triple: product.buildTriple))\(self.triple.dynamicLibraryExtension)")
```
bnbarham, line 246: How likely is it that we're missing other suffix's 😅? Is there any decent way to look for all of them?

bnbarham (approved):

Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
MaxDesiatov: Given that we're constructing these paths in many places in an ad-hoc way, it is still somewhat likely. I'll do an audit of this code and also generalize path handling to avoid ad-hocness in a follow-up PR not to stall this one.

PR #7523, opened 2024-05-02
[ModulesGraph] Fix association between `.tools` targets and their pac…

xedin:
…kages

While inserting updated `.tools` targets to `modulesToPackages` the logic 
should use the package where the target resides instead of the package where it's referenced.

Resolves: rdar://127369576

### DIFF
--- a/Sources/PackageGraph/ModulesGraph.swift
+++ b/Sources/PackageGraph/ModulesGraph.swift
@@ -207,7 +207,8 @@ public struct ModulesGraph {
                         switch dependency {
                         case .target(let targetDependency, _):
                             allTargets.insert(targetDependency)
-                            modulesToPackages[targetDependency.id] = package
+                            modulesToPackages[targetDependency.id] =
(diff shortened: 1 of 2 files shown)


xedin:
@swift-ci please test

Sources/PackageGraph/ModulesGraph.swift
```diff
@@ -207,7 +207,8 @@ public struct ModulesGraph {
                         switch dependency {
                         case .target(let targetDependency, _):
                             allTargets.insert(targetDependency)
-                            modulesToPackages[targetDependency.id] = package
+                            modulesToPackages[targetDependency.id] =
+                                identitiesToPackages[targetDependency.packageIdentity]
```
xedin, line 211: If we change `ResolvedModule` and `ResolvedProduct` to reference `ResolvedPackage` instead of `PackageIdentifier` that should make `modulesToPackages` and `productsToPackages` obsolete...

MaxDesiatov (approved):
Thank you!

PR #7525, opened 2024-05-03
package -> @_spi public due to package exportability rule updates

elsh:
The exportability rules around `package` access level are updated in a way that an `@_spi public` can't be bound to a `package` var, similar to how `@_spi public` can't be bound to a `public` var.  For reference, see https://github.com/apple/swift/pull/73161. 
This PR corrects such use case.


elsh:
@swift-ci smoke test

bnbarham (approved):

elsh:
@swift-ci smoke test macOS

MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):

PR #7524, opened 2024-05-02
Remove the addition of external-plugin-path for dev toolchains

bnbarham:
SDK macro plugins are now provided within
`Platforms/<Platform>.platform/Developer/usr/lib/swift/host/plugins` now, which is always added by the driver.

PR #7526, opened 2024-05-03
[6.0] Remove the addition of external-plugin-path for dev toolchains

bnbarham:
**Explanation**:  SDK macro plugins are now provided within `Platforms` and are added by default. There's no need to add an additional path to the toolchain. Doing so prevents the macros in the dev toolchain from being picked up.
**Scope**: Macros
**Risk**: Low, this path was added prior to macros moving into `Platforms`.
**Testing**: Source compat suite passes (which has various macro projects). We could theoretically test we're not adding the argument, but seems a bit weird to do so.
**Original PR**: https://github.com/apple/swift-package-manager/pull/7524
**Reviewer**: @MaxDesiatov

### DIFF
--- a/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
@@ -439,25 +439,6 @@ package final class SwiftTargetBuildDescription {
         }
         #endif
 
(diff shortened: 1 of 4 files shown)


bnbarham:
@swift-ci please test

MaxDesiatov (approved):

PR #7469, opened 2024-04-18
Update -swift-version to default to 6 for swift-tools-version 6

bnbarham:
We missed updating this when adding 6.0. The change itself is extremely
simple, with the bulk being updates to our tests so they don't fail on
old compiler versions (where language version "6" is unknown).

Also removes disabling the concurrency and string processing imports
from manifest loading, as this was causing the `package` global to not
be main actor isolated by default and thus error in Swift 6. They seemed
to have been disabled to avoid erroring in SDKs that didn't have them,
but that is definitely not an issue any longer.

PR #7439, opened 2024-04-03
Implement per target swift version selection

xedin:
[Swift Forums pitch thread](https://forums.swift.org/t/pitch-swiftpm-swift-language-version-per-target/71067)

Add a new Swift target setting API, similar to `enable{Upcoming, Experimental}Feature`,
to specify a Swift language version that should be used to build the target, if such
version is not specified, fallback to the current language version determination logic.

Resolves: rdar://125732014

PR #7531, opened 2024-05-07
[6.0] Update -swift-version to default to 6 for swift-tools-version 6

bnbarham:
**Explanation**: Changes packages to default to swift version 6 when `swift-tools-version` is set `>= 6.0`. Without this change, all packages will be swift version 5 unless otherwise specified via `swiftLanguageVersions`. Packages default to `swift-tools-version: 6.0` with `swift package init`. Users can opt-out via specifying `swiftLanguageVersions: [.v5]`.
**Scope**: New packages/those updated to 6.0.
**Risk**: Medium. There's inherent risk due to the language mode bump. The change also removes disabling implicit imports for concurrency and string processing in manifest files, which hasn't been widely tested outside of ...
**Testing**: New regression tests. Source compatibility suite run.
**Original PR**: https://github.com/apple/swift-package-manager/pull/7469

### DIFF
--- a/Fixtures/Miscellaneous/Plugins/DependentPlugins/Package.swift
+++ b/Fixtures/Miscellaneous/Plugins/DependentPlugins/Package.swift
@@ -10,7 +10,7 @@ let package = Package(
     targets: [
         .executableTarget(name: "MyExecutable"),
         .executableTarget(name: "MyExecutable2"),
-        
+
         .plugin(
             name: "MyPlugin",
             capability: .buildTool(),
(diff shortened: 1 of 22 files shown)


bnbarham:
@swift-ci please test

bnbarham:
> Source compatibility suite run.

Technically it's currently running (though I would be very surprised if it failed with anything related to this PR).

bnbarham:
Need to cherry-pick a85643e462758d4c271270dbd4ef36eaaa7b5c2d (part of https://github.com/apple/swift-package-manager/pull/7439), looks like we squash commits though so that's more annoying than it otherwise would be :(

bnbarham:
@swift-ci please test

nkcsgexi (approved):

PR #7532, opened 2024-05-07
`swift sdk`: print warnings on stderr instead of stdout

MaxDesiatov:
Tools and scripts may rely on command output to remain parseable, so printing these warnings on `stdout` can break such workflows. We should output these warnings on `stderr` instead.

As these warnings are emitted before any logging is available, we're using `fputs` for output on `stderr`


MaxDesiatov:
@swift-ci test

bnbarham (approved):

PR #7533, opened 2024-05-07
[6.0] `swift sdk`: print warnings on stderr instead of stdout

MaxDesiatov:
Cherry-pick of #7532.

**Explanation**: Tools and scripts may rely on command output to remain parseable, so printing these warnings on `stdout` can break such workflows. We should output these warnings on `stderr` instead.

As these warnings are emitted before any logging is available, we're using `fputs` for output on `stderr`

**Scope**: Isolated to `swift sdk` command and subcommands.
**Risk**: Low, the actual change is only 4 lines touching `print` and `import`, code coverage for those is present.
**Testing**: End-to-end test cases present in `Tests/CommandsTests/SDKCommandTests.swift`
**Issue**: N/A
**Reviewer**: @bnbarham

### DIFF
--- a/Sources/swift-experimental-sdk/Entrypoint.swift
+++ b/Sources/swift-experimental-sdk/Entrypoint.swift
@@ -11,11 +11,12 @@
 //===----------------------------------------------------------------------===//
 
 import SwiftSDKCommand
+import Foundation
 
 @main
 struct Entrypoint {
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
@swift-ci test

bnbarham (approved):

PR #7540, opened 2024-05-08
Add API for background preparation in SourceKit-LSP

ahoppen:
- Add an API to get the name of a target from SourceKit-LSP 
  - We will need this API to prepare a target for indexing.
- Add an API to get all the targets in a build description / modules graph in a topological order
  - We need this for background preparation for indexing so that we can prepare and index lower-level targets first.


ahoppen:
@swift-ci Please test

MaxDesiatov (approved):

PR #7530, opened 2024-05-06
[Package/ModuleGraph] Allow cyclic package dependencies if they don't introduce a cycle in a build graph

xedin:
### Motivation:

It should be possible for packages to depend on each other if such dependence doesn't introduce cycles in the build graph.

### Modifications:

- Introduced a new DFS method to walk over graphs that breaks cycles.
- Replaces use of `findCycle` + `topologicalSort` with `DFS` while building manifest and package graphs. This allows cycles in dependencies to be modeled correctly.
- Removes some of the redundant data transformations from modules graph.
- Modifies `ResolvedPackage` to carry identities of its dependencies instead of resolved dependencies themselves. This helps to simplify logic in `createResolvedPackages`.
- Adds detection of target cycles across packages.

### Result:

Makes it possible for package A to depend on package B and B to depend on A if their targets don't form a cycle.

PR #7541, opened 2024-05-08
[6.0][Package/ModuleGraph] Allow cyclic package dependencies if they don't introduce a cycle in a build graph

xedin:
- Explanation:

   It should be possible for packages to depend on each other if such
    dependence doesn't introduce cycles in the build graph.

    - Introduced a new DFS method to walk over graphs that breaks cycles.
    - Replaces use of `findCycle` + `topologicalSort` with `DFS` while
    building manifest and package graphs. This allows cycles in dependencies
    to be modeled correctly.
    - Removes some of the redundant data transformations from modules graph.
    - Modifies `ResolvedPackage` to carry identities of its dependencies
    instead of resolved dependencies themselves. This helps to simplify
    logic in `createResolvedPackages`.
    - Adds detection of target cycles across packages.

    Makes it possible for package A to depend on package B and B to depend
    on A if their targets don't form a cycle.

- Scope: Package graph

- Main Branch PRs: https://github.com/apple/swift-package-manager/pull/7530

- Risk: Low

- Reviewed By: @MaxDesiatov    

- Testing: Added new test-cases to the test suite

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Note: This is in reverse chronological order, so newer entries are added to the 
 Swift 6.0
 -----------
 
+* [#7530]
+
+  Makes it possible for packages to depend on each other if such dependency doesn't form any target-level cycles. For example,
+  package `A` can depend on `B` and `B` on `A` unless targets in `B` depend on products of `A` that depend on some of the same
+  targets from `B` and vice versa.
+
 * [#7507] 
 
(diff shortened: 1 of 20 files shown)


xedin:
https://github.com/apple/sourcekit-lsp/pull/1248
@swift-ci please test

bnbarham (approved):

xedin:
https://github.com/apple/sourcekit-lsp/pull/1248
@swift-ci please test

PR #7549, opened 2024-05-10
plugins: Global compiler flags should not be used when building plugi…

euanh:
…n dependencies

Build tools and end products are built with separate toolchains:
* Plugins and any targets they depend on, such as helper tools, are built with the host toolchain.
* End products are built with the target toolchain.

Compiler flags set with -Xswiftc are intended to be applied only to end product builds, but they are incorrectly being applied to
plugin tool builds as well.   This can cause problems because target
toolchain flags might not be suitable for a tools build.   In the
worst case, the build might fail completely if the host toolchain
rejects the target toolchain flag.    For example, Linux toochains
accept the `-static-executable` flag, but macOS toolchains reject
it.

Fixes: rdar://127813618

### Motivation:

### Modifications:

### Result:


euanh:
@swift-ci test

euanh:
Test to follow.

euanh:
@swift-ci test

MaxDesiatov (approved):

Sources/CoreCommands/SwiftCommandState.swift
```diff
@@ -741,7 +741,7 @@ package final class SwiftCommandState {
             configuration: options.build.configuration,
             toolchain: toolchain,
             triple: triple,
-            flags: options.build.buildFlags,
+            flags: toolsBuild ? .init() : options.build.buildFlags,
```
xedin, line 744: `buildFlags` is a combination of C/C++/Swift/Linker/XCBuild which means that host compiler doesn't get any of them but are there situations when some of the flags intended for the host and/or tools? 

At the very list I think descriptions of all flags and not just `swiftCompilerFlags` have to be updated to indicate that they are no longer applied to the host.

Sources/CoreCommands/SwiftCommandState.swift
```diff
@@ -796,7 +796,7 @@ package final class SwiftCommandState {
 
     private lazy var _toolsBuildParameters: Result<BuildParameters, Swift.Error> = {
         Result(catching: {
-            try _buildParams(toolchain: self.getHostToolchain())
+            try _buildParams(toolchain: self.getHostToolchain(), toolsBuild: true)
```
xedin, line 799: Maybe instead of a boolean flag here we should just pass in the `options.build.buildFlags` or `[]`?

euanh: Yes, I think that's better and will help when we tidy up the tools/products build option split.

euanh:
@swift-ci test

PR #7555, opened 2024-05-13
Expose the build triple of a `BuildTarget` in `SourceKitLSPAPI`

ahoppen:
If we have a package like this

```swift
let package = Package(
  name: "MyLibrary",
  targets: [
   .target(name: "Lib"),
   .executableTarget(name: "MyExec", dependencies: ["Lib"]),
   .plugin(
     name: "MyPlugin",
     capability: .command(
       intent: .sourceCodeFormatting(),
       permissions: []
     ),
     dependencies: ["MyExec"]
   )
  ]
)
```

Then the package graph contains two targets for `Lib`. We need to be able to differentiate them in SourceKit-LSP. For that, we need to expose the build triple of a `BuildTarget`.

### DIFF
--- a/Sources/SourceKitLSPAPI/BuildDescription.swift
+++ b/Sources/SourceKitLSPAPI/BuildDescription.swift
@@ -23,12 +23,17 @@ import class Build.ClangTargetBuildDescription
 import class Build.SwiftTargetBuildDescription
 import struct PackageGraph.ResolvedModule
 import struct PackageGraph.ModulesGraph
+import enum PackageGraph.BuildTriple
+
+public typealias BuildTriple = PackageGraph.BuildTriple
 
 public protocol BuildTarget {
     var sources: [URL] { get }
 
     /// The name of the target. It should be possible to build a target by passing this name to `swift build --target`
     var name: String { get }
+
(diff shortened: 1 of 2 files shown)


Sources/SourceKitLSPAPI/BuildDescription.swift
```diff
@@ -23,13 +23,16 @@ import class Build.ClangTargetBuildDescription
 import class Build.SwiftTargetBuildDescription
 import struct PackageGraph.ResolvedModule
 import struct PackageGraph.ModulesGraph
+import enum PackageGraph.BuildTriple
```
ahoppen, line 26: @MaxDesiatov Do you prefer to expose the `BuildTriple` type or would you prefer to define an equivalent `BuildTriple` type in `SourceKitLSPAPI` and then map between the two?

MaxDesiatov: For now IMO it's best to create `public typealias BuildTriple = PackageGraph.BuildTriple` in `SourceKitLSPAPI`, which can be replaced with something more advanced later.

ahoppen:
@swift-ci Please test

PR #7557, opened 2024-05-14
[Package/Build] Remove `SwiftTarget.toolsSwiftVersion` in favor of a build setting

xedin:
Follow-up to https://github.com/apple/swift-package-manager/pull/7544 and associated discussion related to how tools version doesn't mean what it used to with introduction of per-target swift version settings.

### Motivation:

Unify handling of swift language version setting to make it harder for users to make mistakes when selected one.

### Modifications:

- Add a way to specify a default setting assignment

   This setting is going to be used if there are no non-default
   assignments that match build environment conditions.

- Remove `toolsSwiftVersion` from SwiftTarget in favor of a build setting

    Start using "default" assignments to represent tools version
    based swift language version setting that is going to be produced
    by `BuildSettings.Scope` if no other assignments match build
    environment conditions.

    This removes ambiguity from `-swift-version` selection for
    downstream clients.

- [PIF] Re-implement swift language version handling

    Since `toolSwiftVersion` is gone and swift language version is
    now handled by the build settings we need to adjust `PIFBuilder`
    to handle that properly.

    PIF build settings can handle per-target platform specific settings
    which are now expressible in SwiftPM via `.swiftLanguageVersion`
    build setting, handling of such settings is implemented in this commit

### Result:

Swift language version is handled uniformly as a build setting.


neonichu (approved):

xedin:
@swift-ci please test

PR #7553, opened 2024-05-13
Add two `Sendable` annotations to enable building SourceKit-LSP in Swift 6 mode

ahoppen:
- Mark `FileRuleDescription` as `Sendable`
- Make `Platform` sendable and make `Platform.current` a constant

PR #7559, opened 2024-05-14
Add two Sendable annotations to enable building SourceKit-LSP in Swift 6 mode

ahoppen:
* **Explanation**: Mark `FileRuleDescription` and `Platform` as `Sendable` and make `Platform.current` a constant. This allows us to build SourceKit-LSP in Swift 6 mode
* **Scope**: `FileRuleDescription` and `Platform` if SwiftPM is used as a library
* **Risk**: Very low, adding `Sendable` conformances is safe and I don’t expect anyone to modify `Platform.current`
* **Testing**: Verified that this helps us build SourceKit-LSP in Swift 6 mode
* **Issue**: n/a
* **Reviewer**:  @xedin on https://github.com/apple/swift-package-manager/pull/7553

### DIFF
--- a/Sources/PackageLoading/Platform.swift
+++ b/Sources/PackageLoading/Platform.swift
@@ -20,23 +20,21 @@ private func isAndroid() -> Bool {
         (try? localFileSystem.isFile(AbsolutePath(validating: "/system/bin/toybox"))) ?? false
 }
 
-public enum Platform: Equatable {
+public enum Platform: Equatable, Sendable {
     case android
     case darwin
     case linux(LinuxFlavor)
     case windows
 
     /// Recognized flavors of linux.
-    public enum LinuxFlavor: Equatable {
+    public enum LinuxFlavor: Equatable, Sendable {
         case debian
         case fedora
     }
 }
 
 extension Platform {
-    // This is not just a computed property because the ToolchainRegistryTests
-    // change the value.
-    public static var current: Platform? = {
+    public static let current: Platform? = {
         #if os(Windows)
         return .windows
         #else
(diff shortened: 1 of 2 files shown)


Sources/PackageLoading/Platform.swift
```diff
@@ -20,23 +20,21 @@ private func isAndroid() -> Bool {
...
-    public enum LinuxFlavor: Equatable {
+    public enum LinuxFlavor: Equatable, Sendable {
         case debian
         case fedora
     }
 }
 
 extension Platform {
-    // This is not just a computed property because the ToolchainRegistryTests
-    // change the value.
```
bnbarham, line 38: I assume this comment wasn't true 😅? If that's the case though, can it just be a computed property?

bnbarham (approved):

Sources/PackageLoading/Platform.swift
ahoppen: It was, `ToolchainRegistryTests` is in SourceKit-LSP: https://github.com/apple/sourcekit-lsp/pull/1276#discussion_r1597909530

I prefer it to be stored so we don’t re-run `uname` every time the variable is accessed.

Sources/PackageLoading/TargetSourcesBuilder.swift
```diff
@@ -563,7 +563,7 @@ public struct TargetSourcesBuilder {
 }
 
 /// Describes a rule for including a source or resource file in a target.
-public struct FileRuleDescription {
+public struct FileRuleDescription: Sendable {
```
giginet, line 566: Don't we need `FileRuleDescription.Rule` to conform to `Sendable`?

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Linux

PR #7573, opened 2024-05-17
Set an environment variable in `swift test` to indicate which testing library is in use.

grynspan:
This PR sets a new environment variable `"SWIFT_PM_TEST_LIBRARY"` to the name of the active testing library (`"XCTest"` or `"swift-testing"`.) Code in these libraries can then use the presence of this environment variable to adjust their behaviour.

Resolves rdar://128272138.

### DIFF
--- a/Tests/CommandsTests/TestCommandTests.swift
+++ b/Tests/CommandsTests/TestCommandTests.swift
@@ -304,12 +304,9 @@ final class TestCommandTests: CommandsTestCase {
     }
 #endif
 
-#if false
-    // DISABLED: This test cannot be enabled until the hosting copy of SwiftPM
-    // has started setting this environment variable.
-    func testLibraryEnvironmentVariable() {
-        let envvar = ProcessInfo.processInfo.environment["SWIFT_PM_TEST_LIBRARY"]
-        XCTAssertEqual(envvar, "XCTest")
+    func testLibraryEnvironmentVariable() throws {
+      try fixture(name: "Miscellaneous/CheckTestLibraryEnvironmentVariable") { fixturePath in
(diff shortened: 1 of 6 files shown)


stmontgomery (approved):

dennisweissmann (approved):

bnbarham (approved):

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

Tests/CommandsTests/TestCommandTests.swift
```diff
@@ -303,4 +303,13 @@ final class TestCommandTests: CommandsTestCase {
         }
     }
 #endif
+
+#if false
```
bnbarham, line 307: Not sure there's much point to this test. We should be able to mock out the process runner and then confirm the environment is what we expect that way (when running the test command in here). In the meantime though, I'd prefer XCTSkip here rather than a `#if false`

grynspan: Yeah that's fair, I can do that.

I'm not using `XCTSkip` because it makes the rest of the code unreachable and produces a warning diagnostic.

grynspan: Fixed the test. :)

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

Tests/CommandsTests/TestCommandTests.swift
bnbarham: Thanks!

PR #7574, opened 2024-05-17
Add prepare for index experimental build argument

dschaefer2:
This will be used by sourcekit-lsp to build the swiftmodule files it needs for indexing.

Adds experimental-prepare-for-indexing flag to swift build. Creates a llbuild manifest specific for the prepare build that skips generating object files and linking of those files and calls swiftc to only create the swiftmodule as quickly as it can.

### Motivation:

To support background indexing in sourcekit-lsp, it will request a prepare for index build to build the swiftmodule files it needs to do indexing. This build should be minimal to ensure indexing is fast so it can respond to language server requests that require the index as soon as possible.

### Modifications:

- Add an experimental-prepare-for-indexing flag to the BuildOptions and pass it around to every that needs it.
- Build a custom llbuild manifest so that only the commands necessary are added to the prepare build
- Add a target property that also ensures tool builds required for the prepare build are performed as usual.
- In SwiftTargetBuildDescription, pass compile options that put the swift compiler in prepare "mode".
- Ensure object files and binaries are not generated when in prepare mode.

### Result:

Implements prepare build mode without affecting regular build mode.

### DIFF
--- a/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
@@ -246,6 +246,12 @@ public final class SwiftTargetBuildDescription {
     /// Whether to disable sandboxing (e.g. for macros).
     private let shouldDisableSandbox: Bool
 
+    /// Whether to build preparing for indexing
+    public var prepareForIndexing: Bool {
+        // Do full build for tools
+        defaultBuildParameters.prepareForIndexing && target.buildTriple != .tools
+    }
+
     /// Create a new target description with target and build parameters.
     init(
         package: ResolvedPackage,
@@ -580,6 +586,14 @@ public final class SwiftTargetBuildDescription {
         // way.
         if self.defaultBuildParameters.driverParameters.enableParseableModuleInterfaces || args.contains("-enable-library-evolution") {
             args += ["-emit-module-interface-path", self.parseableModuleInterfaceOutputPath.pathString]
+        }
+
+        if self.prepareForIndexing {
+            args += [
+                "-Xfrontend", "-experimental-skip-all-function-bodies",
+                "-Xfrontend", "-experimental-allow-module-with-compiler-errors",
+                "-Xfrontend", "-empty-abi-descriptor"
+            ]
         }
 
         args += self.defaultBuildParameters.toolchain.extraFlags.swiftCompilerFlags
--- a/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
@@ -246,12 +246,6 @@ public final class SwiftTargetBuildDescription {
     /// Whether to disable sandboxing (e.g. for macros).
     private let shouldDisableSandbox: Bool
 
-    /// Whether to build preparing for indexing
-    public var prepareForIndexing: Bool {
-        // Do full build for tools
-        defaultBuildParameters.prepareForIndexing && target.buildTriple != .tools
-    }
-
     /// Create a new target description with target and build parameters.
     init(
         package: ResolvedPackage,
@@ -588,9 +582,12 @@ public final class SwiftTargetBuildDescription {
             args += ["-emit-module-interface-path", self.parseableModuleInterfaceOutputPath.pathString]
         }
 
-        if self.prepareForIndexing {
+        if self.defaultBuildParameters.prepareForIndexing {
             args += [
+                "-Xfrontend", "-enable-library-evolution",
                 "-Xfrontend", "-experimental-skip-all-function-bodies",
+                "-Xfrontend", "-experimental-lazy-typecheck",
+                "-Xfrontend", "-experimental-skip-non-exportable-decls",
                 "-Xfrontend", "-experimental-allow-module-with-compiler-errors",
                 "-Xfrontend", "-empty-abi-descriptor"
             ]
--- a/Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
+++ b/Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
@@ -537,7 +537,7 @@ extension LLBuildManifestBuilder {
 
     private func addModuleWrapCmd(_ target: SwiftTargetBuildDescription) throws {
         // Add commands to perform the module wrapping Swift modules when debugging strategy is `modulewrap`.
-        guard target.defaultBuildParameters.debuggingStrategy == .modulewrap else { return }
+        guard target.defaultBuildParameters.debuggingStrategy == .modulewrap, !target.prepareForIndexing else { return }
         var moduleWrapArgs = [
             target.defaultBuildParameters.toolchain.swiftCompilerPath.pathString,
             "-modulewrap", target.moduleOutputPath.pathString,
--- a/Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
+++ b/Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
@@ -41,15 +41,9 @@ extension LLBuildManifestBuilder {
         let inputs = try self.computeSwiftCompileCmdInputs(target)
 
         // Outputs.
-        let objectNodes = try target.objects.map(Node.file)
+        let objectNodes = target.defaultBuildParameters.prepareForIndexing ? [] : try target.objects.map(Node.file)
         let moduleNode = Node.file(target.moduleOutputPath)
-        let cmdOutputs: [Node]
-        if target.defaultBuildParameters.prepareForIndexing {
-            // Don't include the object nodes on prepare builds
-            cmdOutputs = [moduleNode]
-        } else {
-            cmdOutputs = objectNodes + [moduleNode]
-        }
+        let cmdOutputs = objectNodes + [moduleNode]
 
         if target.defaultBuildParameters.driverParameters.useIntegratedSwiftDriver {
             try self.addSwiftCmdsViaIntegratedDriver(
--- a/Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
+++ b/Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
@@ -43,7 +43,13 @@ extension LLBuildManifestBuilder {
         // Outputs.
         let objectNodes = try target.objects.map(Node.file)
         let moduleNode = Node.file(target.moduleOutputPath)
-        let cmdOutputs = objectNodes + [moduleNode]
+        let cmdOutputs: [Node]
+        if target.prepareForIndexing {
+            // Don't include the object nodes on prepare builds
+            cmdOutputs = [moduleNode]
+        } else {
+            cmdOutputs = objectNodes + [moduleNode]
+        }
 
         if target.defaultBuildParameters.driverParameters.useIntegratedSwiftDriver {
             try self.addSwiftCmdsViaIntegratedDriver(
@@ -397,7 +403,8 @@ extension LLBuildManifestBuilder {
             fileList: target.sourcesFileListPath,
             isLibrary: isLibrary,
             wholeModuleOptimization: target.defaultBuildParameters.configuration == .release,
-            outputFileMapPath: try target.writeOutputFileMap() // FIXME: Eliminate side effect.
+            outputFileMapPath: try target.writeOutputFileMap(), // FIXME: Eliminate side effect.
+            prepareForIndexing: target.prepareForIndexing
         )
     }
 
@@ -417,6 +424,8 @@ extension LLBuildManifestBuilder {
             inputs.append(resourcesNode)
         }
 
+        let prepareForIndexing = target.prepareForIndexing
+
         func addStaticTargetInputs(_ target: ResolvedModule) throws {
             // Ignore C Modules.
             if target.underlying is SystemLibraryTarget { return }
@@ -428,7 +437,7 @@ extension LLBuildManifestBuilder {
             if target.underlying is ProvidedLibraryTarget { return }
 
             // Depend on the binary for executable targets.
-            if target.type == .executable {
+            if target.type == .executable && !prepareForIndexing {
                 // FIXME: Optimize.
                 let product = try plan.graph.allProducts.first {
                     try $0.type == .executable && $0.executableTarget.id == target.id
@@ -446,8 +455,15 @@ extension LLBuildManifestBuilder {
             case .swift(let target)?:
                 inputs.append(file: target.moduleOutputPath)
             case .clang(let target)?:
-                for object in try target.objects {
-                    inputs.append(file: object)
+                if prepareForIndexing {
+                    // In preparation, we're only building swiftmodules, need to depend on sources
+                    for source in target.clangTarget.sources.paths {
+                        inputs.append(file: source)
+                    }
+                } else {
+                    for object in try target.objects {
+                        inputs.append(file: object)
+                    }
                 }
             case nil:
                 throw InternalError("unexpected: target \(target) not in target map \(self.plan.targetMap)")
--- a/Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
+++ b/Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
@@ -44,7 +44,7 @@ extension LLBuildManifestBuilder {
         let objectNodes = try target.objects.map(Node.file)
         let moduleNode = Node.file(target.moduleOutputPath)
         let cmdOutputs: [Node]
-        if target.prepareForIndexing {
+        if target.defaultBuildParameters.prepareForIndexing {
             // Don't include the object nodes on prepare builds
             cmdOutputs = [moduleNode]
         } else {
@@ -404,7 +404,7 @@ extension LLBuildManifestBuilder {
             isLibrary: isLibrary,
             wholeModuleOptimization: target.defaultBuildParameters.configuration == .release,
             outputFileMapPath: try target.writeOutputFileMap(), // FIXME: Eliminate side effect.
-            prepareForIndexing: target.prepareForIndexing
+            prepareForIndexing: target.defaultBuildParameters.prepareForIndexing
         )
     }
 
@@ -424,7 +424,7 @@ extension LLBuildManifestBuilder {
             inputs.append(resourcesNode)
         }
 
-        let prepareForIndexing = target.prepareForIndexing
+        let prepareForIndexing = target.defaultBuildParameters.prepareForIndexing
 
         func addStaticTargetInputs(_ target: ResolvedModule) throws {
             // Ignore C Modules.
@@ -456,9 +456,10 @@ extension LLBuildManifestBuilder {
                 inputs.append(file: target.moduleOutputPath)
             case .clang(let target)?:
                 if prepareForIndexing {
-                    // In preparation, we're only building swiftmodules, need to depend on sources
-                    for source in target.clangTarget.sources.paths {
-                        inputs.append(file: source)
+                    // In preparation, we're only building swiftmodules
+                    // propagate the dependency to the header files in this target
+                    for header in target.clangTarget.headers {
+                        inputs.append(file: header)
                     }
                 } else {
                     for object in try target.objects {
@@ -537,7 +538,7 @@ extension LLBuildManifestBuilder {
 
     private func addModuleWrapCmd(_ target: SwiftTargetBuildDescription) throws {
         // Add commands to perform the module wrapping Swift modules when debugging strategy is `modulewrap`.
-        guard target.defaultBuildParameters.debuggingStrategy == .modulewrap, !target.prepareForIndexing else { return }
+        guard target.defaultBuildParameters.debuggingStrategy == .modulewrap else { return }
         var moduleWrapArgs = [
             target.defaultBuildParameters.toolchain.swiftCompilerPath.pathString,
             "-modulewrap", target.moduleOutputPath.pathString,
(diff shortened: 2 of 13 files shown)


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
```diff
@@ -582,6 +588,14 @@ public final class SwiftTargetBuildDescription {
             args += ["-emit-module-interface-path", self.parseableModuleInterfaceOutputPath.pathString]
         }
 
+        if self.prepareForIndexing {
+            args += [
+                "-Xfrontend", "-experimental-skip-all-function-bodies",
+                "-Xfrontend", "-experimental-allow-module-with-compiler-errors",
+                "-Xfrontend", "-empty-abi-descriptor"
+            ]
```
bnbarham, line 596: We could also try out `-experimental-lazy-typecheck` and `-experimental-skip-non-exportable-decls` (sorry, meant to send these to you earlier). They're newer than the others, but should hopefully skip a reasonable amount of typechecking.

Tests/BuildTests/PrepareForIndexTests.swift
```diff
@@ -0,0 +1,109 @@
...
+            observabilityScope: scope
+        )
+
+        let builder = LLBuildManifestBuilder(plan, fileSystem: fs, observabilityScope: scope)
+        let manifest = try builder.generatePrepareManifest(at: "/manifest")
+
+        // Make sure we're still building swift modules
+        XCTAssertNotNil(manifest.commands["<SwiftSyntax-debug.module>"])
+        // Make sure we're not building things that link
+        XCTAssertNil(manifest.commands["C.Core-debug.exe"])
```
bnbarham, line 36: Should this be a per platform check?

Tests/BuildTests/PrepareForIndexTests.swift
```diff
@@ -0,0 +1,109 @@
...
+
+        // Make sure only object files for tools are built
+        let objectFiles = Set(outputs.filter({ $0.hasSuffix(".o") }))
+        XCTAssertEqual(objectFiles, Set([
+            "debug/MMIOMacros-tool.build/source.swift.o",
+            "debug/SwiftSyntax-tool.build/source.swift.o"
+        ]))
+
+        // Check diff with regular build plan
+        let plan0 = try BuildPlan(
```
bnbarham, line 84: Can we rename this `regularPlan` or something similar? Same thing with the vars below. `objectFiles0` could be `extraRegularObjects`.

bnbarham (commented):
Exciting!

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
dschaefer2: Will do. Currently just trying to get all the xplatform tests passing.

Tests/BuildTests/PrepareForIndexTests.swift
dschaefer2: Yup, fix incoming. :)

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
LOL, oh yeah, because the test is actually failing. The object files are getting created on Linux. Need to find out why.

dschaefer2:
@swift-ci Please test

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
```diff
@@ -43,7 +43,13 @@ extension LLBuildManifestBuilder {
         // Outputs.
         let objectNodes = try target.objects.map(Node.file)
         let moduleNode = Node.file(target.moduleOutputPath)
-        let cmdOutputs = objectNodes + [moduleNode]
+        let cmdOutputs: [Node]
+        if target.prepareForIndexing {
+            // Don't include the object nodes on prepare builds
+            cmdOutputs = [moduleNode]
+        } else {
+            cmdOutputs = objectNodes + [moduleNode]
```
xedin, line 51: Have you considered adjusting `objects` to produce `[]` if the target is in "preparation" mode?

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
```diff
@@ -246,6 +246,12 @@ public final class SwiftTargetBuildDescription {
     /// Whether to disable sandboxing (e.g. for macros).
     private let shouldDisableSandbox: Bool
 
+    /// Whether to build preparing for indexing
+    public var prepareForIndexing: Bool {
+        // Do full build for tools
+        defaultBuildParameters.prepareForIndexing && target.buildTriple != .tools
```
xedin, line 252: If the `target` triple ` is `tools` then `defaultBuildParameters` are going to be `toolsBuildParameters` which wouldn't have this flag anyway?

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
```diff
@@ -446,8 +455,15 @@ extension LLBuildManifestBuilder {
             case .swift(let target)?:
                 inputs.append(file: target.moduleOutputPath)
             case .clang(let target)?:
-                for object in try target.objects {
-                    inputs.append(file: object)
+                if prepareForIndexing {
```
xedin, line 458: Could you elaborate on why this is necessary? This is handling a clang target, supposedly preparation doesn't apply here at all and these object files are produced by other clang targets?

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
```diff
@@ -521,7 +537,7 @@ extension LLBuildManifestBuilder {
 
     private func addModuleWrapCmd(_ target: SwiftTargetBuildDescription) throws {
         // Add commands to perform the module wrapping Swift modules when debugging strategy is `modulewrap`.
-        guard target.defaultBuildParameters.debuggingStrategy == .modulewrap else { return }
+        guard target.defaultBuildParameters.debuggingStrategy == .modulewrap, !target.prepareForIndexing else { return }
```
xedin, line 540: I think this should be done inside of `debuggingStrategy` getter - check whether target is for indexing and produce `nil`.

Sources/Build/BuildManifest/LLBuildManifestBuilder.swift
```diff
@@ -127,6 +127,38 @@ public class LLBuildManifestBuilder {
...
+                // Need the clang targets for tools
+                if desc.target.buildTriple == .tools {
+                    try self.createClangCompileCommand(desc)
+                }
+            }
+        }
+
+        for (_, description) in self.plan.productMap {
+            // Need to generate macro products
+            if description.product.type == .macro {
```
xedin, line 153: What about plugins?

Sources/CoreCommands/Options.swift
```diff
@@ -437,6 +437,9 @@ public struct BuildOptions: ParsableArguments {
     @Flag(help: "Enable or disable indexing-while-building feature")
     public var indexStoreMode: StoreMode = .autoIndexStore
 
+    @Flag(name: .customLong("experimental-prepare-for-indexing"))
+    var prepareForIndexing: Bool = false
```
xedin, line 441: If we want this as an option could this be a broader one something `perform-experimental-minimal-build` or something like that which is not necessarily coupled with indexing so tightly but one of the use-cases is indexing?

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
dschaefer2: Good point. I've been slowing moving these checks closer to the source of the values. I'll do one more pass with the help of your comments to clean that up. Thanks!

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
dschaefer2: That is the purpose of the check for `.tools`. This check will always be false if the `buildTriple` is `.tools`. I can probably make that cleaner. I should also make sure `prepareForIndexing` is false in `toolsBuildParameters` though I never check that anywhere.

dschaefer2: Ah, I see what you are saying. I don't need the check if I make the `toolsBuildParameters.prepareForIndexing` set to false.

Sources/CoreCommands/Options.swift
dschaefer2: It's not really a build that can be used for anything other than indexing. The only output is .swiftmodules files.

xedin: I suspect there might be some other uses in the future?

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
dschaefer2: In the normal case this adds a dependency on the object files from the clang modules to force a rebuild when they change. In preparation, there are no object files generated so this was a haphazard attempt to propagate the dependencies to their sources. This should actually be the headers. Let me give this one a bit more thought.

Sources/CoreCommands/Options.swift
bnbarham: We also allow errors + skip a lot that would be somewhat unexpected in more general cases. TBH I find it hard to imagine this would actually be useful for other purposes, though perhaps "only generating swiftmodules" (without the extra flags) may be? CC @kubamracek since IIUC this is somewhat similar to embedded needs?

In an ideal world this wouldn't be a separate arena and could be part of the same build, where any module compiled in this way would be rebuilt during the build, but *not* rebuilt by "preparation" if it is already up to date. Not sure if that helps with naming at all though 🤔.

dschaefer2: That's a great point @bnbarham about reusing modules between regular build and this preparation build. Certainly something we can work towards and probably needs a revisit of the architecture. I am also curious how we integrated embedded into all this (independent of preparation).

At any rate, I'm not sure there is a concept of the experimental experimental options :), but I'm inclined to leave it like this for now. As we get the sourcekit-lsp side lined up with this we can see if anything more natural jumps out.

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
dschaefer2: Wow, those two chopped my previous prepare time in half for sourcekit-lsp which happened in 30 seconds versus 95 for the debug build.

I had to enable-library-evolution for those to work, but that should be fine?

Sources/CoreCommands/Options.swift
bnbarham: > I am also curious how we integrated embedded into all this (independent of preparation).

It isn't today, which is partially why I'm bringing it up - since maybe we should be splitting the "don't need objects" from "preparation". It's unlikely that would be a command line option in the embedded case, ie. we probably want some manifest changes. But maybe still useful to think about.

https://github.com/apple/swift/blob/main/docs/EmbeddedSwift/UserManual.md#libraries-and-modules-in-embedded-swift has a little more information there.

> Certainly something we can work towards and probably needs a revisit of the architecture.

Yeah, just to be clear, I was in no way suggesting this is what we should do now. One day :)

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
bnbarham: Hmmm, enabling library evolution would result in different diagnostics (eg. need to add `@unknown case` for non-frozen enums). @tshortli is library evolution a hard requirement for these?

> Wow, those two chopped my previous prepare time in half for sourcekit-lsp which happened in 30 seconds versus 95 for the debug build.

Just to be clear, 95s is the full debug build and a prepare *used* to be 60s without these and is now 30s?

Sources/CoreCommands/Options.swift
xedin: I'm fine leaving it as is today but I think it's important to consider whether there is a more general thing hiding here that would allow us to cut down on the flags...

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
xedin: Indeed.

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
dschaefer2: > Just to be clear, 95s is the full debug build and a prepare _used_ to be 60s without these and is now 30s?

Correct. For sourcekit-lsp. I've been bouncing around. I need to do a more complete benchmark on a few different projects to get a better picture.

Sources/CoreCommands/Options.swift
dschaefer2: @xedin Agreed. Certainly some of this may prove useful in the embedded case, e.g. skipping object files in the dependencies. But this mode feels like it's violating some of the architecture (like how it reaches deep down to remove the -c option). I think much more thought needs to be put into it to make it more generally applicable.

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
tshortli: > @tshortli is library evolution a hard requirement for these?

Yes, the speed-up with `-experimental-lazy-typecheck` and `-experimental-skip-non-exportable-decls` exploits the fact that much less of the source code is needed to produce a module description for a module with library evolution. The same is unfortunately not true for non-resilient modules.

tshortli: Although that restriction is about making sure the emitted module is suitable for use during compilation of downstream dependents (the ABI of a non-resilient module depends on `internal` and `private` declarations). Maybe in the indexing case it's not necessary to restrict these options to library evolution.

bnbarham: > Although that restriction is about making sure the emitted module is suitable for use during compilation of downstream dependents

Yeah, sounds like we could probably gate that on `-experimental-skip-all-function-bodies` since we don't produce a swiftmodule that would be usable for compilation in that case anyway.

dschaefer2:
@swift-ci please test

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
dschaefer2: Moved the check into the getter.

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
dschaefer2: Made sure the toolsBuildParameters are set properly so I don't need this getter at all any more.

Tests/BuildTests/PrepareForIndexTests.swift
dschaefer2: Actually gave up testing that. It wasn't adding much.

Tests/BuildTests/PrepareForIndexTests.swift
dschaefer2: Made sure tests work on Mac and Linux which forced me to be more platform agnostic.

bnbarham:

bnbarham:

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
tshortli: I've lifted the restriction so that `-enable-library-evolution` is no longer required when indexing:
https://github.com/apple/swift/pull/73905
https://github.com/apple/swift/pull/73905

bnbarham: Nice, thanks @tshortli!

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
domciiiiix: 👍🏻

dschaefer2:
@swift-ci please test

dschaefer2:

dschaefer2:

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
```diff
@@ -582,6 +582,17 @@ public final class SwiftTargetBuildDescription {
             args += ["-emit-module-interface-path", self.parseableModuleInterfaceOutputPath.pathString]
         }
 
+        if self.defaultBuildParameters.prepareForIndexing {
+            args += [
+                "-Xfrontend", "-enable-library-evolution",
```
MaxDesiatov, line 587: Is `-enable-library-evolution` needed here now?

Tests/BuildTests/PrepareForIndexTests.swift
```diff
@@ -0,0 +1,62 @@
+// Copyright (C) 2024 Apple Inc. All rights reserved.
+//
+// This document is the property of Apple Inc.
+// It is considered confidential and proprietary.
+//
+// This document may not be reproduced or transmitted in any form,
+// in whole or in part, without the express written permission of
+// Apple Inc.
```
MaxDesiatov, line 8: This header is wrong, it should be replaced with the Apache 2.0 header.

MaxDesiatov: I also recommend running `./scripts/soundness.sh` before pushing commits, which runs the formatter and also checks if these license headers are valid.

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
dschaefer2: After looking at the clang commands, I realized that the header dependencies are coming from the generated .d files. Since I'm not running the compiler they won't be there. So my original idea of adding the dependencies on the headers instead of the object files seems to be working.

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
dschaefer2: Was waiting for an update toolchain. There's one there now. Should be ready to remove.

Tests/BuildTests/PrepareForIndexTests.swift
dschaefer2: Thanks Max, yeah, I should probably remove that new file template too. :)

dschaefer2:
@ahoppen really wants me to checkout multi target prepare. I will look at that next.

bnbarham:
> @ahoppen really wants me to checkout multi target prepare. I will look at that next.

We can merge this without that. That's also something we should all discuss a bit, since it would be weird to have that for only preparation.

dschaefer2:
> > @ahoppen really wants me to checkout multi target prepare. I will look at that next.
> 
> We can merge this without that. That's also something we should all discuss a bit, since it would be weird to have that for only preparation.

Agreed. I'm not even sure I can make it work cleanly with the current architecture. I'd rather not rush it.

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
MaxDesiatov: Would you mind updating the diff accordingly before merging?

dschaefer2: Will do. That will be the last piece. One I can make sure it works 😀

Sources/CoreCommands/Options.swift
```diff
@@ -437,6 +437,9 @@ public struct BuildOptions: ParsableArguments {
     @Flag(help: "Enable or disable indexing-while-building feature")
     public var indexStoreMode: StoreMode = .autoIndexStore
 
+    @Flag(name: .customLong("experimental-prepare-for-indexing"))
```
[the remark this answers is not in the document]
dschaefer2: Agreed. This option isn't really useful to humans :).

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
dschaefer2: I am still getting 
```
<unknown>:0: warning: ignoring -experimental-skip-non-exportable-decls (requires -enable-library-evolution)
<unknown>:0: warning: ignoring -experimental-lazy-typecheck (requires -enable-library-evolution)
```
in the latest 6.0 toolchain from May 26. Not sure this change got in after that build?

bnbarham: Would have been just after:
https://github.com/apple/swift/pull/73906

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
ahoppen: If this is the only thing blocking the PR from being merged, I’d prefer to get the PR merged with `-enable-library-evolution` still set because it could unblock me to test preparation in sourcekit-lsp. I know that I’ll get spurious warnings because of library evolution but I can ignore those for now.

dschaefer2: Yeah. As mentioned above, I'm not even sure if the change that was made even fixes it. I have one more fix Ben B mentioned about hiding the flag and then I'll kick off the tests.

Sources/CoreCommands/Options.swift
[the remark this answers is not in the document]
dschaefer2: Done.

dschaefer2:
@swift-ci please test

dschaefer2:

bnbarham (approved):

PR #7577, opened 2024-05-17
[6.0] Set an environment variable in `swift test` to indicate which testing library is in use.

grynspan:
**Explanation**: Allows swift-testing and XCTest to detect at runtime which testing library SwiftPM is trying to run so that they can adjust their behaviour (e.g. `XCTestScaffold`) appropriately.
**Scope**: Test build products at runtime.
**Risk**: Low. No obvious risk here to adding a new environment variable nobody is looking for.
**Testing**: New unit test (tested at desk, but must be temporarily disabled in CI until CI is running a version of SwiftPM that includes the change.)
**Original PR**: https://github.com/apple/swift-package-manager/pull/7573

### DIFF
--- a/Sources/Commands/Utilities/TestingSupport.swift
+++ b/Sources/Commands/Utilities/TestingSupport.swift
@@ -116,7 +116,8 @@ enum TestingSupport {
                     experimentalTestOutput: experimentalTestOutput,
                     library: .xctest
                 ),
-                sanitizers: sanitizers
+                sanitizers: sanitizers,
+                library: .xctest
             )
 
             try TSCBasic.Process.checkNonZeroExit(arguments: args, environment: env)
@@ -131,7 +132,8 @@ enum TestingSupport {
                 shouldSkipBuilding: shouldSkipBuilding,
                 library: .xctest
             ),
-            sanitizers: sanitizers
+            sanitizers: sanitizers,
+            library: .xctest
         )
         args = [path.description, "--dump-tests-json"]
         let data = try Process.checkNonZeroExit(arguments: args, environment: env)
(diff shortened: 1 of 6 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci test Windows

grynspan:
@swift-ci please test Windows

bnbarham (approved):

grynspan:
Will cherry-pick over the unit test from the main branch PR on Monday.

grynspan:
Eh, I've got time now.

grynspan:
Wait… I already did the thing. _**MERGING**_.

Sources/Commands/Utilities/TestingSupport.swift
```diff
@@ -156,6 +159,10 @@ enum TestingSupport {
             env["NO_COLOR"] = "1"
         }
 
+        // Set an environment variable to indicate which library's test product
+        // is being executed.
+        env["SWIFT_PM_TEST_LIBRARY"] = String(describing: library)
```
jakepetroules, line 164: This is convenient for shared libraries which want to have support code that might be run by either an XCTest or SwiftTesting test case.

However, I suggest we name this more generally (for example, `APPLE_TEST_LIBRARY`) because it's not SwifPM-specific, and other build tools / IDEs might want to have a uniform way to identify the Apple-provided testing library in use.

I also expect we _wouldn't_ this environment variable if SwiftPM gained direct support for additional testing libraries from other platforms/environments, since those would likely be run via their own test harnesses which might define their own environment variables.

PR #7579, opened 2024-05-18
[PackageGraph] Allow package-level cyclic dependency only for >= 6.0 …

xedin:
…manifests

### Motivation:

Follow-up to https://github.com/apple/swift-package-manager/pull/7530

Otherwise it might be suprising for package authors to discover that their packages cannot be used with older tools because they inadvertently introduced a cyclic dependency in a new version.

### Modifications:

`ModulesGraph.load` has been adjusted to run `findCycle` some of the root manifests support tools version that is older than 6.0. New diagnostic was added to help guide package authors to update their versions if they have cyclic package-level dependencies.

### Result:

Attempts to introduce cyclic package-level dependencies to manifests that support tools-versions less than 6.0 would result in a failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@ Note: This is in reverse chronological order, so newer entries are added to the 
 
 * [#7530]
 
-  Makes it possible for packages to depend on each other if such dependency doesn't form any target-level cycles. For example,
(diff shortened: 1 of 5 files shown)


xedin:
@swift-ci please test

xedin:
@swift-ci please test

xedin:

bnbarham (approved):

PR #7591, opened 2024-05-24
Enable swift-testing by default.

grynspan:
This PR sets the default state of `--{enable,disable}-experimental-swift-testing` to enabled.

This PR is speculative.

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -593,42 +593,8 @@ public struct TestLibraryOptions: ParsableArguments {
             return callerSuppliedValue
         }
 
-        // If the active package has a dependency on swift-testing, automatically enable support for it so that extra steps are not needed.
-        let workspace = try swiftCommandState.getActiveWorkspace()
-        let root = try swiftCommandState.getWorkspaceRoot()
-        let rootManifests = try temp_await {
-            workspace.loadRootManifests(
-                packages: root.packages,
-                observabilityScope: swiftCommandState.observabilityScope,
-                completion: $0
-            )
-        }
-
-        // Is swift-testing among the dependencies of the package being built?
-        // If so, enable support.
-        let isEnabledByDependency = rootManifests.values.lazy
-            .flatMap(\.dependencies)
-            .map(\.identity)
-            .map(String.init(describing:))
-            .contains("swift-testing")
-        if isEnabledByDependency {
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Jerry-Carter:
Hoping to see this change made soon.  Some general comments, mostly related to this issue.

The current behavior is rather unintuitive.  WWDC24 suggests that Swift Testing is now the preferred test framework, but these tests aren't actually executed unless the command line argument `--enable-swift-testing` is given.  The XCTest cases are always executed.  When only Swift Testing is used, the build processes ends with 'Executed 0 tests, with 0 failures (0 unexpected) in 0.000 (0.001) seconds' which is confusing and false.  At a minimum, may I suggest changing
```
Test run started  → Swift Test execution started
Test Suite 'All tests'  → XCTest Suite 'All tests'
Executed 0 tests, with 0 failures  → XCTest complete.  Executed 0 tests, with 0 failures
```
so that the stages are a bit clearer.

I understand from the [forum discussion](https://forums.swift.org/t/swift-testing-with-swiftpm-using-xcode-16-and-swift-6/72372/3) that the code is compiled, then the Swift Testing cases are executed (if the flag is specified), and then the XCTest cases are executed with the 'Executed 0 tests' message originating from the later.  However, it appears that plugins are re-executed between the Swift Testing and XCTest stages which (1) feels unnecessary and (2) breaks up the results from unit testing.

Finally, I appreciate that the unit tests are being performed in parallel but I wish the output from `Suite`s was grouped somehow rather than interwoven.  In my mind, Suites represent the fundamental conceptual units with the underlaying Test functions being supporting details.  I'd love to see something like

```
▢ Test run started.
↳ Testing Library Version: 75.9 (arm64e-apple-macos13.0)
▢ Suite "LongTests" started
      ▢ Test dataDiscombobulation() started.
      ☑️ Test reskewingWidgets() passed after 52.182 seconds.
 ☑️ Suite "ShortTests" passed after 0.390 seconds.
      ☑️ Test checkName() passed after 0.02 seconds.
      ☑️ Test sortOrder() passed after 0.389 seconds.
```
at runtime, but would also welcome an option to generate a final test report.

grynspan:
Thank you for your feedback, @Jerry-Carter. The current behaviour on macOS (where you need to specify the flag) is a known issue with Xcode 16's built-in Swift toolchain. We are working to resolve it.

grynspan:
Superseded by #7766.

PR #7604, opened 2024-05-29
Add arguments to Plugins

BarredEwe:
Adding arguments for running Plugins

### Motivation:

Now, in order to pass any parameters to plugins, it is necessary to use configuration files and read them from Plugin or create several plugins with predefined parameters. This is very inconvenient and time-consuming. This was mentioned during plugin development: https://github.com/apple/swift-evolution/blob/main/proposals/0303-swiftpm-extensible-build-tools.md#type-safe-options , but was not implemented. I propose to add a simple passing of arguments as an array of strings 

### Modifications:

// TODO

### Result:

// TODO


MaxDesiatov:
This would need a corresponding pitch and an evolution proposal to be reviewed and accepted before it could be considered for merging.

BarredEwe:
> This would need a corresponding pitch and an evolution proposal to be reviewed and accepted before it could be considered for merging, as it includes changes to the `PackageDescription` API.

I see, then I'll go make a discussion and write a proposal

PR #6583, opened 2023-05-18
Remove the version parameter for C++ interop mode build setting

egorzhdan:
### Motivation:

The version of C++ interoperability is going to be tied to Swift language version.

### Modifications:

This change removes the explicit version parameter for the `.interoperabilityMode(.Cxx)` build setting.

rdar://109525217

PR #7610, opened 2024-05-29
Add cxx interop support to symbolgraph-extract

rauhul:
In order to extract symbol graphs for modules with transitive imports on C++ modules we need parse headers in C++ interop mode using the option: `-cxx-interoperability-mode=default`. This commit updates `swift-symbolgraph-extract` to pass the option when there is a C++ module in the import graph.

This option is a new addition to `swift-symbolgraph-extract` added in apple/swift#73963.

Prior to this option's introduction, extracting the symbol graph for C++ dependent modules would fail with an error similar to: "unknown type name 'namespace'".

With this SwiftPM commit and `swift-symbolgraph-extract` _prior_ to apple/swift#73963, users will instead see an argument parsing error like: "unknown option -cxx-interoperability-mode".

With this change and `swift-symbolgraph-extract` change in apple/swift#73963, the symbol graph is extracted properly.

### DIFF
--- a/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
@@ -207,6 +207,20 @@ public final class ClangTargetBuildDescription {
         }
     }
 
+    /// Determines the arguments needed to run `swift-symbolgraph-extract` for
+    /// this module.
+    public func symbolGraphExtractArguments() throws -> [String] {
+        var args = [String]()
+
+        if self.clangTarget.isCXX {
+            args += ["-cxx-interoperability-mode=default"]
+        }
+        if let cxxLanguageStandard = self.clangTarget.cxxLanguageStandard {
+            args += ["-Xcc", "-std=\(cxxLanguageStandard)"]
+        }
+        return args
+    }
+
     /// Builds up basic compilation arguments for a source file in this target; these arguments may be different for C++
     /// vs non-C++.
     /// NOTE: The parameter to specify whether to get C++ semantics is currently optional, but this is only for revlock
--- a/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
@@ -547,26 +547,7 @@ public final class SwiftTargetBuildDescription {
             args += ["-color-diagnostics"]
         }
 
-        // If this is a generated test discovery target or a test entry point, it might import a test
-        // target that is built with C++ interop enabled. In that case, the test
-        // discovery target must enable C++ interop as well
-        switch testTargetRole {
-        case .discovery, .entryPoint:
-            for dependency in try self.target.recursiveTargetDependencies() {
-                let dependencyScope = self.buildParameters.createScope(for: dependency)
-                let dependencySwiftFlags = dependencyScope.evaluate(.OTHER_SWIFT_FLAGS)
-                if let interopModeFlag = dependencySwiftFlags.first(where: { $0.hasPrefix("-cxx-interoperability-mode=") }) {
-                    args += [interopModeFlag]
-                    if interopModeFlag != "-cxx-interoperability-mode=off" {
-                        if let cxxStandard = self.package.manifest.cxxLanguageStandard {
-                            args += ["-Xcc", "-std=\(cxxStandard)"]
-                        }
-                    }
-                    break
-                }
-            }
-        default: break
-        }
+        args += try self.cxxInteroperabilityModeArguments(allowDuplicate: false)
 
         // Add arguments from declared build settings.
         args += try self.buildSettingsFlags()
@@ -644,6 +625,67 @@ public final class SwiftTargetBuildDescription {
 
         return args
     }
+    
+    /// Determines the arguments needed to run `swift-symbolgraph-extract` for
+    /// this module.
+    public func symbolGraphExtractArguments() throws -> [String] {
+        var args = [String]()
+        args += try self.cxxInteroperabilityModeArguments(allowDuplicate: true)
+        return args
+    }
+
+    /// Determines the arguments needed for cxx interop for this module.
+    ///
+    /// If the current module or any of its linked dependencies requires cxx
+    /// interop, cxx interop will be enabled on the current module.
+    func cxxInteroperabilityModeArguments(
+        // FIXME: Remove argument
+        // This argument is added as a stop gap to support generating arguments
+        // for tools which currently dont leverage "OTHER_SWIFT_FLAGS". In the
+        // fullness of time, this function should operate on a strongly typed
+        // "interopMode" property of SwiftTargetBuildDescription. Instead of
+        // digging through "OTHER_SWIFT_FLAGS" manually.
+        allowDuplicate: Bool
+    ) throws -> [String] {
+        func _cxxInteroperabilityMode(for module: ResolvedModule) -> String? {
+            let scope = self.buildParameters.createScope(for: module)
+            let flags = scope.evaluate(.OTHER_SWIFT_FLAGS)
+            return flags.first { $0.hasPrefix("-cxx-interoperability-mode=") }
+        }
+
+        // Look for cxx interop mode in the current module, if set exit early,
+        // the flag is already present.
+        var cxxInteroperabilityMode: String?
+        if let mode = _cxxInteroperabilityMode(for: self.target) {
+            if allowDuplicate {
+                cxxInteroperabilityMode = mode
+            }
+        }
+
+        // If the current module doesn't have cxx interop mode set, search
+        // through the module's dependencies looking for the a module that
+        // enables cxx interop and copy it's flag.
+        if cxxInteroperabilityMode == nil {
+            for module in try self.target.recursiveTargetDependencies() {
+                if let mode = _cxxInteroperabilityMode(for: module) {
+                    cxxInteroperabilityMode = mode
+                    break
+                }
+            }
+        }
+
+        var args = [String]()
+        if let cxxInteroperabilityMode {
+            // FIXME: this should reference a local cxxLanguageStandard property
+            // It definitely should _never_ reach back into the manifest
+            args = [cxxInteroperabilityMode]
+            if let cxxStandard = self.package.manifest.cxxLanguageStandard {
+                args += ["-Xcc", "-std=\(cxxStandard)"]
+            }
+        }
+        return args
+    }
+
 
     /// When `scanInvocation` argument is set to `true`, omit the side-effect producing arguments
     /// such as emitting a module or supplementary outputs.
--- a/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
@@ -547,7 +547,8 @@ public final class SwiftTargetBuildDescription {
             args += ["-color-diagnostics"]
         }
 
-        args += try self.cxxInteroperabilityModeArguments(allowDuplicate: false)
+        args += try self.cxxInteroperabilityModeArguments(
+            propagateFromCurrentModuleOtherSwiftFlags: false)
 
         // Add arguments from declared build settings.
         args += try self.buildSettingsFlags()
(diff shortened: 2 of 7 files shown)


MaxDesiatov (commented):
Would you mind adding a test to make sure this doesn't regress in the future/

Sources/Commands/Utilities/SymbolGraphExtract.swift
```diff
@@ -82,7 +82,22 @@ public struct SymbolGraphExtract {
...
+        if !cxxInterop {
+            for module in try module.recursiveTargetDependencies() {
+                if module.cxxInterop() {
+                    cxxInterop = true
+                    break
+                }
+            }
+        }
+        if cxxInterop {
+            commandLine += ["-cxx-interoperability-mode=default"]
```
egorzhdan, line 98: Some projects use a non-default compat version (e.g. `swift-5.9`), we should forward the actual version here.

Sources/Commands/Utilities/SymbolGraphExtract.swift
```diff
@@ -107,3 +122,14 @@ public struct SymbolGraphExtract {
         return try process.waitUntilExit()
     }
 }
+
+extension ResolvedModule {
+    func cxxInterop() -> Bool {
```
egorzhdan, line 127: Could you please pick a more descriptive name for this, e.g. `var enablesCxxInterop: Bool { get { ... } }`?

egorzhdan (changes requested):
This logic mostly duplicates the existing solution that forwards the C++ interop compat version flag to the test discovery target: https://github.com/apple/swift-package-manager/blob/8d419298fe880c44692a757c92ee5ab5d3b6d6f3/Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift#L555-L574

I think it's worth extracting and reusing it.

Sources/Commands/Utilities/SymbolGraphExtract.swift
MaxDesiatov: To follow Swift API guidelines, I'd prefer this to be named `isCxxInteropEnabled`.

Sources/Commands/Utilities/SymbolGraphExtract.swift
rauhul: I dont see anything in SwiftPM that currents sets a non-"default" value and based on the PackageModel I dont see how a user can even select a "non-default compat version"

Sources/Commands/Utilities/SymbolGraphExtract.swift
```diff
@@ -82,7 +82,22 @@ public struct SymbolGraphExtract {
         if includeSPISymbols {
             commandLine += ["-include-spi-symbols"]
         }
-        
+
+        // If this target or any dependencies is is built with C++ interop
+        // enabled then swift-symbolgraph-extract must also enable C++ interop.
+        var cxxInterop = module.cxxInterop()
+        if !cxxInterop {
+            for module in try module.recursiveTargetDependencies() {
```
daniel-grumberg, line 90: Would it make sense to refactor this into a self contained function that tells you whether or not we need the interop flag?

rauhul: Yeah probably, Im struggling to figure out exactly where to put this logic/the logic egor mentioned above so its accessible from both DumpSymbolGraph and SwiftTargetBuildDescription.

Sources/Commands/Utilities/SymbolGraphExtract.swift
rauhul: It seems like this support was removed in #6583

rauhul:
@MaxDesiatov @egorzhdan @daniel-grumberg I've refactored this with (a lot of) guidance from @neonichu, please take another look

rauhul:
@swift-ci smoke test

rauhul:
@swift-ci test

rauhul:
@swift-ci test

rauhul:
@swift-ci smoke test

rauhul:
Using:
- This branch rebased on rel/6.0
- A local copy of `swift-symbolgraph-extract` with support for cxx interop
- A local copy of `swift-mmio` with `swift-docc-plugin`

Via:
```
DOCC_EXEC=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/docc SWIFT_EXEC=/Volumes/Developer/org.swift/build/Ninja-RelWithDebInfoAssert/swift-macosx-arm64/bin/swiftc .build/debug/swift-package --package-path /Users/rauhul/Developer/org.swift/swift-mmio generate-documentation
```
I am able to properly generate docs.

---

Running based on `main`, I see the following errors unrelated to this change:
```
/Volumes/Developer/org.swift/swift-mmio/.build/arm64-apple-macosx/debug/MMIO-tool.build/module.modulemap:1:8: error: redefinition of module 'MMIO'
module MMIO {
       ^
/Volumes/Developer/org.swift/swift-mmio/.build/arm64-apple-macosx/debug/MMIO.build/module.modulemap:1:8: note: previously defined here
module MMIO {
       ^
```

---

When running `swift-package dump-symbol-graph`, I see a crash in `swift-symbolgraph-extract` in `llvm:DenseMapBase:FindAndConstruct` after very deep recursion in `Module.cpp:ollectExportedImports`.

I think both of these issues are unrelated to my change here and should be handled independently.

rauhul:
@swift-ci test

rauhul:
@swift-ci smoke test

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
```diff
@@ -644,6 +625,67 @@ public final class SwiftTargetBuildDescription {
...
+    public func symbolGraphExtractArguments() throws -> [String] {
+        var args = [String]()
+        args += try self.cxxInteroperabilityModeArguments(allowDuplicate: true)
+        return args
+    }
+
+    /// Determines the arguments needed for cxx interop for this module.
+    ///
+    /// If the current module or any of its linked dependencies requires cxx
+    /// interop, cxx interop will be enabled on the current module.
```
egorzhdan, line 640: We don't actually want to implicitly enable C++ interop for anything other than the test discovery target or the symbolgraph target.
One of the reasons for that is that C++ interop is incompatible with Swift's library evolution. If a dependency of your project uses C++ internally, without re-exporting any C++ symbols, you don't need to enable C++ interop and can keep library evolution enabled.

Tests/BuildTests/BuildPlanTests.swift
```diff
@@ -1894,6 +1917,81 @@ final class BuildPlanTests: XCTestCase {
...
+                [.anySequence, "-cxx-interoperability-mode=default", "-Xcc", "-std=c++20", .anySequence]
+            )
+        }
+
+        // Swift module transitively importing cxx module
+        do {
+            try XCTAssertMatch(
+                result.target(for: "swiftLib2").swiftTarget().compileArguments(),
+                [.anySequence, "-cxx-interoperability-mode=default", "-Xcc", "-std=c++20", .anySequence]
+            )
```
egorzhdan, line 1987: This assertion needs to be reversed: unless the Swift target explicitly enables C++ interop, we should not pass these flags to the compiler.

egorzhdan (changes requested):

Sources/Commands/Utilities/SymbolGraphExtract.swift
egorzhdan: It is still possible to pass a custom C++ interop compat version, for example:
```
            swiftSettings: [
                .unsafeFlags(["-cxx-interoperability-mode=upcoming-swift"])
            ]
```

rauhul:
@swift-ci test

rauhul:
@swift-ci smoke test

rauhul:
@swift-ci test

rauhul:
@swift-ci smoke test

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
```diff
@@ -644,6 +626,67 @@ public final class SwiftTargetBuildDescription {
...
         return args
     }
+    
+    /// Determines the arguments needed to run `swift-symbolgraph-extract` for
+    /// this module.
+    public func symbolGraphExtractArguments() throws -> [String] {
+        var args = [String]()
+        args += try self.cxxInteroperabilityModeArguments(
+            propagateFromCurrentModuleOtherSwiftFlags: true)
+        return args
```
egorzhdan, line 636: Can this be simplified to just `return self.cxxInteroperabilityModeArguments(propagateFromCurrentModuleOtherSwiftFlags: true)`?

egorzhdan (approved):
LGTM from the perspective of C++ interop.
I would probably wait for someone from the SwiftPM side of things to give a thumbs up as well before merging 😉

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
```diff
@@ -644,6 +626,67 @@ public final class SwiftTargetBuildDescription {
 
         return args
     }
+    
+    /// Determines the arguments needed to run `swift-symbolgraph-extract` for
+    /// this module.
+    public func symbolGraphExtractArguments() throws -> [String] {
```
MaxDesiatov, line 632:
```suggestion
    package func symbolGraphExtractArguments() throws -> [String] {
```

Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
```diff
@@ -207,6 +207,20 @@ public final class ClangTargetBuildDescription {
         }
     }
 
+    /// Determines the arguments needed to run `swift-symbolgraph-extract` for
+    /// this module.
+    public func symbolGraphExtractArguments() throws -> [String] {
```
MaxDesiatov, line 212:
```suggestion
    package func symbolGraphExtractArguments() throws -> [String] {
```

rauhul:
@MaxDesiatov do you mind if I fix the `public` -> `package` change in my follow up PR here #7621

MaxDesiatov (approved):

Sources/Build/BuildDescription/SwiftTargetBuildDescription.swift
```diff
@@ -644,6 +626,67 @@ public final class SwiftTargetBuildDescription {
...
+        // for tools which currently don't leverage "OTHER_SWIFT_FLAGS". In the
+        // fullness of time this function should operate on a strongly typed
+        // "interopMode" property of SwiftTargetBuildDescription instead of
+        // digging through "OTHER_SWIFT_FLAGS" manually.
+        propagateFromCurrentModuleOtherSwiftFlags: Bool
+    ) throws -> [String] {
+        func cxxInteroperabilityModeAndStandard(
+            for module: ResolvedModule
+        ) -> [String]? {
+            let scope = self.buildParameters.createScope(for: module)
```
xedin, line 654: Is this correct to use `buildParameters` of the context here for dependency scopes? What if if this i.e. executable target that has a plugin or a macro dependency and that in turn has other dependencies, we'd use "destination" parameters for their scope, which seems counter-intuitive? 

It seems like we need to avoid adding new uses of `recursiveTargetDependencies()` in this context when build parameters are involved, because they are strongly connected to a destination, and instead traverse a build plan graph and use `TargetDependencyDescription`.

rauhul: I'd rather not change this logic here, im simply moving the existing code such that symbol graph can use it.

rauhul: I guess the point of that comment is to say, we should not throw away the typed interop information at that layer and try to recover it here by parsing flags, we should instead determine the flags at this layer

xedin: Maybe you could do the same thing as `.SWIFT_VERSION` and add a new declaration for assigments like that which means that you won't need to deal with prefixes here?

xedin: I don't mean it has to be in this PR, just an idea how to avoid having to prefix match things...

PR #7615, opened 2024-05-31
Add support for new Android overlay (`import Android`)

finagolfin:
This adds the overlay to four of the six files that currently `import Glibc`, with those last two showing no difference if this _wasn't_ added, like `PackageDescription.swift`.

### DIFF
--- a/Tests/PackageLoadingTests/PD_5_3_LoadingTests.swift
+++ b/Tests/PackageLoadingTests/PD_5_3_LoadingTests.swift
@@ -505,17 +505,6 @@ final class PackageDescription5_3LoadingTests: PackageDescriptionLoadingTests {
 
     func testNonZeroExitStatusDoesNotAssert() async throws {
         let content = """
-            #if canImport(Glibc)
-            import Glibc
-            #elseif canImport(Musl)
-            import Musl
-            #elseif os(Windows)
-            import MSVCRT
-            import WinSDK
-            #else
-            import Darwin.C
-            #endif
-
             print("crash")
             exit(1)
             """
--- a/Fixtures/DependencyResolution/External/Complex/FisherYates/src/Fisher-Yates_Shuffle.swift
+++ b/Fixtures/DependencyResolution/External/Complex/FisherYates/src/Fisher-Yates_Shuffle.swift
@@ -4,6 +4,8 @@ import Darwin
 import Glibc
 #elseif canImport(Musl)
 import Musl
+#elseif canImport(Android)
+import Android
 #endif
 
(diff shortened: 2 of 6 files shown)


finagolfin:
Updated to remove the libc imports altogether from those two files where not adding the Android overlay didn't matter, and found that everything worked fine on linux x86_64 too. Let's run the CI and see if it makes a difference on _any_ platform, as I suppose this may be handled differently on Darwin or Windows.

@kateinoigakukun, would you run the CI on this now?

finagolfin:
@MaxDesiatov, need a CI run here.

kateinoigakukun:
@swift-ci test

finagolfin:
All CI passed except for Windows, so I put back that one Windows import alone. I can restore all those libc imports, if there's concern they might be needed in other compilation scenarios. I have no idea when these libc imports are needed, so just going off compilation failures now.

Another CI run and this should be ready for review.

kateinoigakukun:
@swift-ci test

finagolfin:
Sigh, I thought I could quickly fix some of those `@_implementationOnly` deprecation warnings, but it looks like that requires fixing other files too, so reverted the part of this pull where I added `internal import`s.

Another CI run and we can see if this passes now.

kateinoigakukun:
@swift-ci test

Tests/PackageLoadingTests/PD_5_3_LoadingTests.swift
```diff
@@ -505,17 +505,6 @@ final class PackageDescription5_3LoadingTests: PackageDescriptionLoadingTests {
 
     func testNonZeroExitStatusDoesNotAssert() async throws {
         let content = """
-            #if canImport(Glibc)
-            import Glibc
-            #elseif canImport(Musl)
-            import Musl
-            #elseif os(Windows)
-            import MSVCRT
-            import WinSDK
```
finagolfin, line 514: @hyp, the Windows CI passed but I notice it doesn't run the SPM tests, just builds the source? Let me know if this change is okay on Windows.

finagolfin:
@xedin, maybe you can review?

finagolfin:
@bnbarham, looking for review on this tiny pull, as I currently have to [manually patch the `swift-package` executable on my Android CI](https://github.com/finagolfin/swift-android-sdk/blob/main/.github/workflows/sdks.yml#L162) because it generates the wrong test file for the current trunk stdlib that uses this new `Android` overlay.

bnbarham (approved):

bnbarham:

PR #7619, opened 2024-05-31
[Build] Adjust `computeLLBuildTargetName` to attempt `host` destinati…

xedin:
…on for test products

### Motivation:

Just like macros and plugins, tests products can also be built for the host depending on their test target(s) destinations (which are inferred based on dependencies). In order to properly support `--test-product` option we need to fallback to `host` lookup for test products just like we do for macros and plugins.

### Modifications:

- Added a fallback lookup to `computeLLBuildTargetName`

### Result:

`--test-product` now supports both `host` and `target` destination products.


xedin:
@swift-ci please test

xedin:
@swift-ci please test

xedin:

bnbarham (approved):

PR #7620, opened 2024-05-31
Adopt language mode naming in `PackageDescription` API

dempseyatgithub:
This PR implements the package manager changes in Swift evolution proposal [SE-0441](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0441-formalize-language-mode-terminology.md).

### DIFF
--- a/Sources/PackageDescription/BuildSettings.swift
+++ b/Sources/PackageDescription/BuildSettings.swift
@@ -422,7 +422,7 @@ public struct SwiftSetting: Sendable {
       _ condition: BuildSettingCondition? = nil
     ) -> SwiftSetting {
         return SwiftSetting(
-            name: "swiftLanguageVersion", value: [.init(describing: languageMode)], condition: condition)
+            name: "swiftLanguageMode", value: [.init(describing: languageMode)], condition: condition)
     }
 }
 
--- a/Sources/PackageDescription/BuildSettings.swift
+++ b/Sources/PackageDescription/BuildSettings.swift
@@ -414,7 +414,7 @@ public struct SwiftSetting: Sendable {
     /// - Since: First available in PackageDescription 6.0.
     ///
     /// - Parameters:
-    ///   - languageMode: The Swift language mode to use.
+    ///   - mode: The Swift language mode to use.
     ///   - condition: A condition that restricts the application of the build setting.
     @available(_PackageDescription, introduced: 6.0)
     public static func swiftLanguageMode(
--- a/Sources/PackageDescription/BuildSettings.swift
+++ b/Sources/PackageDescription/BuildSettings.swift
@@ -414,6 +414,23 @@ public struct SwiftSetting: Sendable {
     /// - Since: First available in PackageDescription 6.0.
     ///
     /// - Parameters:
+    ///   - version: The Swift language version to use.
+    ///   - condition: A condition that restricts the application of the build setting.
+    @available(_PackageDescription, introduced: 6.0, deprecated: 6.0, renamed: "swiftLanguageMode(_:_:)")
+    public static func swiftLanguageVersion(
+      _ version: SwiftVersion,
+      _ condition: BuildSettingCondition? = nil
+    ) -> SwiftSetting {
+        return SwiftSetting(
+            name: "swiftLanguageVersion", value: [.init(describing: version)], condition: condition)
+    }
+
+    /// Defines a `-language-mode` to pass  to the
+    /// corresponding build tool.
+    ///
+    /// - Since: First available in PackageDescription 6.0.
+    ///
+    /// - Parameters:
     ///   - mode: The Swift language mode to use.
     ///   - condition: A condition that restricts the application of the build setting.
     @available(_PackageDescription, introduced: 6.0)
--- a/Sources/PackageDescription/BuildSettings.swift
+++ b/Sources/PackageDescription/BuildSettings.swift
@@ -418,11 +418,11 @@ public struct SwiftSetting: Sendable {
     ///   - condition: A condition that restricts the application of the build setting.
     @available(_PackageDescription, introduced: 6.0)
     public static func swiftLanguageMode(
-      _ languageMode: SwiftLanguageMode,
+      _ mode: SwiftLanguageMode,
       _ condition: BuildSettingCondition? = nil
     ) -> SwiftSetting {
         return SwiftSetting(
-            name: "swiftLanguageMode", value: [.init(describing: languageMode)], condition: condition)
+            name: "swiftLanguageMode", value: [.init(describing: mode)], condition: condition)
     }
 }
 
--- a/Sources/PackageDescription/BuildSettings.swift
+++ b/Sources/PackageDescription/BuildSettings.swift
@@ -414,15 +414,15 @@ public struct SwiftSetting: Sendable {
     /// - Since: First available in PackageDescription 6.0.
     ///
     /// - Parameters:
-    ///   - version: The Swift language version to use.
+    ///   - languageMode: The Swift language mode to use.
     ///   - condition: A condition that restricts the application of the build setting.
     @available(_PackageDescription, introduced: 6.0)
-    public static func swiftLanguageVersion(
-      _ version: SwiftVersion,
+    public static func swiftLanguageMode(
+      _ languageMode: SwiftLanguageMode,
       _ condition: BuildSettingCondition? = nil
     ) -> SwiftSetting {
         return SwiftSetting(
-            name: "swiftLanguageVersion", value: [.init(describing: version)], condition: condition)
+            name: "swiftLanguageVersion", value: [.init(describing: languageMode)], condition: condition)
     }
 }
 
--- a/Sources/PackageDescription/PackageDescription.swift
+++ b/Sources/PackageDescription/PackageDescription.swift
@@ -260,6 +260,7 @@ public final class Package {
     ///   - swiftLanguageVersions: The list of Swift versions with which this package is compatible.
     ///   - cLanguageStandard: The C language standard to use for all C targets in this package.
     ///   - cxxLanguageStandard: The C++ language standard to use for all C++ targets in this package.
+    @_disfavoredOverload
     @available(_PackageDescription, introduced: 5.3)
     @available(_PackageDescription, deprecated: 6, renamed:"init(name:defaultLocalization:platforms:pkgConfig:providers:products:dependencies:targets:swiftLanguageModes:cLanguageStandard:cxxLanguageStandard:)")
     public init(
@@ -271,7 +272,7 @@ public final class Package {
         products: [Product] = [],
         dependencies: [Dependency] = [],
         targets: [Target] = [],
-        swiftLanguageVersions: [SwiftVersion]?,
+        swiftLanguageVersions: [SwiftVersion]? = nil,
         cLanguageStandard: CLanguageStandard? = nil,
         cxxLanguageStandard: CXXLanguageStandard? = nil
     ) {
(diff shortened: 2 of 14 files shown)


dempseyatgithub:
@swift-ci please test

dempseyatgithub:
@swift-ci please test

dempseyatgithub:
@swift-ci please test

Sources/PackageDescription/BuildSettings.swift
```diff
@@ -394,15 +394,15 @@ public struct SwiftSetting: Sendable {
...
     ///
     /// - Parameters:
-    ///   - version: The Swift language version to use.
+    ///   - languageMode: The Swift language mode to use.
     ///   - condition: A condition that restricts the application of the build setting.
     @available(_PackageDescription, introduced: 6.0)
-    public static func swiftLanguageVersion(
-      _ version: SwiftVersion,
+    public static func swiftLanguageMode(
+      _ languageMode: SwiftLanguageMode,
```
xedin, line 401: I think we can go with just `mode` for the label.

Tests/PackageLoadingTests/PD_5_2_LoadingTests.swift
```diff
@@ -414,13 +414,15 @@ final class PackageDescription5_2LoadingTests: PackageDescriptionLoadingTests {
...
-                    ]
+                    ],
+                    swiftLanguageVersions: [.v5]
                 )
                 """
 
             let observability = ObservabilitySystem.makeForTesting()
             await XCTAssertAsyncThrowsError(try await loadAndValidateManifest(content, observabilityScope: observability.topScope), "expected error") { error in
                 if case ManifestParseError.invalidManifestFormat(let error, _, _) = error {
+                    print("BEGIN Error\n\(error)\nEND Error")
```
xedin, line 425: Nit: Leftover `print`

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -252,6 +252,7 @@ public final class Package {
     ///   - cLanguageStandard: The C language standard to use for all C targets in this package.
     ///   - cxxLanguageStandard: The C++ language standard to use for all C++ targets in this package.
     @available(_PackageDescription, introduced: 5.3)
+    @available(_PackageDescription, obsoleted:6, renamed:"init(name:defaultLocalization:platforms:pkgConfig:providers:products:dependencies:targets:swiftLanguageModes:cLanguageStandard:cxxLanguageStandard:)")
```
xedin, line 255: Nit: missing space before `6`

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -278,6 +279,50 @@ public final class Package {
...
+    ///   `<name>.pc` file to get the additional flags required for a system target.
+    ///   - providers: The package providers for a system target.
+    ///   - products: The list of products that this package makes available for clients to use.
+    ///   - dependencies: The list of package dependencies.
+    ///   - targets: The list of targets that are part of this package.
+    ///   - swiftLanguageModes: The list of Swift language modes with which this package is compatible.
+    ///   - cLanguageStandard: The C language standard to use for all C targets in this package.
+    ///   - cxxLanguageStandard: The C++ language standard to use for all C++ targets in this package.
+    @available(_PackageDescription, introduced: 6)
+    public init(
```
xedin, line 299: How is this not ambiguous for scenarios when `swiftLanguageModes` are not provided? It should be similar to i.e.:
```
func test(x: Int, b: Int = 0) {}
func test(x: Int, c: Int = 1) {}

test(x: 42)
```

where `test(x: 42)` is an ambiguous reference.

xedin (commented):
Overall LGTM, I'm not sure about new initializer though, it seems like that could some unfortunate ambiguity issues because renamed parameter is defaulted.

Sources/PackageDescription/PackageDescription.swift
dempseyatgithub: The previous init method is marked as obsoleted in 6. So there is only one valid init method.

This appears to be the approach that has been used for all the previous init methods.

dempseyatgithub:
@swift-ci test please

dempseyatgithub:
@swift-ci test

dempseyatgithub:
@swift-ci test

Sources/PackageDescription/PackageDescription.swift
xedin: “Obsoleted” doesn’t add anything in terms of weight for overload ranking so both methods are equally valid in swift 6 mode.

dempseyatgithub:
@swift-ci please test

Sources/PackageDescription/PackageDescription.swift
xedin: Oh I confused it with “deprecated”, sorry!

dempseyatgithub:
@swift-ci please test

dempseyatgithub:
@swift-ci please test

Sources/PackageDescription/BuildSettings.swift
dempseyatgithub: Yes, that would also follow the convention used by `interoperabilityMode()`. Change made here and in the proposal.

Sources/PackageDescription/PackageDescription.swift
dempseyatgithub: Thank you! Removed.

Tests/PackageLoadingTests/PD_5_2_LoadingTests.swift
dempseyatgithub: Thank you! Removed.

dempseyatgithub:
> Overall LGTM, I'm not sure about new initializer though, it seems like that could some unfortunate ambiguity issues because renamed parameter is defaulted.

Yes, the proposal describes this. Looking back at all the existing init methods, every time a new Package init method is introduced, it is made available in a particular release. The existing init method is obsoleted in the same release, so there is no ambiguity.

When adding parameters, there is no breaking change. Changing a parameter name, as happens here, introduces a breaking change when moving to swift tools version 6.0. The `renamed` annotation allows the compiler to provide a fix-it.

dempseyatgithub:

dempseyatgithub:
@swift-ci please test

dempseyatgithub:
@xedin I believe all of the Swift PM changes in the proposal are included and the Linux and macOS tests pass.

(Something related to the CI itself seems to have gone wrong in the latest macOS test, I will re-run it)

Two questions:
1. Would it be desirable for me to clean up the commits for this PR and then force push it, so the commits and commit messages are nicer? If so, I can do that.

2. Is there anything else I need to do before this PR can be considered a valid implementation of the proposal? (At least the Swift PM piece of the proposal)

Thank you for your help with this.

xedin:
> Would it be desirable for me to clean up the commits for this PR and then force push it, so the commits and commit messages are nicer? If so, I can do that.

We can squash merge everything so you can add all the information you'd like in the commit message to the PR description.

> Is there anything else I need to do before this PR can be considered a valid implementation of the proposal? (At least the Swift PM piece of the proposal)

I think this is pretty much it. 

> How do you kick off Windows Platform ci tests?

`@swift-ci please test Windows platform`

xedin (approved):

MaxDesiatov:
> @swift-ci please test Windows platform

both `please` and `platform` are redundant, and it's case-insensitive, so plain `@swift-ci test windows` also works.

dempseyatgithub:
@swift-ci test

dempseyatgithub:
@swift-ci test windows

Sources/PackageDescription/BuildSettings.swift
```diff
@@ -414,15 +414,15 @@ public struct SwiftSetting: Sendable {
     /// - Since: First available in PackageDescription 6.0.
     ///
     /// - Parameters:
-    ///   - version: The Swift language version to use.
+    ///   - mode: The Swift language mode to use.
     ///   - condition: A condition that restricts the application of the build setting.
     @available(_PackageDescription, introduced: 6.0)
-    public static func swiftLanguageVersion(
-      _ version: SwiftVersion,
+    public static func swiftLanguageMode(
```
MaxDesiatov, line 420: This needs to introduce a new function and mark the old one as deprecated, otherwise existing `Package.swift` manifests will break.

MaxDesiatov (changes requested):

Sources/PackageDescription/BuildSettings.swift
dempseyatgithub: I'm willing to do so, but this would be an API change within the same release cycle. No shipping release version of Swift PM has the original API.

bnbarham: I think we should do this anyway - it would be nice not to break packages that have already started using this in their conversion to Swift 6. I admit it's a little weird, but IMO worth handling.

dempseyatgithub: Assuming this proposal does get reviewed in the Swift 6.0 timeframe, I could completely see adding an 'obsoleted in 6 / renamed in 6' for `swiftLanguageVersion` to give folks an easy migration to move over during pre-release Swift 6.0 with a fix-it - but removing that for the final version so it doesn't have to be baked into the API for all time.

MaxDesiatov:
@swift-ci test

dempseyatgithub:
@swift-ci test

dempseyatgithub:
@swift-ci test

Sources/PackageDescription/BuildSettings.swift
bnbarham: > but removing that for the final version so it doesn't have to be baked into the API for all time

We can just deprecate it and then remove when we remove the other `swiftLanguageVersion` 🤷

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -295,11 +347,10 @@ public final class Package {
     ///   - traits: The set of traits of this package.
     ///   - dependencies: The list of package dependencies.
     ///   - targets: The list of targets that are part of this package.
-    ///   - swiftLanguageVersions: The list of Swift versions with which this package is compatible.
+    ///   - swiftLanguageModes: The list of Swift language modes with which this package is compatible.
     ///   - cLanguageStandard: The C language standard to use for all C targets in this package.
     ///   - cxxLanguageStandard: The C++ language standard to use for all C++ targets in this package.
-    @_spi(ExperimentalTraits)
```
xedin, line 301: @dempseyatgithub Cannot remove this annotation because `Traits` type is currently SPI.

xedin: I think in general this change shouldn't need to change any existing availability annotations?...

MaxDesiatov, line 301: `@_spi` needs to stay on this one with same availability annotation as before, since traits are experimental and not available for any valid tools version.

xedin: This broke a toolchain build, that's how I discovered that.
```
/Users/ec2-user/jenkins/workspace/swift-PR-toolchain-macos/branch-main/swiftpm/Sources/PackageDescription/PackageDescription.swift:361:21: error: cannot use struct 'Trait' here; it is SPI
359 |         providers: [SystemPackageProvider]? = nil,
360 |         products: [Product] = [],
361 |         traits: Set<Trait> = [],
    |                     `- error: cannot use struct 'Trait' here; it is SPI
362 |         dependencies: [Dependency] = [],
363 |         targets: [Target] = [],
```

MaxDesiatov (changes requested):

MaxDesiatov:
@swift-ci test

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -292,14 +300,58 @@ public final class Package {
...
+    ///   `<name>.pc` file to get the additional flags required for a system target.
+    ///   - providers: The package providers for a system target.
+    ///   - products: The list of products that this package makes available for clients to use.
+    ///   - traits: The set of traits of this package.
+    ///   - dependencies: The list of package dependencies.
+    ///   - targets: The list of targets that are part of this package.
+    ///   - swiftLanguageModes: The list of Swift language modes with which this package is compatible.
+    ///   - cLanguageStandard: The C language standard to use for all C targets in this package.
+    ///   - cxxLanguageStandard: The C++ language standard to use for all C++ targets in this package.
+    @available(_PackageDescription, introduced: 6)
```
MaxDesiatov, line 354: Same here

```suggestion
    @_spi(ExperimentalTraits)
    @available(_PackageDescription, introduced: 999.0)
```

Sources/PackageDescription/PackageDescription.swift
xedin: Done!

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

dempseyatgithub:
@swift-ci test

dempseyatgithub:
@swift-ci test windows

dempseyatgithub:
@swift-ci test

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -271,7 +271,7 @@ public final class Package {
         products: [Product] = [],
         dependencies: [Dependency] = [],
         targets: [Target] = [],
-        swiftLanguageVersions: [SwiftVersion]? = nil,
```
xedin, line 266: I think in thise scheme new overload with `swiftLanguageModes` needs default removed as well so we have to versions that accept explicit swiftLanguage* parameter and one that sets a default.

Sources/PackageDescription/PackageDescription.swift
```diff
@@ -290,6 +290,50 @@ public final class Package {
...
+    ///   `<name>.pc` file to get the additional flags required for a system target.
+    ///   - providers: The package providers for a system target.
+    ///   - products: The list of products that this package makes available for clients to use.
+    ///   - dependencies: The list of package dependencies.
+    ///   - targets: The list of targets that are part of this package.
+    ///   - cLanguageStandard: The C language standard to use for all C targets in this package.
+    ///   - cxxLanguageStandard: The C++ language standard to use for all C++ targets in this package.
+    @available(_PackageDescription, introduced: 5.3)
+    @available(_PackageDescription, deprecated: 6, renamed:"init(name:defaultLocalization:platforms:pkgConfig:providers:products:dependencies:targets:swiftLanguageModes:cLanguageStandard:cxxLanguageStandard:)")
+    public init(
```
xedin, line 309: I think we can get the same effect by removing this overload and using `@_disfavoredOverload` on the old one, fewer overloads is usually easier to deal with, do you have a strong reference with this one?

dempseyatgithub: I didn't realize that `@_disfavoredOverload` was acceptable in public API.

But since it seems to be acceptable, I think `@_disfavoredOverload` would be a better way to go.

It would let us avoid adding a completely new init method just to have the overloads work out right.

xedin: Yeah, it's fine to use responsibly :)

dempseyatgithub:
@swift-ci test

dempseyatgithub:
@swift-ci test windows

bnbarham:
@dempseyatgithub given the short runway, I'm happy to merge this (and the 6.0 cherry-pick https://github.com/swiftlang/swift-package-manager/pull/7813) now. We can implement any requested changes or revert (if absolutely necessary) after the fact.

Thank you for all your work here 🙏!

xedin (approved):

dempseyatgithub:
> @dempseyatgithub given the relatively short runway for 6.0, I'm happy to merge this (and the 6.0 cherry-pick #7813) now. We can implement any requested changes or revert on top as necessary, and this way we can actually test it out on the nightlies.

@bnbarham This sounds good to me. I've brought the PR out of draft mode.

I'm not well versed in the nuts and bolts of the release process for the project, so I am fine with whatever you think will work best.

> Thank you for all your work here 🙏!
You're very welcome and thank you and @xedin for all of your help.

dempseyatgithub:

dji1admin (commented):
🙂

PR #7621, opened 2024-06-01
Fix `swift-symbolgraph-extract` search paths

rauhul:
Fixes an architecture and implementation bug where the include paths
for `swift-symbolgraph-extract` were being determined inside build plan
instead of by the target build description. This bug resulted in the
wrong set of paths being determined.

The new code moves the path determination logic into the proper target
build descriptions and releverages the already computed include paths.
The result of this is symbol graphs can be generated for Swift and C
targets properly now without duplicate module in the search paths and
with correct `-fmodule` usage.

### DIFF
--- a/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
@@ -218,7 +218,7 @@ public final class ClangTargetBuildDescription {
             args += ["-Xcc", "-std=\(cxxLanguageStandard)"]
         }
         args += ["-I", self.clangTarget.includeDir.pathString]
-        args += self.additionalFlags
+        args += self.additionalFlags.asSwiftcCCompilerFlags()
         // Unconditionally use clang modules with swift tools.
         args += try self.clangModuleArguments().asSwiftcCCompilerFlags()
         args += try self.currentModuleMapFileArguments().asSwiftcCCompilerFlags()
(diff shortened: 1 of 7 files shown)


rauhul:
builds on top of #7610

Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
```diff
@@ -209,15 +209,19 @@ public final class ClangTargetBuildDescription {
...
         if self.clangTarget.isCXX {
             args += ["-cxx-interoperability-mode=default"]
         }
         if let cxxLanguageStandard = self.clangTarget.cxxLanguageStandard {
             args += ["-Xcc", "-std=\(cxxLanguageStandard)"]
         }
+        args += ["-I", self.clangTarget.includeDir.pathString]
+        args += self.additionalFlags
+        // Unconditionally use clang modules with swift tools.
+        args += try self.clangModuleArguments().asSwiftcCCompilerFlags()
```
xedin, line 223: These are injected based on whether modules are enabled or not, does that matter for `symbolGraphExtractArguments`?

rauhul:
@swift-ci test

rauhul:
@swift-ci test windows

Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
xedin: Alright, sounds good!

rauhul:
@swift-ci test

rauhul:
@swift-ci test windows

MaxDesiatov (approved):

PR #7627, opened 2024-06-03
Add destination C module targets so you can prepare them directly.

dschaefer2:
This allows the C module to be targeted directly with the --target argument. It is a no-op however, since the output of C modules are not required for indexing.


dschaefer2:
@swift-ci Please test

MaxDesiatov (approved):

xedin (approved):

PR #7629, opened 2024-06-03
Fix create symbol graph plugin entrypoint

rauhul:
The functionality of `createSymbolGraphForPlugin` broke when the internal SwiftPM build graph was split into target and host graphs. A client running `createSymbolGraphForPlugin` on a host target would fail with `unspecified("could not find a target named “<Target-Name>”")`

This PR fixes the plugin entry point to search for targets in the host build graph if not found in the target graph.

This PR also includes a cosmetic fix for the above error so it renders as `Unspecified: Could not find a target named “MMIOMacros”.`

### DIFF
--- a/Sources/Commands/Utilities/PluginDelegate.swift
+++ b/Sources/Commands/Utilities/PluginDelegate.swift
@@ -13,6 +13,7 @@
 import Basics
 import CoreCommands
 import Foundation
+import PackageGraph
 import PackageModel
 import SPMBuildCore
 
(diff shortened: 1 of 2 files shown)


rauhul:
@swift-ci test

rauhul:
@swift-ci test

Sources/Commands/Utilities/PluginDelegate.swift
```diff
@@ -383,14 +384,38 @@ final class PluginDelegate: PluginInvocationDelegate {
...
+            throw StringError("Could not find a target named “\(targetName)”")
         }
 
         // Build the target, if needed.
-        try buildSystem.build(subset: .target(target.name))
+        try buildSystem.build(subset: .target(target.name, for: destination))
+
+        let buildParameters: BuildParameters = try if destination == .target {
+                buildSystem.buildPlan.destinationBuildParameters
+            } else {
```
xedin, line 416: The indentation looks off here

Sources/Commands/Utilities/PluginDelegate.swift
```diff
@@ -383,14 +384,38 @@ final class PluginDelegate: PluginInvocationDelegate {
...
+        // Find the target in the build operation's package graph
+        // 1. First look for a matching destination module
+        // 2. If not found, search for a matching tools module, only allowing
+        //    modules that should be built for the host (macro, plugin, test).
+        // 3. Error if no matching targets are found.
+        //
+        // FIXME: dynamic graph
+        // This should work by requesting a build for a target w/ destination:
+        // .destination and generating the graph needed on demand instead of
+        // looking for a preexisting matching target.
```
xedin, line 396: Maybe we should add a `func buildsetFor(target: String) throws -> Buildset` to `BuildSystem` protocol which would encapsulate most of the logic here and allow you to select appropriate build parameters based on resulting buildset.

rauhul: I think the client should be able to request any module built for the target/destination platform and we should create the graph that does that. Is that what you're suggesting here?

xedin: Yes, client would pass a name and we'd start with `.destination` and if that doesn't exist produce a `.host` if it's appropriate.

xedin:
Superseded by https://github.com/apple/swift-package-manager/pull/7646

PR #7630, opened 2024-06-04
[Basics] Switch `IdentifiableSet` to use `OrderedDictionary`

xedin:
### Motivation:

Turns `IdentifiableSet` into an ordered collection which should cut on flakiness in tests and other spots that expect certain ordering of elements.

### Modifications:

Changes `storage` of `IdentifiableSet` from `Dictionary` to `OrderedDictionary`.

### Result:

Less flakiness throughout.

### DIFF
--- a/Sources/Basics/Collections/IdentifiableSet.swift
+++ b/Sources/Basics/Collections/IdentifiableSet.swift
@@ -10,6 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import struct OrderedCollections.OrderedDictionary
+
 /// Replacement for `Set` elements that can't be `Hashable`, but can be `Identifiable`.
 public struct IdentifiableSet<Element: Identifiable>: Collection {
     public init() {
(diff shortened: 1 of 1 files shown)


xedin:
@swift-ci please test

MaxDesiatov (approved):
Thanks!

MaxDesiatov:
IMO definitely worth cherry-picking for 6.0.

rauhul:
I was looking at `IdentifiableSet` recently, is there a reason to duplicate the API surface of SetAlgebra instead of using `Set<Identified<Element>>`?

```swift
struct Identified<Element> where Element: Identifiable {
  var element: Element
}

extension Identified: Equatable {
  static func ==(lhs: Self, rhs: Self) -> Bool { lhs.id == rhs.id }
}

extension Identified: Hashable {
  func hash(into hasher: inout Hasher) -> Bool { hasher.combine(self.id) }
}
```

MaxDesiatov:
How would one look up an element by its ID in `Set<Identified<Element>>`?

rauhul:
> How would one look up an element by its ID in `Set<Identified<Element>>`? We also have `subscript`s defined on `IdentifiableSet`, it's a somewhat weird `Dictionary`/`Set` hybrid.

Clearly I didn't think this through enough!

MaxDesiatov:
For a bit more context, we were storing some of its elements in an `Array` previously, then for deduplication `Set` made more sense, and we were also storing some of those elements in dictionaries too. To make those element types `Sendable` and modifications local and predictable, we had to convert them from `class`es to `struct`s. The synthesized `Hashable` and `Equatable` conformances on `struct`s led to performance regressions. Use of `Identifiable` and `IdentifiableSet` were introduced to fix that. Then the only remaining hole was non-determinism when iterating through elements, which is now fixed with `OrderedDictionary`.

PR #7631, opened 2024-06-04
Require CMake 3.24 in `CMakeLists.txt`

MaxDesiatov:
Versions older than 3.24 don't support `GLOBAL` argument of `find_package` and this leads to obscure errors with older versions of CMake. See https://cmake.org/cmake/help/latest/release/3.24.html#commands for the detailed CMake 3.24 changelog. We should specify the required version explicitly to avoid this.

### DIFF
--- a/BuildSupport/SwiftSyntax/CMakeLists.txt
+++ b/BuildSupport/SwiftSyntax/CMakeLists.txt
@@ -1,5 +1,7 @@
 include(FetchContent)
 
+cmake_minimum_required(VERSION 3.24)
+
 find_package(SwiftSyntax CONFIG GLOBAL)
 if(NOT SwiftSyntax_FOUND)
   set(SWIFT_SYNTAX_INSTALL_TARGETS YES)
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

BuildSupport/SwiftSyntax/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 include(FetchContent)
 
+cmake_minimum_required(VERSION 3.24)
```
bnbarham, line 3: Can you move this to the top level cmake file? An alternative is to just put this top level and avoid the `GLOBAL`. I don't have strong opinions here, update-checkout grabs 3.24 so depending on it seems fine too 🤷

MaxDesiatov:
@swift-ci test

bnbarham (approved):

MaxDesiatov:
@swift-ci test windows

PR #7633, opened 2024-06-04
Fix `clang` linker driver issue in the bootstrap script

MaxDesiatov:
llbuild's `utils/adjust-times` executable can't be linked on Ubuntu Jammy with a plain `clang` invocation, which doesn't link C++ standard library. Replacing that `clang` invocation with `clang++` fixes the issue.


MaxDesiatov:
@swift-ci test

bnbarham (approved):

PR #7639, opened 2024-06-05
Pass through the toolchain to use for manifest loading

bnbarham:
Prefer the toolchain provided in the build plan rather than the default host toolchain so that the returned arguments align with the toolchain used to construct the build plan.

### DIFF
--- a/Sources/SourceKitLSPAPI/BuildDescription.swift
+++ b/Sources/SourceKitLSPAPI/BuildDescription.swift
@@ -24,6 +24,7 @@ import class Build.SwiftModuleBuildDescription
 import struct PackageGraph.ResolvedModule
 import struct PackageGraph.ModulesGraph
 import enum PackageGraph.BuildTriple
+internal import class PackageModel.UserToolchain
 
 public typealias BuildTriple = PackageGraph.BuildTriple
 
@@ -133,6 +134,7 @@ public struct BuildDescription {
                 return PluginTargetBuildDescription(
                     target: target,
                     toolsVersion: package.manifest.toolsVersion,
+                    toolchain: buildPlan.toolsBuildParameters.toolchain,
                     isPartOfRootPackage: modulesGraph.rootPackages.map(\.id).contains(package.id)
                 )
             }
--- a/Sources/SourceKitLSPAPI/PluginTargetBuildDescription.swift
+++ b/Sources/SourceKitLSPAPI/PluginTargetBuildDescription.swift
@@ -18,18 +18,20 @@ import struct PackageGraph.ResolvedModule
 
 private import class PackageLoading.ManifestLoader
 internal import struct PackageModel.ToolsVersion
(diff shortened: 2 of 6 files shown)


Sources/SourceKitLSPAPI/BuildDescription.swift
```diff
@@ -128,6 +129,7 @@ public struct BuildDescription {
                 return PluginTargetBuildDescription(
                     target: target,
                     toolsVersion: package.manifest.toolsVersion,
+                    toolchain: buildPlan.toolsBuildParameters.toolchain as? UserToolchain,
```
bnbarham, line 132: I very much dislike this, but `ManifestLoader` has a bunch of uses of members in `UserToolchain` that are not in `Toolchain`. Should we add requirements for them all?

xedin: If they make sense for Toolchain protocol, why not?

Sources/SourceKitLSPAPI/PluginTargetBuildDescription.swift
```diff
@@ -16,18 +16,20 @@ import struct PackageGraph.ResolvedModule
...
 import enum PackageGraph.BuildTriple
 
 struct PluginTargetBuildDescription: BuildTarget {
     private let target: ResolvedModule
     private let toolsVersion: ToolsVersion
+    private let toolchain: UserToolchain?
     let isPartOfRootPackage: Bool
 
-    init(target: ResolvedModule, toolsVersion: ToolsVersion, isPartOfRootPackage: Bool) {
+    init(target: ResolvedModule, toolsVersion: ToolsVersion, toolchain: UserToolchain?, isPartOfRootPackage: Bool) {
```
xedin, line 28: How about we pass `BuildParameters` here which should allow us make `toolchain` field non-Optional, we can assign it this way in the initializer:

```swift
self.toolchain = if let toolchain = buildParameters.toolchain as? UserToolchain {
  toolchain
} else {
  UserToolchain(swiftSDK: .hostSwiftSDK()))
}
```

Sources/SourceKitLSPAPI/BuildDescription.swift
xedin: There is a FIXME right on top of UserToolchain, might be the time to make it right...

Sources/SourceKitLSPAPI/PluginTargetBuildDescription.swift
bnbarham: Ended up adding a static method to get the arguments and then adding `swiftPMLibrariesLocation` to `Toolchain`, which then gets rid of this optional problem entirely.

bnbarham:
@swift-ci please test

MaxDesiatov (approved):

xedin (approved):

PR #7645, opened 2024-06-07
fix: bundlePath of ClangTarget to using C99ExtendedIdentifier

NakaokaRei:
### Motivation:

When attempting to access a ClangTarget resource bundle from Objective-C’s `SWIFTPM_MODULE_BUNDLE` where the package or target name contains “-”, a crash occurred.

For instance, in the following `Package.swift`, the generated bundle resource is named `SampleObjc-Package_SampleObjc-Target.bundle`, but `SWIFTPM_MODULE_BUNDLE` tries to access `SampleObjc_Package_SampleObjc_Target.bundle`.
```
import PackageDescription

let package = Package(
    name: "SampleObjc-Package",
    products: [
        .library(
            name: "SampleObjc-Target",
            targets: ["SampleObjc-Target"]),
    ],
    targets: [
        .target(
            name: "SampleObjc-Target",
            path: "Sources/SampleObjc",
            resources: [
                .process("Resources")
            ]
        )
    ]
)
```

This means that createResourcesBundle uses `target.bundlePath` for the filename, which is originally generated from the `potentialBundleName` below.:
```
let potentialBundleName = self.manifest.displayName + "_" + potentialModule.name
```

Therefore, when accessing a resource bundle with “-” in the package or target name, the access string uses `c99Name` (where “-” is replaced with “_”), but the actual generated resource bundle name does not replace “-”, leading to a mismatch and crash.

### Modifications:

Hence, this PR changes the resource bundle name (bundlePath) to use `c99Name`.

### Result:

After this change, the resource bundle names will use C99-compliant identifiers, ensuring consistency between the generated bundle resource name and the name accessed via `SWIFTPM_MODULE_BUNDLE`. This will prevent crashes when accessing resource bundles with package or target names that contain hyphens (”-”).

### DIFF
--- a/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
@@ -57,8 +57,11 @@ public final class ClangTargetBuildDescription {
             return .none
         }
 
-        let bundleName = target.underlying.c99name
-        return self.buildParameters.bundlePath(named: bundleName)
+        if let bundleName = target.underlying.potentialBundleName?.spm_mangledToC99ExtendedIdentifier() {
+           return self.buildParameters.bundlePath(named: bundleName)
+       } else {
+           return .none
+       }
     }
 
     /// The modulemap file for this target, if any.
--- a/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
+++ b/Sources/Build/BuildDescription/ClangTargetBuildDescription.swift
@@ -58,10 +58,10 @@ public final class ClangTargetBuildDescription {
         }
 
         if let bundleName = target.underlying.potentialBundleName?.spm_mangledToC99ExtendedIdentifier() {
-           return self.buildParameters.bundlePath(named: bundleName)
-       } else {
-           return .none
-       }
+            return self.buildParameters.bundlePath(named: bundleName)
+        } else {
+            return .none
+        }
     }
 
     /// The modulemap file for this target, if any.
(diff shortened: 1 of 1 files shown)


MaxDesiatov (changes requested):
Would you mind adding a test that ensures this behavior doesn't regress? Thanks!

MaxDesiatov:
@NakaokaRei to elaborate, you can follow https://github.com/apple/swift-package-manager/blob/main/Tests/BuildTests/BuildPlanTests.swift. Construct a mock package manifest and generate a build plan for it, checking that correct options and paths are generated in the build plan.

NakaokaRei:
@MaxDesiatov 
Thanks for the information about adding unit test. However, I  just realized this bug only happens in Xcode. I created a this repository to reproduce this bug.
https://github.com/NakaokaRei/Objc-Resources-Package

In this repo, I created simple objective-c target named "Objc-Resources-Package" having resources. And in the case of running test at Xcode or using `xcodebuild` command, this bug will be reproduced. Bud in the case of running test using `swift test` command, even though objective-c target having "-" in name, the test will be successful.

I guess this is because of the difference of generated `resource_bundle_accessor.m` between Xcode and Swift Package Manager. So I am thinking the bug should be fixed at Xcode side instead of this pr. What do you think ??

MaxDesiatov:
If the bug is not reproducible on CLI with either of SwiftPM subcommands like `swift build` or `swift test`, then this PR should be closed. Xcode is a closed-source project and its development is not tracked in the SwiftPM repository.

NakaokaRei:
Okay. I close this PR. And submitted feed back assistant about Xcode.

PR #7646, opened 2024-06-07
[Build/PackageGraph]  Sink fallback logic for macro/plugin/test products and packages identification into `ModulesGraph`

xedin:
This is a temporary fix until we can figure out a proper way to
handle situations were all we get is a name of a product or a
target.

### Motivation:

Callers or `ModulesGraph.{product, target}(for:destination:)` cannot always know the right `destination` to use at the moment because i.e. for test products and targets its contextual. We need a proper fix for this at the level or BuildSystem but for now sinking the logic down into `ModulesGraph` is the safest option.

### Modifications:

- `ModulesGraph.{product, target}(for:destination:)` can handle fallback if product/target turn out to be a macro, a plugin or a test.

### Result:

`--target` and `--product` options should work correctly regardless of underlying product/target kind.

Resolves: rdar://129400066

### DIFF
--- a/Sources/Commands/Utilities/PluginDelegate.swift
+++ b/Sources/Commands/Utilities/PluginDelegate.swift
@@ -389,8 +389,19 @@ final class PluginDelegate: PluginInvocationDelegate {
             throw StringError("could not find a target named “\(targetName)”")
         }
 
+        // FIXME: This is currently necessary because `target(for:destination:)` can
+        // produce a module that is targeting host when `targetName`` corresponds to
+        // a macro, plugin, or a test. Ideally we'd ask a build system for a`BuildSubset`
+        // and get the destination from there but there are other places that need
+        // refactoring in that way as well.
+        let buildParameters = if target.buildTriple == .tools {
+                try swiftCommandState.toolsBuildParameters
+            } else {
+                try swiftCommandState.productsBuildParameters
+            }
+
         // Build the target, if needed.
-        try buildSystem.build(subset: .target(target.name))
+        try buildSystem.build(subset: .target(target.name, for: buildParameters.destination))
 
         // Configure the symbol graph extractor.
         var symbolGraphExtractor = try SymbolGraphExtract(
@@ -419,7 +430,7 @@ final class PluginDelegate: PluginInvocationDelegate {
         guard let package = packageGraph.package(for: target) else {
             throw StringError("could not determine the package for target “\(target.name)”")
         }
-        let outputDir = try buildSystem.buildPlan.toolsBuildParameters.dataPath.appending(
+        let outputDir = try buildParameters.dataPath.appending(
             components: "extracted-symbols",
             package.identity.description,
             target.name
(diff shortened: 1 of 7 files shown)


xedin:
@swift-ci please test

Sources/Commands/Utilities/PluginDelegate.swift
```diff
@@ -389,8 +389,19 @@ final class PluginDelegate: PluginInvocationDelegate {
             throw StringError("could not find a target named “\(targetName)”")
         }
 
+        // FIXME: This is currently necessary because `target(for:destination:)` can
+        // produce a module that is targeting host when `targetName`` corresponds to
+        // a macro, plugin, or a test. Ideally we'd ask a build system for a`BuildSubset`
+        // and get the destination from there but there are other places that need
+        // refactoring in that way as well.
+        let buildParameters = if target.buildTriple == .tools {
```
xedin, line 397: @rauhul FYI, I also fixed problem in symbol graph creation.

rauhul: Thank you!

rauhul: Feel free to close #7629 when this merges!

xedin: Will do!

xedin:
@swift-ci please test

xedin:

MaxDesiatov (commented):
IIUC this can be covered in `BuildPlanTests`, or in `SwiftCommandStateTests` otherwise?

xedin:
Yeah, I need to figure out where to put the tests for this…

xedin:
@MaxDesiatov I added tests to cross-compilation suite and build operation. While doing that I had an idea:

1. Change `BuildSubset` to use an optional for `destination` parameter and default that to `.none`
2. Switch `destination` parameter of `ModulesGraph.{product, target}(for:destination:)` to use optional as well which is also defaulted to `.none`.
3. This might make "check .destination first and fallback to .tools" better scoped - only applicable when a specific destination is not requested and lets all of the commands use the graph without guessing the destination when its not clear.

WDYT?

xedin:
@swift-ci please test

MaxDesiatov (approved):

MaxDesiatov:
> @MaxDesiatov I added tests to cross-compilation suite and build operation. While doing that I had an idea:
> 
> 1. Change `BuildSubset` to use an optional for `destination` parameter and default that to `.none`
> 2. Switch `destination` parameter of `ModulesGraph.{product, target}(for:destination:)` to use optional as well which is also defaulted to `.none`.
> 3. This might make "check .destination first and fallback to .tools" better scoped - only applicable when a specific destination is not requested and lets all of the commands use the graph without guessing the destination when its not clear.
> 
> WDYT?

SGTM

MaxDesiatov:
@xedin would you be able to confirm that this issue is fixed in this PR? rdar://129400066

xedin:
@MaxDesiatov Indeed it does!

xedin:
🤦 forgot to stage `private` -> `internal` change last night for `computeLLBuildName`...

xedin:
@swift-ci please test

xedin:
@MaxDesiatov I'll follow-up with optional `destination:` changes.

PR #7647, opened 2024-06-07
[Tests] Don't use invalid module to satisfy `canImport` check

cachemeifyoucan:
Stop using a fake invalid module to satisfy `canImport` check for testing as swift compiler will start to evaluate such checks to false in the future.

Instead, using a valid swift interfacefile to satisify the external module import dependency.

### Motivation:

In order for change in https://github.com/apple/swift/pull/74199 to land, update swiftpm tests to stop using invalid swiftmodule.


cachemeifyoucan:
@swift-ci please test

bnbarham (approved):

PR #7650, opened 2024-06-10
[Build/PackageGraph] Switch `BuildSubset.{product, target}` and `Modu…

xedin:
…lesGraph.{product, target}(...)` to use optional destination

This is mostly an NFC change. It makes more sense to default `destination` to "undefined" when applicable because it's not always possible to know intended destination based on user input.

### Motivation:

This kind of design makes more sense for internal APIs, instead of having users to pass `.destination` even though it might not be always correct.

### Modifications:

- `BuildSubset.{product, target}` and `ModulesGraph.{product, target}(...)` accept an optional `destination` that is defaulted to `.none`

### Result:

Easier to understand and use APIs to request a build subset and product and module references from the modules graph.

### DIFF
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -526,9 +526,15 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
             // FIXME: This is super unfortunate that we might need to load the package graph.
             let graph = try getPackageGraph()
 
+            let buildTriple: BuildTriple? = if let destination {
+                destination == .host ? .tools : .destination
+            } else {
+                nil
+            }
+
             let product = graph.product(
                 for: productName,
-                destination: destination == .host ? .tools : .destination
+                destination: buildTriple
             )
 
             guard let product else {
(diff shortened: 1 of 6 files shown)


xedin:
@swift-ci please test

Sources/Build/BuildOperation.swift
```diff
@@ -526,9 +526,15 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
             // FIXME: This is super unfortunate that we might need to load the package graph.
             let graph = try getPackageGraph()
 
+            let buildTriple: BuildTriple? = if let destination {
+                    destination == .host ? .tools : .destination
```
MaxDesiatov, line 530: Is this how `swiftformat` actually indents it? 😦

MaxDesiatov (approved):
LGTM modulo `if` expression formatting

Sources/Build/BuildOperation.swift
xedin: Looks like I might have ran the wrong one again...

xedin: Fixed, thanks!

xedin:
@swift-ci please test

xedin:

PR #7649, opened 2024-06-10
SwiftPackageIndex dependency update

finestructure:
Draft PR for discussion. We're currently being held back from pulling in Vapor and NIO updates due to tighter dependency clauses on swift-collections, swift-system, and swift-certificates in SwiftPM.

We need to depend on SwiftPM in order to use the package collection signing module. We do this by depending on `revision: "swift-5.10.1-RELEASE"` in the SwiftPackageIndex server project.

This change cherry-picks 1c626233f51e66af2e08f3a6d0939504a0fe476a which opens up swift-collections and changes the dependency clauses do to the same for swift-system and swift-certificates.

PR #7655, opened 2024-06-12
Relax version ranges for dependencies in `Package.swift`

finestructure:
This replaces #7649 (which proposed the same change on the 5.10 branch, I'l open a PR to cherry-pick back to 5.10 and 6.0 as well)

Referencing the rationale from the original PR:

We're currently being held back from pulling in Vapor and NIO updates due to tighter dependency clauses on swift-system and swift-certificates in SwiftPM.

We need to depend on SwiftPM in order to use the package collection signing module. We do this by depending on `revision: "swift-5.10.1-RELEASE"` in the SwiftPackageIndex server project.


MaxDesiatov:
@swift-ci test

bnbarham (approved):

MaxDesiatov (approved):

PR #7660, opened 2024-06-13
[Build] NFC: A few minor refactorings to build operation state tracking

xedin:
### Motivation:

Trying to make it possible to share `createBuildOperation` between multiple implementations of `SPMCoreBuild.BuildSystem` in preparation to introduce an operation that would build plugin tools.

### Modifications:

- Rename `BuildOperationBuildSystemDelegateHandler` into `LLBuildProgressTracker` which is a more neutral name that could be used by different llbuild operations if necessary.
- Integrate `commandFailureHandler` into the progress tracker
- Make `BuildOperation.createBuildSystem` stateless and use a single member to set both a new build system and its tracker.

### Result:

The change it make it much easier to move `createBuildSystem` out of `BuildOperation` and into `SPMCoreBuild.BuildSystem` itself.

### DIFF
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -55,11 +55,9 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
     /// The path to scratch space (.build) directory.
     let scratchDirectory: AbsolutePath
 
-    /// The llbuild build delegate reference.
-    private var progressTracker: LLBuildProgressTracker?
-
-    /// The llbuild build system reference.
-    private var buildSystem: SPMLLBuild.BuildSystem?
+    /// The llbuild build system reference previously created
+    /// via `createBuildSystem` call.
+    private var current: (buildSystem: SPMLLBuild.BuildSystem, tracker: LLBuildProgressTracker)?
 
     /// If build manifest caching should be enabled.
     public let cacheBuildManifest: Bool
@@ -194,7 +192,7 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
 
     /// Cancel the active build operation.
     public func cancel(deadline: DispatchTime) throws {
-        buildSystem?.cancel()
+        current?.buildSystem.cancel()
     }
 
     // Emit a warning if a target imports another target in this build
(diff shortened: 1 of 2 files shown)


Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
```diff
@@ -644,10 +644,9 @@ final class CopyCommand: CustomLLBuildCommand {
 }
 
 /// Convenient llbuild build system delegate implementation
-final class BuildOperationBuildSystemDelegateHandler: LLBuildBuildSystemDelegate, SwiftCompilerOutputParserDelegate {
+final class LLBuildProgressTracker: LLBuildBuildSystemDelegate, SwiftCompilerOutputParserDelegate {
```
xedin, line 647: @MaxDesiatov I think we should rename this whole file but there is so much stuff going on here that I don't know what a good name would be maybe something like LLBuildSupport...

MaxDesiatov: The file is big enough to split up into multiple files with names corresponding to names of files they contain

Sources/Build/BuildOperation.swift
```diff
@@ -55,11 +55,9 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
     /// The path to scratch space (.build) directory.
     let scratchDirectory: AbsolutePath
 
-    /// The llbuild build delegate reference.
-    private var buildSystemDelegate: BuildOperationBuildSystemDelegateHandler?
-
-    /// The llbuild build system reference.
-    private var buildSystem: SPMLLBuild.BuildSystem?
+    /// The llbuild build system reference previously created
+    /// via \c createBuildSystem call.
```
MaxDesiatov, line 59:
```suggestion
    /// via `createBuildSystem` call.
```

Sources/Build/BuildOperation.swift
```diff
@@ -461,43 +460,46 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
         // Compile the plugin, getting back a PluginCompilationResult.
         class Delegate: PluginScriptCompilerDelegate {
```
MaxDesiatov, line 461: does this have a subclass? Could become `final` if not
```suggestion
        final class Delegate: PluginScriptCompilerDelegate {
```

Sources/Build/BuildOperation.swift
xedin: Fixed!

xedin:
@swift-ci please test

Sources/Build/BuildOperation.swift
xedin: Fixed!

Sources/Build/BuildOperationBuildSystemDelegateHandler.swift
xedin: I'll do this in a follow-up PR!

PR #7662, opened 2024-06-13
Fix: use bearer auth when pulling binaryTarget or packages from a package registry

pwallrich:
Use bearer auth when pulling binary targets or a package from a package registry and the user is `token`.

### Motivation:

When logging in to a package registry bearer auth works. However when fetching binaryTargets or pulling packages from a registry basic auth is used always.

This either breaks services that don't use basic auth or leads to nasty workarounds like first creating a netrc file with "token" as the user and after login changing the user back to the actual user. Assuming that basic auth is allowed with an identiy token.

### Modifications:

Check if the user is `token` when creating the authorization header and use Bearer auth in these cases.

### Result:

When the user is `token` the `Authorization` header will be set to `Bearer {{access_token}}` instead of `Basic token:{{access_token}}`. 

Before:
Screenshot 2024-06-13 at 22 49 08

After:
Screenshot 2024-06-13 at 22 48 38

### DIFF
--- a/Sources/Basics/AuthorizationProvider.swift
+++ b/Sources/Basics/AuthorizationProvider.swift
@@ -71,6 +71,9 @@ extension AuthorizationProvider {
         guard let (user, password) = self.authentication(for: url) else {
             return nil
         }
+        guard user != "token" else {
+            return "Bearer \(password)"
+        }
         let authString = "\(user):\(password)"
         let authData = Data(authString.utf8)
         return "Basic \(authData.base64EncodedString())"
--- a/Tests/BasicsTests/AuthorizationProviderTests.swift
+++ b/Tests/BasicsTests/AuthorizationProviderTests.swift
@@ -63,6 +63,15 @@ final class AuthorizationProviderTests: XCTestCase {
         }
     }
 
+    func testBasicAPIsBearerToken() {
+        let url = URL("http://\(UUID().uuidString)")
+        let user = "token"
+        let token = UUID().uuidString
+
+        let provider = TestProvider(map: [url: (user: user, password: token)])
+        self.assertBearerAuthentication(provider, for: url, expected: token)
+    }
+
     func testProtocolHostPort() throws {
         #if !canImport(Security)
         try XCTSkipIf(true)
@@ -256,6 +265,20 @@ final class AuthorizationProviderTests: XCTestCase {
         XCTAssertEqual(
             provider.httpAuthorizationHeader(for: url),
             "Basic " + Data("\(expected.user):\(expected.password)".utf8).base64EncodedString()
+        )
+    }
+
+    private func assertBearerAuthentication(
+        _ provider: AuthorizationProvider,
+        for url: URL,
+        expected: String
+    ) {
+        let authentication = provider.authentication(for: url)
+        XCTAssertEqual(authentication?.user, "token")
+        XCTAssertEqual(authentication?.password, expected)
+        XCTAssertEqual(
+            provider.httpAuthorizationHeader(for: url),
+            "Bearer \(expected)"
         )
     }
 }


MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):
Thank you!

MaxDesiatov:
@pwallrich mind cherry-picking it for `release/6.0`? You'll have to follow [the release branch template](https://forums.swift.org/t/swift-6-0-release-process/70220) for release branch PRs.

pwallrich:
Yes, I'll try to open the release PR.

Do I need to do anything regarding the failed linux smoke test?

MaxDesiatov:
@swift-ci test

MaxDesiatov:
Hi @yim-lee, would you have a moment to have a look at this PR, as the original SE proposal author to confirm that this change makes sense? Thanks!

Sources/Basics/AuthorizationProvider.swift
```diff
@@ -71,6 +71,9 @@ extension AuthorizationProvider {
         guard let (user, password) = self.authentication(for: url) else {
             return nil
         }
+        guard user != "token" else {
```
yim-lee, line 74: Using `token` as username for bearer auth is specific to registry auth. Adding this change would mean applying that logic for all SwiftPM auth. Is this what we want @MaxDesiatov?

yim-lee:
> However when fetching binaryTargets or pulling packages from a registry basic auth is used always.

Binaries are not supported by the package registry spec yet.

Can you please elaborate on how you reproduce the problem?

> This either breaks services that don't use basic auth or leads to nasty workarounds like first creating a netrc file with "token" as the user and after login changing the user back to the actual user.

This is sort of how registry auth is supposed to work: the credentials are either found in netrc file or Keychain. I am sorry I don't quite understand why you have to change the username?

Sources/Basics/AuthorizationProvider.swift
MaxDesiatov: cc @bnbarham. I feel like we need to revert it if you think the behavior change (even though for making it consistent) is undesired.

pwallrich:
Took a while to reproduce the original issue again. The package registry stuff is already removed in the project since we had this issue.

A bit more detailed:
We mainly use the package-registry to host prebuilt xcframeworks. 
These are wrapped in a Swift Package that is also in the package-registry.

Package.swift from our package-registry
```
Package(
  "name": "example-package",
  "products": [ .library(name: "example-package", targets: "example-package") ],
  targets: [ .binaryTarget(name: "example-package", url: "{{package-registry-url}}/path/to/example.zip, checksum: "abc") ]
)
```

We would now want to include these packages with 
```
dependencies: [
   .package(id: "registryname.example-package", from: "1.0.0")
```

If we do so with `swift package resolve` the wrapper package is downloaded successfully using Bearer Auth because of the package registry auth mechanism.

Next step is downloading the binaryTarget that is hosted on the same server. This fails, since the binaryArtifactManager doesn't know anything about the package-registry settings and does basic auth with the user `token`.

Therefore we can't use the package-registry for any packages that contain a binaryTarget hosted on the same server.

That's why we had to remove the whole package-registry stuff and went back to repos hosted in git that only download binaryTargets from the server.

And that's what the change in this PR solves and why I opened it.

I admit that the solution is not the best one. A better one could be to also have a configuration for binaryTargets or to be able to pass `--bearer-auth` to `swift package resolve`


> I am sorry I don't quite understand why you have to change the username?

The workaround we tried is
- login to package-registry with a token which creates the netrc  entry with the user `token`
- change the user in the netrc entry to an actual user
- now call `swift package --netrc resolve`

This leads to the package-registry stuff still doing bearer auth (since that's what's specified in the configuration) and the binaryTarget being download with basic auth with the correct username.

It works, but is a bit annoying and error prone since you have to manage the netrc yourself and can't rely on the auto generated entry.

MaxDesiatov:
> And that's what the change in this PR solves and why I opened it.
> 
> I admit that the solution is not the best one. 

Unfortunately, this change seems too disruptive. Until a better solution is found, we'll have to revert this.

PR #7663, opened 2024-06-13
Fix: cache not cleaned up if download fails

pwallrich:
If a server returns an error when trying to download a binaryTarget an invalid file remains in the artifacts cache.

### Motivation:

If the download of a binaryTarget fails a cache file still remains in the artifact cache.

The file only contains the server's response which is usually the status code and the error message.
Example:
```
$ ls ~/Library/Caches/org.swift.swiftpm/artifacts
.
..
https___test_example_local_foo_zip

$ cat ~/Library/Caches/org.swift.swiftpm/artifacts/https___test_example_local_foo_zip
404 page not found
```

Worse, all following `resolve` calls therefore fail because the cached file is used but is an invalid archive.

### Modifications:

If the download fails the cache file in the artifacts cache is deleted.

### Result:
As expected, the artifacts folder is empty.

Example:
```
$ ls ~/Library/Caches/org.swift.swiftpm/artifacts
.
..
```

### DIFF
--- a/Tests/WorkspaceTests/WorkspaceTests.swift
+++ b/Tests/WorkspaceTests/WorkspaceTests.swift
@@ -7471,6 +7471,7 @@ final class WorkspaceTests: XCTestCase {
         // make sure artifact downloaded is deleted
         XCTAssertTrue(fs.isDirectory(AbsolutePath("/tmp/ws/.build/artifacts/root")))
         XCTAssertFalse(fs.exists(AbsolutePath("/tmp/ws/.build/artifacts/root/a.zip")))
+        XCTAssertFalse(fs.exists(AbsolutePath("/home/user/caches/org.swift.swiftpm/artifacts/https___a_com_a_zip")))
     }
 
     func testArtifactDownloaderOrArchiverError() throws {
--- a/Tests/WorkspaceTests/WorkspaceTests.swift
+++ b/Tests/WorkspaceTests/WorkspaceTests.swift
@@ -7416,6 +7416,7 @@ final class WorkspaceTests: XCTestCase {
         let fs = InMemoryFileSystem()
         let sandbox = AbsolutePath("/tmp/ws/")
         try fs.createDirectory(sandbox, recursive: true)
+        let artifactUrl = "https://a.com/a.zip"
 
         let httpClient = LegacyHTTPClient(handler: { request, _, completion in
             do {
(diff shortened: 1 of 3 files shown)


Tests/WorkspaceTests/WorkspaceTests.swift
```diff
@@ -7471,6 +7471,7 @@ final class WorkspaceTests: XCTestCase {
         // make sure artifact downloaded is deleted
         XCTAssertTrue(fs.isDirectory(AbsolutePath("/tmp/ws/.build/artifacts/root")))
         XCTAssertFalse(fs.exists(AbsolutePath("/tmp/ws/.build/artifacts/root/a.zip")))
+        XCTAssertFalse(fs.exists(AbsolutePath("/home/user/caches/org.swift.swiftpm/artifacts/https___a_com_a_zip")))
```
MaxDesiatov, line 7474: Is this the actual path where caches were previously stored? Can we compute it somehow from previously defined paths instead of hardcoding here?

MaxDesiatov: My concern here is that if cache storage somehow changes, this test won't detect that and will continue passing.

pwallrich:
I've updated the test to get the url from the location of the workspace which is also used when initialising the BinaryArtifactsManager in the production code.
The cacheKey is now created the same way as in the BinaryArtifactsManager.

The test could still lead to false positives, if the underlying logic is changed. But I guess to prevent this the BinaryArtifactsManager would need to be refactored which I don't feel comfortable doing at the moment ^^.

Tests/WorkspaceTests/WorkspaceTests.swift
pwallrich: I've updated the test to get the url from the location of the workspace which is also used when initialising the BinaryArtifactsManager in the production code.
The cacheKey is now created the same way as in the BinaryArtifactsManager.

The test could still lead to false positives, if the underlying logic is changed. But I guess to prevent this the BinaryArtifactsManager would need to be refactored which I don't feel comfortable doing at the moment ^^.

MaxDesiatov (approved):
Thank you 👍

MaxDesiatov:
@swift-ci test

MaxDesiatov:
Definitely worth cherry-picking this one for 6.0 👍

PR #7668, opened 2024-06-14
[6.0][Build/PackageGraph] Switch `BuildSubset.{product, target}` and `Modu…

xedin:
…lesGraph.{product, target}(...)` to use optional destination

- Explanation:

  Follow-up to https://github.com/apple/swift-package-manager/pull/7646
  
  This is mostly NFC change. It makes more sense to default `destination` to "undefined" when applicable because it's not always possible to know intended destination based on user input.

  This kind of design makes more sense for internal APIs, instead of having users to pass `.destination` even though it might not be always correct.

- Main Branch PRs: https://github.com/apple/swift-package-manager/pull/7650

- Risk: Very Low

- Reviewed By: @MaxDesiatov 

- Testing: Existing test-cases were modified and new tests were added.

### DIFF
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -516,7 +516,7 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
     }
 
     /// Compute the llbuild target name using the given subset.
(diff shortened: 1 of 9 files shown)


bnbarham (approved):

xedin:
@swift-ci please test

PR #7675, opened 2024-06-18
[Prepare] Don't skip non-exportable-decls when enable-testing

dschaefer2:
When enable-testing is turned on, the test modules need access to the non exportable decls. Make sure we aren't skipping them in this case.


dschaefer2:
@swift-ci please test

MaxDesiatov (approved):

dschaefer2:
rdar://129380615

PR #7679, opened 2024-06-18
[NFC] Vendor `AsyncProcess`, make more tests `async`

MaxDesiatov:
We'd like to implement more changes to `TSCBasic.Process` without fearing of breaking clients, in addition to the fact that TSC is deprecated.

One thing in particular we need is backpressure logic (coming in a future PR) when reading from process instances in plugins communication with the host SwiftPM process. Currently without this logic, a naive rewrite of plugins communication with structured concurrency is quite bug-prone, as it loses messages ordering and can quickly fill up all memory with resource-intensive plugins.

Thus `TSCBasic.Process` is slightly cleaned up and is vendored as `AsyncProcess` in this change. We also give it more coverage in tests by utilizing `async` overloads, which are the overloads we should be using more in the structured concurrency conversion of the SwiftPM codebase.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test

rauhul:
Could we put this in its own module instead of "Basics", that module is massive and has so much conflated code that its really hard to reason about

MaxDesiatov:
I'm not entirely convinced there's much benefit to splitting it, the downside would be that our `Package.swift` is gigantic and it's impossible to get a graph of module dependencies at a glance. Also unclear on what lines it should be split in the first place, although I wouldn't mind discussing that in a separate PR that would make an attempt to do it.

rauhul:
> I'm not entirely convinced there's much benefit to splitting it, the downside would be that our `Package.swift` is gigantic and it's impossible to get a graph of module dependencies at a glance. Also unclear on what lines it should be split on in the first place, although I wouldn't mind discussing that in a separate PR that would make an attempt to do it.
> 
> Here for now I'd prefer to roughly keep the status quo: it was in `TSCBasic` before, which more or less directly corresponds to SwiftPM's `Basics`.

Thats a fair point, but I think moving files over is a great opportunity for better hygiene. Mega modules are both slower to build both clean & incremental and bury the lead of the functionality they provide.

MaxDesiatov:
> IMO a good option for splitting the module is exposing async-process to the rest of SwiftPM via the basics module, e.g.: AsyncProcess -> Basics -> the rest of SwiftPM

This won't work, as `AsyncProcess` depends on `AbsolutePath`, which belongs to `Basics`.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

Sources/Basics/AsyncProcess.swift
```diff
@@ -0,0 +1,1466 @@
...
+    }
+
+    /// The arguments with which the process was launched.
+    package let arguments: [String]
+
+    /// The environment with which the process was launched.
+    package let environmentBlock: ProcessEnvironmentBlock
+
+    @available(*, deprecated, renamed: "env")
+    package var environment: [String:String] {
```
xedin, line 72: Nit: I think all other spots I saw for dictionaries had a space between key and value type, is this a formatter problem?

Sources/Basics/AsyncProcess.swift
```diff
@@ -0,0 +1,1466 @@
...
+            exitStatus = .terminated(code: WEXITSTATUS(exitStatusCode))
+        }
+      #endif
+        self.init(arguments: arguments, environmentBlock: environmentBlock, exitStatus: exitStatus, output: output, stderrOutput: stderrOutput)
+    }
+
+    @available(*, deprecated, message: "use `init(arguments:environmentBlock:exitStatusCode:output:stderrOutput:)`")
+    package init(
+        arguments: [String],
+        environment: [String:String],
```
xedin, line 119: Same formatting thing in here...

Sources/Basics/AsyncProcess.swift
```diff
@@ -0,0 +1,1466 @@
...
+        self.environmentBlock = environmentBlock
+        self.output = output
+        self.stderrOutput = stderrOutput
+        self.exitStatus = exitStatus
+    }
+
+    @available(*, deprecated, message: "use `init(arguments:environmentBlock:exitStatus:output:stderrOutput:)`")
+    package init(
+        arguments: [String],
+        environment: [String:String],
```
xedin, line 153: And here.

xedin (approved):

xedin:
I skimmed through all of the renames and mostly read through AsyncProcess itself which looks good to me.

Sources/Basics/AsyncProcess.swift
MaxDesiatov: I forgot to format that file, cleaning up now...

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/Basics/Environment/Environment.swift
```diff
@@ -84,6 +84,10 @@ extension [String: String] {
 // MARK: - Path Modification
 
 extension Environment {
+    var path: String? {
```
rauhul, line 87: can you use `environment[.path]` instead? its more clear imo that this is a dictionary-ish object

Sources/Basics/Environment/EnvironmentShims.swift
```diff
@@ -39,153 +35,3 @@ extension ProcessEnvironmentBlock {
         }
     }
 }
-
```
rauhul, line 42: nice this entire block was gross

Sources/DriverSupport/SPMSwiftDriverExecutor.swift
```diff
@@ -70,7 +70,7 @@ public final class SPMSwiftDriverExecutor: DriverExecutor {
     }
     
     public func checkNonZeroExit(args: String..., environment: [String : String]) throws -> String {
-        try TSCBasic.Process.checkNonZeroExit(arguments: args, environment: .init(environment))
+        return try AsyncProcess.checkNonZeroExit(arguments: args, environment: .init(environment))
```
rauhul, line 73: extra return?

Sources/Basics/Environment/Environment.swift
MaxDesiatov: Where would `.path` come from?

Sources/DriverSupport/SPMSwiftDriverExecutor.swift
MaxDesiatov: Just a bad merge

Sources/Basics/Environment/Environment.swift
rauhul: Yeah its a static member of `EnvironmentKey`, for example in UserToolchain:
```swift
            // Get the search paths from PATH.
            self.envSearchPaths = getEnvSearchPaths(
                pathString: environment[.path],
                currentWorkingDirectory: fileSystem.currentWorkingDirectory
            )
```

MaxDesiatov: Found it, thanks!

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/Basics/AsyncProcess.swift
```diff
@@ -0,0 +1,1272 @@
...
+        #if os(Windows)
+        let process = Foundation.Process()
+        self._process = process
+        process.arguments = Array(self.arguments.dropFirst()) // Avoid including the executable URL twice.
+        if let workingDirectory {
+            process.currentDirectoryURL = workingDirectory.asURL
+        }
+        process.executableURL = executablePath.asURL
+        process
+            .environment = [String: String](
```
rauhul, line 469: I think this should be `[String: String](environment)`

rauhul (approved):
very happy to see fewer tsc imports. I think with this change we shouldn't need ProcessEnvironmentBlock except for Git

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test

PR #7680, opened 2024-06-18
Revert "[Traits] Package trait description and serialization"

ahoppen:
Reverts apple/swift-package-manager#7669

This PR broke SourceKit-LSP’s tests. It’s reproducible by
1. Cloning https://github.com/apple/sourcekit-lsp.git
2. Running the `testBackgroundIndexingOfSingleFile` test using `SOURCEKITLSP_FORCE_NON_DARWIN_LOGGER=1 swift test --filter testBackgroundIndexingOfSingleFile`

Which fails with the following log output

```
[org.swift.sourcekit-lsp:request-0] debug 2024-06-18 12:13:37.3640 -0700
SwiftPM log: [debug]: resolving and updating 'Package.resolved'
---
[org.swift.sourcekit-lsp:request-0] debug 2024-06-18 12:13:37.3670 -0700
SwiftPM log: [debug]: evaluating manifest for 'testbackgroundindexingofsinglefile-e43cb5da' v. unknown
---
[org.swift.sourcekit-lsp:default] default 2024-06-18 12:13:37.8340 -0700
SwiftPM log: [error]: keyNotFound(CodingKeys(stringValue: "traits", intValue: nil), Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "package", intValue: nil)], debugDescription: "No value associated with key CodingKeys(stringValue: \"traits\", intValue: nil) (\"traits\").", underlyingError: nil))
---
[org.swift.sourcekit-lsp:request-0] error 2024-06-18 12:13:37.8350 -0700
Failed to generate build graph at file:///private/var/folders/9_/vrdblnzs0m762wbv2hsryrkr0000gn/T/sourcekit-lsp-test-scratch/testBackgroundIndexingOfSingleFile-E43CB5DA: The package does not contain a buildable target. 
Add at least one `.target` or `.executableTarget` to your `Package.swift`.
```

PR #7681, opened 2024-06-18
Make traits optional in manifest deserialization

FranzBusch:
# Motivation

We are caching manifest JSONs to speed up loading. It can happen that we cached a manifest with an older SwiftPM version and have to load the manifest from JSON with a newer SwiftPM version. With the first traits PR we made the new fields non-optional. This led to failures when loading the manifest from the cache https://github.com/apple/swift-package-manager/pull/7680.

# Modification
This PR makes the fields optional and we default to empty sets after loading deserialization.


MaxDesiatov:
@swift-ci test

FranzBusch:
@MaxDesiatov i was trying to write a test for it but I wasn't able to find a way to inject things into the cache.

PR #7682, opened 2024-06-18
Fix parameter documentation in `PackageDependency.swift`

SimplyDanny:
* Remove mention of `nil` name which is not allowed due to the parameters being non-optional.
* Synchronize wording describing local package paths.


MaxDesiatov (approved):
Good catch, thanks!

MaxDesiatov:
@swift-ci test

PR #7684, opened 2024-06-18
Fix environment related warnings

rauhul:
Adds a new `Environment` type to replace TSCBasics EnvironmentVariables
and ProcessEnvironmentBlock types. Updates points of use of the older
APIs with the newer API and adds unit tests for `Environment`.


rauhul:
@swift-ci test

rauhul:
@swift-ci test

MaxDesiatov (changes requested):
I'm not convinced `Environment` warrants its own module, I'd prefer to keep it in `Basics` for now. It's unclear to me how that logic should be followed if at all. There's a lot of functionality in `Basics`, a new module for each type (or functionality aspect) would require a dozen of new modules, which IMO overcomplicates things.

MaxDesiatov:
It also could be kept in `Basics` as a stopgap measure, and after `AsyncProcess` PR is merged`Environment` could live in the same new `AsyncProcess` module.

xedin:
It seems like `AsyncProcess` module is indeed a good place for this type.

MaxDesiatov:
Main problem with that is a dependency on `AbsolutePath` in the `AsyncProcess` type itself. For the sake of not opening that can of modularization worms, I'd prefer this PR to be focused on fixing the actual warnings only. We can come back to modularization issues separately. This would also make PRs smaller and easier to review.

rauhul:
@swift-ci test

rauhul:
@swift-ci test

rauhul:
@swift-ci test windows

MaxDesiatov:
@compnerd would you have a moment to look how well this would could on Windows?

Sources/Basics/Environment/Environment.swift
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    func withLock<U>(_ body: (inout T) -> U) -> U {
+        self.lock.lock()
+        defer { self.lock.unlock() }
+        return body(&self.value)
+    }
+}
+
+// FIXME: This should come from Foundation
```
compnerd, line 38: I don't think that this can come from Foundation. Foundation's API is not easy to change and the environment being case insensitive is not currently well handled there.

Sources/Basics/Environment/Environment.swift
```diff
@@ -0,0 +1,326 @@
...
+    package mutating func appendPath(key: EnvironmentKey, value: String) {
+        guard !value.isEmpty else { return }
+        if let existing = self[key] {
+            self[key] = "\(existing)\(Self.pathValueDelimiter)\(value)"
+        } else {
+            self[key] = value
+        }
+    }
+
+    package static var pathValueDelimiter: String {
```
compnerd, line 97: I think `pathEntryDelimiter` is better - the variable is a set of key, value pairs which makes value a bit overloaded.

Sources/Basics/Environment/Environment.swift
```diff
@@ -0,0 +1,326 @@
...
+        var description: String { "\(self.function) returned \(self.code)" }
+    }
+
+    /// Modifies the process's global environment.
+    ///
+    /// > Important: This operation is _not_ concurrency safe.
+    package static func set(key: EnvironmentKey, value: String?) throws {
+        if let value = value {
+            #if os(Windows)
+            guard _putenv("\(key)=\(value)") == 0 else {
```
compnerd, line 200: This is annoying - I don't know if we should use `_putenv` or `SetEnvironmentVariableW`. The former only adjusts the "C" functions view, the latter only changes the Win32 view. I suppose that we could call both?

Sources/Basics/Environment/Environment.swift
```diff
@@ -0,0 +1,326 @@
...
+                first = false
+            } else {
+                description.append(", ")
+            }
+            description.append("\"")
+            description.append(key.rawValue)
+            description.append("=")
+            description.append(value)
+            description.append("\"")
+        }
```
compnerd, line 294:
```suggestion
        sorted.map { #""\#($0.rawValue)"="\#($1)""# }.joined(", ")
```

Sources/Basics/Environment/Environment.swift
rauhul: The intent of the comment that this type _should_ come from a common Swift ecosystem library and foundation seems like the most natural place.

Sources/Basics/Environment/Environment.swift
rauhul: Yeah, I can change the name to your suggestion

Sources/Basics/Environment/Environment.swift
rauhul: I can have it call both, in general I strongly dislike mutating the global environment anyways

Sources/Basics/Environment/Environment.swift
rauhul: I avoided string interpolation and join to avoid intermediate allocations

rauhul: That said it really doesn't matter and I can use an easier to read method

MaxDesiatov: Curious if we have issues on the compiler repo tracking possible optimizer work that hypothetically could get rid of intermediate allocations here?

rauhul:
@swift-ci test

MaxDesiatov (approved):
Thank you for making SwiftPM more environment friendly! 🙈

MaxDesiatov:
@swift-ci test windows

rauhul:
@swift-ci test

rauhul:
@swift-ci test windows

rauhul:
@swift-ci test windows

rauhul:
@swift-ci test windows

rauhul:
looks like windows CI is failing to checkout Swift which is odd because the other checks are working

rauhul:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

xedin (approved):
Looks great, thank you! Could you please run `swiftformat` on new files before merging?

rauhul:
@swift-ci test

rauhul:
@swift-ci test windows

rauhul:
@swift-ci test windows

PR #7685, opened 2024-06-19
Clarify reproduction steps requirements in `BUG_REPORT.yml`

MaxDesiatov:
Since some newly reported issues still don't include SwiftPM CLI reproduction steps and only provide Xcode-specific steps, let's make this requirement more visible in the template.


MaxDesiatov:
@swift-ci test

bnbarham (approved):

PR #7687, opened 2024-06-19
Remove unused `Codable` conformances on `PackageModel` types

MaxDesiatov:
The conformances are unused and pull in an additional `TSCUtility.PolymorphicCodableProtocol` dependency. This change removes about 0.2-0.4 MB from final executables on macOS when comparing with `swift build -c release` to `main`.

Also allows us to rename these types more easily to disambiguate between host/target triples and manifest targets/modules in our code.

If there's an existing user of this JSON manifest serialization we're unaware of, they should rely on `swift package describe --type json` instead, which has stable, consistent, and tested output.

### DIFF
--- a/Sources/PackageModel/BuildConfiguration.swift
+++ b/Sources/PackageModel/BuildConfiguration.swift
@@ -11,7 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 /// The configuration of the build environment.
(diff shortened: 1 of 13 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

xedin (approved):

PR #7691, opened 2024-06-20
Don’t reload package manifest when a header is added

MaxDesiatov:
Currently, any new header file causes a package reload. This means that while background indexing is writing header files into the index build directory, the package will get reloaded, which is unnecessary.


MaxDesiatov:
@swift-ci test

ahoppen (commented):
While this fixes the most common issue that I have seen during a build, there are a couple of others that this change would not cover. For example, SourceKit-LSP is operating on a `.index-build` build folder when background indexing is enabled. Now say I run the first build on the command line into `.build`, which will clone the dependencies and thus generate `.swift` files in `.build`. For all of these files, we’ll consider them as affecting the build settings for SourceKit-LSP because they match the `.swift` rule.

Is there any way in SwiftPM to figure out whether a file is in a `Sources`/`Tests` folder of an existing target or the build plugin equivalent of these? Such a check would fix the scenario described above.

But thanks for the fix. I think this is very valuable to take on its own because it should fix the majority of package manifest reloads.

Sources/Workspace/Workspace.swift
```diff
@@ -1204,7 +1204,7 @@ extension Workspace {
     /// generated by SwiftPM.
     public func fileAffectsSwiftOrClangBuildSettings(filePath: AbsolutePath, packageGraph: ModulesGraph) -> Bool {
         // TODO: Implement a more sophisticated check that also verifies if the file is in the sources directories of the passed in `packageGraph`.
-        FileRuleDescription.builtinRules.contains { fileRuleDescription in
+        filePath.extension != "h" && FileRuleDescription.builtinRules.contains { fileRuleDescription in
```
ahoppen, line 1207: Would it be better to exclude all the extensions from `FileRuleDescription.header`?

MaxDesiatov: Good catch, thanks!

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
> Is there any way in SwiftPM to figure out whether a file is in a `Sources`/`Tests` folder of an existing target or the build plugin equivalent of these? Such a check would fix the scenario described above.

There's a complication as there can be files generated by plugins in the build directory, and we don't know which files until we run those plugins.

ahoppen:
I just drafted an implementation that relies on directories in the build plan instead of file extensions, which also avoids the issue of `.swift` files created in locations that don’t affect the build, which I described above 

https://github.com/apple/sourcekit-lsp/pull/1508

PR #7692, opened 2024-06-20
[6.0] Don’t reload package manifest when a header is added 

MaxDesiatov:
**Explanation**: Currently, any new header file causes a package reload. This means that while background indexing is writing header files into the index build directory, the package will get reloaded, which is unnecessary.
**Scope**: isolated to known clients using this function, primarily SourceKit-LSP
**Risk**: low, the function itself is not used in SwiftPM
**Testing**: added a new unit test
**Issue**: rdar://128573306
**Reviewer**: @ahoppen @bnbarham


MaxDesiatov:
@swift-ci test

PR #7694, opened 2024-06-21
[Traits] Add experimental flags to enable specific or all traits, disable default traits

FranzBusch:
# Motivation
When building, testing or running a root package we want users to be able to change the traits enabled in the packages. This allows them to build binaries with different traits enabled or run tests with different traits.

# Modification
This PR adds trait options to a few commands and wires them through to the ModuleGraph loading and the build description caching. The latter is important to trigger rebuilds when the trait configuration has changed. Along the way I refactored the `GraphLoadingNode` and externalised the logic to calculate traits since I need slightly different logic for the root nodes and for the dependency nodes.

# Result
`swift build/run` now support passing trait configuration.

### DIFF
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -213,7 +213,7 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
                         // return the build description that's on disk.
                         let buildDescription = try BuildDescription.load(fileSystem: self.fileSystem, path: buildDescriptionPath)
 
-                        // We need to check that the build traits enabled for the cached build operation
+                        // We need to check that the build has same traits enabled for the cached build operation
                         // match otherwise we have to re-plan.
                         if buildDescription.traitConfiguration == self.traitConfiguration {
                             return buildDescription
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -55,6 +55,9 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
     /// The path to scratch space (.build) directory.
     let scratchDirectory: AbsolutePath
 
+    /// The trait configuration for this build operation.
+    private let traitConfiguration: TraitConfiguration?
+
     /// The llbuild build system reference previously created
     /// via `createBuildSystem` call.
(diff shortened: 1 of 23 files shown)


FranzBusch:
@swift-ci please test

Sources/Build/BuildOperation.swift
```diff
@@ -109,13 +112,49 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
...
+    }
+
+    package init(
         productsBuildParameters: BuildParameters,
         toolsBuildParameters: BuildParameters,
         cacheBuildManifest: Bool,
         packageGraphLoader: @escaping () throws -> ModulesGraph,
         pluginConfiguration: PluginConfiguration? = .none,
         scratchDirectory: AbsolutePath,
+        traitConfiguration: TraitConfiguration?,
```
MaxDesiatov, line 157: We don't have to keep ABI compatibility here, why not just add this new parameter to the existing initializer by providing a default value for it, instead of adding a new convenient initializer?

Sources/Build/BuildOperation.swift
```diff
@@ -171,7 +211,13 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
                             throw InternalError("could not find build descriptor at \(buildDescriptionPath)")
                         }
                         // return the build description that's on disk.
-                        return try BuildDescription.load(fileSystem: self.fileSystem, path: buildDescriptionPath)
+                        let buildDescription = try BuildDescription.load(fileSystem: self.fileSystem, path: buildDescriptionPath)
+
+                        // We need to check that the build traits enabled for the cached build operation
```
MaxDesiatov, line 216:
```suggestion
                        // We need to check that the build has same traits enabled for the cached build operation
```

Sources/Build/BuildOperation.swift
MaxDesiatov: Ah, we get to keep this parameter hidden as `package`, I see. Thanks for taking care of that!

MaxDesiatov (approved):
SGTM, thank you! I have only a single nit for something that seemed like a typo in a comment.

FranzBusch:
@swift-ci please test

MaxDesiatov:
@swift-ci test windows

PR #7695, opened 2024-06-21
Fix build of host&target destination products with `--static-swift-stdlib`

kateinoigakukun:
### Motivation:

Given the following conditions:
- `--static-swift-stdlib` is enabled (it only affects "target" destination products, "host" destination products are always dynamic)
- the building subset contains both "host" and "target" destination products derived from the same product.
- the product imports `Foundation` (that has private dependency libs)

then the build randomly failed due to the race condition of the Objects.LinkFileList creation.

Reproducible project https://github.com/kateinoigakukun/swift-autolink-issue-repro

### Modifications:

This commit fixes the issue by distinguishing the temporary link file list response file name by the `-tool` suffix.

I think it would be better to add assertions in `LLBuildManifest` to avoid such unintentional target overwrites later.

### Result:

Fix the build for the above case.

### DIFF
--- a/Tests/CommandsTests/PackageCommandTests.swift
+++ b/Tests/CommandsTests/PackageCommandTests.swift
@@ -1574,6 +1574,14 @@ final class PackageCommandTests: CommandsTestCase {
     }
 
     func testBuildToolPlugin() async throws {
+        try await testBuildToolPlugin(staticStdlib: false)
+    }
+
+    func testBuildToolPluginWithStaticStdlib() async throws {
+        try await testBuildToolPlugin(staticStdlib: true)
+    }
+
+    func testBuildToolPlugin(staticStdlib: Bool) async throws {
         // Only run the test if the environment in which we're running actually supports Swift concurrency (which the plugin APIs require).
         try XCTSkipIf(!UserToolchain.default.supportsSwiftConcurrency(), "skipping because test environment doesn't support concurrency")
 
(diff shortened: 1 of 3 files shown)


kateinoigakukun:
@swift-ci test

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -1647,7 +1655,8 @@ final class PackageCommandTests: CommandsTestCase {
             )
 
             // Invoke it, and check the results.
-            let (stdout, stderr) = try SwiftPM.Build.execute(packagePath: packageDir)
+            let args = staticStdlib ? ["--static-swift-stdlib"] : []
+            let (stdout, stderr) = try SwiftPM.Build.execute(args, packagePath: packageDir)
```
xedin, line 1659: Should this be testing produces now since the paths have diverged to make sure that they are placed correctly?

kateinoigakukun: Added the check in a lower-level test case https://github.com/apple/swift-package-manager/pull/7695/commits/ae28ac2dd706bc42bf626881693cffcbaded6f9b

xedin: Thanks!

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test Windows

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test Windows

kateinoigakukun:
@swift-ci test Windows

kateinoigakukun:
@swift-ci test Windows

MaxDesiatov (approved):
Great catch, thank you!

PR #7698, opened 2024-06-22
NFC: Assert that we don't add llbuild commands with the same name twice

kateinoigakukun:
Add debug assertion to avoid llbuild command name conflicts.

Depends on https://github.com/apple/swift-package-manager/pull/7695

### Motivation:

One of the reasons we couldn't reveal https://github.com/apple/swift-package-manager/pull/7695 issue for long time (at least past 2 releases) is that we overlooked name conflicts on `addWriteLinkFileListCommand`. 

Looking into the manifest builder, some of the commands also forgot to assert name conflicts, so it was error-prone.

### Modifications:

Add a convenience method to perform assertions consistently

### Result:

NFC for release build. Just add debug assertions

### DIFF
--- a/Sources/LLBuildManifest/LLBuildManifest.swift
+++ b/Sources/LLBuildManifest/LLBuildManifest.swift
@@ -226,14 +226,18 @@ public struct LLBuildManifest {
         targets[target, default: Target(name: target, nodes: [])].nodes.append(node)
     }
 
+    private mutating func addCommand(name: String, tool: ToolProtocol) {
+        assert(commands[name] == nil, "already had a command named '\(name)'")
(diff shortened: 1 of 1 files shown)


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

rauhul:
Out of curiosity does this catch any current bugs?

kateinoigakukun:
Fortunately, the current SwiftPM test suites didn't detect any bugs 😺

kateinoigakukun:
@swift-ci test Windows

xedin (approved):

PR #7699, opened 2024-06-22
Implement `fileAffectsSwiftOrClangBuildSettings` with the logic from `LLBuildManifestBuilder`

ahoppen:
Instead of inspecting file extensions to decide whether a file might affect build settings, check which files are in directories that are affecting compilation. This is the same logic that `LLBuildManifestBuilder` uses and should be more stable. In particular, this stops us from reloading the package manifest in SourceKit-LSP when a header file is written to a `.build` directory.

rdar://128573306


ahoppen:
https://github.com/apple/sourcekit-lsp/pull/1508

@swift-ci Please test

Sources/Build/BuildPlan/BuildPlan.swift
```diff
@@ -654,6 +662,34 @@ public class BuildPlan: SPMBuildCore.BuildPlan {
         }
         return try description.symbolGraphExtractArguments()
     }
+
+    /// Returns the files and directories that affect the build process of this build plan.
+    public func inputs() -> [Input] {
```
MaxDesiatov, line 667:
```suggestion
    package var inputs: [Input] {
```

Sources/Build/BuildPlan/BuildPlan.swift
```diff
@@ -654,6 +662,34 @@ public class BuildPlan: SPMBuildCore.BuildPlan {
...
+        var inputs: [Input] = []
+        for package in self.graph.rootPackages {
+            inputs += package.modules
```
MaxDesiatov, line 670: Does this include outputs of command plugins?

Sources/SourceKitLSPAPI/BuildDescription.swift
```diff
@@ -142,4 +142,26 @@ public struct BuildDescription {
             getBuildTarget(for: $0, in: modulesGraph)
         }
     }
+
+    /// Returns `true` if the file at the given path might influence build settings for a `swiftc` or `clang` invocation
+    /// generated by SwiftPM.
+    public func fileAffectsSwiftOrClangBuildSettings(_ url: URL) -> Bool {
```
MaxDesiatov, line 148:
```suggestion
    @_spi(SwiftPMInternal)
    public func fileAffectsSwiftOrClangBuildSettings(_ url: URL) -> Bool {
```

Sources/Build/BuildPlan/BuildPlan.swift
```diff
@@ -163,6 +163,14 @@ extension BuildParameters {
 
 /// A build plan for a package graph.
 public class BuildPlan: SPMBuildCore.BuildPlan {
+    /// Return value of `inputs()`
+    public enum Input {
```
MaxDesiatov, line 167:
```suggestion
    package enum Input {
```

Sources/Build/BuildPlan/BuildPlan.swift
ahoppen: Accessing `inputs` is not O(1) because it iterates over all the root packages and that’s why I generally prefer to use a function over a computed property since a computed property access feels like an O(1) operation. Happy to use a computed property though if that matches SwiftPM’s style better.]

Somewhat related to https://www.swift.org/documentation/api-design-guidelines/#document-computed-property-complexity

Sources/Build/BuildPlan/BuildPlan.swift
ahoppen: Isn’t this what the following line a little further down is for? Or am I missing something?

```swift
inputs.append(contentsOf: derivedSourceDirPaths.sorted().map { .directoryStructure($0) })
```

Sources/SourceKitLSPAPI/BuildDescription.swift
ahoppen: This is part of the SourceKitLSP API, so it should be properly public, not SPI, right? Or should we mark all public methods in the SourceKitLSPAPI module as `@_spi(SwiftPMInternal)` and then do `@_spi(SwiftPMInternal) import SourceKitLSPAPI` in SourceKitLSP?

Sources/Build/BuildPlan/BuildPlan.swift
MaxDesiatov: I don't think we've ever committed to computed properties always meaning `O(1)`. `String`'s `count` is `O(n)` and is still a computed property.

Sources/SourceKitLSPAPI/BuildDescription.swift
MaxDesiatov: Ugh, sorry, I missed it's a `SourceKitLSPAPI` module, fine for symbols in that to be plain `public`.

Sources/Build/BuildPlan/BuildPlan.swift
ahoppen: OK, maybe it’s just me who has tried to keep computed properties at O(1) then (and probably not consistently either). Changed it to a computed property.

Sources/Build/BuildPlan/BuildPlan.swift
MaxDesiatov: I've put another comment on relevant lines where context for it is more obvious.

Sources/Build/BuildPlan/BuildPlan.swift
```diff
@@ -654,6 +662,34 @@ public class BuildPlan: SPMBuildCore.BuildPlan {
...
+                .map(\.sources.root)
+                .sorted()
+                .map { .directoryStructure($0) }
+
+            // Add the output paths of any prebuilds that were run, so that we redo the plan if they change.
+            var derivedSourceDirPaths: [AbsolutePath] = []
+            for result in self.prebuildCommandResults.values.flatMap({ $0 }) {
+                derivedSourceDirPaths.append(contentsOf: result.outputDirectories)
+            }
+            inputs.append(contentsOf: derivedSourceDirPaths.sorted().map { .directoryStructure($0) })
```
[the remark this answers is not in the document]
ahoppen: Oh, great catch. I was not aware of that distinction. Do you know why we didn’t include `buildToolPluginInvocationResults` in the inputs before? It seems like they should be included for the llbuild manifest’s package structure as well. Or am I missing something?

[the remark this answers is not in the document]
MaxDesiatov: The reason could be that build plugins are separate from the main build: we need to do build planning and generate corresponding manifests twice. Once to build the plugins, then we run the plugins, get their outputs, and only then we can form a second build plan with a manifest to include plugin outputs, which are dynamic. llbuild can't add nodes to its build graph dynamically while executing an existing graph.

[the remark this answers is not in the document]
ahoppen: Ah, actually I just remembered that SourceKit-LSP currently doesn’t handle the files generated by build plugins anyway (https://github.com/apple/sourcekit-lsp/issues/665). So, adding the build plugin generated files here wouldn’t make a difference at the moment anyway and we couldn’t test it in SourceKit-LSP as well.

I would propose that we ignore the sources from build plugins for now and will consider how to handle them when https://github.com/apple/sourcekit-lsp/issues/665 is implemented. What do you think?

MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

ahoppen:
https://github.com/apple/sourcekit-lsp/pull/1508

@swift-ci Please test

Sources/Build/BuildPlan/BuildPlan.swift
```diff
@@ -654,6 +662,34 @@ public class BuildPlan: SPMBuildCore.BuildPlan {
         }
         return try description.symbolGraphExtractArguments()
     }
+
+    /// Returns the files and directories that affect the build process of this build plan.
+    package var inputs: [Input] {
```
xedin, line 667: I'm not sure whether this is relevant but - traits are currently not considered by package structure command, is it possible for them to affect what files are included in the build and subsequently influence `fileAffectsSwiftOrClangBuildSettings`?

MaxDesiatov: cc @FranzBusch

Sources/SourceKitLSPAPI/BuildDescription.swift
```diff
@@ -142,4 +142,26 @@ public struct BuildDescription {
...
+        guard let filePath = try? AbsolutePath(validating: url.path) else {
+            return false
+        }
+
+        for input in buildPlan.inputs {
```
xedin, line 153: Could we unify this filtering with inputs somehow to avoid having to build inputs just to want them again to find one relevant location?

ahoppen: I don’t think it’s worth optimizing for that:
- I don’t think that building up the inputs is all that expensive and will be any performance impact in practice. The only variables here are:
  - Number of root packages, which will be 1 in almost all cases
  - Number of source files in the root package, which is usually small-ish, definitely doesn’t go quadratic or something like that
  - Number of prebuilt output directories (I also expect this to be <10)
- The code would become quite a bit more complicated if we want to share this method for `LLBuildManifest` and SourceKit-LSP

Sources/Build/BuildPlan/BuildPlan.swift
FranzBusch: Traits aren't affecting what files are included in a single module. The same files will be part of the module just different `SWIFT_ACTIVE_COMPILATION_CONDITIONS` are passed. Traits do effect build planning though. The `swift build/run/test` commands support trait configuration arguments to change the traits of the root manifests. If the root trait changes we have to re-plan the build. This is already implemented and the trait configuration is stored in the build description.

What are the `package structure` commands in question here?

xedin: PackageStructure command determines whether anything changed in the package and would skip loading of modules graph, and rebuilding of the build plan if nothing changed if traits are not part of that consideration there is a possibility of a split-brain scenario if something started directly using `buildPackageStructure()` result as an indicator for a rebuild.

Sources/SourceKitLSPAPI/BuildDescription.swift
xedin: I've recently been shown a project with a large number of targets which was already a problem. I'm not sure how many times would `fileAffectsSwiftOrClangBuildSettings` this be called in SourceKit-LSP but it seems like that would be a problem for that project at least. If we could at least cache the `inputs` that would make a difference.

ahoppen: Oh, interesting. Looks like my performance judging wasn’t correct then.

Caching is the right solution here then. Otherwise we’d still get the worst-case behavior when a file is passed to `fileAffectSwiftOrClangBuildSettings` that’s not part of the inputs.

ahoppen:
https://github.com/apple/sourcekit-lsp/pull/1508

@swift-ci Please test

MaxDesiatov:
https://github.com/apple/sourcekit-lsp/pull/1508

@swift-ci test windows

PR #7702, opened 2024-06-24
[Traits] Correctly diagnose unused dependency with traits

FranzBusch:
# Motivation

We are currently diagnosing unused dependencies by looking at the list of products used by the final modules and check if for every root dependency we are using at least one product of that package. This incorrectly diagnosis unused dependencies if the usage of the dependency is either:
- Just declared to enable a trait if down the graph another package is depending on it
- Used as an optional dependency and the trait is disabled

# Modification

This PR adds two more checks to the diagnostic pass to find unused dependencies. We first create a set of dependency products that are guarded by a trait. When checking each dependency we then check if we enabled any trait on the dependency since this means it is used for setting a trait. We also check if any product of the dependency is guarded by a trait to identify if it was an optional dependency.

# Result
No more incorrect unused dependency warnings.

### DIFF
--- a/Sources/PackageGraph/ModulesGraph+Loading.swift
+++ b/Sources/PackageGraph/ModulesGraph+Loading.swift
@@ -277,6 +277,22 @@ private func checkAllDependenciesAreUsed(
             }
         })
 
+        // List all dependencies of modules that are guarded by a trait.
+        let traitGuardedProductDependencies = Set(package.underlying.modules.flatMap { module in
+            module.dependencies.compactMap { moduleDependency in
+                switch moduleDependency {
+                case .product(let product, let conditions):
+                    if conditions.contains(where: { $0.traitCondition != nil }) {
+                        // This is a product dependency that was enabled by a trait
+                        return product.name
+                    }
+                    return nil
+                case .module:
+                    return nil
+                }
+            }
+        })
+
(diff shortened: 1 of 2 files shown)


FranzBusch:
@swift-ci please test

FranzBusch:
@swift-ci test windows

Sources/PackageGraph/ModulesGraph+Loading.swift
```diff
@@ -300,7 +316,13 @@ private func checkAllDependenciesAreUsed(
             if dependency.products.contains(where: \.isCommandPlugin) {
                 continue
             }
-            
+
+            // Skip this check if traits are enabled since it is valid to add a dependency just
+            // to enable traits on it.
```
MaxDesiatov, line 321: Can you elaborate on this in the comment? Unsure of an exact use case of enabling traits by adding a dependency. It would be great to have an example listed right there to clarify this point, otherwise it doesn't seem obvious enough.

FranzBusch: Updated the comment with more details

MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

PR #7703, opened 2024-06-24
[Traits] Replace `isDefault` with `default` trait

FranzBusch:
# Motivation

The current spelling uses `isDefault` per trait to declare the default traits. This makes it hard to see which traits are actually the default traits.

# Modification

Instead of using an explicit `isDefault` per trait this PR now uses a `.defaults(enabledTraits:)` overload which just uses the `"default"` trait under the hood. This simplifies the logic tremendously to handle default traits.

# Result

Nicer spelling for default traits and easier implementation.

### DIFF
--- a/Sources/PackageLoading/ManifestLoader+Validation.swift
+++ b/Sources/PackageLoading/ManifestLoader+Validation.swift
@@ -107,11 +107,6 @@ public struct ManifestValidator {
                 continue
             }
 
-            guard traitName.lowercased() != "default" && traitName.lowercased() != "defaults" else {
-                diagnostics.append(.defaultTraitName())
-                continue
-            }
-
             for (index, unicodeScalar) in traitName.unicodeScalars.enumerated() {
                 let properties = unicodeScalar.properties
 
@@ -368,10 +363,6 @@ extension Basics.Diagnostic {
 
     static func emptyTraitName() -> Self {
         .error("Empty strings are not allowed as trait names")
-    }
-
-    static func defaultTraitName() -> Self {
-        .error("Traits are not allowed to be named 'default' or 'defaults' to avoid confusion with default traits")
     }
 
     static func invalidFirstCharacterInTrait(firstCharater: UnicodeScalar, trait: String) -> Self {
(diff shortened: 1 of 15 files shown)


Sources/PackageLoading/ManifestLoader+Validation.swift
```diff
@@ -107,11 +107,6 @@ public struct ManifestValidator {
                 continue
             }
 
-            guard traitName.lowercased() != "default" && traitName.lowercased() != "defaults" else {
-                diagnostics.append(.defaultTraitName())
```
MaxDesiatov, line 111: Wouldn't we want to keep this error? A user could still declare both `.default` and a trait named `"defaults"` at the same time with unintended consequences, right?

FranzBusch: I intentionally removed it because a user declaring a `default` trait is just an alternative spelling to using `.default(enabledTraits:)`. I don't want to ban this since a user could add a different description if they wanted. With this change the `default` trait is just serialised as any other trait.
I also thought about keeping the ban on declaring a  `defaults` but decided that it is not worth the restriction since it poses no harm. I could add the ban for `defaults` back and just remove the `default` one but I don't fell strongly about it

MaxDesiatov: The spelling needs to be adjusted then. The helper function is called `.default`, while the trait is called `"defaults"`.  How's about keeping the function `.default` as it is now, but making the string spelling more explicit: `"defaultTraits"`.

FranzBusch: Ah good point I gotta change the backing `String`. I do prefer `default` though. The `traits` are implied since this is part of a set of traits in the end and this would feel redundant to me.

FranzBusch:
@swift-ci please test

FranzBusch:
@swift-ci please test

FranzBusch:
@swift-ci test windows

FranzBusch:
@swift-ci please test

FranzBusch:
@swift-ci please test

FranzBusch:
@swift-ci please test

FranzBusch:
@swift-ci please test

FranzBusch:
@swift-ci test windows

FranzBusch:
@swift-ci test windows

MaxDesiatov (approved):

PR #7705, opened 2024-06-24
Add 'any' to generated test runner code

glbrntt:
Using the upcoming existential any feature via
'-Xswiftc -enable-upcoming-feature -Xswiftc ExistentialAny' causes build failures on non-Darwin platforms as a test observer class is generated which doesn't use explicit 'any'.

This is an issue for library maintainers who want to enforce this in CI.


MaxDesiatov:
@swift-ci test

PR #7706, opened 2024-06-24
[Build/Plugins] Separate plugin tool builds from plugin invocations

xedin:
An attempt to simplify `invokeBuildToolPlugins` and avoid having to use a separate BuildOperation to build plugin accessible tools.

### Motivation:

This is the first step towards the world where plugins are going to be invoked during BuildPlan construction as the modules/products are discovered.

### Modifications:

- Instead of having to pass PluginModule around together with ModulesGraph, refactor accessible tool inference to be based on ResolvedModules.
- BuildOperation: organize various build options and configurations into `LLBuildSystemConfiguration`
- Extract build of plugin tools from `invokeBuildToolPlugins` into a separate method on `BuildOperation`

### Result:

Better concern separation for plugins which enabled further refactoring and plugin integration into the build plan.

### DIFF
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -1020,8 +1020,6 @@ extension BuildOperation {
 }
 
 extension BuildOperation {
-    fileprivate typealias PluginTool = (path: AbsolutePath, triples: [String]?)
-
     private func buildPluginTools(
         graph: ModulesGraph,
         pluginsPerModule: [ResolvedModule.ID: [ResolvedModule]],
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -43,9 +43,9 @@ package struct LLBuildSystemConfiguration {
 
     let traitConfiguration: TraitConfiguration?
 
-    var manifestPath: AbsolutePath
-    var databasePath: AbsolutePath
-    var buildDescriptionPath: AbsolutePath
+    fileprivate(set) var manifestPath: AbsolutePath
+    fileprivate(set) var databasePath: AbsolutePath
+    fileprivate(set) var buildDescriptionPath: AbsolutePath
 
     let fileSystem: any Basics.FileSystem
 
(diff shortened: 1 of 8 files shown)


xedin:
@swift-ci please test

Sources/Build/BuildOperation.swift
```diff
@@ -35,28 +35,141 @@ import DriverSupport
...
+    }
+
+    func configuration(for destination: BuildParameters.Destination) -> BuildConfiguration {
+        switch destination {
+        case .host: self.toolsBuildParameters.configuration
+        case .target: self.destinationBuildParameters.configuration
+        }
+    }
+
+    static func based(
```
MaxDesiatov, line 139: As this is a value type, why not mark `manifestPath`, `databasePath`, and `buildDescriptionPath` with `package` access and avoid this `static func` altogether? Not sure it's worth maintaining a separate function for straightforward updating a few properties on a value type, unless we have bigger plans for it in the future.

Sources/Build/BuildOperation.swift
```diff
@@ -950,19 +1040,111 @@ extension BuildOperation {
     }
 }
 
+extension BuildOperation {
+    fileprivate typealias PluginTool = (path: AbsolutePath, triples: [String]?)
```
MaxDesiatov, line 1044: I'd prefer this to be a separate struct with detailed doc comments on both properties.

Sources/Commands/PackageCommands/PluginCommand.swift
```diff
@@ -373,7 +379,7 @@ struct PluginCommand: SwiftCommand {
         // TODO: We should also emit a final line of output regarding the result.
     }
 
-    static func availableCommandPlugins(in graph: ModulesGraph, limitedTo packageIdentity: String?) -> [PluginModule] {
+    static func availableCommandPlugins(in graph: ModulesGraph, limitedTo packageIdentity: String?) -> [ResolvedModule] {
```
MaxDesiatov, line 382: IIUC we force cast it to `PluginModule` at all call sites, not sure why the return type has changed?

```suggestion
    static func availableCommandPlugins(in graph: ModulesGraph, limitedTo packageIdentity: String?) -> [PluginModule] {
```

Sources/SPMTestSupport/MockBuildTestHelper.swift
```diff
@@ -254,6 +254,29 @@ public func mockBuildPlan(
         observabilityScope: observabilityScope
     )
 }
+
+public func mockPluginTools(
```
MaxDesiatov, line 258:
```suggestion
package func mockPluginTools(
```

Sources/Build/BuildOperation.swift
xedin: Oh, sure, I can do that!

Sources/Commands/PackageCommands/PluginCommand.swift
xedin: It's because of `PluginCommand.run` that uses findPlugins to filter thing and takes a `ResolvedModule` but maybe it doesn't actually need a `ResolvedModule`, I'll take a look.

xedin: We need that for `preparePluginTools` inside of `PluginCommand.run` so yeah, both of these produce `ResolvedModule` to avoid having to re-lookup stuff in the graph.

MaxDesiatov (approved):
LGTM, only `PluginTool` conversion from a tuple to a struct remains as a nit.

xedin:
@swift-ci please test

PR #7725, opened 2024-06-28
Add missing nuget case to system package provider docs

jakepetroules:
Fix the docs so "nuget" shows up under the "Providing Hints to Users of System Packages" heading along with the other cases instead of being off by itself.


jakepetroules:
@swift-ci please smoke test

MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):

PR #7727, opened 2024-06-29
Move getPackageMetadata to async/await

AndrewHoos:
Move more of the PackageCollection APIs to native async/await

### Motivation:

More modern code that will be able to adopt swift 6 data race safety eventually

### Modifications:

PackageMetadataProvider APIs are all now async/await
PackageIndexProtocol APIs are all now async/await
Added question about the need for returning PackageMetadataProviderContext when PackageBasicMetadata can not be read

### Result:

More readable code


AndrewHoos:
@swift-ci test

MaxDesiatov (approved):

PR #7750, opened 2024-07-03
Work around `swift-bootstrap` inability to handle plugins

MaxDesiatov:
`swift-bootstrap` currently can't handle dependencies with plugins, which prevents us from bumping Swift Argument Parser to a newer version, which did add a dependency on a plugin. The easiest workaround is to exclude plugins from the modules graph only when building with `swift-bootstrap`.

### DIFF
--- a/Sources/PackageGraph/ModulesGraph+Loading.swift
+++ b/Sources/PackageGraph/ModulesGraph+Loading.swift
@@ -35,7 +35,9 @@ extension ModulesGraph {
         customXCTestMinimumDeploymentTargets: [PackageModel.Platform: PlatformVersion]? = .none,
         testEntryPointPath: AbsolutePath? = nil,
         fileSystem: FileSystem,
-        observabilityScope: ObservabilityScope
+        observabilityScope: ObservabilityScope,
+        productsFilter: ((Product) -> Bool)?,
+        modulesFilter: ((Module) -> Bool)?
     ) throws -> ModulesGraph {
         try Self.load(
             root: root,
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
@swift-ci test

Sources/PackageGraph/ModulesGraph+Loading.swift
```diff
@@ -35,7 +35,9 @@ extension ModulesGraph {
         customXCTestMinimumDeploymentTargets: [PackageModel.Platform: PlatformVersion]? = .none,
         testEntryPointPath: AbsolutePath? = nil,
         fileSystem: FileSystem,
-        observabilityScope: ObservabilityScope
+        observabilityScope: ObservabilityScope,
+        productsFilter: ((Product) -> Bool)?,
+        modulesFilter: ((Module) -> Bool)?
```
bnbarham, line 40: Should we default in the public interface?

MaxDesiatov, line 40:
```suggestion
        productsFilter: ((Product) -> Bool)? = nil,
        modulesFilter: ((Module) -> Bool)? = nil
```

MaxDesiatov: good point, updated now

MaxDesiatov:
@swift-ci test

xedin (approved):
Reasonable since we are going to cherry-pick this to 6.0 but I hope that it's going to be required in the future...

MaxDesiatov:
@swift-ci test linux

PR #7754, opened 2024-07-04
Clarify unsupported direct macro module imports in `CHANGELOG.md`

MaxDesiatov:
Importing macro modules directly to use those macros was never supposed to work, and now with macro modules built for the host triple, importing such macro modules in modules built for the target triple will fail.


rauhul:
Is there a document anywhere that describes macros in SwiftPM? It would be good to include this information there too.

MaxDesiatov:
No, there's no such document at the moment unfortunately, but contributions would be of course very welcome.

xedin (approved):

PR #7752, opened 2024-07-03
[6.0] Add support for new Bionic module from Android overlay (`import Bionic`) (#7615)

finagolfin:
__Explanation:__ Now that this new overlay was merged into the 6.0 compiler too in swiftlang/swift#74758, this adds the overlay to five of the seven files that currently `import Glibc`, with those last two showing no difference if this _wasn't_ added, like `PackageDescription.swift`, so remove most of those imports.

__Scope:__ Add imports on Android and remove a handful in two files that were unused

__Issue:__ None

__Original PRs:__ #7615 and #7755

__Risk:__ Very low

__Testing:__ Passed all CI on trunk, plus on my daily Android CI

__Reviewer:__ @bnbarham

PR #7755, opened 2024-07-04
Import new Bionic module from Android overlay instead where possible

finagolfin:
Also, add the import in `Environment.swift` too

@rauhul, you added that last file just after I got these imports in with #7615, so I recently had to add this too to keep it building on Android.

### DIFF
--- a/Sources/Basics/Environment/Environment.swift
+++ b/Sources/Basics/Environment/Environment.swift
@@ -19,6 +19,8 @@ import Musl
 #elseif os(Windows)
 import CRT
 import WinSDK
+#elseif canImport(Bionic)
+import Bionic
 #else
 import Darwin.C
 #endif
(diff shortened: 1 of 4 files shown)


Sources/Basics/Environment/Environment.swift
```diff
@@ -19,6 +19,8 @@ import Musl
 #elseif os(Windows)
 import CRT
 import WinSDK
+#elseif canImport(Android)
+import Android
```
compnerd, line 23: Can we use `Bionic` here instead? Note that Darwin is using `Darwin.C` and Windows is using `CRT`. Of course if we need more than the C interfaces, `Android` is fine, but would bee nice to keep this scoped to the same level as the other platforms.

finagolfin: I'll check if that's enough.

finagolfin: Done

finagolfin:
This now makes the changes suggested in #7752, along with the original change here modified to `Bionic`. All except one test fixture from #7615 worked with `import Bionic`.

@compnerd, please review, since you asked for this change.

compnerd (approved):

finagolfin:
@rauhul, a CI run and we can get this in.

bnbarham:
@swift-ci please test

finagolfin:
@kateinoigakukun, passed CI, ready for merge.

PR #7756, opened 2024-07-04
Fix warnings caused by unused `await`

MaxDesiatov:
`UserToolchain.default.skipUnlessAtLeastSwift6` is not `async`, which triggers multiple warnings in tests.


MaxDesiatov:
@swift-ci test

xedin (approved):

PR #7760, opened 2024-07-05
Invalidate manifest cache when `-Xbuild-tools-swiftc` changes

kateinoigakukun:
The manifest cache key should take into account flags passed to `-Xbuild-tools-swiftc`

### Motivation:

SwiftPM did not re-evaluate manifests even after changing `-Xbuild-tools-swiftc`

### Modifications:

Hash those extra flags into manifest cache key

### Result:

Manifest will be properly re-evaluated after changing `-Xbuild-tools-swiftc`.

### DIFF
--- a/Sources/PackageLoading/ManifestLoader.swift
+++ b/Sources/PackageLoading/ManifestLoader.swift
@@ -569,6 +569,7 @@ public final class ManifestLoader: ManifestLoaderProtocol {
                 toolsVersion: toolsVersion,
                 env: Environment.current.cachable,
                 swiftpmVersion: SwiftVersion.current.displayString,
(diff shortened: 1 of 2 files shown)


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test Linux

xedin (approved):
Thank you!

xedin:
I think we should cherry-pick this to 6.0, would you mind opening a PR against release/6.0 branch?

kateinoigakukun:
Okay, opened it https://github.com/swiftlang/swift-package-manager/pull/7762

xedin:
Thank you!

PR #7762, opened 2024-07-06
[6.0] Invalidate manifest cache when `-Xbuild-tools-swiftc` changes

kateinoigakukun:
Cherry-pick of https://github.com/swiftlang/swift-package-manager/pull/7760

**Explanation**: The manifest cache key should take into account flags passed to `-Xbuild-tools-swiftc`, but it didn't. so SwiftPM did not re-evaluate manifests even after changing `-Xbuild-tools-swiftc`. Fix the incorrect cache by hashing those extra flags into manifest cache key.

**Scope**: Affects only when `-Xbuild-tools-swiftc` is used.
**Risk**: low due to isolated less used option
**Testing**: Added a new test case to cover the senario
**Issue**: N/A
**Reviewer**: @xedin


kateinoigakukun:
@swift-ci test

bnbarham (approved):

PR #7764, opened 2024-07-08
`add-target` breaks resulting package when target name is not valid Swift identifier

dmhts:
Executing `swift package add-target my-target --type executable` breaks the updated package because the executable target name is not a valid Swift identifier.

### Motivation:
After executing `swift package add-target my-target --type executable`, I expect the generated stub to not break the modified package. However, in this case, the resulting stub file would contain invalid Swift code:
```swift
@main
struct my-targetMain {
    static func main() {
        print("Hello, world")
    }
}
```
The same issue applies to `macro` and `test` target types. It's also important to note that `dash case` is quite often used for target names, like e.g. in [swift-argument-parser](https://github.com/apple/swift-argument-parser/blob/main/Package.swift#L65-L68).

Additionaly, when `AddTarget` is consumed via an API (as opposed to the CLI), the following parsing error is thrown into `stdout`:

Screenshot 2024-07-08 at 08 40 51

### Modifications:

I believe there are at least two ways to solve this issue:
1. Use a predefined type name in the generated stub, such as `struct ExampleMain`, so it doesn't depend on the target name.
2. Introduce a sanitizer function to ensure that the target name is always a valid Swift identifier. For example, a function that would convert "$my-target-name%" into "myTargetName".

For the sake of simplicity, I opted for approach `#1` in the proposed PR. However, if it's necessary to preserve target names in the generated stubs, we can certainly go with approach `#2`.

### Result:

`swift package add-target my-target --type executable`
```swift
@main
struct ExampleMain {
    static func main() {
        print("Hello, world")
    }
}
```


Tests/PackageModelSyntaxTests/ManifestEditTests.swift
```diff
@@ -449,6 +449,45 @@ class ManifestEditTests: XCTestCase {
         }
     }
 
+    func testAddExecutableTargetWithNameAsInvalidSwiftIdentifier() throws {
+        try assertManifestRefactor("""
```
dmhts, line 453: Instead of adding this test, we could also use a "non-valid Swift identifier target name" instead of "MyProgram" in the original `testAddExecutableTargetWithDependencies` test (the same goes for  `testAddSwiftTestingTestTargetWithNameAsInvalidSwiftIdentifier` test below). 

Alternatively, we could extend the `assertManifestRefactor` function to handle multiple editing operations at once, allowing us to add multiple targets with different names within a single test.

Sources/PackageModelSyntax/AddTarget.swift
```diff
@@ -234,10 +234,10 @@ public struct AddTarget {
         case .macro:
             """
             \(imports)
-            struct \(raw: target.name): Macro {
```
grynspan, line 237: How about using `c99name` instead?

dmhts: That's a very good point. `c99name` would certainly cover most edge cases, such as using dash-case, special characters, and digits in the first characters. Well, it's already used for language-level target name sanitizing so it must be robust enough for this use case. 

The only remaining concern is that `c99` (obviously) doesn't cover Swift's reserved words. However, in that case, we could just escape the reserved word based on a list taken e.g. from [swift-syntax](https://github.com/swiftlang/swift-syntax/blob/main/Sources/SwiftSyntax/generated/Keyword.swift), which `SwiftPM` already depends on.

I believe combining these two safeguards should be a good enough way to proceed. What do you think?

grynspan: I'm just visiting. :) There may be a better way to avoid reserved names—the package owners probably have ideas!

dmhts: Alright :) @DougGregor, maybe you have some ideas about this?

DougGregor: Both of your ideas together (`cc9name` and using the swift-syntax keyword checking) sound reasonable to me.

Tests/PackageModelSyntaxTests/ManifestEditTests.swift
DougGregor: I like the idea of using a "non-valid Swift identifier target name"  in existing tests, because it gives us better coverage without duplication. We also often create new tests from old ones, so it'll help us make sure we cover this case whenever we are introducing a new test (without always introducing two tests).

DougGregor:
@swift-ci please test

Sources/PackageModelSyntax/AddTarget.swift
```diff
@@ -414,3 +414,15 @@ fileprivate extension PackageDependency {
...
+fileprivate extension TargetDescription {
+    var sanitizedName: String {
+        name
+            .spm_mangledToC99ExtendedIdentifier()
+            .capitalizingFirstLetter()
+    }
+}
+
+fileprivate extension String {
+    func capitalizingFirstLetter() -> String { prefix(1).uppercased() + dropFirst() }
```
dmhts, line 427: Using both `prefix(1)` and `dropFirst()` is absolutely safe with strings of any length, including empty ones, as both are concerned with a maximum number of characters to process, not a minimum.

Sources/PackageModelSyntax/AddTarget.swift
```diff
@@ -414,3 +414,15 @@ fileprivate extension PackageDependency {
         )
     }
 }
+
+fileprivate extension TargetDescription {
+    var sanitizedName: String {
+        name
+            .spm_mangledToC99ExtendedIdentifier()
+            .capitalizingFirstLetter()
+    }
```
dmhts, line 423: Using the [Keyword enum ](https://github.com/swiftlang/swift-syntax/blob/main/Sources/SwiftSyntax/generated/Keyword.swift#L15) turned out to be not that trivial for this task, as there are many exceptions due to the keyword's context-specific nature, namely:
- `objc`, `escaping` and some others are considered non-valid identifiers only when they are used in combination with the leading `@` character.
- The same goes for `available` and `selector` but in combination with the `#` character
- `metadata` and `module` would also be valid identifiers, and I'm not even sure in which context they would be considered actual language keywords.
- etc.

As simple solutions are generally better, I preferred to use `c99name` in combination with a capitalization of the first letter. With the capitalization strategy, we can feed two birds with one scone: 
- It forces example type names to follow the overall type naming convention: "sanitizer" -> "struct Sanitizer"
- Reserved words are turned into valid identifiers when capitalized: "macro" -> "struct Macro"

I believe the combination of these two safeguards covers all known edge cases (to the best of my knowledge).

Tests/PackageModelSyntaxTests/ManifestEditTests.swift
dmhts: Good points! I updated the tests.

Sources/PackageModelSyntax/AddTarget.swift
dmhts: I've decided to introduce `capitalizingFirstLetter()` instead of using the existing `capitalized()` function because the latter has an undesirable by-design behavior. It lowercases all remaining letters in words, so "MyNewShinyTarget" would be turned into "Mynewshinytarget". This makes it a rather poor option for snake-case notation, which is the most popular convention for target names.

On the other hand, capitalizing only the first letter would turn "my-new-shiny-target" into "My_new_shiny_target" instead of the potentially preferred "My_New_Shiny_Target" (which `capitalized()` would produce), but using underscores would be a non-preferred Swift naming convention anyway. Therefore I would consider it a very minor downside given that we're talking about example names here.

grynspan: So, worth taking a moment to write some unit tests where the macro name does _not_ use the English Latin alphabet. For instance, does the French "être" correctly capitalize to "Être"? Does the Japanese "マクロ" stay unmodified? These are trivial test cases but worth adding.

More complex to solve would be something like the "Turkish I" capitalization problem (uppercase "i" in Turkish is "İ", not "I", and lowercase "I" is "ı".) Should this code attempt to take the user's localization settings into consideration?

Sources/PackageModelSyntax/AddTarget.swift
bnbarham: Ah sorry, I missed the discussion above. We actually added `isValidSwiftIdentifier` in swift-syntax for the purpose you described above. Having said that, IMO `c99name` + uppercase is probably good enough here anyway.

Sources/PackageModelSyntax/AddTarget.swift
dmhts: Thanks, that's a good point again. I switched to `localizedCapitalized` instead of `uppercased()`, which addresses all the cases mentioned above while taking the current user's locale into account. Interestingly, the case with the Turkish "İ" is [specifically covered in Foundation](https://github.com/apple/swift-foundation/blob/8bf4cb0b9be0cb12ce87d85d2ceafaa605584051/Tests/FoundationEssentialsTests/StringTests.swift#L1510-L1512) :) I'm surprised that the tests are far from being rigorous...

I'd love to write more tests that cover localization edge cases. However, this would require some refactoring of the test helpers (as I [mentioned here](https://github.com/swiftlang/swift-package-manager/pull/7764#discussion_r1668393979)) if we want to avoid significant duplication in tests. 
An alternative approach could be to use the Turkish "I" in the existing test target names to confirm that at least some localization is taken into account. However, I haven't found a way to properly mock `Locale.current` in tests.

At the same time, what we have might already be good enough for this specific use case, as the main goal of this PR was to prevent broken code in modified packages in the first place.

@DougGregor I'd also like to hear your opinion.

Sources/PackageModelSyntax/AddTarget.swift
dmhts: Ah, thanks! That's really nice to know.

Sources/PackageModelSyntax/AddTarget.swift
dmhts: I just checked, and `Locale.current` correctly reports `en_DE` in my case, which is aligned with the system settings:
```
defaults read ~/Library/Preferences/.GlobalPreferences AppleLocale
en_DE
```
At the same, the system `locale` utility reports the default `en_US`, as you said:
```
locale
LANG="en_US.UTF-8"
```
Meaning, `String.localizedCapitalized` should work as expected according to the user's localization preferences.

dmhts:
@DougGregor just a friendly reminder when you have time. I believe this PR can be already merged once the CI tests have passed.

Sidenote: I've been using the new package editing API quite a lot in my own project, and I've come across a few other bugs that I'd be glad to fix in follow-up PRs. For example, sanitizing imports, avoiding the use of product names in imports (only targets can be imported), etc. 

I could also address some FIXMEs, such as plugin and resources support, as I currently have to resort to workarounds to bypass them.

dschaefer2:
@swift-ci please test
@swift-ci please test windows

dschaefer2:
@swift-ci please test

dschaefer2 (approved):

dschaefer2:
Looks good. Sorry about the delay.

DougGregor:
> @DougGregor just a friendly reminder when you have time. I believe this PR can be already merged once the CI tests have passed.

Thank you for the ping, and I'm sorry I lost track of this. Your changes look great to me and we'll get them merged once CI passes. Thank you!

dmhts:
I just fixed some conflicts that popped up only during the tests (which succeeded). Next time, I'll trust GitHub's UI less and merge `main` beforehand.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dmhts:
> --- bootstrap: error: Command '['env', 'SDKROOT=/...' returned non-zero exit status 1.` 

It seems to be a flaky test :(

dschaefer2:
@swift-ci please test

dschaefer2:
> > --- bootstrap: error: Command '['env', 'SDKROOT=/...' returned non-zero exit status 1.`
> 
> It seems to be a CI issue :(

Yup CI is down at the moment. We're trying to get it back.

dschaefer2:
I should say working to get it back. I think we've identified the issue. I'll retrigger the tests once I know we're clear.

dschaefer2:
@swift-ci please test

DougGregor:
All clear, merging! Thank you @dmhts

PR #7768, opened 2024-07-10
Add the formalized Swift Testing event stream arguments to `swift test`.

grynspan:
This PR adds `--configuration-path`, `--event-stream-output-path`, and `--event-stream-version` to `swift test`.

For more information, see the Swift Testing API proposal [here](https://github.com/apple/swift-testing/blob/main/Documentation/Proposals/0002-json-abi.md).

### DIFF
--- a/Sources/Commands/SwiftTestCommand.swift
+++ b/Sources/Commands/SwiftTestCommand.swift
@@ -80,6 +80,38 @@ struct SharedOptions: ParsableArguments {
     var testProduct: String?
 }
 
+struct TestEventStreamOptions: ParsableArguments {
+    /// Legacy equivalent of ``configurationPath``.
+    @Option(name: .customLong("experimental-configuration-path"),
+            help: .private)
+    var experimentalConfigurationPath: AbsolutePath?
+
+    /// Path where swift-testing's JSON configuration should be read.
+    @Option(name: .customLong("configuration-path"),
+            help: .hidden)
+    var configurationPath: AbsolutePath?
+
+    /// Legacy equivalent of ``eventStreamOutputPath``.
+    @Option(name: .customLong("experimental-event-stream-output"),
+            help: .private)
+    var experimentalEventStreamOutputPath: AbsolutePath?
+
+    /// Path where swift-testing's JSON output should be written.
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

grynspan:
See also https://github.com/apple/swift-testing/pull/479

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -80,6 +80,38 @@ struct SharedOptions: ParsableArguments {
     var testProduct: String?
 }
 
+struct TestEventStreamOptions: ParsableArguments {
+    /// Legacy equivalent of ``configurationPath``.
+    @Option(name: .customLong("experimental-configuration-path"),
+            help: .private)
```
bnbarham, line 86: What's private vs hidden? Haven't seen private before 🤔

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -656,6 +677,10 @@ extension SwiftTestCommand {
         @OptionGroup()
         var testLibraryOptions: TestLibraryOptions
 
+        /// Options for Swift Testing's event stream.
+        @OptionGroup()
+        var testEventStreamOptions: TestEventStreamOptions
```
bnbarham, line 682: Is this really for just configuration rather than the stream options?

bnbarham (approved):

Sources/Commands/SwiftTestCommand.swift
grynspan: Hidden means it still shows up in verbose help. Private means it is never shown in help.

Sources/Commands/SwiftTestCommand.swift
grynspan: I don't understand, sorry?

bnbarham: This is in `list`. Does `list` actually need all these, or does it just need the configuration path?

grynspan: `list` can use them all now! See https://github.com/apple/swift-testing/issues/511

PR #7783, opened 2024-07-17
Disable Swift Testing by default in `swift package init`.

grynspan:
This PR turns Swift Testing back off by default when running `swift package init` because it's not available in the toolchain yet and the package dependency may come as a surprise (also, it breaks CI, but who's counting?)

Follow-up to #7766.


grynspan:
@swift-ci please test

Issue #7688, opened 2024-06-19
Package.swift diagnostics on wrong line


award999:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [X] Confirmed reproduction steps with SwiftPM CLI.

### Description

SwiftPM diagnostics for the Package.swift show up at a different line/column than SourceKit-LSP. In this picture the first diagnostics is from SourceKit and shows up at right place, but SwiftPM indicates the diagnostic is on the next line
<img width="596" alt="Screenshot 2024-06-19 at 1 12 31 PM" src="https://github.com/apple/swift-package-manager/assets/78549403/e1724a1a-1b26-4cd7-b959-28e1767d6cf3">

### Steps to reproduce

1. Set macOS platform to v13 in Package.swift
2. Set swift tool version to 5.6: `// swift-tools-version: 5.6`
3. Save file and `swift package resolve`

Notice output:
<img width="614" alt="Screenshot 2024-06-19 at 2 01 58 PM" src="https://github.com/apple/swift-package-manager/assets/78549403/c255973c-3042-4281-8a9b-2221f6f023bf">

As you can see in the picture above the line was line 9, so line is off by one, but the column seems right

### Swift & OS version (output of `swift --version ; uname -a`)

swift-driver version: 1.110 Apple Swift version 6.0 (swiftlang-6.0.0.4.52 clang-1600.0.21.1.3)
Target: arm64-apple-macosx15.0

MaxDesiatov:
Is this reproducible for errors in other files too, or only `Package.swift`?

award999:
Only seems to be this specific case in Package.swift, for example doing something like this put on the right line (22 like I'd expect):
<img width="614" alt="Screenshot 2024-06-19 at 3 42 34 PM" src="https://github.com/apple/swift-package-manager/assets/78549403/b0428865-fc5c-4738-8ff3-7ead813b916c">

All diagnostics from `swiftc` seem correct

PR #7795, opened 2024-07-18
Fix Package.swift diagnostics line off by 1 for tools versions <= 5.7

bripeticca:
Fixes issue #7688

SwiftPM was implicitly prepending the manifest with 'import Foundation' for toolchains using Swift toolchain versions 5.7 and under. The original file itself was never overwritten, thus the diagnostics being reported would always present on the next line.

This is a quick fix that appends the same line to the end of the manifest; doing so assures that the intended behaviour doesn't change, but allows for correct diagnostics to be reported.

### Motivation:

Diagnostics would present on the incorrect line for Package.swift files using tools version 5.7 and under.

### Modifications:

When evaluating the manifest, a temporary directory copies the contents of the manifest to a temporary file; rather than prepending the import to Foundation at the top of the file (causing all lines in the diagnostic reports to be off by one), append the import to the bottom of the temporary manifest when dealing with tools versions 5.7 and under.

### Result:

The intended behaviour that requires users to explicitly import Foundation doesn't change for Swift versions 5.8 and up; diagnostics reported for Package.swift files for those using swift <= 5.7 will now show on the correct line.

### DIFF
--- a/Sources/PackageLoading/ManifestLoader.swift
+++ b/Sources/PackageLoading/ManifestLoader.swift
@@ -769,13 +769,15 @@ public final class ManifestLoader: ManifestLoaderProtocol {
         if toolsVersion >= .v5_8 {
             manifestPreamble = ByteString()
         } else {
-            manifestPreamble = ByteString("import Foundation\n")
+          manifestPreamble = ByteString("\nimport Foundation")
         }
 
         do {
             try withTemporaryDirectory { tempDir, cleanupTempDir in
                 let manifestTempFilePath = tempDir.appending("manifest.swift")
-                try localFileSystem.writeFileContents(manifestTempFilePath, bytes: ByteString(manifestPreamble.contents + manifestContents))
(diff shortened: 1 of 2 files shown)


Sources/PackageLoading/ManifestLoader.swift
```diff
@@ -17,10 +17,10 @@ import Foundation
 import PackageModel
 import SourceControl
 
-import class TSCBasic.BufferedOutputByteStream
-import struct TSCBasic.ByteString
 import class Basics.AsyncProcess
 import struct Basics.AsyncProcessResult
+import class TSCBasic.BufferedOutputByteStream
+import struct TSCBasic.ByteString
```
MaxDesiatov, line 23: It's unclear at a glance which parts of the diff actually resolve the issue. We generally prefer to keep formatting changes to separate PRs, especially those affecting the whole file.

bripeticca:
@swift-ci please test

MaxDesiatov (approved):
Great find, thanks!

PR #7796, opened 2024-07-18
Don't stomp on xUnit output from XCTest when running Swift Testing.

grynspan:
This PR forces Swift Testing to write its xUnit output to a different path from what the user specifies.

In the future, we should have the two testing libraries collate their XML output into a single file, but that requires the ability to parse and reformat XML output and that's a little beyond the capabilities of this feature right now. Since we expect most uses of xUnit output right now are existing ones with existing XCTest-based tests, moving Swift Testing aside seems like the right call.

(If XCTest is explicitly disabled, the exact specified path is used.)

### DIFF
--- a/Sources/Commands/SwiftTestCommand.swift
+++ b/Sources/Commands/SwiftTestCommand.swift
@@ -453,25 +453,30 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
     ) async throws -> TestRunner.Result {
         // Pass through all arguments from the command line to Swift Testing.
         var additionalArguments = additionalArguments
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

xedin (approved):

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

grynspan:
@swift-ci please test Windows

PR #7799, opened 2024-07-19
[PubGrub] Narrow down cases when pre-release decisions are marked invalid

xedin:
Pre-release decisions are valid if the derived range they were inferred from supports pre-release versions.

### Motivation:

Currently, after deriving a pre-release assignment, the solver would assert
in some circumstances depending on order in which the dependencies are
specified in the manifest.

This change makes it possible to infer pre-release verison of a package 
even when intermediate derivations did not support pre-release versions.

For example one package could declare `swift-syntax` dependency as
`508.0.1` ..< `601.0.0` and another could narrow it to `600.0.0-latest` ..< `601.0.0`.

Since the most constrained range in such case supports pre-releases a decision
to use the lastest pre-release version of `600.0.0` should be valid if there
is no `600.0.0` release yet.

It's important to note that on each path a decision for a package is always
made based on the last undecided term's requirements which means -
based on the current semantics - if such derivation doesn't support pre-release,
even though intermediate ones did, the solver would select a released version. 
This PR is not aim to change that.

There is a caveat which we cannot narrowly fix - if package depends on an exact
pre-release version the inference is going to fail when there is another dependency
to the same package that doesn't support pre-releases. For example:

Package `A` depends on packages `B` and `C` as follows:

`A` depends on `B` = `0.0.1-alpha` and `C` = `1.0.0`
`B` version `0.0.1-alpha` has no dependencies
`C` version `1.0.0` depends on `B` - `0.0.1` ..< `0.0.2`

In this situation the solver would fail because `0.0.1-alpha` does not satisfy range - `0.0.1` ..< `0.0.2` inferred from `C`.

### Modifications:

- Delays version decisions for packages with pre-release ranges until there are no more packages without pre-release versions. This also means that if a previous decision narrows down the range to releases-only the package would be made a contender on the next step of the solver.

- Relaxes the `isValidDecision` check for pre-release decisions but makes sure that the last derivation supports pre-release versions.

### Result:

In situations like https://github.com/swiftlang/swift-package-manager/issues/7658 and https://github.com/swiftlang/swift-package-manager/issues/7643 the solver would no longer assert and would produce a solution with a pre-release version for `swift-syntax`.

### DIFF
--- a/Sources/PackageGraph/Resolution/PubGrub/PubGrubDependencyResolver.swift
+++ b/Sources/PackageGraph/Resolution/PubGrub/PubGrubDependencyResolver.swift
@@ -764,7 +764,17 @@ public struct PubGrubDependencyResolver {
                 let start = DispatchTime.now()
                 let counts = try result.get()
                 // forced unwraps safe since we are testing for count and errors above
-                let pkgTerm = undecided.min { counts[$0]! < counts[$1]! }!
+                let pkgTerm = undecided.min {
+                    // Prefer packages that don't allow pre-release versions
+                    // to allow propagation logic to find dependencies that
+                    // limit the range before making any decisions. This means
+                    // that we'd always prefer release versions.
+                    if $0.supportsPrereleases != $1.supportsPrereleases {
+                        return !$0.supportsPrereleases
+                    }
+
+                    return counts[$0]! < counts[$1]!
+                }!
                 self.delegate?.willResolve(term: pkgTerm)
                 // at this point the container is cached
                 let container = try self.provider.getCachedContainer(for: pkgTerm.node.package)
--- a/Sources/PackageGraph/Resolution/PubGrub/Term.swift
+++ b/Sources/PackageGraph/Resolution/PubGrub/Term.swift
@@ -104,9 +104,38 @@ public struct Term: Equatable, Hashable {
     /// - There has to be no decision for it.
     /// - The package version has to match all assignments.
     public func isValidDecision(for solution: PartialSolution) -> Bool {
+        // The intersection between release and pre-release ranges is
+        // allowed to produce a pre-release range. This means that the
+        // solver is allowed to make a pre-release version decision
+        // even when some of the versions didn't allow pre-releases.
+        //
+        // This means that we should ignore pre-release differences
+        // while checking derivations and assert only if the term is
+        // pre-release but the last assignment wasn't.
+        if self.supportsPrereleases {
+            if let assignment = solution.assignments.last(where: { $0.term.node == self.node }) {
+                assert(assignment.term.supportsPrereleases)
+            }
+        }
+
         for assignment in solution.assignments where assignment.term.node == self.node {
             assert(!assignment.isDecision, "Expected assignment to be a derivation.")
-            guard self.satisfies(assignment.term) else { return false }
+
+            // This is not great but dragging `ignorePrereleases` through all the APIs seems
+            // worse. This is valid because we can have a derivation chain which is something
+            // like - "0.0.1"..<"1.0.0" -> "0.0.4-latest"..<"0.0.6" and make a decision
+            // `0.0.4-alpha5` based on that if there is no `0.0.4` release. In vacuum this is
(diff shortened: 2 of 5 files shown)


MaxDesiatov:
@swift-ci test

Sources/PackageGraph/Resolution/PubGrub/PubGrubDependencyResolver.swift
```diff
@@ -764,7 +764,17 @@ public struct PubGrubDependencyResolver {
                 let start = DispatchTime.now()
                 let counts = try result.get()
                 // forced unwraps safe since we are testing for count and errors above
-                let pkgTerm = undecided.min { counts[$0]! < counts[$1]! }!
+                let pkgTerm = undecided.min {
+                    // Prefer packages that don't allow pre-release versions
+                    // to allow propagation logic it find dependencies that
```
MaxDesiatov, line 769: typo?
```suggestion
                    // to allow propagation logic to find dependencies that
```

Sources/PackageGraph/Resolution/PubGrub/Term.swift
```diff
@@ -40,6 +40,10 @@ public struct Term: Equatable, Hashable {
         )
     }
 
+    public var supportsPrereleases: Bool {
```
MaxDesiatov, line 43: obligatory access control nit

```suggestion
    package var supportsPrereleases: Bool {
```

Sources/PackageGraph/Resolution/PubGrub/Term.swift
```diff
@@ -100,9 +104,38 @@ public struct Term: Equatable, Hashable {
     /// - There has to be no decision for it.
     /// - The package version has to match all assignments.
     public func isValidDecision(for solution: PartialSolution) -> Bool {
+        // The intersection between release and pre-release ranges is
+        // allowed to produce a pre-release range which means that the
```
MaxDesiatov, line 108: readability nit
```suggestion
        // allowed to produce a pre-release range. This means that the
```

Sources/PackageGraph/Resolution/PubGrub/Term.swift
```diff
@@ -100,9 +104,38 @@ public struct Term: Equatable, Hashable {
...
         for assignment in solution.assignments where assignment.term.node == self.node {
             assert(!assignment.isDecision, "Expected assignment to be a derivation.")
-            guard self.satisfies(assignment.term) else { return false }
+
+            // This is not great but dragging `ignorePrereleases` through all the APIs seems
+            // worse. This is valid because we can have a derivation chain which is something
+            // like - "0.0.1"..<"1.0.0" -> "0.0.4-latest"..<"0.0.6" and make a decision
+            // `0.0.4-alpha5` based on that if there is no `0.0.4` release. In vacuum this is
+            // (currently) incorrect because `0.0.4-alpha5` doesn't satisfy the initial
+            // range that doesn't support pre-release versions but since the solver is
```
MaxDesiatov, line 129: readability nit
```suggestion
            // range that doesn't support pre-release versions. Since the solver is
```

MaxDesiatov (approved):
LGTM, modulo access control and comments wording nits.

Sources/PackageGraph/Resolution/PubGrub/Term.swift
xedin: ✅

Sources/PackageGraph/Resolution/PubGrub/Term.swift
xedin: ✅

Sources/PackageGraph/Resolution/PubGrub/Term.swift
xedin: ✅

Sources/PackageGraph/Resolution/PubGrub/PubGrubDependencyResolver.swift
xedin: ✅ Thanks!

xedin:
@swift-ci please test

xedin:

PR #7803, opened 2024-07-22
Don't log "no matching tests" diagnostic when running XCTest only.

grynspan:
Fixes a logic error in deciding when to log the "no matching test cases were run" diagnostic if `--disable-swift-testing` is passed.


grynspan:
@swift-ci please test

MaxDesiatov (approved):

PR #7805, opened 2024-07-22
Detect if XCTest is installed on Darwin

grynspan:
This PR checks if XCTest is available before invoking XCTest-based tests on Darwin. There are three possible outcomes:

1. If XCTest is available, we will run XCTest-based tests (as we have historically.)
2. If XCTest is not available and the user explicitly passed `--enable-xctest`, we will attempt to run XCTest-based tests, but in general this code path will continue to fail as `swift test` has historically done when XCTest is not available.
3. If XCTest is not available and the user did not pass `--enable-xctest`, we skip running any XCTest logic.

On Linux/Windows/etc., XCTest is always present via swift-corelibs-xctest and so this change has no practical effect there. On Darwin, XCTest may be missing if the user has installed the Xcode Command Line Tools, but not the full Xcode IDE. XCTest is not included with the Xcode Command Line Tools package.

The purpose of this change is to allow running `swift test` when XCTest is unavailable but Swift Testing _is_ available.

### DIFF
--- a/Sources/Commands/SwiftTestCommand.swift
+++ b/Sources/Commands/SwiftTestCommand.swift
@@ -256,8 +256,10 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
         var results = [TestRunner.Result]()
 
         // Run XCTest.
-        if options.testLibraryOptions.isEnabled(.xctest) {
-            // validate XCTest available on darwin based systems
+        if options.testLibraryOptions.isEnabled(.xctest, swiftCommandState: swiftCommandState) {
+            // Validate XCTest available on Darwin-based systems. If it's not available and we're hitting this code
+            // path, that means the developer must have explicitly passed --enable-xctest (or the toolchain is
+            // corrupt, I suppose.)
             let toolchain = try swiftCommandState.getTargetToolchain()
             if case let .unsupported(reason) = try swiftCommandState.getHostToolchain().swiftSDK.xctestSupport {
                 if let reason {
@@ -278,7 +280,7 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
                     swiftCommandState: swiftCommandState,
                     library: .xctest
                 )
-                if result == .success, let testCount, testCount == 0 {
+                if result == .success, testCount == 0 {
                     results.append(.noMatchingTests)
                 } else {
                     results.append(result)
@@ -324,9 +326,9 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
         }
 
         // Run Swift Testing (parallel or not, it has a single entry point.)
-        if options.testLibraryOptions.isEnabled(.swiftTesting) {
+        if options.testLibraryOptions.isEnabled(.swiftTesting, swiftCommandState: swiftCommandState) {
             lazy var testEntryPointPath = testProducts.lazy.compactMap(\.testEntryPointPath).first
-            if options.testLibraryOptions.isExplicitlyEnabled(.swiftTesting) || testEntryPointPath == nil {
+            if options.testLibraryOptions.isExplicitlyEnabled(.swiftTesting, swiftCommandState: swiftCommandState) || testEntryPointPath == nil {
                 results.append(
                     try await runTestProducts(
                         testProducts,
@@ -412,7 +414,7 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
     public func run(_ swiftCommandState: SwiftCommandState) async throws {
         do {
             // Validate commands arguments
-            try self.validateArguments(observabilityScope: swiftCommandState.observabilityScope)
+            try self.validateArguments(swiftCommandState: swiftCommandState)
         } catch {
             swiftCommandState.observabilityScope.emit(error)
             throw ExitCode.failure
@@ -466,7 +468,7 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
             }
             additionalArguments += commandLineArguments
 
-            if var xunitPath = options.xUnitOutput, options.testLibraryOptions.isEnabled(.xctest) {
+            if var xunitPath = options.xUnitOutput, options.testLibraryOptions.isEnabled(.xctest, swiftCommandState: swiftCommandState) {
                 // We are running Swift Testing, XCTest is also running in this session, and an xUnit path
                 // was specified. Make sure we don't stomp on XCTest's XML output by having Swift Testing
                 // write to a different path.
@@ -634,7 +636,7 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
     /// Private function that validates the commands arguments
     ///
     /// - Throws: if a command argument is invalid
-    private func validateArguments(observabilityScope: ObservabilityScope) throws {
+    private func validateArguments(swiftCommandState: SwiftCommandState) throws {
         // Validation for --num-workers.
         if let workers = options.numberOfWorkers {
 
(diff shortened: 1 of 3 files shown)


Sources/Build/LLBuildCommands.swift
```diff
@@ -259,7 +259,7 @@ final class TestEntryPointCommand: CustomLLBuildCommand, TestBuildCommand {
             #if \#(isXCTMainAvailable)
             \#(generateTestObservationCode(buildParameters: buildParameters))
 
-            import XCTest
+            @_weakLinked import XCTest
```
compnerd, line 262: This might be an issue for Windows. There is no such thing as weak linking, everything must be strongly linked.

compnerd (changes requested):

Sources/Build/LLBuildCommands.swift
grynspan: This is just a proof-of-concept branch. Not gonna merge it. :)

grynspan:
Did some experimenting—I think that what I really need here is not weak-linked XCTest, but simply to detect if `swift test` is using Xcode CL Tools instead of Xcode and to set XCTest as disabled by default if so.

grynspan:
@swift-ci please test

Sources/CoreCommands/Options.swift
```diff
@@ -598,28 +599,30 @@ public struct TestLibraryOptions: ParsableArguments {
...
         case .swiftTesting:
-            explicitlyEnableSwiftTestingLibrarySupport ?? explicitlyEnableExperimentalSwiftTestingLibrarySupport ?? `default`
+            return explicitlyEnableSwiftTestingLibrarySupport ?? explicitlyEnableExperimentalSwiftTestingLibrarySupport ?? `default`
         }
     }
 
     /// Test whether or not a given library is enabled.
-    public func isEnabled(_ library: TestingLibrary) -> Bool {
-        isEnabled(library, default: true)
+    public func isEnabled(_ library: TestingLibrary, swiftCommandState: SwiftCommandState) -> Bool {
```
compnerd, line 619: Similar.

Why does this function exist? Is there an ABI need for it? It seems better to just use a default value for `default`.

Sources/CoreCommands/Options.swift
```diff
@@ -598,28 +599,30 @@ public struct TestLibraryOptions: ParsableArguments {
...
 
     /// Test whether or not a given library was explicitly enabled by the developer.
-    public func isExplicitlyEnabled(_ library: TestingLibrary) -> Bool {
-        isEnabled(library, default: false)
-    }
-
-    /// The list of enabled testing libraries.
-    public var enabledTestingLibraries: [TestingLibrary] {
-        [.xctest, .swiftTesting].filter(isEnabled)
+    public func isExplicitlyEnabled(_ library: TestingLibrary, swiftCommandState: SwiftCommandState) -> Bool {
```
compnerd, line 624: Similar.

Sources/CoreCommands/Options.swift
```diff
@@ -598,28 +599,30 @@ public struct TestLibraryOptions: ParsableArguments {
           help: .private)
     public var explicitlyEnableExperimentalSwiftTestingLibrarySupport: Bool?
 
-    private func isEnabled(_ library: TestingLibrary, `default`: Bool) -> Bool {
+    private func isEnabled(_ library: TestingLibrary, `default`: Bool, swiftCommandState: SwiftCommandState) -> Bool {
```
compnerd, line 602: I think that `state` might be a better label than `swiftCommandState`. The latter doesn't really tell me anything more, and this is a private function in `CoreCommands`.

Sources/CoreCommands/Options.swift
```diff
@@ -598,28 +599,30 @@ public struct TestLibraryOptions: ParsableArguments {
...
+                return explicitlyEnableXCTestSupport
+            }
+            if let toolchain = try? swiftCommandState.getHostToolchain(),
+               toolchain.swiftSDK.xctestSupport == .supported {
+                return `default`
+            }
+            return false
         case .swiftTesting:
-            explicitlyEnableSwiftTestingLibrarySupport ?? explicitlyEnableExperimentalSwiftTestingLibrarySupport ?? `default`
+            return explicitlyEnableSwiftTestingLibrarySupport ?? explicitlyEnableExperimentalSwiftTestingLibrarySupport ?? `default`
```
compnerd, line 614: Is there a task to clean up the `explicitlyEnableExperimentalSwiftTestingLibrarySupport`?

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -256,8 +256,10 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
         var results = [TestRunner.Result]()
 
         // Run XCTest.
-        if options.testLibraryOptions.isEnabled(.xctest) {
-            // validate XCTest available on darwin based systems
+        if options.testLibraryOptions.isEnabled(.xctest, swiftCommandState: swiftCommandState) {
+            // Validate XCTest available on Darwin-based systems. If it's not available and we're hitting this code
```
compnerd, line 260:
```suggestion
            // Validate XCTest availability on Darwin-based systems. If it's not available and we're hitting this code
```

compnerd (approved):

Sources/CoreCommands/Options.swift
grynspan: This label is consistent with all other uses of this type in argument position.

Sources/CoreCommands/Options.swift
grynspan: I'm not sure what you're asking. It exists because, tautologically, we need it to exist so we can call it.

Are you asking why we have two separate calls that both call into the private function? Explicit intent: `isEnabled()` just asks "is the library enabled?" while `isExplicitlyEnabled()` asks "did the user explicitly pass `--enable-xctest`/`--enable-swift-testing`?"

Sources/CoreCommands/Options.swift
grynspan: Nope. It's going to stick around for a while.

Sources/Commands/SwiftTestCommand.swift
grynspan: The typo gives it character. :P

Sources/CoreCommands/Options.swift
compnerd: This is an overload of `isEnabled`, which only exists to provide a value for `default`. I'm asking why have the overload rather than just use a default value for the function that this overloads.

grynspan: I don't intend for the private overload to be part of this type's interface. The private function is just the common implementation of the two public(er) interfaces. I could just as easily call it `checkEnabledButUseThisDefaultValue()`. Or `commonImplForTheIsEnabledFunctions()` or something equally silly.

Sources/Commands/SwiftTestCommand.swift
grynspan: Leaving this alone until CI finishes.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

grynspan:
@swift-ci please test Windows

PR #7806, opened 2024-07-22
Rename `BuildParameters.Testing.Library`.

grynspan:
Since the aforementioned enum no longer has any association with `BuildParameters`, rename it to something less domain-specific and move it to the Basic module.

Since this is a type rename, there's no unit testing to be done. If it compiles, I didn't break it. :)


grynspan:
@swift-ci please test

MaxDesiatov (approved):

xedin (approved):

PR #7808, opened 2024-07-22
[6.0][PubGrub] Narrow down cases when pre-release decisions are marked inv…

xedin:
…alid

- Explanation:

  Pre-release decisions are valid if the derived range they were inferred from supports pre-release versions.

  Currently, after deriving a pre-release assignment, the solver would assert
  in some circumstances depending on order in which the dependencies are specified in the manifest.

  This change makes it possible to infer pre-release verison of a package even when intermediate derivations did not support pre-release versions.

  For example one package could declare `swift-syntax` dependency as `508.0.1` ..< `601.0.0` and another could narrow it to `600.0.0-latest` ..< `601.0.0`.

  Since the most constrained range in such case supports pre-releases a decision
  to use the lastest pre-release version of `600.0.0` should be valid if there
  is no `600.0.0` release yet.

  It's important to note that on each path a decision for a package is always
  made based on the last undecided term's requirements which means - based on the current semantics - if such derivation   doesn't support pre-release,
  even though intermediate ones did, the solver would select a released version.
  This PR is not aim to change that.

  There is a caveat which we cannot narrowly fix - if package depends on an exact
  pre-release version the inference is going to fail when there is another dependency
  to the same package that doesn't support pre-releases. For example:

  Package `A` depends on packages `B` and `C` as follows:

  `A` depends on `B` = `0.0.1-alpha` and `C` = `1.0.0` `B` version `0.0.1-alpha` has no dependencies
  `C` version `1.0.0` depends on `B` - `0.0.1` ..< `0.0.2`

  In this situation the solver would fail because `0.0.1-alpha` does not satisfy range - `0.0.1` ..< `0.0.2` inferred from `C`.

  - Delays version decisions for packages with pre-release ranges until there are no more packages without pre-release versions. This also means that if a previous decision narrows down the range to releases-only the package would be made a contender on the next step of the solver.

  - Relaxes the `isValidDecision` check for pre-release decisions but makes sure that the last derivation supports pre-release versions.

  In situations like
https://github.com/swiftlang/swift-package-manager/issues/7658 and https://github.com/swiftlang/swift-package-manager/issues/7643 the solver would no longer assert and would produce a solution with a pre-release version for `swift-syntax`.

- Main Branch PR: https://github.com/swiftlang/swift-package-manager/pull/7799

- Risk: Low (These changes only affect packages that have pre-release version dependencies in a very narrow circumstances where there is another dependency that isn't a pre-release).

- Reviewed By: @MaxDesiatov 

- Resolves: rdar://130112167 https://github.com/swiftlang/swift-package-manager/issues/7658 https://github.com/swiftlang/swift-package-manager/issues/7643

- Testing: Existing test-cases were modified and new tests were added.

(cherry picked from commit 43fd240bf7263734f1cd730e4e0dd7ae22b54085)

### DIFF
--- a/Sources/PackageGraph/Resolution/PubGrub/PubGrubDependencyResolver.swift
+++ b/Sources/PackageGraph/Resolution/PubGrub/PubGrubDependencyResolver.swift
@@ -760,7 +760,17 @@ public struct PubGrubDependencyResolver {
                 let start = DispatchTime.now()
                 let counts = try result.get()
                 // forced unwraps safe since we are testing for count and errors above
-                let pkgTerm = undecided.min { counts[$0]! < counts[$1]! }!
+                let pkgTerm = undecided.min {
+                    // Prefer packages that don't allow pre-release versions
+                    // to allow propagation logic to find dependencies that
+                    // limit the range before making any decisions. This means
+                    // that we'd always prefer release versions.
+                    if $0.supportsPrereleases != $1.supportsPrereleases {
+                        return !$0.supportsPrereleases
+                    }
+
+                    return counts[$0]! < counts[$1]!
+                }!
                 self.delegate?.willResolve(term: pkgTerm)
                 // at this point the container is cached
                 let container = try self.provider.getCachedContainer(for: pkgTerm.node.package)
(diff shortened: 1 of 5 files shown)


xedin:
@swift-ci please test

bnbarham (approved):

PR #7809, opened 2024-07-22
[6.0] Move the 'GitRepositoryProvider' to compare urls using their canonical representation, correctly accepting results where repositories only differ by '.git'

francescomikulis:
Cherry-pick https://github.com/swiftlang/swift-package-manager/pull/7741.

**Short Explanation**: Fixed an issue where repositories would be re-cloned each build rather than using the cache due to git validation errors.
**Detailed Explanation**: Validating the origin of checked-out repositories no longer ensures that the path extension (often .git) matches the repository's specifier, as different git clients inconsistently preserve the path extension of the remote. Moving to the CanonicalPackageURL also ensures that absolute paths are treated as urls with a file:// scheme, matching git's behavior.
**Scope**: Dependency resolution - helps re-use existing checkouts for repositories where the remotes only differ by path extension.
**Risk**: Ignoring inconsistencies in the path extension could lead to an inconsistent in-memory cache state, if subsystems are keeping references to un-canonicalized remotes as strings.
**Testing**: This change is accompanied by a series of new tests that ensure we preserve this behavior in the future.
**Issue**: rdar://129329704 
**Original PR**: https://github.com/swiftlang/swift-package-manager/pull/7741


francescomikulis:
@swift-ci please test

francescomikulis:

bnbarham (approved):

PR #7811, opened 2024-07-23
Revert "Fix: use bearer auth when pulling binaryTarget or packages from a package registry"

MaxDesiatov:
Reverts swiftlang/swift-package-manager#7662

It would be a breaking change if someone is using `token` as their username, and we don't have a migration path or any other means to clarify this breaking change to the users.


MaxDesiatov:
@swift-ci test

xedin (approved):

xedin:
@swift-ci please test

MaxDesiatov:
@swift-ci test macos

PR #7812, opened 2024-07-23
[Build] NFC: Compute per-module macro requirements on-demand

xedin:
Instead of having to go over all of the products and establish `module -> product` relationships during build plan construction let swift module build descriptions compute required macros based on their underlying modules as necessary.

### Motivation:

Build plan is required to pre-compute macro requirements but that's expensive and shouldn't be necessary.

### Modifications:

- Instead of passing macro products to `SwiftModuleBuildDescription` all we need to pass are build parameters to use for macros and the macros themselves could be found based on the underlying module.

### Result:

Build plan construction is slightly simpler and less expensive to compute, product and module descriptions could now be computed independently.


xedin:
@swift-ci please test

MaxDesiatov (approved):

PR #7815, opened 2024-07-24
Add `--type` argument to `swift package add-dependency`

hi2gage:
Fix issue https://github.com/swiftlang/swift-package-manager/issues/7738

Conversation regarding approach in previous PR #7769

Allow user to add `.package(path:)` and `.package(id, ...)` dependencies using `swift package add-dependency`

### Motivation:

I wanted to be able to add  `.package(path:)` using the command line and was unable to do so with the given parsing.

This PR adds `--type` argument allowing user to choose dependency type. 

### Modifications:

- Introduces `--type` argument which defaults to `url` 
   - accepts `url`, `path` and `registry`
- Uses `--type` argument to determine what type of dependency to create
- Adds `SwiftSyntax` for:
   - `PackageDependency.FileSystem`
   - `PackageDependency.Registry`
   - `PackageDependency.Registry.Requirement`
- Adds more unit test coverage
- Runs `SwiftFormat` in the files that were edited

### Result:

User can run the following commands

#### `Path` Dependency
```bash
swift package add-dependency /ChildPackage --type path
> Updating package manifest at Package.swift... done.
```
resulting in 
```swift
.package(path: "/ChildPackage")
```
---
#### `Registry` Dependency
```bash
swift package add-dependency scope.name --exact 1.0.0
> Updating package manifest at Package.swift... done.
```

resulting in
```swift
.package(id: "scope.name", exact: "1.0.0")
```
---

### DIFF
--- a/Sources/Commands/PackageCommands/AddDependency.swift
+++ b/Sources/Commands/PackageCommands/AddDependency.swift
@@ -24,7 +24,8 @@ import Workspace
 extension SwiftPackageCommand {
     struct AddDependency: SwiftCommand {
         package static let configuration = CommandConfiguration(
-            abstract: "Add a package dependency to the manifest")
+            abstract: "Add a package dependency to the manifest"
+        )
 
         @Argument(help: "The URL or directory of the package to add")
         var dependency: String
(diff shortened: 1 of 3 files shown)


Sources/Commands/PackageCommands/AddDependency.swift
```diff
@@ -50,33 +51,49 @@ extension SwiftPackageCommand {
         @Option(help: "Specify upper bound on the package version range (exclusive)")
         var to: Version?
 
+        @Option(help: "Specify dependency type (url, path, registry)")
+        var type: DependencyType = .url
+
+        enum DependencyType: String, Codable, ExpressibleByArgument {
```
rauhul, line 57: Add CaseIterable so the possible values are displayed by ArgumentParser.

Sources/Commands/PackageCommands/AddDependency.swift
```diff
@@ -50,33 +51,49 @@ extension SwiftPackageCommand {
         @Option(help: "Specify upper bound on the package version range (exclusive)")
         var to: Version?
 
+        @Option(help: "Specify dependency type (url, path, registry)")
```
rauhul, line 54: Drop the possible values from the help here, argument parser handles this.

rauhul (changes requested):

Sources/Commands/PackageCommands/AddDependency.swift
hi2gage: fixed

Sources/Commands/PackageCommands/AddDependency.swift
hi2gage: fixed

bnbarham (approved):
Thanks @hi2gage. LGTM - keeps the current default and allows an override. We can always make the default case smarter in the future.

@rauhul looks like your comments are addressed to?

bnbarham:
@swift-ci please test

rauhul:
> Thanks @hi2gage. LGTM - keeps the current default and allows an override. We can always make the default case smarter in the future.
> 
> @rauhul looks like your comments are addressed to?

Yup

hi2gage:
> > Thanks @hi2gage. LGTM - keeps the current default and allows an override. We can always make the default case smarter in the future.
> > @rauhul looks like your comments are addressed to?
> 
> Yup

@rauhul I think I still need an approval from you because of requested changes.

rauhul (approved):

PR #7817, opened 2024-07-24
Add back `--enable-experimental-swift-testing` to `swift build` as a no-op.

grynspan:
For the benefit of the VSCode Swift plugin, we need to keep this flag around (but as a no-op and hidden from documentation) until the plugin drops support for earlier Swift 6 prereleases.

Resolves rdar://132408655.


grynspan:
@swift-ci please test

MaxDesiatov (approved):

PR #7822, opened 2024-07-25
NFC: Resolve some `Sendable`-related warnings

MaxDesiatov:
Assorted fixes for sendability warnings.

### DIFF
--- a/Tests/BasicsTests/CancellatorTests.swift
+++ b/Tests/BasicsTests/CancellatorTests.swift
@@ -348,6 +348,7 @@ final class CancellatorTests: XCTestCase {
         struct Worker {
             func work()  {}
 
+            @Sendable
             func cancel() {
                 Thread.sleep(forTimeInterval: 5)
             }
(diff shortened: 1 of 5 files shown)


Tests/BasicsTests/CancellatorTests.swift
```diff
@@ -391,6 +392,7 @@ fileprivate struct Worker {
         return self.semaphore.wait(timeout: deadline)
     }
 
+    @Sendable
```
xedin, line 395: I wonder if we should try enabling InferSendableFromCaptures instead of these explicit attributes in few places including properties.

xedin (approved):

Tests/BasicsTests/CancellatorTests.swift
xedin: cc @MaxDesiatov, we can do that without experimental flags too.

MaxDesiatov: How would you do that without experimental flags? Does Swift 6 language mode include support for that?

xedin: It’s an upcoming feature flag.

PR #7825, opened 2024-07-25
[PackageDescription] Deprecated `swiftLanguageVersion` should result …

xedin:
…in `swiftLanguageMode` setting

Otherwise manifest loading is going to fail because `swiftLanguageVersion` has been removed.

Resolves: https://github.com/swiftlang/swift-package-manager/issues/7823
Resolves: rdar://132484168


xedin:
@swift-ci please test

MaxDesiatov (approved):

PR #7845, opened 2024-08-01
Remove `temp_await` [part 2]

AndrewHoos:
Remove usages of `temp_await`.

### Motivation:

`temp_await` is unsafe and it would be better to use true `async` methods.

### Modifications:

Replace 16 uses of `temp_await` with actual async methods.

### Result:

Safer more modern Swift code.

### DIFF
--- a/Sources/Commands/SwiftRunCommand.swift
+++ b/Sources/Commands/SwiftRunCommand.swift
@@ -124,10 +124,12 @@ public struct SwiftRunCommand: AsyncSwiftCommand {
         switch options.mode {
         case .repl:
             // Load a custom package graph which has a special product for REPL.
-            let graphLoader = {
-                try swiftCommandState.loadPackageGraph(
-                    explicitProduct: self.options.executable
-                )
+            let asyncUnsafeGraphLoader = {
+                try unsafe_await {
+                    try await swiftCommandState.loadPackageGraph(
+                        explicitProduct: self.options.executable
+                    )
+                }
             }
 
             // Construct the build operation.
@@ -136,7 +138,7 @@ public struct SwiftRunCommand: AsyncSwiftCommand {
                 explicitBuildSystem: .native,
                 traitConfiguration: .init(traitOptions: self.options.traits),
                 cacheBuildManifest: false,
-                packageGraphLoader: graphLoader
+                packageGraphLoader: asyncUnsafeGraphLoader
             )
 
             // Perform build.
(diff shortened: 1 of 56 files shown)


AndrewHoos:
@swift-ci test

AndrewHoos:
@swift-ci test

AndrewHoos:
@swift-ci test

AndrewHoos:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/Workspace/Workspace.swift
```diff
@@ -606,9 +606,9 @@ extension Workspace {
         revision: Revision? = nil,
         checkoutBranch: String? = nil,
         observabilityScope: ObservabilityScope
-    ) {
+    ) async {
```
MaxDesiatov, line 609: This doesn't account for reentrancy which is now possible with these functions becoming `async`. What happens when `edit` is called again before previous `edit` has finished?

IMO `Workspace` should get a `state` stored property that's asserted on to have `.idle` value and set to a different value indicating that the workspace is no longer idle when  `async` call is started.

MaxDesiatov (changes requested):

MaxDesiatov:
@swift-ci test windows

Sources/Commands/SwiftRunCommand.swift
```diff
@@ -124,10 +124,12 @@ public struct SwiftRunCommand: AsyncSwiftCommand {
         switch options.mode {
         case .repl:
             // Load a custom package graph which has a special product for REPL.
-            let graphLoader = {
-                try swiftCommandState.loadPackageGraph(
-                    explicitProduct: self.options.executable
-                )
+            let asyncUnsafeGraphLoader = {
+                try unsafe_await {
```
MaxDesiatov, line 128: IIUC `unsafe_await` behaves the same way that `temp_await` does, it blocks the current thread? If so, what are we gaining from this change here?

MaxDesiatov (approved):
Concerns addressed in an offline conversation (points of feedback to be addressed in immediately following PRs).

Thanks!

Sources/Commands/SwiftRunCommand.swift
AndrewHoos: This PR lifts the unsafe ness to the highest possible point in the call stack so the benefits are:
1- The unsafe behavior is now explicitly labeled unsafe
2- Multiple unsafe blocking waits are replaced with a single one
In a future PR I make the graph loader async and remove this unsafe_await, but this seemed like a good intermediate spot on the journey to remove all of the unsafe_await style behavior. I also actually have a future PR which will make this async and reduce the usage of unsafe_await to a single call that happens off of the structured concurrency

PR #7849, opened 2024-08-01
Restore XML output for Swift Testing when `--disable-xctest` is passed.

grynspan:
A previous PR, #7796, avoided Swift Testing stomping on XML output from XCTest when both libraries were running tests. However, there's a bug in the implementation where if you pass `--disable-xctest`, we don't end up telling Swift Testing to write any XML at all. This PR fixes that.

Tested with:

```sh
swift build
$(pwd)/.build/debug/swift-test --package-path /Volumes/Dev/Source/swift-testing --disable-xctest --xunit-output /dev/stdout
```

To confirm that XML output was produced for Swift Testing's test target (as an exemplar package that uses Swift Testing.)

### DIFF
--- a/Sources/Commands/SwiftTestCommand.swift
+++ b/Sources/Commands/SwiftTestCommand.swift
@@ -469,15 +469,17 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
             }
             additionalArguments += commandLineArguments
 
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

xedin (approved):

PR #7852, opened 2024-08-02
Remove `temp_await` from swift-bootstrap

AndrewHoos:
Remove temp_await from swift-bootstrap

### Motivation:

`temp_await` is unsafe and it would be better to use true async methods

### Modifications:

Replace `temp_await` with `unsafe_await` in swift-bootstrap (temporarily)

### Result:

Safer more modern Swift code


AndrewHoos:
@swift-ci test

MaxDesiatov (approved):

PR #7854, opened 2024-08-02
[BuildPlan] Switch build plan construction to use graph traversal

xedin:
### Motivation:

This way we can compute destinations as we go instead of relying on `buildTriple` computed during graph construction.

### Modifications:

- Adds a new static method on `BuildPlan` to traverse modules graph and compute destinations for each product and module.
- Switches `BuildPlan.init` from using `allProducts` and `allModules` flat lists over the new incremental destination computation.

### Result:

Removes one of the last places were `buildTriple` is required.

### DIFF
--- a/Sources/Basics/Graph/GraphAlgorithms.swift
+++ b/Sources/Basics/Graph/GraphAlgorithms.swift
@@ -28,7 +28,7 @@ import struct OrderedCollections.OrderedSet
 public func depthFirstSearch<T: Hashable>(
     _ nodes: [T],
     successors: (T) throws -> [T],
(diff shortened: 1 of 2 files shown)


xedin:
Next up we need to switch `productMap` and `targetMap` in `BuildPlan` to use `IdentifiableSet` and remove use of `ResolvedModule.ID` cannot do that without some sourcekit-lsp changes first.

xedin:
@swift-ci please test

MaxDesiatov (approved):
Thank you! 🙏

PR #7855, opened 2024-08-05
[SourceKitLSPAPI] Adjust `BuildDescription` APIs use use build plan

xedin:
These changes represent one my step towards the goal of removing `buildTriple` from `Resolved{Product, Module}`.

### Motivation:

Instead using modules graph and lookup descriptions up, let's provide API to fetch all available modules which has all of the information needed by the sourcekit-lsp at the moment. In the future we could add an additional `getDescription` method if it becomes necessary.

### Modifications:

- Add `paths` and `root` to `BuildTarget` to expose information necessary to construct some sourcekit-lsp internal data structures while reloading packages.

- Remove `BuildDescription.getBuildTarget`
- Replace `allTargetsInTopologicalOrder` with `allModules`.

### Result:

Package reloading that happens in sourcekit-lsp is now based purely on build plan information which facilitates `buildTriple` removal from `Resolved{Product, Module}`.

### DIFF
--- a/Sources/SourceKitLSPAPI/BuildDescription.swift
+++ b/Sources/SourceKitLSPAPI/BuildDescription.swift
@@ -29,7 +29,14 @@ internal import class PackageModel.UserToolchain
 public typealias BuildTriple = PackageGraph.BuildTriple
 
 public protocol BuildTarget {
+    /// The root of the sources.
+    var root: URL { get }
+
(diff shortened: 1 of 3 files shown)


xedin:
https://github.com/swiftlang/sourcekit-lsp/pull/1614
@swift-ci please test

PR #7861, opened 2024-08-06
[PackageModel] SwiftTesting: Fix directory existence check to use tes…

xedin:
…ting directory

The original check was using "lib" directory which would always be there, it should be using "testing" directory instead which might not actually be there always.


xedin:
@swift-ci please test

bnbarham (approved):

PR #7862, opened 2024-08-06
Refactored PackageFingerprintStorage to not use callbacks or async/await

AndrewHoos:
Refactored PackageFingerprintStorage to not use callbacks or async/await

### Motivation:

None of the implementations of PackageFingerprintStorage were actually async so completionHandlers were completely unnecessary. It is much simpler to make these traditional throwing methods and allow the caller to handle threading concerns themselves

### Modifications:

Removed callback and queues from PackageFingerprintStorage APIs
Remove async wrappers of PackageFingerprintStorage APIs
Recursively refactored callers of PackageFingerprintStorage that were actually sync to have sync signatures

### Result:

Significantly simpler and more readable code that can more easily be refactored to async when needed

### DIFF
--- a/Sources/PackageFingerprint/FilePackageFingerprintStorage.swift
+++ b/Sources/PackageFingerprint/FilePackageFingerprintStorage.swift
@@ -35,17 +35,12 @@ public struct FilePackageFingerprintStorage: PackageFingerprintStorage {
     public func get(
         package: PackageIdentity,
         version: Version,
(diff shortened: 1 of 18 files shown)


AndrewHoos:
@swift-ci test

AndrewHoos:
@swift-ci test windows

MaxDesiatov (approved):

PR #7863, opened 2024-08-07
[Build/Commands] Switch `SymbolGraphExtract` to use build plan

xedin:
Yet another step on the path to remove `buildTriple` from `Resolved{Product, Module}`.

### Motivation:

Instead of trying to lookup modules in the module graph and then going the build plan for their descriptions, symbol-graph-extract should refer to the build plan directly.

### Modifications:

- Adds a new protocol `ModuleBuildDescription` to SPMBuildCore
- Conforms `Build.ModuleBuildDescription` to `SPMCore.ModuleBuildDescription`
- Adds a new field to BuildPlan and its protocol to expose build modules
- Uses buildModules to implement symbol graph extraction as a replacement for module graph + build plan lookups.

### Result:

One less place that replies on `buildTriple` which brings us closer to its eventual removal.

### DIFF
--- a/Sources/Build/BuildDescription/ModuleBuildDescription.swift
+++ b/Sources/Build/BuildDescription/ModuleBuildDescription.swift
@@ -12,10 +12,12 @@
 
 import Basics
 import struct PackageGraph.ResolvedModule
+import struct PackageGraph.ResolvedPackage
 import struct PackageModel.Resource
 import struct PackageModel.ToolsVersion
 import struct SPMBuildCore.BuildToolPluginInvocationResult
 import struct SPMBuildCore.BuildParameters
+import protocol SPMBuildCore.ModuleBuildDescription
 
 public enum BuildDescriptionError: Swift.Error {
     case requestedFileNotPartOfTarget(targetName: String, requestedFilePath: AbsolutePath)
@@ -25,7 +27,7 @@ public enum BuildDescriptionError: Swift.Error {
 public typealias TargetBuildDescription = ModuleBuildDescription
 
 /// A module build description which can either be for a Swift or Clang module.
-public enum ModuleBuildDescription {
+public enum ModuleBuildDescription: SPMBuildCore.ModuleBuildDescription {
     /// Swift target description.
     case swift(SwiftModuleBuildDescription)
 
@@ -73,6 +75,19 @@ public enum ModuleBuildDescription {
         }
     }
 
+    public var package: ResolvedPackage {
+        switch self {
+        case .swift(let description):
+            description.package
+        case .clang(let description):
+            description.package
+        }
+    }
+
+    public var module: ResolvedModule {
+        self.target
+    }
+
     /// Paths to the binary libraries the target depends on.
     var libraryBinaryPaths: Set<AbsolutePath> {
         switch self {
@@ -101,7 +116,7 @@ public enum ModuleBuildDescription {
         }
     }
 
-    var buildParameters: BuildParameters {
+    public var buildParameters: BuildParameters {
         switch self {
         case .swift(let buildDescription):
             return buildDescription.buildParameters
@@ -121,7 +136,7 @@ public enum ModuleBuildDescription {
 
     /// Determines the arguments needed to run `swift-symbolgraph-extract` for
     /// this module.
-    package func symbolGraphExtractArguments() throws -> [String] {
+    public func symbolGraphExtractArguments() throws -> [String] {
         switch self {
         case .swift(let buildDescription): try buildDescription.symbolGraphExtractArguments()
         case .clang(let buildDescription): try buildDescription.symbolGraphExtractArguments()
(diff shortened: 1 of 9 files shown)


xedin:
@swift-ci please test

Sources/Build/BuildDescription/ModuleBuildDescription.swift
```diff
@@ -121,7 +136,7 @@ public enum ModuleBuildDescription {
 
     /// Determines the arguments needed to run `swift-symbolgraph-extract` for
     /// this module.
-    package func symbolGraphExtractArguments() throws -> [String] {
+    public func symbolGraphExtractArguments() throws -> [String] {
```
rauhul, line 139: why public?

Sources/Build/BuildPlan/BuildPlan.swift
```diff
@@ -216,6 +216,10 @@ public class BuildPlan: SPMBuildCore.BuildPlan {
         AnySequence(self.productMap.values.map { $0 as SPMBuildCore.ProductBuildDescription })
     }
 
+    public var buildModules: AnySequence<SPMBuildCore.ModuleBuildDescription> {
```
rauhul, line 219: would this be better as `some Sequence<T>`

Sources/Commands/PackageCommands/DumpCommands.swift
```diff
@@ -70,14 +70,14 @@ struct DumpSymbolGraph: AsyncSwiftCommand {
 
         // Run the tool once for every library and executable target in the root package.
         let buildPlan = try buildSystem.buildPlan
+        let modulesGraph = try await buildSystem.getPackageGraph()
         let symbolGraphDirectory = buildPlan.destinationBuildParameters.dataPath.appending("symbolgraph")
-        let targets = try await buildSystem.getPackageGraph().rootPackages.flatMap{ $0.modules }.filter{ $0.type == .library }
-        for target in targets {
-            print("-- Emitting symbol graph for", target.name)
+        for description in buildPlan.buildModules
+            where description.module.type == .library && modulesGraph.rootPackages[description.package.id] != nil
```
rauhul, line 76: imo this is easier to read as a `guard continue`

Sources/Commands/Utilities/PluginDelegate.swift
```diff
@@ -398,25 +398,34 @@ final class PluginDelegate: PluginInvocationDelegate {
...
+        // the destination is, this logic prefers "target" over "host" because that's
+        // historically how this was setup. Ideally we should be building for both "host"
+        // and "target" if module is configured for them but that would require changing
+        // `PluginInvocationSymbolGraphResult` to carry multiple directories.
+        let description: ModuleBuildDescription? = if let targetDescription = try lookupDescription(for: targetName, destination: .target) {
+            targetDescription
+        } else if let hostDescription = try lookupDescription(for: targetName, destination: .host) {
+            hostDescription
+        } else {
+            nil
```
rauhul, line 420: throw here?

Sources/Build/BuildDescription/ModuleBuildDescription.swift
xedin: Because it’s a witness to a public protocol requirement now

Sources/Build/BuildPlan/BuildPlan.swift
xedin: The other member uses AnySequence so I used the same for consistency.

Sources/Commands/Utilities/PluginDelegate.swift
xedin: Cannot throw from if expression I think, I will check if that’s the case.

Sources/Commands/PackageCommands/DumpCommands.swift
xedin: I had it like that initially but this `where` clause is not too complex…

Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
```diff
@@ -80,6 +80,19 @@ extension ProductBuildDescription {
...
+    /// The package the module belongs to.
+    var package: ResolvedPackage { get }
+
+    /// The underlying module this description is for.
+    var module: ResolvedModule { get }
+
+    /// The build parameters.
+    var buildParameters: BuildParameters { get }
+
+    func symbolGraphExtractArguments() throws -> [String]
```
rauhul, line 93: could we add a fixme to remove this?

MaxDesiatov (approved):
Overall LGTM, but I second Rauhul's comments, and CI issues need to be resolved.

xedin:
@swift-ci please test

xedin:
@swift-ci please test Windows platforms

rauhul (approved):

Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
xedin: ✅

PR #7865, opened 2024-08-07
[Build] Use a separate scope while constructing a build plan

xedin:
### Motivation:

We need to use a separate inner scope to capture any problems during build plan construction and make a decision to stop based on that. The current logic checks the scope that is passed in which tracks errors emitted elsewhere.

### Modifications:

- Passes an child scope called "Planning" to the `computeDestinations` functions that is used to construct the plan.

### Result:

sourcekit-lsp no longer fails if there is no Package.resolved - https://github.com/swiftlang/sourcekit-lsp/pull/1618


xedin:
https://github.com/swiftlang/sourcekit-lsp/pull/1618
@swift-ci please test

MaxDesiatov (approved):
Thanks! 👍

PR #7867, opened 2024-08-08
[Build] NFC: Rename `ModuleBuildDescription` property from `target` t…

xedin:
…o `module`

Looks like this was missed by previous refactorings.


xedin:
@swift-ci please test

MaxDesiatov (approved):
Nice catch, thanks!

PR #7871, opened 2024-08-10
Allow relative path with `swift package add-dependency` command

hi2gage:
Enable passing relative path into `swift package add-dependency`

### Motivation:
Currently only `AbsolutePath`s are supported with `--type path` flag
```bash
swift package add-dependency /packagePath --type path
```
There is a need to support `RelativePath` because it's not uncommon to have a package structure inside of an xcodeproj as shown below.

```bash
LocalPackages
├── ChildPackage
│   ├── .gitignore
│   ├── .swiftpm
│   ├── Package.swift
│   ├── Sources
│   └── Tests
└── ParentPackage
    ├── .build
    ├── .gitignore
    ├── .swiftpm
    ├── Package.swift
    ├── Sources
    └── Tests
```

If we want to open `ParentPackage` by itself, it will not be able to resolve that package.

This pr allows for the user to add a package with a `RelativePath` path via the `swift package add-dependency --type path` command.

Access level for 
- `.package(name:url:requirement:traits:)`
- `.package(name:url:requirement:)`
- `.package(id:requirement:traits:)`

were upgraded from `private` to `package` allowing access to these functions from the `PackageCommands` module.

### Modifications:
- Enable passing relative path into `swift package add-dependency`
- Add unit test coverage

### Result:
Both of the following commands are valid

```bash
swift package add-dependency ../relative --type path
```

```bash
swift package add-dependency /absolute --type path
```

### DIFF
--- a/Sources/PackageModelSyntax/AddPackageDependency.swift
+++ b/Sources/PackageModelSyntax/AddPackageDependency.swift
@@ -18,7 +18,7 @@ import SwiftSyntax
 import SwiftSyntaxBuilder
 
 /// Add a package dependency to a manifest's source code.
-public struct AddPackageDependency {
+public enum AddPackageDependency {
     /// The set of argument labels that can occur after the "dependencies"
     /// argument in the Package initializers.
     ///
@@ -28,13 +28,13 @@ public struct AddPackageDependency {
         "targets",
         "swiftLanguageVersions",
         "cLanguageStandard",
-        "cxxLanguageStandard"
+        "cxxLanguageStandard",
     ]
 
     /// Produce the set of source edits needed to add the given package
     /// dependency to the given manifest file.
     public static func addPackageDependency(
-        _ dependency: PackageDependency,
+        _ dependency: MappablePackageDependency.Kind,
         to manifest: SourceFileSyntax
     ) throws -> PackageEditResult {
         // Make sure we have a suitable tools version in the manifest.
@@ -50,19 +50,19 @@ public struct AddPackageDependency {
 
         return PackageEditResult(
             manifestEdits: [
-                .replace(packageCall, with: newPackageCall.description)
+                .replace(packageCall, with: newPackageCall.description),
             ]
         )
     }
 
     /// Implementation of adding a package dependency to an existing call.
     static func addPackageDependencyLocal(
-        _ dependency: PackageDependency,
+        _ dependency: MappablePackageDependency.Kind,
         to packageCall: FunctionCallExprSyntax
     ) throws -> FunctionCallExprSyntax {
         try packageCall.appendingToArrayArgument(
             label: "dependencies",
-            trailingLabels: Self.argumentLabelsAfterDependencies,
+            trailingLabels: self.argumentLabelsAfterDependencies,
             newElement: dependency.asSyntax()
         )
     }
--- a/Sources/PackageModelSyntax/AddTarget.swift
+++ b/Sources/PackageModelSyntax/AddTarget.swift
@@ -18,7 +18,7 @@ import SwiftSyntaxBuilder
 import struct TSCUtility.Version
 
 /// Add a target to a manifest's source code.
-public struct AddTarget {
+public enum AddTarget {
     /// The set of argument labels that can occur after the "targets"
     /// argument in the Package initializers.
     ///
@@ -27,7 +27,7 @@ public struct AddTarget {
     private static let argumentLabelsAfterTargets: Set<String> = [
         "swiftLanguageVersions",
         "cLanguageStandard",
-        "cxxLanguageStandard"
+        "cxxLanguageStandard",
     ]
 
     /// The kind of test harness to use. This isn't part of the manifest
@@ -85,7 +85,7 @@ public struct AddTarget {
             target.dependencies.append(contentsOf: macroTargetDependencies)
 
         default:
-            break;
+            break
         }
 
         var newPackageCall = try packageCall.appendingToArrayArgument(
@@ -103,7 +103,7 @@ public struct AddTarget {
         guard let outerDirectory else {
             return PackageEditResult(
                 manifestEdits: [
-                    .replace(packageCall, with: newPackageCall.description)
+                    .replace(packageCall, with: newPackageCall.description),
                 ]
             )
         }
@@ -114,7 +114,7 @@ public struct AddTarget {
         var auxiliaryFiles: AuxiliaryFiles = []
 
         // Add the primary source file. Every target type has this.
-        addPrimarySourceFile(
+        self.addPrimarySourceFile(
             outerPath: outerPath,
             target: target,
             configuration: configuration,
@@ -125,7 +125,7 @@ public struct AddTarget {
         var extraManifestEdits: [SourceEdit] = []
         switch target.type {
         case .macro:
-            addProvidedMacrosSourceFile(
+            self.addProvidedMacrosSourceFile(
                 outerPath: outerPath,
                 target: target,
                 to: &auxiliaryFiles
@@ -135,7 +135,7 @@ public struct AddTarget {
                 newPackageCall = try AddPackageDependency
                     .addPackageDependencyLocal(
                         .swiftSyntax(
-                          configuration: installedSwiftPMConfiguration
+                            configuration: installedSwiftPMConfiguration
                         ),
                         to: newPackageCall
                     )
@@ -149,7 +149,7 @@ public struct AddTarget {
                             .positionAfterSkippingLeadingTrivia
                         extraManifestEdits.append(
                             SourceEdit(
-                                range: insertPos..<insertPos,
+                                range: insertPos ..< insertPos,
                                 replacement: newImport
                             )
                         )
@@ -158,12 +158,12 @@ public struct AddTarget {
                 }
             }
 
-        default: break;
+        default: break
         }
 
         return PackageEditResult(
             manifestEdits: [
-                .replace(packageCall, with: newPackageCall.description)
+                .replace(packageCall, with: newPackageCall.description),
             ] + extraManifestEdits,
             auxiliaryFiles: auxiliaryFiles
         )
@@ -182,20 +182,18 @@ public struct AddTarget {
         )
 
         // Introduce imports for each of the dependencies that were specified.
-        var importModuleNames = target.dependencies.map {
-            $0.name
-        }
+        var importModuleNames = target.dependencies.map(\.name)
 
         // Add appropriate test module dependencies.
         if target.type == .test {
             switch configuration.testHarness {
-            case .none: 
+            case .none:
                 break
 
             case .xctest:
                 importModuleNames.append("XCTest")
 
-            case .swiftTesting: 
+            case .swiftTesting:
                 importModuleNames.append("Testing")
             }
         }
@@ -306,13 +304,13 @@ public struct AddTarget {
     }
 }
 
-fileprivate extension TargetDescription.Dependency {
+extension TargetDescription.Dependency {
     /// Retrieve the name of the dependency
-    var name: String {
+    fileprivate var name: String {
         switch self {
         case .target(name: let name, condition: _),
-            .byName(name: let name, condition: _),
-            .product(name: let name, package: _, moduleAliases: _, condition: _):
+             .byName(name: let name, condition: _),
+             .product(name: let name, package: _, moduleAliases: _, condition: _):
             name
         }
     }
@@ -320,11 +318,11 @@ fileprivate extension TargetDescription.Dependency {
 
 /// The array of auxiliary files that can be added by a package editing
 /// operation.
-fileprivate typealias AuxiliaryFiles = [(RelativePath, SourceFileSyntax)]
+private typealias AuxiliaryFiles = [(RelativePath, SourceFileSyntax)]
 
-fileprivate extension AuxiliaryFiles {
+extension AuxiliaryFiles {
     /// Add a source file to the list of auxiliary files.
-    mutating func addSourceFile(
+    fileprivate mutating func addSourceFile(
         path: RelativePath,
         sourceCode: SourceFileSyntax
     ) {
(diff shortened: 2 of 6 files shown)


hi2gage:
@bnbarham Just wanted to bump this.

bnbarham (commented):
Thanks @hi2gage! Definitely valuable to support relative paths here.

> Due to access level and not wanting to make certain enums public, I opted to use MappablePackageDependency instead of Package.Dependency to represent the values coming from the CLI command.

Which enums would those be? Ideally `MappablePackageDependency` would not be public, it really shouldn't be visible to commands.

hi2gage:
> Thanks @hi2gage! Definitely valuable to support relative paths here.
> 
> > Due to access level and not wanting to make certain enums public, I opted to use MappablePackageDependency instead of Package.Dependency to represent the values coming from the CLI command.
> 
> Which enums would those be? Ideally `MappablePackageDependency` would not be public, it really shouldn't be visible to commands.


I believe it was to keep `Package.Dependency.SourceControlRequirement` and  `Package.Dependency.RegistryRequirement` private. See comment:
```swift
// intentionally private to hide enum detail
private static func package(
        name: String? = nil,
        url: String,
        requirement: Package.Dependency.SourceControlRequirement
    ) -> Package.Dependency {
        return .init(name: name, location: url, requirement: requirement, traits: nil)
    }
```

but looking back over this again I realized this was written prior to the `package` access level being introduced. Any reason to not move these 3 functions from `private` to `package`?

bnbarham:
> I thought I would need to make Package.Dependency.SourceControlRequirement and Package.Dependency.RegistryRequirement public, but I avoided doing so because of these comments:

Aren't they already 🤔? I assume the issue was that there's no way to create a `Dependency` with them today, but couldn't you call the existing initializers (which end up mapping to these)? 

> Any reason to not move these 3 functions from private -> package?

If they need to be (depending on the above) - they just can't be public as this is the package manifest API.

hi2gage:
Hey @bnbarham I just pushed up the changes to use `Package.Dependency` so we can talk specifics.

> > I thought I would need to make Package.Dependency.SourceControlRequirement and Package.Dependency.RegistryRequirement public, but I avoided doing so because of these comments:
> 
> Aren't they already 🤔?

Yes they are public 🤦I meant to say:
I thought I would need to make the Dependency initializers that contain those enums public. Not the enums themselves public. 



> I assume the issue was that there's no way to create a `Dependency` with them today, but couldn't you call the existing initializers (which end up mapping to these)?

Technically I could call the existing initializers, but it requires some funky switching. The changes I just pushed uses the existing initializers. So let me know what you think about that.

This being said if I made those initializers `package` access level I could take this code:

```swift
let packageDependency: PackageDescription.Package.Dependency
switch (firstRequirement, to) {
case (let .exact(version), nil):
    packageDependency = .package(url: self.dependency, exact: version)
case (let .range(range), let to?):
    packageDependency = .package(url: self.dependency, (range.lowerBound ..< to))
case (let .range(range), nil):
    packageDependency = .package(url: self.dependency, range)
case (let .revision(revision), nil):
    packageDependency = .package(url: self.dependency, revision: revision)
case (let .branch(branch), nil):
    packageDependency = .package(url: self.dependency, branch: branch)
case (.branch, _?), (.revision, _?), (.exact, _?):
    throw StringError("--to can only be specified with --from or --up-to-next-minor-from")
case (_, _):
    throw StringError("unknown requirement")
}
```
-> 
```swift
let packageDependency2: PackageDescription.Package.Dependency = .package(
    url: self.dependency,
    requirement: requirement,
    traits: []
)
```

> 
> > Any reason to not move these 3 functions from private -> package?
> 
> If they need to be (depending on the above) - they just can't be public as this is the package manifest API.

Agreed, they can't be public, making them package would be helpful but we can get around it if you want to keep those initializers locked down.

bnbarham:
I'd be fine making them `package`, especially to avoid that `switch` 😅 . @dschaefer2 / @plemarquand any opinions?

hi2gage:
> I'd be fine making them `package`, especially to avoid that `switch` 😅 . @dschaefer2 / @plemarquand any opinions?

Great! I just pushed up changes making those `package`.

hi2gage:
@bnbarham Just wanted to bump this, let me know if there is anything I can to help move this PR along.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:
@hi2gage that did the trick. Running the tests now. I'm a little worried about adding the PackageDescription dependency since it was really only intended for the package manifest processing, but it actually makes sense how you're using it.

This is actually pretty timely. We've been working at improving support for monorepos where relative packages are common. This will help setting them up.

dschaefer2:
Hmm, the toolchain build in the smoke tests and the Windows builds are failing. Look like the new dependency on SwiftSyntax requires a change to the CMake files (which are used to build swiftpm in these environments).

bnbarham:
> Hmm, the toolchain build in the smoke tests and the Windows builds are failing. Look like the new dependency on SwiftSyntax requires a change to the CMake files (which are used to build swiftpm in these environments).

It's just that `PackageDescription+Syntax.swift` needs to be added to the cmake build (and the old `PackageDependency+Syntax.swift` removed).

hi2gage:
> > Hmm, the toolchain build in the smoke tests and the Windows builds are failing. Look like the new dependency on SwiftSyntax requires a change to the CMake files (which are used to build swiftpm in these environments).
> 
> It's just that `PackageDescription+Syntax.swift` needs to be added to the cmake build (and the old `PackageDependency+Syntax.swift` removed).

Just pushed the changes to the cmake build

hi2gage:
> @hi2gage that did the trick. Running the tests now. I'm a little worried about adding the PackageDescription dependency since it was really only intended for the package manifest processing, but it actually makes sense how you're using it.
> 
> This is actually pretty timely. We've been working at improving support for monorepos where relative packages are common. This will help setting them up.

Yup, we have a monorepo at work so that's why I was motivated to fix this problem. I'm glad that I'm able to help out. If there are other issues / chunks of work in this realm let me know, I've really enjoyed contributing to SPM.

bnbarham:
@swift-ci please test

hi2gage:
@swift-ci please test windows

dschaefer2:
@swift-ci please test windows

hi2gage:
@swift-ci please test windows

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test windows

dschaefer2:
I'm not sure why the Windows build is breaking. It uses the CMakeLists file.

hi2gage:
> I'm not sure why the Windows build is breaking. It uses the CMakeLists file.

Yeah me neither I thought that bc39a72 would have fixed this error.
```
C:\Users\swift-ci\jenkins\workspace\swiftpm-PR-windows\swiftpm\Sources\PackageModelSyntax\AddPackageDependency.swift:14:8: error: no such module 'PackageDescription'
```

Is there a windows VM image or sorts to run this locally?

dschaefer2:
Looks like a missing dependency in the CMakeLists.txt file.

dschaefer2:
I'm trying on my Windows machine. There is a Utilities/bootstrap utility that builds using CMake. I couldn't reproduce it on my Mac, which is surprising.

Sources/PackageModelSyntax/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ target_link_libraries(PackageModelSyntax PUBLIC
   Basics
   PackageLoading
   PackageModel
+  ProductDescription
```
dschaefer2, line 27: I'm confused by this. There is no library called ProductDescription. Did you mean PackageDescription here?

hi2gage: Yeah not sure what I was thinking here, I mixed these up. Just pushed change to `PackageDescription`

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

hi2gage:
@swift-ci please test windows

dschaefer2:
Don't worry about the branching. We'll work to get this into 6.1. Our CI has been under the weather the last few days.

dschaefer2:
Actually, they're going to set up an auto merge of main to releases/6.1 so it'll just go there automatically. I think they just wanted to branch the compiler now, which makes sense. It needs time to settle down.

bnbarham:
@swift-ci please test

hi2gage:
Sounds good! Any further action needed from me on this PR to get those windows tests passing?

dschaefer2:
@swift-ci please test windows

dschaefer2:
I am a little worried about adding the dependency from SwiftPM to PackageDescription. It seems a lot of effort was expended in the past to avoid that, including duplicating most of the types, which I agree is pretty annoying. I assume that was to keep it only for the package manifest library.

@bnbarham any thoughts there? Any toolchain build issue we may run into?

bnbarham:
> @bnbarham any thoughts there? Any toolchain build issue we may run into?

@neonichu probably has the best background there

dschaefer2:
Hmm, the Windows build is failing with the missing dependency. It builds with CMake as well.

Sorry for the delayed responses. Too many things on the go at once :(.

hi2gage:
No worries at all. Holidays are a busy time for me. The more I think about this the more I think that making this as DRY as possible might more pain than it's worth.

Looking at [PackageDescription+Syntax.swift](https://github.com/swiftlang/swift-package-manager/pull/7871/files#diff-e5058a29c17f6875c1edb83f22f5e21478cfcc2acfbf6ca007633b1137f024eb) really all we need is a model that resents  a dependency that is added. 

Now it would have been convenient to use `PackageDescription.Package.Dependency` but maybe we just create a new model that will represent a dependency internally?

I propose we add a lightweight struct inside `PackageModelSyntax` that presents a dependence. Since `PackageModelSyntax` is a dependency of `Commands` then we can create this new model directly inside of `SwiftPackageCommand.AddDependency`

Thoughts on making this a little less DRY?

### Example (Just copied the main properties that are required for this use case):
```swift
public struct Dependency {
    /// The type of dependency.
    @available(_PackageDescription, introduced: 5.6)
    public enum Kind {
        /// A dependency located at the given path.
        /// - Parameters:
        ///    - name: The name of the dependency.
        ///    - path: The path to the dependency.
        case fileSystem(name: String?, path: String)
        /// A dependency based on a source control requirement.
        ///  - Parameters:
        ///    - name: The name of the dependency.
        ///    - location: The Git URL of the dependency.
        ///    - requirement: The version-based requirement for a package.
        case sourceControl(name: String?, location: String, requirement: SourceControlRequirement)
        /// A dependency based on a registry requirement.
        /// - Parameters:
        ///   - id: The package identifier of the dependency.
        ///   - requirement: The version based requirement for a package.
        case registry(id: String, requirement: RegistryRequirement)
    }

    /// A description of the package dependency.
    @available(_PackageDescription, introduced: 5.6)
    public let kind: Kind
}

// MARK: - SourceControlRequirement

extension Dependency {
    @available(_PackageDescription, introduced: 5.6)
    public enum SourceControlRequirement {
        /// An exact version based requirement.
        case exact(Version)
        /// A requirement based on a range of versions.
        case range(Range<Version>)
        /// A commit based requirement.
        case revision(String)
        /// A branch based requirement.
        case branch(String)
    }
}

// MARK: - RegistryRequirement

extension Dependency {
    public enum RegistryRequirement {
        /// A requirement based on an exact version.
        case exact(Version)
        /// A requirement based on a range of versions.
        case range(Range<Version>)
    }
}

// MARK: - Version

public struct Version: Sendable {

    /// The major version according to the semantic versioning standard.
    public let major: Int

    /// The minor version according to the semantic versioning standard.
    public let minor: Int

    /// The patch version according to the semantic versioning standard.
    public let patch: Int

    /// The pre-release identifier according to the semantic versioning standard, such as `-beta.1`.
    public let prereleaseIdentifiers: [String]

    /// The build metadata of this version according to the semantic versioning standard, such as a commit hash.
    public let buildMetadataIdentifiers: [String]

```

the `Command`

dschaefer2:
Agreed. Does the PackageModel module contain types you could use?

hi2gage:
> Agreed. Does the PackageModel module contain types you could use?

Actually yes looking through the `PackageModel` module again `MappablePackageDependency.Kind` works perfectly. 

Pushed up a new commit with that change. Tests are passing locally and no dependencies were added. So hopefully that passes the smell test.

Sources/PackageModelSyntax/AddPackageDependency.swift
```diff
@@ -18,7 +18,7 @@ import SwiftSyntax
 import SwiftSyntaxBuilder
 
 /// Add a package dependency to a manifest's source code.
-public struct AddPackageDependency {
```
hi2gage, line 21: swiftformat made this change

Sources/PackageModelSyntax/AddTarget.swift
```diff
@@ -18,7 +18,7 @@ import SwiftSyntaxBuilder
 import struct TSCUtility.Version
 
 /// Add a target to a manifest's source code.
-public struct AddTarget {
```
hi2gage, line 21: swiftformat made this change

dschaefer2:
@swift-ci please test.

dschaefer2:
@swift-ci please test windows

dschaefer2:
So much cleaner. Thank you!

dschaefer2:
@swift-ci please test windows

hi2gage:
Hey @dschaefer2 just wanted to bump this. Anything else needed from me?

dschaefer2:
Sorry about that. Things got pretty busy. Let me kick off the tests again.

dschaefer2:
@swift-ci please test

dschaefer2:
Finally back to this. Looks good. Thank you!

PR #7866, opened 2024-08-08
Add (partial) support for building on Windows using SwiftPM

jakepetroules:
This adds a dependency on the new swift-toolchain-sqlite package, which allows swift-package-manager to build using swift build on Windows (including tests) without any additional flags or preinstalled dependencies.

It's "partial" because there is still one final blocker, which is that linking fails because we exceed the 65k exported symbol limit due to limitations in the build system, which can be addressed separately.

Issue #5597, opened 2022-06-17
Support static and dynamic linking on Windows


compnerd:
### Description

Windows does not currently handle linking correctly.  When building a library to be linked statically, we must build with `-static`.  The static and shared libraries need a separate build (one for each variant used) including the object files and swiftmodule.

compnerd:
CC: @abertelrud

al45tair:
A complication here that is probably non-obvious to non-Windows developers is that when you make a Windows dynamic library (a `.dll`), you _also_ build a corresponding static library (a `.lib` file) — Windows developers and the Microsoft documentation refers to this as an "import library". The upshot is that it isn't just intermediates that need a separate directory here — you can't build the a static library and a dynamic library into the same directory, because the static library will overwrite the import library.

dschaefer2:
In the Windows development I've done, the libraries are usually dynamic located in the same directory as the executable. We've seen issues with too many exported symbols as we bundle a bunch of static libraries into a single product DLL.

I want to make sure we can build SwiftPM with SwiftPM on Windows so I'll play around with different options to see if it helps to build libraries as DLLs by default and not bags of object files like we do today.

al45tair:
Just noticed I hadn't mentioned above another gotcha that is relevant here, namely that the object files need to be built with different flags on Windows depending on whether you intend them to be part of a DLL or not. SwiftPM's current strategy of building object files and then creating static and/or dynamic library targets from them is therefore a bit broken on Windows, because if you ask for both static and dynamic library output, the code won't be built with the correct flags for both — we'd need to build the objects twice in that case.

That isn't the case on other platforms (for the most part… it is _possible_ on some platforms to build a static library without position independent code, and in some cases that means the objects in that library couldn't be used in a shared library, but PIC is typically just fine in a static library so that's an easy problem to solve).

compnerd:
The dual build being required was the stumbling block for me. We don't have a good way to model the dual builds because targets don't know how they will be used.

dschaefer2:
I definitely noticed that. I was stunned to see the --target arg only produced the objects and I had to use --product to produce the lib/exe. We'll probably need some customization of the build graph when the triple is Windows to fix this.

I'll play with this a bit more. Part of my swiftpm contributor training :).

dschaefer2:
@al45tair do you know off hand what those flags are?

compnerd:
[...]

PR #7873, opened 2024-08-12
Add more support for building on Windows using SwiftPM

jakepetroules:
This continues from the support added in #7866.

While we can't build on Windows as-is due to exceeding the 65k exported symbol limit, in turn due to limitations in the build system, we can apply -static to all of the library targets which are known to be linked statically, in order to get that symbol count down.

Note that this uses unsafeFlags, which is normally not recommended, however given that we are already using unsafeFlags elsewhere in the manifest, this seems reasonable enough for now until we can improve support for static linking on Windows.

Building the test targets still exceeds the symbol count though.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -72,8 +72,10 @@ let swiftPMProduct = (
 )
 
 #if os(Windows)
+let includeDynamicLibrary: Bool = false
 let systemSQLitePkgConfig: String? = nil
 #else
+let includeDynamicLibrary: Bool = true
 let systemSQLitePkgConfig: String? = "sqlite3"
 #endif
 
@@ -90,12 +92,15 @@ let package = Package(
     ],
     products:
     autoProducts.flatMap {
-        [
+        (includeDynamicLibrary ? [
             .library(
                 name: $0.name,
                 type: .dynamic,
                 targets: $0.targets
             ),
+        ] : [])
+        +
+        [
             .library(
                 name: "\($0.name)-auto",
                 targets: $0.targets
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci please smoke test

jakepetroules:
@swift-ci please test

compnerd (approved):

dschaefer2:
Can you create an issue for the Windows symbol count issue so we can track it and gather ideas? Thanks!

compnerd:
> Can you create an issue for the Windows symbol count issue so we can track it and gather ideas? Thanks!

I don't think that the issue is gathering ideas. In fact, I believe a radar exists for the issue. It is caused by SPM not knowing how too build libraries. A static and dynamic libraries are not products, but targets.  The output type changes the required compiler flags, the generated code, and cannot be used interchangeably. The current model of building everything for a shared library, using it as an object library is fundamentally incorrect and results in the over-exported symbol list. The idea here is to simply do what CMake does and model the build properly, tracking the linkage type across each dependency edge and building the dependency appropriately.

owenv:
https://github.com/swiftlang/swift-package-manager/issues/5597 contains some previous discussion

jakepetroules:
@swift-ci please test windows

dschaefer2:
> #5597 contains some previous discussion

Cool. I think that's definitely the root of the problem. In the Windows world, DLLs are usually the default. I want to see if that solves the symbol count issue.

compnerd:
> I want to see if that solves the symbol count issue.

We know that it does - see the CMake build.

owenv (approved):

dschaefer2:
> > I want to see if that solves the symbol count issue.
> 
> We know that it does - see the CMake build.

Ah, I see :)

> option(BUILD_SHARED_LIBS "Build shared libraries by default" YES)

compnerd:
Yes, we default to shared libraries and then selectively convert to static where it is profitable (the changes to make them static usually give stats on the size reduction).

PR #7876, opened 2024-08-13
Ensure package plugin sandbox contents is deterministic

owenv:
Previously, nondeterministic item replacement directory paths could be included in the sandbox content. This can lead to spurious incremental rebuilds of plugin outputs.

Because `itemReplacementDirectories` returns a unique temporary directory on each call, I don't believe including it in the sandbox profile had the intended effect, so removing it should not break existing plugins.


owenv:
@swift-ci please test

owenv:

francescomikulis (approved):

owenv:

owenv:
Reworked this a bit so we still include item replacement directories which should remain stable

owenv:
@swift-ci please test

owenv:

dschaefer2 (approved):

owenv:

owenv:

PR #7877, opened 2024-08-13
[Build] Free `BuildPlan` from `BuildTriple`

xedin:
### Motivation:

This is yet another step (if not the last one) on the path to remove
`buildTriple` from `Resolved{Product, Module}` identifiers and switch
everything that requires destination to use information from the build
plan.

### Modifications:

- Makes build descriptions identifiable just like `Resolved{Product, Module}.ID`
- Adds a way to traverse module  (both direct and recursive) dependencies starting from a module build description.
- Switches `BuildPlan.{target, product}Map` to `IdentifiableSet<{Product, Module}BuildDescription`.

### Result:

`BuildTriple` is no longer used for build related operations such as planning and forming llbuild manifests.

### DIFF
--- a/Sources/Basics/Collections/IdentifiableSet.swift
+++ b/Sources/Basics/Collections/IdentifiableSet.swift
@@ -42,6 +42,10 @@ public struct IdentifiableSet<Element: Identifiable>: Collection {
         Index(storageIndex: self.storage.elements.endIndex)
     }
 
+    public var values: some Sequence<Element> {
+        self.storage.values
+    }
+
     public subscript(position: Index) -> Element {
         self.storage.elements[position.storageIndex].value
     }
--- a/Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
+++ b/Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
@@ -191,10 +191,8 @@ extension LLBuildManifestBuilder {
     public func addTargetsToExplicitBuildManifest() throws {
         // Sort the product targets in topological order in order to collect and "bubble up"
         // their respective dependency graphs to the depending targets.
-        let nodes = self.plan.targets.compactMap {
-            ResolvedModule.Dependency.module($0.module, conditions: [])
-        }
-        let allPackageDependencies = try topologicalSort(nodes, successors: { $0.dependencies })
+        let allPackageDependencies = self.plan.targets.flatMap { $0.recursiveDependencies(using: self.plan) }
+
         // Instantiate the inter-module dependency oracle which will cache commonly-scanned
         // modules across targets' Driver instances.
         let dependencyOracle = InterModuleDependencyOracle()
@@ -206,14 +204,15 @@ extension LLBuildManifestBuilder {
 
         // Create commands for all module descriptions in the plan.
         for dependency in allPackageDependencies.reversed() {
-            guard case .module(let target, _) = dependency else {
+            guard case .module(let module, let description) = dependency else {
                 // Product dependency build jobs are added after the fact.
                 // Targets that depend on product dependencies will expand the corresponding
                 // product into its constituent targets.
                 continue
             }
-            guard target.underlying.type != .systemModule,
-                  target.underlying.type != .binary
+
+            guard module.underlying.type != .systemModule,
+                  module.underlying.type != .binary
             else {
                 // Much like non-Swift targets, system modules will consist of a modulemap
                 // somewhere in the filesystem, with the path to that module being either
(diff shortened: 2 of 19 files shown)


xedin:
@swift-ci please test

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
```diff
@@ -191,10 +191,8 @@ extension LLBuildManifestBuilder {
     public func addTargetsToExplicitBuildManifest() throws {
         // Sort the product targets in topological order in order to collect and "bubble up"
         // their respective dependency graphs to the depending targets.
-        let nodes = self.plan.targets.compactMap {
-            ResolvedModule.Dependency.module($0.module, conditions: [])
-        }
-        let allPackageDependencies = try topologicalSort(nodes, successors: { $0.dependencies })
+        let allPackageDependencies = self.plan.targets.flatMap { $0.recursiveDependencies(using: self.plan) }
```
xedin, line 194: @MaxDesiatov This is something I want to draw your attention to - previously `allPackageDependencies` didn't check conditions associated with dependencies but `recursiveDependencies(using:)` does that and I think that's reasonable for what this logic is trying to archive.

Sources/Basics/Collections/IdentifiableSet.swift
```diff
@@ -42,6 +42,10 @@ public struct IdentifiableSet<Element: Identifiable>: Collection {
         Index(storageIndex: self.storage.elements.endIndex)
     }
 
+    public var values: OrderedDictionary<Element.ID, Element>.Values {
```
MaxDesiatov, line 45: Would `some Collection<Element>` or `some Sequence<Element>` work better here? Then we don't have to commit to the exact collection type in a public API.

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
MaxDesiatov: Seems reasonable on the surface. Does this change some observable behavior that we should add a test for?

xedin: It doesn't that's why I wanted to bring it up.

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
```diff
@@ -422,63 +420,69 @@ extension LLBuildManifestBuilder {
...
-            if target.underlying is BinaryModule { return }
+            if module.underlying is BinaryModule { return }
             // Ignore Plugin Modules.
-            if target.underlying is PluginModule { return }
+            if module.underlying is PluginModule { return }
 
             // Depend on the binary for executable targets.
-            if target.type == .executable && prepareForIndexing == .off {
+            if module.type == .executable && prepareForIndexing == .off {
                 // FIXME: Optimize.
```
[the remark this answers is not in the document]
xedin: Sure, I can add some text to this

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
MaxDesiatov: Even if I show up in `git blame` for that line, I'm pretty sure I didn't write the original logic or change it at any recent point, it was probably some NFC refactor or reformatting. I don't mind this specific change, it makes sense, there's no significant readability regression, and it seems to be NFC

MaxDesiatov (approved):

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
xedin: Yeah, no worries!

xedin:
These changes helped me to discover https://github.com/swiftlang/swift-package-manager/pull/7879

Sources/Build/BuildManifest/LLBuildManifestBuilder+Swift.swift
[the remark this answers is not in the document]
xedin: ✅

Sources/Basics/Collections/IdentifiableSet.swift
xedin: Switched to `some Sequence<Element>`!

xedin:
@swift-ci please test

xedin:

xedin:

xedin:
@swift-ci please test

xedin:

PR #7879, opened 2024-08-13
[Build] BuildPlan: Always discover test modules through their aggrega…

xedin:
…te products

### Motivation:

Fixes a situation when tests were build for both host and target even though they should have been built only for the host (i.e. when one of the modules depends on a macro).

### Modifications:

- Adjusted `BuildPlan.computeDestinations` to never add `test` modules to successor list of a package and allow `test` products to always iterate their module dependencies instead.

### Result:

If one of the test targets depends on a macro we'd never attempt to build others for "target".

### DIFF
--- a/Sources/Build/BuildPlan/BuildPlan.swift
+++ b/Sources/Build/BuildPlan/BuildPlan.swift
@@ -986,9 +986,9 @@ extension BuildPlan {
             }
 
             for module in package.modules {
-                if case .test = module.underlying.type,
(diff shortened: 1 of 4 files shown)


xedin:
@swift-ci please test

xedin:
@swift-ci please test

xedin:

xedin:

MaxDesiatov (approved):
Great catch 👍

PR #7881, opened 2024-08-14
[Build] BuildPlan: Use `Destination` instead of `BuildTriple` to form…

xedin:
… per-module plugin output directory

### Motivation:

This is an NFC since there are no observable differences here from user perspective, just a different mechanism is used to get the same information.

### Modifications:

- `computeDestinations` provides a destination that should be used to invoke build tool plugins per module.

### Result:

Removes one of the last places where `BuildTriple` was still used.

### DIFF
--- a/Sources/Build/BuildPlan/BuildPlan.swift
+++ b/Sources/Build/BuildPlan/BuildPlan.swift
@@ -351,6 +351,7 @@ public class BuildPlan: SPMBuildCore.BuildPlan {
                 if let pluginConfiguration, !buildParameters.shouldSkipBuilding {
                     let pluginInvocationResults = try await Self.invokeBuildToolPlugins(
                         for: module,
+                        destination: destination,
                         configuration: pluginConfiguration,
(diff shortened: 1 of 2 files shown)


xedin:
@swift-ci please test

Sources/Build/BuildPlan/BuildPlan.swift
```diff
@@ -786,7 +788,7 @@ extension BuildPlan {
                 components: [
                     package.identity.description,
                     module.name,
-                    module.buildTriple.rawValue,
```
MaxDesiatov, line 789: Are there any remaining uses of `buildTriple` or can this property be removed now?

xedin: Yeah, we have a few more uses, mostly tests though. I'm planning to follow-up with SourceKitLSPAPI PR and then a PR to remove the rest.

MaxDesiatov (approved):

xedin:
@swift-ci please test

PR #7883, opened 2024-08-15
Un-Revert "[Build] Free `BuildPlan` from `BuildTriple`"

xedin:
Reverts swiftlang/swift-package-manager#7882

The CI problem associated with this revert actually stems from swift-foundation hard-coding a path that is no longer valid because some tests in `FoundationEssentialsTests` test macros which means that xctest bundle is going to be built for `host` and get `-tool` suffix for its resources directory. Using non-prefixed directory was incorrect because it's not related to the bundle that would actually be run.


MaxDesiatov (approved):

xedin:
@swift-ci please test

PR #7885, opened 2024-08-15
Disable part of a test that fails with Swift Testing in the toolchain.

grynspan:
Remove some assertions from `testSwiftTestXMLOutputWhenEmpty()` that aren't strictly necessary and which have started to fail as we add Swift Testing to the toolchain.


grynspan:
@swift-ci test

bnbarham (approved):

PR #7888, opened 2024-08-17
[Build] NFC: Remove uses of `buildTriple` from `computeDependencies(of:)`

xedin:
### Motivation:

Now that we have all the tools to compute destinations for dependencies of a product, 
we no longer need `buildTriple` here and dynamic/static library results could be represented
by build descriptions as a bonus.

### Modifications:

- `BuildPlan.computeDependencies(of:)` has been updated to take a `ProductBuildDescription` and use depth first scan that computes destination;
- Further, dynamic and static library dependencies are now presented by build descriptions which means that we no longer need to look stuff up while planning.

### Result:

One more use of `buildTriple` is gone.

### DIFF
--- a/Sources/Build/BuildPlan/BuildPlan+Product.swift
+++ b/Sources/Build/BuildPlan/BuildPlan+Product.swift
@@ -11,10 +11,11 @@
 //===----------------------------------------------------------------------===//
 
 import struct Basics.AbsolutePath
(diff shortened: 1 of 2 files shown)


xedin:
@swift-ci please test

MaxDesiatov (approved):

PR #7895, opened 2024-08-20
Fix `InternalImportsByDefault` diagnostics in _AsyncFileSystem

tshortli:
Address some `InternalImportsByDefault` diagnostics in the `_AsyncFileSystem` which have crept in because the Swift 6 toolchain isn't enforcing `.enableExperimentalFeature("InternalImportsByDefault")`.

### Motivation:

 Both the `StrictConcurrency` and `InternalImportsByDefault` features became "upcoming" features instead of "experimental" features in the Swift 6 compiler. This means that specifying those features with `.enableExperimentalFeature()` no longer works with newer toolchains, and therefore these settings in SwiftPM's `Package.swift` have become inactive. I'm making the compiler more lenient, so that `-enable-experimental-feature` will enable the corresponding upcoming feature if relevant (https://github.com/swiftlang/swift/pull/75962) but in the meantime it appears that some of the code in SwiftPM needs some fixes to continue to be accepted by the compiler with these features on.

### Modifications:

Adjusted access level on a number of import statements in files belonging to the `_AsyncFileSystem` target.

### Result:

`_AsyncFileSystem` builds successfully when `InternalImportsByDefault` is actually enabled with the Swift 6 toolchain.

### DIFF
--- a/Sources/_AsyncFileSystem/AsyncFileSystem.swift
+++ b/Sources/_AsyncFileSystem/AsyncFileSystem.swift
@@ -11,8 +11,8 @@
 //===----------------------------------------------------------------------===//
 
 import protocol _Concurrency.Actor
-@preconcurrency import struct SystemPackage.Errno
-@preconcurrency import struct SystemPackage.FilePath
+@preconcurrency package import struct SystemPackage.Errno
+@preconcurrency package import struct SystemPackage.FilePath
 
 /// An abstract file system protocol with first-class support for Swift Concurrency.
 package protocol AsyncFileSystem: Actor {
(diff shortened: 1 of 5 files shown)


tshortli:
@swift-ci please test

owenv (approved):

owenv:
There is a possibility the self-hosted jobs may not recognize these features as "upcoming", let's see...

bnbarham:
Yeah, this isn't going to work because of the self hosted jobs (which are still 5.9). We could wrap them in `#if` checks but that gets pretty unmanageable fairly quickly 😓. The code changes themselves LGTM though

tshortli:
Ok, sounds like I should just revert all of the `Package.swift` changes and hope that there won't be any more incompatible changes before https://github.com/swiftlang/swift/pull/75962 becomes effective in SwiftPM CI.

tshortli:
@swift-ci please test

tshortli:
@swift-ci please test Windows

PR #7896, opened 2024-08-20
[PackageGraph/Build] Remove `BuildTriple` and its remaining uses

xedin:
### Motivation:

Last uses of `BuildTriple` have been removed which means that there are no more recursive graph updates or duplicate targets.

### Modifications:

- Removes `BuildTriple` type
- Removes `buildTriple` properties from `Resolved{Product, Module}` types and all of the recursive update logic associated with them
- Updates `BuildOperation` to detect destination for products and modules if one wasn't provided by the caller. Only ambiguity here are tests but the rule is clear - direct macro dependencies cause all tests to be build for the "host".
- Updates tests to use product/module build description destination where appropriate.

### Result:

Hasta la vista, `BuildTriple`.


xedin:
@swift-ci please test

xedin:
Looks like the logic that select a destination for test products is not quiet right, I'll look into that tomorrow.

MaxDesiatov (approved):

xedin:
@swift-ci please test

xedin:

xedin:
@swift-ci please test macOS

PR #7897, opened 2024-08-20
WIP: New and user selectable progress bars

rauhul:
this is very much a work in progress. It works enough to be helpful locally for myself.

https://github.com/user-attachments/assets/710d1fcb-4b33-405b-89ff-aff2a992b7cb

### TODO:
- [ ] Handle % based tasks
- [x] Use milliseconds for tasks with 0 full seconds
- [x] refactor user text interleaving with progress animation (closure based api)
- [ ] Make "Write sources" tasks display name be more informational
- [ ] Indicate host vs target in task name

Larger example output, the new progress bar makes it really easy to see duplicate work:
Screenshot 2024-08-20 at 10 23 50 AM


MaxDesiatov:
I wonder if it would make sense for items that end up rounded as "0s" to drop down to milliseconds instead?

rauhul:
> I wonder if it would make sense for items that end up rounded as "0s" to drop down to milliseconds instead? Otherwise 0s doesn't seem believable on its own, and a lot of "0s" can accumulate to a substantial number, which also would be confusing.

Probably! I'd love feedback and suggestions :D

dschaefer2:
I wouldn't mind some of this now like the number of sources and compile durations. Is it easy to separate into a smaller PR?

dschaefer2:
Also can we show whether something is compiling as a tool (host) or destination (target)?

rauhul:
> Side note: should the "Write sources" tasks display name be more informational?

I'd consider that a "full note"!

rauhul:
> Also can we show whether something is compiling as a tool (host) or destination (target)?
> 
> (update: never mind, we build the plug-in tools before we do the production build, sigh).

It still useful because we build swift-syntax for the host (for macros) as part of the real build.

rauhul:
@dschaefer2 @MaxDesiatov `ms` in the time and a more descriptive task name for some `write-aux-file` tasks.

Screenshot 2024-09-10 at 7 04 22 PM

rauhul:
oops I meant to ping @neonichu not Doug 😓

bkhouri:
This looks nice.  Would it be possible to add some automated tests to validate the expected behaviour.  This way, we will ensure the feature does not regress.

dschaefer2:
> This looks nice. Would it be possible to add some automated tests to validate the expected behaviour? This way, we will ensure the feature does not regress.

Rauhul and I chatted about this today. We acknowledged the need but also the difficulty writing tests for it. Maybe we can get some help with that. Do we have tests for the current progress messages?

Sajjon:
@rauhul nice! Progress bar for building and [progress bar for Swift Testing](https://github.com/swiftlang/swift-testing/pull/19) could/should most likely share implementation - a generic implementation of the CLI based GUI for displaying and updating some progress bars - with the difference being the event stream of updates. 

Swift Testing does **not** have SPM as dependency in [Package.swift](https://github.com/swiftlang/swift-testing/blob/main/Package.swift#L36) though, hmm, perhaps a new Swift package for displaying progress in CLI could be created and added as a dependency to both Swift Testing and SPM?

plemarquand:
This is very cool! I am a big fan of making progress related output more informative and fun, since I think we all spend a good amount of our time looking at it.

That said, I think there are some use cases and nuances to keep in mind. @stmontgomery and @grynspan raised some interesting comments on @Sajjon's [swift-testing progress bar patch](https://github.com/swiftlang/swift-testing/pull/19) that I think apply here as well:
https://github.com/swiftlang/swift-testing/pull/19#issuecomment-1735768717
https://github.com/swiftlang/swift-testing/pull/19#issuecomment-1733909628

The relevant bullet points from their comments:
- We need to make sure that output is legible when written to something like Jenkins or a plain text file
- What do warnings and errors look like (stderr output)?
- Need to ensure this is tested and functional on macOS' Terminal.app, Ubuntu's Terminal, pure-CLI Linux, and Windows' modern Console.

grynspan:
With @plemarquand's notes in mind, I'd love to get this all packaged up somewhere that Swift Testing could link to it and use it too. :)

dschaefer2:
There are other use cases I know that could use good terminal progress animation in a Swift package such as swift-testing. I'd like to see turning this into that as a long term goal.

Also this is an optional view of progress. I think we still need the old one around for tools that depend on it. vscode-swift currently is one of those tools. But talking with @award999, we'd probably be better off with that interface being a streaming machine interface rather than trying to parse human consumable output. We need to be free to make changes to improve human consumption, like we are here.

hassila:
I'd definitely also would be interested in consider using it for the Benchmark package if it's broken out to a separate dependency in the future.

Sources/Basics/ProgressAnimation/Terminal/TerminalColoring.swift
```diff
@@ -0,0 +1,142 @@
...
+    ) -> TerminalColoring? {
+        // Explicitly disabled colors via CLICOLORS == 0
+        if env.cliColor == false { return .noColors }
+        // Explicitly disabled colors via NO_COLORS != nil
+        if env.noColor { return .noColors }
+        // Implicitly disabled colors because Xcode terminal cannot render them.
+        if env.runningInXcode { return .noColors }
+        // Disabled colors because output stream is not a tty, CI == nil,
+        // and CLICOLOR_FORCE == nil.
+        // FIXME: dont use stream.isTTY
```
Sajjon, line 55: Any resolution for this FIXME? Any suggestions for improvement?

Sources/Basics/ProgressAnimation/Misc/ProgressTask.swift
```diff
@@ -0,0 +1,47 @@
...
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+// FIXME: use task tree
+//struct ProgressTask {
```
Sajjon, line 14: dead code

Sajjon (changes requested):
Nice! Any recordings of the various animations? I'm curious about Ninja and Blast :)

Sources/Build/BuildOperation.swift
```diff
@@ -422,7 +428,9 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
 
         let configuration = self.config.configuration(for: .target)
         // delegate is only available after createBuildSystem is called
-        progressTracker.buildStart(configuration: configuration)
+        progressTracker.buildStart(
+            configuration: configuration,
+            action: "Building")
```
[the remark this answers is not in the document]
grynspan: I would want to use an `enum` conforming to `CustomStringConvertible` rather than extending a currency type like `String`.

dschaefer2:
Got my eyes on this now. I may restart it and leverage your Terminal code but need to think about how I want to model progress if it's any different than this.

PR #7903, opened 2024-08-21
[PackageModel] Toolchain: Split SwiftTesting flags between swift comp…

xedin:
…iler and linker

### Motivation:

Keeping them together broke Symbol Graph Extract tool because it cannot handle `-Xlinker` flags.

### Modifications:

- Adjust `deriveMacOSSpecificSwiftTestingFlags` to produce two arrays - one for swift compiler and one for linker flags.

### Result:

`extraFlags.swiftCompilerFlags` no longer includes `-Xlinker` flags.

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -410,23 +410,23 @@ public final class UserToolchain: Toolchain {
     static func deriveMacOSSpecificSwiftTestingFlags(
         derivedSwiftCompiler: AbsolutePath,
         fileSystem: any FileSystem
-    ) -> [String] {
+    ) -> (swiftCFlags: [String], linkerFlags: [String]) {
         // If this is CommandLineTools all we need to add is a frameworks path.
         if let frameworksPath = try? AbsolutePath(
             validating: "../../Library/Developer/Frameworks",
             relativeTo: resolveSymlinks(derivedSwiftCompiler).parentDirectory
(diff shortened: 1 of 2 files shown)


MaxDesiatov (commented):
I know little about this area of code, but how testable is this?

xedin:
That's what I'm trying to figure out right now :)

MaxDesiatov (approved):
Thanks 👍

xedin:
@swift-ci please test

xedin:

Sources/PackageModel/UserToolchain.swift
```diff
@@ -410,23 +410,23 @@ public final class UserToolchain: Toolchain {
...
             validating: "../../Library/Developer/Frameworks",
             relativeTo: resolveSymlinks(derivedSwiftCompiler).parentDirectory
         ), fileSystem.exists(frameworksPath.appending("Testing.framework")) {
-            return [
-                "-F", frameworksPath.pathString,
-                "-Xlinker", "-rpath",
-                "-Xlinker", frameworksPath.pathString
-            ]
+            return (swiftCFlags: [
+                "-F", frameworksPath.pathString
```
briancroom, line 420: Is this framework search path not also needed for the linker?

xedin: This appears to be sufficient, I tested with commandline tools only and -F + `-Xlinker -rpath` builds and runs the swift-testing tests without issues.

PR #7905, opened 2024-08-21
Fix more `InternalImportsByDefault` errors

tshortli:
SwiftPM doesn't build with all toolchains when the `InternalImportsByDefault` upcoming feature is properly enabled.

### Motivation:

With the upcoming feature `InternalImportsByDefault` enabled, some versions of the Swift compiler diagnose this pattern:

```
import protocol _Concurrency.Actor

// error: package protocol cannot refine an internal protocol
package protocol Proto: Actor {
  // ...
}
```

### Modifications:

Address these diagnostics by using correct access levels for direct imports of `_Concurrency`.

### Result:

SwiftPM builds with all relevant Swift compilers with `InternalImportsByDefault` enabled.

### DIFF
--- a/Sources/_AsyncFileSystem/AsyncFileSystem.swift
+++ b/Sources/_AsyncFileSystem/AsyncFileSystem.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import protocol _Concurrency.Actor
+package import _Concurrency
 @preconcurrency package import struct SystemPackage.Errno
 @preconcurrency package import struct SystemPackage.FilePath
 
--- a/Sources/_AsyncFileSystem/ConcurrencySupport.swift
+++ b/Sources/_AsyncFileSystem/ConcurrencySupport.swift
@@ -11,7 +11,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import _Concurrency
+internal import _Concurrency
 internal import class Dispatch.DispatchQueue
 
 extension DispatchQueue {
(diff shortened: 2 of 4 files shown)


tshortli:
This is a follow on to https://github.com/swiftlang/swift-package-manager/pull/7895.

tshortli:
@swift-ci please test

tshortli:
@swift-ci please test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
IIRC `_Concurrency` is not imported implicitly when building with CMake, that's why those imports were added in the first place

tshortli:
> IIRC `_Concurrency` is not imported implicitly when building with CMake, that's why those imports were added in the first place

I see, I wonder if that is due to `-disable-implicit-concurrency-module-import` getting passed down in CMake builds for some reason? It's hard for me to imagine a reason why that flag would be needed by anything in SwiftPM, maybe it's a mistake?

tshortli:
@swift-ci please test

tshortli:
Yeah, the flag seems to be intentional:

```
> rg disable-implicit-concurrency-module-import
Utilities/bootstrap
671:        swiftpm_args += ["-Xswiftc", "-Xfrontend", "-Xswiftc", "-disable-implicit-concurrency-module-import"]

Sources/swift-bootstrap/main.swift
221:            "-Xfrontend", "-disable-implicit-concurrency-module-import",
```

Maybe that dates from the days before `_Concurrency` was available ubiquitously? In any case, I won't question it for now. I've put the imports back and tried to thread the needle differently.

bnbarham:
> Yeah, the flag seems to be intentional:

Can probably remove. It is indeed from before _Concurrency was always available. I recently removed it from the manifest build themselves, I just missed that bootstrap was also doing it.

tshortli:
@swift-ci please test Windows

jakepetroules (approved):

PR #7906, opened 2024-08-22
Remove redundant EXECUTABLE_NAME overrides in PIF

jakepetroules:
This setting shouldn't normally be overridden (except one edge case I'm leaving in for now because I haven't investigated to know whether it's safe to remove the ".o" suffix from the PRODUCT_NAME of library targets, which can be a separate step.


jakepetroules:
@swift-ci please smoke test

jakepetroules:
@swift-ci test linux

jakepetroules:
@swift-ci test macos

owenv (approved):

PR #7916, opened 2024-08-24
Temporarily disable new Swift Testing tests

bnbarham:
`canImport(Testing)` currently fails on the full toolchain test jobs due to various issues when building for both x86 and arm64. Temporarily remove this check until we can fix the underlying problems there.

### DIFF
--- a/Tests/CommandsTests/TestCommandTests.swift
+++ b/Tests/CommandsTests/TestCommandTests.swift
@@ -276,7 +276,7 @@ final class TestCommandTests: CommandsTestCase {
     }
 
     func testBasicSwiftTestingIntegration() async throws {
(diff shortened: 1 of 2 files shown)


bnbarham:
@swift-ci please test

xedin (approved):
Maybe we should put a FIXME in every spot just to make sure that it doesn’t go unnoticed by mistake?

bnbarham:
> Maybe we should put a FIXME in every spot just to make sure that it doesn’t go unnoticed by mistake?

Should fix ASAP, I'll add on Monday if we end up not fixing.

xedin:
Absolutely, that's why I approved :)

PR #7954, opened 2024-09-10
Declare vars nonisolated to solve compiler waning

valeriyvan:


xedin (approved):

xedin:
@swift-ci please test

plemarquand:
@swift-ci please test macOS

MaxDesiatov:
Xcode/Swift installed on macOS CI nodes is too old to support this, so either it has to be conditional on the compiler version or wait (uncertain amount of time) until those nodes are updated.

xedin:
😢

dschaefer2:
I've been wondering if we need to take a step back with all these concurrency warnings and just make everything Swift 6. It certainly would be a ton of work but would be a good showcase for it.

dschaefer2:
I'm just looking for when we should start adopting actors and Sendable and make more things value types and all that. We had that in a previous project I worked on. Stepping into SwiftPM feels like going back in time :).

MaxDesiatov:
The main blocker is macOS CI, which is still on 5.9. Given that we need to make sure the toolchain can be bootstrapped from previous versions of Swift, we'll have to support 5.10 for quite some time.

valeriyvan:
> Xcode/Swift installed on macOS CI nodes is too old to support this, so either it has to be conditional on the compiler version or wait (uncertain amount of time) until those nodes are updated.

I would leave this PR waiting until CI gets upgraded or Cancellator.swift is written using new concurrency.

PR #7961, opened 2024-09-12
ManifestLoader.swift: Address concurrency warning

plemarquand:
ManifestLoaderProtocol.load was forwarding a checked continuation's resume method directily as a completion handler, resulting in the warning:

```
converting a value of type '(__shared sending Result<Manifest, any Error>) -> ()' to type '(Result<Manifest, any Error>) -> Void' risks causing data races; this is an error in the Swift 6 language mode
```

PR #7964, opened 2024-09-13
Address concurrency warnings

plemarquand:
Fix the rest of the concurrency warnings caused by directly passing a continuation's resume method.

Follow on to #7961

### DIFF
--- a/Sources/Basics/Concurrency/AsyncProcess.swift
+++ b/Sources/Basics/Concurrency/AsyncProcess.swift
@@ -786,7 +786,9 @@ package final class AsyncProcess {
     package func waitUntilExit() async throws -> AsyncProcessResult {
         try await withCheckedThrowingContinuation { continuation in
             DispatchQueue.processConcurrent.async {
-                self.waitUntilExit(continuation.resume(with:))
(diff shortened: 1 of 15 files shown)


MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

jakepetroules (approved):

plemarquand:
@MaxDesiatov is it worth back porting any changes to AsyncProcess to [TSCBasics.Process](https://github.com/swiftlang/swift-tools-support-core/blob/main/Sources/TSCBasic/Process/Process.swift)?

MaxDesiatov:
I don't think so, TSC is deprecated, and ideally we should deprecate `TSCBasics.Process`, but we'd have to migrate `swift-driver` first, otherwise they'd unexpectedly get a ton of deprecation warnings.

MaxDesiatov:
FWIW swift-driver would have to be asyncified in lockstep with SwiftPM anyway, because our Driver implementation in SwiftPM is not `async`, and that protocol in Driver is not `async` either.

PR #7983, opened 2024-09-19
AsyncProcess: Apply @Sendable annotation to address compiler warning

valeriyvan:


MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):

valeriyvan:
Does this also fail because of outdated CI?

xedin:
I think so, I believe we downgraded that to a warning at some point.

PR #7986, opened 2024-09-19
[Tests] NFC: Remove overload of `GeneralizedTime.init(_:)`

xedin:
### Motivation:

`swift-certificates` now implements initializer that accepts `Date` and doesn't throw which means this method is redundant but also incorrect because overloading only on `throws` is ambiguous in most cases due to throws checking not being part of the expression checking.

Uses of this overload type-checked only because it was always preferred before due to old type-checker hacks that we are trying to remove.

### Modifications:

- Removes `GeneralizedTime.init(_:) throws` that was added before `swift-certificates` introduced their own non-throwing version.

### Result:

No more duplicate code that is less safe than what `swift-certificates` vends.

### DIFF
--- a/Tests/PackageSigningTests/SigningTests.swift
+++ b/Tests/PackageSigningTests/SigningTests.swift
@@ -511,14 +511,14 @@ final class SigningTests: XCTestCase {
                     throw StringError("Missing OCSP request")
                 }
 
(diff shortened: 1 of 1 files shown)


xedin:
@swift-ci please test

MaxDesiatov (approved):

PR #7998, opened 2024-09-25
Add -F flag for xcframework dependency

### LINKED ISSUE #7580: PackagePlugin's PackageManager.getSymbolGraph does not support binaryTarget

- [X] Confirmed reproduction steps with SwiftPM CLI.

### Description

If a target has a dependency on binaryTarget. Then the call of `PackageManager.getSymbolGraph` to that target will get a directory contains no symbol graph.

It will break `PackageManager.getSymbolGraph`'s client such as swift-docc-plugin.

The downstream user can generate doc using `swift build` and call `docc convert` to generate documentation. But if they migrate to swift-docc-plugin or swift-docc-plugin based service like SwiftPackageIndex, the documentation build will fail.

### Expected behavior

Produce SymbolGraph normally.

### Actual behavior

Call swift-symbolgraph-extract with the arguments manually in terminal or adding the following patch to SymbolGraphExtract.swift

```diff
    public func extractSymbolGraph(
        ...
        try process.launch()
-       try process.waitUntilExit()
+       let result = try process.waitUntilExit()
+       if let error = String(bytes: try result.stderrOutput.get(), encoding: .utf8) {
+            print(error)
+.      }
    }
```

And you will get the following error message.

```
"<unknown>:0: error: missing required module \'MMKV\'\nError: Failed to load the module \'DemoKit\'. Are you missing build dependencies or include/framework directories?\nSee the previous error messages for details. Aborting.\n"
```

I have tried to add the framework manually via `-F`/`-I`/`-L ${MMKV_framework_path}` but none of them will work.

### Steps to reproduce

For user side reproduce steps:

1. Download the following DemoKit.zip package
2. Run `swift package generate-documentation --target DemoKit --verbose` and no documentation will be created.

For swiftpm developer reproduce steps:

1. Set up swiftpm repo development environment
2. Build PackageDescription target first and then choose `swift-package` target 
3. Set working directory to `DemoKit` and the corresponding arguments on launch in Xcode
4.(Optional) Apply the diff mentioned above.

[DemoKit.zip](https://github.com/apple/swift-package-manager/files/15362819/DemoKit.zip)

> Note: Here I use MMKV.xcframework as an example but any binary framework would have the same issue. The binary framework have nothing special here.

5.10

5.10 & macOS 14.4.1

Kyle-Ye:
Close #7580

### Motivation:

If a target has a dependency on binaryTarget. Then the call of `PackageManager.getSymbolGraph` to that target will get a directory contains no symbol graph.

It will break `PackageManager.getSymbolGraph`'s client such as swift-docc-plugin.

The downstream user can generate doc using `swift build` and call `docc convert` to generate documentation. But if they migrate to swift-docc-plugin or swift-docc-plugin based service like SwiftPackageIndex, the documentation build will fail.

```
// ✅ Build Package Documentation via Xcode GUI

// ✅ Build Package Documentation manually
swift build --target DemoKit \
  -Xswiftc -emit-symbol-graph \
  -Xswiftc -emit-symbol-graph-dir -Xswiftc "./symbol-graph"

docc convert --additional-symbol-graph-dir ./symbol-graph ...

// ❌ Build Package Documentation via swift-docc-plugin
swift package generate-documentation --target DemoKit
```

The underlying issue is that xcframework dependencies are being passed to swift-symbolgraph-extract with -I paths, rather than -F paths. (Thanks @QuietMisdreavus)

### Modifications:

Adding -F related additionalFlag for binary swift target.

> Replace -I with -F is enough to fix the issue. Keeping -I for compatibility reason in case there are other cases where -I is needed

### DIFF
--- a/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
@@ -172,6 +172,9 @@ public final class SwiftModuleBuildDescription {
     /// over to the build output directory so executables that use them can find them.
     var windowsDLLBinaryPaths: Set<AbsolutePath> = []
 
+    /// Paths to search when importing binary frameworks.
+    var frameworkSearchPaths: Set<AbsolutePath> = []
+
     /// Any addition flags to be added. These flags are expected to be computed during build planning.
     var additionalFlags: [String] = []
 
@@ -561,10 +564,7 @@ public final class SwiftModuleBuildDescription {
             args += ["-static"]
         }
 
-        // Only add the build path to the framework search path if there are binary frameworks to link against.
-        if !self.libraryBinaryPaths.isEmpty {
-            args += ["-F", BuildOperation.buildProductsPath(for: self.buildParameters).pathString]
-        }
+        args += self.frameworkSearchPathArguments()
 
         // Emit the ObjC compatibility header if enabled.
         if self.shouldEmitObjCCompatibilityHeader {
@@ -696,6 +696,8 @@ public final class SwiftModuleBuildDescription {
         // Include search paths for swift module dependencies.
         args += ["-I", self.modulesPath.pathString]
 
+        args += self.frameworkSearchPathArguments()
+
         // FIXME: Only include valid args
         // This condition should instead only include args which are known to be
         // compatible instead of filtering out specific unknown args.
@@ -704,6 +706,17 @@ public final class SwiftModuleBuildDescription {
         // will silently error failing the operation.
         args = args.filter { !$0.starts(with: "-use-ld=") }
         return args
+    }
+
+    private func frameworkSearchPathArguments() -> [String] {
+        var frameworkSearchPaths = self.frameworkSearchPaths
+        if !self.libraryBinaryPaths.isEmpty {
+            frameworkSearchPaths.insert(BuildOperation.buildProductsPath(for: self.buildParameters))
+        }
+
+        return frameworkSearchPaths.map(\.pathString).sorted().flatMap {
+            ["-F", $0, "-Xcc", "-F", "-Xcc", $0]
+        }
     }
 
     // FIXME: this function should operation on a strongly typed buildSetting
--- a/Sources/Build/BuildPlan/BuildPlan+Swift.swift
+++ b/Sources/Build/BuildPlan/BuildPlan+Swift.swift
@@ -92,6 +92,9 @@ extension BuildPlan {
                         library.headersPaths.forEach {
                             swiftTarget.additionalFlags += ["-I", $0.pathString, "-Xcc", "-I", "-Xcc", $0.pathString]
                         }
+                        if library.libraryPath.extension == "framework", swiftTarget.buildParameters.triple.isDarwin() {
+                            swiftTarget.frameworkSearchPaths.insert(library.libraryPath.parentDirectory)
+                        }
                         swiftTarget.libraryBinaryPaths.insert(library.libraryPath)
                     }
                 }
--- a/Tests/BuildTests/PluginsBuildPlanTests.swift
+++ b/Tests/BuildTests/PluginsBuildPlanTests.swift
@@ -180,8 +180,11 @@ struct PluginsBuildPlanTests {
             .Feature.Command.Package.Plugin,
         ),
         .requireHostOS(.macOS),
+        arguments: SupportedBuildSystemOnAllPlatforms,
     )
-    func docCPluginForBinaryDependency() async throws {
+    func docCPluginForBinaryDependency(
+        buildSystem: BuildSystemProvider.Kind,
+    ) async throws {
         try await fixture(name: "Miscellaneous/Plugins/SymbolGraphForBinaryDependency") { fixturePath in
             let result = try await AsyncProcess.popen(arguments: [
                 fixturePath.appending(RelativePath("FooKit/Scripts/archive_xcframework.sh")).pathString,
@@ -191,14 +194,12 @@ struct PluginsBuildPlanTests {
             try print(result.utf8stderrOutput())
             #expect(result.exitStatus == .terminated(code: 0))
             // Before we add -F support for xcframework, this call will throw since the command will abort with a non-zero exit code
-            do {
+            await #expect(throws: Never.self) {
                 _ = try await executeSwiftPackage(
                     fixturePath,
                     extraArgs: ["generate-symbol-graph"],
-                    buildSystem: .native
+                    buildSystem: buildSystem
                 )
-            } catch {
-                Issue.record("Expected symbol graph generation for an xcframework dependency to succeed, but failed with \(error)")
             }
         }
     }
--- a/Tests/BuildTests/PluginsBuildPlanTests.swift
+++ b/Tests/BuildTests/PluginsBuildPlanTests.swift
@@ -175,5 +175,31 @@ struct PluginsBuildPlanTests {
             expectFileExists(at: pluginToolBinPath.appending(pluginToolName))
         }
     }
-
+    @Test(
+        .tags(
+            .Feature.Command.Package.Plugin,
+        ),
+        .requireHostOS(.macOS),
+    )
+    func docCPluginForBinaryDependency() async throws {
+        try await fixture(name: "Miscellaneous/Plugins/SymbolGraphForBinaryDependency") { fixturePath in
+            let result = try await AsyncProcess.popen(arguments: [
+                fixturePath.appending(RelativePath("FooKit/Scripts/archive_xcframework.sh")).pathString,
+                "FooKit"
+            ])
+            try print(result.utf8Output())
+            try print(result.utf8stderrOutput())
+            #expect(result.exitStatus == .terminated(code: 0))
+            // Before we add -F support for xcframework, this call will throw since the command will abort with a non-zero exit code
+            do {
+                _ = try await executeSwiftPackage(
+                    fixturePath,
+                    extraArgs: ["generate-symbol-graph"],
+                    buildSystem: .native
+                )
+            } catch {
+                Issue.record("Expected symbol graph generation for an xcframework dependency to succeed, but failed with \(error)")
+            }
+        }
+    }
 }
--- a/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/FooKit/Package.swift
+++ b/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/FooKit/Package.swift
@@ -1,0 +1,13 @@
+// swift-tools-version: 5.8
+
+import PackageDescription
+
+let package = Package(
+    name: "FooKit",
+    products: [
+        .library(name: "FooKit", type: .dynamic, targets: ["FooKit"]),
+    ],
+    targets: [
+        .target(name: "FooKit"),
+    ]
+)
--- a/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/FooKit/Scripts/archive_xcframework.sh
+++ b/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/FooKit/Scripts/archive_xcframework.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the Swift open source project
+##
+## Copyright (c) 2024 Apple Inc. and the Swift project authors
+## Licensed under Apache License v2.0 with Runtime Library Exception
+##
+## See http://swift.org/LICENSE.txt for license information
+## See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+##
+##===----------------------------------------------------------------------===##
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd -P)"
+PROJECT_ROOT="$(dirname "$SCRIPT_DIR")"
+
+cd $PROJECT_ROOT
+
+PROJECT_BUILD_DIR="${PROJECT_BUILD_DIR:-"${PROJECT_ROOT}/.build"}"
+XCODEBUILD_BUILD_DIR="$PROJECT_BUILD_DIR/xcodebuild"
+XCODEBUILD_DERIVED_DATA_PATH="$XCODEBUILD_BUILD_DIR/DerivedData"
+
+PACKAGE_NAME=$1
+if [ -z "$PACKAGE_NAME" ]; then
+    echo "No package name provided. Using the first scheme found in the Package.swift."
+    PACKAGE_NAME=$(xcodebuild -list | awk 'schemes && NF>0 { print $1; exit } /Schemes:$/ { schemes = 1 }')
+    echo "Using: $PACKAGE_NAME"
+fi
+
+build_framework() {
+    local sdk="$1"
+    local destination="$2"
+    local scheme="$3"
+
+    local XCODEBUILD_ARCHIVE_PATH="$PROJECT_BUILD_DIR/$scheme-$sdk.xcarchive"
+
+    rm -rf "$XCODEBUILD_ARCHIVE_PATH"
+
+    PROTOBUFKIT_LIBRARY_TYPE=dynamic xcodebuild archive \
+        -scheme $scheme \
+        -archivePath $XCODEBUILD_ARCHIVE_PATH \
+        -derivedDataPath "$XCODEBUILD_DERIVED_DATA_PATH" \
+        -sdk "$sdk" \
+        -destination "$destination" \
+        BUILD_LIBRARY_FOR_DISTRIBUTION=YES \
+        INSTALL_PATH='Library/Frameworks' \
+        OTHER_SWIFT_FLAGS=-no-verify-emitted-module-interface
+
+    if [ "$sdk" = "macosx" ]; then
+        FRAMEWORK_MODULES_PATH="$XCODEBUILD_ARCHIVE_PATH/Products/Library/Frameworks/$scheme.framework/Versions/Current/Modules"
+        mkdir -p "$FRAMEWORK_MODULES_PATH"
+        cp -r \
+        "$XCODEBUILD_DERIVED_DATA_PATH/Build/Intermediates.noindex/ArchiveIntermediates/$scheme/BuildProductsPath/Release/$scheme.swiftmodule" \
+        "$FRAMEWORK_MODULES_PATH/$scheme.swiftmodule"
+        rm -rf "$XCODEBUILD_ARCHIVE_PATH/Products/Library/Frameworks/$scheme.framework/Modules"
+        ln -s Versions/Current/Modules "$XCODEBUILD_ARCHIVE_PATH/Products/Library/Frameworks/$scheme.framework/Modules"
+    else
+        FRAMEWORK_MODULES_PATH="$XCODEBUILD_ARCHIVE_PATH/Products/Library/Frameworks/$scheme.framework/Modules"
+        mkdir -p "$FRAMEWORK_MODULES_PATH"
+        cp -r \
+        "$XCODEBUILD_DERIVED_DATA_PATH/Build/Intermediates.noindex/ArchiveIntermediates/$scheme/BuildProductsPath/Release-$sdk/$scheme.swiftmodule" \
+        "$FRAMEWORK_MODULES_PATH/$scheme.swiftmodule"
+    fi
+
+    # Delete private and package swiftinterface
+    rm -f "$FRAMEWORK_MODULES_PATH/$scheme.swiftmodule/*.package.swiftinterface"
+    rm -f "$FRAMEWORK_MODULES_PATH/$scheme.swiftmodule/*.private.swiftinterface"
+}
+
+build_framework "macosx" "generic/platform=macOS" "$PACKAGE_NAME"
+
+echo "Builds completed successfully."
+
+cd $PROJECT_BUILD_DIR
+
+XCFRAMEWORK_DESTINATION="$PROJECT_ROOT/../$PACKAGE_NAME.xcframework"
+
+rm -rf $XCFRAMEWORK_DESTINATION
+xcodebuild -create-xcframework  \
+    -framework $PACKAGE_NAME-macosx.xcarchive/Products/Library/Frameworks/$PACKAGE_NAME.framework \
+    -output $XCFRAMEWORK_DESTINATION
--- a/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/FooKit/Scripts/archive_xcframework.sh
+++ b/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/FooKit/Scripts/archive_xcframework.sh
@@ -22,14 +22,17 @@ PROJECT_BUILD_DIR="${PROJECT_BUILD_DIR:-"${PROJECT_ROOT}/.build"}"
 XCODEBUILD_BUILD_DIR="$PROJECT_BUILD_DIR/xcodebuild"
 XCODEBUILD_DERIVED_DATA_PATH="$XCODEBUILD_BUILD_DIR/DerivedData"
 
-# SwiftPM CI runs tests with SWIFT_EXEC pointing at the just-built compiler. Do not let xcodebuild
-# inherit those overrides while resolving this fixture package; it should use the selected Xcode
-# toolchain to create the xcframework input for the SwiftPM command under test.
+# SwiftPM CI runs tests with compiler and library overrides pointing at the just-built toolchain and
+# in-tree SwiftPM libraries. Do not let xcodebuild inherit those while resolving this fixture package;
+# it should use the selected Xcode toolchain to create the xcframework input for the command under test.
 run_xcodebuild() {
     env \
+        -u DYLD_LIBRARY_PATH \
         -u SWIFT_EXEC \
         -u SWIFT_DRIVER_SWIFT_EXEC \
         -u SWIFT_EXEC_MANIFEST \
+        -u SWIFTCI_USE_LOCAL_DEPS \
+        -u SWIFTPM_CUSTOM_LIBS_DIR \
         xcodebuild "$@"
 }
 
--- a/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/FooKit/Scripts/archive_xcframework.sh
+++ b/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/FooKit/Scripts/archive_xcframework.sh
@@ -22,10 +22,21 @@ PROJECT_BUILD_DIR="${PROJECT_BUILD_DIR:-"${PROJECT_ROOT}/.build"}"
 XCODEBUILD_BUILD_DIR="$PROJECT_BUILD_DIR/xcodebuild"
 XCODEBUILD_DERIVED_DATA_PATH="$XCODEBUILD_BUILD_DIR/DerivedData"
 
+# SwiftPM CI runs tests with SWIFT_EXEC pointing at the just-built compiler. Do not let xcodebuild
+# inherit those overrides while resolving this fixture package; it should use the selected Xcode
+# toolchain to create the xcframework input for the SwiftPM command under test.
+run_xcodebuild() {
+    env \
+        -u SWIFT_EXEC \
+        -u SWIFT_DRIVER_SWIFT_EXEC \
+        -u SWIFT_EXEC_MANIFEST \
+        xcodebuild "$@"
+}
+
 PACKAGE_NAME=$1
 if [ -z "$PACKAGE_NAME" ]; then
     echo "No package name provided. Using the first scheme found in the Package.swift."
-    PACKAGE_NAME=$(xcodebuild -list | awk 'schemes && NF>0 { print $1; exit } /Schemes:$/ { schemes = 1 }')
+    PACKAGE_NAME=$(run_xcodebuild -list | awk 'schemes && NF>0 { print $1; exit } /Schemes:$/ { schemes = 1 }')
     echo "Using: $PACKAGE_NAME"
 fi
 
@@ -38,7 +49,7 @@ build_framework() {
 
     rm -rf "$XCODEBUILD_ARCHIVE_PATH"
 
-    PROTOBUFKIT_LIBRARY_TYPE=dynamic xcodebuild archive \
+    PROTOBUFKIT_LIBRARY_TYPE=dynamic run_xcodebuild archive \
         -scheme $scheme \
         -archivePath $XCODEBUILD_ARCHIVE_PATH \
         -derivedDataPath "$XCODEBUILD_DERIVED_DATA_PATH" \
@@ -78,6 +89,6 @@ cd $PROJECT_BUILD_DIR
 XCFRAMEWORK_DESTINATION="$PROJECT_ROOT/../$PACKAGE_NAME.xcframework"
 
 rm -rf $XCFRAMEWORK_DESTINATION
-xcodebuild -create-xcframework  \
+run_xcodebuild -create-xcframework  \
     -framework $PACKAGE_NAME-macosx.xcarchive/Products/Library/Frameworks/$PACKAGE_NAME.framework \
     -output $XCFRAMEWORK_DESTINATION
--- a/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/FooKit/Sources/FooKit/FooKit.swift
+++ b/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/FooKit/Sources/FooKit/FooKit.swift
@@ -1,0 +1,4 @@
+/// print "foo"
+public func foo() {
+    print("foo")
+}
--- a/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/Package.swift
+++ b/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/Package.swift
@@ -1,0 +1,25 @@
+// swift-tools-version: 5.8
+import PackageDescription
+
+let package = Package(
+    name: "DemoKit",
+    products: [
+        .library(name: "DemoKit", targets: ["DemoKit"]),
+    ],
+    targets: [
+        .plugin(
+            name: "GenerateSymbolGraphPlugin",
+            capability: .command(
+                intent: .custom(
+                    verb: "generate-symbol-graph",
+                    description: "Generate symbol graph for all Swift source targets."
+                )
+            )
+        ),
+        .binaryTarget(name: "FooKit", path: "FooKit.xcframework"),
+        .target(
+            name: "DemoKit",
+            dependencies: ["FooKit"]
+        ),
+    ]
+)
--- a/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/Plugins/GenerateSymbolGraphPlugin/GenerateSymbolGraphPlugin.swift
+++ b/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/Plugins/GenerateSymbolGraphPlugin/GenerateSymbolGraphPlugin.swift
@@ -1,0 +1,19 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+
+import Foundation
+import PackagePlugin
+
+@main
+struct GenerateSymbolGraphPlugin: CommandPlugin {
+    func performCommand(context: PluginContext, arguments: [String]) throws {
+        for target in context.package.targets where target is SwiftSourceModuleTarget {
+            let _ = try packageManager.getSymbolGraph(for: target, options: .init())
+        }
+    }
+}
--- a/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/Sources/DemoKit/DemoKit.swift
+++ b/Fixtures/Miscellaneous/Plugins/SymbolGraphForBinaryDependency/Sources/DemoKit/DemoKit.swift
@@ -1,0 +1,6 @@
+import FooKit
+
+public func greeting(_ name: String) {
+    FooKit.foo()
+    print("Hello \(name)")
+}


Kyle-Ye:
@swift-ci please test

MaxDesiatov (dismissed):
Please update issue description with more details, otherwise it's unclear what the effect of this PR is, what exactly it's trying to solve, and what's the desired outcome. Additionally, it needs test coverage to prevent future regressions.

Kyle-Ye:
> Please update issue description with more details, otherwise it's unclear what the effect of this PR is, what exactly it's trying to solve, and what's the desired outcome. 

The detail can be found on the issue link. If you do not mind duplication, I can re-summary the issue we encounter in this PR too.

> what exactly it's trying to solve, and what's the desired outcome.

Close the issue linked above.

> Additionally, it needs test coverage to prevent future regressions.

Yes. We need to add some test case to prevent future regressions before merging it.

I'm struggling with the test infra codebase here and wondering is is acceptable to add a small simple xcframework binary file into the codebase.

rauhul:
This does need a test, but I also have no idea where it would go. 

I am curious, do xcframeworks need -I flags at all and should they use -F exclusively?

MaxDesiatov:
Build planning has its own `BuildPlanTests: XCTestCase`, but I'd advise against adding new test functions to that same class, since it's already too big and should be split (even disregarding readability, it needs that purely for making it parallelizable). If no other test that cover build planning fit your change, creating new one is fine, taking the same testing approach that `BuildPlanTests` took.

QuietMisdreavus (approved):
Basic code change looks good. Having both sets of flags _should_ be fine. Were you able to test this on your demo project?

Kyle-Ye:
> Basic code change looks good. Having both sets of flags _should_ be fine. Were you able to test this on your demo project?

Yeah. I have tested it by setting working directory and arguments in Xcode Scheme Editor.

image

The things I'm working on:

1. Familiar with SwiftPM's testing infra to find a good place to write a test case to cover it.

```
// Current plan: PackagePluginAPITests/PackageManagerProxyTests.swift
import PackagePlugin
import XCTest
import _InternalTestSupport

final class PackageManagerProxyTests: XCTestCase {
    func testBinarySwiftTargetSymbolGraph() async throws {
        struct SymbolGraphPlugin: CommandPlugin {
            func performCommand(context: PackagePlugin.PluginContext, arguments: [String]) async throws {
                for target in context.package.targets {
                    let result = try packageManager.getSymbolGraph(for: target, options: .init())
                    let url = result.directoryPath
                    ...
                }
            }
        }
        
        try fixture(name: "BinaryTarget") { path in
            // Construct a package which has a binary dependency and perform the plugin
            let package = ...
            let context: PluginContext = ...
            let plugin = SymbolGraphPlugin()
            plugin.performCommand(context: context, arguments: [])
            ...
        }
    }
}
```

2. Add a simple/small xcframework in `Fixture/BinaryTarget` folder. I'm also plan to add the source package and script to generate the xcframework. But it is easy to make one for Xcode project. Have not found a way for pure Swift package yet.

I guess I'll end up using a pure small xcframework build in my local without package source or a remote one with checksum in the end. (Any recommendation here? eg. Apple's official xcframework example will be great. But I can't find one)

Fixtures/Miscellaneous/Plugins/DocCForBinaryDependency/Package.swift
```diff
@@ -0,0 +1,25 @@
...
+    products: [
+        .library(name: "DemoKit", targets: ["DemoKit"]),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/swiftlang/swift-docc-plugin", from: "1.0.0"),
+    ],
+    targets: [
+        .binaryTarget(
+            name: "MMKV",
+            url: "https://github.com/Kyle-Ye/MMKV/releases/download/v1.3.5-binary/MMKV.xcframework.zip",
```
Kyle-Ye, line 17: Do we have any official/stable xcframework url from Apple? Or is it acceptable to directly place a plain xcframework binary in the repo?

Use `Kyle-Ye/MMKV` only for placeholder. Will replace it before merge.

Kyle-Ye:
@swift-ci please test

Fixtures/Miscellaneous/Plugins/DocCForBinaryDependency/Package.swift
```diff
@@ -0,0 +1,25 @@
...
+
+import PackageDescription
+
+let package = Package(
+    name: "DemoKit",
+    products: [
+        .library(name: "DemoKit", targets: ["DemoKit"]),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/swiftlang/swift-docc-plugin", from: "1.0.0"),
```
Kyle-Ye, line 12: We have avoid `swift-docc-plugin` dependency here with writing the same symbolGraph call from a plugin target in the DemoKit package.

But to keep things simple, I think use `swift-docc-plugin` here is fine.

Fixtures/Miscellaneous/Plugins/DocCForBinaryDependency/Package.swift
Kyle-Ye: Update with a plain FooKit.xcframework build locally.

Kyle-Ye:
@swift-ci test

Kyle-Ye:
@swift-ci test

Fixtures/Miscellaneous/Plugins/DocCForBinaryDependency/Package.swift
MaxDesiatov: There are certain CI jobs that can't access external dependencies, so this dependency needs to be removed.

Kyle-Ye: Fixed.

Kyle-Ye:
@swift-ci test

Kyle-Ye:
@swift-ci test Windows

Fixtures/Miscellaneous/Plugins/DocCForBinaryDependency/FooKit.xcframework/macos-arm64_x86_64/FooKit.framework/Versions/A/Modules/FooKit.swiftmodule/Project/arm64-apple-macos.swiftsourceinfo
MaxDesiatov, line 1: Can you elaborate on what these binaries contain? Would it be possible to build them within the test? I'm not feeling comfortable committing black box binaries that will be linked with and code from them potentially executed on CI.

Kyle-Ye: I understand your concern here. Execute a black box binary indeed bring some security hole to the CI system.

> Add a simple/small xcframework in Fixture/BinaryTarget folder. I'm also plan to add the source package and script to generate the xcframework. It is easy to make one for Xcode project. But I have not found a way for pure Swift package yet.
I guess I'll end up using a pure small xcframework build in my local without package source or a remote one with checksum in the end. (Any recommendation here? eg. Apple's official xcframework example will be great. But I can't find one)

As my previous comments said, I have not found a way for make a pure Swift package to build the required xcframework. The easiest way is via a xcodeproj.

But I don't want to add the xcodeproj file to the SwiftPM repository at the beginning. So I just only add the binary into the repo.

Here is the xcodeproj I used to build the xcframework.

[FooKit.zip](https://github.com/user-attachments/files/17187467/FooKit.zip)

Build it with Release configuration, remove the code signature and create it into xcframework.

```shell
xcodebuild build -configuration Release

cd build/Release

rm -rf FooKit.framework/Versions/A/_CodeSignature

xcodebuild \
-create-xcframework \
-framework FooKit.framework \
-output FooKit.xcframework
```

What do you think I should do here?
1. Find a Swift package way to generate xcframework
2. Add the small Xcode project into the repo and make the binary on the fly.
3. Keep it as it is.

MaxDesiatov: > 1. Find a Swift package way to generate xcframework

If you mean adding a dependency on some other package to generate xcframeworks, that would be hard to justify for a bugfix like this one IMO. Adding new dependencies is a major change and needs additional CI setup.

> 2. Add the small Xcode project into the repo and make the binary on the fly.

Is there a reason this would not be a preferred option? Does that add significant amount of run time to our CI jobs? I see no downsides with it, as long as it's only built on Darwin, and IIUC your change is Darwin only anyway?

> 3. Keep it as it is.

Even with provided instructions, I don't see how that would guarantee that provided binaries were created with these instructions in the first place. The only way to do that is to generate binaries on the fly with option 2.

dschaefer2: We can not accept binaries into the source repo for the reasons @MaxDesiatov mentions. It's too much of a security hole.

Kyle-Ye: 1. No. I means use a pure swift package to generate xcframework instead of a xcodeproj. (eg. `swift build` to generate a dynamic framework. But there is other issue blocking me doing it)

2. There are multi reasons why I do not prefer it.

2-1: It makes SwiftPM's test target has a strong dependency on the Xcode. Even on the Darwin world, there is someone who would use other tools to build swift & swiftpm instead of Xcode. Although such build config is not officially supported by the upstream.

2-2: The xcodeproj file itself is also a binary file in some extend. And thus it can also be considered as unsafe as bad person may embed malicious script in it. We need to use Xcode GUI or some other tool to inspect its content here. However, compared to the Mach-O binary, most of xcodeproj file content is encoded in UTF-8, so we can actually examine it directly with simple text tools.

2-3: If we want to make the binary on the fly, what would be the appropriate time for triggering it? It will add more complexity here and the only possible solution IMO is add a prebuild plugin in the swiftpm package itself.

--------

Let’s return to the initial question—security concerns.

In this specific case, the swiftsourceinfo file and swiftdoc are not essential, and we can simply remove them. The only remaining binary file is FooKit.

Considering its size of only 134K, we can easily use a disassembler to inspect its contents to confirm it does not contain other malicious behavior. The cost of this approach is almost the same to use text tools to check whether the .xcodeproj file contains malicious code.

If we need to cover general cases here (eg. A big binary which is a black box we can't afford to inspect its content), I suggest relying on trusted CI services (like Swift CI or GitHub) to build the artifacts. And we then just use it.

Here’s my suggestion:

- We can use the artifacts built by the GitHub CI. (You can easily verify it by comparing the MD5 checksum of the commit's xcframework with the one produced by the CI. Examples: https://github.com/Kyle-Ye/swift-package-manager-fookit)
  - Because we trust GitHub Action, we change the audit from binary to source code and the build script.
  - Build log: https://github.com/Kyle-Ye/swift-package-manager-fookit/actions/runs/11108169992/job/30860488380
  - Release: https://github.com/Kyle-Ye/swift-package-manager-fookit/releases/tag/0.0.1
- Alternatively, we can host the repository under the swiftlang organization and reference the URL and checksum in `Fixtures/Miscellaneous/Plugins/DocCForBinaryDependency/Package.swift` file instead of using a local .xcframework file.
   - I believe this would be another issue we can discuss if there would be more binary xcframework in the future.

--------

In summary: 

To address the reviewer's security concerns, I've moved this build process to a GitHub Action. The reviewer can simply compare the xcframework's MD5 checksum to confirm that it hasn't been tampered with. (Note that the value calculated on the GitHub Release page is the MD5 checksum of the zip file.)

MaxDesiatov: >  The xcodeproj file itself is also a binary file in some extend

I don't think that's a true, `.pbxproj` files don't have binary encoding and changes in them can be clearly seen in PR diffs.

> And thus it can also be considered as unsafe as bad person may embed malicious script in it.

Again, an introduction of such script would be obvious in the PR diff. 

> xcodeproj file content is encoded in UTF-8, so we can actually examine it directly with simple text tools.

Exactly, that's the main reason to stick to it.

> It makes SwiftPM's test target has a strong dependency on the Xcode.

That's totally fine, the use of `xcframework`s already introduces a dependency on Xcode, so there's no change there.

> we can easily use a disassembler to inspect its contents

"easily" is a very strong assertion here. A disassembly is not shown in a PR diff, and a disassembly of a 134K binary is not easy to read, inspect, and make an audit on by any means either. A `.pbxproj` file is trivial to inspect directly in RP diffs by comparison, with no additional tools needed.

> The reviewer can simply compare the xcframework's MD5 checksum to confirm that it hasn't been tampered with.

Again, one would have to make an audit on the original binary in the first place. And if that has been tampered with, there's no way for us to know, since a binary produced outside of Swift CI could've been tampered with on a self-hosted node.

My other concern is the ability to make easy to miss subsequent changes to such binaries that aren't obvious in PR diffs. That's not the case with actual readable source files, like `.pbxproj`.

MaxDesiatov: > We can use the artifacts built by the GitHub CI

We can't use GitHub Actions for this repository (including transitive use), so this option can't be considered.

Kyle-Ye: > > We can use the artifacts built by the GitHub CI
> 
> We can't use GitHub Actions for this repository (including transitive use), so this option is not considered.

Not mean to use GitHub Actions for this repository. I mean use artifacts(xcframework) build by GitHub Actions.

MaxDesiatov: > this binary will most likely not have any further updates after being introduced in this PR

By PR authors maybe, but in the future some potential malicious actor can sneak in a change to this binary in some future PR, someone runs such binary on Swift CI without examining the diff first or missing the change in the diff (GitHub is not great at highlighting binary changes), and then we get malicious payload executed on CI.

MaxDesiatov: >  I just do not know what would you suggest about be time we build the binary if we choose to build it on the fly? Seems unnecessary and out of scope for me.

Why not build the binary in `testDocCPluginForBinaryDependency` directly? I wouldn't call preparation of test data in a test function unnecessary or out of scope.

Kyle-Ye: Got it. Thanks for the detailed explanation.

Kyle-Ye: I still feel adding project.pbxproj to this repo is like a crime.

But if that's the result you prefer, I'll end up using this way you recommend.

See 120216098e7679a5594e9eaad1b0c80a18b56ba4

MaxDesiatov: Thanks! I don't see any problem with adding that. In fact, we already have an `.pbxproj` on `main` here https://github.com/swiftlang/swift-package-manager/blob/main/IntegrationTests/Fixtures/BinaryTargets/Inputs/SwiftFramework/SwiftFramework.xcodeproj/project.pbxproj

Fixtures/Miscellaneous/Plugins/DocCForBinaryDependency/FooKit/FooKit.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,369 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 77;
```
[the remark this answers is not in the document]
MaxDesiatov: It has to support a much older version of Xcode, at least whatever version shipped with Swift 5.9, maybe even 5.8 to be safe.

Fixtures/Miscellaneous/Plugins/DocCForBinaryDependency/FooKit.xcframework/macos-arm64_x86_64/FooKit.framework/Versions/A/Modules/FooKit.swiftmodule/Project/arm64-apple-macos.swiftsourceinfo
Kyle-Ye: Should we use the existing SwiftFramework instead creating another one?

Kyle-Ye: https://github.com/swiftlang/swift-package-manager/blob/d70b78bb2332855163c25ef769d2310aceaf4c17/IntegrationTests/Tests/IntegrationTests/Helpers.swift#L326-L337

Kyle-Ye: Looks like IntegrationTests is not accessible with swiftpm package itself. I'll keep using my original implementation instead of reusing the existing one on IntegrationTests.

Kyle-Ye:
@swift-ci test

Kyle-Ye:
@swift-ci test

Kyle-Ye:
The "Swift Test macOS Platform (self hosted) " job seems using Xcode-15.1b2 on macOS 13.6.

And it will give the following error for the xcodeproj.

```
2024-09-30 17:30:56.748 xcodebuild[76291:929429] didn't find classname for 'isa' key
2024-09-30 17:30:56.752 xcodebuild[76291:929429] Writing error result bundle to /var/folders/bb/hcrjxg1s0b96pfst0ymhmp240000gn/T/ResultBundle_2024-30-09_17-30-0056.xcresult
xcodebuild: error: Unable to read project 'FooKit.xcodeproj'.
	Reason: The project ‘FooKit’ is damaged and cannot be opened. Examine the project file for invalid edits or unresolved source control conflicts.

Path: /var/folders/bb/hcrjxg1s0b96pfst0ymhmp240000gn/T/Miscellaneous_Plugins_SymbolGraphForBinaryDependency.PcrL7M/Miscellaneous_Plugins_SymbolGraphForBinaryDependency/FooKit/FooKit.xcodeproj
Exception: didn't find classname for 'isa' key
```

I have already set the project format to Xcode 12.0 on Xcode GUI. Any suggestion here? (Create a new project with a lower Xcode version can solve it. But my Mac just upgrades to macOS 15 and it only support running Xcode 16 😢)

Also can't install Xcode-15.1b2 on my Mac because macOS 15 does not support it.

MaxDesiatov:
> I have already set the project format to Xcode 12.0 on Xcode GUI. Any suggestion here? (Create a new project with a lower Xcode version can solve it. But my Mac just upgrades to macOS 15 and it only support running Xcode 16 😢)
> 
> Also can't install Xcode-15.1b2 on my Mac because macOS 15 does not support it.

Have you tried either modifying [the existing project file](https://github.com/swiftlang/swift-package-manager/blob/main/IntegrationTests/Fixtures/BinaryTargets/Inputs/SwiftFramework/SwiftFramework.xcodeproj/project.pbxproj) or using one of the community-maintained Xcode project generation tools?

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
@swift-ci test

Kyle-Ye:
@swift-ci test

Kyle-Ye:
I've update the project file by modify the existing one.

It will still [fail](https://ci.swift.org/job/swift-package-manager-PR-macos-smoke-test/4736/consoleText) on Xcode-15.1b2 CI unfortunately.

If bumping the CI Xcode version to a release is not possible in the short term. I'll try to prepare a macOS VM to reproduce and fix the issue here.

```
    /Applications/Xcode-beta.app/Contents/Developer/usr/bin/xcodebuild archive -project FooKit.xcodeproj -scheme FooKit -configuration Release -destination generic/platform=macOS -archivePath archives/macOS SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES

User defaults from command line:
    IDEArchivePathOverride = /private/var/folders/bb/hcrjxg1s0b96pfst0ymhmp240000gn/T/Miscellaneous_Plugins_SymbolGraphForBinaryDependency.8fIxmZ/Miscellaneous_Plugins_SymbolGraphForBinaryDependency/FooKit/archives/macOS
    IDEPackageSupportToolchainOverrideForManifestLoading = com.apple.dt.toolchain.XcodeDefault
    IDEPackageSupportUseBuiltinSCM = YES

Build settings from command line:
    BUILD_LIBRARY_FOR_DISTRIBUTION = YES
    SKIP_INSTALL = NO
    TOOLCHAINS = default

Prepare packages

ComputeTargetDependencyGraph
note: Building targets in dependency order
note: Target dependency graph (1 target)
    Target 'FooKit' in project 'FooKit' (no dependencies)

GatherProvisioningInputs

CreateBuildDescription
Command CreateBuildDescription failed with a nonzero exit code

error: unexpected service error: The Xcode build system has crashed. Build again to continue.

** ARCHIVE FAILED **
```

Kyle-Ye:
@swift-ci test

Kyle-Ye:
I can't reproduce the crash on CI using the same `Xcode-15.1.0-Beta.2.app` locally.
```
sudo xcode-select -s /Applications/Xcode-15.1.0-Beta.2.app/Contents/Developer
/Applications/Xcode-15.1.0-Beta.2.app/Contents/Developer/usr/bin/xcodebuild archive -project FooKit.xcodeproj -scheme FooKit -configuration Release -destination generic/platform=macOS -archivePath archives/macOS SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES
...
** ARCHIVE SUCCEEDED **
```

Kyle-Ye:
We are encountering a weird Xcode crash issue for `Swift Test macOS Platform (smoke test)` and can't reproduce it locally.

Is there any plan to bump the current Xcode 15.1b2 on CI to any release version of Xcode? cc @shahmishal

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
@swift-ci test

Kyle-Ye:
@swift-ci test

Kyle-Ye:
@swift-ci test

Kyle-Ye:
> Have you tried either modifying [the existing project file](https://github.com/swiftlang/swift-package-manager/blob/main/IntegrationTests/Fixtures/BinaryTargets/Inputs/SwiftFramework/SwiftFramework.xcodeproj/project.pbxproj) or using one of the community-maintained Xcode project generation tools?

Tried to create it via Xcode 15.0.1 / Xcode 15.1 / Xcode 15.1 Beta 2 and modifying the existing project file. They all end up with a strange crash which I can't reproduce locally. (Even with the same Xcode version - 15.1b2)

```swift
error: unexpected service error: The Xcode build system has crashed. Build again to continue.
```

There is already one xcframework zip in the repo. But I can't use it since it only contains a x86_64 one - `IntegrationTests/Fixtures/BinaryTargets/TestBinary/SwiftFramework.zip`.

I suggest we either bump the current Xcode 15.1b2 on CI to any release version of Xcode / just use a xcframework zip (like the existing SwiftFramework.zip) / keep it as it is but mark it as skip like the previous binary test case. @MaxDesiatov 

Or I'd be appreciated if you have other suggestions which may help push the PR to the next level.

https://github.com/swiftlang/swift-package-manager/blob/58f9aa4484763108211a18bd669b8693fe5a170e/IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift#L15-L23.

Kyle-Ye:
@swift-ci test

Kyle-Ye:
@swift-ci test

Kyle-Ye:
@swift-ci test

Kyle-Ye:
Remove xcodeproj and use pure Swift package to generate xcframework still fails on `macOS+smoke test CI` for an unknown error.

```
Command line invocation:
    /Applications/Xcode-beta.app/Contents/Developer/usr/bin/xcodebuild archive -scheme FooKit -archivePath /private/var/folders/bb/hcrjxg1s0b96pfst0ymhmp240000gn/T/Miscellaneous_Plugins_SymbolGraphForBinaryDependency.vYoYE9/Miscellaneous_Plugins_SymbolGraphForBinaryDependency/FooKit/.build/FooKit-macosx.xcarchive -derivedDataPath /private/var/folders/bb/hcrjxg1s0b96pfst0ymhmp240000gn/T/Miscellaneous_Plugins_SymbolGraphForBinaryDependency.vYoYE9/Miscellaneous_Plugins_SymbolGraphForBinaryDependency/FooKit/.build/xcodebuild/DerivedData -sdk macosx -destination generic/platform=macOS BUILD_LIBRARY_FOR_DISTRIBUTION=YES INSTALL_PATH=Library/Frameworks OTHER_SWIFT_FLAGS=-no-verify-emitted-module-interface

User defaults from command line:
    IDEArchivePathOverride = /private/var/folders/bb/hcrjxg1s0b96pfst0ymhmp240000gn/T/Miscellaneous_Plugins_SymbolGraphForBinaryDependency.vYoYE9/Miscellaneous_Plugins_SymbolGraphForBinaryDependency/FooKit/.build/FooKit-macosx.xcarchive
    IDEDerivedDataPathOverride = /private/var/folders/bb/hcrjxg1s0b96pfst0ymhmp240000gn/T/Miscellaneous_Plugins_SymbolGraphForBinaryDependency.vYoYE9/Miscellaneous_Plugins_SymbolGraphForBinaryDependency/FooKit/.build/xcodebuild/DerivedData
    IDEPackageSupportToolchainOverrideForManifestLoading = com.apple.dt.toolchain.XcodeDefault
    IDEPackageSupportUseBuiltinSCM = YES

Build settings from command line:
    BUILD_LIBRARY_FOR_DISTRIBUTION = YES
    INSTALL_PATH = Library/Frameworks
    OTHER_SWIFT_FLAGS = -no-verify-emitted-module-interface
    SDKROOT = macosx14.2
    TOOLCHAINS = default

Resolve Package Graph

<unknown>:0: warning: legacy driver is now deprecated; consider avoiding specifying '-disallow-use-new-driver'
<unknown>:0: error: unable to execute command: <unknown>
```

rauhul:
> Maybe -I is needed for the header search I guess

I think I'm missing how this has been determined, is there a message above about how the build fails without the flag?

My goal here is so minimize the CLI flags because it makes debugging SwiftPM builds _much_ easier.

Kyle-Ye:
> > Maybe -I is needed for the header search I guess
> 
> I think I'm missing how this has been determined, is there a message above about how the build fails without the flag?
> 
> My goal here is so minimize the CLI flags because it makes debugging SwiftPM builds _much_ easier.

Not very familiar with such flags. So I can't confidently remove it.

But I tried to remove `-I` and kick off the test case with a C target. It still success.

image

xedin:
I have a couple of thoughts here:

1. @Kyle-Ye I think you misunderstand what @MaxDesiatov was trying to say initially. I don't think we need an xcframework test here, it would be enough to either modify an existing test in a BuildPlanTests to make sure that we pass the flags to binary targets correct (for example there might be a test that does a similar check but doesn't check compiler arguments related to binary target dependencies).

2. To @rauhul's question in https://github.com/swiftlang/swift-package-manager/pull/7998#issuecomment-2374262198. I don't think we need to use `-I` here at all because correctly emitted `-F` should handle includes and libraries correctly.

 To that end, I think we should remove the code that adds `-I` and `-Xcc -I` from `BuildPlan+Swift.swift` and replace it with 

  ```
  library.headersPaths.forEach {
    swiftTarget.libraryBinaryPaths.insert($0)
  }
  ```

  and modify https://github.com/swiftlang/swift-package-manager/blob/main/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift#L522-L525 to emit `-F` and `-Xcc -F` to each library in the list.

3. We need to add `-F` for each libary in `libraryBinaryPaths` to `SwiftModuleBuildDescription.symbolGraphExtractArguments`.

Kyle-Ye:
> 1. @Kyle-Ye I think you misunderstand what @MaxDesiatov was trying to say initially. I don't think we need an xcframework test here, it would be enough to either modify an existing test in a BuildPlanTests to make sure that we pass the flags to binary targets correct (for example there might be a test that does a similar check but doesn't check compiler arguments related to binary target dependencies) or add a new test (in a separate file i.e. BuildPlanBinaryTargetTests.swift) which build a plan with a binary target dependency and checks output of `compileArguments` and `symbolGraphExtractArguments` for correctness.

Got it. And huge thanks for the point out.

Kyle-Ye:
> 2. To @rauhul's question in [Add -F flag for xcframework dependency #7998 (comment)](https://github.com/swiftlang/swift-package-manager/pull/7998#issuecomment-2374262198). I don't think we need to use `-I` here at all because correctly emitted `-F` should handle includes and libraries correctly.
> 
> To that end, I think we should remove the code that adds `-I` and `-Xcc -I` from `BuildPlan+Swift.swift` and replace it with
> 
> ```
> library.headersPaths.forEach {
>   swiftTarget.libraryBinaryPaths.insert($0)
> }
> ```
> 
> and modify https://github.com/swiftlang/swift-package-manager/blob/main/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift#L522-L525 to emit `-F` and `-Xcc -F` to each library in the list.
> 
> 3. We need to add `-F` for each libary in `libraryBinaryPaths` to `SwiftModuleBuildDescription.symbolGraphExtractArguments`.

d57ee08278177a11dcfd0d8bd6bcdff74b426c4a is what I got. Correct me if I mistake something.

```diff
                    for library in libraries {
                        library.headersPaths.forEach {
-                            swiftTarget.additionalFlags += ["-I", $0.pathString, "-Xcc", "-I", "-Xcc", $0.pathString]
+                            swiftTarget.libraryBinaryPaths.insert($0)
                        }
                        swiftTarget.libraryBinaryPaths.insert(library.libraryPath)
                    }
```

And it will be the following using the DemoKit.zip provided by the [issue](https://github.com/swiftlang/swift-package-manager/issues/7580).

```
(lldb) po library.libraryPath
▿ <AbsolutePath:"/<redacted>/DemoKit/.build/artifacts/demokit/MMKV/MMKV.xcframework/macos-arm64_x86_64/MMKV.framework">
  ▿ underlying : <AbsolutePath:"/<redacted>/DemoKit/.build/artifacts/demokit/MMKV/MMKV.xcframework/macos-arm64_x86_64/MMKV.framework">
    ▿ _impl : UNIXPath
      - string : "/<redacted>/DemoKit/.build/artifacts/demokit/MMKV/MMKV.xcframework/macos-arm64_x86_64/MMKV.framework"
(lldb) po library.headersPaths
▿ 1 element
  ▿ 0 : <AbsolutePath:"/<redacted>/DemoKit/.build/artifacts/demokit/MMKV/MMKV.xcframework/macos-arm64_x86_64">
    ▿ underlying : <AbsolutePath:"/<redacted>/DemoKit/.build/artifacts/demokit/MMKV/MMKV.xcframework/macos-arm64_x86_64">
      ▿ _impl : UNIXPath
        - string : "/<redacted>/DemoKit/.build/artifacts/demokit/MMKV/MMKV.xcframework/macos-arm64_x86_64"
```

xedin:
The line that adds `libraryPath` needs to be removed.

Sources/Build/BuildPlan/BuildPlan+Swift.swift
```diff
@@ -43,7 +43,7 @@ extension BuildPlan {
                     let libraries = try self.parseXCFramework(for: target, triple: swiftTarget.buildParameters.triple)
                     for library in libraries {
                         library.headersPaths.forEach {
-                            swiftTarget.additionalFlags += ["-I", $0.pathString, "-Xcc", "-I", "-Xcc", $0.pathString]
+                            swiftTarget.libraryBinaryPaths.insert($0)
```
Kyle-Ye, line 46: https://github.com/swiftlang/swift-package-manager/pull/7998#issuecomment-2443480388

Do you mean Line 46 or Line 48 here? @xedin 

Considering the `po` result, I guess you mean Line 46. But it would conflict with your previous comment here.

> To that end, I think we should remove the code that adds -I and -Xcc -I from BuildPlan+Swift.swift and replace it with
> 
> library.headersPaths.forEach {
>  swiftTarget.libraryBinaryPaths.insert($0)
> }

xedin: I mean `swiftTarget.libraryBinaryPaths.insert(library.libraryPath)` on line 48, `libraryPath` doesn't actually point to the place that `-F` would recognize anyway.

Tests/BuildTests/PluginsBuildPlanTests.swift
```diff
@@ -79,4 +79,23 @@ final class PluginsBuildPlanTests: XCTestCase {
...
+        try await fixture(name: "Miscellaneous/Plugins/SymbolGraphForBinaryDependency") { fixturePath in
+            let result = try await AsyncProcess.popen(arguments: [
+                fixturePath.appending(RelativePath("FooKit/Scripts/archive_xcframework.sh")).pathString,
+                "FooKit"
+            ])
+            try print(result.utf8Output())
+            try print(result.utf8stderrOutput())
+            XCTAssertEqual(result.exitStatus, .terminated(code: 0))
+            // Before we add -F support for xcframework, this call will throw since the command will abort with a non-zero exit code
+            let _ = try await executeSwiftPackage(fixturePath, extraArgs: ["generate-symbol-graph"])
```
bkhouri, line 98: question: can we add an assertion that validates the expected behaviour of this step? Maybe `XCTAssertNoThrow` or check the return code?

Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
```diff
@@ -521,7 +521,10 @@ public final class SwiftModuleBuildDescription {
 
         // Only add the build path to the framework search path if there are binary frameworks to link against.
         if !self.libraryBinaryPaths.isEmpty {
-            args += ["-F", self.buildParameters.buildPath.pathString]
+            args += [
```
bkhouri, line 524: question: Does this `if` block and the one [here](https://github.com/swiftlang/swift-package-manager/pull/7998/files?w=1#diff-2affeac76f1ef6916e6d06095d4e7f679e592c07c2a7556ad4eaddddd441ed91R664) always need to match? If so, can we refactor by moving the common code into its own function so that updating in one place will automatically update all usage?

Tests/BuildTests/PluginsBuildPlanTests.swift
Kyle-Ye: I'd like to to. But there is a question at present about if we should keep xcframework or not.

I originally thought it is required by the reviewer suggestion. And this is some blocking issues. See comments here https://github.com/swiftlang/swift-package-manager/pull/7998#issuecomment-2399356924

And the latest suggestion is we do not need them. And I'll probably delete this in next commit. See https://github.com/swiftlang/swift-package-manager/pull/7998#issuecomment-2443198351

bkhouri (changes requested):
Hi.

I provided comments using [Conventional Comments](https://conventionalcomments.org).

I'm happy to have a discussion if my comments are invalid.

Sources/Build/BuildPlan/BuildPlan+Swift.swift
Kyle-Ye: This is weird. I do not know the what "-F" is recognizing. But according to the test case, deleting Line 46 is fine while deleting Line 48 would make the test case fail with "No such module FooKit".

Saying that, I do think it is valuable we add real xcframework test case (via a trusted local one or fix the CI to build it on the fly). The simple flags check may be  doubtful here.

There is other reviewer have a +1 for the real test case.

https://github.com/swiftlang/swift-package-manager/pull/7998/files?w=1#r1822483995

Tests/BuildTests/PluginsBuildPlanTests.swift
Kyle-Ye: > issue (blocking): Add assertion to express test intent.

There is a exitStatus check to verify it. - `XCTAssertEqual(result.exitStatus, .terminated(code: 0))`

If the build is failing, the status code would abort with a non-zero exit code.

Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
Kyle-Ye: I think we should. I'll refactor it to a small helper function here.

Kyle-Ye:
@swift-ci please test

Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
Kyle-Ye: Fixed

Kyle-Ye (commented):
BTW, I verified that the original issue no longer seems to reproduce with Xcode 27 beta / Swift 6.4, possibly because SwiftPM now defaults to swiftbuild there. That said, since older build paths/toolchains will likely remain in use for a while, I still think this fix is worthwhile.

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
@swift-ci please test windows

Kyle-Ye:
Sorry for the delay. The CI is passing now and I have resolved your last review suggestions. Would you mind giving another review? @bkhouri

bkhouri:
@Kyle-Ye : Sorry for the delay in response.   The changes in the PR touch the native build system, which is marked ad deprecated on branch `release/6.4.x` and branch `main`.  The Swift Build build system is the default on those branches.

is the issue fixed with the Swift Build build system?  If not, could you re-work the PR so the fix is applied to the Swift Build build system?

https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/6.4

Kyle-Ye:
@bkhouri See this [comment](https://github.com/swiftlang/swift-package-manager/pull/7998#pullrequestreview-4478709533).

I understand SwiftPM moving toward swiftbuild, but deprecating the native build system doesn’t instantly make the ecosystem ready. Some packages (including a few of mine) still need it as a fallback, and fixes there can still matter during the transition IMO.

Tests/BuildTests/PluginsBuildPlanTests.swift
```diff
@@ -175,5 +175,31 @@ struct PluginsBuildPlanTests {
...
+            ])
+            try print(result.utf8Output())
+            try print(result.utf8stderrOutput())
+            #expect(result.exitStatus == .terminated(code: 0))
+            // Before we add -F support for xcframework, this call will throw since the command will abort with a non-zero exit code
+            do {
+                _ = try await executeSwiftPackage(
+                    fixturePath,
+                    extraArgs: ["generate-symbol-graph"],
+                    buildSystem: .native
```
bkhouri: We should add this test to both the native and Swift Build build system

If we apply the above change to add parameterized test, this becomes:
```suggestion
                    buildSystem: buildSystem,
```

Tests/BuildTests/PluginsBuildPlanTests.swift
```diff
@@ -175,5 +175,31 @@ struct PluginsBuildPlanTests {
...
+                )
+            } catch {
+                Issue.record("Expected symbol graph generation for an xcframework dependency to succeed, but failed with \(error)")
+            }
+        }
```
bkhouri: Consider changing the `do... catch` block to use an `#expect` API call
```suggestion
            #expect(throws: Never.self) {
                _ = try await executeSwiftPackage(
                    fixturePath,
                    extraArgs: ["generate-symbol-graph"],
                    buildSystem: .native
                )
            }
        }
```

Tests/BuildTests/PluginsBuildPlanTests.swift
```diff
@@ -175,5 +175,31 @@ struct PluginsBuildPlanTests {
...
         }
     }
-
+    @Test(
+        .tags(
+            .Feature.Command.Package.Plugin,
+        ),
+        .requireHostOS(.macOS),
+    )
+    func docCPluginForBinaryDependency() async throws {
```
bkhouri: Suggestion: we should execute this tests on both native and Swift Build build system.

```suggestion
    @Test(
        .tags(
            .Feature.Command.Package.Plugin,
        ),
        .requireHostOS(.macOS),
        arguments: SupportedBuildSystemOnAllPlatforms,
    )
    func docCPluginForBinaryDependency(
        buildSystem: BuildSystemProvider.Kind,
    ) async throws {
```

bkhouri:
> Some packages (including a few of mine) still need it as a fallback, and fixes there can still matter during the transition IMO.

@Kyle-Ye: Are there GitHub issues for the issues?

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
@bkhouri I’ve applied and pushed the suggestions above.

> > Some packages (including a few of mine) still need it as a fallback, and fixes there can still matter during the transition IMO.
> 
> @Kyle-Ye: Are there GitHub issues for the issues?

The issues I saw in some of my packages were with much earlier versions of Swift Build, and I haven’t retested them with the recent Swift 6.3 or 6.4 toolchains yet.

Once Swift 6.4 is released, I plan to update those packages with --build-system first and then try to adpot and fix the potential issue with swiftbuild. Will file issues then if something does not work.

Kyle-Ye:
@swift-ci please smoke test

bkhouri:
> The issues I saw in some of my packages were with much earlier versions of Swift Build, and I haven’t retested them with the recent Swift 6.3 or 6.4 toolchains yet.

@Kyle-Ye : Could you please try with a recent 6.4 developer snapshot as significant work had been made to ensure feature parity.  There are known issues and differences.  You can read more about them on the SwiftPM documentation [here](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/6.4)

Kyle-Ye:
I'll give it a try this weekend.

BTW, I believe the PR and the potential compatibility issue is 2 unrelated thing right?

bkhouri:
> I'll give it a try this weekend.
> 
> BTW, I believe the PR and the potential compatibility issue is 2 unrelated thing right?

I'm more interested in knowing about any issues building your packages with a recent 6.4 dev snapshot to ensure we have not introduced regressions or we have been documented behavioural changes and known issues.

Kyle-Ye:
> I'm more interested in knowing about any issues building your packages with a recent 6.4 dev snapshot to ensure we have not introduced regressions or we have been documented behavioural changes and known issues.

Confirmed my package is now working fine with both Xcode 27 Beta and the OSS 6.4.x-snapshot-2026-07-11 toolchain.

bkhouri:
> > I'm more interested in knowing about any issues building your packages with a recent 6.4 dev snapshot to ensure we have not introduced regressions or we have been documented behavioural changes and known issues.
> 
> Confirmed my package is now working fine with both Xcode 27 Beta and the OSS 6.4.x-snapshot-2026-07-11 toolchain.

@Kyle-Ye : Since the issue is fixed with Xcode 27 Beta and the OSS 6.4.x-snapshot-2026-07-11, I'm going to close the PR and the related issue.

bkhouri:
The PR and issue author, @Kyle-Ye, indicated the issue is fixed in Xcode 27 Beta and in the OSS 6.4.x developer snapshot from 2026-07-11.  As a result, I'm closing this PR.

---
reference: https://github.com/swiftlang/swift-package-manager/pull/7998#issuecomment-4983901670

PR #8004, opened 2024-09-27
WIP: Introduce SwiftSetting API for controlling whether testability is enabled for a target

jakepetroules:
This introduces a new `enableTesting` API in SwiftSetting which allows targets to explicitly control whether testability is enabled, as some projects do not want to use the @testable feature. This can also improve debug build performance as significantly fewer symbols will be exported from the binary in the case where @testable is disabled.

The default is equivalent to `enableTesting(true, .when(configuration: .debug))`, so there is no behavior change from today without explicitly adopting this new API.

The --enable-testable-imports/--disable-testable-imports command line flags now acts as overrides -- if specified, they will override any build settings configured at the target level, and if unspecified, the target-level settings will be respected.

### DIFF
--- a/Sources/PackageLoading/PackageBuilder.swift
+++ b/Sources/PackageLoading/PackageBuilder.swift
@@ -1192,6 +1192,10 @@ public final class PackageBuilder {
                 }
 
                 values = [version.rawValue]
+
+            case .enableTesting(let enable):
+                decl = .ENABLE_TESTABILITY
+                values = enable ? ["YES"] : ["NO"]
             }
 
             // Create an assignment for this setting.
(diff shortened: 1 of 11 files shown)


jakepetroules:
Untested draft. Will write up a proposal soon.

dschaefer2:
Thanks! Yeah, really want to get some wide community feedback on this. Pretty much everyone fights with enable testing at some point or other.

jakepetroules:
pitch: https://forums.swift.org/t/pitch-swiftpm-testable-build-setting/75084
proposal: https://github.com/swiftlang/swift-evolution/pull/2580

Sources/PackageLoading/PackageBuilder.swift
```diff
@@ -1016,6 +1016,12 @@ public final class PackageBuilder {
             }
         }
 
+        for setting in manifestTarget.settings {
+            if case let .enableTestableImport(enable) = setting.kind, enable, setting.condition?.config == "release" {
```
jakepetroules, line 1020: I might need to check for config==nil as well

LoganDark:
Ow; just ran into this abandoned PR while trying to test my package in release mode. Guess I will have to try to figure out some manual hacks.

PR #8005, opened 2024-09-28
Add a new show-executables package subcommand for all executables

cmcgee1024:
The `swift run` subcommand runs executables that come from a variety of places. Typically, users
are expecting to be able to execute products and targets for their package, iterating quickly with
their code changes. You can also run executable products from packages that are part of the
dependency list. It turns out that all of the executable products of the root package's transitive
closure are available.
    
Discovery of the available executables is limited to either direct inspection of the Package.swift,
the `swift package describe`, and `swift package completion-tool list-executables`. However, these
only work to find the executables of the current package, not in the dependent packages. This has
real implications for a package maintainer since duplicate executable names are forbidden among the
closure of the package's dependencies.
    
Provide the ability to show the complete list of available executables from the package with a new
`swift package show-executables` subcommand. Qualify executables that aren't in the current package
with their package of origin for traceability purposes. The list can be formatted as JSON so that
this information can be used by third-party tools.


Sources/Commands/Utilities/DescribedPackage.swift
```diff
@@ -28,12 +28,13 @@ struct DescribedPackage: Encodable {
...
     let platforms: [DescribedPlatformRestriction]
     let products: [DescribedProduct]
+    let executables: [DescribedExecutable]
     let targets: [DescribedTarget]
     let cLanguageStandard: String?
     let cxxLanguageStandard: String?
     let swiftLanguagesVersions: [String]?
 
-    init(from package: Package) {
+    init(from package: Package, executables: [(package: String?, name: String)]) {
```
MaxDesiatov, line 37: Why not use the newly introduced nominal type instead of a structurally-typed tuple?

```suggestion
    init(from package: Package, executables: [DescribedExecutable]) {
```

Sources/Commands/PackageCommands/Describe.swift
```diff
@@ -41,12 +41,31 @@ extension SwiftPackageCommand {
                 observabilityScope: swiftCommandState.observabilityScope
             )
 
-            try self.describe(package, in: type)
+            let buildSystem = try await swiftCommandState.createBuildSystem(
```
MaxDesiatov, line 44: WDYT about making it conditional on `// swift-tools-version: 6.1`? I would expect some CLI relies on the old behavior. This change seems additive, but I'd rather see it only applied for 6.1 package to make this at least somewhat opt-in.

cmcgee1024: It feels a bit over cautious to me. Perhaps we have some very tight couplings out there that we need to consider?

For the text format, I would be surprised if another tool directly consumes that. I feel that it's mostly to help end-users understand what's contained in their package.

A tool that consumes the output would most likely use the JSON format. Most JSON decoders that I know are capable of ignoring additional fields and objects.

dschaefer2: I'm wondering if this should be a separate subcommand. Partly to avoid surprises, but also I'd like to see this used by bash completions and friends and I don't think we want those things parsing the output of describe. For that matter, I'm not sure we want users to be doing that either :).

cmcgee1024: What's the concern around completion processors using describe to find targets, products, etc. for completing various swift commands? I thought that this might be one of the intended use cases since describe has the JSON output format option. Maybe we're aiming towards a new direction for tools to query package metadata?

dschaefer2: Oh, wait, we have a ```completion-tool``` subcommand! Can you check whether this supports what you're trying to accomplish?

Also, looking at describe right now, It's help does say ```OVERVIEW: Describe the current package```. By listing all executables you are changing what describe does, no? It's is only listing things in the root package. That's why listing more than that is really a different subcommand. And it will cause package resolution to happen which will take a long time. ```describe``` does not do that.

cmcgee1024: Thanks @dschaefer2 I'll check those out.

My rationale for adding the transitive executables to describe was that the package enables you to run all of those executables in the same way you can run the ones in the package. They form part of the package's interface description. Projects intend for their devs to run some of those tools, such as formatters. Plus, people already know about the describe subcommand, which can help with discoverability.

Ah, so it's package resolution that's the concern here. Is it because it can be slow, or something else? Is it much slower than package analysis, which I believe requires compiling and running the Package.swift. Other subcommands like `swift run` do this, but then I suppose that there is an expectation that analysis, resolution, build will happen before the executable is run.

dschaefer2: describe compiles and runs the Package.swift. That's fine. In fact, that's what it's doing, showing the results of that. The resolve does all the checkout and deep dependency analysis. That, of course, takes a while.

It's also the sheer volume of stuff coming out in describe when all I want to see what executables I can run. I just ran describe on swiftpm itself, a pretty standard sized package. Pages and pages of stuff. It's a debug tool, not really intended for human consumption. I don't think I'd want to be using it as a general query mechanism.

cmcgee1024: Thanks all, I've reworked this patch so that it instead adds a `swift package show-executables` subcommand modelled after the existing `swift package show-dependencies` subcommand, which is expected to not only analyze the current package, but also resolve packages transitively to find the entire list of products. I think that this should help to decrease the risk of disruption to existing clients of the describe function.

Sources/Commands/PackageCommands/ShowExecutables.swift
```diff
@@ -0,0 +1,87 @@
...
+
+        switch self.format {
+        case .flatlist:
+            for executable in executables {
+                if let package = executable.package {
+                    print("\(executable.name) (\(package))")
+                } else {
+                    print(executable.name)
+                }
+            }
```
MaxDesiatov, line 51: formatting nit (makes it easier to navigate around between cases with Vim `{`/`}` shortcuts

```suggestion
            }

```

MaxDesiatov (approved):
Thanks!

MaxDesiatov:
@swift-ci test

Sources/Commands/PackageCommands/ShowExecutables.swift
```diff
@@ -0,0 +1,88 @@
...
+
+struct ShowExecutables: AsyncSwiftCommand {
+    static let configuration = CommandConfiguration(
+        abstract: "List the available executables from this package.")
+
+    @OptionGroup(visibility: .hidden)
+    var globalOptions: GlobalOptions
+
+    @Option(help: "Set the output format.")
+    var format: ShowExecutablesMode = .flatlist
```
plemarquand, line 27: This should be made consistent with `ShowDependenciesMode` in [`SwiftPackageCommand`](https://github.com/swiftlang/swift-package-manager/blob/6ad5efb4445c213820897a8784461bc9a7f1314b/Sources/Commands/PackageCommands/ShowDependencies.swift#L72). The help string should be in a similar format, and if possible should have the same options (I believe it supports text, dot (??), json and flatlist)

plemarquand:
@swift-ci test this please

plemarquand:
@swift-ci please test this

plemarquand:
@swift-ci test

cmcgee1024:
The failure in the macOS smoke test is sourceckit-lsp and appears to be unrelated to this change. @MaxDesiatov do you know how to kick off the windows platform test with the latest changes?

dschaefer2:
@swift-ci please test windows

MaxDesiatov:
@swift-ci test macos

dschaefer2 (approved):

PR #8009, opened 2024-09-29
Make OutputClosure @Sendable to address compiler warning

valeriyvan:


dschaefer2:
@swift-ci please test

Sources/SourceControl/GitRepository.swift
```diff
@@ -105,20 +105,37 @@ public struct GitRepositoryProvider: RepositoryProvider, Cancellable {
         progress: FetchProgress.Handler? = nil
     ) throws -> String {
         if let progress {
-            var stdoutBytes: [UInt8] = [], stderrBytes: [UInt8] = []
+            nonisolated(unsafe) var stdoutBytes: [UInt8] = []
+            nonisolated(unsafe) var stderrBytes: [UInt8] = []
```
MaxDesiatov, line 109: I'm concerned we're trading concurrency warnings for `nonisolated(unsafe)` access on mutable arrays. Have you tried using new `AsyncProcess` APIs based on `AsyncSequence` instead? We'd have to re-enable corresponding tests on CI first: https://github.com/swiftlang/swift-package-manager/blob/d70b78bb2332855163c25ef769d2310aceaf4c17/Tests/BasicsTests/AsyncProcessTests.swift#L398

MaxDesiatov (changes requested):
I don't think this change is ready in the current form. Using `nonisolated(unsafe)` is inherently unsafe, we're using legacy `Dispatch` APIs here, and not making the actual transition to structured concurrency with this change. If this API ends up being called from a task on Swift concurrency thread pool, it then blocks that thread for long time, which is also highly undesirable.

It has no improvement over the status quo IMO, but removes the warning and can easily hide data races. We should use `AsyncSequence` APIs of `AsyncProcess` instead and re-enable tests for those first, resolving any issues with that that may come up in the process.

Sources/SourceControl/GitRepository.swift
valeriyvan: why this test is disabled?

MaxDesiatov: There seem to be non-deterministic failures in some CI jobs for it, so those have to be fixed first.

valeriyvan:
I don't think I understand how to accomplish this.

PR #8014, opened 2024-10-01
Reenable testUsage() as testInvalidUsage()

plemarquand:
I believe its expected for the argument `-halp` to show the help string in swift-argument-parser, as if there is a `-h` or `--help` anywhere in the arguments all other inputs are disregarded and the help string is shown.

Instead use an invalid argument to drive the test and reenable it.

### DIFF
--- a/Tests/CommandsTests/PackageCommandTests.swift
+++ b/Tests/CommandsTests/PackageCommandTests.swift
@@ -48,10 +48,9 @@ final class PackageCommandTests: CommandsTestCase {
         XCTAssertMatch(stdout, .contains("USAGE: swift package"))
     }
 
-    func testUsage() async throws {
-        throw XCTSkip("rdar://131126477")
+    func testInvalidUsage() async throws {
         do {
-            _ = try await execute(["-halp"])
+            _ = try await execute(["--xxx"])
             XCTFail("expecting `execute` to fail")
         } catch SwiftPMError.executionFailure(_, _, let stderr) {
             XCTAssertMatch(stderr, .contains("Usage: swift package"))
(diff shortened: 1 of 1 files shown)


plemarquand:
@swift-ci test

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -48,10 +48,9 @@ final class PackageCommandTests: CommandsTestCase {
         XCTAssertMatch(stdout, .contains("USAGE: swift package"))
     }
 
-    func testUsage() async throws {
-        throw XCTSkip("rdar://131126477")
```
MaxDesiatov, line 52: Do we have this radar resolved then? I find it confusing that any option that starts with `-h` leads to this unexpected behavior. We should at least raise this on the SAP repo for them to resolve this first.

plemarquand: I added a comment on the internal radar saying basically I think this is works as intended, but maybe @natecook1000 can chime in.

MaxDesiatov: cc @rauhul

MaxDesiatov: I'm surprised that's the default, given that both `swiftc` and `clang` stick to single dash options, and thus `clang -fblocks` is clearly not equivalent to `clang -f -b -l -o -c -k -s`.

rauhul: swiftc and clang also use options with `-long-name` so I wouldn't take them as amazing prior art. arg parser follows the pattern of `ls -asl` and `git clean -Xdf` where short names are combined.

plemarquand: My experience with this type of combined syntax is as a POSIX convention. See Guideline #5 in the [POSIX Utility Conventions](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html).

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test

Tests/CommandsTests/PackageCommandTests.swift
plemarquand: @MaxDesiatov @rauhul Any lingering concerns with this PR/behaviour?

rauhul: I'm happy with it

MaxDesiatov: Even if `swift build -halp` is parsed the same as `swift build -h -a -l -p` and `-h` is a valid flag, I'd still expect an error message to be printed about invalid `-a -l -p` flags.

bkhouri: I agree with @MaxDesiatov .

I also think swift-argument-parser test need to be augmented to cover this case, if they currently don't!

bkhouri: After looking at the current swift package manager behaviour, I think I'm find with it.  

Running `swift build -halp` and `swift build -alhp` both yield the help usage.  That's ok.

Running `swift build -alp` yields an error. 

```
❯ swift build -alp
error: Unknown option '-alp'. Did you mean '--vv'?
Usage: swift build <options>
  See 'build -help' for more information.
```


The  usage help should probably indicate `See 'build --help for more information` or `See 'build -h' for more information`. it currently has a single `-` before `help`.

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -48,10 +48,9 @@ final class PackageCommandTests: CommandsTestCase {
...
+    func testInvalidUsage() async throws {
         do {
-            _ = try await execute(["-halp"])
+            _ = try await execute(["--xxx"])
```
bkhouri, line 53: issue: Can we add a test, if they don't currently exist, that specified the behaviour when executing `swift build -halp`, `swift build -alhp` and `swift build --help`

bkhouri (changes requested):
Requesting to ensure tests exists the the behaviour we want to preserve.  Looks good otherwise.

Tests/CommandsTests/PackageCommandTests.swift
plemarquand: I think these tests should probably go in swift-argument-parser, since it handles the help arguments

bkhouri: Similar tests should definitely be included in swift-argument-parser. However, from swift-package-manager perspective, it's an implementation details.

This is a `large` (ie: end-to-end) test, where the code under test is not visible. So from the tests's perspective, we don't know we are using swift-argument-parser, or maybe we will use a different argument parser in the future, which breaks the behaviour.  As such, tests should be included in swift-package-manager that verify the expected behaviour as a software developer would use the command line.  We should include `large` (ie: end-to-end) tests that cover the expected behaviour, or maybe at least describe the existing behaviour.

PR #8015, opened 2024-10-01
Mark `LegacyHTTPClientRequest` as Sendable

plemarquand:
`LegacyHTTPClientRequest` can be trivially Sendable.

This suppresses some warnings in `LegacyHTTPClient` which itself should also be made concurrecy safe eventually, but it holds some static shared state and so isn't a straightforward conversion.


MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

plemarquand:
@swift-ci test

PR #8016, opened 2024-10-01
Deprecate tvOS 11.0 and iOS 11.0 deployment target API

jakepetroules:
Also update the text for previously deprecated APIs to match the now oldest supported version (12.0 for either of these OSes).

This change previously landed in Xcode 15.

rdar://108752382


jakepetroules:
@swift-ci please test

francescomikulis (approved):

PR #8022, opened 2024-10-03
Add a fallback extension none os dylibs

### LINKED ISSUE #7893: Cannot cross compile when package contains dylib product

- [X] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

swift-mmio contains a dynamic library product called SVD2LLDB. When trying to build a project that only uses the MMIO product of swift-mmio for `armv7em-apple-none-macho` I hit the following assertion:

```
Basics/Triple+Basics.swift:149: Fatal error: Cannot create dynamic libraries for os "noneOS".
```

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

_No response_

_No response_

_No response_

rauhul:
Adds a fake extension for none os dynamic libraries to avoid a SwiftPM crash during build planning when cross compiling to none OS with a dylib product in the graph.

Fixes #7893

### DIFF
--- a/Sources/Basics/Triple+Basics.swift
+++ b/Sources/Basics/Triple+Basics.swift
@@ -146,7 +146,9 @@ extension Triple {
         case .wasi:
             return ".wasm"
         default:
-            fatalError("Cannot create dynamic libraries for os \"\(os)\".")
+            // Fallback to a nonsense extension for operating systems with
+            // unknown conventions.
+            return ".dynamiclibrary"
         }
     }
 
--- a/Sources/Basics/Triple+Basics.swift
+++ b/Sources/Basics/Triple+Basics.swift
@@ -133,7 +133,7 @@ extension Triple {
     /// The file extension for dynamic libraries (eg. `.dll`, `.so`, or `.dylib`)
     public var dynamicLibraryExtension: String {
         guard let os = self.os else {
-            fatalError("Cannot create dynamic libraries unknown os.")
+            fatalError("Cannot create dynamic libraries for unknown os.")
         }
 
         switch os {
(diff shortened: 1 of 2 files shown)


rauhul:
@swift-ci test

Sources/Basics/Triple+Basics.swift
```diff
@@ -146,7 +146,9 @@ extension Triple {
         case .wasi:
             return ".wasm"
         default:
-            fatalError("Cannot create dynamic libraries for os \"\(os)\".")
+            // Fallback to a nonsense extension for operating systems with
+            // unknown conventions.
+            return ".dynamiclibrary"
```
MaxDesiatov, line 151: Any reason this is preferred over `.so`?

rauhul: I don't really like the idea of the extension for "macho" being "so". I think that would be kinda surprising to the user. 

I was considering this default branch determining the extension based on the object file format instead, thought?

xtremekforever: What is the behavior if building on a non-macho platform like Linux? Would this still work?

rauhul: This code path is only used to plan the build of a dynamic library product for an os none triple. Actually doing this build is nonsensical because the dynamic library would require a specific dynamic loader provided by the OS and by definition, os none does not have a dynamic loader.

One alternative option is to use the object format e.g.:
```swift
switch os {
case _ where isDarwin():
    return ".dylib"
case .linux, .openbsd:
    return ".so"
case .win32:
    return ".dll"
case .wasi:
    return ".wasm"
default:
  break
}

guard let objectFormat = self.objectFormat else {
    fatalError("Cannot create dynamic libraries unknown object format.")
}

switch objectFormat {
case .coff:
    return ".coff"
case .elf:
    return ".elf"
case .macho:
    return ".macho"
case .wasm:
    return ".wasm"
case .xcoff:
    return ".xcoff"
}
```

xtremekforever: So this basically just spoofs the "dynamic library" on a none OS but doesn't actually generate anything? That's kind of weird. But if that's the case then I guess it's just arbitrary what you pick there.

rauhul: yeah its not a great thing to do, but fixing the underlying issue with build planning is something the swift pm is working on

xtremekforever: Well, I'm okay with this if it fixes the none OS issue and if it doesn't break anything!

dschaefer2: I think doing this based on object format looks better.

rauhul:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

dschaefer2 (approved):

PR #8025, opened 2024-10-05
Plugin: Stop building with `-enable-testing` even with release config

kateinoigakukun:
### Motivation:

When building a SwiftPM target through PackagePlugin APIs and overall configuration is debug, they were built with `-enable-testing` even with plugins requests building with release configuration.

This mis-configuration led performance and code size regression only when building through package plugins.

The root problem is that some of the properties under `BuildParameters` store values derived from other properties. 
`buildParameters.testing.enableTestability` was one of them. It depends on `configuration` but it's computed only in `BuildParameters.Testing` initializer. Therefore the property was not updated after `PluginDelegate` modifies `configuration` for requested builds.

https://github.com/swiftlang/swift-package-manager/blob/23fa2816afb930d98f529026f4809cc781fb2e2f/Sources/Commands/Utilities/PluginDelegate.swift#L118-L127
https://github.com/swiftlang/swift-package-manager/blob/23fa2816afb930d98f529026f4809cc781fb2e2f/Sources/SPMBuildCore/BuildParameters/BuildParameters%2BTesting.swift#L107

For those who want an immediate workaround: One of the workaround was to set the overall configuration as release, but it forces building plugins as release too.

### Modifications:

This PR made those properties that depend on other stored properties as computed properties rather than stored ones.

There are still some properties having the same issue (`BuildParameters.Debugging`), so I'll fix them in a follow-up PR separately.

### Result:

Release builds requested through package plugins are truly built as release mode even if the overall configuration is debug.

### DIFF
--- a/Tests/SPMBuildCoreTests/BuildParametersTests.swift
+++ b/Tests/SPMBuildCoreTests/BuildParametersTests.swift
@@ -1,0 +1,31 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@testable import SPMBuildCore
+import Basics
+import struct PackageModel.BuildEnvironment
+import _InternalTestSupport
+import XCTest
+
+final class BuildParametersTests: XCTestCase {
+    func testConfigurationDependentProperties() throws {
+        // Ensure that properties that depend on the "configuration" property are
+        // correctly updated after modifying the configuration.
+        var parameters = mockBuildParameters(
+            destination: .host,
+            environment: BuildEnvironment(platform: .linux, configuration: .debug)
+        )
+        XCTAssertEqual(parameters.enableTestability, true)
+        parameters.configuration = .release
+        XCTAssertEqual(parameters.enableTestability, false)
+    }
+}
--- a/Fixtures/Miscellaneous/Plugins/CommandPluginTestStub/Package.swift
+++ b/Fixtures/Miscellaneous/Plugins/CommandPluginTestStub/Package.swift
@@ -29,11 +29,27 @@ let package = Package(
                 .target(name: "plugintool")
             ]
         ),
+        .plugin(
+            name: "check-testability",
+            capability: .command(intent: .custom(
+                verb: "check-testability",
+                description: "Check testability of a target"
+            ))
+        ),
         .executableTarget(
             name: "placeholder"
         ),
         .executableTarget(
             name: "plugintool"
+        ),
+        .target(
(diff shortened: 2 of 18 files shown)


kateinoigakukun:
@swift-ci test

Tests/SPMBuildCoreTests/BuildParametersTests.swift
```diff
@@ -0,0 +1,31 @@
...
+//===----------------------------------------------------------------------===//
+
+@testable import SPMBuildCore
+import Basics
+import struct PackageModel.BuildEnvironment
+import _InternalTestSupport
+import XCTest
+
+final class BuildParametersTests: XCTestCase {
+    func testConfigrationDependentProperties() throws {
```
MaxDesiatov, line 20:
```suggestion
    func testConfigurationDependentProperties() throws {
```

MaxDesiatov:
@swift-ci test

grynspan:
I haven't dug in but I would suggest making sure the flag _is_ still set when calling `--build-tests` or `swift test` since the intent there is pretty clear.

kateinoigakukun:
@grynspan That's a good point. I added a validation test for `swift test -c release`. For `swift build --build-tests -c release`, I'll make a follow-up fix and test cases later because it currently doesn't work.

MaxDesiatov:
@swift-ci test windows

grynspan:
@plemarquand FYI

dschaefer2:
Wouldn't changing flags with swift test, or --build-tests, trigger an expensive full rebuild?

kateinoigakukun:
@dschaefer2 No full rebuilding triggered between `swift build`, `swit build --build-tests`, and `swift test`. Changing between `swift build -c release` and `swift test -c release` triggers a full rebuild but it's acceptable to provide the best performance for release mode. For those who doesn't really want full-rebuilding even with release mode, they can use `swift build -c release --enable-testable-imports` and `swift test -c release` instead to take build time over runtime performance.

dschaefer2:
The community is very sensitive to build times, especially when macros are involved as is the case with swift-testing. And the worst case is building SwiftSyntax in release mode which can take up to 5 minutes.

kateinoigakukun:
I verified this change doesn't work full toolchain build :) 

Since 1. this change does not change anything in CLI behavior but just for plugin invocations, 2. also it just aligns with the current CLI behavior, and 3. we are suffering from a huge performance regression due to this issue, I'd like to proceed this change and discuss the incremental build topic as a separate issue (https://github.com/swiftlang/swift-package-manager/issues/8031)
Does it make sense for you? @MaxDesiatov @dschaefer2

grynspan:
"Doesn't work" or "doesn't break"? 😬

kateinoigakukun:
"Doesn't break"😅

MaxDesiatov (approved):
LGTM, but I hope @xedin could have a look before this is merged

MaxDesiatov:
cc @rauhul who's building for release for embedded

xedin (approved):
Makes sense to me for both to be consistent. Once it is decided what `release` behavior in regards to testability should be (@rauhul has a thread about that on forums) we could change both places to reflect that.

dschaefer2 (approved):
Agreed.

PR #8026, opened 2024-10-07
Tests: Fix always succeeding `XCTAssertAsyncNoThrow`

kateinoigakukun:
The autoclosure expression `{ throw error }` just creates a closure that won't be executed. It leads to the test always succeeding even if the `expression` throws an error.

Fortunately, I didn't find any issue to enable the assertion except for a single case (https://github.com/swiftlang/swift-package-manager/pull/8026/commits/ab1e7849216aba671ed031993972c03f643649cf)

### DIFF
--- a/Fixtures/Miscellaneous/CheckTestLibraryEnvironmentVariable/Tests/CheckTestLibraryEnvironmentVariableTests/CheckTestLibraryEnvironmentVariableTests.swift
+++ b/Fixtures/Miscellaneous/CheckTestLibraryEnvironmentVariable/Tests/CheckTestLibraryEnvironmentVariableTests/CheckTestLibraryEnvironmentVariableTests.swift
@@ -2,7 +2,7 @@ import XCTest
 
 final class CheckTestLibraryEnvironmentVariableTests: XCTestCase {
     func testEnviromentVariable() throws {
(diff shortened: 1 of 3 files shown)


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

kateinoigakukun:
@grynspan This assertion activation revealed that `SWIFT_TESTING_ENABLED` is set to `"0"` only on macOS and not set on other platforms. (introduced by [this change](https://github.com/swiftlang/swift-package-manager/pull/7789/files#diff-0159593debae8d388e5f622aed4f9c62c4f68a7cef68e08466bc803a0cb5529dR194))

Is this intentional behavior or should we set the env variable regardless of the target platform?

grynspan:
It has no effect on other platforms. We set it on macOS so that XCTest doesn't try to host Swift Testing from within Xcode, like it would if you were using Xcode instead of SwiftPM.

It's a bit of an inelegant hack and we may remove it in a future Xcode update.

kateinoigakukun:
Ok, then skip the test on non-macOS platforms

kateinoigakukun:
@swift-ci test

PR #8045, opened 2024-10-15
Use `CaseIterable` to auto-generate CLI `--help` entries

cmcgee1024:
The swift-argument-parser library will automatically show the available values for options that implement the CaseIterable protocol. There are options in SwiftPM that do either list the available values statically in the help message, which can be stale, or don't provide a list at all, which leaves the user to guess the value that they want. Sweep these options, removing the static list, sometimes adding the help message based on the swift documentation. Since the CaseIterable doesn't provide a description of each value, leave the options alone that provide detail for each of their available values.

The result of these changes is that the online help (and content assist) is able to automatically display all of the available values for certain options. Also, some of the options have a help description for the first time.

### DIFF
--- a/Sources/Commands/PackageCommands/AddProduct.swift
+++ b/Sources/Commands/PackageCommands/AddProduct.swift
@@ -42,7 +42,7 @@ extension SwiftPackageCommand {
         @Argument(help: "The name of the new product")
         var name: String
 
-        @Option(help: "The type of target to add.")
+        @Option(help: "The type of target to add")
         var type: CommandProductType = .library
 
         @Option(
--- a/Sources/Commands/PackageCommands/AddProduct.swift
+++ b/Sources/Commands/PackageCommands/AddProduct.swift
@@ -25,7 +25,7 @@ extension SwiftPackageCommand {
     struct AddProduct: SwiftCommand {
         /// The package product type used for the command-line. This is a
         /// subset of `ProductType` that expands out the library types.
-        enum CommandProductType: String, Codable, ExpressibleByArgument {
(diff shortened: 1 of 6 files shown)


Sources/swift-bootstrap/main.swift
```diff
@@ -116,12 +116,12 @@ struct SwiftBootstrapBuildTool: AsyncParsableCommand {
     @Flag()
     public var useIntegratedSwiftDriver: Bool = false
 
-    /// A flag that indicates this build should check whether targets only import
+    /// An option that indicates this build should check whether targets only import
     /// their explicitly-declared dependencies
-    @Option()
+    @Option(help: "An option that indicates this build should check whether targets only import their explicitly-declared dependencies")
```
MaxDesiatov, line 121: nit: IMO it's a good opportunity to make the help entry more readable.

```suggestion
    @Option(help: "Check that targets only import their explicitly-declared dependencies")
```

MaxDesiatov (approved):

Sources/Commands/PackageCommands/AddProduct.swift
```diff
@@ -42,7 +42,7 @@ extension SwiftPackageCommand {
         @Argument(help: "The name of the new product")
         var name: String
 
-        @Option(help: "The type of target to add, which can be one of 'executable', 'library', 'static-library', 'dynamic-library', or 'plugin'")
+        @Option(help: "The type of target to add.")
```
plemarquand, line 45: nit: its not really standard whether doc strings end with a period `.` or not. Most don't seem to, so maybe we standardize on that.

MaxDesiatov:
@swift-ci test

jakepetroules (approved):
Nice!

PR #8046, opened 2024-10-16
Add descriptive error when attempting to list tests before building

plemarquand:
### Motivation:

Typically a build is performed before a `swift test list` command, but by passing the `--skip-build` flag it is possible to get a cryptic, unhelpful error message if the test artifacts haven't already been built.

### Modifications:

Add a descriptive error for this case, only checking for test artifact existence in the failure case to avoid any performance impact on the happy path.

### DIFF
--- a/Sources/Commands/SwiftTestCommand.swift
+++ b/Sources/Commands/SwiftTestCommand.swift
@@ -32,6 +32,7 @@ import SPMBuildCore
 import func TSCLibc.exit
 import Workspace
 
+import struct TSCBasic.FileSystemError
 import struct TSCBasic.ByteString
 import enum TSCBasic.JSON
 import class Basics.AsyncProcess
@@ -725,6 +726,17 @@ extension SwiftTestCommand {
         var _deprecated_passthrough: Bool = false
 
         func run(_ swiftCommandState: SwiftCommandState) async throws {
+            do {
+                try await self.runCommand(swiftCommandState)
+            } catch let error as FileSystemError {
+                if sharedOptions.shouldSkipBuilding {
+                    throw ErrorWithContext(error, "Test build artifacts were not found in the build folder. Found the --skip-build flag; build the tests first or rerun the command without --skip-build")
+                }
+                throw error
+            }
+        }
+
+        func runCommand(_ swiftCommandState: SwiftCommandState) async throws {
             let (productsBuildParameters, toolsBuildParameters) = try swiftCommandState.buildParametersForTest(
                 enableCodeCoverage: false,
                 shouldSkipBuilding: sharedOptions.shouldSkipBuilding
@@ -774,13 +786,25 @@ extension SwiftTestCommand {
                     )
 
                     // Finally, run the tests.
+                    var output = [String]()
                     let result = runner.test(outputHandler: {
-                        // command's result output goes on stdout
-                        // ie "swift test" should output to stdout
-                        print($0, terminator: "")
+                        output.append($0)
                     })
                     if result == .failure {
                         swiftCommandState.executionStatus = .failure
+                        // If the runner reports failure do a check to ensure
+                        // all the binaries are present on the file system.
+                        try testProducts.map(\.binaryPath).forEach {
+                            if !swiftCommandState.fileSystem.exists($0) {
+                                throw FileSystemError(.noEntry, $0)
+                            }
+                        }
+                    } else {
+                        output.forEach {
+                            // command's result output goes on stdout
+                            // ie "swift test" should output to stdout
+                            print($0, terminator: "")
+                        }
                     }
                 } else if let testEntryPointPath {
                     // Cannot run Swift Testing because an entry point file was used and the developer
--- a/Tests/CommandsTests/TestCommandTests.swift
+++ b/Tests/CommandsTests/TestCommandTests.swift
@@ -284,6 +284,14 @@ final class TestCommandTests: CommandsTestCase {
         }
     }
 
+    func testListWithSkipBuildAndNoBuildArtifacts() async throws {
+        try await fixture(name: "Miscellaneous/TestDiscovery/Simple") { fixturePath in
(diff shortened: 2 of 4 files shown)


Sources/Commands/SwiftTestCommand.swift
```diff
@@ -730,7 +730,7 @@ extension SwiftTestCommand {
                 try await self.runCommand(swiftCommandState)
             } catch let error as FileSystemError {
                 if sharedOptions.shouldSkipBuilding {
-                    throw ErrorWithContext(error, "Test build artifacts were not found in the build folder. Found the --skip-build flag; build the tests first or rerun the command without --skip-build")
+                    throw ErrorWithContext(error, "Test build artifacts were not found in the build folder.\n\tThe --skip-build flag was provided; either build the tests first with \"swift build --build tests\" or rerun the \"swift test list\" command without --skip-build")
```
MaxDesiatov, line 733: formatting nit: line length here clearly exceeds [120 characters specified in the formatter config](https://github.com/swiftlang/swift-package-manager/blob/d28f725139357fbb1b20148964177fef5aab3639/.swiftformat#L15C12-L15C15
), let's use a multi-line string 
```suggestion
                    throw ErrorWithContext(error, """
                        Test build artifacts were not found in the build folder.
                        The `--skip-build` flag was provided; either build the tests first with \
                        `swift build --build tests` or rerun the `swift test list` command without \
                        `--skip-build`
                        """
                    )
```

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -774,13 +786,25 @@ extension SwiftTestCommand {
...
-                        // command's result output goes on stdout
-                        // ie "swift test" should output to stdout
-                        print($0, terminator: "")
+                        output.append($0)
                     })
                     if result == .failure {
                         swiftCommandState.executionStatus = .failure
+                        // If the runner reports failure do a check to ensure
+                        // all the binaries are present on the file system.
+                        try testProducts.map(\.binaryPath).forEach {
```
MaxDesiatov, line 797: If an error is thrown here, does this mean that everything collected in `output` is lost?

Also, coding style nit: we prefer `for` loops instead of `forEach`.

Tests/CommandsTests/TestCommandTests.swift
```diff
@@ -284,6 +284,14 @@ final class TestCommandTests: CommandsTestCase {
         }
     }
 
+    func testListWithSkipBuildAndNoBuildArtifacts() async throws {
+        try await fixture(name: "Miscellaneous/TestDiscovery/Simple") { fixturePath in
+            await XCTAssertThrowsCommandExecutionError(try await SwiftPM.Test.execute(["list", "--skip-build"], packagePath: fixturePath)) { error in
```
MaxDesiatov, line 289: line length nit
```suggestion
            await XCTAssertThrowsCommandExecutionError(
                try await SwiftPM.Test.execute(["list", "--skip-build"], packagePath: fixturePath)
            ) { error in
```

Sources/Commands/SwiftTestCommand.swift
plemarquand: Done

Sources/Commands/SwiftTestCommand.swift
plemarquand: Done

Tests/CommandsTests/TestCommandTests.swift
plemarquand: Done

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -725,6 +726,17 @@ extension SwiftTestCommand {
         var _deprecated_passthrough: Bool = false
 
         func run(_ swiftCommandState: SwiftCommandState) async throws {
+            do {
+                try await self.runCommand(swiftCommandState)
+            } catch let error as FileSystemError {
+                if sharedOptions.shouldSkipBuilding {
+                    throw ErrorWithContext(error, "Test build artifacts were not found in the build folder. Found the --skip-build flag; build the tests first or rerun the command without --skip-build")
```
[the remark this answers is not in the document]
plemarquand: The wrapped error contains the detailed information about what file wasn't found. Here is the error message as it stands right now:

```
error: 	Test build artifacts were not found in the build folder.
	The `--skip-build` flag was provided; either build the tests first with `swift build --build tests` or rerun the `swift test list` command without `--skip-build`
	/Users/plemarquand/work/testapps/DelMeDelMe/.build/arm64-apple-macosx/debug/DelMeDelMePackageTests.xctest doesn't exist in file system
```

Because the xctest codepath throws a FileSystemError as soon as it encounters a missing file we'd need to rework the architecture a bit to keep iterating and collect up all the errors in order to show a list of all the files that aren't found. I can go that route if need be but its going to make the code more complex for a small gain I think.

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -774,13 +792,25 @@ extension SwiftTestCommand {
...
-                        print($0, terminator: "")
+                        output.append($0)
                     })
                     if result == .failure {
                         swiftCommandState.executionStatus = .failure
+                        // If the runner reports failure do a check to ensure
+                        // all the binaries are present on the file system.
+                        for path in testProducts.map(\.binaryPath) {
+                            if !swiftCommandState.fileSystem.exists(path) {
+                                throw FileSystemError(.noEntry, path)
```
MaxDesiatov, line 805: Errors thrown from here still don't flush accumulated `output`, which IIUC is a change from how it previously worked?

plemarquand: I've reverted this. In the failure case nothing was printed, but if there is more than one swift-testing testProduct then the successful ones should still print.

MaxDesiatov:
@swift-ci test

plemarquand:
@swift-ci test

PR #8047, opened 2024-10-16
Use `readData` to read data from a pipe instead of `availableData`

ahoppen:
Using `availableData` causes parallel XCTest execution to hang on Windows if the tests output significant data to stderr.

### DIFF
--- a/Sources/Basics/Concurrency/AsyncProcess.swift
+++ b/Sources/Basics/Concurrency/AsyncProcess.swift
@@ -501,7 +501,7 @@ package final class AsyncProcess {
 
             group.enter()
             stdoutPipe.fileHandleForReading.readabilityHandler = { (fh: FileHandle) in
-                let data = fh.availableData
+                let data = (try? fh.read(upToCount: Int.max)) ?? Data()
                 if data.count == 0 {
                     stdoutPipe.fileHandleForReading.readabilityHandler = nil
                     group.leave()
@@ -516,7 +516,7 @@ package final class AsyncProcess {
 
             group.enter()
             stderrPipe.fileHandleForReading.readabilityHandler = { (fh: FileHandle) in
-                let data = fh.availableData
+                let data = (try? fh.read(upToCount: Int.max)) ?? Data()
                 if data.count == 0 {
                     stderrPipe.fileHandleForReading.readabilityHandler = nil
                     group.leave()


ahoppen:
@swift-ci Please test

Sources/Basics/Concurrency/AsyncProcess.swift
```diff
@@ -501,7 +501,7 @@ package final class AsyncProcess {
 
             group.enter()
             stdoutPipe.fileHandleForReading.readabilityHandler = { (fh: FileHandle) in
-                let data = fh.availableData
+                let data = fh.readData(ofLength: Int.max)
```
MaxDesiatov, line 504: This method is deprecated, `read(upToCount:)` is the recommended replacement.

ahoppen: Good suggestion. Thank you!

ahoppen:
@swift-ci Please test

dschaefer2 (approved):
I think we all have scars from using availableData to read stdout/stderr. Thanks @ahoppen!

jakepetroules (commented):
Are you sure this will work?. [availableData](https://github.com/swiftlang/swift-corelibs-foundation/blob/cb6b2fa46992e04ae3f77fae30e2d86cebd2b440/Sources/Foundation/FileHandle.swift#L220) just calls the same underlying API as [read(upToCount:)](https://github.com/swiftlang/swift-corelibs-foundation/blob/cb6b2fa46992e04ae3f77fae30e2d86cebd2b440/Sources/Foundation/FileHandle.swift#L506), and it still has the ability to block the calling thread.

```swift
    private func _checkFileHandle() {
        precondition(self._handle != INVALID_HANDLE_VALUE, "Invalid file handle")
    }
```

```swift
    open var availableData: Data {
        _checkFileHandle()
        do {
            let readResult = try _readDataOfLength(Int.max, untilEOF: false)
            return readResult.toData()
        } catch {
            fatalError("\(error)")
        }
    }
```

```swift
    public func read(upToCount count: Int) throws -> Data? {
        guard self != FileHandle._nulldeviceFileHandle else { return nil }
        
        let result = try _readDataOfLength(count, untilEOF: true)
        return result.length == 0 ? nil : result.toData()
    }
```

ahoppen:
Yes, I tested that we no longer see the hang when using `read(upToCount:)`. I don’t know why it fixes it but 🤷🏽 maybe it’s because one of them does `untilEOF: true` and the other one `untilEOF: false`.

dschaefer2:
> Yes, I tested that we no longer see the hang when using read(upToCount:). I don’t know why it fixes it but 🤷🏽 maybe it’s because one of them does untilEOF: true and the other one untilEOF: false.

That feels true. The hangs I've seen have been at the end of the stream where EOF isn't properly detected.

PR #8049, opened 2024-10-16
Support separate modules for static and dynamic exporting symbols for Windows

dschaefer2:
On Windows, there is a limit of around 64K to the number of symbols a DLL/EXE can export. We hit that regularly with large projects like SwiftPM itself. This hides those symbols and only exposes the ones requested for a DLL.

For Windows triples only, creates a parallel build graph for Swift modules, one for static linking using -static, and one for exporting linking which is the default. DLL products consume their direct target dependencies as exporting. All other dependencies use the static versions to eliminate unnecessary symbol exports.

The bulk of this is managed by the SwiftModuleBuildDescription which will create a duplicate of itself for the exporting case and set its own type to static linking. Both modules are fed to the planner to create llbuild swift command tasks. Code is added for dynamic libraries to hook up the correct inputs for exporting the symbols in the libraries targets.

This is WIP as we need to do a lot of testing to ensure we didn't break anything. Ensuring this only affects builds for Windows triples helps mitigate that.

### DIFF
--- a/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
@@ -128,8 +128,8 @@ public final class SwiftModuleBuildDescription {
     var modulesPath: AbsolutePath {
         let suffix = self.buildParameters.suffix
         var path = self.buildParameters.buildPath.appending(component: "Modules\(suffix)")
-        if self.windowsTargetType == .exporting {
-            path = path.appending("exporting")
+        if self.windowsTargetType == .dynamic {
+            path = path.appending("dynamic")
         }
         return path
     }
@@ -268,17 +268,18 @@ public final class SwiftModuleBuildDescription {
     /// Whether to disable sandboxing (e.g. for macros).
     private let shouldDisableSandbox: Bool
 
-    /// For Windows we default to static objects and create copies for objects
-    /// That export symbols. This will allow consumers to select which one they want.
+    /// For Windows, we default to static objects and but also create objects
+    /// that export symbols for DLLs. This allows library targets to be used
+    /// in both contexts
     public enum WindowsTargetType {
         case `static`
-        case exporting
+        case dynamic
     }
     /// The target type. Leave nil for non-Windows behavior.
     public let windowsTargetType: WindowsTargetType?
 
-    /// The corresponding target symbols exporting (not -static)
-    public private(set) var windowsExportTarget: SwiftModuleBuildDescription? = nil
+    /// The corresponding target for dynamic library export (i.e., not -static)
+    public private(set) var windowsDynamicTarget: SwiftModuleBuildDescription? = nil
 
     /// Create a new target description with target and build parameters.
     init(
@@ -338,7 +339,7 @@ public final class SwiftModuleBuildDescription {
         if buildParameters.triple.isWindows() {
             // Default to static and add another target for DLLs
             self.windowsTargetType = .static
-            self.windowsExportTarget = .init(windowsExportFor: self)
+            self.windowsDynamicTarget = .init(windowsExportFor: self)
         } else {
             self.windowsTargetType = nil
         }
@@ -366,9 +367,9 @@ public final class SwiftModuleBuildDescription {
 
     /// Private init to set up exporting version of this module
     private init(windowsExportFor parent: SwiftModuleBuildDescription) {
-        self.windowsTargetType = .exporting
-        self.windowsExportTarget = nil
-        self.tempsPath = parent.tempsPath.appending("exporting")
+        self.windowsTargetType = .dynamic
+        self.windowsDynamicTarget = nil
+        self.tempsPath = parent.tempsPath.appending("dynamic")
 
         // The rest of these are just copied from the parent
         self.package = parent.package
@@ -572,7 +573,7 @@ public final class SwiftModuleBuildDescription {
             case .static:
                 // Static on Windows
                 args += ["-static"]
-            case .exporting:
+            case .dynamic:
                 // Add the static versions to the include path
                 // FIXME: need to be much more deliberate about what we're including
                 args += ["-I", self.modulesPath.parentDirectory.pathString]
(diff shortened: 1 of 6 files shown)


Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
```diff
@@ -127,7 +127,11 @@ public final class SwiftModuleBuildDescription {
 
     var modulesPath: AbsolutePath {
         let suffix = self.buildParameters.suffix
-        return self.buildParameters.buildPath.appending(component: "Modules\(suffix)")
+        var path = self.buildParameters.buildPath.appending(component: "Modules\(suffix)")
+        if self.windowsTargetType == .exporting {
+            path = path.appending("exporting")
+        }
+        return path
```
dschaefer2, line 134: I really don't like how we're putting all the modules in a single directory and adding it to the include path. Since the exporting modules have the same name, I had to create another directory to put them in and add that include path for the products that consume them.

dschaefer2:
@swift-ci please test

Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
rauhul: yeah putting all the modules in the same directory is extra annoying because you can end up `import`ing things you haven't declared a dependency on.

Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
```diff
@@ -264,6 +268,18 @@ public final class SwiftModuleBuildDescription {
     /// Whether to disable sandboxing (e.g. for macros).
     private let shouldDisableSandbox: Bool
 
+    /// For Windows we default to static objects and create copies for objects
+    /// That export symbols. This will allow consumers to select which one they want.
+    public enum WindowsTargetType {
+        case `static`
+        case exporting
```
compnerd, line 275: Can we use `dynamic` instead?

dschaefer2: :). I had been using `dynamic` until the very last minute. It's really about exporting symbols from DLLs and was starting to think of this in C terms as in static symbols versus `__declspec(dllexport)`. Maybe dllexport is a better name?

Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
dschaefer2: Question is, who do I break if I change this?

dschaefer2:
One other issue I'm fighting is when building SwiftPM, I'm still getting the exporting imported symbols warning for the yaml_parser functions, despite having the YAML_DECLARE_STATIC flag being passed to the CYaml target.

Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
jakepetroules: DLL is **Dynamic** Link Library, I think the "dynamic" terminology might be clearer / more familiar / approachable for folks compared to something that might be more _technically_ accurate.

dschaefer2: Fair. I'm only voting 1/2 since it literally was "dynamic" until a hour before I did this push :). I'll change it when I add some tests in the next push.

compnerd:
`-DYAML_DECLARE_STATIC` should do the trick IIRC. Are we sure that it is getting to the C compiler and is also getting to the clang importer as `-Xcc -DYAML_DECLARE_STATIC`?

rauhul:
if I recall correctly, CSettings on a C target are _not_ passed to Swift targets which depend on that target as `-Xcc` flags

dschaefer2:
> if I recall correctly, CSettings on a C target are not passed to Swift targets which depend on that target as -Xcc flags

The warnings I'm seeing are:

> lld-link: warning: C:\\Users\\dschaefer2\\swift\\swiftpm2\\.build\\x86_64-unknown-windows-msvc\\debug\\Yams.build\\Parser.swift.o: locally defined symbol imported: yaml_event_delete (defined in C:\\Users\\dschaefer2\\swift\\swiftpm2\\.build\\x86_64-unknown-windows-msvc\\debug\\CYaml.build\\src\\api.c.o) [LNK4217]

Which does imply the issue is on the Swift side. That's something worth trying.

dschaefer2:
> if I recall correctly, CSettings on a C target are not passed to Swift targets which depend on that target as -Xcc flags

Yup, @DougGregor ran into that with swift-java too. I've created https://github.com/swiftlang/swift-package-manager/issues/8095 to track that. (I'm sure that's another issue on that somewhere but I didn't spend enough time trying to find it :))

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
dschaefer2: Done.

dschaefer2:
This is ready for review. I'd like to start giving it some soak time with the snapshot builds once approved.

jakepetroules (approved):

hjyamauchi:
Hi @dschaefer2 it seems like the following issue started after this PR/commit: https://github.com/swiftlang/swift/issues/77497

I'm not sure if it's an issue on the user side or the spm side.

Wdyt?

dschaefer2:
> Hi @dschaefer2 it seems like the following issue started after this PR/commit: [swiftlang/swift#77497](https://github.com/swiftlang/swift/issues/77497)
> 
> I'm not sure if it's an issue on the user side or the spm side.
> 
> Wdyt?

Thanks! I'll take a look. I am worried about side effects I didn't think about. This could be one.

compnerd:
I think that this is a modelling issue. The problem is that there are two libraries SwiftWin32UI and SwiftWin32. SwiftWin32UI depends on SwiftWin32. SwiftWin32 however has a dependency on SwiftCOM which is package external and has a dynamic library product. When the SwiftWin32 product is built (dynamic library), we link dynamically to SwiftCOM and everything is happy. When building SwiftWin32UI, we statically link against SwiftWin32. This uses the object library for SwiftWin32. However, this object library was built against the static library (object library) for SwiftCOM. We now attempt to link with `SwiftCOM.lib` (the import library for SwiftCOM) which assumes dynamic linking for symbols that we are expecting statically as we are building against the static library.

dschaefer2:
I've brought that issue over here as #8110 . We can continue diagnosis there.

PR #8050, opened 2024-10-16
Ensure a newline is printed after Planning build

plemarquand:
If a PackageStructure build needs to occur during a `swift test list --skip-build` command there is no newline printed to the terminal, causing the first line of listed tests to appear on the same line as the progress, i.e:

`[1/1] Planning buildTestPackageTests.example()`

This doesn't occur when you omit `--skip-build` because there is code in the `LLBuildProgressTracker.swift` that calls
`progressAnimation.complete(success:)` when a build completes. Ideally I'd like to contain this logic to the LLBuildProgressTracker, however for a PackageStructure build the relevant buildComplete delegate method is never called.


jakepetroules (approved):

plemarquand:
@swift-ci test

PR #8051, opened 2024-10-17
SE-0387: add `--toolset` option to build-related subcommands

MaxDesiatov:
### Motivation:

The only feature proposed in [SE-0387](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0387-cross-compilation-destinations.md) that remains unimplemented is the `--toolset` CLI option:

> We propose that users also should be able to pass `--toolset <path_to_toolset.json>` option to `swift build`, `swift test`, and `swift run`.
>
> We'd like to allow using multiple toolset files at once. This way users can "assemble" toolchains on the fly out of tools that in certain scenarios may even come from different vendors. A toolset file can have an arbitrary name, and each file should be passed with a separate `--toolset` option, i.e. `swift build --toolset t1.json --toolset t2.json`.
>
> All of the properties related to names of the tools are optional, which allows merging configuration from multiple
toolset files. For example, consider `toolset1.json`:
>```json5
>{
>  "schemaVersion": "1.0",
>  "swiftCompiler": {
>    "path": "/usr/bin/swiftc",
>    "extraCLIOptions": ["-Xfrontend", "-enable-cxx-interop"]
>  },
>  "cCompiler": {
>    "path": "/usr/bin/clang",
>    "extraCLIOptions": ["-pedantic"]
>  }
>}
>```
>
> and `toolset2.json`:
>
> ```json5
>{
>  "schemaVersion": "1.0",
>  "swiftCompiler": {
>    "path": "/custom/swiftc"
>  }
>}
>```
>
> With multiple `--toolset` options, passing both of those files will merge them into a single configuration. Tools passed in subsequent `--toolset` options will shadow tools from previous options with the same names. That is, 
>`swift build --toolset toolset1.json --toolset toolset2.json` will build with `/custom/swiftc` and no extra flags, as specified in `toolset2.json`, but `/usr/bin/clang -pedantic` from `toolset1.json` will still be used.
>
> Tools not specified in any of the supplied toolset files will be looked up in existing implied search paths that are used without toolsets, even when `rootPath` is present. We'd like toolsets to be explicit in this regard: if a tool would like to participate in toolset path lookups, it must provide either a relative or an absolute path in a toolset.
>
> Tools that don't have `path` property but have `extraCLIOptions` present will append options from that property to a tool with the same name specified in a preceding toolset file. If no other toolset files were provided, these options will be appended to the default tool invocation.

### Modifications:

Added `toolsetPaths` on `LocationOptions`, which passes new `customToolsets` argument on `SwiftSDK.deriveTargetSwiftSDK`. Added corresponding tests.

### Result:

New `--toolset` option is able to accept toolset JSON files.

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -130,7 +130,7 @@ public struct LocationOptions: ParsableArguments {
         help: "Specify to a toolset JSON file to use when building for the target platform",
         completion: .file(extensions: [".json"])
     )
-    public var toolsetPath: AbsolutePath?
+    public var toolsetPaths: [AbsolutePath] = []
 
     @Option(
         name: .customLong("pkg-config-path"),
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -127,7 +127,11 @@ public struct LocationOptions: ParsableArguments {
     
     @Option(
         name: .customLong("toolset"),
-        help: "Specify to a toolset JSON file to use when building for the target platform",
+        help: """
+            Specify to a toolset JSON file to use when building for the target platform. \
+            Use the option multiple times to specify more than one toolset. Toolsets will be merged in the order \
+            they're specified into a single final toolset for the current build.
+            """,
         completion: .file(extensions: [".json"])
     )
     public var toolsetPaths: [AbsolutePath] = []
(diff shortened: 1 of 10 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

Issue #8055, opened 2024-10-18
Regression: Binary artifacts in C++ targets break the build planning phase.


furby-tm:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [X] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

> [!IMPORTANT]
> This is only a regression in the development `main` branch, **not** in the official release. In the following I refer to the `main` branch of SwiftPM, the `swift-build` executable built with `swift build -c release` by the alias `swiftpmdev`.

Running `swift build` (debug or release) can no longer resolve binary artifacts in package dependencies, with an error message like the following:
```swift
error: InternalError(description: "Internal error. Please file a bug at https://github.com/swiftlang/swift-package-manager/issues with this info. unknown module: <ResolvedModule: Python, binary>")
```

> [!TIP]
> This can only be reproduced within C++ targets within SwiftPM Package manifests, Swift targets remain unaffected.

### Expected behavior

Running `swift build` (debug or release) should have no problems resolving binary artifacts in package dependencies, as all versions of **Swift 5.9** - **Swift 6.0** did not experience this issue.

```swift
$ git clone git@github.com:wabiverse/SwiftPMBinaryTargetRegression.git
$ cd SwiftPMBinaryTargetRegression

$ swift build
```

```swift
Building for debugging...
```

### Actual behavior

Running `swift build` (debug or release) now has problems resolving binary artifacts in C++ package dependencies, as of Swift 6.1.0-dev.

```swift
$ git clone git@github.com:wabiverse/SwiftPMBinaryTargetRegression.git
$ cd SwiftPMBinaryTargetRegression

$ swiftpmdev build
```

```swift
error: InternalError(description: "Internal error. Please file a bug at https://github.com/swiftlang/swift-package-manager/issues with this info. unknown module: <ResolvedModule: Python, binary>")
```

### Steps to reproduce

1. Run the following in your terminal against the `main` branch of SwiftPM:
    ```swift
    git clone git@github.com:wabiverse/SwiftPMBinaryTargetRegression.git
    cd SwiftPMBinaryTargetRegression
    
    swiftpmdev build
    ```

2. Notice the error when resolving binary artifacts.
    ```swift
[...]

PR #8056, opened 2024-10-19
Binary dependencies of C++ module break build planning

furby-tm:
Fix a regression where the build planning phase will early terminate if SwiftPM packages contain binary targets that C++ products refer to, https://github.com/swiftlang/swift-package-manager/issues/8055.

### Motivation:

When running `swift build` on SwiftPM packages with binary targets that C++ products refer to, they will early terminate during the build planning phase with the following error:

```swift
error: InternalError(description: "Internal error. Please file a bug at https://github.com/swiftlang/swift-package-manager/issues with this info. unknown module: <ResolvedModule: Python, binary>")
```

### Modifications:

Remove the guard, so we do not fail ~~if the module dependency does not yet exist in the build plan's target map, which I think is normal~~ on binary targets that C++ products refer to, since they don't have build descriptions.

### Result:

When running `swift build` on SwiftPM packages with binary targets that C++ products refer to, the build planning phase will proceed successfully without error.


MaxDesiatov (changes requested):
What does "large packages" mean here and how that and binary artifacts are related to changed lines of code?

But first of all, this needs a self-contained test case added to the test suite, so that we can verify that the change actually fixes the issue and doesn't regress with future changes.

furby-tm:
@MaxDesiatov Sorry, I inferred from the issue linked above #8055 should give you more insight.

> What does "large packages" mean here (what exactly would make a package large in this case?)

This is all you have to do to replicate binary artifacts not resolving in the `main` branch of swiftpm:

```swift
git clone git@github.com:wabiverse/MetaverseKit.git
cd MetaverseKit

swift build
```

> and how that and binary artifacts are related to changed lines of code?

The guard early exits swiftpm, because the binary target: `Python` in this case was not present in the `BuildPlan`'s `targetMap`, which is a `IdentifiableSet<ModuleBuildDescription>`.

> But first of all, this needs a self-contained test case added to the test suite, so that we can verify that the change actually fixes the issue and doesn't regress with future changes.

I can attempt to re-create a pretty large dependency cycle as seen in the `MetaverseKit` SwiftPM package, if that's something suitable to be added to the test suite.

furby-tm:
My underlying assumption here is that it is considered normal for some modules to not be in that `targetMap: Set<ModuleBuildDescription>` at the time of invocation, also given the optional parameter in `addStaticTargetInputs(_ description: ModuleBuildDescription?)`.

This was also the previous behavior (no `guard`) before this revision:
https://github.com/swiftlang/swift-package-manager/pull/7883/files#diff-3119c9dc1513ef3b1cfb8088dce4a14fbb9d33f692c96f578161b6abc7d1cd47L57-L58

furby-tm:
@MaxDesiatov good news, I was able to [minimally reproduce this](https://github.com/wabiverse/SwiftPMBinaryTargetRegression), it isn't related to a large dependency cycle.

This issue is not related to Swift targets, you can only reproduce this regression in C++ targets, the underlying issue is that binary target dependencies specified within C++ targets do not resolve without this PR.

```swift
git clone git@github.com:wabiverse/SwiftPMBinaryTargetRegression.git
cd SwiftPMBinaryTargetRegression

swift build
```

furby-tm:
@MaxDesiatov 

I have added a [reproducible test to the existing XCFramework fixture](https://github.com/swiftlang/swift-package-manager/pull/8056/commits/81e0cbfa55417629ca4c16eb4710837ad1ebd3ad), one final thing to note, I realized that this regression only affects binary target dependencies brought in from external **_package dependencies_**, within C++ targets.

Fixtures/DependencyResolution/External/XCFramework/Inner/Package.swift
MaxDesiatov, line 1: Where is this newly fixture package built? I don't see any new test cases utilizing this fixture added in this PR.

Fixtures/DependencyResolution/External/XCFramework/Package.swift
```diff
@@ -5,10 +5,14 @@ import PackageDescription
 let package = Package(
     name: "Foo",
     products: [
-        .library(name: "Foo", targets: ["Foo", "Bar"]),
+        .library(name: "Foo", targets: ["Foo", "Bar", "Baz"]),
```
MaxDesiatov, line 8: Mind setting up a separate test case for this? I'm worried that editing an existing fixture could have unintended effects on already existing test cases.

furby-tm: No problem, I will add it to a new test case.

furby-tm: I have setup a separate test case for this: `XCFrameworkExternal`, so we don't introduce unintended effects on existing test cases utilizing `XCFramework`.

Fixtures/DependencyResolution/External/XCFrameworkExternal/Package.swift
MaxDesiatov, line 1: With a new test fixture added this needs a new test case function exercising it, which would fail on `main`, but pass with this PR.

MaxDesiatov (changes requested):

Fixtures/DependencyResolution/External/XCFrameworkExternal/Package.swift
furby-tm: Sorry, I'm still quite new to the `swift-package-manager` codebase and its respective test cases.

I see the `XCFramework` test fixture utilized in these 4 places:
- https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Tests/CommandsTests/PackageCommandTests.swift#L115
- https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Tests/CommandsTests/PackageCommandTests.swift#L132
- https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Tests/CommandsTests/SwiftCommandStateTests.swift#L168
- https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Tests/CommandsTests/SwiftCommandStateTests.swift#L203

For which of the existing tests do you think is a good place to add a new test case function to exercise `XCFrameworkExternal`?

MaxDesiatov: It depends on what you need to have access to and what exactly the fix covers. `SwiftCommandStateTests` is almost exclusively for testing the `SwiftCommandState` API. `PackageCommandTests` is for testing `swift package` CLI invocations. I don't think your fixture belongs to either, but then the nature of your fix is unclear to me in general, in your "a needless early-out at the guard" explanation, what exactly makes you think it's needless? 

The title of the PR says "fix ... not resolving", but then your change is in the build planning phase and not in the modules resolution phase. You'd have to pinpoint the exact logic you're changing and test that bit specifically.

furby-tm: > The title of the PR says "fix ... not resolving", but then your change is in the build planning phase and not in the modules resolution phase. You'd have to pinpoint the exact logic you're changing and test that bit specifically

@MaxDesiatov I have updated the title to reflect the purpose of this PR.

> I don't think your fixture belongs to either, but then the nature of your fix is unclear to me in general, in your "a needless early-out at the guard" explanation, what exactly makes you think it's needless?

For these reasons I [mentioned previously](https://github.com/swiftlang/swift-package-manager/pull/8056#issuecomment-2424234649):

1. This PR retains the previous (no `guard`) behavior of SwiftPM before [this revision](https://github.com/swiftlang/swift-package-manager/pull/7883/files#diff-3119c9dc1513ef3b1cfb8088dce4a14fbb9d33f692c96f578161b6abc7d1cd47L57-L58) (added a `guard`) on **Aug 15th**.

2. Also of note, this PR now matches the same (no `guard`) logic in [L.485 **`LLBuildManifestBuilder+Swift.swift`**](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/Build/BuildManifest/LLBuildManifestBuilder%2BSwift.swift#L482-L490).

3. Given the optional parameter in `addStaticTargetInputs(_ description: ModuleBuildDescription?)`.

<hr/>

The removal of the `guard` is not new behavior, it is to match the current behavior in point 2, and maintain the previous behavior of SwiftPM, when binary artifact dependencies in C++ targets did not break (early terminate) the build planning phase.

Fixtures/DependencyResolution/External/XCFrameworkExternal/Foo/Foo.swift
```diff
@@ -0,0 +1,3 @@
+public func foo() {
```
MaxDesiatov, line 1: Is this function called from anywhere? Can this file stay empty and the test still valid that way?

furby-tm: whoops, sorry on the title change, I had not realized you changed it when I changed the title.

This function is not called from anywhere, I was copying over the fixture behavior of `XCFramework` which had a `Foo` target with the same code, I can actually just go ahead and delete this `Foo` Swift target if you'd like, and only build the `Baz` target, since the underlying problem is with the `Baz` C++ target.

MaxDesiatov: Yes, we should make new tests as clean as possible. But as I said, we may not need a fixture a tool if you can construct build planning inputs like existing build planning tests already do.

furby-tm: I went ahead and removed the `Foo` Swift target from the new `XCFrameworkExternal` fixture, this fixture (if we are keeping it) now only has a single `Baz` C++ target, and I went ahead and just reconfirmed this error is still reproducible in this fixture when running `swift build` from the `main` branch.

Fixtures/DependencyResolution/External/XCFrameworkExternal/Package.swift
furby-tm: That's more than fine with me, whatever is the preferred approach here that still allows the bug to be minimally reproducible.

Just to confirm, would it be preferable to add a new test case to replicate this bug in https://github.com/swiftlang/swift-package-manager/blob/main/Tests/BuildTests/ClangTargetBuildDescriptionTests.swift, or would it be better to setup a whole new file for this test under `BuildTests`?

MaxDesiatov: That file is not too big, so adding a new test function to it is fine by me. Thanks!

xedin:
Just came here to say that the fix is correct, there shouldn't have been `guard` in there in the first place because binary targets that C++ products refer to don't have build descriptions and `addStaticTargetInputs` is expected to handle only swift library targets. Might be worth it to rename `addStaticTargetInputs` into something like `addStaticSwiftLibraryInputs` or something similar.

xedin:
@furby-tm and @MaxDesiatov I think all we need to test this issue is to form a BuildPlan, build and analyze the compile command, we don't actually need fixtures here.

furby-tm:
> Just came here to say that the fix is correct, there shouldn't have been `guard` in there in the first place because binary targets that C++ products refer to don't have build descriptions and `addStaticTargetInputs` is expected to handle only swift library targets. Might be worth it to rename `addStaticTargetInputs` into something like `addStaticSwiftLibraryInputs` or something similar.

Thanks for the clarification! That makes much more sense than my earlier attempt to explain:
> “we do not fail if the module dependency does not yet exist in the build plan’s target map, which I think is normal.”
In retrospect, that doesn’t make sense at all. Please disregard my earlier comments; I’m still learning the nuances of SwiftPM and familiarizing myself with the codebase, feel free to write me off as the crazy new guy lol. 😅

@MaxDesiatov, I will remove the new fixture I added and include the test case we discussed. I apologize for my chaotic explanation, which understandably caused some confusion. I appreciate your willingness to help me with this PR, even if you probably felt like throwing your MacBook at me in the process! 😭🙏

Fixtures/DependencyResolution/External/XCFrameworkExternal/Package.swift
furby-tm: As an aside, forgive my confusion here a bit but after running into a:
```
failed: caught error: "The package does not contain a buildable target.
Add at least one `.target` or `.executableTarget` to your `Package.swift`."
```
error while attempting to re-create a SwiftPM package in this completely different approach to the current SwiftPM Package Manifest API (using internal APIs and things that don't reflect a "consistent with the available API" approach of how any package manifests are authored), I find it odd that we are creating tests in this way.

I'm sure I will fix this error and get a test case that eventually works, I'm not too worried about that. But I suppose I just find these tests a bit odd, at least on the outset.

furby-tm:
I offered some general feedback on SwiftPM's current method of creating many of these tests [here](https://github.com/swiftlang/swift-package-manager/pull/8056#discussion_r1815779685), if helpful.

xedin (approved):
LGTM! We are going to add a test in a separate PR.

xedin:
@swift-ci please test

bnbarham:
@swift-ci please test

bnbarham:
Added a test case in https://github.com/swiftlang/swift-package-manager/pull/8212

xedin:

MaxDesiatov:
@swift-ci test macos

MaxDesiatov (approved):
LGTM contingent on getting https://github.com/swiftlang/swift-package-manager/pull/8212 in.

bkhouri (changes requested):
praise: Thank you very much for addressing this regression, but would it be possible to add a test to ensure we do re-introduce this regression?

bnbarham:

bnbarham:

bnbarham:

PR #8057, opened 2024-10-20
Enable dynamic dispatch for package path manifest loading

dmhts:
This PR elevates the default [ManifestLoaderProtocol.load(packagePath...](https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Sources/PackageLoading/ManifestLoader.swift#L192-L234) implementation to a protocol requirement, enabling dynamic dispatch for types conforming to the protocol.

### Motivation:

The Workspace class provides a convenient way to fine-tune the manifest loading process by allowing the use of custom loaders. For example, [RegistryAwareManifestLoader](https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Sources/Workspace/Workspace%2BRegistry.swift#L55) already enables loading manifests from registries. Similarly, it would be beneficial for other Workspace clients to implement their own manifest loaders to address their specific needs (if you'd like, I could go into the details of our specific need, but I believe it might be irrelevant in the context of this PR).

However, one of the default manifest loading [implementations](https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Sources/PackageLoading/ManifestLoader.swift#L192-L234) used by [Workspace](https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Sources/Workspace/Workspace%2BManifests.swift#L715-L728) (and other clients as well) is not currently reflected in the protocol requirements, making it impossible for implementers to override due to the static dispatch. To address this, I propose elevating the default implementation to a protocol requirement. This would enable dynamic dispatch for this method, providing greater flexibility for Workspace clients.

### Modifications:

This PR simply copies the signature of the default method implementation in question to the protocol body, while also fixing some outdated documentation along the way. This modification introduces no interface changes, so no adaptations are required at the calling sites.

### Result:

```swift
class MyCustomManifestLoader: ManifestLoaderProtocol { 

    func load(packagePath: AbsolutePath, packageIdentity: PackageIdentity...) {
        // I'll be called now since I'm already dynamically dispatched.
    }
    
    ...

}
```

### DIFF
--- a/Sources/PackageLoading/ManifestLoader.swift
+++ b/Sources/PackageLoading/ManifestLoader.swift
@@ -99,10 +99,10 @@ extension ManifestParseError: CustomStringConvertible {
 
 /// Protocol for the manifest loader interface.
 public protocol ManifestLoaderProtocol {
-    /// Load the manifest for the package at `path`.
+    /// Load the manifest at `path`.
     ///
     /// - Parameters:
-    ///   - manifestPath: The root path of the package.
+    ///   - manifestPath: The path of the manifest file to load.
     ///   - manifestToolsVersion: The version of the tools the manifest supports.
     ///   - packageIdentity: the identity of the package
     ///   - packageKind: The kind of package the manifest is from.
@@ -112,6 +112,7 @@ public protocol ManifestLoaderProtocol {
     ///   - dependencyMapper: A helper to map dependencies.
     ///   - fileSystem: File system to load from.
     ///   - observabilityScope: Observability scope to emit diagnostics.
+    ///   - delegateQueue: The dispatch queue to perform the delegate callbacks on.
     ///   - callbackQueue: The dispatch queue to perform completion handler on.
     ///   - completion: The completion handler .
     func load(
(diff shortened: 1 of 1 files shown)


jakepetroules:
> if you'd like, I could go into the details of our specific need, but I believe it might be irrelevant in the context of this PR

It would definitely be interesting to hear about use cases!

dmhts:
@jakepetroules sure!

### Background:

We use some of SwiftPM's data models and tools to precompile our dependency package targets as XCFrameworks, which are later consumed via binary targets from automatically generated _mirrored_ packages. Specifically, we use `Workspace` to [load a package's module graph](https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Sources/Workspace/Workspace.swift#L881-L889), which we then [use in the PIFBuilder](https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Sources/XCBuildSupport/PIFBuilder.swift#L73-L89) to generate the corresponding PIF file for more direct communication with the underlying XCBuild system.

Additionally, we require that the produced XCFrameworks be compatible with modules built with different Swift versions. To achieve this, we attempt to compile them with module stability enabled by default (and hence library evolution). However, packages that do not support module stability must be precompiled in an unstable, compiler-specific module format. This requires using a specific toolchain (typically in the form of an Xcode bundle) for each supported Swift (and sometimes SDK) version.

Since we need to apply different toolchains to the same package, this can lead to the same package being resolved into different [module graphs](https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Sources/PackageGraph/ModulesGraph.swift#L88). This typically occurs when the package has version-specific manifests (where each manifest may define its own set of dependencies). This situation is especially common nowadays as many 3rd party packages are in the process of gradually migrating from Swift 5.x to Swift 6.0

### Existing issue and use cases:

Unfortunately, [the existing version-specific manifest loading mechanism](https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Sources/PackageLoading/ToolsVersionParser.swift#L612-L618) doesn't allow loading manifests compatible with arbitrary toolchains. As you can see from the logic, the default implementation first always tries to look for version-specific manifests that are compatible with the [_**hardcoded**_ SwiftPM tools version](https://github.com/swiftlang/swift-package-manager/blob/4095b908832f25c492af72728dc6687924978604/Sources/PackageModel/ToolsVersion.swift#L37-L41). I'm not sure if there was a specific reason for implementing it this way, but it limits the flexibility of the current loading logic for multi-toolchain scenarios. 

For our use case, we need a more flexible implementation where Workspace loads version-specific manifests based on the toolchain version provided dynamically, rather than relying on the hardcoded toolchain version in the currently used SwiftPM instance. I can also imagine other, less straightforward use cases where a manifest loader could generate manifests and the surrounding package structure on the fly to create adapters for non-Swift libraries, but that might be too far-fetched at this point.

### Possible solutions:

The good news is that a greater flexibility can be easily achieved by incorporating the changes from this PR. Alternatively, the goal could also be met by modifying the existing loading logic, but that would likely involve a more invasive change, requiring careful examination. In contrast, the proposed solution of enabling dynamic dispatch is simpler and non-disruptive.

dmhts:
@jakepetroules Sorry for the rather long write-up. My main point was that such a small change could enable a lot of flexibility without any visible downsides. Do you think there’s a chance to get it through?

dmhts:
Sorry to bother you. Just a friendly reminder about this PR whenever you have time. If you feel it doesn’t add value, I’m happy to close it to avoid contributing to the number of dangling PRs.

Issue #8060, opened 2024-10-21
Build tool plugin is not seeing target dependencies when there is exactly one


DougGregor:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [X] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

I have a build tool plugin that looks at the dependencies of the target it is applied to. For example:

```swift
        .target(
            name: "JavaMath",
            dependencies: [
              .product(name: "JavaKit", package: "swift-java"),
              // Note: Uncomment the following to unbreak test.swift
              // .product(name: "JavaKitJar", package: "swift-java"),
            ],
            swiftSettings: [
              .unsafeFlags(["-I\(javaIncludePath)", "-I\(javaPlatformIncludePath)"])
            ],
            plugins: [
              .plugin(name: "Java2SwiftPlugin", package: "swift-java"),
            ]
        ),
```

The Java2SwiftPlugin doesn't see the dependency of JavaMath on JavaKit. However, if I uncomment the second line (to add the JavaKitJar dependency), the Java2SwiftPlugin sees both the JavaKit and JavaKitJar dependencies correctly. It seems to only ever happen when there is exactly one dependency, it is missing.

### Expected behavior

The plugin should see the target dependency even when there is only one.

### Actual behavior

The dependency is missing when there is a single dependency; all dependencies are visible if there is more than one.

### Steps to reproduce

1. `git clone https://github.com/DougGregor/swift-java-deps`
2. `cd swift-java-deps`
3. `swift build`

The build will fail. If you uncomment the marked line in `Package.swift`, the build will succeed because the Java2SwiftPlugin is correctly seeing the dependency.

### Swift Package Manager version/commit hash

6.0

### Swift & OS version (output of `swift --version ; uname -a`)

```
swift-driver version: 1.115 Apple Swift version 6.0 (swiftlang-6.0.0.9.10 clang-1600.0.26.2)
Target: arm64-apple-macosx15.0
```

dschaefer2:
@DougGregor just to confirm, were you getting errors like:

`.../swift-java-deps/.build/plugins/outputs/swift-java-deps/JavaMath/destination/Java2SwiftPlugin/generated/RoundingMode.swift:20:13: error: value of type 'RoundingMode' has no member 'equals'`

[...]

PR #8083, opened 2024-10-29
Ensure plug-in C modules don't mix with target ones.

dschaefer2:
When a Swift module depends on a C library that is also a dependent of plug-in that the module depends on, we were seeing includes of the module map for both versions of that library, host and target, causing types to be redefined.

This change filters out the plug-in version of these dependencies.

### DIFF
--- a/Sources/Build/BuildPlan/BuildPlan+Swift.swift
+++ b/Sources/Build/BuildPlan/BuildPlan+Swift.swift
@@ -20,16 +20,13 @@ extension BuildPlan {
     func plan(swiftTarget: SwiftModuleBuildDescription) throws {
         // We need to iterate recursive dependencies because Swift compiler needs to see all the targets a target
         // depends on.
-        for case .module(let dependency, let description) in swiftTarget.recursiveDependencies(using: self) {
+        for case .module(let dependency, let description) in swiftTarget.recursiveLinkDependencies(using: self) {
             switch dependency.underlying {
             case let underlyingTarget as ClangModule where underlyingTarget.type == .library:
                 guard case let .clang(target)? = description else {
                     throw InternalError("unexpected clang target \(underlyingTarget)")
                 }
-                // Only if the targets are for the same destination, i.e. not a plugin for a target target
-                guard target.buildParameters.destination == swiftTarget.buildParameters.destination else {
-                    break
-                }
+
                 // Add the path to modulemap of the dependency. Currently we require that all Clang targets have a
                 // modulemap but we may want to remove that requirement since it is valid for a target to exist without
                 // one. However, in that case it will not be importable in Swift targets. We may want to emit a warning
--- a/Sources/Build/BuildPlan/BuildPlan+Swift.swift
+++ b/Sources/Build/BuildPlan/BuildPlan+Swift.swift
@@ -26,6 +26,10 @@ extension BuildPlan {
                 guard case let .clang(target)? = description else {
                     throw InternalError("unexpected clang target \(underlyingTarget)")
                 }
+                // Only if the targets are for the same destination, i.e. not a plugin for a target target
+                guard target.buildParameters.destination == swiftTarget.buildParameters.destination else {
+                    break
+                }
                 // Add the path to modulemap of the dependency. Currently we require that all Clang targets have a
                 // modulemap but we may want to remove that requirement since it is valid for a target to exist without
                 // one. However, in that case it will not be importable in Swift targets. We may want to emit a warning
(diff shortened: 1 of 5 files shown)


dschaefer2:
Still have tests to write.

Sources/Build/BuildPlan/BuildPlan+Swift.swift
```diff
@@ -26,6 +26,10 @@ extension BuildPlan {
                 guard case let .clang(target)? = description else {
                     throw InternalError("unexpected clang target \(underlyingTarget)")
                 }
+                // Only if the targets are for the same destination, i.e. not a plugin for a target target
+                guard target.buildParameters.destination == swiftTarget.buildParameters.destination else {
+                    break
+                }
```
MaxDesiatov, line 32: formatting nit: IMO more readable with visual separation, and makes it easier to navigate with `{`/`}` when using Vi keybindings 😉 

```suggestion
                }

                // Only if the targets are for the same destination, i.e. not a plugin for a target target
                guard target.buildParameters.destination == swiftTarget.buildParameters.destination else {
                    break
                }

```

MaxDesiatov (commented):
> we were seeing includes of the module map for both versions of that library, host and target, causing types to be redefined.

Would you mind adding a test case that reproduce that situation?

dschaefer2:
@swift-ci please test

dschaefer2:
Updated to push this all the way to the dependency walking so we never get plug-in rooted dependencies in this part of the planning. And added a test.

Also missed adding the issue link #8060 which points at the @DougGregor's repo where I first reproduced this.

Tests/BuildTests/BuildPlanTests.swift
```diff
@@ -6736,4 +6736,68 @@ final class BuildPlanTests: XCTestCase {
             )
         }
     }
+
+    func testHostTargetDontCross() async throws {
```
MaxDesiatov, line 6740: naming nit: `Cross` could refer to cross compilation here, but I thought we're actually referring to overlapping builds?

```suggestion
    func testHostTargetDontOverlap() async throws {
```

dschaefer2: It's referring to crossing the streams. Been watching too much Ghostbusters lately :p.

dschaefer2: Renamed to better reflect.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

Sources/Build/BuildPlan/BuildPlan+Swift.swift
```diff
@@ -20,12 +20,13 @@ extension BuildPlan {
     func plan(swiftTarget: SwiftModuleBuildDescription) throws {
         // We need to iterate recursive dependencies because Swift compiler needs to see all the targets a target
         // depends on.
-        for case .module(let dependency, let description) in swiftTarget.recursiveDependencies(using: self) {
+        for case .module(let dependency, let description) in swiftTarget.recursiveLinkDependencies(using: self) {
```
xedin, line 23: I think instead of a special method that is filtering out plugins what you want to do is check whether the dependency targets the same destination and if not skip it, that would filter out plugins/macros and other things for "target" destination.

dschaefer2: I did that originally but felt that was hiding the real issue of including plug-in dependencies in my list of recursive dependencies when I don't want them when constructing the command line arguments for my module build.

xedin: Maybe add a optional destination as a parameter to recursive dependencies method then, it would be easier to understand than having to navigate multiple slightly different methods.

xedin: Or maybe it would be okay to adjust `recursiveDependencies` to always respect destination of the requesting module I don't think we ever want to have things from different destinations used anyway, it's not only plugins though it's also applicable to macros.

dschaefer2: I've been hesitant to rely on destinations. Could a plug-in tool depend on another plug-in tool? Wouldn't both be host?

dschaefer2:
@swift-ci please test

Sources/Build/BuildPlan/BuildPlan+Swift.swift
xedin: Yes, both would be host, just like macros. There are other places in Build which filter based on the destination so it won’t be unique.

dschaefer2:
@swift-ci please test windows

Sources/Build/BuildPlan/BuildPlan+Swift.swift
xedin: Btw, note that plugins actually have a special description not `SwiftModuleBuildDescription`.

dschaefer2:
Fixed this for real https://github.com/swiftlang/swift-package-manager/pull/8524

PR #8087, opened 2024-10-31
Use async `HTTPClient` in `BinaryArtifactsManager`

MaxDesiatov:
### Motivation:

To adopt Swift concurrency in `ManifestLoader`, we need to convert `RegistryClient` first, which then would require use of `HTTPClient` instead of `LegacyHTTPClient`. That would also cascade to `BinaryArtifactsManager`, so I'm starting there first to make PRs smaller.

### Modifications:

Converted (and as a consequence, simplified) code in `Workspace+BinaryArtifacts.swift`, also adopted `HTTPClient` in `WorkspaceTests` to fix tests after the change.

Converted some error diagnostics to conform to `Error`, which makes control flow with `ObservabilityScope.trap` wrapper easier to write and read.

### Result:

One major previous adopter of `LegacyHTTPClient` is now compatible with Swift concurrency.

Since `BinaryArtifactsManager` was blocking Swift concurrency thread pool and also used `.sharedConcurrent` dispatch queue, this potentially fixes thread explosion/deadlocks that may have been caused by previously broken code.

### DIFF
--- a/Sources/Basics/Concurrency/ThreadSafeBox.swift
+++ b/Sources/Basics/Concurrency/ThreadSafeBox.swift
@@ -30,6 +30,12 @@ public final class ThreadSafeBox<Value> {
             self.underlying = value
         }
     }
+    
+    public func mutate(body: (inout Value?) throws -> ()) rethrows {
+        try self.lock.withLock {
+            try body(&self.underlying)
+        }
+    }
 
     @discardableResult
     public func memoize(body: () throws -> Value) rethrows -> Value {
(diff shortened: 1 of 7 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test linux

bnbarham:
@swift-ci test linux

bnbarham:
@swift-ci test linux

bnbarham:
@swift-ci test linux

plemarquand (approved):

PR #8099, opened 2024-11-05
Test: Add platform helpers

bkhouri:
Add helpers related to the current platform to assist with conditionally running test cases.

### Motivation:

Some XCTests have the following implementation, or a variation thereof

```
#if os(macOS)
    func testLibraryEnvironmentVariable() async throws {
       //test implementation
    }
#endif
```

The idiomatic Swift Testing is to use the `enabled`/`disabled` traits, thus allowing the test to be compiled.  As such, create some helpers function. that will allow enabling/disabling a test using these helper functions.

### Modifications:

Add helper functions and add some Swift Testing tests.

### Result:

Ran the equivalent of the following and ensured there were no test-related failures 

```
for _ in $(seq 0 100);
do
    swift test --enable-swift-testing --disable-xctest
done
```

Blocked by #8137 
Requires https://github.com/swiftlang/swift/pull/78300

### DIFF
--- a/Sources/_InternalTestSupport/PlatformHelpers.swift
+++ b/Sources/_InternalTestSupport/PlatformHelpers.swift
@@ -1,0 +1,53 @@
+
+
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+import Basics
+
+import Testing
+
+public func isWindows() -> Bool {
+    #if os(Windows)
+    return true
+    #else
+    return false
+    #endif
+}
+
+public func isLinux() -> Bool {
+    #if os(Linux)
+    return true
+    #else
+    return false
+    #endif
+}
+
+public func isMacOS() -> Bool {
+    #if os(macOS)
+    return true
+    #else
+    return false
+    #endif
+}
+
+public func isRealSigningIdentityTestEnabled() -> Bool {
+    #if ENABLE_REAL_SIGNING_IDENTITY_TEST
+    return true
+    #else
+    return false
+    #endif
+}
+
+public func isEnvironmentVariableSet(_ variableName: EnvironmentKey) -> Bool {
+    guard let value = Environment.current[variableName] else { return false }
+    return !value.isEmpty
+}
--- a/Tests/_InternalTestSupportTests/PlatformHelpersTests.swift
+++ b/Tests/_InternalTestSupportTests/PlatformHelpersTests.swift
@@ -1,0 +1,35 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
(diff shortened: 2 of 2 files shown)


bkhouri:
Ready for review, though it may be blocked as not all pipeline builds run with Swift 6.0!

Sources/_InternalTestSupport/PlatformHelpers.swift
```diff
@@ -0,0 +1,33 @@
...
+public func isRealSigningIdentityTestEnabled() -> Bool {
+    #if ENABLE_REAL_SIGNING_IDENTITY_TEST
+    return true
+    #else
+    return false
+    #endif
+}
+
+public func isEnvironmentVariableSet(_ variableName: EnvironmentKey) -> Bool {
+    guard let value = Environment.current[variableName] else {return false}
```
plemarquand, line 31: nit: run the formatter

Sources/_InternalTestSupport/PlatformHelpers.swift
```diff
@@ -0,0 +1,33 @@
...
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+import Basics
+
+public func isMacOs() -> Bool {
```
plemarquand, line 14: If these are for swift-testing tests it may make more sense to implement these as test traits

Sources/_InternalTestSupport/PlatformHelpers.swift
bkhouri: Although I understand the desire to have consistent formatting, it is not currently enforced via the pipelines.  Having the tools enforce "formatting" and "linting" is the correct solution, in my opinion.  Until such time, I will do my best to run the formatter, however it will cause significant changes for the sources we touch as not all lines of code adhere to the formatter settings.

Sources/_InternalTestSupport/PlatformHelpers.swift
plemarquand: I think here is Ok; TSC is deprecated and there aren't plans to expand its use.

grynspan, line 14:
```suggestion
public func isMacOS() -> Bool {
```

bkhouri:
This PR is blocked by #8137

bkhouri:
Please test with following pull request:
https://github.com/swiftlang/swift/pull/78300

@swift-ci Please test

grynspan:
Be aware: this won't work reliably for tests where the test content is only compilable on one platform or another.

bkhouri:
> Be aware: this won't work reliably for tests where the test content is only compilable on one platform or another.

@grynspan : Could you please clarify `this`?

Tests/_InternalTestSupportTests/PlatformHelpersTests.swift
```diff
@@ -0,0 +1,35 @@
...
+import _InternalTestSupport
+import Basics
+import Testing
+
+struct testisEnvironmentVariableSet {
+    @Test(
+        arguments: [
+            (name: "", expected: false),
+            (name: "DOES_NOT_EXIST", expected: false),
+            (name: "HOME", expected: true)
```
bkhouri, line 22: thought: this will likely needs to be revisited when addressing #8121 and #8176

bkhouri:
@swift-ci please test

bkhouri:
@swift-ci please test

bkhouri:
@swift-ci please test windows

bkhouri:
@swift-ci please test windows

bkhouri:
This PR will not be pursued.  Closing.

PR #8106, opened 2024-11-07
[DNM] Support safe.bareRepository=explicit for .gitconfig

### LINKED ISSUE #8068: Git config setting `safe.bareRepository=explicit` can break dependency resolution

- [X] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

A recent software update from Sourcetree added or modified the `safe.bareRepository=explicit` configuration option to git.

This interfers with the package resolution, it seems to only involve our private packages on github.

The packages refuse to resolve and we get error like saying it cannot fetch revision (see below)

### Expected behavior

```
% swift package resolve
Fetching git@github.com:<MyOrg>/<MyDependencyPackage>.git
Fetched git@github.com:<MyOrg>/<MyDependencyPackage>.git from cache (5.32s)
Creating working copy for git@github.com:<MyOrg>/<MyDependencyPackage>.git
Working copy of git@github.com:<MyOrg>/<MyDependencyPackage>.git resolved at <BranchName> (f96e2ec)
```

### Actual behavior

```
% swift package resolve
Fetching git@github.com:<MyOrg>/<MyDependencyPackage>.git
Fetched git@github.com:<MyOrg>/<MyDependencyPackage>.git from cache (5.32s)
error: Couldn’t get revision ‘<A Commit Hash>^{commit}’:
    fatal: cannot use bare repository '/Users/<myuserdir>/Repos/Fixes/<MyPackage>/.build/repositories/<MyDependencyPackage>-ad20a496' (safe.bareRepository is 'explicit') in git@github.com:<MyOrg>/<MyDependencyPackage>.git
```

### Steps to reproduce

A Package with a dependency to a private github repo
In the CLI
1. git config --global --replace safe.bareRepository explicit
2. swift package resolve

Swift Package Manager - Swift 6.0.0-dev

```
swift-driver version: 1.115 Apple Swift version 6.0 (swiftlang-6.0.0.9.10 clang-1600.0.26.2)
Target: arm64-apple-macosx15.0
Darwin rlee-MacBook-Pro 24.0.0 Darwin Kernel Version 24.0.0: Tue Sep 24 23:37:36 PDT 2024; root:xnu-11215.1.12~1/RELEASE_ARM64_T6020 arm64
```

Kyle-Ye:
Close #8068 

- **Add safe.bareRepository=explicit support**

### Motivation:

Git has add `safe.bareRepository=explicit` to opt-in mitigations for buried bare repos attack since `2.38.0`.

Also some Git GUI app(SourceTree) is enforcing the new rule which breaking the building system including SwiftPM here.

See https://github.com/justinsteven/advisories/blob/main/2022_git_buried_bare_repos_and_fsmonitor_various_abuses.md
 for detail.

### Modifications:

Add `--git-dir` explicitly for bareRepo.

> https://fossies.org/linux/git/Documentation/config/safe.txt

### Result:

When `safe.bareRepository=explicit` is enabled globally on .gitconfig. SwiftPM resolve dependency successfully.

### Steps to verify

[SPMBareTest.zip](https://github.com/user-attachments/files/17658832/SPMBareTest.zip)

1. Run `git config set safe.bareRepository explicit --global`.
2. Download `SPMBareTest.zip` and unzip it.
3. Run `cd SPMBareTest && swift package resolve`.

### Context

Other package manager system tracking the issue:

- https://github.com/composer/composer/issues/11855
- https://github.com/rubygems/rubygems/issues/7515

### DIFF
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -204,13 +204,25 @@ public struct GitRepositoryProvider: RepositoryProvider, Cancellable {
     }
 
     public func isValidDirectory(_ directory: Basics.AbsolutePath) throws -> Bool {
-        let result = try self.git.run(["-C", directory.pathString, "rev-parse", "--git-dir"])
-        return result == ".git" || result == "." || result == directory.pathString
+        do {
+            let result = try self.git.run(["-C", directory.pathString, "rev-parse", "--git-dir"])
+            return result == ".git" || result == "." || result == directory.pathString
+        } catch let error as GitShellError {
+            guard try error.result.utf8stderrOutput().contains("safe.bareRepository is 'explicit'") else { return false }
+            let result = try self.git.run(["-C", directory.pathString, "--git-dir", directory.pathString, "rev-parse", "--git-dir"])
+            return result == directory.pathString
+        }
     }
 
     public func isValidDirectory(_ directory: Basics.AbsolutePath, for repository: RepositorySpecifier) throws -> Bool {
-        let remoteURL = try self.git.run(["-C", directory.pathString, "config", "--get", "remote.origin.url"])
-        return CanonicalPackageURL(remoteURL) == CanonicalPackageURL(repository.url)
+        do {
+            let remoteURL = try self.git.run(["-C", directory.pathString, "config", "--get", "remote.origin.url"])
+            return CanonicalPackageURL(remoteURL) == CanonicalPackageURL(repository.url)
+        } catch let error as GitShellError {
+            guard try error.result.utf8stderrOutput().contains("safe.bareRepository is 'explicit'") else { return false }
+            let remoteURL = try self.git.run(["-C", directory.pathString,  "--git-dir", directory.pathString, "config", "--get", "remote.origin.url"])
+            return CanonicalPackageURL(remoteURL) == CanonicalPackageURL(repository.url)
+        }
     }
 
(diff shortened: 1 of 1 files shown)


Kyle-Ye:
The PR is only a POC for adding the support. And I think we need to figure out the following issue before merging.

## Technic challenge:

When user has opt-in this security feature, we need to add `--git-dir` or set `GIT_DIR` for every call `git` call happening on a bare repository.

### Set GIT_DIR env for git command

If we set it case by case, it is the same as adding `--git-dir` to every git command.

So we are talking about set it "globally" for the whole repo.

One way is modifying Git.environmentBlock when a repo start git operation and restore it when it ends.

The problem is I don't know if we have concurrent git operation for multi repo/dependency. And the documentation states that "modification is not thread safe, designed for testing only"

### Pass --git-dir to every git command

Generally speaking, most git operation will call `GitRepository/callGit`. So we can add the check to add `["--git-dir", self.path.pathString]` if the repo is a bare one.

But the problem is the way we detetmine if a repo is a bare one is through a getter function which will also call ``GitRepository/callGit`.

Even we do not call `callGit` in `isBare()` function, the problem still exists. Because `git rev-parse --is-bare-repository` will also fail if it is a bare one.

## Other tasks/todo:

- [ ] Add test case for covering the changes

Issue #8090, opened 2024-11-01
Values containing hash (#) symbols are not included fully


garyodonoghue:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [X] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

if i use a sample .netrc file as below:
```
machine blah.com
login test_user
password first#second
```

the # symbol in the password causes it to be not read entirely, it gets processed as just 'first', with the remaining characters "second" treated as a comment. 

I believe this comes from [this function](https://github.com/swiftlang/swift-package-manager/blob/0b569a157ed14044ffd86c20304d63dcff50c61c/Sources/Basics/Netrc.swift#L128)

I understand comments need to be handled, but i think I would expect that if the password is all one 'word', i.e. no spaces, that it should still be treated as part of the password

### Expected behavior

Values in .netrc containing hash symbols are processed as they are, including hash symbols, i.e. in the above example, the password value of "first#second" is used

### Actual behavior

Value are truncated at the point of the location of the hash symbol due to the following characters being treated as comments, i.e. in the above example, the password value of "first" is used

### Steps to reproduce

Have a .netrc file containing the following:

```
machine blah.com
login test_user
password first#second
```

Capture "Authorisation" header used in request from SPM to fetch binary target 
Notice that it has the password truncated, so be "first", and does not include the entire value "first#second"

Sample unit test added to `NetrcTests.swift` to reproduce it :

```
func testHashSymbolInPassword() throws {
        let content = """
                    machine example.com
                    login anonymous
                    password qw#erty
                    """

        let netrc = try NetrcParser.parse(content)

        let machine = netrc.machines.first
        XCTAssertEqual(machine?.name, "example.com")
        XCTAssertEqual(machine?.login, "anonymous")
        XCTAssertEqual(machine?.password, "qw#erty")
    }
```

See output below: 

<img width="1002" alt="Screenshot 2024-11-01 at 16 47 09" src="https://github.com/user-attachments/assets/239d5b9b-2d08-4f30-907b-8e18e2f24073">

### Swift Package Manager version/commit hash

main (0b569a157)

### Swift & OS version (output of `swift --version ; uname -a`)

[...]

PR #8093, opened 2024-11-01
Tests: Convert some tests BasicsTests to Swift Testing

bkhouri:
Convert some BasicsTests from XCTest to Swift Testing to make use of parallelism and, in some cases, test parameterization

### Motivation:

The XCTest run, by default, sequentially. Convert most of the WorkspaceTests from XCTests to Swift Testing to make use of parallelism, and in some cases, parameterized test cases.

Not all Test Suites in BasicsTests have been converted as some use helpers in swift-tools-core-support, which don't have a matching swift testing helper.

### Modifications:

Convert XCTests to Swift Testing

### Result:

Ran the equivalent of
```
for _ in $(seq 0 100); 
do
    swift test --enable-swift-testing --disable-xctest
done
```
and ensured there were no test-related  failures.

Blocked by #8137 
Requires https://github.com/swiftlang/swift/pull/78300
Might conflict with #8210

PR #8109, opened 2024-11-08
Handle hash characters (#) in netrc fields

plemarquand:
Hash symbols were always treated as the beginning of a comment even if they were in the middle of a username or password. To address this require some whitespace before a hash character before a comment is parsed.

That patch also implements support for quoted fields so that passwords can contain the sequence "foo #bar" without dropping "bar" as a comment.

Issue: #8090

### DIFF
--- a/Tests/BasicsTests/NetrcTests.swift
+++ b/Tests/BasicsTests/NetrcTests.swift
@@ -431,5 +431,65 @@ class NetrcTests: XCTestCase {
         XCTAssertEqual(netrc.machines[1].login, "fred")
         XCTAssertEqual(netrc.machines[1].password, "sunshine4ever")
     }
-}
 
+    func testComments() throws {
+        let content = """
+            # A comment at the beginning of the line
+            machine example.com # Another comment
+            login anonymous
+            password qw#erty
+            """
+
+        let netrc = try NetrcParser.parse(content)
+
+        let machine = netrc.machines.first
+        XCTAssertEqual(machine?.name, "example.com")
+        XCTAssertEqual(machine?.login, "anonymous")
+        XCTAssertEqual(machine?.password, "qw#erty")
+    }
+
+    func testHashSymbolInPassword() throws {
+        let content = """
+            machine example.com
+            login anonymous
+            password qw#erty
+            """
+
+        let netrc = try NetrcParser.parse(content)
+
+        let machine = netrc.machines.first
+        XCTAssertEqual(machine?.name, "example.com")
+        XCTAssertEqual(machine?.login, "anonymous")
+        XCTAssertEqual(machine?.password, "qw#erty")
+    }
+
+    func testQuotedPasswordWithSpace() throws {
+        let content = """
+            machine example.com
+            login anonymous
+            password "qw erty"
+            """
+
+        let netrc = try NetrcParser.parse(content)
+
+        let machine = netrc.machines.first
+        XCTAssertEqual(machine?.name, "example.com")
+        XCTAssertEqual(machine?.login, "anonymous")
+        XCTAssertEqual(machine?.password, "qw erty")
+    }
+
+    func testQuotedPasswordWithSpaceAndHash() throws {
+        let content = """
+            machine example.com
+            login anonymous
+            password "qw #erty"
+            """
+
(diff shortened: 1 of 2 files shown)


plemarquand:
@swift-ci test

dschaefer2:
@swift-ci please test Mac

Tests/BasicsTests/NetrcTests.swift
```diff
@@ -431,5 +431,65 @@ class NetrcTests: XCTestCase {
...
+        XCTAssertEqual(machine?.name, "example.com")
+        XCTAssertEqual(machine?.login, "anonymous")
+        XCTAssertEqual(machine?.password, "qw#erty")
+    }
+
+    func testQuotedPasswordWithSpace() throws {
+        let content = """
+            machine example.com
+            login anonymous
+            password "qw erty"
```
bkhouri, line 470: Suggestion: in order to cover all the scenarios, can we add a flavour of that appends a comment at the end of this line

Ie
`password "we #erty" #a comment`

bkhouri (changes requested):

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

jakepetroules:
Is there a reference implementation or standard detailing the specific format so we can ensure we're compliant with that?

plemarquand:
@jakepetroules frustratingly, no. I went looking before picking this up and there doesn't seem to be a defined spec. 

It seems like the [reference implementation](https://linux.die.net/man/5/netrc) is generally accepted to be the [Linux ftp](https://linux.die.net/man/1/ftp) client, however none of their documentation mentions comments.

The [Python netrc library](https://docs.python.org/3/library/netrc.html) does seem to support comments, [but only if they're on their own line](https://github.com/python/cpython/blob/3.13/Lib/netrc.py#L125).

Because our implementation already supports comments both on their own line and trailing declarations I think modifying our implementation to match Python's or Linux's would break the parsing of user's existing .netrc files that we successfully parse today.

plemarquand:
@swift-ci test macOS

plemarquand:
@swift-ci test macOS

Tests/BasicsTests/NetrcTests.swift
```diff
@@ -431,5 +431,80 @@ class NetrcTests: XCTestCase {
...
+
+        let netrc = try NetrcParser.parse(content)
+
+        let machine = netrc.machines.first
+        XCTAssertEqual(machine?.name, "example.com")
+        XCTAssertEqual(machine?.login, "anonymous")
+        XCTAssertEqual(machine?.password, "qw #erty")
+    }
+
+    func testQuotedPasswordWithSpaceAndHashAndCommentAtEndOfLine() throws {
```
bkhouri, line 496: suggestion (blocking): the more I read about .netrc, the more I see test gaps :S. can we add a test case sphere the machine, login and password are all on a single line,  them add the same with comments.

Also (non-blocking), is there value is describing the behaviour of the parser should a `#` character appear in the value of `login` and/or `machine`?   This will probably never happen, but maybe we should document the parsers expected behaviour.

plemarquand: I've reworked the tests to cover these cases, and made a note to convert them to parameterized tests when we switch to swift-testing.

bkhouri: This PR may conflict with #8093.  So depending on which is merge first, the other will need to be updated to be Swift Testing compatible :)

bkhouri (approved):

Tests/BasicsTests/NetrcTests.swift
bkhouri: That should be fixed now: https://ci.swift.org/job/swift-package-manager-with-xcode-self-hosted-PR-osx/4511/execution/node/49/log/

plemarquand:
@swift-ci test

plemarquand:
@swift-ci please test macOS self hosted

plemarquand:
@swift-ci please test Windows

plemarquand:
@swift-ci please test macOS self hosted

PR #8117, opened 2024-11-14
Attempt two at solving Windows symbol linking issues

dschaefer2:
Instead of creating two modules, one that exports symbols and one that doesn't and letting consuming products pick which one they want, we now mark all targets on Windows as static and then reset that if a target is in a DLL product. This gets by the missing feature of marking a target as a DLL (only products can be DLLs).

We also turn off the exporting imported symbol warning that is generated when a module in a DLL is also used in a static context confusing the linker.


dschaefer2:
@swift-ci please test

jakepetroules (approved):

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

Sources/Build/BuildDescription/ProductBuildDescription.swift
```diff
@@ -193,6 +193,12 @@ public final class ProductBuildDescription: SPMBuildCore.ProductBuildDescription
         // radar://112671586 supress unnecessary warnings
         if triple.isMacOSX {
             args += ["-Xlinker", "-no_warn_duplicate_libraries"]
+        } else if triple.isWindows() {
+            // locally defined symbol imported
+            // Occurs when an object file exports a symbol is imported
+            // into another module in the same executable.
+            // FIXME: Need to support targets as DLLs
+            args += ["-Xlinker", "/ignore:4217"]
```
compnerd, line 201: Please do not do this - it is hiding a real issue.

dschaefer2: The ones I'm seeing aren't actionable by the user. They're caused by our lack of support for dynamic targets. There's so many you miss any other warning that might be actionable.

If a target appears in a DLL, it doesn't get the -static arg. But then if you include it in a EXE, like tests, you get the warning.

dschaefer2: @compnerd brings up a good point about hiding user caused warnings. I'll take this out for now and see what the fallout is. And we'll continue to work to make sure SwiftPM isn't causing these warnings.

dschaefer2:
@swift-ci enable tests

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

PR #8135, opened 2024-11-22
Bump the Swift version to 6.2

shahmishal:
https://github.com/swiftlang/swift/pull/77799

### DIFF
--- a/Sources/Basics/SwiftVersion.swift
+++ b/Sources/Basics/SwiftVersion.swift
@@ -58,7 +58,7 @@ public struct SwiftVersion: Sendable {
 extension SwiftVersion {
     /// The current version of the package manager.
     public static let current = SwiftVersion(
-        version: (6, 1, 0),
+        version: (6, 2, 0),
         isDevelopment: true,
         buildIdentifier: getBuildIdentifier()
     )


shahmishal:
@swift-ci test

jakepetroules (approved):

shahmishal:
When we merge this change, we need to make sure to revert it from `release/6.1` branch.

Sources/Basics/SwiftVersion.swift
```diff
@@ -58,7 +58,7 @@ public struct SwiftVersion: Sendable {
 extension SwiftVersion {
     /// The current version of the package manager.
     public static let current = SwiftVersion(
-        version: (6, 1, 0),
```
bkhouri, line 61: It's great we are increasing the Swift Version, however we should also add a `6.1` entry in the `ToolsVersion` and replace any referenced of tools version `999.0` with `6.1`.  I'm happy to push a commit to the branch with the appropriate changes! 

Then, once this is merged, we would only need to ensure the version is updated back to 6.1 in the `release/6.1` branch!

shahmishal: > I'm happy to push a commit to the branch with the appropriate changes!


Once this change is merged, let's create a new PR so it can be discussed separately.

bkhouri: Reverted in #8173 8173

PR #8138, opened 2024-11-22
Return compiler arguments for invalid package manifests

ahoppen:
Currently, when there‘s a syntax error in a package manifest, we don’t get any build settings from it in SourceKit-LSP and thus loose almost all semantic functionality. If we can’t parse the package manifest, fall back to providing build settings by assuming it has the current Swift tools version.

Fixes swiftlang/sourcekit-lsp#1704
rdar://136423767

### DIFF
--- a/Sources/Workspace/Workspace+Manifests.swift
+++ b/Sources/Workspace/Workspace+Manifests.swift
@@ -384,33 +384,23 @@ extension Workspace {
     }
 
     /// Returns manifest interpreter flags for a package.
-    // TODO: should this be throwing instead?
-    public func interpreterFlags(for packagePath: AbsolutePath) -> [String] {
-        do {
-            guard let manifestLoader = self.manifestLoader as? ManifestLoader else {
-                throw StringError("unexpected manifest loader kind")
-            }
-
-            let manifestPath = try ManifestLoader.findManifest(
-                packagePath: packagePath,
-                fileSystem: self.fileSystem,
-                currentToolsVersion: self.currentToolsVersion
-            )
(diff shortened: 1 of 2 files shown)


ahoppen:
https://github.com/swiftlang/sourcekit-lsp/pull/1846

@swift-ci Please test

Sources/Workspace/Workspace+Manifests.swift
```diff
@@ -384,33 +384,23 @@ extension Workspace {
...
-            }
-            return manifestLoader.interpreterFlags(for: manifestToolsVersion)
-        } catch {
-            // We ignore all failures here and return empty array.
-            return []
+
+        guard self.currentToolsVersion >= manifestToolsVersion || SwiftVersion.current.isDevelopment,
+                manifestToolsVersion >= ToolsVersion.minimumRequired
+        else {
+            throw StringError("invalid tools version")
```
plemarquand, line 401: nit: can you include the invalid tools version in the string to make the issue easier to debug?

plemarquand (approved):

Sources/Workspace/Workspace+Manifests.swift
MaxDesiatov: Does this error path also have corresponding test coverage here?

ahoppen: I just shuffled code around and I don’t know how I would trigger this error scenario.

ahoppen: Addressed in https://github.com/swiftlang/swift-package-manager/pull/8172

PR #8139, opened 2024-11-25
Disable development mode

bkhouri:
Disable development mode in the release 6.1 branch.

### Motivation:

We should never release Swift Package Manager in development mode. So this disables dev mode and brings in the functionality of `999.0` over being introduced specific swift tool version (primarily in tests)

### Modifications:

Modify `SwiftVersion` `isDevelopment` to `false` and update references to swift tool version `999.0` to  a specific version.

### Result:

`swift test`

### DIFF
--- a/Sources/Basics/SwiftVersion.swift
+++ b/Sources/Basics/SwiftVersion.swift
@@ -59,7 +59,7 @@ extension SwiftVersion {
     /// The current version of the package manager.
     public static let current = SwiftVersion(
         version: (6, 1, 0),
-        isDevelopment: true,
+        isDevelopment: false,
         buildIdentifier: getBuildIdentifier()
     )
 }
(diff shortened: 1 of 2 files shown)


dschaefer2:
This will not get approved until we are out of development phase or the Traits proposal gets approved. It is correctly marked 999.0 since it's a development only feature of the manifest.

bkhouri:
@swift-ci please test

dschaefer2:
I think it makes sense to wait for the auto merger to be turned off before doing this in case we have some traits work to push in that will conflict with this. Not sure it's likely but we are also not technically out of developer mode until then either.

bkhouri:
> I think it makes sense to wait for the auto merger to be turned off before doing this in case we have some traits work to push in that will conflict with this. Not sure it's likely but we are also not technically out of developer mode until then either.

Just getting the PR ready to merge whenever we are ready to merge this

bkhouri:
@swift-ci please test

plemarquand (approved):

PR #8143, opened 2024-11-28
PackageDescription.Target.Dependency type have been changed to conform to Sendable

coffmark:
PackageDescription.Target.Dependency type have been changed to conform to Sendable.

### Motivation:

Since PackageDescription.Target.Dependency type does not conform to Sendable, we need to add `@preconcurrency` to ignore the warning from the dependency.

The following is an example where `@preconcurrency` is required.

```swift
// swift-tools-version: 6.0
@preconcurrency import PackageDescription

// ...

private extension PackageDescription.Target.Dependency {
    static let collections: Self = .product(name: "Collections", package: "swift-collections")
    static let asyncAlgorithms: Self = .product(name: "AsyncAlgorithms", package: "swift-async-algorithms")
}
```

### Modifications:

PackageDescription.Target.Dependency type and TargetDependencyCondition type have been changed to conform to Sendable.

### Result:

In the example above, `@preconcurrency` is not required.

### DIFF
--- a/Sources/PackageDescription/Target.swift
+++ b/Sources/PackageDescription/Target.swift
@@ -42,7 +42,7 @@ public final class Target {
     }
 
     /// The different types of a target's dependency on another entity.
-    public enum Dependency {
+    public enum Dependency: Sendable {
         /// A dependency on a target.
         ///
         ///  - Parameters:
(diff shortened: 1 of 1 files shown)


plemarquand:
@swift-ci test

xedin (approved):

PR #8148, opened 2024-11-30
Android: switch to the new swift-toolchain-sqlite package instead of requiring an external dependency

finagolfin:
I've been building with this patch both natively on Android and [on my Android CI](https://github.com/finagolfin/swift-android-sdk/blob/3493a85cd1ae767792ba01c8634e3d9456d19901/swift-android-ci-except-release.patch#L34), confirming that it [got rid of the external dependency on `libsqlite3.so`](https://github.com/finagolfin/swift-android-sdk/actions/runs/12095721068/job/33729218006#step:8:20591) and works well.

@jakepetroules, along with your recent llbuild pull, this gets rid of the external SQLite3 dependency for the Android toolchain except for one llbuild executable, `swift-build-tool`, whose build is configured by CMake and appears to be unused, so we can clean that up last.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -76,7 +76,10 @@ let includeDynamicLibrary: Bool = false
 let systemSQLitePkgConfig: String? = nil
 #else
 let includeDynamicLibrary: Bool = true
-let systemSQLitePkgConfig: String? = "sqlite3"
+var systemSQLitePkgConfig: String? = "sqlite3"
+if ProcessInfo.processInfo.environment["SWIFTCI_INSTALL_RPATH_OS"] == "android" {
+    systemSQLitePkgConfig = nil
+}
 #endif
 
 /** An array of products which have two versions listed: one dynamically linked, the other with the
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -76,7 +76,10 @@ let includeDynamicLibrary: Bool = false
 let systemSQLitePkgConfig: String? = nil
 #else
 let includeDynamicLibrary: Bool = true
-let systemSQLitePkgConfig: String? = "sqlite3"
+var systemSQLitePkgConfig: String? = "sqlite3"
+if ProcessInfo.processInfo.environment["SWIFTCI_INSTALL_RPATH_OS"] == "android" {
```
finagolfin, line 80: This is [only set by the official `build-script`](https://github.com/swiftlang/swift-package-manager/blob/cb829d83d016beb6607fde42136a079c37297d90/Utilities/bootstrap#L757), so it will only work in that scenario, which is fine as most builds of this repo happen that way.

Eventually, `systemLibrary` dependencies should be conditional on the platform, but that is not available yet.

jakepetroules (approved):

finagolfin:
@MaxDesiatov, need a CI run here.

plemarquand:
@swift-ci test

finagolfin:
Passed CI as expected, ready for merge.

finagolfin:
@jakepetroules, thanks for the quick review, would you be okay with getting this small patch into 6.1 next? It would allow me to retire it from [my downstream patches](https://github.com/finagolfin/termux-packages/blob/swift-devel/packages/swift/swiftpm-sqlite-dependency.patch#L31).

dschaefer2:
For SwiftPM, main is currently auto merged to 6.1 until Dec 20'th. No manual effort necessary.

PR #8151, opened 2024-12-02
Mark all test classes fileprivate

jakepetroules:
This reduces the number of unnecessary exported symbols.


jakepetroules:
@swift-ci please smoke test

jakepetroules:
Actually, this technique doesn't work with XCTest on Windows (only Swift Testing), never mind.

dschaefer2:
The -static changes in the build manifest generator are working so far.

jakepetroules:
I consider this best practice regardless. :)

PR #8152, opened 2024-12-03
Show human-readable git errors

yyvch:
Show human-readable git errors

### Motivation:

I noticed that when swiftpm cache is empty and there's a network issue, then `swift build` throws an error which is not helpful, something like this:

```
error: GitShellError(result: <AsyncProcessResult: exit: terminated(code: 128), output:
 
>)
```

where it does say "output" but doesn't actually include output or show any details about what dependency caused the error.

### Modifications:

 - `GitShellError` now conforms to CustomStringConvertible protocol, so that when it is thrown then it will print what was the git error
 - `GitShelError` has now package access to allow catching it explicitly.

### Result:

Errors now tell details of the failure: 

```
error: Git command 'git -C /Users/yy/Library/Caches/org.swift.swiftpm/repositories/postgres-kit-e4358808 config --get remote.origin.url' failed: fatal: cannot change to '/Users/yy/Library/Caches/org.swift.swiftpm/repositories/postgres-kit-e4358808': No such file or directory
```

### DIFF
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -1168,7 +1168,8 @@ public struct GitShellError: Error, CustomStringConvertible {
         let stdout = (try? self.result.utf8Output()) ?? ""
         let stderr = (try? self.result.utf8stderrOutput()) ?? ""
         let output = (stdout + stderr).spm_chomp()
-        return output
+        let command = result.arguments.joined(separator: " ")
+        return "Git command '\(command)' failed: \(output)"
     }
 }
 
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift open source project
 //
-// Copyright (c) 2014-2020 Apple Inc. and the Swift project authors
+// Copyright (c) 2014-2024 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See http://swift.org/LICENSE.txt for license information
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -204,21 +204,13 @@ public struct GitRepositoryProvider: RepositoryProvider, Cancellable {
     }
 
     public func isValidDirectory(_ directory: Basics.AbsolutePath) throws -> Bool {
-        do {
-            let result = try self.git.run(["-C", directory.pathString, "rev-parse", "--git-dir"])
-            return result == ".git" || result == "." || result == directory.pathString
-        } catch _ as GitShellError {
-            return false
-        }
+        let result = try self.git.run(["-C", directory.pathString, "rev-parse", "--git-dir"])
+        return result == ".git" || result == "." || result == directory.pathString
     }
 
     public func isValidDirectory(_ directory: Basics.AbsolutePath, for repository: RepositorySpecifier) throws -> Bool {
-        do {
-            let remoteURL = try self.git.run(["-C", directory.pathString, "config", "--get", "remote.origin.url"])
-            return CanonicalPackageURL(remoteURL) == CanonicalPackageURL(repository.url)
-        } catch _ as GitShellError {
-            return false
-        }
+        let remoteURL = try self.git.run(["-C", directory.pathString, "config", "--get", "remote.origin.url"])
+        return CanonicalPackageURL(remoteURL) == CanonicalPackageURL(repository.url)
     }
 
     public func copy(from sourcePath: Basics.AbsolutePath, to destinationPath: Basics.AbsolutePath) throws {
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -1161,8 +1161,15 @@ extension GitFileSystemView: @unchecked Sendable {}
 
 // MARK: - Errors
 
-private struct GitShellError: Error {
+private struct GitShellError: Error, CustomStringConvertible {
     let result: AsyncProcessResult
+
+    public var description: String {
+        let stdout = (try? self.result.utf8Output()) ?? ""
+        let stderr = (try? self.result.utf8stderrOutput()) ?? ""
+        let output = (stdout + stderr).spm_chomp()
+        return output
+    }
 }
 
 private enum GitInterfaceError: Swift.Error {
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -1161,7 +1161,7 @@ extension GitFileSystemView: @unchecked Sendable {}
 
 // MARK: - Errors
 
-public struct GitShellError: Error, CustomStringConvertible {
+package struct GitShellError: Error, CustomStringConvertible {
     let result: AsyncProcessResult
 
     public var description: String {
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -1161,7 +1161,7 @@ extension GitFileSystemView: @unchecked Sendable {}
 
 // MARK: - Errors
 
-private struct GitShellError: Error, CustomStringConvertible {
+public struct GitShellError: Error, CustomStringConvertible {
     let result: AsyncProcessResult
 
     public var description: String {
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -204,13 +204,21 @@ public struct GitRepositoryProvider: RepositoryProvider, Cancellable {
     }
 
     public func isValidDirectory(_ directory: Basics.AbsolutePath) throws -> Bool {
-        let result = try self.git.run(["-C", directory.pathString, "rev-parse", "--git-dir"])
-        return result == ".git" || result == "." || result == directory.pathString
+        do {
+            let result = try self.git.run(["-C", directory.pathString, "rev-parse", "--git-dir"])
+            return result == ".git" || result == "." || result == directory.pathString
+        } catch _ as GitShellError {
+            return false
+        }
     }
 
     public func isValidDirectory(_ directory: Basics.AbsolutePath, for repository: RepositorySpecifier) throws -> Bool {
-        let remoteURL = try self.git.run(["-C", directory.pathString, "config", "--get", "remote.origin.url"])
-        return CanonicalPackageURL(remoteURL) == CanonicalPackageURL(repository.url)
+        do {
+            let remoteURL = try self.git.run(["-C", directory.pathString, "config", "--get", "remote.origin.url"])
+            return CanonicalPackageURL(remoteURL) == CanonicalPackageURL(repository.url)
+        } catch _ as GitShellError {
+            return false
+        }
     }
 
     public func copy(from sourcePath: Basics.AbsolutePath, to destinationPath: Basics.AbsolutePath) throws {
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -1168,7 +1168,7 @@ package struct GitShellError: Error, CustomStringConvertible {
         let stdout = (try? self.result.utf8Output()) ?? ""
         let stderr = (try? self.result.utf8stderrOutput()) ?? ""
         let output = (stdout + stderr).spm_chomp()
-        let command = result.arguments.joined(separator: " ")
+        let command = self.result.arguments.joined(separator: " ")
         return "Git command '\(command)' failed: \(output)"
     }
 }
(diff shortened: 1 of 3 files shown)


Sources/SourceControl/GitRepository.swift
```diff
@@ -204,13 +204,21 @@ public struct GitRepositoryProvider: RepositoryProvider, Cancellable {
...
 
     public func isValidDirectory(_ directory: Basics.AbsolutePath) throws -> Bool {
-        let result = try self.git.run(["-C", directory.pathString, "rev-parse", "--git-dir"])
-        return result == ".git" || result == "." || result == directory.pathString
+        do {
+            let result = try self.git.run(["-C", directory.pathString, "rev-parse", "--git-dir"])
+            return result == ".git" || result == "." || result == directory.pathString
+        } catch _ as GitShellError {
+            return false
+        }
```
MaxDesiatov, line 212: This means the error is now completely ignored, where exactly is its description printed then? This might also break  assumptions down the call stack about control flow, which previously relied on throwing this error.

yyvch: Right.

As I read it, there are two ways to interact with git repo:
 - low-level [GitHelper.run](https://github.com/swiftlang/swift-package-manager/blob/c237352345ff9deca691cc5c5e2e7b1cb216ae48/Sources/SourceControl/GitRepository.swift#L45) which doesn't have a context of what repo it's called for. It throws private exception [GitShellError](https://github.com/swiftlang/swift-package-manager/blob/c237352345ff9deca691cc5c5e2e7b1cb216ae48/Sources/SourceControl/GitRepository.swift#L1164).
 - high-level [GitRepository.callGit](https://github.com/swiftlang/swift-package-manager/blob/c237352345ff9deca691cc5c5e2e7b1cb216ae48/Sources/SourceControl/GitRepository.swift#L100) which does have a context of what repo it's called for and what human-readable message should be. It throws public exception [GitCloneError](https://github.com/swiftlang/swift-package-manager/blob/c237352345ff9deca691cc5c5e2e7b1cb216ae48/Sources/SourceControl/GitRepository.swift#L1200).

`isValidDirectory` uses the low-level, so it doesn't carry context of what was the repo and what should be the human-readable message. Its `GitShellError` also not caught anywhere further in the control flow. So if `isValidDirectory` raises exception, that's fatal for any Swift Package Manager operation.

For example, Swift Package Manager fails with this error if cache doesn't exist and there's no network connection:
```
error: GitShellError:     fatal: cannot change to '/Users/yy/Library/Caches/org.swift.swiftpm/repositories/postgres-kit-e4358808': No such file or directory
```

The error is accurate, but it's not helpful, because it doesn't tell what's wrong and how to fix that. Moreover, the raised exception is private, so there's no way of catching it gracefully. So changing the flow is absolutely the point.


```
error: Failed to clone repository https://github.com/vapor/postgres-kit.git:
    Cloning into bare repository '/Users/yy/work/pub/tst/testpackage/.build/repositories/postgres-kit-e4358808'...
    fatal: unable to access 'https://github.com/vapor/postgres-kit.git/': Could not resolve host: github.com
```

Where that message is coming from: for that specific use-case (offline and no pre-populated cache) there's [already code](https://github.com/swiftlang/swift-package-manager/blob/c237352345ff9deca691cc5c5e2e7b1cb216ae48/Sources/SourceControl/RepositoryManager.swift#L363  ) which checks offline, checks the cache, attempts to fetch, etc.

For the places where `isValidDirectory` is used, above is one of them,

 - no GitShellError exceptions are caught, because it's private
 - the flow would be the same for exception as well as non-valid dir

I did due diligence for "a control flow, which previously relied on throwing this error" at least within Swift Package Manager codebase, I believe it's not a huge factor for this change. I understand hesitation on changing behavior for public API, but the promise is ["returns true if the directory is valid git location"](https://github.com/swiftlang/swift-package-manager/blob/c237352345ff9deca691cc5c5e2e7b1cb216ae48/Sources/SourceControl/RepositoryManager.swift#L450), and I believe hiding private exception from public interface is within the promise.

dschaefer2: We still want it to throw the error in case there are other utilities using this API and want to display the error in their own way. All I thought this change needed to do was to make sure stderr was captured in the description which it wasn't previously and which you did fix below.

i.e. I agree with Max. :)

dschaefer2 (changes requested):

Sources/SourceControl/GitRepository.swift
yyvch: :/ oh well

Can we make the `GitShelError` exception public then? Since the expectation that somebody may want to catch it and show in their own way outside of SourceControl module?

dschaefer2: I'd be OK with that.

yyvch: - change in the public behavior of the API is reverted in 2525234
 - the error struct is now public in 301fa08
 - description now includes the git command itself 18dea2c

Sources/SourceControl/GitRepository.swift
```diff
@@ -1161,8 +1161,16 @@ extension GitFileSystemView: @unchecked Sendable {}
 
 // MARK: - Errors
 
-private struct GitShellError: Error {
+public struct GitShellError: Error, CustomStringConvertible {
```
MaxDesiatov, line 1164: I'd rather proceed cautiously and make this `package` or `@_spi(SwiftPMTestSuite)` first, unless it's shown that making it `public` is absolutely necessary.

yyvch: Package it is, [4dc1860](https://github.com/swiftlang/swift-package-manager/pull/8152/commits/4dc1860fb7f78ae3b59a4481dbb27a88af89e6a0).

plemarquand:
@swift-ci test

Tests/SourceControlTests/GitRepositoryProviderTests.swift
```diff
@@ -57,23 +57,59 @@ class GitRepositoryProviderTests: XCTestCase {
...
+            "Error string '\(errorString)' should contain '\(stdOut)'")
+        XCTAssertTrue(
+            errorString.contains(stdErr),
+            "Error string '\(errorString)' should contain '\(stdErr)'")
         XCTAssertTrue(
             errorString.contains(command),
             "Error string '\(errorString)' should contain '\(command)'")
         }
+
+    func testGitShelErrorEmptyStdOut() throws {
```
bripeticca, line 84: typo: `testGitShellErrorEmptyStdOut`

Tests/SourceControlTests/GitRepositoryProviderTests.swift
```diff
@@ -57,23 +57,59 @@ class GitRepositoryProviderTests: XCTestCase {
...
+            stderrOutput: .success(Array(stdErr.utf8))
+        )
+        let error = GitShellError(result: result)
+        let errorString = "\(error)"
+        XCTAssertTrue(
+            errorString.contains(stdErr),
+            "Error string '\(errorString)' should contain '\(stdErr)'")
+        }
+
+    func testGitShelErrorEmptyStdErr() throws {
```
bripeticca, line 100: typo: `testGitShellErrorEmptyStdErr`

Tests/SourceControlTests/GitRepositoryProviderTests.swift
```diff
@@ -40,4 +40,76 @@ class GitRepositoryProviderTests: XCTestCase {
             XCTAssertThrowsError(try provider.isValidDirectory(notGitChildPath))
         }
     }
+
+    func testIsValidDirectoryThrowsPrintableError() throws {
+        try testWithTemporaryDirectory { temp in
+            let provider = GitRepositoryProvider()
+            let expectedErrorMessage = "not a git repository"
+            do {
```
bkhouri, line 48: question: out of curiosity, why did we not use `XCTAssertThrowsError` here?

https://developer.apple.com/documentation/xctest/1500795-xctassertthrowserror

bripeticca (changes requested):
LGTM, just noticed a few typos 😄

bkhouri (approved):

Tests/SourceControlTests/GitRepositoryProviderTests.swift
```diff
@@ -40,4 +40,76 @@ class GitRepositoryProviderTests: XCTestCase {
...
+            } catch let error {
+                let errorString = String(describing: error)
+                XCTAssertTrue(
+                    errorString.contains(expectedErrorMessage),
+                    "Error string '\(errorString)' should contain '\(expectedErrorMessage)'")
+            }
+        }
+    }
+
+    func testGitShelErrorIsPrintable() throws {
```
bripeticca, line 59: typo: `testGitShellErrorIsPrintable`

Tests/SourceControlTests/GitRepositoryProviderTests.swift
```diff
@@ -40,4 +40,76 @@ class GitRepositoryProviderTests: XCTestCase {
             XCTAssertThrowsError(try provider.isValidDirectory(notGitChildPath))
         }
     }
+
+    func testIsValidDirectoryThrowsPrintableError() throws {
+        try testWithTemporaryDirectory { temp in
+            let provider = GitRepositoryProvider()
+            let expectedErrorMessage = "not a git repository"
+            do {
+                try _ = provider.isValidDirectory(temp)
```
weissi, line 49: don't you need to either use `XCTAssertThorwsError` or an `XCTFail` after this line? It's wrong if it doesn't throw but at the moment, the test will pass, right?

yyvch: Thank you, you're absolutely right. It is `XCTAssertThrowsError` now in [8840c58](https://github.com/swiftlang/swift-package-manager/pull/8152/commits/8840c58ea6d14eb153ee01220e22c4763e2ac183)

Tests/SourceControlTests/GitRepositoryProviderTests.swift
yyvch: No good reason. Fixed in [8840c58](https://github.com/swiftlang/swift-package-manager/pull/8152/commits/8840c58ea6d14eb153ee01220e22c4763e2ac183).

Tests/SourceControlTests/GitRepositoryProviderTests.swift
yyvch: Thanks for vigilance. Fixed in [7c2b0ea](https://github.com/swiftlang/swift-package-manager/pull/8152/commits/7c2b0eabaa21d59cf10a34a81914b6f91d4647c3).

Tests/SourceControlTests/GitRepositoryProviderTests.swift
yyvch: Thanks for vigilance. Fixed in [7c2b0ea](https://github.com/swiftlang/swift-package-manager/pull/8152/commits/7c2b0eabaa21d59cf10a34a81914b6f91d4647c3).

plemarquand:
@swift-ci test

Tests/SourceControlTests/GitRepositoryTests.swift
```diff
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift open source project
 //
-// Copyright (c) 2014-2017 Apple Inc. and the Swift project authors
+// Copyright (c) 2014-2024 Apple Inc. and the Swift project authors
```
MaxDesiatov, line 5:
```suggestion
// Copyright (c) 2014-2017 Apple Inc. and the Swift project authors
```

Tests/SourceControlTests/GitRepositoryProviderTests.swift
```diff
@@ -40,4 +40,74 @@ class GitRepositoryProviderTests: XCTestCase {
...
+        XCTAssertTrue(
+            errorString.contains(stdOut),
+            "Error string '\(errorString)' should contain '\(stdOut)'")
+        XCTAssertTrue(
+            errorString.contains(stdErr),
+            "Error string '\(errorString)' should contain '\(stdErr)'")
+        XCTAssertTrue(
+            errorString.contains(command),
+            "Error string '\(errorString)' should contain '\(command)'")
+        }
```
MaxDesiatov, line 80: nit: formatting is off here and subsequent functions as well. Would you mind running SwiftFormat on this file, as [described in the corresponding section of `CONTRIBUTING.md`](https://github.com/swiftlang/swift-package-manager/blob/main/CONTRIBUTING.md#creating-pull-requests)?

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2 (approved):

bripeticca (approved):

MaxDesiatov:
@swift-ci test

Tests/SourceControlTests/GitRepositoryProviderTests.swift
yyvch: I do not mind running SwiftFormat, however I would rather prefer linter checks. Partially formatted codebase and no linter checks doesn't make a great experience for contributing. For example, running SwiftFormat `Sources/SourceControl/GitRepository.swift` reformats far more than I modified, which made be believe styling wasn't enforced.

Requested changed done in [28580c7](https://github.com/swiftlang/swift-package-manager/pull/8152/commits/28580c7376900bad57597c034c01ae8cb6d065b7).

MaxDesiatov:
@swift-ci test windows

xedin (approved):

PR #8154, opened 2024-12-03
Unify env vars for `Utilities/bootstrap` commands

MaxDesiatov:
This is a prerequisite for enabling incremental builds when running `./Utillities/bootstrap build --release` and `./Utilities/bootstrap test --release` in a sequence.

### Motivation:

Environment variables passed to `swift-bootstrap` from `./Utilities/bootstrap build` and `./Utilities/bootstrap test` commands unnecessarily differed slightly, which caused generated build manifests to be different. This precluded incremental builds when running these commands in a sequence one after each other, slowing down our CI jobs.

### Modifications:

Moved env var computation logic to the existing `get_swiftpm_env_cmd` function in `Utilities/bootstrap`.

### Result:

`swift-bootstrap` is consistently called with the same environment variables, ensuring that generated build manifests can be the same, eventually enabling incremental builds.

### DIFF
--- a/Utilities/bootstrap
+++ b/Utilities/bootstrap
@@ -359,16 +359,9 @@ def test(args):
     """Builds SwiftPM, then tests itself."""
     build(args)
 
-    if '-macosx' in args.build_target:
-        extra_env = ["SWIFTCI_DISABLE_SDK_DEPENDENT_TESTS=YES"]
-    else:
-        extra_env = []
-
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

dschaefer2 (approved):

jakepetroules (approved):

PR #8158, opened 2024-12-05
Enable cross-import overlays in test targets.

grynspan:
This PR enables cross-import overlays when building test targets. Swift Testing cannot link to most modules because it would/could introduce circular dependencies that prevent those modules from being tested using Swift Testing. However, we need to introduce some API that uses types from Foundation and other modules, which means we need to use the cross-import overlays feature supported by the compiler.

Cross-import overlays haven't been through Swift Evolution, so we don't want to enable the feature in the general case. We talked with @beccadax who agreed that because our use case is relatively unique, enabling these overlays just for test targets was probably the right approach until/unless the feature can be fully reviewed and supported.

We're also looking at consulting available information (the dependency scanner) to determine if there is any dependency on Swift Testing. If there isn't one, we don't need this flag. That's a "future direction" for this PR at any rate.


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test Windows

grynspan:
Need to do a full toolchain build before merging.

dschaefer2 (approved):

PR #8160, opened 2024-12-06
[cxx-interop] Do not pass `-lc++`/`-lstdc++` flags to the compiler

egorzhdan:
### Motivation:

The Swift compiler has recently introduced support for building with C++ stdlibs which aren't the default on the target platform, specifically libc++ on Linux. The Swift Driver has logic to determine which C++ stdlib should be used, depending on the target and explicit `-Xcc -stdlib=xyz` flags, and link against the requested standard library. Similar functionality exists in Clang Driver. We should defer the C++ stdlib selection to the compiler driver, it shouldn't be necessary to duplicate it in the build system.

### Modifications:

This makes sure that SwiftPM does not explicitly ask the compiler to link against a particular C++ stdlib.

### Result:

Projects that use libc++ on Linux build correctly with SwiftPM.

rdar://141047307

### DIFF
--- a/Sources/Build/BuildPlan/BuildPlan+Product.swift
+++ b/Sources/Build/BuildPlan/BuildPlan+Product.swift
@@ -54,27 +54,6 @@ extension BuildPlan {
             }
         }
 
-        // Don't link libc++ or libstd++ when building for Embedded Swift.
-        // Users can still link it manually for embedded platforms when needed,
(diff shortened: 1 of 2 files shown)


egorzhdan:
@swift-ci please test

Xazax-hun (approved):

dschaefer2 (approved):

jakepetroules (approved):

PR #8165, opened 2024-12-09
Fix comment typo in `SwiftCommandState.swift`

MaxDesiatov:
`the manifest file should *not* present` -> `the manifest file should *not* be present`

### DIFF
--- a/Sources/CoreCommands/SwiftCommandState.swift
+++ b/Sources/CoreCommands/SwiftCommandState.swift
@@ -238,8 +238,8 @@ public final class SwiftCommandState {
 
     /// Holds the currently active workspace.
     ///
-    /// It is not initialized in init() because for some of the commands like package `init`, usage etc,
-    /// workspace is not needed, in-fact it would be an error to ask for the workspace object
+    /// It is not initialized in init() because for some of the commands like `package init`, usage etc,
+    /// a workspace is not needed. In fact it would be an error to ask for the workspace object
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

Sources/CoreCommands/SwiftCommandState.swift
```diff
@@ -238,9 +238,9 @@ public final class SwiftCommandState {
 
     /// Holds the currently active workspace.
     ///
-    /// It is not initialized in init() because for some of the commands like package init , usage etc,
+    /// It is not initialized in init() because for some of the commands like package `init`, usage etc,
```
plemarquand, line 241:
```suggestion
    /// It is not initialized in init() because for some of the commands like `package init`, usage etc,
```

Sources/CoreCommands/SwiftCommandState.swift
```diff
@@ -238,9 +238,9 @@ public final class SwiftCommandState {
...
     /// workspace is not needed, in-fact it would be an error to ask for the workspace object
```
plemarquand, line 242:
```suggestion
    /// a workspace is not needed. In fact it would be an error to ask for the workspace object
```

MaxDesiatov:
@swift-ci test

plemarquand (approved):

jakepetroules (approved):

PR #8168, opened 2024-12-09
Revert "Workspace/InitPackage: Add `.vscode` to `.gitignore`"

plemarquand:
Reverts swiftlang/swift-package-manager#8164

### DIFF
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -439,7 +439,6 @@ public final class InitPackage {
                 .swiftpm/configuration/registries.json
                 .swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
                 .netrc
-                .vscode
 
                 """
             )


Sources/Workspace/InitPackage.swift
```diff
@@ -439,7 +439,6 @@ public final class InitPackage {
                 .swiftpm/configuration/registries.json
                 .swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
                 .netrc
-                .vscode
```
MaxDesiatov, line 442: WDYT of just changing this to `.vscode/launch.json`?

dschaefer2 (approved):

Sources/Workspace/InitPackage.swift
MaxDesiatov: My objection here is that auto-generated files should not be committed in the first place. I've bit gitignoring this file in my every single Swift package because VS Code kept recreating it with no way to prevent that, and I don't see a use case for keeping it really. At the very least, VS Code should keep auto-generated and manually maintained data in separate files.

plemarquand: What about something like:
```
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
!.vscode/*.code-snippets
```

Those un-ignored files prefixed with `!` are some you may want to share.

czechboy0: Same, every project I use VS Code with, I add this line into gitignore.

I'm curious, to make this more data-driven: could we search for packages that have Package.swift and also either have `.vscode` in the gitignore, or have at least one checked-in file from the `.vscode` directory? That way we could see what's more common and which group to optimize for (and the other group can always update their gitignore post-init).

MahdiBM: I think the thing is a bunch of stuff that `launch.json` supports are unsupported by SwiftPM itself (regardless if they make sense in SwiftPM at all) and are usually ignored in majority of the OpenSource projects.
For example I've never seen a reputable OpenSource project sharing their Xcode schemes, possibly containing some env vars as an example.
I think realistically most, specially OpenSource Swift projects currently won't want to share their `launch.json` file, so I'd be okay with ignoring this file.
That is, if there are no plans to change things, or if nobody foresees the community needing to share `launch.json`.

Overall I'd slightly lean on the side of ignoring `launch.json` although it can be useful to share in specific projects, in which case it's not hard to just remove that line from `.gitignore`.

plemarquand:
@swift-ci test

jakepetroules (approved):

Sources/Workspace/InitPackage.swift
bkhouri: Question: Can teams simply modify the generated `.gitignore` if it does not suite their needs?

plemarquand: @bkhouri yes. This is just about providing sensible defaults.

plemarquand:
@swift-ci test windows

Sources/Workspace/InitPackage.swift
czechboy0: So I actually did a search and unless I did it wrong, a tiny number of Swift packages are checking in their .vscode directory: https://github.com/search?q=language%3Aswift+path%3A%2F%5E%5C.vscode%5C%2F%2F&type=code

If anyone can confirm that I haven't done the search wrong, I feel that we should keep the `.vscode` entry in the generated `.gitignore`, and let the minority of teams for whom that's not the desired setup to change it by hand. Otherwise we're signing up the vast majority of new package authors for adding `.vscode` in their by hand for no good reason.

czechboy0 (changes requested):
~See my comment here, I think we should follow what the majority of packages prefer: https://github.com/swiftlang/swift-package-manager/pull/8168?notification_referrer_id=NT_kwDOACFL6bMxMzc2NzY3MjQwMzoyMTgyMTIx#discussion_r1877457236~

See clarification in the comment

Sources/Workspace/InitPackage.swift
czechboy0: Nevermind, I indeed did it wrong, this looks for Swift files in .vscode. Let me try to fix it.

czechboy0: I can't figure out the right search term here, let's ignore it.

I still think we should do what the majority of packages do, but I'll let the group decide what y'all believe that is.

plemarquand: @czechboy0 After giving it some thought I'd rather err on the side of caution and not prescribe something that users may have to undo. This template should be minimal and not opinionated.

czechboy0: That's fair, I'm fine either way, I'll just call out that the existing generated .gitignore is quite opinionated, ignoring several Xcode files that some people also want to check in: https://github.com/swiftlang/swift-package-manager/blob/11ebe9e07a1b99c6bbbc09ec1ec18093a3ada688/Sources/Workspace/InitPackage.swift#L431 So not sure if you want to remove those too, to avoid users have to remove them. Happy either way, it would just be good to be consistent between the various IDEs.

plemarquand: I'm unfamiliar with some of the xcode related entries in the gitignore. Seems like `xcuserdata` makes sense to ignore. I'm not sure about `contents.xcworkspacedata`.

Is the underlying issue that the VS Code Swift Extension is creating and modifying launch/tasks.json files in the .vscode folder? If so I think it makes sense to address this in the extension and if possible avoid auto-generating these configurations. That would mean that anything in `.vscode` would be put there explicitly by the user and then its totally up to them how to set up their `.gitconfig` around `.vscode`.

dschaefer2: > Is the underlying issue that the VS Code Swift Extension is creating and modifying launch/tasks.json files in the .vscode folder? If so I think it makes sense to address this in the extension and if possible avoid auto-generating these configurations. That would mean that anything in `.vscode` would be put there explicitly by the user and then its totally up to them how to set up their `.gitconfig` around `.vscode`.

I'm all for that. I actually find the autogenerated tasks for SwiftPM to be annoying. There's so many of them. That said I do appreciate help creating them. I'd hate to have to remember where my executable ended up.

czechboy0: > I'm unfamiliar with some of the xcode related entries in the gitignore. Seems like xcuserdata makes sense to ignore. I'm not sure about contents.xcworkspacedata.

Right, might make sense to declare bankruptcy and remove them all, that way at least we start from a consistent point.

xcuserdata is user data, but the argument for not having `.vscode` was that some people do want their user data committed, so I'm curious if we can get to a set of rules that help contributors understand what is and isn't in the gitignore.

plemarquand: @czechboy0 While you can configure whats in `.vscode` to be user data specific to a user I'd argue that its mainly for project configuration data. Settings, recommended extensions, launch tasks, etc are all project specific and not user specific. They can all be configured with relative paths so that when a new user checks out the project it builds in VS Code with no extra setup. I'd argue the default is to commit it and the less common path is to have non-portable configuration in there that should be .gitignored. If the VS Code Swift extension is generating launch configs that aren't portable or useful then we should have that discussion in that repository or on the forums.

For now I propose we go forward with this revert and continue to discuss.

czechboy0: Sure, go ahead. I've personally not worked on a single repo that checked in `.vscode` contents, all the ones I've worked with ignore it, but it's also not too difficult to add it to `.gitignore` on every project I work on, so happy with the revert 👍

czechboy0 (approved):

PR #8193, opened 2025-01-01
Fix building and testing on FreeBSD; Add FreeBSD platform support

michael-yuji:
Fix building and testing on FreeBSD and add support for "FreeBSD" platform

### Motivation:
Adding support for "FreeBSD" is swiftpm is necessary to get a usable toolchain building on FreeBSD.

#### Archiver
On FreeBSD, the `unzip` command is available in base but not the `zip` command. Therefore; we use libarchive (via `bsdtar`) to produce ZIP archive instead.

The implementation is also based on `libarchive` which produces different error messages; Tests are updated accordingly

#### compile / link arguments

FreeBSD ships with libc++ by default hence should use `-lc++`. Like Linux, `defsym` is also required to link targets with `_main` symbol.

#### FreeBSD native package manager (`pkgng`) support:

FreeBSD uses `pkgng` as its package manager, 

### Result:

Should be able to build swiftpm on FreeBSD. Testing on `x86_64 FreeBSD 14.2 RELEASE`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,7 +52,11 @@ find_package(Foundation QUIET)
 find_package(SQLite3 REQUIRED)
 
 # Enable `package` modifier for the whole package.
-add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:-package-name;SwiftPM>")
+if(CMAKE_SYSTEM_NAME STREQUAL FreeBSD)
+  add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:-package-name;SwiftPM>" -L/usr/local/lib)
+else()
+  add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:-package-name;SwiftPM>")
+endif()
 
 add_subdirectory(BuildSupport/SwiftSyntax)
 add_subdirectory(Sources)
--- a/Package.swift
+++ b/Package.swift
@@ -201,7 +201,7 @@ let package = Package(
             name: "Basics",
             dependencies: [
                 "_AsyncFileSystem",
-                .target(name: "SPMSQLite3", condition: .when(platforms: [.macOS, .iOS, .tvOS, .watchOS, .visionOS, .macCatalyst, .linux])),
+                .target(name: "SPMSQLite3", condition: .when(platforms: [.macOS, .iOS, .tvOS, .watchOS, .visionOS, .macCatalyst, .linux, .custom("freebsd")])),
                 .product(name: "SwiftToolchainCSQLite", package: "swift-toolchain-sqlite", condition: .when(platforms: [.windows, .android])),
                 .product(name: "DequeModule", package: "swift-collections"),
                 .product(name: "OrderedCollections", package: "swift-collections"),
--- a/Sources/Basics/Archiver/ZipArchiver.swift
+++ b/Sources/Basics/Archiver/ZipArchiver.swift
@@ -109,6 +109,11 @@ public struct ZipArchiver: Archiver, Cancellable {
             arguments: [windowsTar, "-a", "-c", "-f", destinationPath.pathString, directory.basename],
             workingDirectory: directory.parentDirectory
         )
+        #elseif os(FreeBSD)
+        let process = AsyncProcess(
+          arguments: ["tar", "-c", "--format", "zip", "-f", destinationPath.pathString, directory.basename],
+          workingDirectory: directory.parentDirectory
+        )
         #else
         // This is to work around `swift package-registry publish` tool failing on
         // Amazon Linux 2 due to it having an earlier Glibc version (rdar://116370323)
--- a/Sources/Basics/Archiver/ZipArchiver.swift
+++ b/Sources/Basics/Archiver/ZipArchiver.swift
@@ -110,6 +110,8 @@ public struct ZipArchiver: Archiver, Cancellable {
             workingDirectory: directory.parentDirectory
         )
         #elseif os(FreeBSD)
+        // On FreeBSD, the unzip command is available in base but not the zip command.
+        // Therefore; we use libarchive(bsdtar) to produce the ZIP archive instead.
         let process = AsyncProcess(
           arguments: ["tar", "-c", "--format", "zip", "-f", destinationPath.pathString, directory.basename],
           workingDirectory: directory.parentDirectory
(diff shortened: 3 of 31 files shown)


MaxDesiatov:
@swift-ci test

Sources/PackageModel/Manifest/SystemPackageProviderDescription.swift
```diff
@@ -16,11 +16,12 @@ public enum SystemPackageProviderDescription: Hashable, Codable, Sendable {
     case apt([String])
     case yum([String])
     case nuget([String])
+    case pkg([String])
```
jakepetroules, line 19: Is FreeBSD's `pkg` sufficiently "branded" to call it this? OpenBSD's package manager is named quite similarly, and while it doesn't seem to have a formal name, all of its commands begin with `pkg_`. `pkg` also isn't a particularly searchable term.

`pkg` _might_ still be the right choice, but I just want to make sure we're doing the thought exercise of considering the best name in the context of all the largest 4 BSDs (FreeBSD, OpenBSD, NetBSD, and DragonflyBSD) and what would be least confusing, even if we're not adding those others right now.

Here's some of my thoughts:
- NetBSD's `pkgsrc` is not exclusive to NetBSD, and has its own website / identity / branding. So I think an enum named `pkgsrc` would make sense there.
- OpenBSD's package system doesn't have a name. Call the enum `openbsd_pkg`?
- FreeBSD's package system doesn't have a name, and DragonflyBSD's package management software (as far as I can tell) is a fork of FreeBSD's (as is the OS generally). Two enums `freebsd_pkg` and `dragonflybsd_pkg`? Or a single enum `pkg` that applies to both FreeBSD and DragonflyBSD? That might be a bad idea though, as they are different OSes which may have diverged.

jakepetroules (commented):
Looks great, thanks! Just one thing: wanted to prime some discussion on the name of the `pkg` enum case and see if we might want to do anything differently there.

Sources/PackageModel/Manifest/SystemPackageProviderDescription.swift
michael-yuji: This is also in my mind, originally I name this `pkgng`, but `pkgng` renamed to `pkg` quite a while ago. https://wiki.freebsd.org/pkg. DragonflyBSD is using the same `pkg` from FreeBSD https://github.com/DragonFlyBSD/DPorts/blob/master/ports-mgmt/pkg/Makefile.

I'm not sure about OpenBSD tho, and currently the `SystemPackageProviderDescription ` enum does not include one for it.

Sources/Basics/Archiver/ZipArchiver.swift
```diff
@@ -85,6 +85,11 @@ public struct ZipArchiver: Archiver, Cancellable {
             arguments: ["tar.exe", "-a", "-c", "-f", destinationPath.pathString, directory.basename],
             workingDirectory: directory.parentDirectory
         )
+        #elseif os(FreeBSD)
+        let process = AsyncProcess(
```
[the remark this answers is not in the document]
michael-yuji: I'm not sure on Linux, since the `zip` support in `tar` is really a `bsdtar` aka `libarchive` feature (which windows also use)

Sources/PackageDescription/SupportedPlatforms.swift
```diff
@@ -69,6 +69,10 @@ public struct Platform: Equatable, Sendable {
     /// The OpenBSD platform.
     @available(_PackageDescription, introduced: 5.8)
     public static let openbsd: Platform = Platform(name: "openbsd")
+
+    /// The FreeBSD platform.
+//    @available(_PackageDescription, introduced: ??.??)
```
jakepetroules, line 74: Convention is to used an introduction version of 999.0 for non-finalized API.

Sources/PackageModel/Manifest/SystemPackageProviderDescription.swift
jakepetroules: OK, might be worth starting a Swift evolution thread on the naming. In the meantime you could land this behind an API availability of 999.0

Package.swift
```diff
@@ -201,7 +201,7 @@ let package = Package(
             name: "Basics",
             dependencies: [
                 "_AsyncFileSystem",
-                .target(name: "SPMSQLite3", condition: .when(platforms: [.macOS, .iOS, .tvOS, .watchOS, .visionOS, .macCatalyst, .linux])),
+                .target(name: "SPMSQLite3", condition: .when(platforms: [.macOS, .iOS, .tvOS, .watchOS, .visionOS, .macCatalyst, .linux, .custom("freebsd")])),
                 .product(name: "SwiftToolchainCSQLite", package: "swift-toolchain-sqlite", condition: .when(platforms: [.windows, .android])),
```
finagolfin, line 205: Great to see this port moving forward. 🥲

Have you tried using this SQLite package instead? The only reason most OSs above don't is that this is new: I switched my upcoming native 6.1 toolchain for Android over to using it and was able to [remove the external libsqlite package dependency](https://github.com/finagolfin/termux-packages/commit/a3d06c314b0e77e026cbcb19c58883b19d85af0d#diff-a3bcec5381cf1ae076d0c16e6be6792a91869d20be55c919085f0704fa11424cL11) of this Swift toolchain as a result. I suggest you try the same, will require [patching llbuild also](https://github.com/swiftlang/swift-llbuild/pull/954/files#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR17).

michael-yuji: Ah, I just saw your comment. This sounds like a good idea! Thanks!

michael-yuji:
@swift-ci please test

jakepetroules (approved):

jakepetroules:
@swift-ci please test

jakepetroules:
@swift-ci please test Windows

michael-yuji:
@swift-ci please test

CMakeLists.txt
```diff
@@ -52,7 +52,11 @@ find_package(Foundation QUIET)
 find_package(SQLite3 REQUIRED)
 
 # Enable `package` modifier for the whole package.
-add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:-package-name;SwiftPM>")
+if(CMAKE_SYSTEM_NAME STREQUAL FreeBSD)
+  add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:-package-name;SwiftPM>" -L/usr/local/lib)
```
jakepetroules, line 56: nit: may be worth a comment that the /usr vs /usr/local separation is a formal FreeBSD policy for all ports not part of the base system, otherwise this difference may look suspicious to future readers.

jakepetroules:
@swift-ci test windows

PR #8194, opened 2025-01-02
Fix registry package swizzling when package name casing differs

fortmarek:
Fix registry package swizzling when package name casing differs

### Motivation:

When swizzling dependencies when running `swift package --replace-scm-with-registry resolve`, package references that have different casing don't get swizzled.

You can find an example of this scenario in the [Firebase SDK](https://github.com/firebase/firebase-ios-sdk/blob/main/Package.swift) that boils down to:
```swift
import PackageDescription


let package = Package(
  name: "Firebase",
  dependencies: [
    .package(
      url: "https://github.com/google/promises.git", // lowercase
      "2.4.0" ..< "3.0.0"
    ),
  ],
  targets: [
    .target(
      name: "FirebaseCrashlytics",
      dependencies: [
        .product(name: "FBLPromises", package: "Promises"), // uppercase
      ]
    )
  ]
)
```

The `.product(name: "FBLPromises", package: "Promises")`should be swizzled to use the `google.promises` identifier, but because of the different casing, it's not.

### Modifications:

When creating the map for packages to be swizzled, make the key lowercase. When we access the map, we again use a lowercase variant of the package name, so we don't unexpectedly miss an entry because of a different casing.

### Result:

Packages with different casing are swizzled when using the `swift package --replace-scm-with-registry resolve` command.

### DIFF
--- a/Sources/Workspace/Workspace+Registry.swift
+++ b/Sources/Workspace/Workspace+Registry.swift
@@ -220,7 +220,7 @@ extension Workspace {
                                     info: "swizzling '\(dependency.locationString)' with registry dependency '\(registryIdentity)'."
                                 )
                             targetDependencyPackageNameTransformations[dependency
-                                .nameForModuleDependencyResolutionOnly] = registryIdentity.description
+                                .nameForModuleDependencyResolutionOnly.lowercased()] = registryIdentity.description
                             modifiedDependency = .registry(
                                 identity: registryIdentity,
                                 requirement: requirement,
@@ -261,7 +261,7 @@ extension Workspace {
                            let packageName
                         {
                             // makes sure we use the updated package name for target based dependencies
-                            if let modifiedPackageName = targetDependencyPackageNameTransformations[packageName] {
+                            if let modifiedPackageName = targetDependencyPackageNameTransformations[packageName.lowercased()] {
                                 modifiedDependency = .product(
                                     name: name,
                                     package: modifiedPackageName,
--- a/Sources/PackageModel/Manifest/PackageDependencyDescription.swift
+++ b/Sources/PackageModel/Manifest/PackageDependencyDescription.swift
@@ -152,7 +152,7 @@ public enum PackageDependency: Equatable, Hashable, Sendable {
                 return settings.nameForTargetDependencyResolutionOnly ?? PackageIdentityParser.computeDefaultName(fromURL: url)
             }
         case .registry:
-            return self.identity.description
+            return self.identity.description.lowercased()
         }
     }
 
--- a/Tests/WorkspaceTests/WorkspaceTests.swift
+++ b/Tests/WorkspaceTests/WorkspaceTests.swift
@@ -13000,6 +13000,70 @@ final class WorkspaceTests: XCTestCase {
             }
         }
     }
+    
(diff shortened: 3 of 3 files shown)


pepicrft (approved):

MaxDesiatov:
@swift-ci test

Sources/Workspace/Workspace+Registry.swift
```diff
@@ -220,7 +220,7 @@ extension Workspace {
                                     info: "swizzling '\(dependency.locationString)' with registry dependency '\(registryIdentity)'."
                                 )
                             targetDependencyPackageNameTransformations[dependency
-                                .nameForModuleDependencyResolutionOnly] = registryIdentity.description
+                                .nameForModuleDependencyResolutionOnly.lowercased()] = registryIdentity.description
```
plemarquand, line 223: Because package [names/identifiers are case insensitive](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/Registry.md#361-package-scope) I think it makes more sense to update the [.registry case in nameForModuleDependencyResolutionOnly](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/PackageModel/Manifest/PackageDependencyDescription.swift#L155) to return a lowercased name so we're consistent everywhere going forward.

plemarquand (changes requested):
Thanks for the patch, appreciate the contribution! I just have one small comment.

plemarquand:
@swift-ci please test

Sources/Workspace/Workspace+Registry.swift
fortmarek: I did that in the latest commit 👍  

However, `.lowercased()` still needs to be kept as `dependency` is not necessarily the `registry` case as that's done the line below. This could also be rewritten to:
```swift
                            modifiedDependency = .registry(
                                identity: registryIdentity,
                                requirement: requirement,
                                productFilter: settings.productFilter,
                                traits: settings.traits
                            )
                            targetDependencyPackageNameTransformations[modifiedDependency
                                .nameForModuleDependencyResolutionOnly] = registryIdentity.description
```

Let me know what you prefer, I don't have a clear preference here.

plemarquand: I think your proposed rewrite here to create the modifiedDependency first makes more sense, since it doesn't duplicate the lowercasing logic in  nameForModuleDependencyResolutionOnly

fortmarek: Actually, that rewrite is wrong. `modifiedDependency.nameForModuleDependencyResolutionOnly` yields the registry identifier (`org.foo`) whereas `dependency.nameForModuleDependencyResolutionOnly` yields the module name (`foo`). So, this needs to be kept as-is.

plemarquand:
@swift-ci test

plemarquand (approved):

fortmarek:
Hey @plemarquand 👋 

Is there anything blocking the merge of this PR?

plemarquand:
@fortmarek so sorry, this fell off my queue! I've kicked off the windows tests and enabled auto merge.

plemarquand:

plemarquand:
@swift-ci test windows

Issue #8197, opened 2025-01-06
SwiftRunCommand.exec(path:args:) causes crash with debug-build libdispatch


michael-yuji:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [X] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

In [`execute(path:args)`](https://github.com/swiftlang/swift-package-manager/blob/0401a2ae55077cfd1f4c0acd43ae0a1a56ab21ef/Sources/Commands/SwiftRunCommand.swift#L322), it closes all file descriptors unconditionally, this includes file descriptors owned by `libdispatch`, which triggers [an assertion](https://github.com/swiftlang/swift-corelibs-libdispatch/blob/b2a1fad6b49a9394cf864b87ab0034ea8300e2de/src/event/event_kevent.c#L750) and crash the process with the message: "BUG IN CLIENT OF LIBDISPATCH: Do not close random Unix descriptors"

This causes failure in building the toolchain with debuggable libdispatch on systems using kqueue with option `DISPATCH_USE_MGR_THREAD`.

Some details of what's happening:

1. [`_dispatch_kq_create`](https://github.com/swiftlang/swift-corelibs-libdispatch/blob/b2a1fad6b49a9394cf864b87ab0034ea8300e2de/src/event/event_kevent.c#L579) creates a kqueue file descriptor and set `dispatch_mgr_q.do_ctxt` to it.

2. [`execute(path:args)`](https://github.com/swiftlang/swift-package-manager/blob/0401a2ae55077cfd1f4c0acd43ae0a1a56ab21ef/Sources/Commands/SwiftRunCommand.swift#L322) closes all file descriptors, which includes the kqueue created in (1), now `execve` enters a race with `libdispatch`.

[...]

PR #8201, opened 2025-01-07
use closefrom(2) on platforms supports it to reduce race condition window

michael-yuji:
Use `closefrom(2)` instead of `close(2)` in a loop to reduce the window of race condition with other part of the process cloud be polling/processing the file descriptors. 

### Motivation:

Currently, a loop is used to close all possible file descriptors prior to `execve(2)` being called. This has the side effect that file descriptors owned by different components of the same process is being closed unexpectedly, which cause issue #8197.

Use `closefrom(2)` instead of `close` in a loop to significantly reduce the window between closing the file descriptors and calling `execve(2)`, to minimize the impact.

In #8197, linking a debug build of `libdispatch` can randomly crash `swift-run`, due to dispatch catching that the file descriptors it works on no longer being valid, which causes an assertion and crash the process.

### Modifications:

Use `closefrom(2)` instead of `close(2)` in a loop on platforms supports it.

### Result:

Reduce the probability `swift-run` crash due to race conditions.

### DIFF
--- a/Sources/Commands/SwiftRunCommand.swift
+++ b/Sources/Commands/SwiftRunCommand.swift
@@ -312,16 +312,20 @@ public struct SwiftRunCommand: AsyncSwiftCommand {
         sigfillset(&sig_set_all)
         sigprocmask(SIG_UNBLOCK, &sig_set_all, nil)
 
+        #if os(FreeBSD) || os(OpenBSD)
+        closefrom(3)
+        #else
         #if os(Android)
         let number_fds = Int32(sysconf(_SC_OPEN_MAX))
         #else
         let number_fds = getdtablesize()
-        #endif
+        #endif /* os(Android) */
         
         // 2. close all file descriptors.
         for i in 3..<number_fds {
             close(i)
         }
+        #endif /* os(FreeBSD) || os(OpenBSD) */
         #endif
 
         try TSCBasic.exec(path: path, args: args)


michael-yuji:
@3405691582 I'm making the assumption that `closefrom(2)` is also available on the OpenBSD Swift toolchain.

jakepetroules:
> @3405691582 I'm making the assumption that `closefrom(2)` is also available on the OpenBSD Swift toolchain.

Looks like it should be: https://man.openbsd.org/closefrom

jakepetroules (approved):

3405691582:
Yes, it's on the platform, but don't we need to `import Glibc` for that? Or am I missing something?

jakepetroules:
Ah, that I don't know. Are there any libc imports needed for FreeBSD?

michael-yuji:
On FreeBSD, no further imports are needed. 🤔

3405691582:
It's probably fine then. It's been a while for me so I'm guessing it's probably some transitive import?

In the worst case, when I get back to tweaking SPM, can revisit.

3405691582 (approved):

michael-yuji:
@swift-ci please test

jakepetroules:
@swift-ci please test Windows

jakepetroules:
Merged at @michael-yuji's request as he does not yet have write access.

PR #8205, opened 2025-01-08
Move traits into resolution phase

bripeticca:
### Motivation:

Currently, target dependencies that are guarded by traits are being calculated in the modules graph loading phase - there is a desire to move this functionality into the resolution phase of SwiftPM to avoid doing redundant calculations on dependencies that will not be used further down the pipeline.

### Modifications:

Since there is an inherent reliance on expecting the resolution of unused dependencies in SwiftPM's test suite, a flag `--prune-unused-dependencies` has been added to act as a stopgap for unexpected test failures and should only be considered as a feature for trait-guarded target dependencies. This should be treated as an experimental feature for any other unused dependencies for the time being.

The `Manifest` is responsible for providing trait-related endpoints since it stores all defined traits; there are new methods allowing for the computation of which traits are enabled given a set of explicitly enabled traits (either defined by the user passing in a trait configuration through CLI, or defined by a manifest on its `PackageDependency`), whether a trait guards a particular `TargetDependency`, etc. The `dependenciesRequired` method considers traits in its computation as well, depending on whether pruning dependencies is enabled.

When a `Workspace` calls `computeDependencies`, there is a new set of package identities being calculated in the result that represent the set of unused dependencies (named `unusedDependencies`) that are described in the root package's set of dependencies - this assures that we can track which dependency manifests will be unnecessary to load. Note that dependencies that are guarded by traits are inherently included in this calculation.

A set of traits representing the **enabled traits** of a package have been added as a property to both `PackageContainerConstraint` as well as `DependencyResolutionNode`, since it is necessary to track this as a requirement during dependency resolution and acts as a constraint on the computation.

### Result:

With traits being factored into dependency resolution, SwiftPM can omit guarded dependencies from its calculation thus avoiding unnecessary checkouts and - subsequently - dependency version computations.


Sources/PackageModel/Manifest/Manifest.swift
```diff
@@ -560,3 +563,171 @@ extension Manifest: Encodable {
         try container.encode(self.packageKind, forKey: .packageKind)
     }
 }
+
+/// Helper methods that enable data collection through traits configurations in manifests.
+extension Manifest {
```
bkhouri, line 568: suggestion (possibly-blocking): In order to ensure this extension behaves as intended, can we write `small` tests that verify the behaviour of each  [public] method in isolation, taking into account fault injections (whenever possible), boundary conditions, collections (empty, single item, two items, multiple items), etc...

If we are unable to control the test the way we want, we may need to update the production code to make it support testability.

Sources/PackageModel/Manifest/TargetDescription.swift
```diff
@@ -43,6 +43,27 @@ public struct TargetDescription: Hashable, Encodable, Sendable {
             }
         }
 
+        public var name: String {
```
bkhouri, line 46: suggestion (possibly-blocking): Same comment here about creating `small` tests to verify the function/attributes in isolation.

bkhouri (commented):
I know this is a WIP, but I thought I'd provide some feedback from a testing perspective.

bripeticca:
@swift-ci please test

Sources/Workspace/Workspace+Dependencies.swift
```diff
@@ -1234,3 +1268,14 @@ extension Workspace.ManagedDependencies {
         })
     }
 }
+
+extension Manifest.TraitConfiguration {
+    public init(_ traitConfiguration: TraitConfiguration) {
+        self.init(enabledTraits: traitConfiguration.enabledTraits, enableAllTraits: traitConfiguration.enableAllTraits)
+    }
+
+    public init?(_ traitConfiguration: TraitConfiguration?) {
```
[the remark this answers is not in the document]
MaxDesiatov: What's the benefit of having `NullTraitConfiguration` instead of `nil` here? In theory, it could make sense in languages with null references like Java, as the article you linked mentions. Swift has neither implicit null reference, nor `TraitConfiguration` is an object, nor it is a reference, it's a value type that has no reference semantics whatsoever.

Sources/CoreCommands/Options.swift
```diff
@@ -294,6 +294,17 @@ public struct ResolverOptions: ParsableArguments {
     public var sourceControlToRegistryDependencyTransformation: SourceControlToRegistryDependencyTransformation =
         .disabled
 
+    /// Enables pruning unused dependencies to omit redundant calculations during resolution, and each phase thereafter.
+    /// Hidden from the generated help text as this feature is only currently being considered for traits.
+    @Flag(
+        name: .customLong("prune-unused-dependencies"),
```
FranzBusch, line 300: Are we intending for this to flip to `true` or removed soon? I think having the default be `false` will make this almost unused.

Sources/CoreCommands/SwiftCommandState.swift
```diff
@@ -256,6 +256,9 @@ public final class SwiftCommandState {
     // should use sandbox on external subcommands
     public var shouldDisableSandbox: Bool
 
+    /// Flag that determines whether a manifest's dependencies should be pruned if unused.
+//    public var shouldPruneDependencies: Bool = false
+
```
FranzBusch, line 261: Can this be removed?

bripeticca: Yup! Thanks for catching that. :)

Sources/CoreCommands/Options.swift
bripeticca: The intent is to flip it to true once there is better support for the ability to omit unused dependencies during resolution - currently a lot of tests in SwiftPM rely on the current behaviour which resolves dependencies regardless of their usage. This is a part of another feature that I have on my plate at the moment!

FranzBusch: So is the problem that this PR is also removing unused dependencies not gated by traits but just generally unused?

dschaefer2: Correct. We have some tests that test in those scenarios that were failing. Made us worried that people may actually be depending on them not being removed. Needs more investigation.

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

dschaefer2 (approved):

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

PR #8217, opened 2025-01-14
Revert "Retrieve resolved package versions in parallel"

dschaefer2:
Reverts swiftlang/swift-package-manager#8203

This is causing crashes in SwiftPM during the parallelized checkout. The checkouts update the workspace state and then save it. The workspace save data is not protected against multi-thread access while the save is walking the data which is causing the crash.

### DIFF
--- a/Sources/Workspace/Workspace+Dependencies.swift
+++ b/Sources/Workspace/Workspace+Dependencies.swift
@@ -426,31 +426,26 @@ extension Workspace {
         }
 
         // Retrieve the required resolved packages.
-        await withThrowingTaskGroup(of: Void.self) { taskGroup in
-            for resolvedPackage in requiredResolvedPackages {
-                let observabilityScope = observabilityScope.makeChildScope(
-                    description: "retrieving resolved package versions for dependencies",
-                    metadata: resolvedPackage.packageRef.diagnosticsMetadata
-                )
-                taskGroup.addTask {
-                    await observabilityScope.trap {
-                        switch resolvedPackage.packageRef.kind {
-                        case .localSourceControl, .remoteSourceControl:
-                            _ = try await self.checkoutRepository(
-                                package: resolvedPackage.packageRef,
-                                at: resolvedPackage.state,
-                                observabilityScope: observabilityScope
-                            )
-                        case .registry:
-                            _ = try await self.downloadRegistryArchive(
-                                package: resolvedPackage.packageRef,
-                                at: resolvedPackage.state,
-                                observabilityScope: observabilityScope
-                            )
-                        default:
-                            throw InternalError("invalid resolved package type \(resolvedPackage.packageRef.kind)")
-                        }
-                    }
+        for resolvedPackage in requiredResolvedPackages {
+            await observabilityScope.makeChildScope(
+                description: "retrieving resolved package versions for dependencies",
+                metadata: resolvedPackage.packageRef.diagnosticsMetadata
+            ).trap {
+                switch resolvedPackage.packageRef.kind {
+                case .localSourceControl, .remoteSourceControl:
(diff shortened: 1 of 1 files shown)


dschaefer2:
@swift-ci please test

MaxDesiatov (approved):

bkhouri (approved):

fortmarek:
@dschaefer2 what do you think about parallelizing only downloads of registry packages? That is, instead of parallelizing both `checkoutRepository` and `downloadRegistryArchive`, we'd parallelize only the latter? Alternatively, `checkoutRepository` could be made thread-safe, but I'm not sure how complex that would be.

The parallelization makes a really big difference in performance, so still would love to make it work.

fortmarek:
This is what I had in mind: https://github.com/swiftlang/swift-package-manager/pull/8218

dschaefer2:
I think the right solution is to make `Workspace` an actor so that its state is concurrency safe. That's one thing I've noticed working on SwiftPM the last few months is how so much of it predates Swift concurrency and really hasn't adopted it fully. However that's also a pretty big change since all clients would need to become async as well.

I was actually excited about this change for the non-registry case since that's a common complaint I've seen as well. But looking at the registry case, it is also changing the list of dependencies in the Workspace so would have the same problem anyway.

fortmarek:
> I think the right solution is to make Workspace an actor so that its state is concurrency safe. That's one thing I've noticed working on SwiftPM the last few months is how so much of it predates Swift concurrency and really hasn't adopted it fully. However that's also a pretty big change since all clients would need to become async as well.

Is this something the SwiftPM team plans to work on? Any way we can push it forward? The sequential download of registry is quite a drag on its performance – and so is for source control checkouts. I would love to push this initiative further, but there's also so much time I can dedicate to this and it sounds like making `Workspace` an actor could have a little of ripple effects.

> But looking at the registry case, it is also changing the list of dependencies in the Workspace so would have the same problem anyway.

Got it. Let's go with the revert. I do wonder in the interim, if we could at least split the download operation and updating the state into two steps. However, it's definitely preferred to make `Workspace` an actor, I'm just uncertain how complex that would be.

fortmarek:
> If you can think of a compromise that makes sure this parallelization serializes access to the Workspace state I would be OK with that for a short term solution. Though looking at the code, it might be sufficient to make WorkspaceState an actor which has a much smaller API service and the affects might not be so bad. Looking at the updateCheckout for example, it's already async.

I'll try having a look at this. Thanks for the pointers and sorry for breaking things!

jakepetroules (approved):

dschaefer2:
> I'll try having a look at this. Thanks for the pointers and sorry for breaking things!

All good. It helped highlight we really need to make SwiftPM more thread safe. As a central piece of the Swift toolchain, we really should be able to hold it up as a good example of modern Swift. :).

fortmarek:
@dschaefer2 here's a PR that makes `WorkspaceState` into an `actor`, it wasn't too bad: https://github.com/swiftlang/swift-package-manager/pull/8220

Were you able to consistently reproduce the crash? I ran the test suite multiple times and things look a-ok, but would love to double check.

bkhouri:
@swift-ci please test windows

bkhouri:
> @dschaefer2 here's a PR that makes `WorkspaceState` into an `actor`, it wasn't too bad: https://github.com/swiftlang/swift-package-manager/pull/8220
> 
> 
> 
> Were you able to consistently reproduce the crash? I ran the test suite multiple times and things look a-ok, but would love to double check.

@fortmarek personally, the WorkspaceTests failed for me 7 out of 30. I can try the same with #8220 tomorrow!

dschaefer2:
@swift-ci please test windows

PR #8218, opened 2025-01-14
Parallelize only downloading registry archives, not repository checkouts

fortmarek:

### Motivation:

### Modifications:

### Result:


fortmarek:
Closing as it would have the same issues as the [current solution](https://github.com/swiftlang/swift-package-manager/pull/8217#issuecomment-2590875467):
> I was actually excited about this change for the non-registry case since that's a common complaint I've seen as well. But looking at the registry case, it is also changing the list of dependencies in the Workspace so would have the same problem anyway.

Issue #8215, opened 2025-01-14
Package Registry client implementation does not support pagination using Link headers


ehyche:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

In the specification for the [Swift Package Registry Service](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/Registry.md#41-list-package-releases), the specification for the `GET /{scope}/{name}` endpoint says:

> A server MAY paginate results by responding with a Link header field that includes any of the following relations:

and provides these examples of a `Link` header:

```
Link: <https://packages.example.com/mona/HashMap/5.0.3>; rel="latest-version",
      <https://packages.example.com/mona/HashMap?page=1>; rel="first",
      <https://packages.example.com/mona/HashMap?page=2>; rel="previous",
      <https://packages.example.com/mona/HashMap?page=4>; rel="next",
      <https://packages.example.com/mona/HashMap?page=10>; rel="last"
```

**However, it does not appear that the SPM Package Registry client implementation supports pagination for the `GET /{scope}/{name}` endpoint.**

This is a burden on Swift Package Registry Service (SPRS) implementations. For example, if you were implementing an SPRS service which proxied the Github API, you would most likely use [this Github API endpoint](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repository-tags) to implement the `GET /{scope}/{name}`.

Take an example of implementing `GET /{scope}/{name}` for  [swift-syntax](https://github.com/swiftlang/swift-syntax).

# You get the request for `GET /swiftlang/swift-syntax` from SPM to your SPRS
# You proxy this request to the Github API: `GET https://api.github.com/repos/swiftlang/swift-syntax/tags`
# This response uses the default value of `per_page` which is 30, and responds with the first 30 tags. Here is the `Link` header in that response:

```
Link: <https://api.github.com/repositories/143079594/tags?per_page=30&page=2>; rel="next", <https://api.github.com/repositories/143079594/tags?per_page=30&page=47>; rel="last"
```

[...]

PR #8219, opened 2025-01-14
Handle paginated registry metadata responses

plemarquand:
### Motivation:

In [4.1 List Package Releases](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/Registry.md#41-list-package-releases) it is stated that a server may respond with a `Link` header that contains a pointer to a subsequent page of results. SPM is not checking for this link in the `Link` header and so if a registry returns paginated results, only the first page of versions is searched when resolving. This would result in a "version not found" error when in reality the version is present in a subsequent page of results.

### Modifications:

Respect the `next` link in the `Link` header by loading the next page of results and building up a list of versions, continuing until there is no `next` link present in the `Link` header of the last result.

### Result:

Registry servers that serve paginated results now have all their results read.

Issue: #8215


plemarquand:
I still need to write some tests for this PR.

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -151,129 +151,181 @@ public final class RegistryClient: Cancellable {
...
+    @available(*, noasync, message: "Use the async alternative")
+    public func getPackageMetadata(
+        package: PackageIdentity,
+        timeout: DispatchTimeInterval? = .none,
         observabilityScope: ObservabilityScope,
         callbackQueue: DispatchQueue,
         completion: @escaping (Result<PackageMetadata, Error>) -> Void
     ) {
         let completion = self.makeAsync(completion, on: callbackQueue)
+        _ = Task {
```
MaxDesiatov, line 194: With the unstructured task spawned, how does this handle cancellation?

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -2290,16 +2372,60 @@ extension HTTPClientHeaders {
...
+    }
+}
+
+extension HTTPClientHeaders {
+    /*
+    https://github.com/swiftlang/swift-package-manager/blob/0340bb12a56f9696b3966ad82c2aee1594135377/Documentation/PackageRegistry/Registry.md?plain=1#L403-L411
+    <https://github.com/mona/LinkedList?page=2>; rel="next",
+    <ssh://git@github.com:mona/LinkedList.git?page=40>; rel="last",
+    */
+    fileprivate func parsePagniationLinks() -> [RegistryClient.NextLocationLink] {
```
MaxDesiatov, line 2414:
```suggestion
    fileprivate func parsePaginationLinks() -> [RegistryClient.NextLocationLink] {
```

Sources/PackageRegistry/RegistryClient.swift
plemarquand: Good point, cancellation is not handled, though was it handled previously? As far as I can tell the `callbackQueue` typically provided is `DispatchQueue.sharedConcurrent` and DispatchQueue doesn't have a nice mechanism for this.

MaxDesiatov: `Task.isCancelled` will always be `false` since only structured child tasks get cancellation propagated from parent tasks. `Task` is an unstructured task with no parents, the only primitives for structured concurrency that propagate cancellation are `async let` and `withTaskGroup`.

plemarquand: OK, that is consistent with my understanding. Seeing as the sync version of the API didn't handle cancellation before I think its reasonable to keep that behaviour.

However you're correct that the async version _should_ handle cancellation, and I've updated the PR accordingly.

plemarquand:
@swift-ci test

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -151,129 +151,183 @@ public final class RegistryClient: Cancellable {
...
-            return completion(.failure(RegistryError.invalidPackageIdentity(package)))
+            throw RegistryError.invalidPackageIdentity(package)
         }
 
         guard let registry = self.configuration.registry(for: registryIdentity.scope) else {
-            return completion(.failure(RegistryError.registryNotConfigured(scope: registryIdentity.scope)))
+            throw RegistryError.registryNotConfigured(scope: registryIdentity.scope)
         }
 
-        observabilityScope.emit(debug: "registry for \(package): \(registry)")
```
cmcgee1024, line 185: question: Where'd this observability message go?

Tests/PackageRegistryTests/RegistryClientTests.swift
```diff
@@ -88,12 +88,205 @@ final class RegistryClientTests: XCTestCase {
         let registryClient = makeRegistryClient(configuration: configuration, httpClient: httpClient)
         let metadata = try await registryClient.getPackageMetadata(package: identity)
         XCTAssertEqual(metadata.versions, ["1.1.1", "1.0.0"])
-        XCTAssertEqual(metadata.alternateLocations!, [
+        XCTAssertEqual(metadata.alternateLocations, [
```
cmcgee1024, line 91: praise: There's one more happy kitten and this is less loud.

Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -151,129 +151,183 @@ public final class RegistryClient: Cancellable {
...
-                    case 200:
-                        let packageMetadata = try response.parseJSON(
-                            Serialization.PackageMetadata.self,
-                            decoder: self.jsonDecoder
-                        )
+        observabilityScope.emit(info: "retrieving \(package) metadata from \(url)")
+
+        let response: LegacyHTTPClient.Response
+        do {
+            response = try await self.httpClient.get(
```
cmcgee1024, line 285: thought: I wonder if there is an openapi spec for the package registries. It probably wouldn't handle pagination out of the box (an assumption). But, it might eliminate some boilerplate and make mocking with Swift datatypes instead of JSON payloads.

cmcgee1024 (approved):
Overall this looks good to me. I've just added a few thoughts and questions.

dschaefer2 (approved):

Sources/PackageRegistry/RegistryClient.swift
jakepetroules: @czechboy0 would know if OpenAPI specs have a way to communicate pagination.

jakepetroules (approved):

Sources/PackageRegistry/RegistryClient.swift
czechboy0: Yes I refreshed the OpenAPI doc a few months ago, and it would remove a bunch of the manual networking code here: https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/registry.openapi.yaml

OpenAPI itself doesn't get as high level as providing conventions for pagination. But the generated code will give you the Link header, it's documented in the OpenAPI doc.

czechboy0: If you adopt Swift OpenAPI Generator, I think it'd be reasonable to file an issue on us to provide the Link header parsing out of the box, so you'd get typesafe access to the individual link values. That way you could drop most of this manual parsing code.

PR #8220, opened 2025-01-14
Paralellize retrieving resolved packages

fortmarek:
Paralellize retrieving resolved packages

### Motivation:

This PR brings back the parallelization reverted [here](https://github.com/swiftlang/swift-package-manager/pull/8217).

As raised in the revertal, checkouts and registry downloads mutate the `WorkspaceState` – and running these operations in parallel can lead to a crash since the `WorkspaceState` is not thread-safe.

### Modifications:

Based on [this](https://github.com/swiftlang/swift-package-manager/pull/8217#issuecomment-2590905617) suggestion, I'm making `WorkspaceState` an `actor` to make access to it thread-safe.

Additionally, I turned `ManagedDependencies` into a `struct` to ensure its thread-safety, too. There are two methods of this struct that mutate self – `add` and `remove`. I changed this to return a new copy of `ManagedDependencies` instead and add counterpart methods in `WorkspaceState` where the mutation is done in-place. Since `ManagedDependencies` is a `Collection`, making it into an `actor` because of two mutable methods didn't feel like the right fit. However, let me know what you think about it.

I added comments to the key modifications to make it easier to find the more interesting modifications since there's a lot of noise since I had to add loads of `async` and `await`. You can find these comments below. 

### Result:

[Retrieving resolved packages](https://github.com/swiftlang/swift-package-manager/blob/0340bb12a56f9696b3966ad82c2aee1594135377/Sources/Workspace/Workspace%2BDependencies.swift#L428) in parallel should not lead to crashes.

As suggested by @bkhouri [here](https://github.com/swiftlang/swift-package-manager/pull/8217#issuecomment-2591337202), I ran `WorkspaceTests` multiple times to see if I run into any thread-safety issues:
```
for ((i=1; i<=30; i++)); do swift test --filter WorkspaceTests.WorkspaceTests; done
for ((i=1; i<=10; i++)); do swift test --filter WorkspaceTests; done
```


Sources/Workspace/Workspace+Dependencies.swift
```diff
@@ -426,26 +427,31 @@ extension Workspace {
...
-                    )
-                case .registry:
-                    _ = try await self.downloadRegistryArchive(
-                        package: resolvedPackage.packageRef,
-                        at: resolvedPackage.state,
-                        observabilityScope: observabilityScope
-                    )
-                default:
-                    throw InternalError("invalid resolved package type \(resolvedPackage.packageRef.kind)")
+        await withThrowingTaskGroup(of: Void.self) { taskGroup in
```
fortmarek, line 430: Returning back the parallelization of retrieving resolved packages

Sources/Workspace/Workspace+State.swift
```diff
@@ -19,7 +19,7 @@ import SourceControl
 import struct TSCUtility.Version
 
 /// Represents the workspace internal state persisted on disk.
-public final class WorkspaceState {
+public actor WorkspaceState {
```
fortmarek, line 22: Making `WorkspaceState` into actor to make it thread-safe

Sources/Workspace/ManagedDependency.swift
```diff
@@ -172,12 +172,18 @@ extension Workspace.ManagedDependency: CustomStringConvertible {
 
 extension Workspace {
     /// A collection of managed dependencies.
-    final public class ManagedDependencies {
+    public struct ManagedDependencies {
```
fortmarek, line 175: Turning `ManagedDependencies` into a `struct`. The alternative would be to make it an `actor` but since it conforms to `Collection`, it wouldn't be that easy.

There are only two methods of `ManagedDependencies` that are mutable – `add` and `remove`. Instead of mutating `self`, we can return copy of `Self` and do the mutation in `WorkspaceState`.

Sources/Workspace/Workspace+State.swift
```diff
@@ -88,6 +88,14 @@ public final class WorkspaceState {
...
         self.prebuilts = storedState.prebuilts
     }
+    
+    public func add(dependency: Workspace.ManagedDependency) {
+        dependencies = dependencies.add(dependency)
+    }
+
+    public func remove(identity: PackageIdentity) {
+        dependencies = dependencies.remove(identity)
+    }
```
fortmarek, line 98: To keep the setter of `dependencies` `private`, we need to add helper methods for the two operations that mutate `ManagedDependencies`. Alternatively, we could make the setter public, however, I do find this makes callsites of `WorkspaceState` consumers nicer, so I'm leaning to keep the current solution.

fortmarek:
@swift-ci test

bkhouri:
I ran, from the root of the repository, the following shell script against this PR and there were no entries in the associated `failed` directory

```
#!/bin/bash
# cmd="swift test --very-verbose --filter CommandsTests.RunCommandTests/testMultipleExecutableAndExplicitExecutable"
cmd=(
    "git log -n1"
    "time swift test --filter \"Workspace*Tests\" --parallel --disable-swift-testing"
)
BASEDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"


custom_run=$(date -u +'%Y%m%dT%H%M%S%Z')
root_log_dir="${HOME}/Downloads/can_be_deleted/swiftpm_logs/${custom_run}"
failed_dir="${root_log_dir}/failed"

caffeinate_cmd=$(command -v caffeinate)

function cleanup() {
    echo ""
    echo ""
    echo "******************************************************"
    echo "Root log directory  : ${root_log_dir}"
    echo "Failed log directory: ${failed_dir}"
}

trap cleanup EXIT

mkdir -p ${failed_dir}


num_iterations=300
# for num in $(seq -w 2); 
echo "Executing         : ${cmd}"
echo "Root Log Directory: ${root_log_dir}"
set -x 
which swift 
xcode-select -p
swift --version
set +x
for num in $(seq -w $num_iterations); 
do 
    (
        cd ${BASEDIR}
        log_dir=${root_log_dir}/${num}
        message="[${num}/${num_iterations}] executing and writing log to ${log_dir}"
        echo "${message} ..."
        mkdir -p ${log_dir}
        test_log=${log_dir}/swift_test_console_${num}.txt
        rm -rf ${test_log}
        start_time=$(date +%s)
        for ((i = 0; i < ${#cmd[@]}; i++))
        do
            full_cmd="${caffeinate_cmd} ${cmd[${i}]}"
            echo "❯❯❯ Executing: ${full_cmd}" >> ${test_log}
            # set -x
            # caffeinate ${cmd} 2>&1 | tee -a "${test_log}"
            ${full_cmd} >> "${test_log}" 2>&1
            rc=$?
            # set +x
            if [ ${rc} -ne 0 ] ; then
            # if [ ${PIPESTATUS[0]} -ne 0 ] ; then
                ln -s ${log_dir} ${failed_dir}
            fi

        done
        end_time=$(date +%s)
        echo "${message} completed in $(( ($end_time - $start_time) )) seconds"
        
        # echo "Waiting 0.5 second..."
        # echo "Done run number ${num}.  Log dir: ${log_dir}"
        sleep 0.5 
    )
done
echo "Done all ${um_iterations} interations"
```

fortmarek:
Thanks for checking @bkhouri. I believe then the previous thread-safety issues should be fixed now 🙏 cc @dschaefer2

bkhouri:
I added the utility script in #8227

bkhouri:
@fortmarek Have you tried executing all of the swift tests on your change?  Maybe it's me, but I'm seeing a few test failures when running `swift test --parallel`.

fortmarek:
Thanks @bkhouri! I don't think it has to do anything with running the tests in parallel, but some `PackageCommandTests` are actually failing. I will look into it 🙂

fortmarek:
@bkhouri the issue should be fixed – `SwiftCommand `should now conform to `AsyncParsableCommand` instead of `ParsableCommand` as the `run` method has been changed to being async.

fortmarek:
@swift-ci test

bkhouri:
> @bkhouri the issue should be fixed – `SwiftCommand `should now conform to `AsyncParsableCommand` instead of `ParsableCommand` as the `run` method has been changed to being async.

@fortmarek Sorry for the delay.  Let me rerun the script against your latest changes to ensure there no regression or intermittent issues were introduced. 

I'll trigger the CI builds against the change in the interim

bkhouri:
@swift-ci please test

bkhouri:
@swift-ci please test macos linux

plemarquand:
@swift-ci please test

fortmarek:
Hey @bkhouri @plemarquand @MaxDesiatov – sorry for the broad ping, but would love to see this get in 🙏

pepicrft:
This seems to be passing on CI 🚀

Sources/CoreCommands/SwiftCommandState.swift
```diff
@@ -98,14 +98,14 @@ extension _SwiftCommand {
     }
 }
 
-public protocol SwiftCommand: ParsableCommand, _SwiftCommand {
-    func run(_ swiftCommandState: SwiftCommandState) throws
+public protocol SwiftCommand: AsyncParsableCommand, _SwiftCommand {
```
MaxDesiatov, line 101: How is this protocol different now from the existing `AsyncSwiftCommand`? Why that protocol is not used instead?

Sources/Commands/PackageCommands/ResetCommands.swift
```diff
@@ -45,8 +45,8 @@ extension SwiftPackageCommand {
         @OptionGroup(visibility: .hidden)
         var globalOptions: GlobalOptions
 
-        func run(_ swiftCommandState: SwiftCommandState) throws {
-            try swiftCommandState.getActiveWorkspace().reset(observabilityScope: swiftCommandState.observabilityScope)
+        func run(_ swiftCommandState: SwiftCommandState) async throws {
```
MaxDesiatov, line 48: Is there a specific reason `SwiftPackageCommand.Reset` shouldn't conform to `AsyncSwiftCommand` now?

MaxDesiatov (changes requested):
With this PR as is, the relationship between `SwiftCommand` and `AsyncSwiftCommand` becomes unclear, they essentially duplicate each other with this change.

Sources/CoreCommands/SwiftCommandState.swift
fortmarek: I reverted this, that was overlooked from my side. Thanks for pointing that out.

fortmarek:
Thanks @MaxDesiatov for your feedback, should be resolved now 🙏

MaxDesiatov (approved):
LGTM, but I'd like this to get at least one review from someone on the SwiftPM team before merging.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

plemarquand (approved):
Looks great to me! Just one minor comment, otherwise its good to go.

Sources/CoreCommands/BuildSystemSupport.swift
```diff
@@ -40,7 +40,7 @@ private struct NativeBuildSystemFactory: BuildSystemFactory {
         return try BuildOperation(
             productsBuildParameters: try productsBuildParameters ?? self.swiftCommandState.productsBuildParameters,
             toolsBuildParameters: try toolsBuildParameters ?? self.swiftCommandState.toolsBuildParameters,
-            cacheBuildManifest: cacheBuildManifest && self.swiftCommandState.canUseCachedBuildManifest(),
+            cacheBuildManifest: await self.swiftCommandState.canUseCachedBuildManifest() && cacheBuildManifest,
```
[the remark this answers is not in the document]
fortmarek: This doesn't compile as `await` call can't be on the right side of a `&&` operator. If you do this, you get the following error:
```swift
'async' call in an autoclosure that does not support concurrency
```

I used an `if-else` to skip calling `canUseCachedBuildManifest` when not needed. Alternatively, this could have been kept as a one-line by writing `cacheBuildManifest ? await self.swiftCommandState.canUseCachedBuildManifest() : false` but that felt a bit esoteric, so I opted in the imho more readable `if-else`. lmk what you think.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

plemarquand (approved):

MaxDesiatov:
@swift-ci test windows

plemarquand:
@fortmarek thanks for this great contribution! Looking forward to the improvements it brings.

PR #8227, opened 2025-01-16
Utility: Introduce command repeater script

bkhouri:
Introduce a command repeater script to help run a set of commands repeatedly.

Downside:
 - the script must be manually modified to define the set of commands to be executed


MaxDesiatov (commented):
Is there a particular reason this is written in Bash instead of Python or Swift? What's the plan for running this utility on Windows?

MaxDesiatov:
Additionally, what's the goal of this script in general? Is this supposed to be used on CI or locally? I see it measures completion time, which is not meaningful to measure on CI in the vast majority of cases. Then if it's only for local use, why prefer this instead of existing utilities like `hyperfine`?

bkhouri:
> Additionally, what's the goal of this script in general? Is this supposed to be used on CI or locally? 

I have been using it to determine if my changes introduced intermittent test failures.  I wrote it in bash as that was alot easier to get started.

I originally created this is bash as I was using it for myself for a while and didn't want to copy it from one worktree to the next.  We can investigate converting to Swift or Python at a later date, if it is deems to be useful enough.

bkhouri:
The more I think about it, the more I like the idea of creating a python script.  I'm converting this PR to draft to work on this.

bkhouri:
Script does not belong in SwiftPM. closing PR.

PR #8229, opened 2025-01-17
Use prebuilts for macro tests. Add debug URL for swift-syntax.

dschaefer2:
Re-enable use of prebuilts for tests that depend on macro targets. These are guaranteed to be built for host as the macros are.

Add a hidden option to override the prebuilts download URL for swift-syntax for testing. Also added support for this to be a file URL.

Add support for using the ssh GitHub URL for swift-syntax, allowing multiple PackageReferences for prebuilts. Also elevated the PackageIdentity to the root structures that need it.

Fixed the build prebuilts script to generate into a versioned directory so we can support generating multiple versions at the same time.

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -188,10 +188,19 @@ public struct CachingOptions: ParsableArguments {
     }
 
     /// Whether to use macro prebuilts or not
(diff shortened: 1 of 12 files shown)


dschaefer2:
@swift-ci please test

jakepetroules (approved):

Issue #8216, opened 2025-01-14
`No target named 'swift-openapi-generator-arm64-apple-macosx15.0-debug-tool.exe' in build description` when doing `--experimental-prepare-for-indexing`


ahoppen:
Running `swift build --target HelloWorldHummingbirdServer --experimental-prepare-for-indexing` on the attached project fails with

```
error: No target named 'swift-openapi-generator-arm64-apple-macosx15.0-debug-tool.exe' in build description
```

[testpackage.zip](https://github.com/user-attachments/files/18413483/testpackage.zip)

plemarquand:
I started digging in to this issue, and while I don't know if I've got enough context to fix it completely I can state my findings.

Building with `--experimental-prepare-for-indexing` triggers a `LLBuildManifestBuilder.generatePrepareManifest(at:)`, which generates a stripped down manifest. This manifest generates compile commands and _only_ creates `.macro` and `.plugin` product commands. 

https://github.com/swiftlang/swift-package-manager/blob/1d637b38d4526f13d8995b2669393d75953ea9e6/Sources/Build/BuildManifest/LLBuildManifestBuilder.swift#L151-L159

Because `swift-openapi-generator-arm64-apple-macosx15.0-debug-tool` is an [executable](https://github.com/apple/swift-openapi-generator/blob/84b693f9d0559dc488e691edb4837bafbce2aaea/Package.swift#L121) that is depended on by the [OpenAPIGenerator plugin target](https://github.com/apple/swift-openapi-generator/blob/84b693f9d0559dc488e691edb4837bafbce2aaea/Package.swift#L130), the product command for the plugin is added to the manifest but the executable that it depends on is not. If I comment out the `switch` and simply do a `createProductCommand` for all descriptions in the `self.plan.productMap` the build succeeds.

Obviously its not ideal, but I'm not sure the best approach to pair down the products while still building what is required.



The other, much more minor thing is the `.exe` in the target name on all platforms (i.e. macOS). It doesn't really break anything but is pretty confusing. That looks like a simple omission in the function that generates the executable product name:

https://github.com/swiftlang/swift-package-manager/blob/1d637b38d4526f13d8995b2669393d75953ea9e6/Sources/Build/BuildManifest/LLBuildManifestBuilder%2BProduct.swift#L149-L154

Needs to take in to account the platform, i.e:

```swift
fileprivate func executableName(
    for product: String,
    buildParameters: BuildParameters
) -> String {
[...]

PR #8230, opened 2025-01-17
Fix prepare for indexing so plugin tools are properly built.

dschaefer2:
Eliminate the special generatePrepareManifest and just reuse the real one and add a couple of checks for the target/product prepareForIndexing buildParameter instead. This then enables us to make sure we're doing a full build of the plugin tools so they can generate code to be properly indexed.

Addresses #8216 for openapi generator.

### DIFF
--- a/Sources/Build/BuildManifest/LLBuildManifestBuilder.swift
+++ b/Sources/Build/BuildManifest/LLBuildManifestBuilder.swift
@@ -107,54 +107,26 @@ public class LLBuildManifestBuilder {
                 case .swift(let desc):
                     try self.createSwiftCompileCommand(desc)
                 case .clang(let desc):
-                    try self.createClangCompileCommand(desc)
+                    if desc.buildParameters.prepareForIndexing == .off {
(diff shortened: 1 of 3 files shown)


dschaefer2:
@swift-ci please test

jakepetroules (approved):

MahdiBM:
I think and hope this PR related to this issue in this comment: https://github.com/swiftlang/sourcekit-lsp/issues/1896#issuecomment-2599592492

dschaefer2:
> I think and hope this PR related to this issue in this comment: [swiftlang/sourcekit-lsp#1896 (comment)](https://github.com/swiftlang/sourcekit-lsp/issues/1896#issuecomment-2599592492) ?

Not sure it does. The focus here is to make sure we run the plugin tools so that the code the may generate gets indexed.

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test windows

PR #8231, opened 2025-01-17
Basics,_AsyncFileSystem: internalise SystemPackage

compnerd:
`SystemPackage` is only used as an internal implementation detail here. This package is not needed for the public API surface. Furthermore, as this dependency is a single use dependency, we are able to save some nominal space on the builds with this:

Pre:
SystemPackage.dll: 212992 bytes
_AsyncFileSystem.dll: 60416 bytes

Post:
SystemPackage.dll: 0 bytes (static)
_AsyncFileSystem.dll: 217600

This effectively shaves ~56K off the toolchain.


compnerd:
@swift-ci please test

jakepetroules (approved):

Issue #8233, opened 2025-01-21
Swift package manager is not working with 'git lfs' in Xcode.


ankushkushwaha:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

Hi,

I am created a library, which has dependency in heremap sdk. To use hereSDK I need to use git lfs. But When I try to install the swift package to integrate my library, swift package manager shows error. 

Is the git lfs does not work with swift package manager ? 

### Expected behavior

Swift package manager should integrate library with all dependencies (sub modules)

robmaceachern:
We've been suffering with spm/gitlfs issues for a while too and it'd be great if this could be improved. In our case the only usage of lfs is for snapshot test reference images.

The sample project is really helpful @plemarquand. Here is the verbose output of the failure:

```
~/Downloads/GitLFS_SampleProject swift package resolve --verbose
Running resolver because the following dependencies were added: 'lib-here-map-package' (https://github.com/ankushkushwaha/lib-here-map-package.git)
Updating https://github.com/ankushkushwaha/lib-here-map-package.git
Updated https://github.com/ankushkushwaha/lib-here-map-package.git (0.32s)
Creating working copy for https://github.com/ankushkushwaha/lib-here-map-package.git
error: 'lib-here-map-package': Couldn’t check out revision ‘e048e833d808ad8ed2c7e430da60c0ff3e7dbb3e’:
    Downloading Frameworks/heresdk.xcframework/Info.plist (1.1 KB)
    Error downloading object: Frameworks/heresdk.xcframework/Info.plist (2f567e3): Smudge error: Error downloading Frameworks/heresdk.xcframework/Info.plist (2f567e30dd98cc4be512f456f9b1440a0381ce45f3816437221c5f1ffd4b04f2): error transferring "2f567e30dd98cc4be512f456f9b1440a0381ce45f3816437221c5f1ffd4b04f2": [0] remote missing object 2f567e30dd98cc4be512f456f9b1440a0381ce45f3816437221c5f1ffd4b04f2
    Errors logged to '/Users/robmaceachern/Downloads/GitLFS_SampleProject/.build/checkouts/lib-here-map-package/.git/lfs/logs/20250214T163019.148823.log'.
[...]

PR #8235, opened 2025-01-21
Fetch Git LFS files after clone

plemarquand:
### Motivation:

Currently SPM clones repositories with `--mirror`, creating a bare repository. If the cloned repository uses git lfs, the files in lfs are never pulled and so the package resolution ultimately fails.

### Modifications:

This is a naive implementation of LFS support that simply does a `lfs fetch` on the mirrored repository, without regard for whether there are LFS files to clone or even if the user has git lfs installed. It does, however, allow packages using LFS to resolve correctly.

Issue: #8233

### DIFF
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -201,6 +201,24 @@ public struct GitRepositoryProvider: RepositoryProvider, Cancellable {
             ["--mirror"],
             progress: progressHandler
         )
+
+        do {
+            try self.callGit(
+                [
+                    "-C",
+                    path.pathString,
+                    "lfs",
+                    "fetch"
+                ],
+                repository: repository,
+                failureMessage: "Failed to fetch Git LFS files for \(repository.location)"
+            )
+        } catch {
+            // Ignore the error if lfs fetch fails, likely the repository either isn't using
+            // git lfs or it isn't installed.
+
+            // TODO: If the repo _is_ using git lfs and it isn't installed can we show a nice error?
+        }
     }
 
     public func isValidDirectory(_ directory: Basics.AbsolutePath) throws -> Bool {


dschaefer2:
The main question I have is that we do two clones. One into the cache and then another into the scratch dir (.build). Can we git-lfs through them both?

plemarquand:
Once the files have been fetched in to the bare repo (the cached folder) they're treated as regular files in the subsequent clone in to the .build folder, so we should only need to work with git lfs the one time.

sbelbin:
I'd appreciate to learn about the state/progress of this PR.

We've been relying on Git LFS for our development cycle, since we rely on storing large binary files that are generated.

This PR is a step forward to address this limitation within Swift Package Manager & progress in the review of this PR is appreciated by myself and I'd suspect by other developers as well.

plemarquand:
@sbelbin I'm interested in gathering more example usages for git-lfs in the package ecosystem so we can make sure this solution covers all the usecases. Do you have an example package or usecase where lfs would be necessary?

sbelbin:
> @sbelbin I'm interested in gathering more example usages for git-lfs in the package ecosystem so we can make sure this solution covers all the usecases. Do you have an example package or usecase where lfs would be necessary?

A simple use case that I should be able to spin up (hopefull by Thursday - Eastern time zone) is one that contains a folder with prebuilt libraries of various open source libraries such as SuiteSparse, OpenMP, GMP, & MPFR for different Apple platforms on which our ML libraries depend upon.

P.S. I've seen the commits that were done earlier today and the approach that you've taken of making it configurable opt-in choice seems reasonable to me.

sbelbin:
@plemarquand 
You can experiment with this Git repositories.
Swift Package: https://github.com/sbelbin/sample_lfs
Sampler App: https://github.com/sbelbin/sampler

It's Swift Package contains open source libraries that are built externally as to allow our colleagues to incorporate into their Swift Package Manager.

The Swift Package contains C function called `compute` having a modulemap as to export that C function.

At this moment the `sampler` Swift App has a local dependency on the `sample_lfs` but the goal with your modifications to Swift Package Manager is use Git repository URL.

plemarquand:
Closing in favour of #9625

PR #8236, opened 2025-01-21
Fix prepare for indexing so plugin tools are properly built. (#8230)

dschaefer2:
Back port of #8230 to release/6.1

Eliminate the special generatePrepareManifest and just reuse the real one and add a couple of checks for the target/product prepareForIndexing buildParameter instead. This then enables us to make sure we're doing a full build of the plugin tools so they can generate code to be properly indexed.

Addresses #8216 for openapi generator.


dschaefer2:
@swift-ci please test

bkhouri (approved):
Looks like a clean cherry-pick.  I quickly scanned the diff against #8230 and nothing stood out as being completely different.

jakepetroules (approved):

PR #8237, opened 2025-01-21
Basics: internalise `Deque` usage

compnerd:
This module is not part of the public interface. Mark this is as an `internal` import which helps reduce the spread of this dependency into SourceKit-LSP.


compnerd:
@swift-ci please test

PR #8240, opened 2025-01-22
SourceKitLSPAPI: export library as an interface

compnerd:
This is consumed by SourceKit-LSP, so export the library target to allow the library to be consumed properly to manage dependencies and flags.


compnerd:
@swift-ci please test

PR #8232, opened 2025-01-20
Clean up warnings, fix handling of dep version changes

dschaefer2:
Assume dependent products from prebuilts are being used.

PR #8243, opened 2025-01-23
Forward port fixes for macro build prebuilts

dschaefer2:
Forward port of macro build prebuilts fixes from release/6.1.

#8229

Re-enable use of prebuilts for tests that depend on macro targets. These are guaranteed to be built for host as the macros are.

Add a hidden option to override the prebuilts download URL for swift-syntax for testing. Also added support for this to be a file URL.

Add support for using the ssh GitHub URL for swift-syntax, allowing multiple PackageReferences for prebuilts. Also elevated the PackageIdentity to the root structures that need it.

Fixed the build prebuilts script to generate into a versioned directory so we can support generating multiple versions at the same time.

 #8232 

I found a bug where if you change swift syntax version in your manifest, we were not changing the use of prebuilts to match. I have added a scan to do so. Also requires that we add the version of the prebuilt to the workspace state. Assume dependent products from prebuilts are being used to remove the warning for swift-syntax.

Fixes warning about swift-syntax not being used when all uses were replaced by the prebuilts.

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -188,10 +188,19 @@ public struct CachingOptions: ParsableArguments {
     }
 
     /// Whether to use macro prebuilts or not
-    @Flag(name: .customLong("experimental-prebuilts"),
-          inversion: .prefixedEnableDisable,
-          help: "Whether to use prebuilt swift-syntax libraries for macros")
+    @Flag(
+        name: .customLong("experimental-prebuilts"),
+        inversion: .prefixedEnableDisable,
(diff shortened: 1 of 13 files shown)


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please smoke test

dschaefer2:
@swift-ci Please smoke test macOS platform

dschaefer2:
@swift-ci please smoke test

dschaefer2:
Will do another cut of this once I get a final fix into 6.1.1

PR #8244, opened 2025-01-24
Basics,Commands,PackageLoading,SPMBuildCore: internalize Collections

compnerd:
These modules do not require Collections to leak into the public interface. Use the opportunity to reduce the publicly visible dependencies and reduce some unnecessary imports of `OrderedCollections`.


compnerd:
@swift-ci please test

jakepetroules (approved):

PR #8247, opened 2025-01-24
Utilities: remove Yams from the bootstrap script

compnerd:
Yams is no longer a dependency for swift-driver, remove the dependency building to reduce the build times.


compnerd:
@swift-ci please test

owenv (approved):

PR #8250, opened 2025-01-26
Fix: preserve symlinks in ZipArchive compress function

### LINKED ISSUE #8248: swift package-registry publish incorrectly handles packages containing symlinks

### Description

I'm trying to push certain open-source packages to a Swift Package Registry, but am having an unexpected issue when trying to publish the realm-swift library. Within the realm-swift sources, there's [a directory containing a large number of symlinks](https://github.com/realm/realm-swift/tree/master/include/Realm). However, when I run `swift package-registry publish realm.realm-swift {version}`, the resulting archive is not buildable when unarchived.

### Expected behavior

I would expect that after publishing a package such as realm-swift to my registry, I should be able to resolve and build it successfully. 

### Actual behavior

In reality, it appears that the archiving logic does not preserve symlinks in package sources, but instead overwrites them with copies of the referenced file, which is causing errors such as `duplicate interface definition for class` and `property has a previous declaration` when trying to resolve and build the realm-swift package from the registry.

### Steps to reproduce

To reproduce this issue, you can clone the realm-swift repository, and use `swift package-registry publish` to push it to a Swift registry. Unzipping the resulting archive will show that the symlinks are not preserved as would be needed for proper usability of the package. 

Swift Package Manager - Swift 6.0.2-dev

swift-driver version: 1.115 Apple Swift version 6.0.2 (swiftlang-6.0.2.1.2 clang-1600.0.26.4)
Darwin [hostname redacted] 23.6.0 Darwin Kernel Version 23.6.0: Wed Jul 31 20:49:39 PDT 2024; root:xnu-10063.141.1.700.5~1/RELEASE_ARM64_T6000 arm64

sphanley:
This PR fixes #8248, by adding the `y` flag to the `zip` command that is executed when archiving a package's sources for publication to a registry on a non-Windows OS.

### Motivation:

Without this change, packages which contain symlinks in their sources will produce invalid and potentially non-buildable archives when publishing to a registry, since any symlinks are currently transformed into duplicate files during the archive creation, instead of being preserved.  

### Modifications:

I've added the following flag to the `zip` call:
>   -y   store symbolic links as the link instead of the referenced file

I've additionally modified the associated `testCompress()` test case, adding logic to ensure that symlinks are preserved correctly. 

### Result:

With this change, the resulting zipfile that is uploaded to a registry will contain symlinks consistent with the original package sources.

### DIFF
--- a/Sources/Basics/Archiver/ZipArchiver.swift
+++ b/Sources/Basics/Archiver/ZipArchiver.swift
@@ -120,7 +120,7 @@ public struct ZipArchiver: Archiver, Cancellable {
             arguments: [
                 "/bin/sh",
                 "-c",
-                "cd \(directory.parentDirectory.underlying.pathString) && zip -r \(destinationPath.pathString) \(directory.basename)",
+                "cd \(directory.parentDirectory.underlying.pathString) && zip -ry \(destinationPath.pathString) \(directory.basename)",
             ]
         )
         #endif
--- a/Tests/BasicsTests/Archiver/ZipArchiverTests.swift
+++ b/Tests/BasicsTests/Archiver/ZipArchiverTests.swift
@@ -119,6 +119,8 @@ final class ZipArchiverTests: XCTestCase {
              try localFileSystem.writeFileContents(dir2.appending("file3.txt"), string: "Hello World 3!")
              try localFileSystem.writeFileContents(dir2.appending("file4.txt"), string: "Hello World 4!")
 
+             try localFileSystem.createSymbolicLink(rootDir.appending("file2.txt"), pointingAt: dir1.appending("file2.txt"), relative: true)
+
              let archivePath = tmpdir.appending(component: UUID().uuidString + ".zip")
              try await archiver.compress(directory: rootDir, to: archivePath)
              XCTAssertFileExists(archivePath)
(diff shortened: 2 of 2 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):

sphanley:
Thanks for the review, @MaxDesiatov! Apologies for the failing test on Linux, I'll try to dig into what's going on there and make sure this works cross-platform.

sphanley:
@MaxDesiatov, I believe I've got this working– there's an issue that seems specific to Ubuntu 20.04 when the symlink in the unit test is at the root of the test's temp directory, but moving it to a subdirectory seems to behave correctly across MacOS and Linux and provides the same coverage for the functionality.

MaxDesiatov:
@swift-ci test

sphanley:
I hate to be a bother, but any chance someone can re-invoke the tests? Looks like the Windows job never reported its status, but I believe the tests should have passed.

jakepetroules:
@swift-ci test Windows

jakepetroules (approved):

dschaefer2:
Thanks!

Issue #8180, opened 2024-12-18
PackageCommandTests.testCommandPluginDiagnostics is flakey


plemarquand:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [X] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

Periodically `PackageCommandTests.testCommandPluginDiagnostics` will fail. This test has been temporarily disabled. This issue tracks the work to reenable it.

It is failing with:
```
Test Case 'PackageCommandTests.testCommandPluginDiagnostics' started at 2024-12-18 13:15:21.289
/home/build-user/swiftpm/Tests/CommandsTests/PackageCommandTests.swift:2447: error: PackageCommandTests.testCommandPluginDiagnostics : XCTAssertTrue failed - unexpected failure matching '' against pattern equal("command plugin: print\n")
Test Case 'PackageCommandTests.testCommandPluginDiagnostics' failed (32.126 seconds)
```

https://ci.swift.org/job/oss-swift-package-debian-12/708/

PR #8251, opened 2025-01-26
[FreeBSD] suspend all other threads before closing file descriptors

michael-yuji:
Suspend all other threads first before closing file descriptors and exec.

### Motivation:

See #8197 #8180, `swift run` can crash due to race condition between libdispatch controlled threads. To properly mitigate this, suspend all other threads (including those managed by libdispatch) before closing the file descriptors.

### Modifications:

Call `pthread_suspend_all_np` before closing file descriptors, this api is not portable but exists on both FreeBSD and OpenBSD. However, since the Glibc overlay for OpenBSD does not include <pthread_np.h>, we guard this and make it available for FreeBSD only.


michael-yuji:
@swift-ci please test

jakepetroules (approved):

PR #8252, opened 2025-01-27
Apply toolset's `debugger` property in `swift run`

MaxDesiatov:
### Motivation:

Per [SE-0378 toolsets can specify an optional `debugger` property](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0387-cross-compilation-destinations.md#toolsetjson-files), which so far had no effect. We should respect this property in `swift run` invocations when any toolsets or Swift SDKs are selected.

This allows running and debugging cross-compiled products in an environment other than the host, e.g. a container runtime when cross-compiled to Linux, QEMU or a firmware flasher and serial port setup script with Swift Embedded for microcontrollers, or in a Wasm runtime for a Wasm binary.

### Modifications:

Updated `SwiftRunCommand` to check for the `debugger` property in toolsets or Swift SDKs when such are specified. Its value is then used to launch a freshly built binary.

### Result:

Cross-compiled binaries can be directly launched and debugged with a `swift run` invocation.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Tests/CommandsTests/RunCommandTests.swift
```diff
@@ -40,6 +40,25 @@ final class RunCommandTests: CommandsTestCase {
         XCTAssert(stdout.contains("Swift Package Manager"), "got stdout:\n" + stdout)
     }
 
+// echo.sh script from the toolset won't work on Windows
+#if !os(Windows)
```
bkhouri, line 44: suggestion: can we create an equivalent target that will run on Windows to ensure the functions work on that host platform?

kateinoigakukun (approved):

Tests/CommandsTests/RunCommandTests.swift
MaxDesiatov: I'm not sure, I don't have experience with Windows and don't have means to test on Windows. IIRC the test suite doesn't even run on the Windows CI job. I'd prefer to do this in a separate PR.

PR #8261, opened 2025-01-30
Add `package` access level to access level enums

daniel-grumberg:
Currently the symbol graph generator defaults to "public" minimum access level even if the user tries to initialize the options with "package" this is because the internal enums representing access level don't have a notion of `package` access control.

### Motivation:

Currently it is not possible to generate documentation only for `package` and `public` symbols. This changes that

### Modifications:

Adds `package` values to enums representing access control along the symbol graph generation code path.

### Result:

When invoking [swift-docc-plugin](https://github.com/apple/swift-docc-plugin) with `--symbol-graph-minimum-access-level package` the resulting documentation contains `public` and `package` access controlled symbols.

### DIFF
--- a/Sources/Commands/Utilities/PluginDelegate.swift
+++ b/Sources/Commands/Utilities/PluginDelegate.swift
@@ -442,6 +442,8 @@ final class PluginDelegate: PluginInvocationDelegate {
             symbolGraphExtractor.minimumAccessLevel = .fileprivate
         case .internal:
             symbolGraphExtractor.minimumAccessLevel = .internal
+        case .package:
+            symbolGraphExtractor.minimumAccessLevel = .package
         case .public:
             symbolGraphExtractor.minimumAccessLevel = .public
         case .open:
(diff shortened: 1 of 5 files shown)


daniel-grumberg:
@swift-ci please test

MaxDesiatov (commented):
Are there no test cases covering this code? If there are, would you mind extending them to cover the newly added access level?

daniel-grumberg:
The closest I could find was https://github.com/swiftlang/swift-package-manager/blob/4878aba305087b5d5daad6dd9c161976cd9b4efb/Tests/CommandsTests/PackageCommandTests.swift#L2979 which isn't easy to modify to test this behavior.

daniel-grumberg:
Nonetheless, the problem is that the enums aren't kept in-sync with the language, arguably we could just forward the raw string along to not have to worry about this going forward.

plemarquand (approved):

PR #8262, opened 2025-01-30
[6.1] Handle paginated registry metadata responses (#8219)

plemarquand:
In [4.1 List Package
Releases](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/Registry.md#41-list-package-releases) it is stated that a server may respond with a `Link` header that contains a pointer to a subsequent page of results. SwiftPM is not checking for the `next` link in the `Link` header and so if a registry returns paginated results, only the first page of versions is searched when resolving. This would result in a "version not found" error when in reality the version is present in a subsequent page of results.

Respect the `next` link in the `Link` header by loading the next page of results and building up a list of versions, continuing until there is no `next` link present in the `Link` header of the last result.

Registry servers that serve paginated results now have all their results read.

Issue: #8215

### DIFF
--- a/Sources/PackageMetadata/PackageMetadata.swift
+++ b/Sources/PackageMetadata/PackageMetadata.swift
@@ -352,7 +352,7 @@ public struct PackageSearchClient {
         ) { result in
             do {
                 let metadata = try result.get()
-                let alternateLocations = metadata.alternateLocations ?? []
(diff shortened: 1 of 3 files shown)


plemarquand:
@swift-ci test

jakepetroules (approved):

PR #8263, opened 2025-01-30
Deprecate sync methods in RegistryClient

plemarquand:
Deprecate the synchronous methods in RegistryClient and move the logic from then sync versions in to the async versions. This async-ifys the previously callback based methods, flattening them out.

The synchronous versions of these methods will now warn when they're used. After some time they can be removed completely.

### DIFF
--- a/Sources/Basics/HTTPClient/LegacyHTTPClient.swift
+++ b/Sources/Basics/HTTPClient/LegacyHTTPClient.swift
@@ -363,25 +363,6 @@ extension LegacyHTTPClient {
         body: Data?,
         headers: HTTPClientHeaders = .init(),
         options: Request.Options = .init(),
-        observabilityScope: ObservabilityScope? = .none
-    ) async throws -> Response {
-        try await withCheckedThrowingContinuation { continuation in
-            self.put(
-                url,
-                body: body,
-                headers: headers,
-                options: options,
-                observabilityScope: observabilityScope,
-                completion: { continuation.resume(with: $0) }
-            )
-        }
-    }
-    @available(*, noasync, message: "Use the async alternative")
-    public func put(
-        _ url: URL,
-        body: Data?,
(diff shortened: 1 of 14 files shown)


Sources/Basics/HTTPClient/LegacyHTTPClient.swift
```diff
@@ -358,6 +358,25 @@ extension LegacyHTTPClient {
...
 
+    public func put(
+        _ url: URL,
+        body: Data?,
+        headers: HTTPClientHeaders = .init(),
+        options: Request.Options = .init(),
+        observabilityScope: ObservabilityScope? = .none
+    ) async throws -> Response {
+        try await withCheckedThrowingContinuation { continuation in
+            self.put(
```
jakepetroules, line 369: I suggest having this call `execute` directly, and switching the completion handler versions to call the async versions. That way when it comes time to remove the completion handler overloads it's just a plain delete.

plemarquand:
@swift-ci test

Sources/Basics/HTTPClient/LegacyHTTPClient.swift
plemarquand: Instead of building out these missing pieces in LegacyHTTPClient I switched the RegistryClient over to use the newer and asyc-ier HttpClient.

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test windows

jakepetroules (approved):

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci please test windows

PR #8271, opened 2025-02-01
Create an alternate SwiftPM build system called Swift Build

cmcgee1024:
Add basic test coverage for the new build system for the supported platforms: Linux, macOS, and Windows. Provide basic progress reporting, and an option for "very verbose" mode so that that individual build tasks and the arguments can be inspected for correctness.

Build a package with the new build system like this:

    swift build --build-system=swiftbuild

Enable very verbose task outputs with their flags, options, and arguments:

    swift build --build-system=swiftbuild --vv


jakepetroules:
@swift-ci please test

jakepetroules (approved):

rauhul:
Not a review comment, but I can't wait for this to land :D this is awesome!

cmcgee1024:
@swift-ci please test

dschaefer2:
I just tried this out and get:

`error: Dependencies could not be resolved because root depends on 'swift-argument-parser' 1.4.0..<1.5.0 and 'swift-driver' depends on 'swift-argument-parser' 1.2.2..<1.3.0.`

dschaefer2:
Never mind, had a really old version of the deps. Did a `swift package update` and it's going now.

dschaefer2:
Build works on my Mac. A few weird warnings. Also looks like swift-test has hardcoded the location of the xctest bundle to the old scratch dir (.build) layout. Lots of work to do indeed :).

Thanks @cmcgee1024. This is a great start!

dschaefer2 (approved):

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

jakepetroules:
@swift-ci please test Windows

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test Windows

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/swiftlang/swift/pull/79106

@swift-ci please test

bkhouri (commented):
I have not reviewed this extensively, however there are a few minor comments.

cmcgee1024:
@swift-ci please test

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -0,0 +1,537 @@
...
+                        {
+                            true
+                        } else {
+                            false
+                        }
+
+                        if report {
+                            progressAnimation.complete(success: false)
+                            // FIXME: a whole range of pretty printed location, context, and fixups
+                            print("\(info.kind): \(info.location) \(info.message) \(info.fixIts)")
```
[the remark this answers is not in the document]
MaxDesiatov: I agree, we should avoid using print and rely on existing logging facilities instead, especially since those can propagate diagnostic information like message level and location already.

Sources/XCBuildSupport/XcodeBuildSystem.swift
```diff
@@ -230,6 +230,8 @@ public final class XcodeBuildSystem: SPMBuildCore.BuildSystem {
 
             throw Diagnostics.fatalError
         }
+
+        print("Build complete!")
```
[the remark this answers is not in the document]
MaxDesiatov: This should use existing logging facilities that respect `-v`/`-q` settings.

MaxDesiatov (commented):
I am concerned by uses of `print` here. We have commands like `swift-run` that are supposed to only print freshly-built product's output. The rest of diagnostic messages even successful ones like "Build complete" should go to stderr, since there is an existing body of scripts and utilities that read from `swift-run` invocations stdout via a pipe. Stray `print` polluting such output with diagnostics will break those scripts.

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -412,23 +412,43 @@ final class BuildCommandTests: CommandsTestCase {
...
             // try await building using XCBuild with default parameters.  This should succeed.  We build verbosely so we get
             // full command lines.
-            let defaultOutput = try await execute(["-c", "debug", "-v"], packagePath: fixturePath).stdout
+            let defaultOutput = try await execute(["--build-system", buildsystem, "-c", "debug", "-v"], packagePath: fixturePath).stdout
 
             // Look for certain things in the output from XCBuild.
             XCTAssertMatch(
                 defaultOutput,
-                try .contains("-target \(UserToolchain.default.targetTriple.tripleString(forPlatformVersion: ""))")
+                try .contains("Build complete!")
```
bkhouri, line 424: issue (possibly-blocking): Since the native build system was outputting the target triple when verbose was set, should we ensure we are still "showing" this target triple with the Swift Build build system?

MaxDesiatov: +1 here, unclear why the old expectation was removed

Sources/XCBuildSupport/XcodeBuildSystem.swift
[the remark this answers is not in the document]
cmcgee1024: I've converted this over to use an observability scope emit. One problem though, is that we probably want this output to be emitted in all log levels, but I don't want to indicate that it is an error.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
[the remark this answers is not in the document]
cmcgee1024: I've converted the print calls to observability scopes. See my note above about what to do about certain kinds of messages like "Build complete!"

Tests/CommandsTests/BuildCommandTests.swift
MaxDesiatov: `-target <triple>` string is supposed to be emitted with all build systems, if it's absent, I would assume Swift Build doesn't support cross-compilation?

Either way, removing this expectation, we can regress with cross-compilation in other build systems and won't notice it. If it's currently not expected that Swift Build can cross-compile, the expectation should be excluded for Swift Build, but stay for everything else to prevent regressions.

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -412,23 +412,43 @@ final class BuildCommandTests: CommandsTestCase {
         }
     }
 
-    func testXcodeBuildSystemDefaultSettings() async throws {
-        #if !os(macOS)
-        try XCTSkipIf(true, "test requires `xcbuild` and is therefore only supported on macOS")
-        #endif
+    private func testBuildSystemDefaultSettings(buildsystem: String) async throws {
```
MaxDesiatov, line 415: nit: inconsistent capitalization
```suggestion
    private func testBuildSystemDefaultSettings(buildSystem: String) async throws {
```

MaxDesiatov (changes requested):
IMO removing existing tests for cross-compilation for existing build systems is a significant regression in the test suite. I'm totally ok with merging Swift Build support if it can't cross-compile yet, but it shouldn't come at the expense of reducing the test coverage for other build systems.

cmcgee1024:
@swift-ci please test

cmcgee1024:
> IMO removing existing expectations for cross-compilation for existing build systems is a significant regression in the test suite. I'm totally ok with merging Swift Build support if it can't cross-compile yet, but it shouldn't come at the expense of reducing the test coverage for other build systems.

This is an issue with the CI infrastructure using an old Xcode that can't handle Swift 6, which is required by the SwiftBuild code. I will ask about getting the CI updated with a newer version.

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test windows

Tests/CommandsTests/BuildCommandTests.swift
MaxDesiatov: If that was the motivation for the change, then the check should be moved to the appropriate business layer, not removed completely.

cmcgee1024: I'll restore this check for the `-target` in the output in the case of the XCBuild (aka. Xcode) build system so that we don't lose this part of the verification.

MaxDesiatov: Why not llbuild though? I haven't seen Swift SDKs used in the wild with XCBuild, users rely on the llbuild as the default CLI build system, so what would be the reason for regressing existing llbuild test coverage?

dschaefer2:
We have switched focus to fix the CI so we have the right toolchain lineup.

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test windows

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -418,7 +418,15 @@ final class BuildCommandTests: CommandsTestCase {
             // full command lines.
             let output = try await execute(["--build-system", buildSystem, "-c", "debug", "-v"], packagePath: fixturePath)
 
-            // Look for certain things in the output from XCBuild.
+            // In the case of XCBuild (ie. xcode build system) check for the cross-compile target
+            if buildSystem == "xcode" {
```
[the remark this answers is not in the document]
MaxDesiatov: +1 here, also highlighted above: Swift SDKs for cross-compilation are used with llbuild, not XCBuild.

Tests/CommandsTests/BuildCommandTests.swift
cmcgee1024: I was confused by the comment that was there. Changing it over to the native build system, and also the comment.

Tests/CommandsTests/BuildCommandTests.swift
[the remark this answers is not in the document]
cmcgee1024: This should be changed back to the native build system. The buildSystem parameter is a string, so there is the string comparison.

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/322

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/322

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/322

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/322

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/322

@swift-ci please test

cmcgee1024:
@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/322

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/322

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/322

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/322

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/323

@swift-ci please test

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -412,23 +412,53 @@ final class BuildCommandTests: CommandsTestCase {
...
+    private func testBuildSystemDefaultSettings(buildSystem: String) async throws {
         try await fixture(name: "ValidLayouts/SingleModule/ExecutableNew") { fixturePath in
             // try await building using XCBuild with default parameters.  This should succeed.  We build verbosely so we get
             // full command lines.
-            let defaultOutput = try await execute(["-c", "debug", "-v"], packagePath: fixturePath).stdout
+            let output = try await execute(["--build-system", buildSystem, "-c", "debug", "-v"], packagePath: fixturePath)
 
-            // Look for certain things in the output from XCBuild.
+            // In the case of the native build system check for the cross-compile target, only for macOS
+#if os(macOS)
```
MaxDesiatov, line 422: This indicates in the change of behavior in llbuild, the `-target` option was previously passed on all platforms, not just macOS. What's the reason for the change then?

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/324

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/324

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/325

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/325

@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/325

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci please test

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326
https://github.com/apple/swift-asn1/pull/88

@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326
https://github.com/apple/swift-asn1/pull/88

@swift-ci test windows

cmcgee1024:
Please test with following pull request:
https://github.com/apple/swift-crypto/pull/326

@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

dschaefer2:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test windows

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test windows

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test windows

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test windows

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test windows

dschaefer2:
DriverSupport needs swift-driver always independent of swift-build.

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test windows

Sources/PackageGraph/ResolvedPackagesStore.swift
```diff
@@ -560,7 +560,12 @@ extension ResolvedPackagesStore.ResolvedPackage {
         let location = mirrors.effective(for: pin.location)
         switch pin.kind {
         case .localSourceControl:
-            packageRef = try .localSourceControl(identity: identity, path: AbsolutePath(validating: location))
+            // FIXME investigate why local source control packages resolve to a non-local URL location
```
owenv, line 563: Is this change intended to be part of the PR?

owenv (approved):

Package.swift
```diff
@@ -989,24 +1000,48 @@ if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
...
         .package(path: "../swift-system"),
         .package(path: "../swift-collections"),
         .package(path: "../swift-certificates"),
         .package(path: "../swift-toolchain-sqlite"),
+        .package(path: "../swift-driver"),
+    ]
+}
+
+if ProcessInfo.processInfo.environment["SWIFTPM_SWBUILD_FRAMEWORK"] == nil &&
+    ProcessInfo.processInfo.environment["SWIFTPM_NO_SWBUILD_DEPENDENCY"] == nil {
```
dschaefer2, line 1026: Really confused by having both of these here. It should always build with the dependency unless NO_SWBUILD is set.

owenv: SWIFTPM_SWBUILD_FRAMEWORK is used to link a prebuilt copy of Swift Build which is assumed to already exist outside the usual SwiftPM dependency resolution process.

cmcgee1024:
@swift-ci please test

Package.swift
cmcgee1024: Agreed. There are these two cases where we don't include the swift-build dependency:

1. We're building with it as a common framework, which will bring it in via the `# if canImport()` check
2. The dependency is explicitly removed by the bootstrap script

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test windows

Package.swift
```diff
@@ -989,24 +1000,48 @@ if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
...
         // dependency version changes here with those projects.
         .package(url: "https://github.com/apple/swift-argument-parser.git", .upToNextMinor(from: "1.4.0")),
-        .package(url: "https://github.com/apple/swift-driver.git", branch: relatedDependenciesBranch),
         .package(url: "https://github.com/apple/swift-crypto.git", .upToNextMinor(from: "3.0.0")),
         .package(url: "https://github.com/swiftlang/swift-syntax.git", branch: relatedDependenciesBranch),
         .package(url: "https://github.com/apple/swift-system.git", from: "1.1.1"),
         .package(url: "https://github.com/apple/swift-collections.git", "1.0.1" ..< "1.2.0"),
         .package(url: "https://github.com/apple/swift-certificates.git", "1.0.1" ..< "1.6.0"),
         .package(url: "https://github.com/swiftlang/swift-toolchain-sqlite.git", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-driver.git", branch: relatedDependenciesBranch),
```
dschaefer2, line 1009: This should come from swiftlang.

Package.swift
dschaefer2: Ah, get it.

cmcgee1024:
@swift-ci test windows

cmcgee1024:
@swift-ci test macOS

owenv:
@swift-ci build toolchain

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test windows

Utilities/build-using-self
```diff
@@ -138,6 +138,9 @@ def main() -> None:
             shlex.split("swift --version"),
         )
 
+        call(
```
bkhouri, line 141: Out of curiosity, why was this added it?  Are previous CI builds impacting the current PR build?

cmcgee1024: This was added because the package may have been built earlier using the bootstrap with the environment variable set to exclude swift-build as a dependency. This is one of the issues with using environment variables in packages. If you flip them then SwiftPM may not be aware of the change and produce inconsistent build results since its build graph is not obviously out of sync with the current package state.

I believe that adding this will force the package to be entirely reset, and cleaned before building it again with swift-build support.

Sources/PackageGraph/ResolvedPackagesStore.swift
cmcgee1024: This could be an artifact of one of the previous approaches to getting the CI working. I will try disable it and see if everything continues to succeed.

cmcgee1024:
@swift-ci please test

Package.swift
cmcgee1024: I've basically undone this diff, so it is the same as what is currently in `main`.

cmcgee1024:
@swift-ci test windows

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test windows

cmcgee1024:
@swift-ci test windows

cmcgee1024:
@swift-ci test windows

cmcgee1024:
@swift-ci test windows

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test windows

cmcgee1024:
@swift-ci test Linux

cmcgee1024:
@swift-ci test Windows

Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
```diff
@@ -172,6 +173,7 @@ extension BuildSystemProvider.Kind {
     public var usesXcodeBuildEngine: Bool {
         switch self {
             case .native: return false
+            case .swiftbuild: return false
```
bkhouri, line 176: Chore (blocking): this should return "true"

Sources/SwiftBuildSupport/BuildSystem.swift
```diff
@@ -0,0 +1,26 @@
+import SPMBuildCore
+import PackageModel
+
+extension BuildConfiguration {
+    public var swiftbuildName: String {
+        switch self {
+        case .debug: "Debug"
```
bkhouri, line 7: Question: can we have the build configuration confirm to "ConvertibleString" ?

bkhouri (changes requested):

Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
cmcgee1024: SwiftBuild isn't using any Xcode or its build engine. It's unclear what this check is for.

dschaefer2: SwiftBuild doesn't use the same directory structure.

PR #8278, opened 2025-02-04
[WIP] Provide mechanism to disable color diagnostics

johnbute:
Added the no-color-diagnostics option to disable colorful diagnostics
### Motivation:

Users prefer disabling coloured diagnostics due to readability, color blindness, consistency with their terminal, and performance. Based on the following [issue](https://github.com/swiftlang/swift-package-manager/issues/4776)

### Modifications:

Added the "no-color-diagnostics" subcommand in loggingOptions, in order to initialize the noColorDiagnostics boolean. Then, a simple check is made to determine if the outputParameters are coloured (if are, then checked for if the terminal is a tty). Finally, edited the handleDiagnostics function by adding a formatDiagnostics function that checks for if user wanted colorDiagnostics and emitting warnings and errors, accordingly.

### Result:

Users will now have the option of specifying the no-color-diagnostics option when running SPM in order to disable coloured output

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -208,8 +208,11 @@ public struct LoggingOptions: ParsableArguments {
     /// Whether logging output should be limited to `.error`.
     @Flag(name: .shortAndLong, help: "Decrease verbosity to only include error output.")
     public var quiet: Bool = false
+    
+    @Flag(name: .customLong("no-color-diagnostics"), help: "Disable colored diagnostics")
+    public var noColorDiagnostics: Bool = false
 }
-
+  
 public struct SecurityOptions: ParsableArguments {
     public init() {}
 
@@ -534,7 +537,7 @@ public struct BuildOptions: ParsableArguments {
     // this can be removed once the backtracer uses DWARF instead of frame pointers
     @Flag(inversion: .prefixedNo,  help: .hidden)
     public var omitFramePointers: Bool? = nil
-
+    
     // @Flag works best when there is a default value present
     // if true, false aren't enough and a third state is needed
     // nil should not be the goto. Instead create an enum
--- a/Sources/CoreCommands/SwiftCommandObservabilityHandler.swift
+++ b/Sources/CoreCommands/SwiftCommandObservabilityHandler.swift
@@ -34,9 +34,9 @@ public struct SwiftCommandObservabilityHandler: ObservabilityHandlerProvider {
     ///   - outputStream: an instance of a stream used for output.
     ///   - logLevel: the lowest severity of diagnostics that this handler will forward to `outputStream`. Diagnostics
     ///   emitted below this level will be ignored.
-    public init(outputStream: OutputByteStream, logLevel: Basics.Diagnostic.Severity) {
+    public init(outputStream: OutputByteStream, logLevel: Basics.Diagnostic.Severity, noColorDiagnostics: Bool = false) {
         let threadSafeOutputByteStream = outputStream as? ThreadSafeOutputByteStream ?? ThreadSafeOutputByteStream(outputStream)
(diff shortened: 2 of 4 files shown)


Sources/CoreCommands/Options.swift
```diff
@@ -208,8 +208,11 @@ public struct LoggingOptions: ParsableArguments {
     /// Whether logging output should be limited to `.error`.
     @Flag(name: .shortAndLong, help: "Decrease verbosity to only include error output.")
     public var quiet: Bool = false
+    
+    @Flag(name: .customLong("no-color-diagnostics"), help: "Disable colored diagnostics")
```
cmcgee1024, line 212: suggestion: Maybe the help text could be "Disable colorized diagnostics" instead?

plemarquand (commented):
Question: Do we respect the `NO_COLOR` environment variable to do the same thing? See https://no-color.org

Sources/CoreCommands/SwiftCommandObservabilityHandler.swift
```diff
@@ -68,49 +68,55 @@ public struct SwiftCommandObservabilityHandler: ObservabilityHandlerProvider {
         internal let outputStream: ThreadSafeOutputByteStream
         private let writer: InteractiveWriter
         private let progressAnimation: ProgressAnimationProtocol
-
+        private let noColorDiagnostics: Bool
```
cmcgee1024, line 71: suggestion: Same as above. I think it makes clearer code to use the affirmative where possible instead of the negation. So, this could be `colorDiagnostics`.

Sources/CoreCommands/SwiftCommandObservabilityHandler.swift
```diff
@@ -68,49 +68,55 @@ public struct SwiftCommandObservabilityHandler: ObservabilityHandlerProvider {
...
-                case .error:
-                    output = self.writer.format("error: ", inColor: .red, bold: true)
-                case .warning:
-                    output = self.writer.format("warning: ", inColor: .yellow, bold: true)
-                case .info:
-                    output = self.writer.format("info: ", inColor: .white, bold: true)
-                case .debug:
-                    output = self.writer.format("debug: ", inColor: .white, bold: true)
-                }
+                let (prefix, color, bold): (String, TerminalController.Color, Bool) = formatDiagnostic(for: diagnostic)
```
cmcgee1024, line 93: thought: This is going to be useful to know how to do this for the diagnostics that come from the new swift build build system.

Sources/CoreCommands/SwiftCommandObservabilityHandler.swift
```diff
@@ -68,49 +68,55 @@ public struct SwiftCommandObservabilityHandler: ObservabilityHandlerProvider {
...
+                
+                output = self.writer.format(prefix, inColor: color, bold: bold)
 
                 if let diagnosticPrefix = diagnostic.metadata?.diagnosticPrefix {
-                    output += diagnosticPrefix
-                    output += ": "
+                    output += "\(diagnosticPrefix): "
```
cmcgee1024, line 98: praise: This is a nice fix here using the string interpolation. Less cognitive load, and more idiomatic Swift.

Sources/CoreCommands/SwiftCommandState.swift
```diff
@@ -112,7 +112,7 @@ extension SwiftCommand {
             workspaceDelegateProvider: self.workspaceDelegateProvider,
             workspaceLoaderProvider: self.workspaceLoaderProvider
         )
-
+
```
cmcgee1024, line 115: suggestion: See above about the whitespace changes. They make the code reviews more noisy.

cmcgee1024 (commented):
Overall this is looking good. I've left some issues, suggestions, and thoughts to consider.

PR #8284, opened 2025-02-07
Signing of the prebuilts manifest file.

dschaefer2:
To ensure confidence that the manifest file and zip files are the ones that were published, we use checksums for the zip files which are listed in the manifest file. We now add signing for the manifest file to ensure it isn't compromised. We use the same scheme as we used for package collections where a signature is generated for the core manifest data and then injected into the JSON file at the top level.

Note that we are adding the ManifestSigning to the Workspace module with the eventual intention of making this a module we can reuse with package collections once we forward port to main. We don't want to make changes to the package collections on the 6.1 branch.

Also added a quick fix to the host platform computation that was causing compile errors if not the two major ones (i.e. 32-bit Android ARM hosts).


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test windows

dschaefer2:

dschaefer2:
@swift-ci please test

dschaefer2:
For those wondering the contents of the ManifestSigning folder come from the PackageCollectionSigning. The key signing parts of that are a copy with the front end generalized to handle Codables. It encodes them and then generates the signature from the resulting Data block. When I forward port this to main, I'll move that into it's own module, port the package collection signing to use it as well and delete it's copy if this.

dschaefer2:
@swift-ci please test Windows

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test

Sources/Basics/FileSystem/FileSystem+Extensions.swift
```diff
@@ -677,3 +677,17 @@ extension FileSystem {
...
     }
 }
+
+extension FileSystem {
+    /// Do a deep enumeration, passing each file to block
+    public func enumerate(directory: AbsolutePath, block: (AbsolutePath) throws -> ()) throws {
+        for file in try getDirectoryContents(directory) {
+            let path = directory.appending(file)
+            if isDirectory(path) {
+                try enumerate(directory: path, block: block)
```
cmcgee1024, line 687: thought: It might be good to put a limit on the recursion in case it blows up a stack somewhere. Alternatively, there is this API from FileManager that I suspect iterates instead of recurses to avoid that kind of problem: https://developer.apple.com/documentation/foundation/filemanager/1408726-enumerator

Sources/Workspace/Workspace+Prebuilts.swift
```diff
@@ -246,13 +268,37 @@ extension Workspace {
...
-                        as: PrebuiltsManifest.self
+                        as: SignedPrebuiltsManifest.self
                     )
+
+                    // Check the signature
+                    // Ignore errors coming from the certificate loading, that will shutdown the build
+                    // instead of letting it continue with build from source.
+                    if let rootCertPath {
+                        try await withTemporaryDirectory(fileSystem: fileSystem) { tmpDir in
+                            try fileSystem.copy(from: rootCertPath, to: tmpDir.appending(rootCertPath.basename))
```
cmcgee1024, line 284: question: Why copy the rootCert here? Can it be used in-place instead, or in-memory instead? I'm wondering if there can be a [toctou](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) potential here.

Sources/Workspace/Workspace+Prebuilts.swift
```diff
@@ -264,16 +310,20 @@ extension Workspace {
                 }
             }
 
-            if fileSystem.exists(destination), let manifest = try? loadManifest() {
+            if fileSystem.exists(destination), let manifest = try? await loadManifest() {
```
cmcgee1024, line 313: question: Why not try the cache first, and the fall-back to the loadManifest() if that fails? Fewer filesystem exists() checks might reduce the risk of toctou types of vulnerabilities.

Sources/Workspace/Workspace+Prebuilts.swift
dschaefer2: The signing code I'm using scans the directory for certs so the root cert needs to be alone in this directory. It tried to use whatever certs it can find there and only the root cert works.

Sources/Workspace/Workspace+Prebuilts.swift
```diff
@@ -359,112 +402,120 @@ extension Workspace {
...
-                version.description,
-                artifactFile
-            )
+            let artifactFile = artifactName + (hostPlatform.os == .linux ? ".tar.gz" : ".zip")
+            let destination = scratchDir.appending(artifactFile)
+            let cacheFile = cachePath?.appending(components: package.identity.description, version.description, artifactFile)
 
             let zipExists = fileSystem.exists(destination)
             if try (!zipExists || !check(path: destination, checksum: artifact.checksum)) {
-                if zipExists {
```
cmcgee1024, line 382: question: What happened to this zipExists check?

Sources/Workspace/Workspace+Prebuilts.swift
dschaefer2: The key feature of loadManifest is checking the signature. I want to do that every time to make sure the manifest isn't compromised. If the one in the scratch dir isn't valid we fall back to the cache and if that's not valid we download again and if that's not valid we skip the prebuilts all together. The cache is optional so we can't assume it's there.

Sources/Workspace/Workspace+Prebuilts.swift
```diff
@@ -359,112 +402,120 @@ extension Workspace {
...
-                    observabilityScope.emit(info: errorString)
-                    self.delegate?.didDownloadPrebuilt(
-                        from: artifactURL.absoluteString,
-                        result: .failure(StringError(errorString)),
-                        duration: fetchStart.distance(to: .now())
-                    )
-                    return nil
+                    if let cacheFile {
+                        // Cache the zip file
+                        if fileSystem.exists(cacheFile) {
```
cmcgee1024, line 511: suggestion (non-blocking): Just go ahead and delete the cacheFile with removeFileTree() without checking if it already exists first. Checking first doesn't seem to buy anyone much. Ideally, the copy could just be destructive, so just do that, after a pre-emptive createDirectory() to ensure that it has a home.

dschaefer2: If it doesn't exist, the removeFileTree fails. I know that the hard way :).

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -93,8 +93,47 @@ struct BuildPrebuilts: AsyncParsableCommand {
...
+            }
+
+            guard !certChainPathStrs.isEmpty else {
+                throw ValidationError("No certificates provided")
+            }
+        }
+    }
+
     mutating func run() async throws {
-        let fm = FileManager.default
```
cmcgee1024, line 97: praise: It's good to have everything standardized on one filesystem API.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -131,8 +170,8 @@ struct BuildPrebuilts: AsyncParsableCommand {
 
             for version in repo.versions {
                 let versionDir = prebuiltDir.appending(version.tag)
-                if !fm.fileExists(atPath: versionDir.pathString) {
-                    try fm.createDirectory(atPath: versionDir.pathString, withIntermediateDirectories: true)
+                if !fileSystem.exists(versionDir) {
```
cmcgee1024, line 173: suggestion (non-blocking): Create the directory without first checking if it exists. It can just be an assertion that it is there, and the success whether true or false can be ignored.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -169,82 +206,148 @@ struct BuildPrebuilts: AsyncParsableCommand {
 
                         // Copy the library to staging
                         let lib = "lib\(library.name).a"
```
cmcgee1024, line 208: question: Does the filename change on Windows at all? Does it become `"\(library.name).lib"`?

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -169,82 +206,148 @@ struct BuildPrebuilts: AsyncParsableCommand {
...
+                            .filter({ $0.hasSuffix(".zip.json")})
+                            .compactMap({
+                                let data: Data = try fileSystem.readFileContents(versionDir.appending($0))
+                                return try? decoder.decode(Workspace.PrebuiltsManifest.Library.Artifact.self, from: data)
+                            })
                     )
                     newLibraries.insert(newLibrary)
 
-                    try await shell("git reset --hard", cwd: repoDir)
+                    try await shell("git restore .", cwd: repoDir)
```
cmcgee1024, line 270: praise: It's good to use the high-level git commands where possible. Git reset has some sharp edges.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -169,82 +206,148 @@ struct BuildPrebuilts: AsyncParsableCommand {
...
+                            for file in try fileSystem.getDirectoryContents(srcIncludeDir) {
+                                try fileSystem.copy(from: srcIncludeDir.appending(file), to: destIncludeDir.appending(file))
                             }
                         }
 
                         // Zip it up
-                        let zipFile = versionDir.appending("\(swiftVersion)-\(library.name)-\(platform).zip")
                         let contentDirs = ["lib", "Modules"] + (library.cModules.isEmpty ? [] : ["include"])
 #if os(Windows)
+                        let zipFile = versionDir.appending("\(swiftVersion)-\(library.name)-\(platform).zip")
```
bkhouri, line 231: suggestion (non-blocking): The prebuilt file name pattern appears to be duplicated here and in the code base. Consider creating a function that returns the filename to avoid having to duplicate the code.  this way, if we decide to augment the filename in the future, we only need to update the function API and we'll get notified for all the uses.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -169,82 +206,148 @@ struct BuildPrebuilts: AsyncParsableCommand {
...
                         try await shell("tar -acf \(zipFile.pathString) \(contentDirs.joined(separator: " "))", cwd: stageDir)
+                        let contents = try ByteString(Data(contentsOf: zipFile.asURL))
+#elseif os(Linux)
+                        let tarFile = versionDir.appending("\(swiftVersion)-\(library.name)-\(platform).tar.gz")
+                        try await shell("tar -zcf \(tarFile.pathString) \(contentDirs.joined(separator: " "))", cwd: stageDir)
+                        let contents = try ByteString(Data(contentsOf: tarFile.asURL))
 #else
+                        let zipFile = versionDir.appending("\(swiftVersion)-\(library.name)-\(platform).zip")
```
bkhouri, line 239: issue (blocking): Will this cause a compile error as `zipFile` is defined on line 231 above.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -169,82 +206,148 @@ struct BuildPrebuilts: AsyncParsableCommand {
...
                         try await shell("zip -r \(zipFile.pathString) \(contentDirs.joined(separator: " "))", cwd: stageDir)
+                        let contents = try ByteString(Data(contentsOf: zipFile.asURL))
```
bkhouri, line 241: issue (blocking): Will this cause a compile error as `contents` is defined on line 237 above.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -93,8 +93,47 @@ struct BuildPrebuilts: AsyncParsableCommand {
     @Option(help: "The command to use for docker")
     var dockerCommand: String = "docker"
 
+    @Flag(help: "Building the prebuilts")
```
plemarquand, line 96:
```suggestion
    @Flag(help: "Whether to build the prebuilts")
```

Sources/swift-build-prebuilts/BuildPrebuilts.swift
bkhouri, line 270: question: If I have local changes, and then decide to run the `swift-build-prebuilts` executable targets, will this wipe away my local changes? If so, we may want to error out early in the script informing the user there are unstaged changed.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: Nope, Windows doesn't care what the library extension is, or at least it is accepting the .a.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -169,82 +206,148 @@ struct BuildPrebuilts: AsyncParsableCommand {
...
                         // Copy the C module headers
                         for cModule in library.cModules {
                             let cModuleDir = version.cModulePaths[cModule] ?? ["Sources", cModule]
                             let srcIncludeDir = repoDir.appending(components: cModuleDir).appending("include")
                             let destIncludeDir = includesDir.appending(cModule)
-                            try fm.createDirectory(atPath: destIncludeDir.pathString, withIntermediateDirectories: true)
-                            for file in try fm.contentsOfDirectory(atPath: srcIncludeDir.pathString) {
-                                try fm.copyItem(atPath: srcIncludeDir.appending(file).pathString, toPath: destIncludeDir.appending(file).pathString)
+                            try fileSystem.createDirectory(destIncludeDir, recursive: true)
+                            // TODO: should enumerate the directory to pick up deep headers
```
plemarquand, line 222: Still a valid TODO?

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -169,82 +206,148 @@ struct BuildPrebuilts: AsyncParsableCommand {
...
+                    privateKeyPathStr = certsPath.appending("Test_rsa_key.pem").pathString
+                    certChainPathStrs = [
+                        certsPath.appending("Test_rsa.cer").pathString,
+                        certsPath.appending("TestIntermediateCA.cer").pathString,
+                        certsPath.appending("TestRootCA.cer").pathString
+                    ]
+                }
+
+                guard let privateKeyPathStr else {
+                    fatalError("No private key path provided")
```
plemarquand, line 319: throwing an error would be more consistent with the other SwiftPM commands and lead to a better message

Sources/Workspace/Workspace+Prebuilts.swift
```diff
@@ -246,13 +268,37 @@ extension Workspace {
...
+                        as: SignedPrebuiltsManifest.self
                     )
+
+                    // Check the signature
+                    // Ignore errors coming from the certificate loading, that will shutdown the build
+                    // instead of letting it continue with build from source.
+                    if let rootCertPath {
+                        try await withTemporaryDirectory(fileSystem: fileSystem) { tmpDir in
+                            try fileSystem.copy(from: rootCertPath, to: tmpDir.appending(rootCertPath.basename))
+                            let validator = ManifestSigning(trustedRootCertsDir: tmpDir, observabilityScope: ObservabilitySystem.NOOP)
```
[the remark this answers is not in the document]
dschaefer2: The signing code actually emits error diagnostics when the signature validation fails which stops the build. That shouldn't happen since we'll just build from source if the validation fails.

Sources/Workspace/Workspace+Prebuilts.swift
```diff
@@ -246,13 +268,37 @@ extension Workspace {
...
-                    return try JSONDecoder().decode(
+                    let signedManifest = try JSONDecoder().decode(
                         path: destination,
                         fileSystem: fileSystem,
-                        as: PrebuiltsManifest.self
+                        as: SignedPrebuiltsManifest.self
                     )
+
+                    // Check the signature
+                    // Ignore errors coming from the certificate loading, that will shutdown the build
```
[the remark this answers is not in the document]
dschaefer2: That's what the NOOPs do. I could make that clearer.

Sources/Workspace/Workspace+Prebuilts.swift
```diff
@@ -331,7 +373,7 @@ extension Workspace {
                 }
             }
 
-            if let manifest = try loadManifest() {
+            if let manifest = try await loadManifest() {
```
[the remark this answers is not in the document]
dschaefer2: Yes. The other ones we ignore end up forcing the download. At this point it's the last chance to get things right. And really the only thing throwing are filesystem errors which point at bigger problems.

Sources/Workspace/Workspace+Prebuilts.swift
```diff
@@ -359,112 +402,120 @@ extension Workspace {
...
             }
 
-            let artifactFile = artifactName + ".zip"
-            let prebuiltsDir = cachePath ?? scratchPath
-            let destination = prebuiltsDir.appending(components:
-                package.identity.description,
-                version.description,
-                artifactFile
-            )
+            let artifactFile = artifactName + (hostPlatform.os == .linux ? ".tar.gz" : ".zip")
```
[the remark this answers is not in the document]
dschaefer2: This was to get the tests to pass at the last minute. The tests mock everything out as being a macOS host. In those other cases, the mocking overrides them so are avoided.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: Sure, this utility only has one or two customers, the infra team and us testing it. I can clean that up when we forward port this to main.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: Nope, also fixed at the last minute. I'll do a clean up as we forward port this to main or I need to do another commit.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: This isn't really like the other SwiftPM commands. Ideally this is just a Swift script that's not in the package, but we can't do scripts that have dependencies yet.

Sources/Workspace/ManifestSigning/Base64URL.swift
```diff
@@ -0,0 +1,76 @@
...
+// SPDX-License-Identifier: MIT
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+// Source: https://github.com/vapor/jwt-kit/blob/master/Sources/JWTKit/Utilities/Base64URL.swift
+
+extension DataProtocol {
+    func base64URLDecodedBytes() -> Data? {
```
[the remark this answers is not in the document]
dschaefer2: This is code copied over from the PackageCollectionSigning. I was not intending to change it at this time.

Sources/Workspace/ManifestSigning/Base64URL.swift
```diff
@@ -0,0 +1,76 @@
...
+        data.base64URLEscape()
+        return data
+    }
+}
+
+extension Data {
+    /// Converts base64-url encoded data to a base64 encoded data.
+    ///
+    /// https://tools.ietf.org/html/rfc4648#page-7
+    mutating func base64URLUnescape() {
```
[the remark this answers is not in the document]
dschaefer2: This is code copied over from the PackageCollectionSigning. I was not intending to change it at this time.

Sources/Workspace/Workspace+Prebuilts.swift
```diff
@@ -158,13 +172,20 @@ extension Workspace {
...
+            prebuiltsDownloadURL: String?,
+            rootCertPath: AbsolutePath?
         ) {
             self.fileSystem = fileSystem
             self.hostPlatform = hostPlatform
             self.authorizationProvider = authorizationProvider
             self.httpClient = customHTTPClient ?? HTTPClient()
+
+#if os(Linux)
+            self.archiver = customArchiver ?? TarArchiver(fileSystem: fileSystem)
```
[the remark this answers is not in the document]
dschaefer2: The docker images used for building the archives for Linux don't have zip installed. We can't assume it's there. Tar on the other hand has been universal since we had Tape ARchives :). And this is just how the Archive is structured, there are separate ones for Zip and Tar.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: Only one of the three conditions are selected. How can that be true?

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: If you do it while the script is running between the clone above and before we remove the source tree at the end, sure, but I'm not expecting the to happen.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -169,82 +206,148 @@ struct BuildPrebuilts: AsyncParsableCommand {
...
                     )
                     newLibraries.insert(newLibrary)
 
-                    try await shell("git reset --hard", cwd: repoDir)
+                    try await shell("git restore .", cwd: repoDir)
                 }
+            }
+        }
+
+        _ = FileManager.default.changeCurrentDirectoryPath(stageDir.pathString)
```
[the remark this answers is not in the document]
dschaefer2: Probably. The switch to FileSystem was done late when I added it to the signing API. I can do that when I clean things up forward porting to main.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: Same as above.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: That's a good point. I'll do that when we get this to main.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: That's what I was trying to hint at in the comment at the top of the file. :).

Sources/Workspace/Workspace+Prebuilts.swift
cmcgee1024: Right, if the remoteFileTree() fails then just ignore that and proceed with the copy. The copy operation is the important one, so it should fail the entire operation if that fails.

dschaefer2: I often debug my code with swift error breakpoints turned on. It's super annoying to have those hit on success conditions. That's why I'd rather check and avoid throwing errors unless something is really broken. And the cost is super minimal here since it's only called once.

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: Actually those messages should never come out. This is checked by the `validate` function above.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

Sources/swift-build-prebuilts/BuildPrebuilts.swift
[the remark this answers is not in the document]
dschaefer2: Done

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: Switched this to actually do the enumerate. Done.

plemarquand (approved):
My concerns are addressed, thanks!

Issue #8277, opened 2025-02-04
`add-target` command doesn't work with binary targets


grantjbutler:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

The `add-target` command doesn't seem to properly support either creating a target that points to a remote binary artifact or creating a target that points to a local binary artifact. The arguments in the command exist, but the command doesn't seem to property process them and create the correct target type. For example, the following two commands fail, but in different ways (see "Actual Behavior" for how these fail):

```bash
swift package add-target Capacitor \
  --url "https://github.com/ionic-team/capacitor-swift-pm/releases/download/7.0.1/Capacitor.xcframework.zip" \
  --checksum "7781e0a8d8072c26132fd1170e4fe08ae72a98ef664e28176c7e43f241c3479e"
```

```bash
swift package add-target LocalBinaryDependency --path Artifacts/LocalBinaryDependency.xcframework
```

This seems to be because the presence of these arguments doesn't change the type of the target to be created from `.regular` to `.binary`. Some of these values being present gets caught in `TargetDescription`'s validation and fails the command line call. Binary targets aren't supported, in the sense of the `--type` argument does not accept a value named `binary` to explicitly change the type of the created target to `.binary`, which would then map to a `.binaryTarget` value in the manifest.

I have limited knowledge of the internals of SPM, but I would hazard a guess that this could be addressed by adding a new `binary` case to the [`TargetType` enumeration](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/Commands/PackageCommands/AddTarget.swift#L29), and then properly [mapping that](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/Commands/PackageCommands/AddTarget.swift#L93-L98) to a `TargetDescription.TargetKind.binary`.

### Expected behavior

Providing the information for a binary target would properly create a `.binaryTarget` target in the SPM package's manifest.

### Actual behavior

For remote binary targets, the command will fail with the following error:

```
error: target 'Capacitor' contains a value for disallowed property 'url'
```

For local binary targets, the command succeeds, but creates a normal target, but with its `path` set in the manifest to the path provided to the command. The command will also create the target's source folders in the `Sources/` folder.

### Steps to reproduce

[...]

PR #8290, opened 2025-02-12
Add `binary` type for binary targets to `add-target` command.

grantjbutler:
Updates the `add-target` subcommand to allow creating a target that references a remote or local binary artifact.

### Motivation:

The `add-target` command currently accepts a set of arguments that make it seem like it should be able to create a `.binaryTarget` in the package's manifest. However, when trying to use the command in this way, you'll either get an error because of validation if trying to create a target for a remote binary artifact, or a regular source target will be created if trying to create a target for a local binary artifact. Issue #8277 has the full details.

### Modifications:

This PR adds `.binary` as a new value for the `TargetType` enum (which is used for parsing the `--type` argument of the `add-target` command). This value gets mapped to `TargetDescription.TargetKind.binary` when actually creating the target in the manifest.

### Result:

By specifying `--type binary` for the `add-target` command and providing a value for `--url` or `--path`, a `.binaryTarget` will be added to the package's manifest, with the appropriate configuration depending on if the binary artifact is sourced from a remote URL or a local path.

### DIFF
--- a/Tests/CommandsTests/PackageCommandTests.swift
+++ b/Tests/CommandsTests/PackageCommandTests.swift
@@ -1067,6 +1067,65 @@ final class PackageCommandTests: CommandsTestCase {
             XCTAssertMatch(contents, .contains(#""OtherLib""#))
         }
     }
+    
+    func testPackageAddRemoteBinaryTarget() async throws {
+        try await testWithTemporaryDirectory { tmpPath in
+            let fs = localFileSystem
+            let path = tmpPath.appending("PackageB")
+            try fs.createDirectory(path)
+
+            try fs.writeFileContents(path.appending("Package.swift"), string:
+                """
+                // swift-tools-version: 5.9
+                import PackageDescription
+                let package = Package(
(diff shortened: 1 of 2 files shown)


plemarquand:
@swift-ci please test

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -1067,6 +1067,65 @@ final class PackageCommandTests: CommandsTestCase {
...
+                """
+                // swift-tools-version: 5.9
+                import PackageDescription
+                let package = Package(
+                    name: "client"
+                )
+                """
+            )
+
+            _ = try await execute(["add-target", "client", "--url", "https://example.com/client.xcframework.zip", "--checksum", "abcdef1234567890", "--type", "binary"], packagePath: path)
```
hi2gage, line 1087: Would it make sense to add checksum validation to ensure that the user is inputting a valid SHA-256 checksum? As called out in `PackageDescription.Target.checksum`

PR #8293, opened 2025-02-13
Tests: Improve testVersion assertion and add duplicate symbol test

bkhouri:
The various CommandTestCases 'testVersion' tests were only asserting the output contains a string.  Update the assertion to check the output matches a regular expression.

Also, add a test that ensured the command does not emit duplicate symbol on stdout and stderr.

rdar://144103998


bkhouri:
@swift-ci please test

dschaefer2 (approved):

PR #8297, opened 2025-02-17
Add new swift-crypto on asn1 to the bootstrap script

cmcgee1024:
Add new swift-crypto on asn1 to the bootstrap script

### Motivation:

The newest version of swift-crypto has a dependency on swift-asn1. Without providing the directory the CMakeLists.txt of swift-crypto will attempt to vendor a specific version of that library that may not match what is in the update-checkout version using in Swift CI, which can lead to mismatched symbols during linking.

### Modifications:

This fix adds the swift-asn1 dependency to swift-crypto by adding the directory for swift-asn1 to cmake when building swift-crypto.

### Result:

After the change the new dependency from swift-crypto should be ready for when the new version of the lib is adopted.

### DIFF
--- a/Utilities/bootstrap
+++ b/Utilities/bootstrap
@@ -386,8 +386,9 @@ def build(args):
         ]
         build_dependency(args, "swift-driver", swift_driver_cmake_flags)
         build_dependency(args, "swift-collections")
-        build_dependency(args, "swift-crypto")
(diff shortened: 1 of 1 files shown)


plemarquand:
@swift-ci please test

jakepetroules (approved):

PR #8298, opened 2025-02-19
build: define build path structure in BuildSystemProvider.Kind

### LINKED ISSUE #8272: swift-test doesn't find xctest bundle with swiftbuild

### Description

swift-test is using the old native build path layout to find the xctest bundle when used with `--build-system swiftbuild`

### Expected behavior

swift-test should run the tests.

### Actual behavior

I'm getting an error:

`swiftpm_testing_helper/Entrypoint.swift:29: Fatal error: Failed to open test bundle at path /Users/dschaefer/swift/douglib/.build/arm64-apple-macosx/debug/douglibPackageTests.xctest/Contents/MacOS/douglibPackageTests: dlopen(/Users/dschaefer/swift/douglib/.build/arm64-apple-macosx/debug/douglibPackageTests.xctest/Contents/MacOS/douglibPackageTests, 0x0101)`

it's actually at the following with swiftbuild.

`.build/arm64-apple-macosx/Products/Debug/douglibPackageTests.xctest/Contents/MacOS/douglibPackageTests`

### Steps to reproduce

Run `swift test --build-system swiftbuild`, or `swift-test` until the initial PR lands.

_No response_

_No response_

bkhouri:
There are a few places that check the BuildSystemProvider.Kind is Xcode to determine the build directory structure.  With the upcoming Swift Build integration, which uses the "Xcode path" structure, we would need to update all instances to check the two build system provider kinds.

Add an extension on the BuildSystemProvider.Kind that create a boolean variable `useXcodeBuildEngine`, which determines whether the build system should use the xcode path structure or not.  In addition, update the code that requires a  "isXcodeBuildSystemEnabled" to return this build system provider variable.

This will hopefully help address #8272 when #8271 is merged and added `case .swiftbuild: return true` to the `useXcodeBuildEngine `

Fixes #8272
rdar://144023142

### DIFF
--- a/Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
+++ b/Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
@@ -168,6 +168,14 @@ public struct BuildSystemProvider {
     }
 }
 
+extension BuildSystemProvider.Kind {
+    public var usesXcodeBuildEngine: Bool {
+        switch self {
+            case .native: return false
+            case .xcode: return true
+        }
+    }
+}
 private enum Errors: Swift.Error {
     case buildSystemProviderNotRegistered(kind: BuildSystemProvider.Kind)
 }
--- a/Sources/CoreCommands/SwiftCommandState.swift
+++ b/Sources/CoreCommands/SwiftCommandState.swift
@@ -441,7 +441,7 @@ public final class SwiftCommandState {
             self.observabilityHandler.progress,
             self.observabilityHandler.prompt
         )
-        let isXcodeBuildSystemEnabled = self.options.build.buildSystem == .xcode
+        let isXcodeBuildSystemEnabled = self.options.build.buildSystem.usesXcodeBuildEngine
         let workspace = try Workspace(
             fileSystem: self.fileSystem,
             location: .init(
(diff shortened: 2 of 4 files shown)


Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
```diff
@@ -168,6 +168,14 @@ public struct BuildSystemProvider {
     }
 }
 
+extension BuildSystemProvider.Kind {
+    public var useXcodeBuildSystemPath: Bool {
```
bkhouri, line 172: suggestion: is this not just a Build System path. Maybe update the variable name to `useXcodeBuildEngine`

bkhouri: Done

bkhouri:
@swift-ci please test

Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
```diff
@@ -168,6 +168,14 @@ public struct BuildSystemProvider {
     }
 }
 
+extension BuildSystemProvider.Kind {
+    public var useXcodeBuildEngine: Bool {
```
MaxDesiatov, line 172: This should be spelled as `usesXcodeBuildEngine` to conform to the API Design Guidelines: https://www.swift.org/documentation/api-design-guidelines/#strive-for-fluent-usage

> Uses of Boolean methods and properties should read as assertions about the receiver when the use is nonmutating, e.g. `x.isEmpty`, `line1.intersects(line2)`.

```suggestion
    public var usesXcodeBuildEngine: Bool {
```

MaxDesiatov (commented):
I'm not sure if `usesXcodeBuildEngine` is more readable than `== .xcode`. If this were more than an enum with two cases (or there were plans to make it more complex in the future), one might argue that the check is complex enough, but for an enum as trivial as this I don't see much benefit.

bkhouri:
> I'm not sure if `usesXcodeBuildEngine` is more readable than `== .xcode`. If this were more than an enum with two cases (or there were plans to make it more complex in the future), one might argue that the check is complex enough, but for an enum as trivial as this I don't see much benefit.

This enum will be expanding with the introduction of #8271.  And this change will force us to set the  `case .swiftbuild: return true`, thus having automatically fixed #8272.

bkhouri:
@swift-ci please test

MaxDesiatov (approved):

bkhouri:
@swift-ci please test

bkhouri:
@swift-ci please test windows

PR #8308, opened 2025-02-26
[6.1] Fix retain cycle in `LLBuildProgressTracker`

ahoppen:
- **Explanation**: `self.taskTracker.onTaskProgressUpdateText` captured `self` strongly, creating a reference cycle in `LLBuildProgressTracker`, which caused it to never get deallocated.
- **Scope**: Manifested as a memory leak in SourceKit-LSP
- **Issue**: n/a
- **Original PR**: https://github.com/swiftlang/swift-package-manager/pull/8305
- **Risk**: Low, just captures a delegate weakly
- **Testing**: Verified that this fixes the memory leak we saw
- **Reviewer**: @MaxDesiatov


ahoppen:
@swift-ci Please test

bnbarham (approved):

plemarquand (approved):

PR #8313, opened 2025-02-26
Align --build-system xcode language mode validation with --build-system native

owenv:
Fix a bug where builds using --build-system xcode would fail if a target was configured to use a language mode which was not also specified at the top level of the package manifest


owenv:
@swift-ci smoke test

jakepetroules (approved):

dschaefer2 (approved):

owenv:
@swift-ci test

PR #8316, opened 2025-02-27
Add `color-diagnostics/no-color-diagnostics` CLI flags

johnbute:
Added the no-color-diagnostics flag to allow users to toggle on/off color diagnostics
### Motivation:
Not everyone has access to tty, or like colors in their terminal, or are color blind. This feature allows them to control if they want to see colors or not.
### Modifications:

1. Added the color-diagnostics/no-color-diagnostics flag
2. Added test cases to test for colors
3. edited snippets to include this feature too
4. Created a ManualWriter class to manage ANSI color codes without requiring a TTY environment, facilitating better testability in non-terminal scenarios

### Result:
Developers can now control their color diagnostics when building, running, testing, installing sdks, or browsing code snippets within the swift package learn feature.

Furthermore, ManualWriter allows for better testability when working in a test environment, when a terminal is not always available, and testing outputs in tty/non-tty scenarios is needed.


Sources/Basics/Observability.swift
```diff
@@ -161,6 +162,14 @@ public protocol DiagnosticsHandler: Sendable {
     func handleDiagnostic(scope: ObservabilityScope, diagnostic: Diagnostic)
 }
 
+//helper protocol to configurate the style of diagnostics
```
MaxDesiatov, line 165: Please use doc comments formatting to make these more discoverable.

```suggestion
/// Helper protocol to configurate the style of diagnostics
```

Sources/Basics/Observability.swift
```diff
@@ -161,6 +162,14 @@ public protocol DiagnosticsHandler: Sendable {
...
+protocol SeverityConfig {
+    var prefix: String { get }
```
MaxDesiatov, line 167: IMO at least this property warrants a doc comment to explain what kind of prefix this is in more details.

Sources/Basics/Observability.swift
```diff
@@ -161,6 +162,14 @@ public protocol DiagnosticsHandler: Sendable {
...
+    var color: TerminalController.Color { get }
+    var isBold: Bool { get }
+}
+
+
 // helper protocol to share default behavior
```
MaxDesiatov, line 173: Let's convert this to a doc comment while we're at it.

```suggestion
/// Helper protocol to share default behavior.
```

Sources/Basics/Observability.swift
MaxDesiatov: Having another look at it, does this even warrant a separate protocol if there's only one implementation of it?

Sources/Basics/Observability.swift
```diff
@@ -409,6 +418,36 @@ public struct Diagnostic: Sendable, CustomStringConvertible {
...
+            switch self{
+            case .debug:
+                return "debug: "
+            case .info:
+                return "info: "
+            case .warning:
+                return "warning: "
+            case .error:
+                return "error: "
+
```
MaxDesiatov, line 432: Stray newline. Please the formatter per [the `CONTRIBUTING.md` guidelines](https://github.com/swiftlang/swift-package-manager/blob/main/CONTRIBUTING.md#creating-pull-requests).

Sources/Basics/ProgressAnimation/PercentProgressAnimation.swift
```diff
@@ -35,27 +36,42 @@ extension ProgressAnimation {
...
 
     /// Creates repeating string for count times.
     /// If count is negative, returns empty string.
     private func repeating(string: String, count: Int) -> String {
         return String(repeating: string, count: max(count, 0))
     }
+    
+    func isBold() -> Bool {
+        return self.isColorized
+    }
```
MaxDesiatov, line 56: Since this function is used in a single place to initialize the `let isBold` constant binding, I don't think it's worth to maintain it separately.

Sources/Basics/ProgressAnimation/PercentProgressAnimation.swift
```diff
@@ -35,27 +36,42 @@ extension ProgressAnimation {
...
+    func colorizeText(color: TerminalController.Color = .noColor) -> TerminalController.Color{
+        if self.isColorized{
+            return color
+        }
+        return .noColor
+    }
 
     func update(step: Int, total: Int, text: String) {
         assert(step <= total)
+        let isBold = isBold()
```
MaxDesiatov, line 67: I would find `let isBold = self.isColorized` less convoluted than creating a separate function used only once for this.

Sources/CoreCommands/SwiftCommandObservabilityHandler.swift
```diff
@@ -34,9 +35,10 @@ public struct SwiftCommandObservabilityHandler: ObservabilityHandlerProvider {
     ///   - outputStream: an instance of a stream used for output.
     ///   - logLevel: the lowest severity of diagnostics that this handler will forward to `outputStream`. Diagnostics
     ///   emitted below this level will be ignored.
-    public init(outputStream: OutputByteStream, logLevel: Basics.Diagnostic.Severity) {
+
+    public init(outputStream: OutputByteStream, logLevel: Basics.Diagnostic.Severity, colorDiagnostics: Bool = true, manualWriterParams: [String: Bool] = ["use": false]) {
```
MaxDesiatov, line 39: What's the reasoning for maintaining a stringly-typed dictionary for `manualWriterParams` instead of creating a separate strongly-typed struct? Besides, naming is obscure here, quite unclear what `use` is referring to here.

Sources/CoreCommands/SwiftCommandObservabilityHandler.swift
```diff
@@ -173,38 +175,90 @@ extension SwiftCommandObservabilityHandler.OutputHandler: DiagnosticsHandler {}
...
+            return "\(stringColor)\(bold ? "\u{001B}[1m" : "")\(string)\u{001B}[0m"
+        } else {
+            return string
+        }
+    }
+
+    private func getColorString(color: TerminalController.Color) -> String {
+        switch color {
+            case .noColor: return ""
+            case .red: return "\u{001B}[31m"
```
MaxDesiatov, line 214: These escape code strings are already defined in TSC, no need to duplicate them here: https://github.com/swiftlang/swift-tools-support-core/blob/bd9e2de5fdada4a0249de3be88f9ad95fe8511f6/Sources/TSCBasic/TerminalController.swift#L46

cmcgee1024: It appears that there's no way to get at those strings since they are file private. What do you suggest here?

cmcgee1024:
@swift-ci please test

Sources/CoreCommands/SwiftCommandObservabilityHandler.swift
MaxDesiatov: Feels totally fine to me to make them `public` in TSC first.

Sources/Basics/ProgressAnimation/PercentProgressAnimation.swift
bkhouri: I find this clarifies the intent and I actually like having this here, and it will allow us to change the "isBold()" behaviour in a single location should we ever decide to change the "bold" condition.

Sources/Basics/Observability.swift
```diff
@@ -409,6 +418,36 @@ public struct Diagnostic: Sendable, CustomStringConvertible {
...
+            case .error:
+                return "error: "
+            
+            }
+        }
+        
+        public var color: TerminalController.Color {
+            switch self {
+            case .debug:
+                return .white
```
bkhouri, line 439: question: how does this work with a terminals light and dark theme?

Sources/Basics/Observability.swift
```diff
@@ -161,7 +162,14 @@ public protocol DiagnosticsHandler: Sendable {
     func handleDiagnostic(scope: ObservabilityScope, diagnostic: Diagnostic)
 }
 
-// helper protocol to share default behavior
+/// Helper protocol to configurate the style of diagnostics.
+protocol SeverityConfig {
```
MaxDesiatov, line 166: What's the purpose of this protocol if it is only used in a single place? Why not use that same type directly?

MaxDesiatov (commented):
I suggest applying the formatter only to files that were modified, ideally only the PR should only reformat lines that were modified. Otherwise there are a lot of seemingly unrelated changes, making it hard to review.

MaxDesiatov:
This also needs conflicts with `main` resolved, which are supposedly minor.

cmcgee1024:
@swift-ci please test

bripeticca:
@swift-ci please test

johnbute:
Turning into draft to fix issues that arrived with the new build system

Sources/Basics/Observability.swift
johnbute: Yep! fixed it. Thanks for the find max :)

Sources/Basics/Observability.swift
johnbute: Ran Formatting!

Sources/Basics/ProgressAnimation/PercentProgressAnimation.swift
johnbute: You are completely right, I reverted the changes.

Sources/CoreCommands/SwiftCommandObservabilityHandler.swift
johnbute: This is removed, due to the inability to efficiently test color diagnostics.

Sources/Basics/Observability.swift
johnbute: Originally, SPM never took into account the theme of the terminal. This PR is merely aimed at fixing color diagnostics to give users options to enable or disable them. In the future, it would be nice to either let users set the diagnostics colors themselves, or for diagnostics to be responsive based on the theme of the terminal.

Sources/Basics/Observability.swift
johnbute: This protocol originally was also used in testing, but due to the pitfalls of testing color diagnostics, it's only being used for a single purpose. Although I agree with you that this added level of abstraction may be unnecessary, if ever someone does figure out a better way of testing color diagnostics, I believe that this protocol would help them out. Thus, for the sake of not adding test debt, I would want to keep this level of abstraction.

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swiift-ci test Windows

PR #8321, opened 2025-03-03
Found another place where we specify the swift-syntax version.

dschaefer2:
Updates the config.json to swift-syntax 601.

### DIFF
--- a/Utilities/config.json
+++ b/Utilities/config.json
@@ -1,3 +1,3 @@
 {"version":1,
-  "swiftSyntaxVersionForMacroTemplate":{"major":600,"minor":0,"patch":0, "prereleaseIdentifier":"latest"},
-  "swiftTestingVersionForTestTemplate":{"major":0,"minor":8,"patch":0}}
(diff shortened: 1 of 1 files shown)


dschaefer2:
@swift-ci please test

Utilities/config.json
```diff
@@ -1,3 +1,3 @@
 {"version":1,
-  "swiftSyntaxVersionForMacroTemplate":{"major":600,"minor":0,"patch":0, "prereleaseIdentifier":"latest"},
-  "swiftTestingVersionForTestTemplate":{"major":0,"minor":8,"patch":0}}
+  "swiftSyntaxVersionForMacroTemplate":{"major":601,"minor":0,"patch":0, "prereleaseIdentifier":"latest"},
```
bkhouri, line 2: question: how is tie config.json use?

dschaefer2: It's used with the swift package init --type macro.

dschaefer2: I guess that lets us override the hardcoded default in the code by updating that file in the toolchain.

PR #8324, opened 2025-03-03
update Swift-driver dependency

bkhouri:
The swift-driver dependency was refering to the Apple organization. The swift-driver repo is now in swiftlang


bkhouri:
@swift-ci please test

PR #8327, opened 2025-03-05
swift run: Speed up fd closes on macOS/Linux for high rlimits

dcantah:
Today, for everything that isn't the BSDs, we grab the maximum number of open fds the process supports and then loop through from 3 -> max, calling close(2) on everything. Even for a low open fd count of 65k this will typically result in 99+% of these close's being EBADF. At a 65k nofile RLIMIT the sluggishness is not really felt, but if you start raising this it becomes much more apparent. `swift run` on a hello world program can take 15 seconds/minutes before the program is actually ran depending on what fs.nr_open is + soft rlimit on linux.

There's a couple ways to work around this:

1. For Linux, on kernels 5.9 and up we have a handy friend named close_range, which is similar to closefrom on the BSDs which we already use.
2. If close_range doesn't exist we can read /proc/self/fd and close everything manually above 3. It seems from some trial runs there's really not all too many fds we hold open, so this is around the same runtime as close_range. A similar avenue exists on macOS in /dev/fd that we can use as well.

This change employs both of them for Linux, and only /dev/fd for macOS. For Linux, we will try close_range first and if we get -1 (either for ENOSYS as the syscall doesn't exist, or for any other error) we'll fallback to the /proc method.

Below is the delta between two runs of `swift run` on a simple hello world program on an m1 max. The shell I'm running these in has a nofile rlimit of 1 billion for an extreme example. At 100 million it falls to about 20 seconds on my machine, 10 million is around 3 seconds and then gets progressively smaller until the two approaches aren't really any different at all.

With the patch:
```
Build of product 'closerange' complete! (0.17s)
Hello, world!

real	0m0.865s
user	0m0.702s
sys	0m0.110s
```

Without:
```
Build of product 'closerange' complete! (0.15s)
Hello, world!

real	2m43.203s
user	0m47.357s
sys	1m55.344s
```


Sources/Commands/SwiftRunCommand.swift
```diff
@@ -331,36 +335,30 @@ public struct SwiftRunCommand: AsyncSwiftCommand {
...
+        closeFdsAbove(3)
         #endif
 
         try TSCBasic.exec(path: path, args: args)
     }
 
+    #if !os(Windows)
+    private func closeFdsAbove(_ from: UInt32) {
+        #if os(FreeBSD)
+        pthread_suspend_all_np()
```
weissi, line 356: This function is only ever called after `fork()` and pre-`execve()`, right? There should be precisely one active thread so this seems like a noop.

Sources/CExecHelpers/exec_helpers.c
```diff
@@ -0,0 +1,90 @@
...
+    struct rlimit rl;
+    if (getrlimit(RLIMIT_NOFILE, &rl) == 0) {
+        return rl.rlim_cur;
+    }
+
+    // Fallback to sysconf if our pal above didn't work.
+    return sysconf(_SC_OPEN_MAX);
+}
+
+void close_fds_from(unsigned int from) {
```
weissi, line 73: either make this `static` (and put it into the header file) or give it a name space, like `spm_close_fds_from` or so. This is rather likely to clash with other C functions in the global namespace.

Similarly for the other functions in this file.

Sources/CExecHelpers/exec_helpers.c
```diff
@@ -0,0 +1,90 @@
...
+    #elif defined(__linux__)
+    highest_fd = highest_open_fd_dir("/proc/self/fd");
+    #endif
+
+    if (highest_fd != -1) {
+        return highest_fd;
+    }
+
+    struct rlimit rl;
+    if (getrlimit(RLIMIT_NOFILE, &rl) == 0) {
```
weissi, line 65: I don't _think_ this is right because you can have file descriptor numbers above the allowed number of open files, right? E.g. I can do `dup2(12, 100000)` but the rlimit might say I can only have 20 open fds.

Sources/Commands/SwiftRunCommand.swift
dcantah: I was always somewhat confused on this. This is existing code though so didn't want to remove in case I was missing something..

Package.swift
```diff
@@ -690,6 +691,13 @@ let package = Package(
             ]
         ),
 
+        // MARK: C helpers
+
+        .target(
+            name: "CExecHelpers",
```
weissi, line 697: Swift has a global namespace for module names, so I'd suggest to give this a slightly more unique name too.

dcantah:
>Would you be able to provide the source code for such program?

@MaxDesiatov  Sure, it's just the default that `swift package init --type executable` produces today. I'll link in the description.

Sources/Commands/SwiftRunCommand.swift
```diff
@@ -331,36 +335,30 @@ public struct SwiftRunCommand: AsyncSwiftCommand {
...
-        // 2. close all file descriptors.
-        for i in 3..<number_fds {
-            close(i)
-        }
-        #endif /* os(FreeBSD) || os(OpenBSD) */
+        // 2. Close all file descriptors above stdio.
+        closeFdsAbove(3)
         #endif
 
         try TSCBasic.exec(path: path, args: args)
```
weissi, line 350: I realise that doesn't originate from your new code but I think it's super dangerous to run Swift between `fork()` and `execve()`. We're essentially in a broken process, so this might just hang. For example this situation could happen:

- program runs happily
- Thread N acquires some lock (e.g. an allocator lock or a Swift runtime lock)
- `fork()` gets called on thread M
- new process starts, continuing to run thread M exclusively
- the Swift code here allocates or calls into the Swift runtime
- can't acquire the lock because it's marked as already in use and will never be freed because the acquisition happened in the old process.

Sources/Commands/SwiftRunCommand.swift
weissi: wow, that is weird. Tbh I would still err on the side of removing the code. If it does anything, we have a huge issue.

Sources/CExecHelpers/exec_helpers.c
dcantah: I don't think Linux (macOS I can test) allows a dup2 higher than the soft limit for nofile.

dcantah: RLIMIT_NOFILE
              This specifies a value one greater than the maximum file
              descriptor number that can be opened by this process.

The definition states that it's actually related to the maximum fd number that can be opened in general.

Sources/Commands/SwiftRunCommand.swift
dcantah: Yea, fork cludgery in a runtime based language is always tricky. Fwiw we do allocate in opendir, although we could swap to getdents64 and whatever the equivalent is on macOS.

dcantah: Y'know, I just took a look outside of this code, and I don't even see anywhere that fork takes place. spm might just be doing execve directly with no fork?

Sources/CExecHelpers/exec_helpers.c
weissi: Ha, interesting, thanks. Wonder what happens if you lower rlimit to less than your current max. But this is more out of interest, don't feel like you need  to change the code, this seems correct enough

Sources/Commands/SwiftRunCommand.swift
dcantah: https://github.com/swiftlang/swift-package-manager/pull/8327/files#r1981142987

dcantah: This might make sense if the above is true ^

Sources/CExecHelpers/exec_helpers.c
dcantah: Ack

Sources/Commands/SwiftRunCommand.swift
weissi: Iiinteresting, yes, for `swift run` that's actually possible. Of course that theoretically opens up a whole can of works of course because we're multi-threaded. So all the signal masking/resetting stuff is technically racy unless we do the equivalent of `pthread_suspend_all_np()`.

But of course, none of this is the issue of this PR. Maybe leave a bunch of comments about what we're aiming to do and why the suspension of the other threads is now necessary (but harder to do in at least Linux which doesn't have `pthread_suspend_all_np` from what I know)...

Sources/Commands/SwiftRunCommand.swift
weissi: After your latest findings: Ignore this, maybe move it out of the `closeFdsAbove` function higher up (alongside a comment) because it seems important now.

Plus, tiny nit: Swift style guidelines would have `closeFDsAbove` because "fd" is an abbreviation so it's `FD` unless it's the start of a word (e.g. `fdNumber` but `highestFD`)

jakepetroules:
@dcantah Why not switch all of this over to use `Process` (which wraps `posix_spawn`)?

dcantah:
@jakepetroules I can trial! I wasn't sure if there was any reason for the more manual approach that was taken here today so didn't want to depart too far. It seems Process (if I'm reading the right code) does a lot of the same tricks. I guess the main departure for that would be `swift run` would not be replaced with the users binary, but rather be a child of it if we used Foundation.Process. Does anyone know if that would cause issues or we relied on the current behavior today?

jakepetroules:
> I guess the main departure for that would be `swift run` would not be replaced with the users binary, but rather be a child of it if we used Foundation.Process. Does anyone know if that would cause issues or we relied on the current behavior today?

Asked around and nobody could think of a specific reason.

dcantah:
@jakepetroules Sweet! I'll move this over to `Process` and see if that's a good replacement

MaxDesiatov:
There are `swift run --debugger` and `swift run --repl` modes, those could in theory break if debugger attachment logic relies on how it is currently set up.

dcantah:
@MaxDesiatov Good point. I'm somewhat concerned on changing it to be honest the more I look at this. The other thing is if we went with `Process` we'd need to forward signals from `swift run` to the child now which adds some complexity.

jakepetroules:
> There are `swift run --debugger` and `swift run --repl` modes, those could in theory break if debugger attachment logic relies on how it is currently set up.

Good point, thanks for that clarification, @MaxDesiatov.

We might want to add a paraphrased version of your reply as a code comment so it's clear from reading the sources _why_ it may be using the lower-level APIs here.

Sources/Commands/SwiftRunCommand.swift
dcantah: Added a lengthy comment on the original reason this was added (dug through the history, the change was recent also)

Sources/CExecHelpers/exec_helpers.c
dcantah: Changed to spm_close_fds_from!

dcantah:
Makes me wonder though, if someone had recently introduced that pthread_suspend_all_np [to avoid a race with libdispatch](https://github.com/swiftlang/swift-package-manager/issues/8197), and we'd like to stay with just raw execve to keep things chugging along, should we just CLOEXEC everything instead of closing? Presumably we'd avoid the race conditions detailed with closing the kqueue fds from dispatch as we'd just let the system do it during exec for us.

dcantah:
@swift-ci please test

dschaefer2:
> Updated. Makes me wonder though, if someone had recently introduced that pthread_suspend_all_np [to avoid a race with libdispatch](https://github.com/swiftlang/swift-package-manager/issues/8197), and we'd like to stay with just raw execve to keep things chugging along, should we just `fcntl(...CLOEXEC)` everything instead of closing? Presumably we'd avoid the race conditions detailed with closing the kqueue fds from dispatch as we'd just let the system do it during exec for us.

Sorry I missed this PR while working on getting `swift run` to work with Swift Build. In my PR, #8331, I actually did switch it to CLOEXEC. Closing the fd's was causing havoc with the Swift Build shutdown and crashing SwiftPM.

dcantah:
@dschaefer2 Makes sense. I'll swap to just /dev/fd or /proc/self/fd + cloexec for all platforms then

dcantah:
@dschaefer2 I've amended this change to just read /proc/self/fd and /dev/fd on macOS/Linux and cloexec everything. I've left the BSDs as is though, as I truly don't know what the right thing to do on them is, but it seems some folks here recently changed bits there already to make things better. It seems FreeBSD is probably fine with closefrom as we suspend every thread, but OpenBSD's in a weird spot. I'm open to suggestions.

dcantah:
@swift-ci please test

PR #8328, opened 2025-03-05
Remove trailing comma from `downloadAndPopulateCache`

ahoppen:
This restores the ability to build SwiftPM using a Swift 6.0 compiler.


ahoppen:
@swift-ci Please test

plemarquand (approved):
Whoops, sorry, this was my fault. Is there a way we can catch this kind of thing in CI going forward?

MaxDesiatov:
This is caught by CI in other repositories. IIUC jobs on this one now always use the latest toolchain, which doesn't allow testing for compatibility with older toolchains without setting up additional jobs.

jakepetroules (approved):

PR #8334, opened 2025-03-06
Include Swift Build support in second stage bootstrap builds

owenv:
Stop setting SWIFTPM_NO_SWBUILD_DEPENDENCY in the bootstrap script - we've made changes to Swift Build which should allow it to be bootstrapped in these environments (it's still not part of swift-bootstrap itself though).

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -146,6 +146,12 @@ public struct LocationOptions: ParsableArguments {
         completion: .directory
     )
     public var pkgConfigDirectories: [AbsolutePath] = []
+    
+    @Option(
+        help: .init("Specify alternate path to search for SwiftPM resources.", visibility: .hidden),
+        completion: .directory
+    )
+    public var packageManagerResourcesDirectory: AbsolutePath?
 
     @Flag(name: .customLong("ignore-lock"), help: .hidden)
     public var ignoreLock: Bool = false
--- a/Utilities/bootstrap
+++ b/Utilities/bootstrap
@@ -809,7 +809,6 @@ def get_swiftpm_env_cmd(args):
 
     if args.llbuild_link_framework:
         env_cmd.append("SWIFTPM_LLBUILD_FWK=1")
-    env_cmd.append("SWIFTPM_NO_SWBUILD_DEPENDENCY=1")
     env_cmd.append("SWIFTCI_USE_LOCAL_DEPS=1")
     env_cmd.append("SWIFTPM_MACOS_DEPLOYMENT_TARGET=%s" % g_macos_deployment_target)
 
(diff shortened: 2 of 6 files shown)


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test macOS

owenv:
@swift-ci test macOS

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:

owenv:
@swift-ci test

Utilities/bootstrap
```diff
@@ -518,6 +518,26 @@ def install_swiftpm(prefix, args):
     dest = os.path.join(prefix, "lib", "swift", "pm", "PluginAPI")
     install_dylib(args, "PackagePlugin", dest, ["PackagePlugin"])
 
+    # Install required resource bundles
+    resource_bundles = [
```
jakepetroules, line 522: No way we can keep this part in the swift-build repo? Will be easy to forget to update it...

owenv: changed it to enumerate the build products for resource bundles instead of maintaining a list

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test Linux

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

dschaefer2 (approved):

Sources/CoreCommands/Options.swift
```diff
@@ -146,6 +146,12 @@ public struct LocationOptions: ParsableArguments {
         completion: .directory
     )
     public var pkgConfigDirectories: [AbsolutePath] = []
+    
+    @Option(
+        help: .init("Specify alternate path to search for SwiftPM resources.", visibility: .hidden),
+        completion: .directory
+    )
+    public var packageManagerResourcesDirectory: AbsolutePath?
```
cmcgee1024, line 154: question: Should this be added to the command-line options in `swift-bootstrap/main.swift` too so that we can bootstrap building SwiftPM itself with the swiftbuild system someday?

Utilities/bootstrap
```diff
@@ -809,7 +814,6 @@ def get_swiftpm_env_cmd(args):
 
     if args.llbuild_link_framework:
         env_cmd.append("SWIFTPM_LLBUILD_FWK=1")
-    env_cmd.append("SWIFTPM_NO_SWBUILD_DEPENDENCY=1")
```
cmcgee1024, line 812: thought: I wonder if the smoke test builds will pass with the dependency.

Sources/CoreCommands/Options.swift
owenv: Swift Build still has a fallback to the resources directory in the build directory, which I think will cover this case

cmcgee1024 (approved):
This looks good to me. I posted a question and thought.

Utilities/bootstrap
owenv: I was able to get an earlier run passing, the latest test run just has minor tweaks so I think it'll work

PR #8336, opened 2025-03-06
Attempt to fix build by reverting #8193 and related commits

kperryua:
Despite all reason, these changes appear to be a candidate for the cause of the current linux build breakage. Putting up a PR so we can test the theory.


jakepetroules:
Closing since this didn't turn out to be the case.

PR #8340, opened 2025-03-07
Remove trailing comma in RegistryDownloadsManager.swift

ahoppen:
This restores the ability to build SwiftPM using a Swift 6.0 compiler.

### DIFF
--- a/Sources/PackageRegistry/RegistryDownloadsManager.swift
+++ b/Sources/PackageRegistry/RegistryDownloadsManager.swift
@@ -98,7 +98,7 @@ public class RegistryDownloadsManager: AsyncCancellable {
                             packagePath: packagePath,
                             observabilityScope: observabilityScope,
                             delegateQueue: delegateQueue,
-                            callbackQueue: callbackQueue,
+                            callbackQueue: callbackQueue
                         )
                         // inform delegate that we finished to fetch
                         let duration = start.distance(to: .now())


ahoppen:
@swift-ci Please test

bkhouri (commented):
All CI jobs in Swift PM use the nightly toolchain.   As such, I feel SwiftPm should support the nightly toolchain. 

Then again, with a lack of a .swift-version, there is nothing in the repository that stipulate the supported Swift toolchain version to build Swift package Manager.

ahoppen:
I think it would be very unfortunate if you couldn’t contribute to SwiftPM (and thus also SourceKit-LSP) with the latest released Swift version, so I think we should make an effort to keep SwiftPM building using the latest released Swift version.

bnbarham (approved):
To be clear, this doesn't stop supporting the nightly toolchain, it just allows us to also build with the released toolchain. We should very much keep the ability to do so, it makes it very difficult to contribute to the project otherwise.

plemarquand (approved):

jakepetroules (approved):

MaxDesiatov:
> Then again, with a lack of a .swift-version, there is nothing in the repository that stipulate the supported Swift toolchain version to build Swift package Manager.

There is, like for any other Swift package it's the tools version stanza on the first line of `Package.swift`: `// swift-tools-version:5.9`, which stipulates that SwiftPM should be buildable with Swift 5.9.

plemarquand:
@MaxDesiatov As of relatively recently swift-package-manager now depends on swift-build, which requires 6.0. I think we should probably update the tools version in Package.swift.

PR #8348, opened 2025-03-10
Remove code that supports < Swift 6.0

plemarquand:
`swift-build` is a dependency of swift-package-manager that requires >= Swift 6.0 to build. SPM also requires swift-testing for tests, which requires >= Swift 6.0.

Clean up legacy code that skipped tests on 5.10 and below, as well as conditional compilation directives on imports.


plemarquand:
@swift-ci please test

bripeticca (approved):

PR #8353, opened 2025-03-11
Remove unnecessary `@retroactive` conformances

plemarquand:
Several extensions to conform to common Swift protocols were unnecessarily marked as `@retroactive`, causing compiler warnings. The `@retroactive` annotation is only necessary when the type being extended is defined outside the package.


plemarquand:
@swift-ci please test

MaxDesiatov (approved):

PR #8357, opened 2025-03-11
Fix call to SWBBuildService.createSession

plemarquand:
https://github.com/swiftlang/swift-build/pull/266 changed the API of SwiftBuildService.createSession.

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -52,7 +52,14 @@ func withSession(
         _ diagnostics: [SwiftBuild.SwiftBuildMessage.DiagnosticInfo]
     ) async throws -> Void
 ) async throws {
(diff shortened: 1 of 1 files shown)


plemarquand:
@swift-ci please test

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -52,7 +52,14 @@ func withSession(
         _ diagnostics: [SwiftBuild.SwiftBuildMessage.DiagnosticInfo]
     ) async throws -> Void
 ) async throws {
-    switch await service.createSession(name: name, resourceSearchPaths: packageManagerResourcesDirectory.map { [$0.pathString] } ?? [], cachePath: nil, inferiorProductsPath: nil, environment: nil) {
+    switch await service.createSession(
+        name: name,
+        developerPath: nil,
```
owenv, line 57: developerPath defaults to nil, so it shouldn't be needed here. Were you seeing the build fail in some configuration?

plemarquand: I think my dependencies got in to a bad state, I switched to a known good toolchain, cleaned and rebuilt and its not an issue.

PR #8364, opened 2025-03-12
Add toolchain swift stdlib to env when swift run.

dschaefer2:
Swift Build doesn't set the rpath of the generated executable to point at the toolchains swift standard library. We shouldn't be hardcoding those paths anyway since that restricts deployment options. Instead we point the LD_LIBRARY_PATH at swift run time to pick them up.

### DIFF
--- a/Sources/Commands/SwiftRunCommand.swift
+++ b/Sources/Commands/SwiftRunCommand.swift
@@ -11,7 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import ArgumentParser
-import Basics
+@_spi(SwiftPMInternal) import Basics
 import CoreCommands
 import Foundation
 import PackageGraph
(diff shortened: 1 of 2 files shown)


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

Sources/Commands/SwiftRunCommand.swift
```diff
@@ -242,11 +244,23 @@ public struct SwiftRunCommand: AsyncSwiftCommand {
                     arguments = options.arguments
                 }
 
+                // For Linux, need to point LD_LIBRARY_PATH at the swift runtime
+                let environment: [String: String]?
+                if toolchain.targetTriple.isLinux() {
```
jakepetroules, line 249: nit: could be `if toolchain.targetTriple.objectFormat == .elf` to cover other relevant platforms as well

(can also be a follow-on change)

jakepetroules (approved):

PR #8366, opened 2025-03-13
Revert "Include Swift Build support in second stage bootstrap builds"

owenv:
Reverts https://github.com/swiftlang/swift-package-manager/pull/8334

This seems to cause an issue in the incremental buildbots


owenv:
@swift-ci test

PR #8369, opened 2025-03-13
add support to TarArchiver for more common compression suffixes

toffaletti:
- include more suffixes for: gzip, bzip2, lzma, xz

### Motivation:

Add support for compression algorithms that can achieve higher compression ratios than gzip.
Support another common extension for gzip as well.

### Modifications:

I modified the list of supported extensions for the TarArchiver.

I researched various manpages to see what was commonly supported.

- The [GNU tar manpage](https://www.gnu.org/software/tar/manual/html_section/Compression.html) has the most extensive list of suffixes.
- [FreeBSD tar](https://man.freebsd.org/cgi/man.cgi?tar(1)) i
- [OpenBSD tar](https://man.openbsd.org/tar) doesn't support the `a` flag used for auto-compress which `TarArchiver.compress` uses [ref](https://github.com/swiftlang/swift-package-manager/pull/6368/files#r1156324814). It also doesn't seem to have anything beyond gzip and bzip2 (no lzma or xz).
- [NetBSD tar](https://man.netbsd.org/tar.1) has gzip, bzip2, lzma, xz
- [Windows tar.exe](https://ss64.com/nt/tar.html) seems to be a port of BSD tar similar in support to NetBSD.
- macOS tar is an older FreeBSD tar, similar support to NetBSD.

### Result:

More compression algorithms will be supported.

### DIFF
--- a/Sources/Basics/Archiver/TarArchiver.swift
+++ b/Sources/Basics/Archiver/TarArchiver.swift
@@ -16,7 +16,14 @@ import struct TSCBasic.FileSystemError
 
 /// An `Archiver` that handles Tar archives using the command-line `tar` tool.
 public struct TarArchiver: Archiver {
-    public let supportedExtensions: Set<String> = ["tar", "tar.gz"]
+
+    public let supportedExtensions: Set<String> = [
+        "tar",
+        "gz", "tgz", // gzip
+        "bz2", "tz2", "tbz2", "tbz", // bzip2
+        "lzma", // lzma
+        "xz" // xz
+    ]
 
     /// The file-system implementation used for various file-system operations and checks.
     private let fileSystem: FileSystem


toffaletti:
I wasn't sure about adding test files for all these extensions?

Sources/Basics/Archiver/TarArchiver.swift
```diff
@@ -16,7 +16,14 @@ import struct TSCBasic.FileSystemError
 
 /// An `Archiver` that handles Tar archives using the command-line `tar` tool.
 public struct TarArchiver: Archiver {
-    public let supportedExtensions: Set<String> = ["tar", "tar.gz"]
```
toffaletti, line 19: It wasn't clear to me whether "tar.gz" is really needed or if "gz" is good enough here? The `tar` commands don't seem to care:

```
 tar tzf test.gz | head
6.0.3-RELEASE_rhel_ubi9_x86_64.artifactbundle/
6.0.3-RELEASE_rhel_ubi9_x86_64.artifactbundle/6.0.3-RELEASE_rhel_ubi9_x86_64/
6.0.3-RELEASE_rhel_ubi9_x86_64.artifactbundle/info.json
6.0.3-RELEASE_rhel_ubi9_x86_64.artifactbundle/6.0.3-RELEASE_rhel_ubi9_x86_64/x86_64-unknown-linux-gnu/
```

dschaefer2:
> I wasn't sure about adding test files for all these extensions?

Tests would be extremely helpful. It's hard to know whether these things work or will stay working without them.

MaxDesiatov:
Another thing to keep in mind that not all of these formats are supported by `tar` on all platforms that SwiftPM supports. We need to make sure that appropriate diagnostic message is emitted to the user and that this support is detected appropriately on a given platform.

toffaletti:
> Another thing to keep in mind that not all of these formats are supported by `tar` on all platforms that SwiftPM supports. We need to make sure that in case of unsupported formats appropriate diagnostic message is emitted to the user and that this support is detected correctly on a given platform.

Is there a canonical list of supported platforms and their versions?

MaxDesiatov:
> Is there a canonical list of supported platforms and their versions?

Yes, it's available at https://swift.org/install

tbkka:
There are many, many different tar implementations on many different platforms.  If you want consistent cross-platform behavior, you'll need to implement this yourself.   A few notes:

* GNU tar only detects the compression and format when reading from a file on disk.  It can't auto-detect when reading from stdin.
* bsdtar can auto-detect from any input source.
* Few other tar implementations do any auto-detection at all.  So for portability, you'll need to pass appropriate command-line flags to tar.
* The specific compression formats supported by either of the above depends on how it was built and what platform it's running on.  Both will attempt to run command-line `gzip`, `xz`, etc to handle decompression, which means that it requires the appropriate command-line tool to be installed.   bsdtar can be built to use compression libraries for fully internal compression, but those are all optional.  So the only way to know for sure is to actually try feeding something to tar -- you can't just look at the OS version and know anything useful.  (In particular, different Linux distros vary widely in what tar implementation they include by default.  Many use GNU tar, but by no means all.)
* Windows only recently started including `tar` as a standard command-line utility.  Older Windows systems lack this.

PR #8370, opened 2025-03-13
[Traits] Change `TraitConfiguration` to be an enum instead of struct

### LINKED ISSUE #8355: Change `TraitConfiguration` struct to an enum
### Description

Currently, the `TraitConfiguration` struct describes a user-designated trait configuration wherein the user can decide which combination of traits are enabled/disabled for a given package. The way this struct is currently designed allows for an optional set of enabled traits, as well as a `bool` that describes the flag to `enableAllTraits`. Since there are distinct use cases depending on whether or not the enabled traits are `nil`, empty, or contains a list of traits, it makes more sense to treat this as an enum:

```swift
enum TraitConfiguration: Codable, Hashable { 
    case enableAllTraits // when the enableAllTraits flag is set, since this will simply enable everything
    case noConfiguration // when there isn't a configuration set, meaning that the default traits will be used
    case noTraits // when there is a configuration set and the set of enabled traits is empty, which overrides the default traits 
    case traits(Set<String>) // when there is a configuration set and there is a set of traits present
}
```

### Expected behavior

This will make the checks done on `TraitConfiguration` to be more ergonomic, as there would be a defined set of cases that are explicitly named and would omit redundant checks.

### Actual behavior

The current use of `TraitConfiguration` can be a bit painful - there are plenty of condition checks involving repeated terms (i.e. `if let enabledTraits = traitConfiguration.enabledTraits, traitConfiguration.enableAllTraits == false { ... }` and other variations of this) that can be condensed for ease of use.

### Steps to reproduce

_No response_

_No response_

_No response_

### LINKED ISSUE #8356: [Traits] Integrate 'disallow disabling default traits of a package without traits' into resolution phase

### Description

The PR #8326 has been successfully implemented in the modules graph loading phase, where traits were initially supported. Now that #8205 has been merged, this feature should be considered earlier on in resolution.

### Expected behavior

The check for disallowing disabling default traits for a package without traits should be checked during resolution to omit any further calculations/checkouts/etc. if this condition fails.

### Actual behavior

This check is being done after resolution has already completed.

### Steps to reproduce

_No response_

_No response_

_No response_

bripeticca:
Fixes #8355.

Since the way a `TraitConfiguration` is set up can affect how traits are calculated and whether default traits are overridden or not, it's more ergonomic to change this to an `enum` that lists the various possible cases of configuration:

 - No trait configuration, which defaults to using default traits if applicable
 - A configuration with an empty set of traits that overrides the default traits
 - A configuration that enables all traits
 - A configuration that specifies an exact set of traits that are enabled

This way we can avoid doing multiple checks on the properties of the `struct` and instead rely on the exact case that describes how to continue calculating enabled traits.

The refactoring done as a part of this change also fixes #8356.


Sources/PackageGraph/TraitConfiguration.swift
```diff
@@ -11,18 +11,96 @@
...
+    /// Default instance of `TraitConfiguration`.
+    public static var `default`: TraitConfiguration {
+        .init(enabledTraits: nil)
     }
+
+    public var enabledTraits: Set<String>? {
+        switch self {
+        case .enabledTraits(let traits):
+            return traits
+        case .noConfiguration, .enableAllTraits:
```
bkhouri, line 52: thought: it seems odd to me that the `.enableAllTraits` returns `nil`.  Should it return a set with all the traits instead?

Sources/PackageGraph/TraitConfiguration.swift
```diff
@@ -11,18 +11,96 @@
...
 /// The trait configuration.
-public struct TraitConfiguration: Codable, Hashable {
-    /// The traits to enable for the package.
-    package var enabledTraits: Set<String>?
-
-    /// Enables all traits of the package.
-    package var enableAllTraits: Bool
+public enum TraitConfiguration: Codable, Hashable {
+    case enableAllTraits
+    case noConfiguration
```
bkhouri, line 16: question: what does `noConfiguration` represent?  how is it different than the `noEnabledTraits`?  is there an actual behavioural difference between `noConfiguration` and `noEnabledTraits`?

Sources/PackageGraph/TraitConfiguration.swift
```diff
@@ -11,18 +11,96 @@
...
+        if let enabledTraits {
+            if enabledTraits.isEmpty {
+                self = .noEnabledTraits
+            } else {
+                self = .enabledTraits(enabledTraits)
+            }
+        } else {
+            // Since enableAllTraits isn't enabled and there isn't a set of traits set,
+            // there is no configuration passed by the user.
+            self = .noConfiguration
```
bkhouri, line 39: question: could we have also set `.noEnabledTraits` to get the same behaviour?

bkhouri (commented):
I have initial question on the enum `TraitConfiguration`.

Sources/PackageGraph/TraitConfiguration.swift
bripeticca: There is, yes: I've since changed the nomenclature for these cases and will be pushing that up today, but essentially `noConfiguration` encapsulates the case where there is no configuration of traits passed in by the user and as such will default to using a package's default traits if there are any. If a user _does_ explicitly pass in a configuration with an empty set of traits enabled (so essentially, no traits enabled), this overrides the default traits behaviour and instead disables all traits which is what `.noEnabledTraits` accomplishes. :)

Sources/PackageGraph/TraitConfiguration.swift
bripeticca: Since `.enableAllTraits` represents a flag that a user would pass in (and as such the `TraitConfiguration` is constructed before we begin parsing the `Manifest` which is where we'd have access to the entire list of traits that a manifest defines), that was the reasoning for it returning `nil` at this stage. Though since the `enum` now allows me to do specific checks for these cases where this information would be relevant, there may not be a need for this computed property - haven't fully refactored everything just yet.

bripeticca: This has made me notice that it would probably make more sense to have `.noConfiguration` (or rather `.none` as the updated case name) return `["default"]` here. :)

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

bripeticca:

bripeticca:
@swift-ci  please test self hosted windows

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

Sources/PackageModel/Manifest/TraitConfiguration.swift
```diff
@@ -0,0 +1,64 @@
...
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// The trait configuration.
+public enum TraitConfiguration: Codable, Hashable {
+    case enableAllTraits
+    case disableAllTraits
+    case enabledTraits(Set<String>)
+    case none
```
FranzBusch, line 18: Should we call this `default` instead?

FranzBusch (approved):
LGTM to me in general. Just a small naming question

Sources/PackageModel/Manifest/TraitConfiguration.swift
bripeticca: Yup! Renaming it to default will also get rid of the need for the static variable as well.

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

bripeticca:
@swift-ci please smoke test linux platform

bripeticca:

bripeticca:
@swift-ci please test linux

dschaefer2 (approved):

PR #8371, opened 2025-03-14
Use `@main` instead of top level code when initializing executable

plemarquand:
### Motivation:

Since `@main` is the general purpose way of marking the entry point to executables, use it over top level code when generating an executable template.

### Modifications:

Update the template output when the specified `--type` is executable.

### Result:

The template is now:

```swift
@main
struct MyProject {
  static func main() {
    print("Hello, world!")
  }
}
```

This patch reverts https://github.com/swiftlang/swift-package-manager/pull/6197 and partially reverts https://github.com/swiftlang/swift-package-manager/pull/6144

### DIFF
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -577,7 +577,7 @@ public final class InitPackage {
         let moduleDir = sources.appending("\(pkgname)")
         try makeDirectories(moduleDir)
 
-        // If we're creating an executable we can't 't have both a @main declaration and a main.swift file.
+        // If we're creating an executable we can't have both a @main declaration and a main.swift file.
         // Handle the edge case of a user creating a project called "main" by give the generated file a different name.
         let sourceFileName = ((packageType == .executable || packageType == .tool) && typeName == "main") ? "MainEntrypoint.swift" : "\(typeName).swift"
         let sourceFile = try AbsolutePath(validating: sourceFileName, relativeTo: moduleDir)
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -581,7 +581,7 @@ public final class InitPackage {
 
         // If we're creating an executable we can't 't have both a @main delcaration and a Main.swift file.
         // Handle the edge case of a user creating a project called "Main" by give the generated file a different name.
-        let sourceFileName = (packageType == .executable && typeName.lowercased() == "main") ? "MainEntrypoint.swift" : "\(typeName).swift"
+        let sourceFileName = (packageType == .executable && typeName == "main") ? "MainEntrypoint.swift" : "\(typeName).swift"
         let sourceFile = try AbsolutePath(validating: sourceFileName, relativeTo: moduleDir)
 
         let content: String
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -579,12 +579,7 @@ public final class InitPackage {
         }
         try makeDirectories(moduleDir)
 
-        let sourceFileName: String
-        if packageType == .executable {
-            sourceFileName = "main.swift"
-        } else {
-            sourceFileName = "\(typeName).swift"
-        }
+        let sourceFileName = "\(typeName).swift"
         let sourceFile = try AbsolutePath(validating: sourceFileName, relativeTo: moduleDir)
 
         let content: String
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -597,7 +597,7 @@ public final class InitPackage {
 
                 @main
                 struct \(typeName) {
-                    static func main() {
+                    static func main() async throws {
                         print("Hello, world!")
                     }
                 }
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -600,7 +600,12 @@ public final class InitPackage {
                 // The Swift Programming Language
                 // https://docs.swift.org/swift-book
 
-                print("Hello, world!")
+                @main
+                struct \(typeName) {
+                    static func main() {
+                        print("Hello, world!")
+                    }
+                }
 
                 """
         case .tool:
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -239,7 +239,8 @@ public final class InitPackage {
                         // Products define the executables and libraries a package produces, making them visible to other packages.
                         .library(
                             name: "\(pkgname)",
-                            targets: ["\(pkgname)"]),
+                            targets: ["\(pkgname)"]
+                        ),
                     ]
                 """)
             } else if packageType == .buildToolPlugin || packageType == .commandPlugin {
@@ -248,7 +249,8 @@ public final class InitPackage {
                         // Products can be used to vend plugins, making them visible to other packages.
                         .plugin(
                             name: "\(pkgname)",
-                            targets: ["\(pkgname)"]),
+                            targets: ["\(pkgname)"]
+                        ),
                     ]
                 """)
             } else if packageType == .macro {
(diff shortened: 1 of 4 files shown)


plemarquand:
@swift-ci test

MaxDesiatov (commented):
This was changed to use top-level code instead of `@main` not so long ago in https://github.com/swiftlang/swift-package-manager/pull/6144, what's the reason for reverting?

plemarquand:
@swift-ci test

dschaefer2:
Yeah, this is a pretty major change and not very new user friendly. Maybe this is a separate template?

dnadoba:
Top level code has various bugs that make it [very easy to write code that miscompiles](https://forums.swift.org/t/on-the-behavior-of-variables-in-top-level-code/52230) e.g.:
```swift
print(x)
var x: Set = ["Hello World"]
```
which crashes at runtime because x isn't initialized before we attempt to print it.

This isn't a great user experience. I think until this is sorted we should prefer an `@main` entry point.

bitjammer:
I generally support this change. There should ideally be only *one* way to define a main entry point, and it should be an explicit form. If top-level code is truly problematic, then we should remove it from the language and be done with it. It would be nice if we could put `@main` on a bare function, since the function within the structure is static.

I would ask that you consider naming the file/structure something that reflects the target, package, or a noun that reflects what it is, rather than just `Main`.

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

bitjammer (approved):

dnadoba (approved):

plemarquand:
@swift-ci test windows

IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift
```diff
@@ -51,7 +51,7 @@ final class SwiftPMTests: XCTestCase {
 
     func testSwiftBuild() throws {
         #if os(Linux)
```
bkhouri, line 53: suggestion: can we add a test where we try to instantiate an executable wth the name `main` ?  

something like 
```
swift package --package-path main init --type executable
swift run  or
```

plemarquand: Good idea, I've updated the PR to handle this edge case and added a new test.

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test windows

bkhouri (approved):

dschaefer2:
Another aspect of this template, if you are going to @main the main, then you should probably also create the module in a subdirectory of Sources.

I believe that's why the current template is like it is, this is a different use case. The original template is actually an executable package. There is no easy mechanism to add a second module, as you well know.

Which is a back to my point that this, where you're adding a module that's an executable is a different template.

Sources/Workspace/InitPackage.swift
```diff
@@ -579,12 +579,9 @@ public final class InitPackage {
...
 
-        let sourceFileName: String
-        if packageType == .executable {
-            sourceFileName = "main.swift"
-        } else {
-            sourceFileName = "\(typeName).swift"
-        }
+        // If we're creating an executable we can't 't have both a @main delcaration and a Main.swift file.
+        // Handle the edge case of a user creating a project called "Main" by give the generated file a different name.
+        let sourceFileName = (packageType == .executable && typeName.lowercased() == "main") ? "MainEntrypoint.swift" : "\(typeName).swift"
```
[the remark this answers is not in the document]
plemarquand: Good catch, I've updated the PR

plemarquand:
@swift-ci test

Sources/Workspace/InitPackage.swift
```diff
@@ -579,12 +579,9 @@ public final class InitPackage {
...
         try makeDirectories(moduleDir)
 
-        let sourceFileName: String
-        if packageType == .executable {
-            sourceFileName = "main.swift"
-        } else {
-            sourceFileName = "\(typeName).swift"
-        }
+        // If we're creating an executable we can't 't have both a @main delcaration and a Main.swift file.
+        // Handle the edge case of a user creating a project called "Main" by give the generated file a different name.
```
dnadoba, line 583:
```suggestion
        // If we're creating an executable we can't 't have both a @main declaration and a main.swift file.
        // Handle the edge case of a user creating a project called "main" by give the generated file a different name.
```

Sources/Workspace/InitPackage.swift
```diff
@@ -579,12 +579,9 @@ public final class InitPackage {
...
+        let sourceFileName = (packageType == .executable && typeName == "main") ? "MainEntrypoint.swift" : "\(typeName).swift"
```
dnadoba, line 584: How about just using uppercase `Main.swift` instead of `MainEntrypoint.swift` if we already uppercase `main` anyway? Don't feel too strongly about it.

dnadoba (approved):
Thanks! LGTM otherwise

Sources/Workspace/InitPackage.swift
plemarquand: I was trying to steer people away from having a "Main.swift" in any form. The [guidance on swift.org](https://www.swift.org/getting-started/cli-swiftpm/#a-small-application) states `You can either have a main.swift file or a @main entrypoint, but not both.`, and it would be easy to miss that `main.swift` and `Main.swift` are different.

Also, its my understanding that Windows filenames are typically case insensitive. I haven't tested whether this impacts how the compiler treats main.swift vs Main.swift, but it could factor in.

dnadoba: Fair enough!

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci please test macos

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test windows

FranzBusch:
Love this change! I think we should give the community a heads up in the forums though.

plemarquand:
@swift-ci test

plemarquand:
@FranzBusch Good call, I posted https://forums.swift.org/t/update-to-swift-package-init-executable-template/78813.

Sources/Workspace/InitPackage.swift
```diff
@@ -570,21 +574,12 @@ public final class InitPackage {
...
+        let moduleDir = sources.appending("\(pkgname)")
         try makeDirectories(moduleDir)
 
-        let sourceFileName: String
-        if packageType == .executable {
-            sourceFileName = "main.swift"
-        } else {
-            sourceFileName = "\(typeName).swift"
-        }
+        // If we're creating an executable we can't 't have both a @main declaration and a main.swift file.
```
heckj, line 580:
```suggestion
        // If we're creating an executable we can't have both a @main declaration and a main.swift file.
```

heckj (changes requested):
minor typo

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test

gavineadie:
> I've updated the PR to place the generated Source files for executables/tools into `/Sources/PkgName` instead of directly in `Sources`.

Do you mean `PkgName` literally, or the name of the Package?  If the former, I'd prefer it unabbreviated -- `PackageName`.

plemarquand:
@gavineadie Sorry, I was unclear, `PkgName` refers to the name of the package being created. So `swift package init --type executable --package-path ./Foo` would create the source file in `./Sources/Foo/Foo.swift`.

plemarquand:
@swift-ci please test windows

bripeticca (approved):

heckj (approved):

plemarquand:
@swift-ci test

plemarquand:
@swift-ci please test windows

plemarquand:

plemarquand:
@swift-ci test macOS

plemarquand:
@swift-ci please test

bkhouri:
@swift-ci please test windows

bkhouri:

PR #8379, opened 2025-03-18
Avoid testing library package with swiftbuild build system on Windows

cmcgee1024:
This section of the test will compile on Windows, but it doesn't succeed with Windows. Do an OS condition on there to avoid failures when tests are run with Windows.

### DIFF
--- a/IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift
+++ b/IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift
@@ -69,6 +69,8 @@ final class SwiftPMTests: XCTestCase {
             }
         }
 
+        #if !os(Windows)
+        // SWBINTTODO: Windows fails to link this library package due to a "lld-link: error: subsystem must be defined" error. See https://github.com/swiftlang/swift-build/issues/310
         do {
             try withTemporaryDirectory { tmpDir in
                 let packagePath = tmpDir.appending(component: "foo")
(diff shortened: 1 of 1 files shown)


cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Windows

MaxDesiatov (approved):

IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift
```diff
@@ -69,6 +69,8 @@ final class SwiftPMTests: XCTestCase {
             }
         }
 
+        #if !os(Windows)
+        // SWBINTTODO: Windows fails to link this library package due to a "lld-link: error: subsystem must be defined" error. See https://github.com/swiftlang/swift-build/issues/310
```
bkhouri, line 73: suggestion: Instead of compile this out of the test, can we move this to it's own test and mark it skipped on windows?

jakepetroules (approved):

IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift
jakepetroules: Let's do that in a separate change.

cmcgee1024:
@swift-ci test Windows

dschaefer2 (approved):

PR #8381, opened 2025-03-18
Remove ALTERNATE_LINKER build setting when building for Windows

cmcgee1024:
The default linker for Windows in swift-build is now lld-link so this is no longer needed.


cmcgee1024:
@swift-ci test Windows

dschaefer2 (approved):

PR #8383, opened 2025-03-19
PluginAPI: Optionally animate build progress in console

kateinoigakukun:
Introduce a way to animate build progress in console triggered by PackagePlugin API.

### Motivation:

When a package plugin asks SwiftPM to build a target, SwiftPM captures the whole build output including the progress messages and compiler diagnostics into a non-TTY output buffer stream. Use of the non-TTY output buffer stream makes the `[n/N] Compiling ...` progress messages always to be newlined. This is not ideal when building a large scale project as it consumes a lot of space in the terminal buffer.

### Modifications:

This commit introduces a new parameter `progressToConsole` to the `BuildParameters` to allow the package plugin to decide whether to write the progress messages to the console or capture them into `logText` buffer.

Compiler diagnostics are still "tee"ed and they are not colored and are included in `logText`.

### Result:

Package plugin developers will be able to provide a better build output to plugin users.

| `progressToConsole: false` | `progressToConsole: true` |
|:---------------------------:|:------------------------:|
| <video src="https://github.com/user-attachments/assets/6dae08dc-e53d-43c7-b574-9db3354be185" /> |  <video src="https://github.com/user-attachments/assets/c0502a42-36b6-4f90-9cb9-21de56e13a72" /> |

### DIFF
--- a/Sources/PackagePlugin/PackageManagerProxy.swift
+++ b/Sources/PackagePlugin/PackageManagerProxy.swift
@@ -62,6 +62,12 @@ public struct PackageManager {
         /// Whether to print build logs to the console
         public var echoLogs: Bool
 
+        /// Whether to print build progress to the console
+        ///
+        /// If this is set to true, `BuildResult.logText` will not contain any
+        /// progress information.
+        public var progressToConsole: Bool
+
         /// Additional flags to pass to all C compiler invocations.
         public var otherCFlags: [String] = []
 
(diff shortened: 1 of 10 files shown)


Sources/PackagePlugin/PackageManagerProxy.swift
```diff
@@ -62,6 +62,12 @@ public struct PackageManager {
         /// Whether to print build logs to the console
         public var echoLogs: Bool
 
+        /// Whether to print build progress to the console
+        ///
+        /// If this is set to true, `BuildResult.logText` will not contain any
+        /// progress information.
+        public var progressToConsole: Bool
```
jakepetroules, line 69: This is new public API and will therefore need a Swift Evolution proposal API availability annotation.

Can you please start a pitch on the Swift Forums to propose this new API?

kateinoigakukun: Sure, I started it here: https://forums.swift.org/t/pitch-build-progress-display-control-for-swiftpm-plugins/78678

PR #8384, opened 2025-03-19
Exclude unused files from SwiftBuildSupport target

bkhouri:
Remove warnings generated by unused files in the SwiftBuildSupport target.


bkhouri:
This is fixed in https://github.com/swiftlang/swift-package-manager/pull/8361.  Closing the PR.

PR #8396, opened 2025-03-20
Move integration tests to use swift-testing

kcieplak:
- Move all integration tests to use swift-testing
- Add some test skipping traits for OS types, Xcode etc.
- Create a common test module
  - Move the helpers from test source to common test module.
  - TSCBasic requires a minimum os of 10.3, force produce to 13.0
  - Update the IntegrationTests sub package to use the common test module.
  - Remove XCTest macros from within the StringChecker for #expect usage.
- Run swift-format on all files


kcieplak:
@swift-ci test

IntegrationTests/Sources/IntegrationTestSupport/Filesystem.swift
```diff
@@ -0,0 +1,18 @@
...
+*/
+
+import TSCBasic
+
+extension FileSystem {
+    // Returns true if a path does not exist. Makes #expect statements more obvious.
+    public func notExists(_ path: AbsolutePath, followSymlink: Bool = false) -> Bool {
+        return !exists(path, followSymlink: followSymlink)
+    }
+}
```
MaxDesiatov, line 18: missing trailing newline
```suggestion
}

```

IntegrationTests/Sources/IntegrationTestSupport/Helpers.swift
```diff
@@ -177,7 +184,8 @@ func _sh(
 
     environment.merge(env, uniquingKeysWith: { $1 })
 
-    let result = try Process.popen(arguments: arguments.map { $0.description }, environment: environment)
+    let result = try Process.popen(
+        arguments: arguments.map { $0.description }, environment: environment)
```
MaxDesiatov, line 188: Does this use SwiftFormat or swift-format? Per [`CONTRIBUTING.md`](https://github.com/swiftlang/swift-package-manager/blob/main/CONTRIBUTING.md#creating-pull-requests`) we're using SwiftFormat here, which IIUC would format it differently.
```suggestion
        arguments: arguments.map { $0.description }, environment: environment
    )
```

IntegrationTests/Sources/IntegrationTestSupport/SkippedTestSupport.swift
```diff
@@ -0,0 +1,44 @@
...
+    public static func skipIfXcodeBuilt() -> Self {
+        disabled {
+            #if Xcode
+                true
+            #else
+                false
+            #endif
+        }
+    }
+}
```
MaxDesiatov, line 44: Missing final newline
```suggestion
}

```

IntegrationTests/Sources/IntegrationTestSupport/SkippedTestSupport.swift
```diff
@@ -0,0 +1,44 @@
...
+    public static func skip(_ comment: Comment? = nil) -> Self {
+        disabled(comment ?? "Unconditional skip, a comment should be added for the reason", { true } )
+    }
+    // Skip test will be skipped if the environment is self hosted.
+    public static func skipSwiftCISelfHosted(_ comment: Comment? = nil) -> Self {
+        disabled(comment ?? "SwiftCI is self hosted") {
+            ProcessInfo.processInfo.environment["SWIFTCI_IS_SELF_HOSTED"] != nil
+        }
+    }
+    // Skip test if the built by XCode.
```
MaxDesiatov, line 34: nit: typo fix
```suggestion
    // Skip test if built by Xcode.
```

MaxDesiatov, line 34: nit: formatting-wise we separate declarations with a single newline.
```suggestion
    // Skip test if the host operating system does not match the running OS.
    public static func requireHostOS(_ os: OperatingSystem, when condition: Bool = true) -> Self {
        enabled("This test requires a \(os) host OS.", { ProcessInfo.processInfo.hostOperatingSystem() == os && condition })
    }

    // Skip test if the host operating system matches the running OS.
    public static func skipHostOS(_ os: OperatingSystem, _ comment: Comment? = nil) -> Self {
        disabled(comment ?? "This test cannot run on a \(os) host OS.", { ProcessInfo.processInfo.hostOperatingSystem() == os })
    }

    // Skip test unconditionally
    public static func skip(_ comment: Comment? = nil) -> Self {
        disabled(comment ?? "Unconditional skip, a comment should be added for the reason", { true } )
    }

    // Skip test will be skipped if the environment is self hosted.
    public static func skipSwiftCISelfHosted(_ comment: Comment? = nil) -> Self {
        disabled(comment ?? "SwiftCI is self hosted") {
            ProcessInfo.processInfo.environment["SWIFTCI_IS_SELF_HOSTED"] != nil
        }
    }

    // Skip test if the built by XCode.
```

IntegrationTests/Sources/IntegrationTestSupport/Helpers.swift
kcieplak: Looks like the vscode swift extension picked swift-format.  I will re-run with SwiftFormat.

kcieplak:
@swift-ci test

IntegrationTests/Sources/IntegrationTestSupport/Helpers.swift
```diff
@@ -138,14 +136,17 @@ func sh(
     let stderr = try result.utf8stderrOutput()
 
     if result.exitStatus != .terminated(code: 0) {
-        XCTFail("Command failed with exit code: \(result.exitStatus) - \(result.integrationTests_debugDescription)", file: file, line: line)
+        XCTFail(
```
bkhouri, line 139: issue (blocking): Since we convert the Integration Tests to Swift Testing, `XCTFail` will not work, or may not properly report the failure message.

Can we use `Issue.record(...)` instead?  See the [Migration from XCTest](https://developer.apple.com/documentation/testing/migratingfromxctest) documentation..

IntegrationTests/Sources/IntegrationTestSupport/Helpers.swift
```diff
@@ -156,28 +157,33 @@ func shFails(
     let stderr = try result.utf8stderrOutput()
 
     if result.exitStatus == .terminated(code: 0) {
-        XCTFail("Command unexpectedly succeeded with exit code: \(result.exitStatus) - \(result.integrationTests_debugDescription)", file: file, line: line)
+        XCTFail(
```
bkhouri, line 160: issue (blocking): Since we convert the Integration Tests to Swift Testing, `XCTFail` will not work, or may not properly report the failure message.

Can we use `Issue.record(...)` instead?  See the [Migration from XCTest](https://developer.apple.com/documentation/testing/migratingfromxctest) documentation..

IntegrationTests/Sources/IntegrationTestSupport/SkippedTestSupport.swift
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    // Skip test if built by XCode.
+    public static func skipIfXcodeBuilt() -> Self {
+        disabled {
+            #if Xcode
+            true
+            #else
+            false
+            #endif
```
bkhouri, line 49: suggestion: can we add a message as it's always nice to know why a test was skipped.

IntegrationTests/Tests/IntegrationTests/BasicTests.swift
```diff
@@ -11,69 +11,82 @@
...
+@Suite
+fileprivate struct BasicTests {
+    @Test
     func testVersion() throws {
-        XCTAssertMatch(try sh(swift, "--version").stdout, .contains("Swift version"))
+        #expect(try sh(swift, "--version").stdout.contains("Swift version"))
     }
 
+    @Test(
+        nil,
```
bkhouri, line 25: thought (non-blocking):  I believe the first argument can be omitted, at least I have omitted them in a previous PR - which are still open.

IntegrationTests/Sources/IntegrationTestSupport/SkippedTestSupport.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    // Skip test if the host operating system matches the running OS.
+    public static func skipHostOS(_ os: OperatingSystem, _ comment: Comment? = nil) -> Self {
+        disabled(comment ?? "This test cannot run on a \(os) host OS.") {
+            ProcessInfo.hostOperatingSystem == os
+        }
+    }
+
+    // Skip test unconditionally
+    public static func skip(_ comment: Comment? = nil) -> Self {
```
bkhouri, line 31: thought: can we use `.disabled(...)` instead of creating an extra indirection?

IntegrationTests/Sources/IntegrationTestSupport/Helpers.swift
```diff
@@ -1,49 +1,51 @@
...
-*/
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+ */
 
 import Foundation
-import XCTest
 import TSCBasic
 import TSCTestSupport
+import XCTest
```
bkhouri, line 14: issue (blocking): Since the `IntegrationTests` have been converted to Swift Testing, can we remove this import once we convert the `XCTFails` to the appropriate Swift Testing equivalent, which is posted in another comment.

IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift
```diff
@@ -1,106 +1,131 @@
...
 
-    func testSwiftBuild() throws {
-        #if os(Linux)
-        if FileManager.default.contents(atPath: "/etc/system-release").map { String(decoding: $0, as: UTF8.self) == "Amazon Linux release 2 (Karoo)\n" } ?? false {
-            throw XCTSkip("Skipping SwiftBuild testing on Amazon Linux because of platform issues.")
-        }
-        #endif
-
-        // Test SwiftBuildSystem
+    @Test(nil, .skipHostOS(.linux, "Amazon Linux has platform issues"))
```
bkhouri, line 69: issue (blocking): this will skip the test on all linux host, but the issue is only on Amazon Linux 2.  Can we update the `skipHostOS` trait that optionally accepts an "enum" which we can use skip against a specific distribution?

if the optional is not set, skip on all host.  if the optional is set, skip only on that distribution.

IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift
```diff
@@ -1,106 +1,131 @@
...
+                // #expect(!stderr.contains("error:"))
+                // #expect(stdout.contains("Hello, world!"))
             }
         }
+    }
 
-        #if !os(Windows)
-        // SWBINTTODO: Windows fails to link this library package due to a "lld-link: error: subsystem must be defined" error. See https://github.com/swiftlang/swift-build/issues/310
+    @Test(
+        nil, .skipHostOS(.linux, "Amazon Linux has platform issues"),
```
bkhouri, line 89: issue (blocking): this will skip the test on all linux host, but the issue is only on Amazon Linux 2.  Can we update the `skipHostOS` trait that optionally accepts an "enum" which we can use skip against a specific distribution?

if the optional is not set, skip on all host.  if the optional is set, skip only on that distribution.

IntegrationTests/Tests/IntegrationTests/XCBuildTests.swift
```diff
@@ -1,40 +1,41 @@
...
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See http://swift.org/LICENSE.txt for license information
  See http://swift.org/CONTRIBUTORS.txt for Swift project authors
  */
 
-import XCTest
 import TSCBasic
 import TSCTestSupport
+import XCTest
```
bkhouri, line 13: issue (blocking): Since we converted to swift testing, can we remove this import and replace all XCTest API with the associated Swift Testing call?

See https://developer.apple.com/documentation/testing/migratingfromxctest#Record-issues

bkhouri (changes requested):
Looks good, though I still see hints of XCTest and some test  are no longer executed on linux host, while the XCTest version skipped the test only on Amazon Linux 2.

All comments that require code changes have been marked  with `(blocking)`,

IntegrationTests/Sources/IntegrationTestSupport/Helpers.swift
jakepetroules: Yes, `Issue.record` is the canonical replacement for `XCTFail`.

Never mix XCTest and Swift Testing APIs in the same test.

IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift
jakepetroules: Just copy the code from https://github.com/swiftlang/swift-build/blob/cd632ba33e264dcac4bcf3d09f83e404c3408f09/Sources/SWBTestSupport/SkippedTestSupport.swift#L140; the sole issue with Amazon Linux 2 is the lack of posix_spawn_file_actions_addchdir, so let's be explicit about that. Technically the same issue would impact OpenBSD.

IntegrationTests/Tests/IntegrationTests/BasicTests.swift
kcieplak: For some reason when I first converted the tests, compilation was failing as the @Test was expecting String? as the first argument.  I removed the nil, and is seems to work now.  Not sure what changed.

IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift
kcieplak: Sorry, yeah not sure why I missed that the code was checking not just for #if os(linux), but probing the distro as well.  

I will borrow more from the swift-build test support ;)

kcieplak:
@swift-ci test

kcieplak:
@swift-ci test windows

IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift
```diff
@@ -1,106 +1,133 @@
...
+                // #expect(!stderr.contains("error:"))
+                // #expect(stdout.contains("Hello, world!"))
             }
         }
+    }
 
-        #if !os(Windows)
-        // SWBINTTODO: Windows fails to link this library package due to a "lld-link: error: subsystem must be defined" error. See https://github.com/swiftlang/swift-build/issues/310
+    @Test(
+        .skipHostOS(
```
bkhouri, line 91: thought (non-blocking): I really wish Swift Testing has a `.all(...)` and `.any(...)` trait to make this more explicit as we don't know if this is currently an `OR` or `AND` condition.


Here, we need to skip the test if any of the conditions are true.

I created https://github.com/swiftlang/swift-testing/issues/1034 to have an option to be more explicit.

kcieplak:
@swift-ci test macos

bkhouri (approved):

IntegrationTests/Tests/IntegrationTests/SwiftPMTests.swift
kcieplak: From https://developer.apple.com/documentation/testing/enablinganddisabling

```
@Test(
  "Ice cream is cold",
  .enabled(if: Season.current == .summer),
  .disabled("We ran out of sprinkles"),
  .bug(id: "12345")
)
func isCold() async throws { ... }
```
If a test has multiple conditions applied to it, they must all pass for it to run. Otherwise, the test notes the first condition to fail as the reason the test is skipped.

The traits are .enable or .disable depending on the extension of the trait. In this case it will have:

skipHostOS
enabled() with a condition of true if os != windows

requireThreadSafeWorkingDirectory
disabled() with a condition of true  if the os does not support thread safe working directories

I think this logic is correct for this test-case.

kcieplak:
Jenkins instance was rebooted during macOS legs causing failures

kcieplak:
@swift-ci test macos

Issue #8393, opened 2025-03-20
`swift package init --package-path <dir-does-not-exists>` does not create the directory


bkhouri:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

When trying to create a new package, using the `--package-path` option fails if the directory does not exist.

### Expected behavior

`swift package init --package-path <dir>` should:
- create the directory, and populate the package if it does not exists
- should populate the package if the directory exists and is empty
- prompt the user for action if the directory exists and is non-empty 
- should error if `<dir>` exists as a file

### Actual behavior

# directory does not exists
```
❯ swift package init --package-path Foo
error: chdir error: No such file or directory (2): /Users/bkhouri/Downloads/Foo
```

# Directory exists and is empty
```
❯ swift package init --package-path Foo
Creating library package: Foo
Creating Package.swift
Creating .gitignore
Creating Sources
Creating Sources/Foo/Foo.swift
Creating Tests/
Creating Tests/FooTests/
Creating Tests/FooTests/FooTests.swift
```

# Directory Exists and contains a `Package.swift`
```
❯ swift package init --package-path Foo
Creating library package: Foo
error: a manifest file already exists in this directory
```

# Directory exists and is not empty (does not contains `Package.swift`)
```
❯ swift package init --package-path Foo
<...SNIP...>
❯ rm -rf Foo/Package.swift
❯ swift package init --package-path Foo
Creating library package: Foo
Creating Package.swift
```

### Steps to reproduce

```
swift package init --package-path ./does-not-exists-yadda-yadda
```

### Swift Package Manager version/commit hash

Swift Package Manager - Swift 6.2.0-dev

### Swift & OS version (output of `swift --version ; uname -a`)

❯ swift --version
Apple Swift version 6.2-dev (LLVM 162ee50b401fff2, Swift 57288d13c9f3c02)
Target: arm64-apple-macosx15.0
Build config: +assertions

❯ uname -a
Darwin BKs-Work-M1Pro-MacBook-Pro-2.local 24.4.0 Darwin Kernel Version 24.4.0: Tue Mar 11 22:26:20 PDT 2025; root:xnu_development-11417.101.12~4/DEVELOPMENT_ARM64_T6000 arm64

PR #8401, opened 2025-03-21
Allow creation of non-existent directory during `swift package init`

plemarquand:
Running `swift package init --package-path <dir-does-not-exists>` results in an error when SwiftPM attempts to change directories in to the directory that doesn't exist.

### Motivation:

It is rare that a user wants to do a `swift package init` in a folder that already has content. A typical pattern is to `mkdir mypackage && cd my package && swift package init`. SwiftPM already has a `--package-path` flag indicating the package folder that the command should operate on, but attempting to use this to create a folder that doesn't exist during `swift package init` results in an error when SwiftPM attempts to chdir to the --package-path that doesn't exist.

### Modifications:

Add a new boolean to the `_SwiftCommand` protocol that lets commands opt in to creating the directory at `--package-path` if it doesn't exist. Opt in `InitCommand` to this behaviour.

### Result:

`swift package init --package-path ./mypackage` successfully initializes a package in `./mypackage`, even if `./mypackage` doesn't exist.

Issue: #8393

### DIFF
--- a/Tests/CommandsTests/SwiftCommandStateTests.swift
+++ b/Tests/CommandsTests/SwiftCommandStateTests.swift
@@ -489,7 +489,7 @@ final class SwiftCommandStateTests: CommandsTestCase {
     }
 
     func testPackagePathWithMissingFolder() async throws {
-        try fixture(name: "Miscellaneous/Simple") { fixturePath in
+        try withTemporaryDirectory { fixturePath in
             let packagePath = fixturePath.appending(component: "Foo")
             let options = try GlobalOptions.parse(["--package-path", packagePath.pathString])
 
--- a/Tests/CommandsTests/SwiftCommandStateTests.swift
+++ b/Tests/CommandsTests/SwiftCommandStateTests.swift
@@ -22,8 +22,10 @@ import func PackageGraph.loadModulesGraph
 import _InternalTestSupport
 import XCTest
 
+import ArgumentParser
 import class TSCBasic.BufferedOutputByteStream
 import protocol TSCBasic.OutputByteStream
+import enum TSCBasic.SystemError
 import var TSCBasic.stderrStream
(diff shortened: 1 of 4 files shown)


Tests/CommandsTests/SwiftCommandStateTests.swift
```diff
@@ -485,12 +487,34 @@ final class SwiftCommandStateTests: CommandsTestCase {
         XCTAssertEqual(try targetToolchain.getClangCompiler(), targetClangPath)
         XCTAssertEqual(targetToolchain.librarianPath, targetArPath)
     }
+
+    func testPackagePathWithMissingFolder() async throws {
+        try fixture(name: "Miscellaneous/Simple") { fixturePath in
```
bkhouri, line 492: suggestion: Since we the test does not care about the existence of the directory, we can use `withTemporaryDirectory` instead of `fixture("...")`.

bkhouri (approved):
Looks good to me, though I would not close #8393 as one of the expected behaviours is not currently implemented.


`swift package init --package-path <dir>` should:


- [x] create the directory, and populate the package if it does not exists 
- [x] should populate the package if the directory exists and is empty
- [ ] prompt the user for action if the directory exists and is non-empty
- [x] should error if `<dir>` exists as a file

plemarquand:
@swift-ci test

jakepetroules (approved):

cmcgee1024 (approved):
This is a really welcome addition to the init package workflow.

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test linux
@swift-ci test windows

plemarquand:
@swift-ci test linux

dschaefer2 (approved):

PR #8415, opened 2025-03-26
NFC: Fix trailing commas

DmT021:


xedin (approved):
LGTM. I'm a bit surprised that formatter doesn't remove them.

MaxDesiatov:
Even if the formatter did remove them, CI doesn't run the formatter on PR jobs to verify that trailing commas aren't merged. And we have no Swift 6.0 PR jobs currently for builds to fail on this and provide some indication either.

PR #8417, opened 2025-03-26
[Traits] Support traits in `swift package edit` and `swift package unedit` commands

### LINKED ISSUE #8354: `swift package edit` and `swift package unedit` should support trait options

### Description

The `swift package edit` and `swift package unedit` commands don't currently support passing in a trait configuration despite having a path for computing dependency resolution - this will disrupt the dependency graph for packages relying on enabled/disabled trait configurations.

### Expected behavior

The `edit` and `unedit` commands should support traits (as `swift run`, `swift build`, `swift package resolve` etc. do) in their available options.

### Actual behavior

Traits are not supported as a current option.

### Steps to reproduce

No concrete steps of reproduction besides taking note of the fact that there is no entry point into `swift package edit` and `swift package unedit` for configuring traits - can make note of this in both the source code where the command structs themselves do not have a `TraitOptions` property, and in CLI when running `swift package edit --help`/`swift package unedit --help`.

_No response_

_No response_

bripeticca:
Fixes #8354

### Motivation:

Since the `edit` and `unedit` commands can affect dependency resolution, we must support the ability to specify trait configurations.

### Modifications:

- Add entrypoint for trait options in the edit/unedit commands
- Made appropriate trait-related structs and properties public for accessibility in test suite

### Result:

`swift package edit` and `swift package unedit` should now be able to support traits. Current test mocks don't fully support edit/unedit by pointing to an entirely different manifest and parsing its contents at the moment, so further improvements to the mocks is needed for more robust testing of this feature which will be covered by a future change.


bripeticca:
@swift-ci please test

bripeticca:
After chatting with @dschaefer2 and investigating the `edit`/`unedited`/`build` commands further, we decided that it might not make sense to have `resolve` called from the `edit`/`unedit` commands as subsequently users will generally do a `swift build`/`run`/etc. which will either trigger a resolution if required, or if there is a cached build operation it does a check to assure that the trait configurations match and if not it will generate a new build plan (which resolves as well).

bripeticca:
Upon further discussion, it seems like this might need more work to refactor - a lot of the logic for an edit relies on the Workspace state having managed dependencies which is only done after a dependency resolution, as they will have information about the checkout state (for remote source control deps). When executing an edit on a remote source control dependency, the edit command also attempts to checkout to the same revision but without a call to resolve the dependencies, the reference to the revision does not exist at this stage.

Ultimately, it seems like edit is making changes in the wrong places - users shouldn’t have to be concerned about the result of resolution when making an edit, as the command is the equivalent to manually modifying the manifest file.

Will close this PR as we've re-evaluated the needs here.

PR #8434, opened 2025-04-01
Sever the package dependency if using SwiftBuild Framework

bkhouri:
The SwiftBuild project has a dependency on Swift Driver, which has a dependency on Swift Tools Support Core (STSC).  Swift Package Manager (SwiftPM) executable targets also have a dependency on STSC.

Sever SwiftPM package depenedency on STSC if the
`SWIFTPM_SWBUILD_FRAMEWORK` environment variable is set so we can indirectly pull the STCS dependency on the dynamic library that will be pulled in via SwiftDriver.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -87,6 +87,26 @@ if ProcessInfo.processInfo.environment["SWIFTCI_INSTALL_RPATH_OS"] == "android" 
  */
 let autoProducts = [swiftPMProduct, swiftPMDataModelProduct]
 
+private func useSwiftBuildPackageDependency() -> Bool {
+    return (ProcessInfo.processInfo.environment["SWIFTPM_SWBUILD_FRAMEWORK"] == nil &&
+    ProcessInfo.processInfo.environment["SWIFTPM_NO_SWBUILD_DEPENDENCY"] == nil)
+}
+
+let swiftDriverDep: [Target.Dependency]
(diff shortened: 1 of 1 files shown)


bkhouri:
@swift-ci please test

bkhouri:

jakepetroules (approved):

Package.swift
```diff
@@ -87,6 +87,26 @@ if ProcessInfo.processInfo.environment["SWIFTCI_INSTALL_RPATH_OS"] == "android"
  */
 let autoProducts = [swiftPMProduct, swiftPMDataModelProduct]
 
+private func useSwiftBuildPackageDependency() -> Bool {
+    return (ProcessInfo.processInfo.environment["SWIFTPM_SWBUILD_FRAMEWORK"] == nil &&
+    ProcessInfo.processInfo.environment["SWIFTPM_NO_SWBUILD_DEPENDENCY"] == nil)
+}
```
MaxDesiatov, line 93: Is there a particular reason this uses a function instead of a simple `let` binding? I also don't think `private` is needed here, there's no chance this function could be used accidentally elsewhere, since multi-file package manifest is not a thing.

bkhouri: This is used in two places, so instead of duplicating the logic, I opted to put it in a function call.

PR #8438, opened 2025-04-01
Android: remove build flags that are no longer needed and force unwrap dlerror()

finagolfin:
Removing the definition of `_GNU_SOURCE` is no longer needed since swiftlang/swift-tools-support-core#500 made sure the right `strerror_r()` is used, and `dlerror()` has to be unwrapped because of its `_Nullable` annotation.


finagolfin:
@swift-ci please test

finagolfin:

PR #8443, opened 2025-04-02
Add executorFactory option to package manifest.

al45tair:
This lets you specify which executor factory to use in the `Package.swift`.

rdar://148430450

### DIFF
--- a/Sources/PackageDescription/BuildSettings.swift
+++ b/Sources/PackageDescription/BuildSettings.swift
@@ -501,7 +501,7 @@ public struct SwiftSetting: Sendable {
       _ condition: BuildSettingCondition? = nil
     ) -> SwiftSetting {
         return SwiftSetting(
-          name: "swiftExecutorFactory", value: [name], condition: condition)
+          name: "executorFactory", value: [factory], condition: condition)
     }
 }
 
--- a/Sources/PackageDescription/BuildSettings.swift
+++ b/Sources/PackageDescription/BuildSettings.swift
@@ -486,6 +486,23 @@ public struct SwiftSetting: Sendable {
         return SwiftSetting(
             name: "defaultIsolation", value: [isolationString], condition: condition)
     }
+
+    /// Defines an `-executor-factory` to pass to the
+    /// corresponding build tool.
+    ///
+    /// - Since: First available in PackageDescription 6.2
+    ///
+    /// - Parameters:
+    ///   - factory: The type name of the executor factory that should be used.
+    ///   - condition: A condition that restricts the application of the build setting.
+    @available(_PackageDescription, introduced: 6.2)
+    public static func executorFactory(
+      _ factory: String,
+      _ condition: BuildSettingCondition? = nil
+    ) -> SwiftSetting {
+        return SwiftSetting(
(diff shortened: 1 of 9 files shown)


al45tair:
@swift-ci Please smoke test

MaxDesiatov (commented):
What's the corresponding Swift Evolution proposal/pitch for this? IIUC changes to `PackageDescription` public API are usually required to be pitched before merging.

dschaefer2:
Yup, a quick pitch should be fine just to let the community know. Also some tests :).

al45tair:
> What's the corresponding Swift Evolution proposal/pitch for this?

Pitch is here:
https://forums.swift.org/t/pitch-2-custom-main-and-global-executors/78437

> Also some tests :).

Indeed, I'd like to add some tests, though I find the SwiftPM repo hard to navigate so it would be useful if someone could point me in the right place(s) for that.

MaxDesiatov:
You could modify the existing [`testBuildSettings`](https://github.com/swiftlang/swift-package-manager/blob/c4f0e4b7f90d09a985fcb3d3433382b757fb9ed5/Tests/BuildTests/BuildPlanTests.swift#L4240) in `BuildPlantTests.swift`, although it probably be better to create a separate test function so that it can run in parallel when these are migrated to Swift Testing.

jakepetroules:
@al45tair Please add a new build setting in Swift Build for this as well. When using the Swift Build build engine backend we're not going to be using OTHER_SWIFT_FLAGS to pass down these sort of things.

bkhouri (changes requested):
chore (blocking): Would it be possible to add an automated tests that verify, at a minimum:
- A valid value is passed for the executor factory
- an invalid value is passed to the executor Factory
- the executor factory is not specified

This will ensure we the behaviour works as intended and that it does not regress

al45tair:
> Would it be possible to add an automated tests that verify, at a minimum:
> 
> * A valid value is passed for the executor factory
> * an invalid value is passed to the executor Factory
> * the executor factory is not specified
> 
> This will ensure we the behaviour works as intended and that it does not regress

It doesn't make sense to test all of that here. Tests of compiler options belong in the compiler tests. SwiftPM might reasonably test that the option is passed through to the compiler — it shouldn't be testing compiler behaviour.

al45tair:
(I'll add to that that I think it would make a lot of sense if SwiftPM and Swift Build picked up options automatically from the `.def` files in the main Swift repo; there could be a flag that causes the option to be added to the `swiftSettings` automatically. For a _lot_ of compiler options, including this one, that would be more than sufficient and would avoid people working on compiler features having to open two separate PRs and write separate sets of tests for each of these things. `swift-driver` already does pretty much this, though it isn't wholly automated and you need to open a PR against it to get an option added.)

al45tair:
@swift-ci Please smoke test

al45tair:
@swift-ci Please smoke test

al45tair:
@swift-ci Please test

al45tair:
Closing in favour of https://github.com/swiftlang/swift/pull/80758

dschaefer2:
Ah, excellent. Thanks Alastair! Yeah, I think that's better ergonomics than a build setting. Can't wait to try it out :).

PR #8452, opened 2025-04-07
Implemented support in the new PIF builder for main module products with multiple modules.

cwakamo:
This was previously ignored as unsupported, but packages with multiple modules are fully-supported by the "native" SwiftPM build system.

This addresses <rdar://problem/147793804>.


cwakamo:
I still need to add a test for this, and actually test that this works (because the new PIF builder isn't hooked up to `--build-system swiftbuild` yet), but I wanted to get this up for review while I had the changes fresh.

jakepetroules (approved):

PR #8458, opened 2025-04-08
Add missing products from the swift-syntax prebuilts.

dschaefer2:
We were focused on including things for macros but missed a number of products, including new ones for 601. With this change we now just include all the products and make sure all the targets they depend on end up in the library.


dschaefer2:
@swift-ci please test

PR #8459, opened 2025-04-08
Link test command test failures on Linux to issue

cmcgee1024:
The TestCommandTests fail on Linux due to a common linker problem with swift build build system,
which is the missing 'main' symbol problem. This is described further detail in #8439.

Provide the link in the skip messages and remove the TODO tags since the investigation is complete
and the reason for the failures is now known.

Update the common test case logic so that it dumps the command execution stdout and stderr
in the case of failure to help diagnose these result.xml failures immediately in the logs.

### DIFF
--- a/Tests/CommandsTests/TestCommandTests.swift
+++ b/Tests/CommandsTests/TestCommandTests.swift
@@ -704,23 +704,23 @@ class TestCommandSwiftBuildTests: TestCommandTestCase {
     }
 
     override func testSwiftTestXMLOutputVerifyMultipleTestFailureMessageWithFlagEnabledXCTest() async throws {
-        throw XCTSkip("Result XML could not be found. The build fails due to a linker error on missing main symbol. https://github.com/swiftlang/swift-package-manager/issues/8439"")
+        throw XCTSkip("Result XML could not be found. The build fails due to a linker error on missing main symbol. https://github.com/swiftlang/swift-package-manager/issues/8439")
     }
 
     override func testSwiftTestXMLOutputVerifySingleTestFailureMessageWithFlagEnabledXCTest() async throws {
-        throw XCTSkip("Result XML could not be found. The build fails due to a linker error on missing main symbol. https://github.com/swiftlang/swift-package-manager/issues/8439"")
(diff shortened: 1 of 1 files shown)


cmcgee1024:
@swift-ci please test

Tests/CommandsTests/TestCommandTests.swift
```diff
@@ -239,6 +239,12 @@ class TestCommandTestCase: CommandsBuildProviderTestCase {
                 throwIfCommandFails: false
             )
 
+            if !FileManager.default.fileExists(atPath: xUnitUnderTest.pathString) {
```
bkhouri, line 242: suggestion: can we include the output of `stdout` and `stderr` as part of the assertion message?

cmcgee1024: Won't it look long and unwieldy to have an entire command output in an assert message? The stdout would be the entire verbose build log from running the `swift test` command. It's probably 50 lines long, or more in some cases with dense command-line arguments.

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Linux

bkhouri:

Tests/CommandsTests/TestCommandTests.swift
bkhouri: If the assertion changes, we need to remember to update this part of the test.  Instead, if a test assertion fails, we want to output the contents.  We could always add `\n` characters in the assertion message to make the output a bit better.

cmcgee1024: The API from TSC for this doesn't appear to offer a way to customize the assert message.

cmcgee1024:
@swift-ci test smoke test Linux

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Windows

Tests/CommandsTests/TestCommandTests.swift
bkhouri: That's a shame!

bkhouri (approved):

bkhouri:

PR #8464, opened 2025-04-08
Initialize Swift Build with a toolchain-based developer directory

owenv:
This allows us to correctly construct the stack of toolchains used for tool/library lookup internally, fixing builds of test targets (among other things).

Depends on https://github.com/swiftlang/swift-build/pull/385

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -62,13 +62,22 @@ func withService(
 func withSession(
     service: SWBBuildService,
     name: String,
+    toolchainPath: String,
     packageManagerResourcesDirectory: Basics.AbsolutePath?,
     body: @escaping (
         _ session: SWBBuildServiceSession,
         _ diagnostics: [SwiftBuild.SwiftBuildMessage.DiagnosticInfo]
     ) async throws -> Void
 ) async throws {
-    switch await service.createSession(name: name, resourceSearchPaths: packageManagerResourcesDirectory.map { [$0.pathString] } ?? [], cachePath: nil, inferiorProductsPath: nil, environment: nil) {
+    switch await service.createSession(
+        name: name,
+        swiftToolchainPath: toolchainPath,
+        resourceSearchPaths: packageManagerResourcesDirectory.map {
+            [$0.pathString]
+        } ?? [],
+        cachePath: nil, inferiorProductsPath: nil,
+        environment: nil
+    ) {
     case (.success(let session), let diagnostics):
         do {
             try await body(session, diagnostics)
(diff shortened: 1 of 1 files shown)


Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -62,13 +62,14 @@ func withService(
...
     name: String,
+    toolchainPath: String,
     packageManagerResourcesDirectory: Basics.AbsolutePath?,
     body: @escaping (
         _ session: SWBBuildServiceSession,
         _ diagnostics: [SwiftBuild.SwiftBuildMessage.DiagnosticInfo]
     ) async throws -> Void
 ) async throws {
-    switch await service.createSession(name: name, resourceSearchPaths: packageManagerResourcesDirectory.map { [$0.pathString] } ?? [], cachePath: nil, inferiorProductsPath: nil, environment: nil) {
+    switch await service.createSession(name: name, swiftToolchainPath: toolchainPath, resourceSearchPaths: packageManagerResourcesDirectory.map { [$0.pathString] } ?? [], cachePath: nil, inferiorProductsPath: nil, environment: nil) {
```
MaxDesiatov, line 72: line length nit

```suggestion
    switch await service.createSession(
        name: name, 
        swiftToolchainPath: toolchainPath, 
        resourceSearchPaths: packageManagerResourcesDirectory.map { 
            [$0.pathString] 
        } ?? [], 
        cachePath: nil,
        inferiorProductsPath: nil,
        environment: nil
    ) {
```

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -265,7 +266,11 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
             )
 
             do {
-                try await withSession(service: service, name: self.buildParameters.pifManifest.pathString, packageManagerResourcesDirectory: self.packageManagerResourcesDirectory) { session, _ in
+                let toolchainPath = self.buildParameters.toolchain.swiftCompilerPath
+                    .parentDirectory // remove swift
+                    .parentDirectory // remove bin
+                    .parentDirectory // remove usr
+                try await withSession(service: service, name: self.buildParameters.pifManifest.pathString, toolchainPath: toolchainPath.pathString, packageManagerResourcesDirectory: self.packageManagerResourcesDirectory) { session, _ in
```
MaxDesiatov, line 273:
```suggestion
                try await withSession(
                    service: service, 
                    name: self.buildParameters.pifManifest.pathString, 
                    toolchainPath: toolchainPath.pathString,
                    packageManagerResourcesDirectory: self.packageManagerResourcesDirectory
                ) { session, _ in
```

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -265,7 +266,11 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
             )
 
             do {
-                try await withSession(service: service, name: self.buildParameters.pifManifest.pathString, packageManagerResourcesDirectory: self.packageManagerResourcesDirectory) { session, _ in
+                let toolchainPath = self.buildParameters.toolchain.swiftCompilerPath
```
jakepetroules, line 269: I think the toolchain base path should also be part of the Toolchain protocol itself rather than expanded here in SwiftBuildSystem.

Also because on some platforms it's usr/local/bin instead of usr/local, so I'd rather that logic be kept In UserToolchain itself.

jakepetroules (changes requested):

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
owenv: There's not an obvious way to add that requirement right now because the SwiftPM-level toolchain type conflates toolchains, SDKs, and cross compilation bundles

owenv:
@swift-ci test

owenv:
Obsoleted by https://github.com/swiftlang/swift-package-manager/pull/8696

Issue #8436, opened 2025-04-01
`swift build` failed when macro and target depend on the same C target


qazyn951230:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

When `marco` and `target` depend on the same C target, running `swift build` will produce a `redefinition of module` error.

A minimal example can be found [here](https://github.com/qazyn951230/MacroTest). The project structure is as follows:

- CLib
- MyLib
    - CLib
    - MyMacro
- MyMacro
    - CLib

MyLib and MyMacro only import CLib and have no other code.

### Expected behavior

Compiled successfully.

### Actual behavior

```shell
$ swift build
[1/1] Planning build
Building for debugging...
error: emit-module command failed with exit code 1 (use -v to see invocation)
<REDACTED>/CLib-tool.build/module.modulemap:1:8: error: redefinition of module 'CLib'
module CLib {
       ^
<REDACTED>/CLib.build/module.modulemap:1:8: note: previously defined here
module CLib {
       ^
<REDACTED>/CLib-tool.build/module.modulemap:1:8: error: redefinition of module 'CLib'
module CLib {
       ^
<REDACTED>/CLib.build/module.modulemap:1:8: note: previously defined here
module CLib {
       ^
```

### Steps to reproduce

1. `git clone https://github.com/qazyn951230/MacroTest.git`
2. `cd MacroTest`
3. `swift build` 

### Swift Package Manager version/commit hash

6.1.0

### Swift & OS version (output of `swift --version ; uname -a`)

Swift 6.1 from Xcode 16.3

```shell
$ swift --version
swift-driver version: 1.120.5 Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)
Target: arm64-apple-macosx15.0
```

dschaefer2:
I see. We are picking up the dependencies from the macros and plugins and hooking them up. That's why we get them twice. I'll fix up the algorithm.

qazyn951230:
@dschaefer2 I'm glad you fixed it. Should it be cherry-picked into the branch "release/6.2" since it's a bug fix?

dschaefer2:
Fixed by #8472

dschaefer2:
[...]

PR #8472, opened 2025-04-09
Fix duplicate modulemap errors with macro and plugin deps

dschaefer2:
We were including flags to hook up modulemaps and include files to C library dependencies in macros and plugin tools through to the modules that depend on those. This adds the capability to prune the depth first searches through the dependencies to ensure these are skipped when crossing macro and plugin boundaries.


dschaefer2:
@swift-ci please test

dschaefer2:

dschaefer2:
For #8436

jakepetroules (approved):

PR #8485, opened 2025-04-11
Ensure prebuilt Modules include path added to generated modules

dschaefer2:
For Windows and Linux, SwiftPM generates modules for test discovery and the test entry point. When testing macro targets, we add the swift-syntax prebuilt Modules directory to the include path for the test target. We also need to do this for the generated modules.

As a side effect, this also adds all include directories to the include path for these generated modules to ensure they build for other use cases that may need this.


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test smoke test linux

dschaefer2:
@swift-ci please test linux

cmcgee1024 (approved):

jakepetroules (approved):

PR #8487, opened 2025-04-11
Adds DocC catalog to host SwiftPM documentation

heckj:
Adds a `Documentation.docc` directory (a bare DocC catalog) as a stub for migrating existing content in `Documentation/` from general markdown into DocC, as well as providing a base for expanding that documentation. Follow up PRs will bring over and convert content from the `Documentation` directory more piecemeal

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -419,7 +419,7 @@ let package = Package(
         // MARK: Documentation
         
         .target(
-            name: "PackageManagerDocs",
+            name: "PackageManagerDocs"
         ),
         
         // MARK: Package Manager Functionality
--- a/Package.swift
+++ b/Package.swift
@@ -417,11 +417,11 @@ let package = Package(
         ),
 
         // MARK: Documentation
-        
+
         .target(
             name: "PackageManagerDocs"
         ),
-        
+
         // MARK: Package Manager Functionality
 
         .target(
@@ -1020,12 +1020,6 @@ if ProcessInfo.processInfo.environment["SWIFTPM_LLBUILD_FWK"] == nil {
     ]
 }
 
-if ProcessInfo.processInfo.environment["SWIFTPM_DOCC_ENABLED"] != nil {
-    package.dependencies += [
-        .package(url: "https://github.com/swiftlang/swift-docc-plugin", from: "1.1.0"),
-    ]
-}
-
 if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
     package.dependencies += [
         .package(url: "https://github.com/swiftlang/swift-tools-support-core.git", branch: relatedDependenciesBranch),
@@ -1040,6 +1034,8 @@ if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
         .package(url: "https://github.com/apple/swift-collections.git", "1.0.1" ..< "1.2.0"),
         .package(url: "https://github.com/apple/swift-certificates.git", "1.0.1" ..< "1.6.0"),
         .package(url: "https://github.com/swiftlang/swift-toolchain-sqlite.git", from: "1.0.0"),
+        // For use in previewing documentation
+        .package(url: "https://github.com/swiftlang/swift-docc-plugin", from: "1.1.0"),
     ]
 } else {
     package.dependencies += [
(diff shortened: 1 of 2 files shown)


bkhouri:
Thanks for this @heckj . Is it currently possible to generate the documentation at-desk?

bkhouri:

heckj:
@bkhouri yes, definitely. The easiest path, since this is a bare DocC catalog with no symbols, is to use the command "docc preview" - on macOS, quick path to that is `xcrun docc preview`. Right now this is only a placeholder into which to shift (or rewrite) content into from the existing `Documentation` directory and elsewhere. I also have a parallel PR to add the docc-plugin to this project (gated on an environment variable), with the intention to make it easy to use the `swift package generate-documentation` command to generate documentation for targets in the Swift package, but ran into some bugs with that flow (#8486).

Once #8486 is sorted, it may be handle to reference the bare docc catalog within it's own target to make content generation a bit easier (using the package commands instead of the raw docc commands), or to link up additional targets for reference documentation (such as PackageDescription) - but all that's quite possible to add later and after we've sorted what's happening with the doc generation failure bug.

Note: Running DocC locally is known to have issues on Windows, as I understand it because of the use of ":" in the filenames.

heckj:
@swift-ci test

bkhouri:
I'm glad to know that documentation generation from `swift package generate-documentation` is on the roadmap .

bkhouri (approved):

heckj:
I moved adding an ENV variable gated dependency to docc-plugin into this PR and move the DocC catalog into a target. The `EmptyFile.swift` in the target is a workaround to a Package target not supporting hosting a DocC catalog by itself. With these in place, we can preview documentation updates using the command:

```
SWIFTPM_DOCC_ENABLED=1 swift package --disable-sandbox preview-documentation --target PackageManagerDocs
```

You can also build and view the documentation in Xcode 16.2 (but **not** Xcode 16.3 - there's a regression in DocC between 6.0 and 6.1 that causes the documentation builds to crash when using the latest release, tracked down in #8486)

Putting this into a target also allows us to use the Snippets feature to verify Swift code samples used in the documentation compile and don't fall out of date.

heckj:
@swift-ci test

bripeticca:
@swift-ci test windows

Package.swift
```diff
@@ -1014,6 +1020,12 @@ if ProcessInfo.processInfo.environment["SWIFTPM_LLBUILD_FWK"] == nil {
     ]
 }
 
+if ProcessInfo.processInfo.environment["SWIFTPM_DOCC_ENABLED"] != nil {
+    package.dependencies += [
```
bripeticca, line 1024: suggestion: I don't think we'll need to guard this dependency on swift-package-manager, might be better to just remove this :)

bripeticca (dismissed):

Package.swift
bripeticca: (meant to mark this as "Requested Changes"!)

bripeticca: Using this comment to mark this as a request for the above changes.

bripeticca:
@swift-ci please test

heckj:
@swift-ci please test windows

Package.swift
```diff
@@ -1028,6 +1034,8 @@ if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
         .package(url: "https://github.com/apple/swift-collections.git", "1.0.1" ..< "1.2.0"),
         .package(url: "https://github.com/apple/swift-certificates.git", "1.0.1" ..< "1.6.0"),
         .package(url: "https://github.com/swiftlang/swift-toolchain-sqlite.git", from: "1.0.0"),
+        // For use in previewing documentation
+        .package(url: "https://github.com/swiftlang/swift-docc-plugin", from: "1.1.0"),
```
MaxDesiatov, line 1038: What's the reason for unconditionally including this dependency? If I'm building SwiftPM only for development or testing, this will still be cloned, with its targets included in build manifests, making the build slower. What happened to the environment variable that only conditionally included this dependency?

Package.swift
MaxDesiatov: This dependency makes all the build slower even if it's not used

MaxDesiatov:
@heckj would you mind making the dependency conditional in a follow-up PR? Thanks!

PR #8508, opened 2025-04-15
Add default values to testing framework options

johnbute:
Adding default values when initializing packages

Motivation:

Currently, we try to encourage users to use swift testing as much as possible, as the current logic is as follows:

if users do not explicitly mention disabling swift testing, the initialized package will initialize as if the user wrote --enable-swift-test.
Likewise, currently, unless the package initialized is a macro, or if the user explicitly mentions --enable-xctest, then the xctest framework would not be used.
This change will allow users to see that the current defaults, improving the --help screen.

Modifications:

The only modifications I have done was change the flags' types from optional to boolean, then added a validation function that checks if the user explicitly mentioned either enabling or disabling swift testing.

Result:

Users, when writing --help, will see the default values for --enable-swift-test and --enable-xctest.

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -615,7 +615,7 @@ public struct LinkerOptions: ParsableArguments {
 @_spi(SwiftPMInternal)
 public struct TestLibraryOptions: ParsableArguments {
     public init() {}
-
+    
     /// Whether to enable support for XCTest (as explicitly specified by the user.)
     ///
     /// Callers will generally want to use ``enableXCTestSupport`` since it will
@@ -623,8 +623,14 @@ public struct TestLibraryOptions: ParsableArguments {
     @Flag(name: .customLong("xctest"),
           inversion: .prefixedEnableDisable,
           help: "Enable support for XCTest")
-    public var explicitlyEnableXCTestSupport: Bool?
+    public var enableXCTestSupport: Bool = false
 
+    private var didSetEnableXCTestSupport = false
+
+    public var explicitlyEnableXCTestSupport: Bool? {
+        didSetEnableXCTestSupport ? enableXCTestSupport : nil
+    }
+    
     /// Whether to enable support for Swift Testing (as explicitly specified by the user.)
     ///
     /// Callers will generally want to use ``enableSwiftTestingLibrarySupport`` since it will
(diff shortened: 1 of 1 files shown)


Sources/CoreCommands/Options.swift
```diff
@@ -615,25 +615,36 @@ public struct LinkerOptions: ParsableArguments {
 @_spi(SwiftPMInternal)
 public struct TestLibraryOptions: ParsableArguments {
     public init() {}
-
+
```
MaxDesiatov, line 618: trailing whitespace change
```suggestion

```

Sources/CoreCommands/Options.swift
```diff
@@ -642,6 +653,20 @@ public struct TestLibraryOptions: ParsableArguments {
...
+        let args = CommandLine.arguments
+        
+        didSetEnableXCTestSupport = args.contains {
+            $0 == "--enable-xctest" || $0 == "--disable-xctest"
+        }
+
+        didSetEnableSwiftTestingLibrarySupport = args.contains {
+            $0 == "--enable-swift-testing" || $0 == "--disable-swift-testing"
+        }
+
```
MaxDesiatov, line 667: redundant whitespace

```suggestion
```

Sources/CoreCommands/Options.swift
```diff
@@ -615,25 +615,36 @@ public struct LinkerOptions: ParsableArguments {
...
           inversion: .prefixedEnableDisable,
           help: "Enable support for Swift Testing")
-    public var explicitlyEnableSwiftTestingLibrarySupport: Bool?
+    public var enableSwiftTestingLibrarySupport: Bool = true
 
+    private var didSetEnableSwiftTestingLibrarySupport = false
+    
+    public var explicitlyEnableSwiftTestingLibrarySupport: Bool? {
+        didSetEnableSwiftTestingLibrarySupport ? enableSwiftTestingLibrarySupport : nil
+    }
```
MaxDesiatov, line 647: Properties should be separated with a newline

```suggestion
    }

```

Sources/CoreCommands/Options.swift
```diff
@@ -642,6 +653,20 @@ public struct TestLibraryOptions: ParsableArguments {
           help: .private)
     public var explicitlyEnableExperimentalSwiftTestingLibrarySupport: Bool?
 
+    /// Validate if user explicitly enables/disables testing framework.
+    mutating public func validate() throws {
+        let args = CommandLine.arguments
+        
+        didSetEnableXCTestSupport = args.contains {
+            $0 == "--enable-xctest" || $0 == "--disable-xctest"
```
MaxDesiatov, line 661: @rauhul @natecook1000 is this even the right way to achieve this with ArgumentParser? Raw string comparison feels icky here, but AFAIU it's only done to somehow achieve this?

> Users, when writing --help, will see the default values for --enable-swift-test and --enable-xctest.

johnbute: This was the only solution I could have figured out that kept the logic in place, but even with default values for swift testing and enable xctesting, it could be confusing for users who initialize a macro packages knowing swift-testing is enabled by default that their new macro package does not contain swift testing.

Plus, raw string comparison from the command line is indeed brittle and icky

PR #8518, opened 2025-04-17
Add an .editorconfig file

jakepetroules:
This encodes our indentation size and line ending preferences and such. This is an exact copy of the .editorconfig in swift-build, which uses the same conventions.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+xcode_trim_whitespace_on_empty_lines = true
+
+[*.{yml,yaml}]
+indent_size = 2


jakepetroules:
@swift-ci test

MaxDesiatov (approved):

MaxDesiatov:

.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+xcode_trim_whitespace_on_empty_lines = true
```
tonyarnold, line 10: Does Xcode not respect the unadorned version of this option?

.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
```
bkhouri, line 1: thought: In the long term, we will want to enforce Swift Format.  How does the `.editorconfig` differ from the `.swiftformat` configuration file?  Which one takes precedence?

Would it make more sense to incorporate these setting in a single place, the `.swiftformat` file (or equivalent)?

.editorconfig
jakepetroules: Xcode has had longstanding behavior that treats trailing whitespace on empty lines distinctly from trailing whitespace on non-empty lines and so the standard editorconfig key applies only to the former. This extra nonstandard key gets us the reasonable behavior we want.

.editorconfig
jakepetroules: .editorconfig and .swift-format serve different purposes. One is _editor_ settings, the other is _formatter_ settings. The former applies to all text files under a directory hierarchy, not only to Swift. And the latter is much more comprehensive and concerned with the semantics and syntax of specific programming languages. They aren't in conflict; we should have both.

(Also, we shouldn't use .swiftformat; that's a third party tool. We should use .swift-format, which is the one that's officially part of the Swift ecosystem.)

.editorconfig
tonyarnold: Thanks, Jake! I’ve wondered about that for a while now - I appreciate the explanation 😊

PR #8525, opened 2025-04-17
Adopt MemberImportVisibility

plemarquand:
Adopt the experimental feature MemberImportVisibility. 

### Motivation:

Due to a weakness in the existing Swift language implementation, a public extension declared in a library module can be accessible to source code that doesn’t directly import that module.

### Modifications:

Adopt MemberImportVisibility on all targets and adjust imports to be explicit. There were a few places where ambiguities were exposed. The most apparent one is the ambiguity between Basics.AbsolutePath and TSCBasics.AbsolutePath. In this case they're functionally equivalent since the Basics version just type aliases the TSCBasics version. There were a few others where identically named methods were defined in both a SwiftPM module and an imported dependency.

### Result:

SwiftPM compiles with no code included without a corresponding explicit import.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -153,6 +153,7 @@ let package = Package(
             name: "PackageDescription",
             exclude: ["CMakeLists.txt"],
             swiftSettings: [
+                .enableExperimentalFeature("MemberImportVisibility"),
                 .define("USE_IMPL_ONLY_IMPORTS"),
                 .unsafeFlags(["-package-description-version", "999.0"]),
                 .unsafeFlags(["-enable-library-evolution"]),
--- a/Package.swift
+++ b/Package.swift
@@ -296,6 +296,7 @@ let package = Package(
             dependencies: ["Basics"],
             exclude: ["CMakeLists.txt"],
             swiftSettings: [
+                .enableExperimentalFeature("MemberImportVisibility"),
                 .unsafeFlags(["-static"]),
             ]
         ),
--- a/Package.swift
+++ b/Package.swift
@@ -876,7 +876,10 @@ let package = Package(
         ),
         .testTarget(
             name: "PackageGraphTests",
-            dependencies: ["PackageGraph", "_InternalTestSupport"]
+            dependencies: ["PackageGraph", "_InternalTestSupport"],
+            swiftSettings: [
+                .enableExperimentalFeature("MemberImportVisibility"),
+            ],
         ),
         .testTarget(
             name: "PackageGraphPerformanceTests",
--- a/Package.swift
+++ b/Package.swift
@@ -203,6 +203,7 @@ let package = Package(
             ],
             exclude: ["CMakeLists.txt"],
             swiftSettings: [
+                .enableExperimentalFeature("MemberImportVisibility"),
                 .enableExperimentalFeature("AccessLevelOnImport"),
                 .unsafeFlags(["-static"]),
             ]
--- a/Package.swift
+++ b/Package.swift
@@ -466,6 +466,7 @@ let package = Package(
             ],
             exclude: ["CMakeLists.txt"],
             swiftSettings: [
+                .enableExperimentalFeature("MemberImportVisibility"),
                 .unsafeFlags(["-static"]),
             ]
         ),
--- a/Package.swift
+++ b/Package.swift
@@ -220,6 +220,7 @@ let package = Package(
             ],
             exclude: ["CMakeLists.txt"],
             swiftSettings: [
+                .enableExperimentalFeature("MemberImportVisibility"),
                 .enableExperimentalFeature("StrictConcurrency"),
                 .enableExperimentalFeature("AccessLevelOnImport"),
                 .enableExperimentalFeature("InternalImportsByDefault"),
--- a/Package.swift
+++ b/Package.swift
@@ -171,6 +171,7 @@ let package = Package(
             // messing with the manifest loader.
             dependencies: ["PackageDescription"],
             swiftSettings: [
+                .enableExperimentalFeature("MemberImportVisibility"),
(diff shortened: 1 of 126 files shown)


plemarquand:
@swift-ci test

jakepetroules (approved):
Thanks!

Can I ask you to factor the `enableExperimentalFeature("MemberImportVisibility")` statements into a common swiftSettings array at the top of the package, though, as in [Swift Build](https://github.com/swiftlang/swift-build/blob/main/Package.swift#L26)?

We'll want to add more of these over time, and it's much easier if we can apply to the whole package at once without having to copy-paste 20 times.

plemarquand:
Looks like there is a [compile failure](https://ci.swift.org/job/swift-package-manager-with-xcode-self-hosted-PR-osx/4754/) on macOS when adopting this feature on 6.0.3. Going to have to dig deeper to see if this is something we can work around.

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test windows

plemarquand:

plemarquand:

plemarquand:
@swift-ci please test macos

plemarquand:

Sources/PackageGraph/ModulesGraph.swift
```diff
@@ -399,12 +401,12 @@ enum GraphError: Error {
 ///
 /// - Complexity: O(v + e) where (v, e) are the number of vertices and edges
 /// reachable from the input nodes via the relation.
-func topologicalSort<T: Identifiable>(
+func topologicalSortIdentifiable<T: Identifiable>(
```
cmcgee1024, line 404: question: why this rename here? Maybe there was a collision in TSC?

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
```diff
@@ -989,7 +990,7 @@ extension PackagePIFProjectBuilder {
 
 private struct PackageRegistrySignature: Encodable {
     enum Source: Encodable {
-        case registry(url: URL)
+        case registry(url: Foundation.URL)
```
cmcgee1024, line 993: praise: It's good to be explicit here given that TSCUtility has its own URL, deprecated in favour of the much more well known and expected Foundation one.

cmcgee1024 (approved):
The changes look good to me, and this change can make unexpected usages much more apparent moving forward.

I've added a question and one other comment, nothing blocking.

Sources/PackageGraph/ModulesGraph.swift
plemarquand: Correct, there is a `TSCBasic.topologicalSort`. This file requires the `spm_createDictionary`, which is defined as an extension method on `Array` inside `TSCBasic`. Because its defined in an extension it means we can't simply import the methods/structs/classes we need from TSCBasic and have to import the whole module. This imports `TSCBasic.topologicalSort` and creates the naming conflict.

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci test windows

Package.swift
```diff
@@ -152,7 +165,7 @@ let package = Package(
         .target(
             name: "PackageDescription",
             exclude: ["CMakeLists.txt"],
-            swiftSettings: [
+            swiftSettings: commonExperimentalFeatures + [
```
bkhouri, line 168: question: should we be using `swift6CompatibleExperimentalFeatures` instead of `commonExperimentalFeatures` in the event a developer is using Swift 6.0.x to build the project?

Sources/PackageGraph/ModulesGraph.swift
bkhouri: Is `TSCBasic.topologicalSort` used by this module?  Is there advantages to only importing TSC symbols that are used by this file instead of importing all of them?

bkhouri (commented):
I very much favour this change to make import more explicit.

My primary concern is posted as a comment in `Package.swift`.

Sources/PackageGraph/ModulesGraph.swift
plemarquand: This file requires the `spm_createDictionary`, which is defined as an extension method on Array inside TSCBasic. You can't import extension methods, so you must import the whole module, which implicitly imports `TSCBasic.topologicalSort` and so conflicts.

Package.swift
plemarquand: We want these to apply to any toolchain. The `swift6CompatibleExperimentalFeatures` var only exists for targets that currently fail to build with `MemberImportVisibility` on in 6.0.3.

Package.swift
```diff
@@ -235,7 +248,7 @@ let package = Package(
                 .product(name: "SystemPackage", package: "swift-system"),
             ],
             exclude: ["CMakeLists.txt", "Vendor/README.md"],
-            swiftSettings: [
+            swiftSettings: swift6CompatibleExperimentalFeatures + [
```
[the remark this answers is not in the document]
bkhouri: Replied here: https://github.com/swiftlang/swift-package-manager/pull/8525#discussion_r2056239029

Package.swift
bkhouri: Do we understand why those targets fail to build on 6.0.3?   Can we raise an issue to ensure we get this addressed?

bkhouri (approved):

Package.swift
plemarquand: They don't fail on 6.1, so my assumption is it's already fixed.

PR #8532, opened 2025-04-20
[Commands] Initial implementation of `swift package add-setting` command

xedin:
### Motivation:

Adds a way to programmatically insert new settings into a package manifest.

Currently only some Swift settings are supported, namely: `enable{Upcoming, Experimental}Feature`, `swiftLanguageMode` and `strictMemorySafety`; but the command could be expanded to support more Swift (C, C++, linker) settings in the future.

### Modifications:

- Add a new "package" sub-command named "add-setting" that accepts a target and a list of `--swift Name[=Value]?` pairs for each new swift setting to add. Each setting would make sure that manifest is new enough to support it.
- Add new manifest refactoring action - AddSwiftSetting that would add `swiftSettings:` to a target or modify the existing one.
- Expands existing way to check whether manifest is too old to support custom versions. This is doe to make sure that users don't add settings that are not supported by the manifest tools version i.e. `swiftLanguageMode` was introduced in 6.0 tools.

### Result:

A new `swift package add-setting --target <name> [--swift Name[=Value]?]+`  command that allows users to programmatically add new settings to package manifests.

### DIFF
--- a/Sources/Commands/PackageCommands/AddSetting.swift
+++ b/Sources/Commands/PackageCommands/AddSetting.swift
@@ -1,0 +1,148 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import Basics
+import CoreCommands
+import Foundation
+import PackageGraph
+import PackageModel
+import PackageModelSyntax
+import SwiftParser
+import TSCBasic
+import TSCUtility
+import Workspace
+
+extension SwiftPackageCommand {
+    struct AddSetting: SwiftCommand {
+        /// The Swift language setting that can be specified on the command line.
+        enum SwiftSetting: String, Codable, ExpressibleByArgument, CaseIterable {
+            case experimentalFeature
+            case upcomingFeature
+            case languageMode
+            case strictMemorySafety
+        }
+
+        package static let configuration = CommandConfiguration(
+            abstract: "Add a new setting to the manifest"
+        )
+
+        @OptionGroup(visibility: .hidden)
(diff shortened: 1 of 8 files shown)


Sources/Commands/PackageCommands/AddSetting.swift
```diff
@@ -0,0 +1,148 @@
...
+                            Parser.parse(source: buffer)
+                        }
+                    }
+                }
+
+                let editResult: PackageEditResult
+
+                switch setting {
+                case .experimentalFeature:
+                    editResult = try AddSwiftSetting.experimentalFeature(
```
plemarquand, line 107: What happens if the user adds the same setting a second time? And if that setting has a different value than the already added setting?

xedin: The compiler would take the last value.

xedin: Yeah, I looked into that while working on that and without some support from swift-syntax to match the nodes in general we won't be able to do match...

Sources/Commands/PackageCommands/AddSetting.swift
```diff
@@ -0,0 +1,148 @@
...
+        @OptionGroup(visibility: .hidden)
+        var globalOptions: GlobalOptions
+
+        @Option(help: "The target to add the setting to")
+        var target: String
+
+        @Option(
+            name: .customLong("swift"),
+            parsing: .unconditionalSingleValue,
+            help: "The Swift language setting(s) to add. Supported settings: \(SwiftSetting.allCases.map(\.rawValue))"
```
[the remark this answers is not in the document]
xedin: Sure, done!

Sources/Commands/PackageCommands/AddSetting.swift
xedin: I didn't do that because it would lead to false positive warnings since we cannot really figure out what is what so if they have the setting with different conditions we'll just keep warning about it for no reason.

xedin:
@swift-ci please test

xedin:
Re-organized imports a bit to avoid imported what we don't need here.

xedin:
@swift-ci please test

xedin:
@swift-ci please test

xedin:
@swift-ci please test

xedin:

Issue #8519, opened 2025-04-17
`swift package add-dependency` will add the same dependency multiple times causing build failures


euanh:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

`swift package add-dependency` is not idempotent.  It will add a dependency entry for a package to `Package.swift` even if that package is already a dependency.   This makes the package fail to build.

### Expected behavior

`swift package add-dependency` should refuse to add an existing dependency.

### Actual behavior

`swift package add-dependency` adds duplicate dependencies, leading to build failures.

### Steps to reproduce

1. Create a new package
```
% swift package init --type executable
Creating executable package: dependencies
Creating Package.swift
Creating Sources/
Creating Sources/main.swift
% cat Package.swift 
// swift-tools-version: 6.1
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "dependencies",
    targets: [
        // Targets are the basic building blocks of a package, defining a module or a test suite.
        // Targets can depend on other targets in this package and products from dependencies.
        .executableTarget(
            name: "dependencies"),
    ]
)
% swift build
Building for debugging...
[8/8] Applying dependencies
Build complete! (1.41s)
```

2. Add a dependency on swift-argument-parser and rebuild - this succeeds.
```
% swift package add-dependency https://github.com/apple/swift-argument-parser --from 1.5.0      
Updating package manifest at Package.swift... done.
% cat Package.swift
// swift-tools-version: 6.1
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "dependencies",
    dependencies: [
        .package(url: "https://github.com/apple/swift-argument-parser", from: "1.5.0"),
    ],
    targets: [
        // Targets are the basic building blocks of a package, defining a module or a test suite.
        // Targets can depend on other targets in this package and products from dependencies.
        .executableTarget(
            name: "dependencies"),
    ]
)
% swift build
Fetching https://github.com/apple/swift-argument-parser from cache
Fetched https://github.com/apple/swift-argument-parser from cache (0.56s)
Computing version for https://github.com/apple/swift-argument-parser
[...]

PR #8534, opened 2025-04-21
Make add-dependency idempotent

plemarquand:
### Motivation:

If the same dependency is added twice with different version specifiers the manifest is no longer valid. If the same dependency is added twice with the same version specifier it remains valid, but less maintainable.

### Modifications:

If a manifest already contains the exact dependency and version that is being added via a `add-dependency` call, don't add it twice. Instead, do nothing.

If a manifest contains a dependency with the supplied URL but a different version qualifier then throw an error explaining that the supplied dependency has already been added.

### Result:

A manifest that continues to parse correctly.

Issue: #8519

### DIFF
--- a/Sources/PackageModelSyntax/AddPackageDependency.swift
+++ b/Sources/PackageModelSyntax/AddPackageDependency.swift
@@ -44,7 +44,7 @@ public enum AddPackageDependency {
             throw ManifestEditError.cannotFindPackage
         }
 
-        guard try checkExistingDependency(
+        guard try !dependencyAlreadyAdded(
             dependency,
             in: packageCall
         ) else {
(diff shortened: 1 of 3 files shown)


plemarquand:
@swift-ci please test

Sources/PackageModelSyntax/AddPackageDependency.swift
```diff
@@ -55,6 +62,52 @@ public enum AddPackageDependency {
         )
     }
 
+    /// Check that the given package dependency doesn't already exist in the manifest.
+    /// If the exact same dependency already exists, `false` is returned to indicate
+    /// that no edits are needed. If a different dependency with the same id or url
+    /// with different arguments exists, an error is thrown.
+    private static func checkExistingDependency(
```
bkhouri, line 69: question: I know this is marked as a private function, but is there value is validate this function in isolation to ensure we can test all code paths to ensure the behaviour is as expected, as opposed to testing the implementation?

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -875,109 +880,178 @@ class PackageCommandTestCase: CommandsBuildProviderTestCase {
...
+                    ],
+                    targets: [ .target(name: "client", dependencies: [ "library" ]) ]
+                )
+            """
+            let expected = #".package(url: "https://github.com/swiftlang/swift-syntax.git", exact: "601.0.1"),"#
+
+            try await executeAddURLDependencyAndAssert(
+                packagePath: path,
+                initialManifest: manifest,
+                url: "https://github.com/swiftlang/swift-syntax.git",
```
bkhouri, line 966: suggestion: can we replace string `"http://.../swift-syntax.git"` with `url`?

bkhouri, line 966: question: Does SwiftPM treat a package dependency of `"https://github.com/swiftlang/swift-syntax.git"` the same as `"https://github.com/swiftlang/swift-syntax"`?  Is there value if documenting this specific case?

what about repositories that has the same name, but belong to different organizations/users?

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

PR #8548, opened 2025-04-22
build-using-self use a self build of swift-test instead

dschaefer2:
We are seeing hangs in swift test on Windows. Instead of build-using-self using the underlying toolchain, build swift-test and use that for the tests. When we make improvements to it to fix the underlying problem, we will be able to take advantage of that fix right away.

### DIFF
--- a/Utilities/build-using-self
+++ b/Utilities/build-using-self
@@ -152,19 +152,20 @@ def main() -> None:
             shlex.split(f"swift build --configuration {args.config} {ignore}"),
         )
 
-        swift_testing_arg= "--enable-swift-testing" if args.enable_swift_testing else ""
-        xctest_arg= "--enable-xctest" if args.enable_swift_testing else ""
-        call(
-            shlex.split(f"swift run swift-test --configuration {args.config} --parallel {swift_testing_arg} {xctest_arg} --scratch-path .test {ignore}"),
-        )
+        if os.name != "nt": # turn off for Windows until we get the hang resolved
+            swift_testing_arg= "--enable-swift-testing" if args.enable_swift_testing else ""
+            xctest_arg= "--enable-xctest" if args.enable_swift_testing else ""
+            call(
+                shlex.split(f"swift run swift-test --configuration {args.config} --parallel {swift_testing_arg} {xctest_arg} --scratch-path .test {ignore}"),
+            )
 
-        integration_test_dir = REPO_ROOT_PATH / "IntegrationTests"
-        call(
-            shlex.split(f"swift package --package-path {integration_test_dir} update"),
-        )
-        call(
-            shlex.split(f"swift run swift-test --package-path {integration_test_dir} --parallel {ignore}"),
-        )
+            integration_test_dir = REPO_ROOT_PATH / "IntegrationTests"
+            call(
+                shlex.split(f"swift package --package-path {integration_test_dir} update"),
+            )
+            call(
+                shlex.split(f"swift run swift-test --package-path {integration_test_dir} --parallel {ignore}"),
+            )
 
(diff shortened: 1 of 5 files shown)


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test windows

Utilities/build-using-self
```diff
@@ -144,31 +145,27 @@ def main() -> None:
             shlex.split("swift --version"),
         )
 
-        # call(
```
bkhouri, line 147: praise: It's always nice seeing commented code being removed!

Utilities/build-using-self
```diff
@@ -144,31 +145,27 @@ def main() -> None:
...
-        call(
-            shlex.split("swift package update"),
-        )
-        call(
-            shlex.split(
-                f"{swiftpm_bin_dir / 'swift-test'} --parallel",
-                posix=(os.name == "posix"), # must be set correctly, otherwhsie shlex.split("C:\\Foo\\bar") become ['CFoobar']
-            ),
-        )
+        if os.name != "nt": # turn off for Windows until we get the hang resolved
```
bkhouri, line 155: Do we know what may be causing the hangs?  It's quite coincidental the hangs started to occurs after the self hosted windows pipeline was changes from using the nightly toolchain, to the nightly 6.1 toolchain.

The self-hosted windows toolchain used the official docker image.  should we try to see if the released 6.1 image has the same hangs?

Utilities/build-using-self
```diff
@@ -144,31 +145,27 @@ def main() -> None:
...
+            swift_testing_arg= "--enable-swift-testing" if args.enable_swift_testing else ""
+            xctest_arg= "--enable-xctest" if args.enable_swift_testing else ""
+            call(
+                shlex.split(f"swift run swift-test --configuration {args.config} --parallel {swift_testing_arg} {xctest_arg} --scratch-path .test {ignore}"),
+            )
+
+            integration_test_dir = REPO_ROOT_PATH / "IntegrationTests"
```
bkhouri, line 162: to make this code more robust to changes, it would be better to change the `change_directory(REPO_ROOT_PATH` context manager to return the directory, and use that path here instead of `REPO_ROOT_PATH`.

That is, change  `with change_directory(REPO_ROOT_PATH):` to `with change_directory(REPO_ROOT_PATH) as my_path:` and then set `integration_test_dir = my_path / "IntegrationTests"` or just set it to a relative path (ie: `integration_test_dir = Path("./IntegrationTests")`

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test

Utilities/build-using-self
dschaefer2: Yes, the 6.1 swiftpm has a bug in it. It's still there in 6.2. We should probably fix it.

Utilities/build-using-self
dschaefer2: We'll be back here to re-enable them on Windows. We can clean things up then. Let's just focus on re-enabling the Windows CI for now.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

jakepetroules (approved):

PR #8549, opened 2025-04-23
Multi-thread pulling of out put from AsyncProcess

dschaefer2:
We are seeing hangs on Windows when SwiftPM tests itself. Looking at how AsyncProcess pulls stdout and stderr, I can see that there may be conditions where the streams aren't fully drained causing the hang. The proper way to do this, like it's done on the other platforms is to create a thread for each stream.


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

jakepetroules (approved):
Heh, I've lost count of how many times I've fixed this same problem with uses of `Process`. Really looking forward to adopting swift-subprocess everywhere, which is async-native.

Looks like readabilityHandler is invoked on a private serial Dispatch queue owned by the FileHandle; I'm not sure this is a scenario where a hang could be caused, but it's hard to tell since the Swift Concurrency / GCD interaction is complex. This approach is worth trying though, since a pthread provides a pretty strong independence guarantee.

Tests/PackageLoadingTests/PDLoadingTests.swift
```diff
@@ -17,7 +17,6 @@ import _InternalTestSupport
 import XCTest
 
 class PackageDescriptionLoadingTests: XCTestCase, ManifestLoaderDelegate {
-    lazy var manifestLoader = ManifestLoader(toolchain: try! UserToolchain.default, delegate: self)
```
bkhouri, line 20: suggestion: can we instantiate `manifestLoader` in a `setUp(...)` call to ensure each test instantiates/creates its own object?

Tests/PackageLoadingTests/PDLoadingTests.swift
```diff
@@ -17,7 +17,6 @@ import _InternalTestSupport
...
     var parsedManifest = ThreadSafeBox<AbsolutePath>()
```
bkhouri, line 20: suggestion: can we instantiate `parsedManifest` in a `setUp(...)` call to ensure each test instantiates/creates its own object?

dschaefer2:
Still hanging, even removing the --parallel. Trying one more attempt at going pure win32.

But, yes, subprocess would come in handy right now.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:

dschaefer2:

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test self hosted windows.

dschaefer2:

dschaefer2:

dschaefer2:
Huh, that was it. I'm going to abandon this and go back to using Foundation.Process but with the multi-threaded reads. And then make sure multiple gits aren't running at the same time.

PR #8555, opened 2025-04-24
Tests: Augment SwiftBuild build provider test coverage

bkhouri:
Augment some test to run against the `native` and `swiftbuild` build providers to gain extra confidence with the swift build integration.


bkhouri:
@swift-ci test

Sources/_InternalTestSupport/XCTAssertHelpers.swift
```diff
@@ -50,6 +50,10 @@ public func XCTSkipIfCI(file: StaticString = #filePath, line: UInt = #line) thro
     }
 }
 
+public func XCTSkipSwiftBuildTodo(because reason: String) throws {
+    throw XCTSkip("SWBINTTODO: \(reason)")
```
cmcgee1024, line 54: thought: The unique string "SWBINTTODO" was introduced for "grep"ability over the code back. Since grep will now only find one instance of the label here, it's not very useful, and can probably be removed since the new grep expression will be "SwiftBuildTodo" instead.

Note that there are other instances of this string in the production code (ie. in Sources/), so those will need to be changed to match this new grepable name so that there's one string that will find everything. Those string don't need to call this function however. They just need to be changed to match its suffix.

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -1412,15 +1414,10 @@ class PackageCommandTestCase: CommandsBuildProviderTestCase {
             let binFile = buildPath.appending(components: try UserToolchain.default.targetTriple.platformBuildPathComponent, "debug", "Bar")
             XCTAssertFileExists(binFile)
             XCTAssert(localFileSystem.isDirectory(buildPath))
-            // Clean, and check for removal of the build directory but not Packages.
```
cmcgee1024, line 1415: question: why remove this clean and check afterwards?

bkhouri: Based on the test, it's verifying the `package reset`, which would apply the same operation as `package clean` and more.

Sources/_InternalTestSupport/XCTAssertHelpers.swift
bkhouri: We can rename the function from `XCTSkipSwiftBuildTodo` to `XCTSkipSWBINTTODO` to preserve consistency.

cmcgee1024: I think that it would be better to go with the new name since it looks much better as a function name. It just needs to changed in Sources/ too.

bkhouri:
@swift-ci test

bkhouri:
@swift-ci test

bkhouri:
Closing PR is favour of smaller PR's.

PR #8561, opened 2025-04-26
Factor the working directory skip into a common function

jakepetroules:
Also skips on OpenBSD, which has the same issue.


jakepetroules:
@swift-ci test

dschaefer2 (approved):

bkhouri (approved):

PR #8563, opened 2025-04-27
Fix a number of minor curated documentation issues

jakepetroules:
Fixing wrong parameter names, adding missing APIs to the curation, removing a couple non-existent things.


jakepetroules:
@swift-ci test

plemarquand (approved):

PR #8568, opened 2025-04-28
Avoid building targets that are conditional on platform when building all

cmcgee1024:
When using the swiftbuild build system, and building all targets with a project
that contains a target that is depended on only when on a specific platform that
target should be skipped. Make the All(Excluding|Including)Tests generated
targets depend only on targets that are reachable from the roots through
dependencies that are either unconditional or are conditioned on the current
platform.

### DIFF
--- a/Sources/SwiftBuildSupport/PIFBuilder.swift
+++ b/Sources/SwiftBuildSupport/PIFBuilder.swift
@@ -100,7 +100,8 @@ public final class PIFBuilder {
     func generatePIF(
         prettyPrint: Bool = true,
         preservePIFModelStructure: Bool = false,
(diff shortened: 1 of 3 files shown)


cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Windows

jakepetroules (approved):
Wow, that is surprisingly unintuitive... but it does appear to match the current behavior (I had to check to make sure I was understanding correctly what's happening).

PR #8571, opened 2025-04-29
Rename Build module to NativeBuildSupport

jakepetroules:
This module is intended to contain code exclusive to the native build system, in order to help more clearly identify the delineation between the three build system backends, and identify layering violations and new higher-level abstractions that will need to be created to delegate to the appropriate backend (SourceKit support is one major example).


jakepetroules:
@swift-ci test

MaxDesiatov (approved):

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test windows

Package.swift
```diff
@@ -500,7 +499,9 @@ let package = Package(
                 "PackageGraph",
             ],
             exclude: ["CMakeLists.txt", "README.md"],
-            swiftSettings: commonExperimentalFeatures
+            swiftSettings: commonExperimentalFeatures + [
+                .unsafeFlags(["-static"]),
```
bkhouri, line 503: question: out of curiosity, why did we add the `-static` flag here if we this PR is renaming `BuildSupport` to `NativeBuildSupport`?

jakepetroules: All those targets are supposed to have it, it was accidentally missed from this one before.

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test windows

Package.swift
dschaefer2: I'm not sure we need it any more when building with Swift 6.1.

PR #8574, opened 2025-04-30
Use correct FileSystem in the PIF builder for Swift Build

pmattos:
### Motivation:

Ensure the PIF builder follows the same `FileSystem` as the remaining code in the `SwiftBuildSupport` target.

### Modifications:

Keep around a file system reference in `PackagePIFBuilder` and use that across the PIF builder.

### Result:

All code in `SwiftBuildSupport` is now based on the same `FileSystem`.


pmattos:
@swift-ci test

jakepetroules (approved):

PR #8612, opened 2025-05-02
Add more context for validation errors for target properties.

yyvch:
Add more context for validation errors for target properties.

### Motivation:

I came across a post https://forums.swift.org/t/getting-error-on-package-resolve/75419 which reports the error message to be not clear. I thought I could clarify the message.

```
error: 'xerr': target 'Penny' contains a value for disallowed property 'settings'
```

### Modifications:

I've updated the message for the error, so it includes more details to narrow down what is the culprit:
 - highlight that the property depends on the target type
 - show the value of the property, so that there's a hint what line in the Package.swift lead to the error (for example, linkerSettings or cxxSettings)

### Result:

Before:

```
error: 'xerr': target 'Penny' contains a value for disallowed property 'settings'
```

After:

```
error: 'xerr': target 'Penny' is assigned a property 'settings' which is not accepted for
the binary target type. The current value of the property has the following representation: 
[PackageModel.TargetBuildSettingDescription.Setting(
tool: PackageModel.TargetBuildSettingDescription.Tool.linker,
kind: PackageModel.TargetBuildSettingDescription.Kind.linkedFramework("AVFoundation"),
condition: nil)].
```

### DIFF
--- a/Sources/PackageModel/Manifest/TargetDescription.swift
+++ b/Sources/PackageModel/Manifest/TargetDescription.swift
@@ -596,12 +596,12 @@ private enum Error: LocalizedError, Equatable {
     var errorDescription: String? {
         switch self {
         case .binaryTargetRequiresEitherPathOrURL(let targetName):
-            "binary target '\(targetName)' neither defines neither path nor URL for its artifacts"
+            "binary target '\(targetName)' must define either path or URL for its artifacts"
         case .pluginTargetRequiresPluginCapability(let targetName):
             "plugin target '\(targetName)' does not define any plugin capability"
         case .disallowedPropertyInTarget(let targetName, let targetType, let propertyName, let value):
             "target '\(targetName)' is assigned a property '\(propertyName)' which is not accepted " +
-            "for the \(targetType) target type. The current value of the property has " +
+            "for the \(targetType) target type. The current property value has " +
             "the following representation: \(value)."
         }
     }
--- a/Sources/PackageModel/Manifest/TargetDescription.swift
+++ b/Sources/PackageModel/Manifest/TargetDescription.swift
@@ -212,51 +212,237 @@ public struct TargetDescription: Hashable, Encodable, Sendable {
         checksum: String? = nil,
         pluginUsages: [PluginUsage]? = nil
     ) throws {
+        let targetType = String(describing: type)
         switch type {
         case .regular, .executable, .test:
-            if url != nil { throw Error.disallowedPropertyInTarget(targetName: name, propertyName: "url") }
-            if pkgConfig != nil { throw Error.disallowedPropertyInTarget(targetName: name, propertyName: "pkgConfig") }
-            if providers != nil { throw Error.disallowedPropertyInTarget(targetName: name, propertyName: "providers") }
-            if pluginCapability != nil { throw Error.disallowedPropertyInTarget(targetName: name, propertyName: "pluginCapability") }
-            if checksum != nil { throw Error.disallowedPropertyInTarget(targetName: name, propertyName: "checksum") }
+            if url != nil { throw Error.disallowedPropertyInTarget(
+                targetName: name,
+                targetType: targetType,
+                propertyName: "url",
+                value: url ?? "<nil>"
+            ) }
+            if pkgConfig != nil { throw Error.disallowedPropertyInTarget(
+                targetName: name,
+                targetType: targetType,
+                propertyName: "pkgConfig",
+                value: pkgConfig ?? "<nil>"
+            ) }
+            if providers != nil { throw Error.disallowedPropertyInTarget(
+                targetName: name,
+                targetType: targetType,
+                propertyName: "providers",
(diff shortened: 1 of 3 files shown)


yyvch:
@swift-ci please test

Sources/PackageModel/Manifest/TargetDescription.swift
```diff
@@ -212,51 +212,237 @@ public struct TargetDescription: Hashable, Encodable, Sendable {
...
-            if url != nil { throw Error.disallowedPropertyInTarget(targetName: name, propertyName: "url") }
-            if pkgConfig != nil { throw Error.disallowedPropertyInTarget(targetName: name, propertyName: "pkgConfig") }
-            if providers != nil { throw Error.disallowedPropertyInTarget(targetName: name, propertyName: "providers") }
-            if pluginCapability != nil { throw Error.disallowedPropertyInTarget(targetName: name, propertyName: "pluginCapability") }
-            if checksum != nil { throw Error.disallowedPropertyInTarget(targetName: name, propertyName: "checksum") }
+            if url != nil { throw Error.disallowedPropertyInTarget(
+                targetName: name,
+                targetType: targetType,
+                propertyName: "url",
+                value: url ?? "<nil>"
```
bkhouri, line 222: suggestion: can `value` be an optional?  If so, can we handle the `nil` case in the `Error` itself so this can be `value: url`?

Sources/PackageModel/Manifest/TargetDescription.swift
```diff
@@ -404,14 +590,19 @@ import protocol Foundation.LocalizedError
 
 private enum Error: LocalizedError, Equatable {
     case binaryTargetRequiresEitherPathOrURL(targetName: String)
-    case disallowedPropertyInTarget(targetName: String, propertyName: String)
-    
+    case pluginTargetRequiresPluginCapability(targetName: String)
```
bkhouri, line 593: thought (possibly-blocking): It's great to see we added an additional error.   However, would it be possible to add a test case to validate this path (if one does no already exist)?

Sources/PackageModel/Manifest/TargetDescription.swift
```diff
@@ -404,14 +590,19 @@ import protocol Foundation.LocalizedError
...
+    case disallowedPropertyInTarget(targetName: String, targetType: String, propertyName: String, value: String)
+
     var errorDescription: String? {
         switch self {
         case .binaryTargetRequiresEitherPathOrURL(let targetName):
-            return "binary target '\(targetName)' neither defines neither path nor URL for its artifacts"
-        case .disallowedPropertyInTarget(let targetName, let propertyName):
-            return "target '\(targetName)' contains a value for disallowed property '\(propertyName)'"
+            "binary target '\(targetName)' neither defines neither path nor URL for its artifacts"
```
bkhouri, line 599: suggestion: this is not related to your change, but would it be possible to clarify this error message?  maybe something like `"binary target '\(targetName)' must define either path or URL for it's artifacts"`

Sources/PackageModel/Manifest/TargetDescription.swift
```diff
@@ -404,14 +590,19 @@ import protocol Foundation.LocalizedError
...
+        case .pluginTargetRequiresPluginCapability(let targetName):
+            "plugin target '\(targetName)' does not define any plugin capability"
+        case .disallowedPropertyInTarget(let targetName, let targetType, let propertyName, let value):
+            "target '\(targetName)' is assigned a property '\(propertyName)' which is not accepted " +
+            "for the \(targetType) target type. The current value of the property has " +
```
bkhouri, line 604: suggestion: `s/The current value of the property has/The current property value has`

Sources/PackageModel/Manifest/TargetDescription.swift
yyvch: Ack. I will add it.

Sources/PackageModel/Manifest/TargetDescription.swift
bkhouri: This is fine.

Sources/PackageModel/Manifest/TargetDescription.swift
yyvch: Done in 24d87fb.

Sources/PackageModel/Manifest/TargetDescription.swift
yyvch: Done in 24d87fb.

Sources/PackageModel/Manifest/TargetDescription.swift
yyvch: Done in 3816496.

yyvch:
@swift-ci please test

bkhouri:
@swift-ci test windows

bkhouri (approved):

bripeticca (approved):

PR #8613, opened 2025-05-03
[Commands] `swift migrate` command

xedin:
### Motivation:

`swift migrate` could be used to migrate whole package or its individual targets to use the given feature(s) that support migration mode. This is paired with Swift compiler changes to add `:migration` mode to upcoming/experimental features.

### Modifications:

- `UserToolchain` gained an ability to list features supported by the Swift compiler.
- Implementation of the `add-setting` package command has been slightly refactored to allow use of manifest modification logic without having to call the command itself.
- `ModuleBuildDescription` protocol can now list diagnostic files
- Added SwiftMigratedCommand that accepts targets (optional) and features, performs a build and uses the newly added `SwiftFixIt` API to apply fix-its and `AddSwiftSetting` API to enable features.

### Result:

SwiftPM gained a new command `swift migrate --targets <target,...> --to-feature <feature> [--to-feature <...>]`


xedin:
I'm going to add more tests once we actually get a toolchain that supports all the new functionality.

xedin:
@swift-ci please test

Sources/Commands/SwiftMigrateCommand.swift
```diff
@@ -0,0 +1,211 @@
...
+        features: [SwiftCompilerFeature]
+    ) async throws -> BuildSystem {
+        let toolsBuildParameters = try swiftCommandState.toolsBuildParameters
+        var destinationBuildParameters = try swiftCommandState.productsBuildParameters
+
+        // Inject feature settings as flags. This is safe and not as invasive
+        // as trying to update manifest because in adoption mode the features
+        // can only produce warnings.
+        for feature in features {
+            destinationBuildParameters.flags.swiftCompilerFlags.append(contentsOf: [
```
jakepetroules, line 165: This will need to be done through build settings for the swiftbuild backend, not as flags.

Sources/_InternalTestSupport/misc.swift
```diff
@@ -382,6 +382,28 @@ public func executeSwiftTest(
...
+@discardableResult
+public func executeSwiftMigrate(
+    _ packagePath: AbsolutePath?,
+    configuration: Configuration = .Debug,
+    extraArgs: [String] = [],
+    Xcc: [String] = [],
+    Xld: [String] = [],
+    Xswiftc: [String] = [],
+    env: Environment? = nil,
+    buildSystem: BuildSystemProvider.Kind = .native
```
jakepetroules, line 394: Don't default the build system to `native`, we're actively replacing that with the Swift Build backend and don't want to miss places where the tests have to be parameterized across multiple build system backends.

jakepetroules:
Just so I understand this correctly, the flow is:

- `swift migrate` does a build with some extra compiler flags passed, based on the migration options chosen
- Compiler generates some diagnostics files with "fixits" in the form of file paths + line/column ranges + replacement text
- SPM reads the diagnostics files and applies the fixits to the source files programmatically
- SPM programmatically updates the package manifest if possible, based on the migration options chosen

Right?

Sources/_InternalTestSupport/misc.swift
xedin: Every other execute* in this file is defaulted to `.native`, all I did was copy package and chance one line. It sounds like if we want to change that it would have to be done separately and all together so we don't miss anything.

Sources/Commands/SwiftMigrateCommand.swift
xedin: Yeah, I'm not sure how to exactly plug that in, if flags don't work it would also mean that `swift-build` won't support `-Xswiftc` because I think that's where these end up being stored?...

xedin:
@jakepetroules Yes, that is correct.

xedin:

Sources/Commands/SwiftMigrateCommand.swift
xedin: As in, ideally we'd just inject that into the build directly for selected targets or entire build but I don't know an easy way to do that besides specifying them as flags in build options. I don't think we can simply modify ModulesGraph to inject them, that was the original idea I had and I couldn't think of a way to do that during manifest loading which won't be super invasive.

xedin: We could make `Module.buildSettings` to be `package(set) var` instead of `let` I guess but that seems excessive just to support this.

xedin: We'd have to implement our own loading then too that loads then modifies and returns.

xedin: I also checked that `.flags.swiftCompilerFlags` are recorded as `OTHER_SWIFT_FLAGS` in `SwiftBuildSystem.makeBuildParameters()`

xedin:
On thing I forgot to mention - we need to implement category filtering as part of SwiftFixIt still and only apply fix-its related to selected features. I am planning to do that in a follow-up.

Sources/Commands/SwiftMigrateCommand.swift
dschaefer2: There's going to be a lot like this that we'll need to make sure works with the SwiftBuildSystem, e.g. the macro prebuilts feature. All in due time.

Sources/_InternalTestSupport/misc.swift
dschaefer2: We're still working on integrating SwiftBuild. Default will still be .native in this next release.

dschaefer2 (approved):

xedin:
@swift-ci please test

xedin:

xedin:

Sources/_InternalTestSupport/misc.swift
jakepetroules: > Every other execute* in this file is defaulted to `.native`, all I did was copy package and change one line. It sounds like if we want to change that it would have to be done separately and all together so we don't miss anything.

OK, yeah, that sounds fine.

jakepetroules (approved):

xedin:
@dschaefer2, we have decided to make it a sub-command of package instead, so I'll have to make some changes here but the functionally everything is going to stay the same.

xedin:
@swift-ci please test

xedin:

xedin:
@swift-ci please test

xedin:
@swift-ci please test

xedin:

xedin:

xedin:

xedin:

xedin:

xedin:

xedin:

xedin:

xedin:

xedin:

xedin:
@swift-ci please test

xedin:

PR #8620, opened 2025-05-05
[SwiftFixIt] Allow filtering diagnostics by category

xedin:
### Motivation:

Helps with applying only certain fix-its i.e. for features in migration mode or concurrency.

### Modifications:

- Add `categories: Set<String>` parameter to `SwiftFixIt` initializers that is defaulted to `[]`.

### Result:

`swift migrate` can use this to filter based on the feature categories to avoid applying unrelated fix-its.


xedin:
@swift-ci please test

bnbarham (approved):

jakepetroules (approved):

PR #8629, opened 2025-05-07
[Commands] Adjust `swift package migrate` to only apply fix-its to th…

xedin:
…e selected features

### Motivation:

Use new ability of `SwiftFixIt` to filter by category, to make sure that the command only applies fix-its that are related to the selected features and nothing else.

### Modifications:

- `swift package migrate` - Add an argument to `SwiftFixIt` API that lists all of the features selected by the user.

### Result:

Even if there are other warnings with a single fix-it the command is not going to apply them i.e. for `var` -> `let` or `<name>` to `_`.


xedin:
@swift-ci please test

bnbarham (approved):

PR #8633, opened 2025-05-08
Make PackageCollectionModel.V1.Collection and PackageCollectionModel.V1.SignedCollection Sendable

finestructure:
### Motivation:

We are using SwiftPM as a dependency to vend signed package collections on the [Swift Package Index website](https://swiftpackageindex.com).

(There are some more details around why we do this and don't think we have an alternative laid out [in this issue](https://github.com/swiftlang/swift-package-manager/issues/8622).)

After updating to Swift 6, we've had to introduce a `@retroactive @unchecked Sendable` on `SignedCollection` in order to conform it to `Vapor.Content` and send it in web responses.

### Modifications:

This change marks the related types as `Sendable`.

### Result:

With this change, we can drop the retroactive unchecked conformance. We've tested this with our project on a fork here:

https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/actions/runs/14808870777

NB: I'm pretty sure `release/6.1` is not the proper target to land these changes (if they are acceptable) but it was the easiest place for us to test them and keep the change set minimal.

### DIFF
--- a/Sources/PackageCollectionsModel/PackageCollectionModel+v1.swift
+++ b/Sources/PackageCollectionsModel/PackageCollectionModel+v1.swift
@@ -17,7 +17,7 @@ extension PackageCollectionModel {
 }
 
 extension PackageCollectionModel.V1 {
-    public struct Collection: Equatable, Codable {
+    public struct Collection: Equatable, Codable, Sendable {
         /// The name of the package collection, for display purposes only.
         public let name: String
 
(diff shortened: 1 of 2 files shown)


finestructure:
Thanks for taking a look, Max!

Since I can't change the source branch in this PR here, I've cherry-picked the change to a new one against `main` that I've opened here: https://github.com/swiftlang/swift-package-manager/pull/8635

PR #8638, opened 2025-05-08
Support Swift SDKs w/ many metadata files in same directory

MaxDesiatov:
### Motivation:

Swift SDK artifact bundle for WASI now supports both embedded and non-embedded modes. Majority of their content is overlapping, thus to avoid duplication they live in the same source tree.

Currently, SwiftPM requires that Swift SDKs are located in separate directories, with `swift-sdk.json` metadata at the root of each directory.

### Modification:

SwiftPM can now parse Swift SDK metadata that points directly to metadata JSON files instead of their root directories.

### Result:

We can avoid changing file system layout in the artifact bundles for WASI and just place sibling `nonembedded.json` and `embedded.json` Swift SDK metadata files in the same directory in their existing layout.

The change is transparent to Swift SDK users and is an optional and incremental addition for Swift SDK authors.

As Swift SDKs are required to exactly match the version of the host toolchain, there are no backward compatibility concerns with this change.

### DIFF
--- a/Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift
+++ b/Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift
@@ -370,9 +370,13 @@ public final class SwiftSDKBundleStore {
             var variants = [SwiftSDKBundle.Variant]()
 
             for variantMetadata in artifactMetadata.variants {
-                let variantConfigurationPath = bundlePath
+                var variantConfigurationPath = bundlePath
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

kateinoigakukun (approved):
Seems reasonable to me

PR #8639, opened 2025-05-08
Add `.staticLibrary` support for binary targets

FranzBusch:
Implements https://github.com/swiftlang/swift-evolution/blob/main/proposals/0482-swiftpm-static-library-binary-target-non-apple-platforms.md


FranzBusch:
@swift-ci test

FranzBusch:
@swift-ci test

FranzBusch:
@swift-ci test

FranzBusch:
@swift-ci test

Sources/Workspace/Workspace+State.swift
```diff
@@ -425,17 +425,17 @@ extension WorkspaceStateStorage {
                 }
             }
 
-            enum Kind: String, Codable {
+            enum Kind: Codable {
                 case xcframework
-                case artifactsArchive
+                case artifactsArchive([String])
```
dschaefer2, line 430: This will require a version update to v8 for the workspace state.

FranzBusch: Good catch. I just pushed a change that should allow us to do this without having to version the JSON format and will in the worst case require a clean build. Do you think that's okay?

FranzBusch:
@swift-ci test

Fixtures/BinaryLibraries/Static/Package1/Simple.artifactbundle/info.json
```diff
@@ -0,0 +1,35 @@
+{
```
bkhouri, line 1: suggestion: This file does not contain support for Windows, and the associated test does not skip this test on Windows.

I suspect we will either (1) need to add Support for windows, or (2) skip this on windows

If we need to support this feature for Windows and we decide to skip the test on Windows, please create a GitHub issue and reference the issue in the comment with skipping the test.  We can use `XCTSkipOnWindows(...)` in SwiftPM's `_InternalTestSupport` module for ths.

Tests/FunctionalTests/StaticBinaryLibrary.swift
```diff
@@ -0,0 +1,34 @@
...
+import TSCBasic
+import Testing
+import _InternalTestSupport
+
+@Suite
+struct StaticBinaryLibraryTests {
+    @Test
+    func staticLibrary() async throws {
+        try await fixture(name: "BinaryLibraries") { fixturePath in
+            let (stdout, stderr) = try await executeSwiftRun(
```
bkhouri, line 23: issue: Thanks for writing this test using SwiftTesting, however, there is currently an issue in `_InternalTestHelpers` that will cause the `executeSwift*(...)` functions to fail.  We should convert this to XCTestCase until the issue is resolved.

FranzBusch:
@swift-ci test

Sources/Workspace/Workspace+State.swift
FranzBusch: @dschaefer2 addressed this break by introducing a forwards compatibility new case and mapped the old one to it.

FranzBusch:
@swift-ci test windows

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -68,7 +68,7 @@ func withSession(
         _ diagnostics: [SwiftBuild.SwiftBuildMessage.DiagnosticInfo]
     ) async throws -> Void
 ) async throws {
-    switch await service.createSession(name: name, resourceSearchPaths: packageManagerResourcesDirectory.map { [$0.pathString] } ?? [], cachePath: nil, inferiorProductsPath: nil, environment: nil) {
+    switch await service.createSession(name: name, cachePath: nil, inferiorProductsPath: nil, environment: nil) {
```
owenv, line 71: This shouldn't be deleted

FranzBusch:
@swift-ci test

FranzBusch:
@swift-ci test windows

dschaefer2 (approved):
Thanks!

PR #8642, opened 2025-05-08
CLI command docs

heckj:
Adds documentation into the stubs for current Swift Package commands and subcommands.

This additionally marks a number of commands that are `hidden` as also `private` to prevent them from showing up in future dump commands.

The process was somewhat automated, but quite manual - as the current iteration of the argument parser plugin has issues with dumping docs for executables that are expected to be run under different names (such as `swift build`, `swift run`, etc)  - which Package Manager hosts, as well as dumps hidden commands, but not private commands.

resolves https://github.com/swiftlang/swift-package-manager/issues/8643


heckj:
@swift-ci please test

bripeticca (approved):

PR #8649, opened 2025-05-09
Test: Skip test on Amazon Linux 2

bkhouri:
This test caused a stack trace on Amazon Linux 2. Skip it for now.

### DIFF
--- a/Tests/PackageSigningTests/FilePackageSigningEntityStorageTests.swift
+++ b/Tests/PackageSigningTests/FilePackageSigningEntityStorageTests.swift
@@ -20,6 +20,7 @@ import Testing
 
 import struct TSCUtility.Version
 
+
 struct FilePackageSigningEntityStorageTests {
     @Test
     func happyCase() async throws {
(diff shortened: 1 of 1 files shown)


bkhouri:
@swift-ci test

Tests/PackageSigningTests/FilePackageSigningEntityStorageTests.swift
```diff
@@ -326,7 +327,12 @@ struct FilePackageSigningEntityStorageTests {
         #expect(packageSigners.signers[davinci]?.origins == [.registry(URL("http://foo.com"))])
     }
 
-    @Test
+    @Test(
```
jakepetroules, line 330: Could you refactor this into its own trait so it's reusable?

```swift
extension Trait where Self == Testing.ConditionTrait {
    package static var skipAmazonLinux2: Self {
        disabled("Test fails on Amazon Linux 2", {
            ProcessInfo.hostOperatingSystem == .linux && (FileManager.default.contents(atPath: "/etc/system-release").map({ String(decoding: $0, as: UTF8.self) == "Amazon Linux release 2 (Karoo)\n" }) ?? false)
        })
    }
}
```

scriptforfivem:
Very clean code with good test coverage. This is a great example of best practices.

Tests/PackageSigningTests/FilePackageSigningEntityStorageTests.swift
bkhouri: I definitely thought of this, but that's not currently possible as adding an `import Testing` in the `_InternalTestSupport` target will cause other pipelines to start failing since, in the `build-script`, swift PM will build itself using Cmake, and then will build the projects using the artifacts that was produced with the Cmake.

bkhouri: This should be possible once https://github.com/swiftlang/swift/pull/81401 is merged, hopefully.

bkhouri:
Closing PR as it appears to be stale.

PR #8668, opened 2025-05-14
Use WasmKit by default for Wasm triples in toolsets

MaxDesiatov:
As WasmKit is included in recent development snapshots of the swift.org toolchain, we should select it as a default debugger and test runner when a Wasm triple is selected. User toolsets can still override it if needed through toolset merging algorithm described in SE-0387.

rdar://150382758

### DIFF
--- a/Sources/CoreCommands/SwiftCommandState.swift
+++ b/Sources/CoreCommands/SwiftCommandState.swift
@@ -925,12 +925,6 @@ public final class SwiftCommandState {
     private lazy var _targetToolchain: Result<UserToolchain, Swift.Error> = {
         let swiftSDK: SwiftSDK
         let hostSwiftSDK: SwiftSDK
-        let store = SwiftSDKBundleStore(
-            swiftSDKsDirectory: self.sharedSwiftSDKsDirectory,
-            fileSystem: self.fileSystem,
-            observabilityScope: self.observabilityScope,
-            outputHandler: { print($0.description) }
-        )
         do {
             let hostToolchain = try _hostToolchain.get()
             hostSwiftSDK = hostToolchain.swiftSDK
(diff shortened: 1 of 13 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:

kateinoigakukun:
I think it would be nice if we can generalize it by introducing a way to express a toolchain binary path in toolset.json like `{ "testRunner": "path": { "type": "toolchain-lookup", "name": "wasmkit" } }`, but let's leave it as a future work.

kateinoigakukun (approved):

dschaefer2:
> I think it would be nice if we can generalize it by introducing a way to express a toolchain binary path in toolset.json like `{ "testRunner": "path": { "type": "toolchain-lookup", "name": "wasmkit" } }`, but let's leave it as a future work.

Agreed. We need to stop hardcoding so much stuff. We need to allow for someone to come along and change these things without having to put up a PR.

MaxDesiatov:
> > I think it would be nice if we can generalize it by introducing a way to express a toolchain binary path in toolset.json like `{ "testRunner": "path": { "type": "toolchain-lookup", "name": "wasmkit" } }`, but let's leave it as a future work.
> 
> Agreed. We need to stop hardcoding so much stuff. We need to allow for someone to come along and change these things without having to put up a PR.

As I've mentioned in the PR description, if you need to override things you can just apply your own toolset on top of the default toolset, which is the mechanism I've explicitly described in the Swift SDKs proposal foreseeing this use case.

What's hardcoded here is the toolchain default. Since WasmKit ships with the toolchain, it's toolchain's job to know where it is. If it shipped within a Swift SDK instead, then we would specify it in the Swift SDK. Since Swift SDK has no control over the toolchain, it's natural for the toolchain stuff to be hardcoded in the toolchain itself. It's no different from Clang hardcoding `wasm-ld`, knowing that `wasm-ld` ships together with Clang.

MaxDesiatov:
https://github.com/swiftlang/sourcekit-lsp/pull/2157
@swift-ci test

jakepetroules (approved):
Nice, this will be a great improvement to the out-of-the-box experience for using Wasm.

PR #8670, opened 2025-05-14
Include search path for Swift Testing's macro plugin from toolchain if present

stmontgomery:
This modifies the logic for determining compiler flags for locating Swift Testing content such that it will always include a search path for Swift Testing's macro plugin at its newer, testing-specific location in the toolchain, if that directory exists.

### Motivation:

Swift Testing installs a macro plugin library named `TestingMacros` into the official Swift toolchains for several platforms. By convention, most Swift macro plugins are installed into the toolchain's `usr/lib/swift/host/plugins` directory. However, testing libraries are somewhat special in the sense that they should generally only be available for use in "non-product" targets, i.e. targets which are not intended for distribution and only intended for use in qualifying the main product targets which _will_ be distributed.

For that reason, in Darwin toolchains, Swift Testing's plugin is installed into a different location than for other platforms: `usr/lib/swift/host/plugins/testing` (note the final `testing/` directory) — see the [CMake rules](https://github.com/swiftlang/swift-testing/blob/72afbb418542654781a6b7853479c7e70a862b6f/Sources/TestingMacros/CMakeLists.txt#L67-L75) where that is controlled. Over time, we'd like to move the macro plugin on other platforms to that location for consistency, because the current _inconsistent_ install path has caused friction for users if they attempt to form search paths to the plugin on their own — see https://github.com/swiftlang/swift-testing/issues/1039. So this PR paves the way for Swift Testing to begin to relocate its plugin by ensuring that _if_ this distinct plugin directory exists in a toolchain, it will be preferred.

### Modifications:

- Modify logic in `UserToolchain.swift` to begin passing `-plugin-path` whenever the relevant, testing-specific plugin path exists in a toolchain.
- Add a new test which validates this, by simulating the scenario of _not_ having a custom toolchain and validating the expected flags are passed. (This most closely matches the current scenario today of using the toolchain included in Xcode, and in that toolchain the plugin is already installed into the `testing/`-suffixed path.)

rdar://151319768

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -636,24 +636,26 @@ public final class UserToolchain: Toolchain {
         )
 
         if triple.isMacOSX, let swiftTestingPath {
-            // swift-testing in CommandLineTools, needs extra frameworks search path
+            // Swift Testing is a framework (e.g. from CommandLineTools) so use -F.
             if swiftTestingPath.extension == "framework" {
                 swiftCompilerFlags += ["-F", swiftTestingPath.pathString]
-            }
 
-            // Otherwise we must have a custom toolchain, add overrides to find its swift-testing ahead of any in the
-            // SDK. We expect the library to be in `lib/swift/macosx/testing` and the plugin in
-            // `lib/swift/host/plugins/testing`
-            if let pluginsPath = try? AbsolutePath(
-                validating: "../../host/plugins/testing",
-                relativeTo: swiftTestingPath
-            ) {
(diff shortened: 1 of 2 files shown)


stmontgomery:
@swift-ci please test

dschaefer2 (approved):

jakepetroules (approved):

jakepetroules:
Does SwiftBuild need any similar change here?

stmontgomery:
> Does SwiftBuild need any similar change here?

Nope, SwiftBuild does the correct thing already! See the [relevant code](https://github.com/swiftlang/swift-build/blob/1d66c4fdc1b1caac33c015236ee26e03ef054402/Sources/SWBCore/Settings/Settings.swift#L4297).

PR #8693, opened 2025-05-19
Detect embedded swift in flags

rauhul:
SwiftPM currently only detects Embedded Swift set via a swift settings and uses this to enable/disable a couple features. This is nice, but doesn't handle the case where Embedded is enabled via a user flag, typically set in a toolset.json. This commit updates the logic to also search for "-enable-experimental-feature Embedded" in the various place user flags can be placed.

### DIFF
--- a/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
@@ -1034,10 +1034,35 @@ public final class SwiftModuleBuildDescription {
         return arguments
     }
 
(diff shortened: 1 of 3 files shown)


kubamracek:
IIUC, this addresses https://github.com/swiftlang/swift-package-manager/issues/8688

MaxDesiatov:
@swift-ci test

kubamracek:
@swift-ci test

dschaefer2 (approved):
We'll add tests later once we get swift build support for the same thing.

Issue #7257, opened 2024-01-16
Failed to generate a static or dynamic Swift library called by C code


oliviermartin:
### Description

I am trying to generate a Swift library that can be called from a C binary.

### Expected behavior

1) Generate a static or dynamic library based on Swift code
2) Include the C header linked to this swift library in my C source code
3) Link the Swift based library with my C object
4) Demonstrate my C binary can call my Swift function

### Actual behavior

1) Generate a static or dynamic library based on Swift code [OK]
2) Include the C header linked to this swift library in my C source code [FAILED]
      The generated C header cannot be include because of error `error: module 'MySwiftLib' requires feature 'objc'`

### Steps to reproduce

1. Create sample project:

* `Package.swift`

```
// swift-tools-version: 5.9
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "TestApp",
    platforms: [.macOS(.v14), .iOS(.v13), .watchOS(.v6)],
    products: [
        .executable(name: "MyApp", targets: ["MyApp"]),
        .library(name: "MySwiftLib", type: .static, targets: ["MySwiftLib"])
    ],
    targets: [
        .executableTarget(name: "MyApp",
            cSettings: [
                .headerSearchPath("../../.build/x86_64-apple-macosx/debug/MySwiftLib.build")
            ]/*,
            linkerSettings: [
                .unsafeFlags(["-Xlinker", "-rpath", "-Xlinker", "../../.build/x86_64-apple-macosx/debug/"]),
                .linkedLibrary("MySwiftLib")
            ]*/
        ),
        .target(name: "MySwiftLib",
            swiftSettings: [
                .interoperabilityMode(.C)
            ]
        )
    ]
)
```

* `Sources/MySwiftLib/MySwiftLib.swift`

```
@_cdecl("mySwiftFunc")
public func mySwiftFunc() {
    print("Hello from Swift")
}
```

* `Sources/MyApp/main.c`

```
#include <stdio.h>

#include "MySwiftLib-Swift.h"

void main() {
    puts("Hello World from C!");
    mySwiftFunc();
}
```

2. Build the static or dynamic (replace `.static` by `.dynamic` in `Package.swift`) library  with: `swift build --product MySwiftLib`
[...]

PR #8695, opened 2025-05-19
Generate the compatibility header for libraries on all platforms

xymus:
Enable generating the compatibility header for library targets on all platforms by lifting the restriction on Darwin.

As part of the work to formalize `@cdecl` we've been adding content to the compatibility header that is usable by C compilers without the need for Objective-C or modules. Lifting this restriction will allow Linux and Windows users to generate and import the compatibility header to call `@cdecl` functions in their package.

### DIFF
--- a/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
@@ -801,7 +801,7 @@ public final class SwiftModuleBuildDescription {
 
     /// Returns true if ObjC compatibility header should be emitted.
     private var shouldEmitObjCCompatibilityHeader: Bool {
-        self.buildParameters.triple.isDarwin() && self.target.type == .library
+        self.target.type == .library
     }
 
     func writeOutputFileMap(to path: AbsolutePath) throws {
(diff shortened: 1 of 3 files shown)


xymus:
@swift-ci Please test

nkcsgexi:
Woohoo!

xymus:
@swift-ci Please test

xymus:
@swift-ci test windows

jakepetroules:
Took a quick look at Swift Build and I think we don't need any changes there; I see in my Android builds that the -Swift.h file is already being generated.

jakepetroules (approved):

DougGregor (commented):
This looks reasonable to me. I recommend doing a quick search of open GitHub issues that mention the generated header, because this probably fixes some.

xymus:
@swift-ci Please test windows

xymus:
@swift-ci Please test windows

xymus:
@swift-ci Please test windows

xymus:
@swift-ci Please test windows

DougGregor (approved):
Thank you!

xymus:
I found two related issues. #7319 should be fixed by this PR as they were only asking for the compatibility header to be generated on non-Darwin platforms. However there are still hurdles to import the compatibility header from a C/C++ source file and actually use it. I'm aware of two restrictions, both mentioned in #7257: the module map for the compatibility header is printed with `requires objc` and we don't pass a header search path to C targets that would allow non-modular includes. I'll see to lift those restrictions in my next PR.

PR #8700, opened 2025-05-20
[Commands] Migrate: Avoid injecting feature settings into every Swift…

xedin:
… module

### Motivation:

Previously the feature flags in `:migrate` mode were injected globally into "destination" parameter and affected every module in the graph, this is technically harmless but creates a lot of noise and precludes some of the future optimizations i.e. avoid injecting `@concurrent` into some closure literals when context is already `@concurrent`.

Instead of using build parameters, let's make the `Module.buildSettings` mutable in SwiftPM package and inject feature settings directly into `buildSettings` of the user-requested or all root package modules instead.

### Modifications:

- Make `Module.buildSettings` to be a `package(set) var` instead of a `let` to allow for mutation my migration command.
- Update `Migrate` command to inject the requested upcoming/experimental features in `:migrate` mode only into request targets or all non-plugin modules of root packages.

### Result:

No more global build flags and extraneous warnings from dependency modules that shouldn't have migration enabled.

### DIFF
--- a/Sources/Commands/PackageCommands/Migrate.swift
+++ b/Sources/Commands/PackageCommands/Migrate.swift
@@ -87,6 +87,7 @@ extension SwiftPackageCommand {
 
             let buildSystem = try await createBuildSystem(
                 swiftCommandState,
+                targets: self.options.targets,
                 features: features
             )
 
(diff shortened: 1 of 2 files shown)


xedin:
@swift-ci please test

xedin:

xedin:
@swift-ci please test self hosted Windows platform

dschaefer2 (approved):

jakepetroules (approved):

PR #8703, opened 2025-05-20
Simplify Swift SDK installation/selection with aliases

MaxDesiatov:
### Motivation:

As Swift SDKs require host toolchain version to exactly match the version used to build a Swift SDK that's used, we need to automate Swift SDK installation and selection process. Proposed solution is for the users to rely on predefined aliases, that SwiftPM uses as inputs to compute the exact Swift SDK installation URL and selection ID. New CLI options are added, which in the future could be unified with existing CLI arguments and options for Swift SDKs.

### Modifications:

Implemented for installation of Swift SDKs from swift.org:
```
swift sdk install --experimental-alias static-linux
```
For `swift build`, `swift run`, and `swift test`:
```
swift build --experimental-swift-sdk-alias static-linux
```

This assumes `version.json` is available in `usr/lib/swift` directory of the host toolchain, looking like this
```
{
  "branch": "development",
  "tag": "DEVELOPMENT-SNAPSHOT-2025-04-24-a",
  "platform": "ubuntu2004",
  "architecture": "aarch64"
}
```

Information from this file is used to compute Swift SDK URLs and IDs.

### Result:

Users no longer need to find the exactly matching Swift SDK for the currently selected toolchain version.

### DIFF
--- a/Sources/SwiftSDKCommand/SwiftSDKInstall.swift
+++ b/Sources/SwiftSDKCommand/SwiftSDKInstall.swift
@@ -81,7 +81,7 @@ struct SwiftSDKInstall: SwiftSDKSubcommand {
             try SwiftToolchainVersion(
                 toolchain: hostToolchain,
                 fileSystem: self.fileSystem
-            ).generateURL(aliasString: experimentalAlias)
+            ).urlForSwiftSDK(aliasString: experimentalAlias)
         } else if let bundlePathOrURL {
             bundlePathOrURL
         } else {
--- a/Sources/SwiftSDKCommand/SwiftSDKInstall.swift
+++ b/Sources/SwiftSDKCommand/SwiftSDKInstall.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift open source project
 //
-// Copyright (c) 2023 Apple Inc. and the Swift project authors
+// Copyright (c) 2023-2025 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See http://swift.org/LICENSE.txt for license information
@@ -20,7 +20,7 @@ import PackageModel
 import var TSCBasic.stdoutStream
 import class Workspace.Workspace
 
-struct InstallSwiftSDK: SwiftSDKSubcommand {
+struct SwiftSDKInstall: SwiftSDKSubcommand {
     static let configuration = CommandConfiguration(
         commandName: "install",
         abstract: """
@@ -76,6 +76,17 @@ struct InstallSwiftSDK: SwiftSDKSubcommand {
                 )
                 .throttled(interval: .milliseconds(300))
         )
+
+        let bundlePathOrURL = if let experimentalAlias {
+            try SwiftToolchainVersion(
+                toolchain: hostToolchain,
+                fileSystem: self.fileSystem
+            ).generateURL(aliasString: experimentalAlias)
+        } else if let bundlePathOrURL {
+            bundlePathOrURL
+        } else {
+            throw InternalError("foo")
+        }
 
         try await store.install(
             bundlePathOrURL: bundlePathOrURL,
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -483,6 +483,12 @@ public struct BuildOptions: ParsableArguments {
     )
     public var swiftSDKSelector: String?
 
+    @Option(
+        name: .customLong("experimental-swift-sdk-alias"),
+        help: "Alias for a compatible Swift SDK to build with."
+    )
+    public var swiftSDKAlias: String?
+
     /// Which compile-time sanitizers should be enabled.
     @Option(
         name: .customLong("sanitize"),
--- a/Sources/CoreCommands/SwiftCommandState.swift
+++ b/Sources/CoreCommands/SwiftCommandState.swift
@@ -945,6 +945,13 @@ public final class SwiftCommandState {
                 outputHandler: { print($0.description) }
             )
 
+            let swiftSDKSelector = if let swiftSDKAliasString = self.options.build.swiftSDKAlias {
+                try SwiftToolchainVersion(toolchain: hostToolchain, fileSystem: self.fileSystem)
+                    .idForSwiftSDK(aliasString: swiftSDKAliasString)
+            } else {
+                self.options.build.swiftSDKSelector ?? self.options.build.deprecatedSwiftSDKSelector
+            }
+
             swiftSDK = try SwiftSDK.deriveTargetSwiftSDK(
                 hostSwiftSDK: hostSwiftSDK,
                 hostTriple: hostToolchain.targetTriple,
@@ -953,7 +960,7 @@ public final class SwiftCommandState {
                 customCompileTriple: self.options.build.customCompileTriple,
                 customCompileToolchain: self.options.build.customCompileToolchain,
                 customCompileSDK: self.options.build.customCompileSDK,
-                swiftSDKSelector: self.options.build.swiftSDKSelector ?? self.options.build.deprecatedSwiftSDKSelector,
+                swiftSDKSelector: swiftSDKSelector,
                 architectures: self.options.build.architectures,
                 store: store,
                 observabilityScope: self.observabilityScope,
--- a/Sources/PackageModel/CMakeLists.txt
+++ b/Sources/PackageModel/CMakeLists.txt
@@ -53,9 +53,11 @@ add_library(PackageModel
   SupportedLanguageExtension.swift
   SwiftLanguageVersion.swift
   SwiftSDKs/SwiftSDK.swift
+  SwiftSDKs/SwiftSDKAlias.swift
   SwiftSDKs/SwiftSDKConfigurationStore.swift
   SwiftSDKs/SwiftSDKBundle.swift
   SwiftSDKs/SwiftSDKBundleStore.swift
+  SwiftToolchainVersion.swift
   Toolchain.swift
   ToolchainConfiguration.swift
   Toolchain+SupportedFeatures.swift
--- a/Sources/PackageModel/SwiftSDKs/SwiftSDKAlias.swift
+++ b/Sources/PackageModel/SwiftSDKs/SwiftSDKAlias.swift
@@ -16,7 +16,7 @@ package struct SwiftSDKAlias {
         self.kind = kind
     }
     
-    enum Kind: String {
+    enum Kind: String, CaseIterable {
         case staticLinux  = "static-linux"
         case wasi         = "wasi"
         case wasiEmbedded = "embedded-wasi"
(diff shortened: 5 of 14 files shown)


MaxDesiatov:
@swift-ci test

Sources/SwiftSDKCommand/SwiftSDKInstall.swift
```diff
@@ -73,6 +77,17 @@ struct InstallSwiftSDK: SwiftSDKSubcommand {
...
 
+        let bundlePathOrURL = if let experimentalAlias {
+            try SwiftToolchainVersion(
+                toolchain: hostToolchain,
+                fileSystem: self.fileSystem
+            ).urlForSwiftSDK(aliasString: experimentalAlias)
+        } else if let bundlePathOrURL {
+            bundlePathOrURL
+        } else {
+            throw InternalError("foo")
```
jakepetroules, line 88: This should be changed to a real message before merging.

jakepetroules (changes requested):
Is there any way this could be more data driven, so we don't have to change the toolchain every time there's a new Swift SDK? Or is this meant to be geared towards supporting just the Swift SDKs that are officially distributed by swift.org, and you need to do manual selection for others?

kateinoigakukun (changes requested):
Totally agree with @jakepetroules. I think we should explore more data-driven way for installation process by even changing the current distribution scheme. At least, SwiftPM should not know about the list of aliases.

MaxDesiatov:
>  Or is this meant to be geared towards supporting just the Swift SDKs that are officially distributed by swift.org, and you need to do manual selection for others?

Yes

MaxDesiatov:
> Totally agree with @jakepetroules. I think we should explore more data-driven way for installation process by even changing the current distribution scheme. At least, SwiftPM should not know about the list of aliases.

Can you elaborate? How would it get the list of aliases otherwise? Any server-based solutions are out of scope for this change.

al45tair:
> As Swift SDKs require host toolchain version to exactly match the version used to build a Swift SDK that's used

 I keep wondering whether we should really be fixing *that* problem, instead of trying to work around it (either with this PR or by giving people complicated instructions about matching them up). The fundamental issue here is that we aren't ABI stable, so we're relying on `swiftmodule` files, which are themselves not ABI stable and that's why the toolchain and SDK must match.

> Totally agree with @jakepetroules. I think we should explore more data-driven way for installation process by even changing the current distribution scheme. At least, SwiftPM should not know about the list of aliases.

I think I also agree with that sentiment. Could we perhaps have an option to specify the SDK by pointing at a JSON file that provides the required information to locate the _actual_ SDK according to the current toolchain version? So, e.g. you'd write something like

```
$ swift sdk install swift.org/sdks/static-linux
```

or (to give a purely imaginary third-party example)

```
$ swift sdk install microsoft.com/swift-sdks/xbox
```

Then if the JSON data at the given URL doesn't contain a version matching the toolchain, we'd respond with something like

```
Unable to find a matching SDK for this toolchain (<toolchain version>). Available versions are:

<list of available SDK versions>
```

MaxDesiatov:
> I keep wondering whether we should really be fixing _that_ problem, instead of trying to work around it (either with this PR or by giving people complicated instructions about matching them up). The fundamental issue here is that we aren't ABI stable, so we're relying on `swiftmodule` files, which are themselves not ABI stable and that's why the toolchain and SDK must match.

Making Swift ABI-stable on non-Darwin platforms is out of scope of this PR 🙂 

> I think I also agree with that sentiment. Could we perhaps have an option to specify the SDK by pointing at a JSON file that provides the required information to locate the _actual_ SDK according to the current toolchain version?

That's the problem, what would generate that JSON? As server-side solutions are out of scope (due to time constraints), we'd have to hardcode that JSON somewhere, either in SwiftPM or in the toolchain. Since SwiftPM is responsible for Swift SDK installation, hardcoding it in SwiftPM makes most since, and as we're already hardcoding it, JSON brings no benefits when compared to just writing this logic in plain Swift.

MaxDesiatov:
> that would try to fetch the JSON file at e.g. `https://swift.org/sdks/static-linux/sdk-index.json`.

As mentioned above, any server-side solutions are out of scope of this PR. We're looking for something very targeted and focused that can resolve the installation issues for user's in the short term. When a solution for hosting a Swift SDKs index on swift.org is available, I'd be very happy to implement it in a separate PR.

kateinoigakukun:
> As mentioned above, any server-side solutions are out of scope of this PR. 

What do you mean by "server-side solutions"? I think we can implement something what @al45tair suggested by putting static files like what we do in https://github.com/swiftwasm/swift-sdk-index or https://github.com/swiftlang/swift-org-website/blob/main/api/v1/install/dev/6.0/static-sdk.json, and it shouldn't require "server" process other than just a static HTTP server we already have on swift.org.

I'm ok with hardcoding the SDK source URL locations and limiting the alias with known ones for now for time constraints, but I think we should at least define what it will be eventually to allow third-parties to provide their SDK index and we need to make sure we can implement it without changing CLI interface.

MaxDesiatov:
> What do you mean by "server-side solutions"? I think we can implement something what @al45tair suggested by putting static files like what we do in https://github.com/swiftwasm/swift-sdk-index or https://github.com/swiftlang/swift-org-website/blob/main/api/v1/install/dev/6.0/static-sdk.json, and it shouldn't require "server" process other than just a static HTTP server we already have on swift.org.

I mean deploying payloads on any server, including a static server. This suggestion glosses over complexities of deployment on a static server. What's very easy to set on an arbitrary GitHub organization can't be universally replicated in all scenarios. There's a lot of mechanics involved in keeping these payloads up-to-date, tying them to existing CI jobs etc. That's definitely out of scope of this PR. When these deployment and CI issues are resolved, I'd very happy to come back and overhaul everything to be dynamic, generalized, and clean, avoiding any hardcoding etc. But before that we need something that solves immediate problems that users already have been having for quite a long time.

Sources/SwiftSDKCommand/SwiftSDKInstall.swift
MaxDesiatov: Sorry I missed that, I've added a separate error type for this. I hope this addresses all of the remaining concerns.

MaxDesiatov:
@swift-ci test

marcprux:
@finagolfin Can you add Android alias support as a Todo on the https://github.com/orgs/swiftlang/projects/17 project board? It'd be great if we also supported the command like:

```
swift sdk install --experimental-alias android
```

finagolfin:
I agree that we should alias Android also, but I'm skeptical of adding [a bunch more SDK names to this hard-coded list in SwiftPM](https://github.com/swiftlang/swift-package-manager/pull/8703/files#diff-5e0199a1b8ce365cda8052c6aca3517d18ffae881172d860047784938b601c5cR22). I suppose this current list is fine since these three are already officially supported bundles, but as Android and more official and third-party SDK bundles are added, it's not going to scale to keep hard-coding aliases into SwiftPM.

I'd prefer to have SDK bundle registries, as @dschaefer2 has talked about in the past, both one for official swift.org SDK bundles and the possibility of third-party SDK registries, that will list aliases too. Given the time constraints Max mentions, not asking for that now, but I agree with the others above that such server-side solutions should be the medium-term solution.

All that said, if you think we should add an Android alias here in SwiftPM once the Android SDK is official, feel free to open an issue for that and I can then add it to the roadmap.

dschaefer2:
> I agree that we should alias Android also, but I'm skeptical of adding a bunch more SDK names to this hard-coded list in SwiftPM

Working with the embedded Swift community we're envisioning many SDKs that people can choose from. Requiring a SwiftPM pull request for each one doesn't make much sense and will only slow down the growth of the ecosystem and prevent developers who might want to have private SDKs.

I'm much more of the opinion that SDKs should be elements of Packages. To use an SDK you just create a package dependency on the package defining it and add constraints to your targets, expecially executable targets, to build for those SDKs.

I'm not sure of the details yet of how that will all work but as I get started, I'll be reaching out to all SDK providers to see if we can come up with a solution that works for everyone.

MaxDesiatov:
As tags are specified in `swiftCompilerTag` output of `swiftc -print-target-info` in latest snapshots, this command can be used in the meantime for selecting a Swift SDK with ID based on tags:

```bash
swift build --swift-sdk "$(swiftc -print-target-info | jq -r '.swiftCompilerTag')_wasm-embedded"
```

ephemer:
What I’m unsure about: I think this introduces a complex dependency graph where none was present before.

I mean: doesn’t this require the SDK to be built and deployed before SwiftPM can even begin to be built? We‘d need to manually update this list of SDKs to bundle SwiftPM with Swift - and Swift (maybe even SwiftPM) had to be already built to build the SDKs?

This doesn’t seem pretty to me, but I’m probably missing something. If that situation does truly exist, to me it seems similarly complex to solving the issues surrounding a web-based registry.

While I have no interest to (and am in no position to) block this work going forward, personally I’m not convinced by the mentality of „let’s do this ‚for now‘“. From experience, this tends to become „the“ solution by default. And in this case I‘m not yet certain this is the best solution.

———

It would be nice if the SDK itself had an indication of the compiler version it was built with. Then SwiftPM could match SDKs to compiler versions and filter the list (at least visually and functionally marking others as unavailable)

marcprux:
> Working with the embedded Swift community we're envisioning many SDKs that people can choose from. Requiring a SwiftPM pull request for each one doesn't make much sense and will only slow down the growth of the ecosystem and prevent developers who might want to have private SDKs.

Could we have the command check the [API endpoints](https://github.com/swiftlang/swift-org-website/tree/main/api/v1/install/dev/main) for the SDK? E.g., `swift build --experimental-swift-sdk-alias android`  would fetch https://www.swift.org/api/v1/install/dev/main/android-sdk.json and grab the right SDK for the current toolchain. That way, new SDKs would "just work" as they are released.

Private or beta SDKs could be supported with something like:

```
swift build --swift-sdk-alias android --swift-sdk-endpoint-base https://www.swift.org/api/v1/install/dev/main/
```

> I'm much more of the opinion that SDKs should be elements of Packages. To use an SDK you just create a package dependency on the package defining it and add constraints to your targets, expecially executable targets, to build for those SDKs.

This confuses me. Do you mean the Package.swift for the project being built? Not all packages know which SDKs they want to be able to support. Perhaps in the case of an executable target you might want to call out that your program is designed for some_exotic_sdk, but for packages that want to support many different platforms, having to individually enumerate every SDK they might be able to build with seems to shift a lot of burden on the package developer. Perhaps I am misunderstanding your suggestion?

(BTW, I know the comment I'm responding to is 5 months old, but I never got around to responding the first time around and only noticed it after the most recent comment)

MaxDesiatov:
I would like to remind that this PR is in a draft state, hasn't had any new commits for ~6 months, and in no way reflects any proposed design approach even if such would be pitched or accepted at all.

I will post an update and remove the draft marker when it's ready, otherwise providing feedback on a PR that's not marked as ready for review feels premature to me.

dschaefer2:
Great feedback though :). I look forward to what we can do here.

PR #8707, opened 2025-05-21
Dynamically determine the swift compiler version.

dschaefer2:
Prebuilts for macros and the automated downloading of SwiftSDKs need to know the version of the compiler so we can fetch compatible binaries. The swiftc compiler has a --print-target-info options which dumps out a JSON structure that contains the compilerVersion. We already use info in this structure to determine the hostTriple for the UserToolchain.

This adds the swiftCompilerVersion to UserToolchain that uses a couple of regex's to pull out the Swift compiler version. This is then used by the prebuilts feature instead of our current hardcodeing of the swift toolchain version.

This also turns the prebuilts feature on by default which was supposed to be done in the last update.

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -13,6 +13,7 @@
 import Basics
 import Foundation
 import TSCUtility
+import enum TSCBasic.JSON
 
 import class Basics.AsyncProcess
 
@@ -73,6 +74,9 @@ public final class UserToolchain: Toolchain {
     public var triple: Basics.Triple { targetTriple }
 
     public let targetTriple: Basics.Triple
+
+    // A version string that can be used to identify the swift compiler version
+    public let swiftCompilerVersion: String?
 
     /// The list of CPU architectures to build for.
     public let architectures: [String]?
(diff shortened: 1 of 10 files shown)


Sources/PackageModel/UserToolchain.swift
```diff
@@ -160,6 +164,73 @@ public final class UserToolchain: Toolchain {
...
+    private static func computeSwiftCompilerVersion(targetInfo: JSON) -> String? {
+        // Get the compiler version from the target info
+        let compilerVersion: String
+        do {
+            compilerVersion = try targetInfo.get("compilerVersion")
+        } catch {
+            return nil
+        }
+
+        // Extract the swift version using regex from the description if available
```
jakepetroules, line 216: I'm pretty sure what you have captures the necessary variations, but here is what SwiftBuild is doing for the record: https://github.com/swiftlang/swift-build/blob/8e71bde5ac572174bc7e65f505abf3559bf05455/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift#L3607

jakepetroules (approved):

Sources/PackageModel/UserToolchain.swift
bnbarham: I'm about to put a patch up for swift-build. It's *way* too specific right now 😅

bnbarham: https://github.com/swiftlang/swift-build/pull/519

dschaefer2: All the mocks failed. Another commit incoming.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

PR #8708, opened 2025-05-21
Dynamically determine the swift compiler version.

dschaefer2:
Prebuilts for macros and the automated downloading of SwiftSDKs need to know the version of the compiler so we can fetch compatible binaries. The swiftc compiler has a --print-target-info options which dumps out a JSON structure that contains the compilerVersion. We already use info in this structure to determine the hostTriple for the UserToolchain.

This adds the swiftCompilerVersion to UserToolchain that uses a couple of regex's to pull out the Swift compiler version. This is then used by the prebuilts feature instead of our current hardcodeing of the swift toolchain version.

This also turns the prebuilts feature on by default which was supposed to be done in the last update.

Cherry pick of https://github.com/swiftlang/swift-package-manager/pull/8707

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -198,7 +198,7 @@ public struct CachingOptions: ParsableArguments {
     @Flag(name: .customLong("experimental-prebuilts"),
           inversion: .prefixedEnableDisable,
           help: "Whether to use prebuilt swift-syntax libraries for macros.")
-    public var usePrebuilts: Bool = false
+    public var usePrebuilts: Bool = true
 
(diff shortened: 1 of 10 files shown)


dschaefer2:
@swift-ci please test

jakepetroules (approved):

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test self-hosted windows

dschaefer2:
@swift-ci please test self-hosted windows

dschaefer2:
@swift-ci please test windows

PR #8719, opened 2025-05-23
[CODEOWNERS] Move code owners for specific directories/files under wi…

xedin:
…ldcard

According to GitHub Docs the order is important and the last matching pattern takes precedence.


xedin:
@swift-ci please test

PR #8721, opened 2025-05-24
Async git repository opening

plemarquand:
The `RepositoryProvider` protocol's `open` method is defined as synchronous. Convert this to `async` to avoid concurrency workarounds in clients that might be prone to deadlocks.

This involved porting `RepositoryManager.lookup` to structured concurrency. The method has some concurrency constraints that are preserved:

- There is a limit (`maxConcurrentOperations`) on how many simultaneous lookups can be performed concurrently.
- If a lookup is requested and one is in flight for the same `RepositorySpecifier`, the in flight request completes before the new one is started.

### DIFF
--- a/Sources/Basics/Concurrency/ConcurrencyHelpers.swift
+++ b/Sources/Basics/Concurrency/ConcurrencyHelpers.swift
@@ -235,7 +235,8 @@ public final class AsyncOperationQueue: @unchecked Sendable {
                 waitingTasks.remove(at: taskIndex)
             }
 
-            // This could be simplified if `waitingTasks` could be an `OrderedDictionary` from swift-collections.
+            // We cannot remove elements from `waitingTasks` while iterating over it, so we make
+            // a pass to collect operations and then apply them after the loop.
             func createTaskListOperations() -> (CollectionDifference<WorkTask>?, WaitingContinuation?) {
                 var changes: [CollectionDifference<WorkTask>.Change] = []
                 for (index, task) in waitingTasks.enumerated() {
--- a/Sources/Basics/Concurrency/ConcurrencyHelpers.swift
+++ b/Sources/Basics/Concurrency/ConcurrencyHelpers.swift
@@ -81,16 +81,37 @@ extension DispatchQueue {
 /// A queue for running async operations with a limit on the number of concurrent tasks.
 public final class AsyncOperationQueue: @unchecked Sendable {
 
-    // This implementation is adapted from the one in swift-build,
-    // modified to respect cancellation of the parent Task.
-    // https://github.com/swiftlang/swift-build/blob/054f2300ad83fd1633f1b50a06b82eea9e7c6901/Sources/SWBUtil/AsyncOperationQueue.swift#L13
+    // This implementation is identical to the AsyncOperationQueue in swift-build.
+    // Any modifications made here should also be made there.
+    // https://github.com/swiftlang/swift-build/blob/main/Sources/SWBUtil/AsyncOperationQueue.swift#L13
 
-    typealias ID = UUID
+    fileprivate typealias ID = UUID
+    fileprivate typealias WaitingContinuation = CheckedContinuation<Void, any Error>
 
     private let concurrentTasks: Int
     private var activeTasks: Int = 0
-    private var waitingTasks: [(ID, CheckedContinuation<Void, any Error>)] = []
+    private var waitingTasks: [WaitingTask] = []
     private let waitingTasksLock = NSLock()
+
+    fileprivate enum WaitingTask {
+        case creating(ID)
+        case waiting(ID, WaitingContinuation)
+        case cancelled(ID)
+
+        var id: ID {
+            switch self {
+            case .creating(let id), .waiting(let id, _), .cancelled(let id):
+                return id
+            }
+        }
+
+        var continuation: WaitingContinuation? {
+            guard case .waiting(_, let continuation) = self else {
+                return nil
+            }
+            return continuation
+        }
+    }
 
     /// Creates an `AsyncOperationQueue` with a specified number of concurrent tasks.
     /// - Parameter concurrentTasks: The maximum number of concurrent tasks that can be executed concurrently.
@@ -100,7 +121,7 @@ public final class AsyncOperationQueue: @unchecked Sendable {
 
     deinit {
         waitingTasksLock.withLock {
-            if !waitingTasks.isEmpty {
+            if !waitingTasks.filter({ $0.continuation != nil }).isEmpty {
                 preconditionFailure("Deallocated with waiting tasks")
             }
         }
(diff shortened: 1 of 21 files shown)


plemarquand:
@swift-ci please test

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +76,33 @@ extension DispatchQueue {
         }
     }
 }
+
+package actor ConcurrencyLimiter {
```
jakepetroules, line 80: Can you instead copy https://github.com/swiftlang/swift-build/blob/main/Sources/SWBUtil/AsyncOperationQueue.swift? We should avoid having slightly different implementations for the same problem; also that implementation is safer because it guards against unbalanced acquire/release by only providing a closure-based API.

plemarquand: Done.

plemarquand:
@swift-ci please test

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +76,55 @@ extension DispatchQueue {
...
+        _ operation: @Sendable () async throws -> sending ReturnValue
+    ) async throws -> ReturnValue {
+        await waitIfNeeded()
+        defer { signalCompletion() }
+        return try await operation()
+    }
+
+    private func waitIfNeeded() async {
+        if activeTasks >= concurrentTasks {
+            await withCheckedContinuation { continuation in
```
FranzBusch, line 115: We should add a `withTaskCancellationHandler` here and potentially a `withTaskPriorityEscalationHandler`

FranzBusch: No that's not enough since the task can be cancelled after the continuation has been appended. In general, to be structured concurrency and cancellation compliant whenever you create a continuation you need to wrap it into a `withTaskCancellationHandler`. In this case you probably want to remove that continuation from the `waitingTasks` and resume it with a `CancellationError`.

Since you are currently using an `actor` this won't work easily without an unstructured task. So I would recommend you to switch over to a `class` and protect the state with a `Mutex`.

plemarquand: Ah, unfortunately SwiftPM's supported `platforms` is set to `.macOS(.v13)` and `Mutex` is only available on >= .v15, so I'll have to use an NSLock as we do elsewhere in SwiftPM.

FranzBusch: Is there a reason why we can't bump to macOS 15.0?

jakepetroules: Yes. We cannot require macOS 15 at this time.

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +77,87 @@ extension DispatchQueue {
         }
     }
 }
+
+/// A queue for running async operations with a limit on the number of concurrent tasks.
+public final class AsyncOperationQueue: @unchecked Sendable {
+
+    // This implementation is adapted from the one in swift-build,
```
jakepetroules, line 84: Instead of adding this comment, can you please also open a PR to update the copy in SwiftBuild?

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +77,87 @@ extension DispatchQueue {
...
+                        continuation.resume(throwing: CancellationError())
+                    }
+                }
+            } onCancel: {
+                // If the parent task is cancelled then we need to manually handle resuming the
+                // continuation for the waiting task with a `CancellationError`.
+                self.waitingTasksLock.withLock {
+                    if let taskIndex = self.waitingTasks.firstIndex(where: { $0.0 == taskId }) {
+                        let task = self.waitingTasks.remove(at: taskIndex)
+                        task.1.resume(throwing: CancellationError())
```
jakepetroules, line 148: resuming continuations while a lock is held is dangerous and can lead to deadlocks. Note withTaskCancellationHandler's documentation:

> Cancellation handlers which acquire locks must take care to avoid deadlock. The cancellation handler may be invoked while holding internal locks associated with the task or other tasks. Other operations on the task, such as resuming a continuation, may acquire these same internal locks. Therefore, if a cancellation handler must acquire a lock, other code should not cancel tasks or resume continuations while holding that lock.

I think here you can return the continuation out of the lock and then resume it after releasing the lock.

jakepetroules (changes requested):

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +77,87 @@ extension DispatchQueue {
...
+
+    private func signalCompletion() {
+        let continuationToResume = waitingTasksLock.withLock {
+            activeTasks -= 1
+            return waitingTasks.popLast()?.1
+        }
+
+        continuationToResume?.resume()
+    }
+}
```
jakepetroules, line 163: nit: missing newline at EOF

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +77,87 @@ extension DispatchQueue {
...
+            } onCancel: {
+                // If the parent task is cancelled then we need to manually handle resuming the
+                // continuation for the waiting task with a `CancellationError`.
+                self.waitingTasksLock.withLock {
+                    if let taskIndex = self.waitingTasks.firstIndex(where: { $0.0 == taskId }) {
+                        let task = self.waitingTasks.remove(at: taskIndex)
+                        task.1.resume(throwing: CancellationError())
+                    }
+                }
+            }
```
FranzBusch, line 151: @jakepetroules raised a good point with `resuming` while a lock is held. In general, never do any outcalls while a lock is held. FWIW, we have run into exactly this deadlock before.

Now the first part of this code is also not 100% correct. There is subtle race that can happen in this code in this scenario:
1. Task is running not cancelled
2. You are checking `Task.isCancelled` returns `false`
3. Something is canceling your task
4. The task cancellation handler runs and acquires the lock. Tries to remove the continuation but it hasn't been stored yet so resumes nothing
5. The code after the `Task.isCancelled` runs and stores the continuation

In this case you now stored a continuation of a cancelled task and the task cancellation handler already ran. What you have to do instead is to model a tri-state for a waiter:
```
enum Waiter {
  case creating
  case waiting(Continuation)
  case cancelled
}
```

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test windows

plemarquand:
@FranzBusch @jakepetroules I've addressed comments and added some tests for `AsyncOperationQueue`. Could you take another look?

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +77,183 @@ extension DispatchQueue {
...
+
+    /// Creates an `AsyncOperationQueue` with a specified number of concurrent tasks.
+    /// - Parameter concurrentTasks: The maximum number of concurrent tasks that can be executed concurrently.
+    public init(concurrentTasks: Int) {
+        self.concurrentTasks = concurrentTasks
+    }
+
+    deinit {
+        waitingTasksLock.withLock {
+            if !waitingTasks.filter({ $0.continuation != nil }).isEmpty {
```
jakepetroules, line 124: Isn't it unexpected if there is _anything at all_ in waitingTasks, not just those with a non-nil continuation?

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci please test windows

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +77,184 @@ extension DispatchQueue {
...
+            activeTasks += 1
+            return shouldWait
+        }) else {
+            return // Less tasks are in flight than the limit.
+        }
+
+        let taskId = ID()
+        waitingTasksLock.withLock {
+            waitingTasks.append(.creating(taskId))
+        }
```
FranzBusch: We could move this into the above lock and return an optional `UUID`. This way we don't need to acquire the lock more than once for the initial setup

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +77,184 @@ extension DispatchQueue {
...
+
+    // This implementation is identical to the AsyncOperationQueue in swift-build.
+    // Any modifications made here should also be made there.
+    // https://github.com/swiftlang/swift-build/blob/main/Sources/SWBUtil/AsyncOperationQueue.swift#L13
+
+    fileprivate typealias ID = UUID
+    fileprivate typealias WaitingContinuation = CheckedContinuation<Void, any Error>
+
+    private let concurrentTasks: Int
+    private var activeTasks: Int = 0
```
FranzBusch: Do we need this counter at all or can we just use `waitingTasks.count`? Keeping the two in sync can be tricky and I had to triple check that the code does it correctly. That requires us to change the `WaitingTask` to be just `WorkTask` with a new case but IMO that would make this code a lot clearer.

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +77,184 @@ extension DispatchQueue {
...
+            try await withCheckedThrowingContinuation { (continuation: WaitingContinuation) -> Void in
+                let action: TaskAction? = waitingTasksLock.withLock {
+                    guard let index = waitingTasks.firstIndex(where: { $0.id == taskId }) else {
+                        // If the task was cancelled in onCancelled it will have been removed from the waiting tasks list.
+                        return .cancel(continuation)
+                    }
+
+                    // If the task was cancelled in between creating the task cancellation handler and aquiring the lock,
+                    // we should resume the continuation with a `CancellationError`.
+                    if case .cancelled = waitingTasks[index] {
```
FranzBusch: Instead of writing an `if case` I would prefer if we exhaustively switch over the returned task here similar to how you did it in the `onCancel` case. I always recommend doing that since it makes sure every case is properly handled.

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +77,184 @@ extension DispatchQueue {
...
+
+            while let lastTask = waitingTasks.first {
+                switch lastTask {
+                case .creating:
+                    // If the next task is in the process of being created, let the
+                    return Optional<WaitingContinuation>.none
+                case .waiting:
+                    activeTasks -= 1
+                    // Begin the next waiting task
+                    return waitingTasks.remove(at: 0).continuation
```
FranzBusch: Since we are operating under FIFO here it is worth using a `Deque` instead of an array since this `remove(at:)` is going to reallocate the entire array.

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +77,184 @@ extension DispatchQueue {
...
+        enum TaskAction {
+            case start(WaitingContinuation)
+            case cancel(WaitingContinuation)
+        }
+
+        try await withTaskCancellationHandler {
+            try await withCheckedThrowingContinuation { (continuation: WaitingContinuation) -> Void in
+                let action: TaskAction? = waitingTasksLock.withLock {
+                    guard let index = waitingTasks.firstIndex(where: { $0.id == taskId }) else {
+                        // If the task was cancelled in onCancelled it will have been removed from the waiting tasks list.
```
FranzBusch: This comment is not true right? We are always going to get an index by just looking at the code in `onCancel`. Either we find a `.cancelled` case or we have run before. There is no way for `onCancel` and this check to interleave in the way you describe here. Now, what can interleave is the `signalCompletion` that also appears to remove `cancelled()` tasks. Can we update this comment?

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
plemarquand: I was trying to keep in mind this will make its way back in to swift-build, which has no dependency on swift-collections and [looks to be trying to keep its dependencies minimal](https://github.com/swiftlang/swift-build/blob/main/Package.swift#L457). @jakepetroules is that accurate, or would swift-build accept a swift-collections dependency for this?

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
plemarquand: We can, but it that would mean the waitingTasks would no longer be able to be modelled as a queue since the actively running tasks would be contained within it.

I think an OrderedDictionary could work instead, but runs in to the swift-collections in swift-build question.

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
plemarquand: Thanks, I've tightend up this comment to reflect how this guard code would actually get called.

plemarquand:
@swift-ci please test

plemarquand:
@FranzBusch I've tried to simplify the code as suggested by removing the counter state. It did complicate the code in `signalCompletion` a bit because we can't start processing immediately from the head of the collection.

plemarquand:
@swift-ci please test

plemarquand:
@swift test macos

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci please test macOS

plemarquand:
@swift-ci test windows

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -76,3 +77,203 @@ extension DispatchQueue {
...
+        }
+    }
+
+    /// Executes an asynchronous operation, ensuring that the number of concurrent tasks
+    // does not exceed the specified limit.
+    /// - Parameter operation: The asynchronous operation to execute.
+    /// - Returns: The result of the operation.
+    /// - Throws: An error thrown by the operation, or a `CancellationError` if the operation is cancelled.
+    public func withOperation<ReturnValue>(
+        _ operation: @Sendable () async throws -> sending ReturnValue
```
FranzBusch: The operation doesn't have to be `Sendable` right?

plemarquand:
Thank you for your tireless reviews @FranzBusch!

jakepetroules (approved):

PR #8722, opened 2025-05-25
Non-darwin test discovery with Swift Build

owenv:
Still needs a fair amount of work and cleanup, I'll look at breaking off some parts that can land independently. Depends on the larger patch in https://github.com/swiftlang/swift-build/pull/499

### DIFF
--- a/Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
+++ b/Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
@@ -314,11 +314,26 @@ public struct BuildParameters: Encodable {
         case .library(.automatic), .plugin:
             fatalError()
         case .test:
-            let base = "\(product.name).xctest"
-            if self.triple.isDarwin() {
-                return try RelativePath(validating: "\(base)/Contents/MacOS/\(product.name)")
-            } else {
-                return try RelativePath(validating: base)
+            switch buildSystemKind {
+            case .native, .xcode:
+                let base = "\(product.name).xctest"
+                if self.triple.isDarwin() {
+                    return try RelativePath(validating: "\(base)/Contents/MacOS/\(product.name)")
+                } else {
+                    return try RelativePath(validating: base)
+                }
+            case .swiftbuild:
+                if self.triple.isDarwin() {
+                    let base = "\(product.name).xctest"
+                    return try RelativePath(validating: "\(base)/Contents/MacOS/\(product.name)")
+                } else {
+                    var base = "\(product.name)-test-runner"
+                    let ext = self.triple.executableExtension
+                    if !ext.isEmpty {
+                        base += ext
+                    }
+                    return try RelativePath(validating: base)
+                }
             }
         case .macro:
             #if BUILD_MACROS_AS_DYLIBS
--- a/Sources/SwiftBuildSupport/PIFBuilder.swift
+++ b/Sources/SwiftBuildSupport/PIFBuilder.swift
@@ -633,13 +633,13 @@ fileprivate func buildAggregateProject(
                         continue
                     }
                 }
-                
+
                 aggregateProject[keyPath: allIncludingTestsTargetKeyPath].common.addDependency(
                     on: target.id,
                     platformFilters: [],
                     linkProduct: false
                 )
-                if target.productType != .unitTest {
+                if ![.unitTest, .swiftpmTestRunner].contains(target.productType) {
                     aggregateProject[keyPath: allExcludingTestsTargetKeyPath].common.addDependency(
                         on: target.id,
                         platformFilters: [],
--- a/Sources/SwiftBuildSupport/PackagePIFBuilder.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFBuilder.swift
@@ -362,6 +362,7 @@ public final class PackagePIFBuilder {
         case framework
         case executable
         case unitTest
+        case unitTestRunner
         case bundle
         case resourceBundle
         case packageProduct
@@ -385,6 +386,7 @@ public final class PackagePIFBuilder {
             case .framework: .framework
             case .executable: .executable
             case .unitTest: .unitTest
+            case .swiftpmTestRunner: .unitTestRunner
             case .bundle: .bundle
             case .packageProduct: .packageProduct
             case .hostBuildTool: fatalError("Unexpected hostBuildTool type")
@@ -520,7 +522,11 @@ public final class PackagePIFBuilder {
         settings[.WATCHOS_DEPLOYMENT_TARGET] = builder.deploymentTargets[.watchOS] ?? nil
         settings[.DRIVERKIT_DEPLOYMENT_TARGET] = builder.deploymentTargets[.driverKit] ?? nil
         settings[.XROS_DEPLOYMENT_TARGET] = builder.deploymentTargets[.visionOS] ?? nil
-        settings[.DYLIB_INSTALL_NAME_BASE] = "@rpath"
+
+        for machoPlatform in [ProjectModel.BuildSettings.Platform.macOS, .macCatalyst, .iOS, .watchOS, .tvOS, .xrOS, .driverKit] {
+            settings.platformSpecificSettings[machoPlatform]![.DYLIB_INSTALL_NAME_BASE]! = ["@rpath"]
+        }
+
         settings[.USE_HEADERMAP] = "NO"
         settings[.OTHER_SWIFT_FLAGS].lazilyInitializeAndMutate(initialValue: ["$(inherited)"]) { $0.append("-DXcode") }
 
--- a/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+import TSCBasic
 import TSCUtility
 
 import struct Basics.AbsolutePath
@@ -121,13 +122,15 @@ extension PackagePIFProjectBuilder {
         if mainModule.type == .test {
             // FIXME: we shouldn't always include both the deep and shallow bundle paths here, but for that we'll need rdar://31867023
             settings[.LD_RUNPATH_SEARCH_PATHS] = [
-                "@loader_path/Frameworks",
-                "@loader_path/../Frameworks",
+                "$(RPATH_ORIGIN)/Frameworks",
+                "$(RPATH_ORIGIN)/../Frameworks",
                 "$(inherited)"
             ]
             settings[.GENERATE_INFOPLIST_FILE] = "YES"
             settings[.SKIP_INSTALL] = "NO"
             settings[.SWIFT_ACTIVE_COMPILATION_CONDITIONS].lazilyInitialize { ["$(inherited)"] }
+            // Enable index-while building for Swift compilations to facilitate discovery of XCTest tests.
+            settings[.SWIFT_INDEX_STORE_ENABLE] = "YES"
         } else if mainModule.type == .executable {
             // Setup install path for executables if it's in root of a pure Swift package.
             if pifBuilder.delegate.hostsOnlyPackages && pifBuilder.delegate.isRootPackage {
@@ -502,9 +505,13 @@ extension PackagePIFProjectBuilder {
             linkedPackageBinaries: linkedPackageBinaries,
             swiftLanguageVersion: mainModule.packageSwiftLanguageVersion(manifest: packageManifest),
             declaredPlatforms: self.declaredPlatforms,
-            deploymentTargets: self.deploymentTargets
+            deploymentTargets: mainTargetDeploymentTargets
         )
         self.builtModulesAndProducts.append(moduleOrProduct)
+
+        if moduleOrProductType == .unitTest {
+            try makeTestRunnerProduct(for: moduleOrProduct)
+        }
     }
 
     private mutating func handleProduct(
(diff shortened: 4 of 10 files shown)


Sources/SwiftBuildSupport/PIFBuilder.swift
```diff
@@ -362,20 +362,20 @@ fileprivate func buildAggregateProject(
                     // conflicts with those from "PACKAGE-TARGET:Foo-dynamic".
                     continue
                 }
-
+/*
```
owenv, line 365: fixme: undo this

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
```diff
@@ -498,7 +500,15 @@ public final class PackagePIFBuilder {
         settings[.WATCHOS_DEPLOYMENT_TARGET] = builder.deploymentTargets[.watchOS] ?? nil
         settings[.DRIVERKIT_DEPLOYMENT_TARGET] = builder.deploymentTargets[.driverKit] ?? nil
         settings[.XROS_DEPLOYMENT_TARGET] = builder.deploymentTargets[.visionOS] ?? nil
-        settings[.DYLIB_INSTALL_NAME_BASE] = "@rpath"
+
```
owenv, line 503: This is a targeted fix, but we should try to mixup how we set sonames more generally on linux

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
```diff
@@ -995,6 +1001,102 @@ extension PackagePIFProjectBuilder {
...
     }
+
+    // MARK: - Test Runners
+    mutating func makeTestRunnerProduct(for unitTestProduct: PackagePIFBuilder.ModuleOrProduct) throws {
+        // Only generate a test runner for root packages with tests.
+        guard pifBuilder.delegate.isRootPackage else {
+            return
+        }
+
+        guard let unitTestModuleName = unitTestProduct.moduleName else {
```
owenv, line 1012: todo: throw a real error

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
```diff
@@ -995,6 +1001,102 @@ extension PackagePIFProjectBuilder {
...
+        var settings: BuildSettings = self.package.underlying.packageBaseBuildSettings
+        let impartedSettings = BuildSettings()
+
+        settings[.TARGET_NAME] = name
+        settings[.PACKAGE_RESOURCE_TARGET_KIND] = "regular"
+        settings[.PRODUCT_NAME] = "$(TARGET_NAME)"
+        settings[.PRODUCT_MODULE_NAME] = moduleName
+        settings[.PRODUCT_BUNDLE_IDENTIFIER] = "\(self.package.identity).\(name)"
+            .spm_mangledToBundleIdentifier()
+        settings[.EXECUTABLE_NAME] = name
```
owenv, line 1038: The PIFBuilder is overriding EXECUTABLE_NAME for a bunch of target types, which sometimes strips extensions and prefixes of product types, we might need to rework some of that

owenv:
Locally, I can pass a decent chunk of TestDiscoveryTests on linux. macOS will need a bit of work to skip building the test runner executables, and I expect Windows will need a few tweaks

@swift-ci test

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
jakepetroules: $ORIGIN is an appropriate value roughly equivalent to @rpath for ELF platforms

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
```diff
@@ -498,7 +500,15 @@ public final class PackagePIFBuilder {
...
+        for machoPlatform in [ProjectModel.BuildSettings.Platform.macOS, .macCatalyst, .iOS, .watchOS, .tvOS, .xrOS, .driverKit] {
+            settings.platformSpecificSettings[machoPlatform]![.DYLIB_INSTALL_NAME_BASE]! = ["@rpath"]
+        }
+        for elfPlatform in [ProjectModel.BuildSettings.Platform.linux, .android] {
```
jakepetroules, line 507: We have FreeBSD and OpenBSD in there right now, don't we? Should add those too.

owenv:
@swift-ci test

owenv:
@swift-ci test

Tests/FunctionalTests/TestDiscoveryTests.swift
```diff
@@ -12,159 +12,156 @@
...
-            XCTAssertMatch(stderr, .contains("Build complete!"))
+            #expect(stderr.contains("Build complete!"))
             // in "swift test" test output goes to stdout
-            XCTAssertMatch(stdout, .contains("Executed 3 tests"))
+            #expect(stdout.contains("Executed 3 tests"))
         }
     }
 
-    func testNonStandardName() async throws {
+    // FIXME: Fix --build-system swiftbuild link file list quoting on non-Darwin
```
bkhouri: question: do we have a GitHub issue for this? or is to plan to address this FIXME before merge?  Can we create a GitHub issue and add a `.bug()` trait that references said issue?

Tests/FunctionalTests/TestDiscoveryTests.swift
```diff
@@ -12,159 +12,156 @@
...
-            XCTAssertMatch(stdout, .contains("Executed 4 tests"))
+            #expect(stdout.contains("Executed 4 tests"))
         }
     }
 
-    func testDiscovery_whenNoTests() async throws {
-        #if os(macOS)
-        try XCTSkipIf(true)
-        #endif
+    // FIXME: eliminate extraneous warnings with --build-system swiftbuild
```
bkhouri, line 79: question: do we have a GitHub issue for this? or is to plan to address this FIXME before merge?  Can we create a GitHub issue and add a `.bug()` trait that references said issue?

Tests/FunctionalTests/TestDiscoveryTests.swift
```diff
@@ -12,159 +12,156 @@
...
+            #expect(stdout.contains("Executed 0 tests"))
         }
     }
 
-    func testEntryPointOverride() async throws {
-        #if os(macOS)
-        try XCTSkipIf(true)
-        #endif
-
+    // FIXME: --build-system swiftbuild should support hand-authored entry points.
```
bkhouri, line 93: question: do we have a GitHub issue for this? or is to plan to address this FIXME before merge? Can we create a GitHub issue and add a .bug() trait that references said issue?

bkhouri (commented):
Thanks for converting the TestDiscoverTests from XCTest to Swift Testing.  I have a couple questions/comments, but nothing blocking.

Tests/FunctionalTests/TestDiscoveryTests.swift
owenv: added links for each one in the latest push

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
Working to address linker failures in https://github.com/swiftlang/swift-build/pull/576

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
jakepetroules: Can we not use SDK conditions here?

jakepetroules: Also, maybe we add a `USE_RELOCATABLE_DYLIB_INSTALL_NAME_BASE` boolean setting which causes `DYLIB_INSTALL_NAME_BASE` to default to an appropriate platform-specific value for relocatable libraries?

Then internally in Swift Build we can just check if the triple is Apple/MachO based and use @rpath, otherwise use $ORIGIN and it should cover almost every conceivable platform.

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -4106,11 +4106,12 @@ class PackageCommandSwiftBuildTests: PackageCommandTestCase {
         throw XCTSkip("SWBINTTODO: Build plan is not currently supported")
     }
 
+#if !os(macOS)
```
bkhouri: question: why are we not executing this test on macOS?

Tests/FunctionalTests/TestDiscoveryTests.swift
```diff
@@ -12,159 +12,166 @@
...
-final class TestDiscoveryTests: XCTestCase {
-    func testBuild() async throws {
+@Suite
+struct TestDiscoveryTests {
+
+    // Conditional parameters to tests are not supported by Swift Testing currently, so we construct the list of params here.
+    static var buildSystems: [BuildSystemProvider.Kind] = {
+        if ProcessInfo.hostOperatingSystem == .windows {
+            // Windows currently fails due to long path handling issues on Windows
+            return [BuildSystemProvider.Kind.native]
```
bkhouri: issue (blocking): we should execute these tests on Windows and mark each test as `withKnownIssue` so we get signal, so we can ensure the test is enabled once the underlying windows path issue is fixed.

Have a looks at `Tests/FunctionalTests/TraitTests.swift` for how those tests handle it.

bkhouri (changes requested):
I only reviewed the tests changes.  Although the overcharge is good, we should ensure to run the TestDicoveryTests against SwiftBuild on Windows and use the `withKnownIssue` Swift Testing API to essentially mark the test as "expected fail".  This will give us signal when the underlying Windows issue is fix, thus prompting the removal of the `withKnownIssue` as opposed to having to remember to augment the test.

Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
```diff
@@ -314,11 +314,26 @@ public struct BuildParameters: Encodable {
...
+            switch buildSystemKind {
+            case .native, .xcode:
+                let base = "\(product.name).xctest"
+                if self.triple.isDarwin() {
+                    return try RelativePath(validating: "\(base)/Contents/MacOS/\(product.name)")
+                } else {
+                    return try RelativePath(validating: base)
+                }
+            case .swiftbuild:
+                if self.triple.isDarwin() {
```
jakepetroules, line 326: nit: could do if-macOS `"\(base)/Contents/MacOS/\(product.name)"` else-if-Darwin "\(base)/\(product.name)" so it'll support making `swift test` work for simulator targets (acknowledging that would need additional changes elsewhere for that to be fully working)

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
```diff
@@ -520,7 +522,15 @@ public final class PackagePIFBuilder {
         settings[.WATCHOS_DEPLOYMENT_TARGET] = builder.deploymentTargets[.watchOS] ?? nil
         settings[.DRIVERKIT_DEPLOYMENT_TARGET] = builder.deploymentTargets[.driverKit] ?? nil
         settings[.XROS_DEPLOYMENT_TARGET] = builder.deploymentTargets[.visionOS] ?? nil
-        settings[.DYLIB_INSTALL_NAME_BASE] = "@rpath"
+
+        for machoPlatform in [ProjectModel.BuildSettings.Platform.macOS, .macCatalyst, .iOS, .watchOS, .tvOS, .xrOS, .driverKit] {
+            settings.platformSpecificSettings[machoPlatform]![.DYLIB_INSTALL_NAME_BASE]! = ["@rpath"]
+        }
+        for elfPlatform in [ProjectModel.BuildSettings.Platform.linux, .android, .openbsd, .freebsd] {
+            // FIXME: Treating $ORIGIN as part of the soname is not really the right thing to do here
```
jakepetroules: I think we can just not set it at all. The soname on ELF platforms should just be the leaf name. $ORIGIN is only used in rpaths. I realize I might've misled you in my previous comment, I was misremembering what the conventions were before.

(Referencing Qbs, that's what we did there -- libraries set cpp.sonamePrefix (DYLIB_INSTALL_NAME_BASE equivalent) to @rpath on Darwin and nothing elsewhere, while cpp.rpaths (LD_RUNPATH_SEARCH_PATHS equivalent) referenced cpp.rpathOrigin, a helper variable which is @loader_path on Darwin and $ORIGIN on ELF)

Tests/CommandsTests/PackageCommandTests.swift
owenv: it was a vestige of an earlier diff, dropped it

Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
owenv: I think I'll probably leave this to future work

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
owenv: Got this working but this patch now depends on https://github.com/swiftlang/swift-build/pull/604

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
Windows: 'TestCommandSwiftBuildTests.testListWithSkipBuildAndNoBuildArtifacts' failed (14.613 seconds)

owenv:
@swift-ci test

owenv:
@swift-ci test

Tests/FunctionalTests/TestDiscoveryTests.swift
```diff
@@ -12,159 +12,185 @@
...
-    func testBuild() async throws {
-        try await fixture(name: "Miscellaneous/TestDiscovery/Simple") { fixturePath in
-            let (stdout, _) = try await executeSwiftBuild(fixturePath)
-            // in "swift build" build output goes to stdout
-            XCTAssertMatch(stdout, .contains("Build complete!"))
+@Suite
+struct TestDiscoveryTests {
+    static var buildSystems: [BuildSystemProvider.Kind] = [BuildSystemProvider.Kind.native, .swiftbuild]
+
+    @Test(arguments: buildSystems)
```
bkhouri, line 25: suggestion (non-blocking, if-pr-update):  using `arguments: SupportedBuildSystemOnAllPlatforms` does essentially the same thing :)

Tests/FunctionalTests/TestDiscoveryTests.swift
```diff
@@ -12,159 +12,185 @@
...
+                // in "swift test" test output goes to stdout
+                #expect(stdout.contains("Executed 3 tests"))
+            }
+        } when: {
+            buildSystem == .swiftbuild && ProcessInfo.hostOperatingSystem == .windows
         }
     }
 
-    func testNonStandardName() async throws {
+    @Test(.bug("https://github.com/swiftlang/swift-build/issues/13"), arguments: [BuildSystemProvider.Kind.native])
```
bkhouri, line 53: question: out of curiosity, why aren't we executing this test against the SwiftBuild build system?

bkhouri (approved):

bkhouri:
@swift-ci test windows

Tests/FunctionalTests/TestDiscoveryTests.swift
owenv: There are some design questions I need to sort out related to response file quoting on non-Darwin platforms and it's not 100% clear to me marking this as a known issue is the right thing to do yet

owenv:
@swift-ci smoke test macOS

owenv:
@swift-ci test macOS

owenv:
macOS checkout issue
@swift-ci test macOS

PR #8723, opened 2025-05-26
build-using-self: Add ability to pass extra build/run/test arguments

kcieplak:
Add the ability to inject arbitrary arguments to the invocations of swift <build|run|test> from the build-using-self-script.
Add a boolean flag to skip running the bootstrap subscript. Default is to run it.

This allows one to add options like '--jobs' or '--event-stream-output-path' .
Thus allowing for different use cases in pipelines, like collection of test output json to a specific path.

For example:
`build-using-self -v --skip-bootstrap --additional-test-args '--event-stream-output-path spm-test.json'`


kcieplak:
@swift-ci test

PR #8724, opened 2025-05-26
Allow staging in new BuildSettings.Declaration cases without introducing failures

owenv:
Until we get https://github.com/swiftlang/swift-evolution/blob/main/proposals/0487-extensible-enums.md or a similar mechanism, add some redundant default clauses so that changes like https://github.com/swiftlang/swift-build/pull/526 can be staged in without revlock.


jakepetroules (approved):
This will generate warnings that can't be silenced, right? Might want to note in the comments that these should not be removed until SE-0487 is available.

owenv:
@swift-ci test Windows

owenv:
@swift-ci test macOS

PR #8733, opened 2025-05-27
SwiftFixIt: Always rely on the UTF-8 offset of an input source location

AnthonyLatsis:
As it turns out, contrary to an old comment in the diagnostic deserializer, the Swift compiler *does* serialize the offset of a source location, so stop assuming that it may have been left blank.

### Motivation:

 This slightly simplifies the diagnostic conversion logic and also enables us to stop hashing the line and column in the diagnostic filter.


AnthonyLatsis:
@swift-ci please test

xedin:
@swift-ci please test self hosted macOS platform

AnthonyLatsis:

PR #8736, opened 2025-05-28
Allow C targets to import the compatibility header generated for Swift libraries

### LINKED ISSUE #7257: Failed to generate a static or dynamic Swift library called by C code

xymus:
Address two issues preventing a C source file to import the generated compatibility header:

1. Don't mark the module map generated for the compatibility header as `requires objc`. This triggers an error when importing it from a C source file. The compatibility header is already printed in a way where the Objective-C code is protected behind a language check. C clients can safely import it even if they may not see any content. Plus we're adding C content to the compatibility header with the official support for `@cdecl` that is independent of Objective-C.

2. Emit the generated compatibility header and module map in a new `include` directory under the temporary build directory of each Swift library. This directory is then used as a header search path for the dependent clang targets.

   This replaces the previous strategy that relied only the generated module map to resolve imports. It was incompatible with C source files and non-modular textual includes. The use of a header search path preserves the support for module imports via the same module map and supports textual includes of the compatibility header.

Fixes #7257.

Follow up to #8695.

### DIFF
--- a/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
@@ -898,7 +898,6 @@ public final class SwiftModuleBuildDescription {
             #"""
             module \#(self.target.c99name) {
                 header "\#(self.objCompatibilityHeaderPath.pathString)"
-                requires objc
             }
 
             """#.utf8
--- a/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
@@ -890,9 +890,15 @@ public final class SwiftModuleBuildDescription {
         try self.fileSystem.writeFileContents(path, bytes: .init(encodingAsUTF8: content), atomically: true)
     }
 
+    /// Directory for the the compatibility header and module map generated for this target.
+    /// The whole directory should be usable as a header search path.
+    private var compatibilityHeaderDirectory: AbsolutePath {
+        tempsPath.appending("include")
+    }
+
     /// Generates the module map for the Swift target and returns its path.
     private func generateModuleMap() throws -> AbsolutePath {
-        let path = self.tempsPath.appending(component: moduleMapFilename)
+        let path = self.compatibilityHeaderDirectory.appending(component: moduleMapFilename)
 
         let bytes = ByteString(
             #"""
@@ -916,7 +922,7 @@ public final class SwiftModuleBuildDescription {
 
     /// Returns the path to the ObjC compatibility header for this Swift target.
     var objCompatibilityHeaderPath: AbsolutePath {
-        self.tempsPath.appending("\(self.target.name)-Swift.h")
+        self.compatibilityHeaderDirectory.appending("\(self.target.name)-Swift.h")
     }
 
     /// Returns the build flags from the declared build settings.
--- a/Sources/Build/BuildPlan/BuildPlan+Clang.swift
+++ b/Sources/Build/BuildPlan/BuildPlan+Clang.swift
@@ -30,7 +30,8 @@ extension BuildPlan {
             case is SwiftModule:
                 if case let .swift(dependencyTargetDescription)? = description {
                     if let moduleMap = dependencyTargetDescription.moduleMap {
-                        clangTarget.additionalFlags += ["-fmodule-map-file=\(moduleMap.pathString)"]
+                        // C languages clients should either import the module or include the compatibility header next to it.
+                        clangTarget.additionalFlags += ["-I", moduleMap.dirname]
                     }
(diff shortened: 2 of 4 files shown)


xymus:
@swift-ci Please test

xymus:
@swift-ci Please test Windows

xymus:
@swift-ci Please test

xymus:
@swift-ci Please test

xymus:
@swift-ci Please test Windows

xymus:
Is someone available to review this change?

xymus:
@jakepetroules do you think we need a similar change in swift-build? It appears to print the `requires objc` as well, however I don’t think it uses the same header search strategy relying only on the module map.

jakepetroules:
> @jakepetroules do you think we need a similar change in swift-build? It appears to print the `requires objc` as well, however I don’t think it uses the same header search strategy relying only on the module map.

I think so, yeah, at least for the `requires objc` bit.

dschaefer2 (approved):

PR #8741, opened 2025-05-29
Binary Static Library Artifact auditing tool

daniel-grumberg:
This is an implementation of the auditing tool described in [SE-0482](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0482-swiftpm-static-library-binary-target-non-apple-platforms.md).

### Motivation:

As part of SE-0482 and https://github.com/swiftlang/swift-package-manager/pull/8639 we introduced the ability to depend on prebuilt static libraries (that expose a C interface) and that don't have any dependencies outside of the C standard library. This PR introduces an auditing tool that checks the ABI of static library artifact bundle and checks if it's compatible with the current host platform.

### Modifications:

- Create a new package subcommand that checks a local artifact bundle for unexpected external dependencies.
- New internal APIs to inspect the ABI of a binary object (object file, static archive, dynamic library).

### Result:

Users will be able to validate that their static library binary artifacts won't cause runtime issues for users.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -319,6 +319,19 @@ let package = Package(
             ]
         ),
 
+        .target(
+            /** API for inspecting symbols defined in binaries */
+            name: "BinarySymbols",
+            dependencies: [
+                "Basics",
+                .product(name: "TSCBasic", package: "swift-tools-support-core"),
+            ],
+            exclude: ["CMakeLists.txt"],
+            swiftSettings: commonExperimentalFeatures + [
+                .unsafeFlags(["-static"]),
+            ]
+        ),
+
         // MARK: Project Model
 
         .target(
@@ -580,6 +593,7 @@ let package = Package(
                 .product(name: "ArgumentParser", package: "swift-argument-parser"),
                 .product(name: "OrderedCollections", package: "swift-collections"),
                 "Basics",
+                "BinarySymbols",
                 "Build",
                 "CoreCommands",
                 "PackageGraph",
--- a/Package.swift
+++ b/Package.swift
@@ -955,6 +955,10 @@ let package = Package(
             dependencies: ["SwiftFixIt", "_InternalTestSupport"]
         ),
         .testTarget(
+            name: "BinarySymbolsTests",
+            dependencies: ["BinarySymbols", "_InternalTestSupport"]
+        ),
+        .testTarget(
             name: "XCBuildSupportTests",
             dependencies: ["XCBuildSupport", "_InternalTestSupport", "_InternalBuildTestSupport"],
             exclude: ["Inputs/Foo.pc"]
--- a/Sources/BinarySymbols/ClangHostDefaultObjectsDetector.swift
+++ b/Sources/BinarySymbols/ClangHostDefaultObjectsDetector.swift
@@ -1,0 +1,90 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2014 - 2025 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+ */
+
+import Basics
+import Foundation
+
+import protocol TSCBasic.WritableByteStream
+
+package func detectDefaultObjects(
+    clang: AbsolutePath, fileSystem: any FileSystem, hostTriple: Triple
+) async throws -> [AbsolutePath] {
+    let clangProcess = AsyncProcess(args: clang.pathString, "-###", "-x", "c", "-")
+    let stdinStream = try clangProcess.launch()
+    stdinStream.write(
+        #"""
+        #include <stdio.h>
+        int main(int argc, char *argv[]) {
+            printf("Hello world!\n")
+            return 0;
+        }
+        """#
+    )
+    stdinStream.flush()
+    try stdinStream.close()
+    let clangResult = try await clangProcess.waitUntilExit()
+    guard case .terminated(let status) = clangResult.exitStatus,
+        status == 0
+    else {
+        throw StringError("Couldn't run clang on sample hello world program")
+    }
+    let commandsStrings = try clangResult.utf8stderrOutput().split(whereSeparator: \.isNewline)
+
+    let commands = commandsStrings.map { $0.split(whereSeparator: \.isWhitespace) }
+    guard let linkerCommand = commands.last(where: { $0.first?.contains("ld") == true }) else {
+        throw StringError("Couldn't find default link command")
+    }
+
+    // TODO: This logic doesn't support Darwin and Windows based, c.f. https://github.com/swiftlang/swift-package-manager/issues/8753
+    let libraryExtensions = [hostTriple.staticLibraryExtension, hostTriple.dynamicLibraryExtension]
+    var objects: Set<AbsolutePath> = []
+    var searchPaths: [AbsolutePath] = []
+
+    var linkerArguments = linkerCommand.dropFirst().map {
+        $0.replacingOccurrences(of: "\"", with: "")
+    }
+
+    if hostTriple.isLinux() {
+        // Some platform still separate those out...
+        linkerArguments.append(contentsOf: ["-lm", "-lpthread", "-ldl"])
+    }
+
+    for argument in linkerArguments {
+        if argument.hasPrefix("-L") {
+            searchPaths.append(try AbsolutePath(validating: String(argument.dropFirst(2))))
+        } else if argument.hasPrefix("-l") && !argument.hasSuffix("lto_library") {
+            let libraryName = argument.dropFirst(2)
+            let potentialLibraries = searchPaths.flatMap { path in
+                if libraryName == "gcc_s" && hostTriple.isLinux() {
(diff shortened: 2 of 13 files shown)


Sources/BinarySymbols/ClangHostDefaultObjectsDetector.swift
```diff
@@ -0,0 +1,86 @@
...
+        throw StringError("Couldn't run clang on sample hello world program")
+    }
+    let commandsStrings = try clangResult.utf8stderrOutput().split(whereSeparator: \.isNewline)
+
+    let commands = commandsStrings.map { $0.split(whereSeparator: \.isWhitespace) }
+    guard let linkerCommand = commands.last(where: { $0.first?.contains("ld") == true }) else {
+        throw StringError("Couldn't find default link command")
+    }
+
+    // TODO: extend this to support dylib and other formats for Darwin and dll and lib files on Windows.
```
FranzBusch: Let's create a GH issue for this to not forget it

Sources/Commands/PackageCommands/AuditBinaryArtifact.swift
```diff
@@ -0,0 +1,124 @@
...
+import Foundation
+import PackageModel
+import SPMBuildCore
+import Workspace
+
+import struct TSCBasic.StringError
+
+struct AuditBinaryArtifact: AsyncSwiftCommand {
+    static let configuration = CommandConfiguration(
+        abstract: "Compute the checksum for a binary artifact.")
```
FranzBusch:
```suggestion
        abstract: "Audit a static binary artifact for undefined symbols.")
```

Sources/Commands/PackageCommands/AuditBinaryArtifact.swift
```diff
@@ -0,0 +1,124 @@
...
+                    "Invalid artifact binary \(library.libraryPath.pathString), found undefined symbols:"
+                )
+                for name in symbols.undefined {
+                    print("- \(name)")
+                }
+                throw ExitCode(1)
+            }
+        }
+
+        print("Artifact is safe to use across supported Swift deployment targets!")
```
FranzBusch: Should this rather print for current build target?

Sources/Commands/PackageCommands/AuditBinaryArtifact.swift
```diff
@@ -0,0 +1,124 @@
...
+import Workspace
+
+import struct TSCBasic.StringError
+
+struct AuditBinaryArtifact: AsyncSwiftCommand {
+    static let configuration = CommandConfiguration(
+        abstract: "Compute the checksum for a binary artifact.")
+
+    @OptionGroup(visibility: .hidden)
+    var globalOptions: GlobalOptions
```
FranzBusch, line 31: These are not used right?

daniel-grumberg: They are required to conform to `SwiftCommand` unfortunately.

rauhul:
Please correct me if I'm wrong, but I don't see how this tool checks for compatible ABIs at all. It seems to only deal with symbol names and doesn't have any knowledge the actual function ABI. 

Also is there a reason this only works for the host triple? This seems like an odd requirement. 

Lastly, why does this manually call shell tools? This functionality should likely be expressed in an action graph (via llbuild). The his would mean it could be included in larger graphs as well individually operations can be cached.

daniel-grumberg:
> We would also like to provide docker image that bundles the tool and has the oldest supported libc in the Swift ecosystem which is glibc 2.26 as of Swift 6.

Upon further reflection we can tell users to the 6.2 amazonlinux2 image once it comes out, until we figure out a more robust strategy in a future release.

daniel-grumberg:
@rauhul answers inline:

> Please correct me if I'm wrong, but I don't see how this tool checks for compatible ABIs at all. It seems to only deal with symbol names and doesn't have any knowledge the actual function ABI.
> Also is there a reason this only works for the host triple? This seems like an odd requirement.

This might a terminology misnomer (I am not a low level runtime person), but what it does validate is that any referenced symbols are either provided by default (e.g. compiler runtimes) or come from the system libc. It needs to run on the host triple because we don't actually have access to the system libc on the deployment target currently. We will recommend users run this on amazonlinux2 images as they have the oldest libc of the supported images.

> Lastly, why does this manually call shell tools? This functionality should likely be expressed in an action graph (via llbuild). The his would mean it could be included in larger graphs as well individually operations can be cached.

I am not familiar with this, but this is not part of the build, it is a purely one of operation intended to be run in CI by binary static library artifact vendors.

daniel-grumberg:
@swift-ci please smoke-test

daniel-grumberg:
@swift-ci please test

Sources/BinarySymbols/LLVMObjdumpSymbolProvider.swift
```diff
@@ -0,0 +1,155 @@
...
+            // case "*UND*":
+            //     guard recordUndefined else {
+            //         continue
+            //     }
+            //     if line[line.index(line.startIndex, offsetBy: Int)]
+            //         symbols.addUndefined(String(name))
+            //     }
+            // default:
+            //     symbols.addDefined(String(name))
+            // }
```
MaxDesiatov: Is this supposed to go in as a comment?

Sources/BinarySymbols/ReferencedSymbols.swift
```diff
@@ -0,0 +1,31 @@
...
+ Copyright (c) 2014 - 2025 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See http://swift.org/LICENSE.txt for license information
+ See http://swift.org/CONTRIBUTORS.txt for Swift project authors
+ */
+
+package struct ReferencedSymbols {
+    package var defined: Set<String>
+    package var undefined: Set<String>
```
MaxDesiatov: Are these supposed to be `package` if corresponding `addUndefined` and `addDefined` are `internal`?
```suggestion
    var defined: Set<String>
    var undefined: Set<String>
```

Sources/PackageModel/UserToolchain.swift
```diff
@@ -429,6 +429,15 @@ public final class UserToolchain: Toolchain {
         )
     }
 
+    /// Returns the path to llvm-objdump tool.
+    public func getLLVMObjdump() throws -> AbsolutePath {
```
MaxDesiatov: Let's avoid adding `public` APIs unless absolutely necessary
```suggestion
    package func getLLVMObjdump() throws -> AbsolutePath {
```

Package.swift
```diff
@@ -319,6 +319,20 @@ let package = Package(
...
+                "Basics",
+                .product(name: "TSCBasic", package: "swift-tools-support-core"),
+            ],
+            exclude: ["CMakeLists.txt"],
+            swiftSettings: commonExperimentalFeatures + [
+                .unsafeFlags(["-static"]),
+            ]
+        ),
+
+
```
MaxDesiatov: formatting nit
```suggestion

```

Sources/BinarySymbols/ReferencedSymbols.swift
daniel-grumberg: This needs to be available so that the tool can have a look. I made the `add*` APIs package as well.

Sources/BinarySymbols/LLVMObjdumpSymbolProvider.swift
daniel-grumberg: Oops 😅

Sources/BinarySymbols/ReferencedSymbols.swift
MaxDesiatov: How's about `package internal(set)` or `package private(set)`?

daniel-grumberg:
@swift-ci please test

FranzBusch (approved):

MaxDesiatov:
@swift-ci test windows

dschaefer2 (approved):

daniel-grumberg:
@swift-ci please test windows

daniel-grumberg:
@swift-ci please test

MaxDesiatov:

daniel-grumberg:
@swift-ci test self-hosted windows

Sources/Commands/PackageCommands/AuditBinaryArtifact.swift
```diff
@@ -0,0 +1,136 @@
...
+
+        let archiveDirectory = scratchDirectory.appending(
+            components: "artifact-auditing",
+            path.basenameWithoutExt, UUID().uuidString
+        )
+        try fileSystem.forceCreateDirectory(at: archiveDirectory)
+
+        try await archiver.extract(from: path, to: archiveDirectory)
+
+        let artifacts = try fileSystem.getDirectoryContents(archiveDirectory)
```
FranzBusch, line 123: Just to check are we filtering somewhere to only use the `.a` that also applies to the current host triplet?

daniel-grumberg: `parseLibraryArtifactArchives` does it for us.

daniel-grumberg:

daniel-grumberg:
@swift-ci test

daniel-grumberg:
@swift-ci test windows

PR #8750, opened 2025-06-01
Leverage package graph, add include paths.

dschaefer2:
Take advantage of the fact we're in swiftpm and calculate things out of the package graph instead of hard coding. In particular, we compute the list of targets for the requested list of products, as well as the list of C modules and their include paths.

Swift build is better at ensuring the include path is added to all targets that need it. Unfortunately when mixing libraries that use swift-syntax and macros, we end up adding both the C modules from the checkouts and the prebuilts causing duplicate definition of the C modules. By adding the above include paths and leveraging that we always checkout the source anyway, switch to using the checked out include paths, thus avoiding the duplicate.

Also some prep for adding a config.json file so the list of prebuilts can be changed without a PR to SwiftPM. Removes the magic we were using to add the prebuilt library to the package manifest and just add them at the end.

### DIFF
--- a/Tests/BuildTests/BuildPlanTests.swift
+++ b/Tests/BuildTests/BuildPlanTests.swift
@@ -4721,6 +4721,7 @@ class BuildPlanTestCase: BuildSystemProviderTestCase {
             identity: .plain("swift-syntax"),
             libraryName: "MacroSupport",
             path: "/MyPackage/.build/prebuilts/swift-syntax/600.0.1/6.1-MacroSupport-macos_aarch64",
+            checkoutPath: "/MyPackage/.build/checkouts/swift-syntax",
             products: [
                 "SwiftBasicFormat",
                 "SwiftCompilerPlugin",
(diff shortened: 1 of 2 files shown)


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test macOS

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -121,69 +90,21 @@ var prebuiltRepos: IdentifiableSet<PrebuiltRepos> = [
...
 
-let dockerImageRoot = "swiftlang/swift:nightly-6.1-"
-
 @main
 struct BuildPrebuilts: AsyncParsableCommand {
     @Option(help: "The directory to generate the artifacts to.")
     var stageDir = try! AbsolutePath(validating: FileManager.default.currentDirectoryPath).appending("stage")
 
-    @Flag(help: "Whether to build artifacts using docker.")
-    var docker = false
```
cmcgee1024, line 180: question: Is this a breaking change to the command? If so, are all clients migrated away from the docker flag and command?

cmcgee1024 (approved):

Sources/swift-build-prebuilts/BuildPrebuilts.swift
dschaefer2: There are only two users of this command. One of them is me. :) The prebuilts for Linux are created by running this command inside a docker container so this option isn't used.

PR #8758, opened 2025-06-03
Temporarily skip a Swift Build test to allow staging in changes to the test product type.

owenv:
We're staging in changes so tests build as dylibs with a separate runner executable on non-Darwin platforms, which more closely matches the native build system. Skip this test temporarily as it will fail after the Swift Build changes land but before the SwiftPM ones do, to avoid the need for coordinated merges

### DIFF
--- a/Tests/CommandsTests/PackageCommandTests.swift
+++ b/Tests/CommandsTests/PackageCommandTests.swift
@@ -4081,14 +4081,13 @@ class PackageCommandSwiftBuildTests: PackageCommandTestCase {
         throw XCTSkip("SWBINTTODO: Build plan is not currently supported")
     }
 
-#if !os(macOS)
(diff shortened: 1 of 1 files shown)


Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -4081,14 +4081,13 @@ class PackageCommandSwiftBuildTests: PackageCommandTestCase {
         throw XCTSkip("SWBINTTODO: Build plan is not currently supported")
     }
 
-#if !os(macOS)
```
owenv, line 4084: Testing locally this test works for me on macOS, so I don't think this is needed anymore

owenv:
@swift-ci test

jakepetroules (approved):

PR #8761, opened 2025-06-03
[PackageModel] Toolchain: A few fixes for features supported by Swift compiler

xedin:
### Motivation:

- Fix "supported features" parsing with older toolchains. `-print-supported-features` flag used to output "enabled_in" version as an integer (major only).
- Handle absence of optional features section. `JSON.get` would throw if the key couldn't be found, let's handle that gracefully for an optional "optional features" section.

### Modifications:

- Update `Toolchain.swiftCompilerSupportedFeatures` to use `try?` when querying JSON for an optional "optional" key.
- Update `Toolchain.swiftCompilerSupportedFeatures` to attempt to retrieve "enabled_in" key as a String first and if that fails - as an integer.

### Result:

`swift package migrate` can now support older toolchains that have "enabled_in" printed as an integer and have no "optional features" section in their `-print-supported-features` output.

### DIFF
--- a/Sources/PackageModel/Toolchain+SupportedFeatures.swift
+++ b/Sources/PackageModel/Toolchain+SupportedFeatures.swift
@@ -102,7 +102,9 @@ extension Toolchain {
 
             let features: JSON = try parsedSupportedFeatures.get("features")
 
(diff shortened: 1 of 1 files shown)


xedin:
@swift-ci please test

jakepetroules (approved):

PR #8767, opened 2025-06-04
Provide library paths to plugin tools when run as custom tasks

cmcgee1024:
Linux build tool plugins won't work because they require a library
path to the Swift standard libraries. Add the toolchain's library path
to the environment for the custom task. Enable the tests that cover
this area for non-macOS platforms.

### DIFF
--- a/Sources/SwiftBuildSupport/PIFBuilder.swift
+++ b/Sources/SwiftBuildSupport/PIFBuilder.swift
@@ -362,6 +362,35 @@ public final class PIFBuilder {
                                     env[key.rawValue] = value
                                 }
 
+                                if let libDir = try? buildParameters.toolchain.toolchainLibDir {
+                                    var libPathVar: String? = nil
+                                    var libPathEntry: String? = nil
+                                    var libPathSeparator: String? = nil
+#if os(macOS)
+                                    libPathVar = "DYLD_LIBRARY_PATH"
+                                    libPathEntry = libDir.appending("swift").appending("macosx").pathString
+                                    libPathSeparator = ":"
+#elseif os(Linux)
+                                    libPathVar = "LD_LIBRARY_PATH"
+                                    libPathEntry = libDir.appending("swift").appending("linux").pathString
+                                    libPathSeparator = ":"
+#elseif os(Windows)
+                                    libPathVar = "PATH"
+                                    libPathEntry = libDir.pathString
+                                    libPathSeparator = ";"
+#endif
+
+                                    if let libPathVar, let libPathEntry, let libPathSeparator {
+                                        let existingEntry = env[libPathVar] ?? ""
+                                        if case var parts = existingEntry.split(separator: libPathSeparator).map({ String($0) }), !parts.contains(libPathEntry) {
+                                            parts.append(libPathEntry)
+                                            env[libPathVar] = parts.joined(separator: libPathSeparator)
+                                        }
+                                    } else {
+                                        observabilityScope.emit(warning: "Unable to set the library path for this platform. Some plugin build tools may not run without this.")
+                                    }
+                                }
+
                                 let workingDir = buildCommand.configuration.workingDirectory
 
                                 let writableDirectories: [AbsolutePath] = buildCommand.outputFiles
--- a/Sources/SwiftBuildSupport/PIFBuilder.swift
+++ b/Sources/SwiftBuildSupport/PIFBuilder.swift
@@ -10,7 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import Basics
+@_spi(SwiftPMInternal) import Basics
 import Foundation
 import PackageGraph
 import PackageLoading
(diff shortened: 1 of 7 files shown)


cmcgee1024:
@swift-ci please test

Sources/SwiftBuildSupport/PIFBuilder.swift
```diff
@@ -362,6 +362,35 @@ public final class PIFBuilder {
                                     env[key.rawValue] = value
                                 }
 
+                                if let libDir = try? buildParameters.toolchain.toolchainLibDir {
+                                    var libPathVar: String? = nil
+                                    var libPathEntry: String? = nil
+                                    var libPathSeparator: String? = nil
+#if os(macOS)
```
jakepetroules: This should remove Linux and add a terminating "else" case, since anything but Windows and Apple is ELF -- we want this to work for FreeBSD and other Unixes as well. But we can do even better...

Sources/SwiftBuildSupport/PIFBuilder.swift
```diff
@@ -362,6 +362,35 @@ public final class PIFBuilder {
                                     env[key.rawValue] = value
```
jakepetroules: You should switch `env` here over to the `Environment` abstraction from `Basics` rather than using a raw `[String: String]` -- this will handle the platform path list entry separator for you, and has convenience methods for adding a path to the path list, so you don't need to deal with all the split/join business.

`Environment` is also particularly important on Windows because the casing of PATH or Path can vary in different environments and using a raw dictionary can lead to reading/writing the environment incorrectly.

jakepetroules: I suggest also adding:

```swift
extension EnvironmentKey {
    package static var libraryPath: Self {
        #if os(Windows)
        path
        #elseif canImport(Darwin)
        "DYLD_LIBRARY_PATH"
        #else
        "LD_LIBRARY_PATH"
        #endif
    }
}
```

in EnvironmentKey.swift to make the platform-specific library path environment variable name easily reusable.

Sources/SwiftBuildSupport/PIFBuilder.swift
jakepetroules: In UserToolchain.swift's initializer, we call getTargetInfo, which runs `swiftc -print-target-info` and returns info like this:

```
{
  "compilerVersion": "Apple Swift version 6.1 (swift-6.1-RELEASE)",
  "target": {
    "triple": "arm64-apple-macosx15.0",
    "unversionedTriple": "arm64-apple-macosx",
    "moduleTriple": "arm64-apple-macos",
    "compatibilityLibraries": [ ],
    "librariesRequireRPath": false
  },
  "paths": {
    "runtimeLibraryPaths": [
      "/Users/jakepetroules/Library/Developer/Toolchains/swift-6.1-RELEASE.xctoolchain/usr/lib/swift/macosx",
      "/usr/lib/swift"
    ],
    "runtimeLibraryImportPaths": [
      "/Users/jakepetroules/Library/Developer/Toolchains/swift-6.1-RELEASE.xctoolchain/usr/lib/swift/macosx"
    ],
    "runtimeResourcePath": "/Users/jakepetroules/Library/Developer/Toolchains/swift-6.1-RELEASE.xctoolchain/usr/lib/swift"
  }
}
```

The `runtimeLibraryPaths` are what we want. Change the UserToolchain initializer to capture those and store them on the UserToolchain (and add that property on Toolchain).

jakepetroules: After all that, your code here in the PIF builder can become super simple and removes all the platform-specific considerations:

```swift
if let libDirs = try? buildParameters.toolchain.runtimeLibraryPaths {
    for libDir in libDirs {
        env.appendPath(key: .libraryPath, value: libDir.pathString)
    }
}
```

jakepetroules (changes requested):

cmcgee1024:
@swift-ci test Linux

cmcgee1024:
@swift-ci test Linux

cmcgee1024:
@swift-ci test Windows

Sources/SwiftBuildSupport/PIFBuilder.swift
cmcgee1024: I've made adjustments along the lines of these recommendations.

Sources/SwiftBuildSupport/PIFBuilder.swift
cmcgee1024: The CustomTask API still takes a `[String: String]` so while this now uses the Environment struct to do all of the path work, it converts to the dictionary before sending it off to that API.

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test Windows

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -182,7 +172,7 @@ final class PluginTests: XCTestCase {
             XCTAssert(stdout.contains("Build complete!"), "stdout:\n\(stdout)")
         }
 
-#if os(macOS) // See https://github.com/swiftlang/swift-package-manager/issues/8416 for errors running build tools on Linux
+#if os(macOS)
```
bkhouri, line 183: question: are these tests not supported on other platform, or they currently fail?  If they later, can we convert this Suite to Swift Testing and make use of the `withKnownIssue` API instead of skipping the test?  This would allow us to "re-enable" the test once the production code gets at par.

jakepetroules (approved):

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test Linux

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: It's the latter. I can try and convert these to Swift Testing in a future PR and then we will get notification when features start working, such as with Windows.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Windows

Issue #8081, opened 2024-10-29
`diagnose-api-breaking-changes` diagnoses API breaking change if `swift-argument-parser` is declared as a dependency


ahoppen:
When declaring `swift-argument-parser` as a dependency, `swift package diagnose-api-breaking-changes` will declare all public declarations having been removed. 

To reproduce, run

```bash
#!/usr/bin/env bash

cd $(mktemp -d)
echo $(realpath .)

echo '
// swift-tools-version: 6.0

import PackageDescription

let package = Package(
  name: "mypackage",
  products: [
    .library(name: "mypackage", targets: ["mypackage"])
  ],
  dependencies: [
    .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.2.2"),
  ],
  targets: [
    .target(name: "mypackage")
  ]
)
' > Package.swift

mkdir -p Sources/mypackage

echo 'public let myPublicString: String = "ABC"' > Sources/mypackage/mypackage.swift

git init
git add Package.swift Sources/mypackage/mypackage.swift
git commit -m "Initial commit"

swift package diagnose-api-breaking-changes main
```

Prints 
```
1 breaking change detected in mypackage:
  💔 API breakage: var myPublicString has been removed
```

Interestingly, no error is emitted if 
```
.package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.2.2"),
``` 
is removed

Oliver-Binns:
I am also seeing this issue, see failing build on this pull request: https://github.com/Oliver-Binns/Versioning/pull/14

ahoppen:
No, I am not aware of a workaround.

PR #8776, opened 2025-06-06
Do not include all of the build dir for API digester to avoid conflicting definitions.

yyvch:
Do not include all of the build dir for API digester to avoid conflicting definitions.

### Motivation:

Issue #8081 : when a package or its dependencies define a tool plugin, then diagnose-api-breaking-changes report false API breakage.

Investigation shown that diagnose-api-breaking-changes passes for analysis (with `-I`) all of the build dir, with all of the dependencies. When a tool plugin is in the dependency tree, then dependencies for the plugin itself are built with `.host` destination. If it happens that these modules are also built for `.target` destination, then `swift-api-digester` fails to analyze them, and public API declared there is printed as removed.

```
/Users/yy/work/pub/tst/.build/arm64-apple-macosx/debug/MyLib-tool.build/module.modulemap:1:8: error: redefinition of module 'MyLib'
module MyLib {
       ^
/Users/yy/work/pub/tst/.build/arm64-apple-macosx/debug/MyLib.build/module.modulemap:1:8: note: previously defined here
module MyLib {
       ^
Failed to load module: MyLib
```

### Modifications:

The goal for this PR is to reduce analyzed surface  from "all of the build directory" down to modules built for target.

### Result:

Before the change, example in Issue #8081 

```
1 breaking change detected in mypackage:
  💔 API breakage: var myPublicString has been removed
```

After the change:

```
No breaking changes detected in mypackage
```

### DIFF
--- a/Fixtures/Miscellaneous/APIDiff/WithPlugin/Package.swift
+++ b/Fixtures/Miscellaneous/APIDiff/WithPlugin/Package.swift
@@ -1,0 +1,16 @@
+// swift-tools-version: 6.1
+import PackageDescription
+
+let package = Package(
+    name: "package-with-plugin",
+    products: [.library(name: "PackageLib", targets: ["TargetLib"])],
+    targets: [
+        .target(name: "TargetLib"),
+        .executableTarget(name: "BuildTool", dependencies: ["TargetLib"]),
+        .plugin(
+            name: "BuildPlugin",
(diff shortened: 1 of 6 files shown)


yyvch:
@swift-ci please test

yyvch:
@swift-ci Please test Windows Platform

PR #8777, opened 2025-06-06
[DNM] cross repo self hosted testing

owenv:


owenv:
@swift-ci please test self hosted

PR #8780, opened 2025-06-09
Add SwiftBuildSupport to the primary libSwiftPM product

owenv:
This will make it accessible to clients like sourcekit-lsp


owenv:
@swift-ci test

PR #8785, opened 2025-06-09
[6.2] Allow prebuilt manifests to be updated as platforms are added.

dschaefer2:
One thing we're finding with snapshot toolchain builds is that not all platforms are completed at the same time. The prebuilts publishing job builds for the libraries using the latest available toolchains and then merges them together when signing. Havoc ensues when the toolchains aren't all the same version.

This change adds more metadata to the per zip json files including the swiftVersion used to build the library. The signing task will then see if a manifest already exists for that version and merge in the new platforms. If it doesn't it creates a new manifest from the other metadata in the zip json files. This takes out the use of the current swiftVersion when signing since it may not have been the version that created the libraries.

Cherry-pick from main https://github.com/swiftlang/swift-package-manager/pull/8784

### DIFF
--- a/Sources/Workspace/Workspace+Prebuilts.swift
+++ b/Sources/Workspace/Workspace+Prebuilts.swift
@@ -61,7 +61,7 @@ extension Workspace {
 
             public struct Artifact: Identifiable, Codable {
                 public let platform: Platform
-                public let checksum: String
+                public var checksum: String
 
                 public var id: Platform { platform }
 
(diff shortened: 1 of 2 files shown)


dschaefer2:
@swift-ci please test windows

dschaefer2:

dschaefer2:
Holding off putting this in right now. We want to use the current version one more time.

jakepetroules (approved):

Issue #8416, opened 2025-03-26
[Linux] (--build-system=swiftbuild): swift run fails to run executable


kcieplak:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

On Linux when running a binary using 'swift run' if the environment does not have a the swift runtime libraries in the LD_LIBRARY_PATH, the executable will fail to run.

```
.build/x86_64-unknown-linux-gnu/Products/Debug-linux/test_package: error while loading shared libraries: libswiftCore.so: cannot open shared object file: No such file or directory
```

### Expected behavior

built binary executes and prints greeting

### Actual behavior

`error while loading shared libraries: libswiftCore.so: cannot open shared object file: No such file or directory`

### Steps to reproduce

Remove any swift runtime libraries from **LD_LIBRARY_PATH** and **/usr/lib/swift/**

```
swift-package init --type executable
swift-run --build-system swiftbuild
```

kcieplak:
This issue occurs due to a difference in RUNPATH embedded in the binary.

### --build-system native

```
/test-swiftpm/swift-package-manager/.build/x86_64-unknown-linux-gnu/debug/swift-build --build-system native
/test-swiftpm/swift-package-manager/.build/x86_64-unknown-linux-gnu/debug/swift-run --build-system native
[1/1] Planning build
Building for debugging...
[1/1] Write swift-version--1CEC03321F1812F6.txt
Build of product 'test_package' complete! (0.63s)
Hello, world!

root@5ff52252beab:/test-swiftpm/test_package# readelf -d .build/x86_64-unknown-linux-gnu/debug/test_package

Dynamic section at offset 0x5d70 contains 31 entries:
  Tag        Type                         Name/Value

 0x000000000000001d (RUNPATH)            Library runpath: [/test-swiftpm/swift_toolchain/swift-DEVELOPMENT-SNAPSHOT-2024-10-08-a-ubuntu20.04/usr/lib/swift/linux:$ORIGIN]
```

### --build-system swiftbuild

```
/test-swiftpm/swift-package-manager/.build/x86_64-unknown-linux-gnu/debug/swift-build --build-system swiftbuild
[...]

PR #8793, opened 2025-06-09
Migrate PluginTests to Swift Testing

cmcgee1024:
Migrate the test suite to Swift Testing.

Explicitly mark tests that are enabled/disabled.

Add bug links to issues and radars for better tracking combined
withKnownIssues so that when bugs are fixed there is immediate
notification that the tests can be enabled.


cmcgee1024:
@swift-ci please test

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -20,117 +20,146 @@ import PackageModel
...
-
-final class PluginTests: XCTestCase {
+import Testing
+import Foundation
+
+@Suite
+final class PluginTests {
+    @Test(
+        .bug("https://github.com/swiftlang/swift-package-manager/issues/8602"),
+        .bug("https://github.com/swiftlang/swift-package-manager/issues/8791"),
```
bkhouri, line 30: question: is this meant to be issue #8416 per line 39 of the left file?

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -20,117 +20,146 @@ import PackageModel
...
-        // Try again with the Swift Build build system
-        try await fixture(name: "Miscellaneous/Plugins") { fixturePath in
-            let (stdout, _) = try await executeSwiftBuild(fixturePath.appending("MySourceGenPlugin"), configuration: .Debug, extraArgs: ["--product", "MyLocalTool", "--build-system", "swiftbuild"])
-            XCTAssert(stdout.contains("Build complete!"), "stdout:\n\(stdout)")
-        }
-#endif
+        try await withKnownIssue {
+            // Try again with the Swift Build build system
+            try await fixture(name: "Miscellaneous/Plugins") { fixturePath in
+                let (stdout, _) = try await executeSwiftBuild(fixturePath.appending("MySourceGenPlugin"), configuration: .Debug, extraArgs: ["--vv", "--product", "MyLocalTool", "--build-system", "swiftbuild"])
```
bkhouri: suggestion: instead of creating a separate code block in. the same "test case", consider using parameterized tests.

See #8714 as an example.

This comment applies to all applicable tests in suite.

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -20,117 +20,146 @@ import PackageModel
...
 
+    @Test(
+        .enabled(if: (try? UserToolchain.default)!.supportsSwiftConcurrency(), "skipping because test environment doesn't support concurrency")
+    )
     func testInternalExecutableAvailableOnlyToPlugin() async throws {
-        // Only run the test if the environment in which we're running actually supports Swift concurrency (which the plugin APIs require).
-        try XCTSkipIf(!UserToolchain.default.supportsSwiftConcurrency(), "skipping because test environment doesn't support concurrency")
-
         try await fixture(name: "Miscellaneous/Plugins") { fixturePath in
             await XCTAssertThrowsCommandExecutionError(try await executeSwiftBuild(fixturePath.appending("InvalidUseOfInternalPluginExecutable")), "Illegally used internal executable") { error in
```
bkhouri: Issue: (blocking): This test has a mixture of XCTest and Swift Testing.  We need convert this to Swift Testing.

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -143,123 +172,131 @@ final class PluginTests: XCTestCase {
...
         // Try again with Swift Build build system
         try await fixture(name: "Miscellaneous/Plugins") { fixturePath in
             let (stdout, _) = try await executeSwiftBuild(fixturePath.appending("SandboxTesterPlugin"), configuration: .Debug, extraArgs: ["--build-system", "swiftbuild", "--product", "MyLocalTool"])
-            XCTAssert(stdout.contains("Build complete!"), "stdout:\n\(stdout)")
+            #expect(stdout.contains("Build complete!"), "stdout:\n\(stdout)")
         }
     }
 
+    @Test(
+        .enabled(if: (try? UserToolchain.default)!.supportsSwiftConcurrency(), "skipping because test environment doesn't support concurrency"),
```
bkhouri, line 301: suggestion (non-blocking): do all the tests in this suite required SwiftConcurrency?  if so, consider adding the `.enabled(if:)` trait to the `@Suite`

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -643,62 +676,63 @@ final class PluginTests: XCTestCase {
...
+                        try #require(!success, "expected command to fail, but it succeeded")
                     }
                     else {
-                        XCTAssertTrue(success, "expected command to succeed, but it failed", file: file, line: line)
+                        try #require(success, "expected command to succeed, but it failed")
                     }
                 }
                 catch {
-                    XCTFail("error \(String(describing: error))", file: file, line: line)
+                    try #require(true == false, "error \(String(describing: error)) at \(file) and \(line)")
```
bkhouri: issue: We should use `Issue.record("error \(String(describing: error)) at \(sourceLocation)")" here instead

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -590,21 +627,17 @@ final class PluginTests: XCTestCase {
                 line: UInt = #line,
```
bkhouri: issue (blocking): I believe we should be using `sourceLocation: SourceLocation = #_sourceLocation` in Swift Testing instead of the `file: StaticString = #file, #line: UInt = #line`

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -643,62 +676,63 @@ final class PluginTests: XCTestCase {
                       )
                     }
                     if expectFailure {
-                        XCTAssertFalse(success, "expected command to fail, but it succeeded", file: file, line: line)
+                        try #require(!success, "expected command to fail, but it succeeded")
```
bkhouri: issue (possibly-blocking): if the expectation fails, this will stop all remaining test case execution.  Is this what we want?

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -643,62 +676,63 @@ final class PluginTests: XCTestCase {
...
                     }
                     else {
-                        XCTAssertTrue(success, "expected command to succeed, but it failed", file: file, line: line)
+                        try #require(success, "expected command to succeed, but it failed")
```
bkhouri: issue (possibly-blocking): if the expectation fails, this will stop all remaining test case execution.  Is this what we want?

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -727,17 +761,16 @@ final class PluginTests: XCTestCase {
...
             }
         }
     }
 
+    @Test(
+        .enabled(if: (try? UserToolchain.default)!.supportsSwiftConcurrency(), "skipping because test environment doesn't support concurrency"),
+        .disabled(if: ProcessInfo.hostOperatingSystem == .windows, "This hangs intermittently on windows in CI")
+    )
     func testCommandPluginCancellation() async throws {
-        try XCTSkipOnWindows(because: "This hangs intermittently on windows in CI", skipSelfHostedCI: true)
```
bkhouri, line 736: I noticed we did not add a `withKnownIssue(Comment:isIntermittent)` call here,  Does this mean the test runs reliably in a self hosted windows pipeline?

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -925,15 +956,15 @@ final class PluginTests: XCTestCase {
                         do {
                             try scriptRunner.cancel(deadline: .now() + .seconds(5))
                         } catch {
-                            XCTFail("Cancelling script runner should not fail: \(error)")
+                            fatalError("Cancelling script runner should not fail: \(error)")
```
bkhouri: issue (blocking): Can we call `Issue.record("Cancelling script runner should not fail: \(error)")` to record this as a test failure instead of aborting the entire execution?

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -925,15 +956,15 @@ final class PluginTests: XCTestCase {
...
                         }
                     }
                 }
                 group.addTask {
                     do {
                         try await Task.sleep(nanoseconds: UInt64(DispatchTimeInterval.seconds(3).nanoseconds()!))
                     } catch {
-                        XCTFail("The plugin should not finish within 3 seconds")
+                        try #require(true == false, "The plugin should not finish within 3 seconds")
```
bkhouri: issue (blocking)" replace the `try #require(...)` with `Issue.record("The plugin should not finish within 3 seconds")`.

See https://developer.apple.com/documentation/testing/migratingfromxctest

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -945,15 +976,16 @@ final class PluginTests: XCTestCase {
                 delegateQueue.sync {
                     pid = delegate.parsedProcessIdentifier
                 }
-                guard let pid else {
-                    throw XCTSkip("skipping test because no pid was received from the plugin; being investigated as rdar://88792829\n\(delegate.diagnostics.description)")
+                guard let pid = pid else {
```
bkhouri, line 1018: issue (possibly-blocking): if we want to fail the test, we should invoke `try #require(pid, "Test failed because no pid was received from the plugin; being investigated as rdar://88792829\n\(delegate.diagnostics.description)")`


Can we add `.bug("rdar://88792829")` trait to the test, and preferably reference a GitHub issue.

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -1201,140 +1234,152 @@ final class PluginTests: XCTestCase {
             // Check that the build succeeds if we disable the sandbox.
             try await fixture(name: "Miscellaneous/Plugins/SandboxViolatingBuildToolPluginCommands") { path in
                 let (stdout, stderr) = try await executeSwiftBuild(path.appending("MyLibrary"), configuration: .Debug, extraArgs: ["--build-system", buildSystem, "--disable-sandbox"])
```
bkhouri: suggestion: instead of adding `"--build-system", buildSystem` to the `extraArgs`, consider setting the `buildSystem` argument to `executeSwiftBuild`, which is of type `BuildSystemProvider.Kind`

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -1201,140 +1234,152 @@ final class PluginTests: XCTestCase {
...
-        try await fixture(name: "Miscellaneous/Plugins/MySourceGenPluginUsingURLBasedAPI") { fixturePath in
-            let (stdout, _) = try await executeSwiftBuild(fixturePath, configuration: .Debug)
-            XCTAssert(stdout.contains("Build complete!"), "stdout:\n\(stdout)")
-        }
-#if os(macOS)
-        // Try again with the Swift Build build system
-        try await fixture(name: "Miscellaneous/Plugins/MySourceGenPluginUsingURLBasedAPI") { fixturePath in
-            let (stdout, _) = try await executeSwiftBuild(fixturePath, configuration: .Debug, extraArgs: ["--build-system", "swiftbuild"])
-            XCTAssert(stdout.contains("Build complete!"), "stdout:\n\(stdout)")
+            try await withKnownIssue {
```
bkhouri: issue (possibly-blocking): The XCTest version of this does not have this being skipped on linux.  So why is this currently a known issue in the Swift Testing migration?

bkhouri (changes requested):
There are still references to custom `XCTAssertion` calls, which mixes XCTest and Swift Testing. Not to mention I believe the use of `#file` and `#line` should be replaced with `#_sourceLocation` based on the some Swift Testing trait documentation: https://developer.apple.com/documentation/testing/trait/enabled(if:_:sourcelocation:).

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: No, the comment was stale. There have been PR's that make build tools work correctly with Linux. #8416 remains for other cases where `swift run` can't run a target on Linux platforms.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: As far as I'm aware the parameterized tests don't permit skipping certain parameter combinations, or have separate parts that have different known issue when conditions. I expect there to be too many parameters to make the parameterization worthwhile.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: Thanks, do you know if anyone has make equivalents for Swift Testing for these utilities in `_InternalTestSupport`? There's significant logic in these utility methods, and not entirely obvious how to convert them to the new testing API.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: All but one...

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: Thanks, I couldn't find the API to raise an issue directly.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: Perhaps it does, but then why does this test run in a smoke test pipeline at all? It's probably best that we skip it on Windows if it hangs (not just fails) builds there.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: Thanks, I couldn't find this API to directly report an issue and fail the test. This is fixed now.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: The code that's written here before appears to want to skip the test in middle in this case, not fail it. I'm not sure how to do that with Swift Test API, other than to just return. How would one print this information out without just printing it?

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: I've raised a new GH issue (linked with the .bug trait at the top) for this one. The existing code was guarded with `#if os(macOS)`, and I just discovered that this fails on Linux for a novel reason, so that's why there's a new `withKnownIssue` and bug link for tracking purposes.

cmcgee1024:
@swift-ci please test

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: No, this was an oversight. Thanks, I've converted this to an expect instead.

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

Tests/FunctionalTests/PluginTests.swift
bkhouri: The `withKnownIssue` can accept a `when:` argument.

e.g.:

```
withKnownIssue {
   // test code
} when: {
 buildSystem == .swiftbuild
}
```

See the [known issues](https://developer.apple.com/documentation/testing/known-issues) documentation.

Regarding skipping: It is my opinion that a test should be skipped if, and only if, the test is not applicable.  Otherwise, we ought to use `withKnownIssue` to report the currently failing behaviour.

Tests/FunctionalTests/PluginTests.swift
bkhouri: Have a look at `BuildCommandTests.swift` here for a possible conversion option:

https://github.com/swiftlang/swift-package-manager/pull/8714/files#diff-b1845cad8f64e37500b06818ff71629726c14cf5a8d34add6ef3a9d38236cfe5

In addition, has a look a the [Record issues](https://developer.apple.com/documentation/testing/migratingfromxctest#Record-issues) in the [Migrate a test from XCTest](https://developer.apple.com/documentation/testing/migratingfromxctest).

Tests/FunctionalTests/PluginTests.swift
bkhouri: @grynspan or @stmontgomery might have ideas on how to handle this.

bkhouri:
@swift-ci test windows

Tests/FunctionalTests/PluginTests.swift
stmontgomery: Consider using `withKnownIssue` and use `try #require` in its body closure to unwrap the `pid` and use it in the subsequent code. You could use a `matching:` closure on `withKnownIssue` to only match the specific expectation which is attempting to unwrap `pid`.

Tests/FunctionalTests/PluginTests.swift
bkhouri: The Windows Platform Test (smoke test) runs in a different environment than the windows self-hosted pipeline, which runs in a docker container.  These environmental differences sometimes have differing test execution, for some reason.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: Changed this to an `#expect` so that the test continues.

cmcgee1024:
@swift-ci test Linux

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Windows

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -20,246 +20,297 @@ import PackageModel
...
-        try XCTSkipIf(!UserToolchain.default.supportsSwiftConcurrency(), "skipping because test environment doesn't support concurrency")
-
-        try await fixture(name: "Miscellaneous/Plugins") { fixturePath in
-            let (stdout, _) = try await executeSwiftBuild(fixturePath.appending("MySourceGenPlugin"), configuration: .Debug, extraArgs: ["--product", "MyOtherLocalTool"])
-            XCTAssert(stdout.contains("Compiling MyOtherLocalTool bar.swift"), "stdout:\n\(stdout)")
-            XCTAssert(stdout.contains("Compiling MyOtherLocalTool baz.swift"), "stdout:\n\(stdout)")
-            XCTAssert(stdout.contains("Linking MyOtherLocalTool"), "stdout:\n\(stdout)")
-            XCTAssert(stdout.contains("Build of product 'MyOtherLocalTool' complete!"), "stdout:\n\(stdout)")
-        }
+        try await withKnownIssue {
```
bkhouri, line 114: nit (non-blocking, if-updating-pr): Can we wrap the `withKnownIssue` around each build system? This should give more fine grained indication if this test will be supported on windows for native and SwiftBuild build system.

This comment applies to all applicable test cases.

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -20,246 +20,297 @@ import PackageModel
...
+
+            #expect(
+                    stderr.contains("product 'PluginExecutable' required by package 'invaliduseofinternalpluginexecutable' target 'RootTarget' not found in package 'PluginWithInternalExecutable'."), "stderr:\n\(stderr)"
+            )
         }
 
         try await fixture(name: "Miscellaneous/Plugins") { fixturePath in
-            await XCTAssertThrowsCommandExecutionError(try await executeSwiftBuild(fixturePath.appending("InvalidUseOfInternalPluginExecutable")), "Illegally used internal executable"
-) { error in
+            let error =  try await #require(throws: SwiftPMError.self, "Illegally used internal executable") {
```
bkhouri, line 184: question: Are we able to add an expectation to verify we failed in the expected way?

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -20,246 +20,297 @@ import PackageModel
...
+                try await executeSwiftBuild(fixturePath.appending("InvalidUseOfInternalPluginExecutable"))
```
bkhouri, line 185: issue (possibly-blocking): This running the same action as lie 162.   Did we intend to build this using SwiftBuild build system?  If so, can we update this test?   It's ok to do this in a follow-up PR.

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -20,246 +20,297 @@ import PackageModel
...
 
-#if os(macOS)
-        // Try again with the Swift Build build system
-        try await fixture(name: "Miscellaneous/Plugins") { fixturePath in
-            let (stdout, _) = try await executeSwiftBuild(fixturePath.appending("MyBuildToolPluginDependencies"), extraArgs: ["--build-system", "swiftbuild"])
-            XCTAssert(stdout.contains("Build complete!"), "stdout:\n\(stdout)")
+            try await withKnownIssue {
+                // Try again with the Swift Build build system
+                try await fixture(name: "Miscellaneous/Plugins") { fixturePath in
+                    let (stdout, _) = try await executeSwiftBuild(fixturePath.appending("MyBuildToolPluginDependencies"), extraArgs: ["--vv", "--build-system", "swiftbuild"])
```
bkhouri: question: does this test now need the `--vv` argument?

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -643,62 +698,63 @@ final class PluginTests: XCTestCase {
                       )
                     }
                     if expectFailure {
-                        XCTAssertFalse(success, "expected command to fail, but it succeeded", file: file, line: line)
+                        #expect(!success, "expected command to fail, but it succeeded")
```
bkhouri, line 717: question: should we pass through the `sourceLocation` argument to the expectation so the failure has a more meaningful file and line number?

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -643,62 +698,63 @@ final class PluginTests: XCTestCase {
...
                     }
                     else {
-                        XCTAssertTrue(success, "expected command to succeed, but it failed", file: file, line: line)
+                        #expect(success, "expected command to succeed, but it failed")
```
bkhouri: question: should we pass through the `sourceLocation` argument to the expectation so the failure has a more meaningful file and line number?

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -643,62 +698,63 @@ final class PluginTests: XCTestCase {
...
                     }
                 }
                 catch {
-                    XCTFail("error \(String(describing: error))", file: file, line: line)
+                    Issue.record("error \(String(describing: error)) at \(sourceLocation)")
```
bkhouri: suggestion (non-blocking): Instead of include the source location in failure message, set the `sourceLocation` argument.

```suggestion
                    Issue.record("error \(String(describing: error))", sourceLocation: sourceLocation)`
```


See https://developer.apple.com/documentation/testing/issue/record(_:_:sourcelocation:)

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -925,15 +979,15 @@ final class PluginTests: XCTestCase {
                         do {
```
bkhouri, line 995: suggestion (non-blocking): this do block could be replace with 

```swift
#expect(throws: Never.self) {
    try scriptRunner.cancel(deadline: .now() + .seconds(5))
}
```

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -1147,194 +1203,207 @@ final class PluginTests: XCTestCase {
...
-            XCTAssert(stderr.contains("Creating foo.swift from foo.dat"), "stdout:\n\(stdout)\nstderr:\n\(stderr)")
-            XCTAssert(stdout.contains("Build complete!"), "stdout:\n\(stdout)\nstderr:\n\(stderr)")
-        }
-#endif
+            try await withKnownIssue {
+                try await fixture(name: "Miscellaneous/Plugins") { fixturePath in
+                    let (stdout, stderr) = try await executeSwiftBuild(
+                        fixturePath.appending(component: "MySourceGenPlugin"),
+                        configuration: .Debug,
+                        extraArgs: ["--vv", "--product", "MyLocalTool", "-Xbuild-tools-swiftc", "-DUSE_CREATE", "--build-system", "swiftbuild"]
```
bkhouri: question: is `--vv` required to to pass this test?

Sources/_InternalTestSupport/Observability.swift
```diff
@@ -103,6 +103,18 @@ public func XCTAssertNoDiagnostics(
...
+public func expectNoDiagnostics(
+    _ diagnostics: [Basics.Diagnostic],
+    problemsOnly: Bool = true,
+    sourceLocation: SourceLocation = #_sourceLocation
+) {
+    let diagnostics = problemsOnly ? diagnostics.filter { $0.severity >= .warning } : diagnostics
+    if diagnostics.isEmpty { return }
+
+    let description = diagnostics.map { "- " + $0.description }.joined(separator: "\n")
+    Issue.record("Found unexpected diagnostics: \n\(description) \(sourceLocation.fileName):\(sourceLocation.line)")
```
bkhouri: issue (if-patch-updated): instead of putting the source location in the error message, can we call `Issue.record` with the `sourceLocation` argument set.  That's, Swift Testing will accurately report the filename and line number.

```suggestion
    Issue.record("Found unexpected diagnostics: \n\(description)", sourceLocation: sourceLocation)
```

See https://developer.apple.com/documentation/testing/issue/record(_:sourcelocation:)

bkhouri (approved):
Nothing blocking, however there are changes that should be addressed. Addressing them in a follow-up PR is acceptable, unless the branch will be updated.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: The "--vv" will give us enough information from swift build for problem diagnosis for things like linker failures. Otherwise, it's just a single error message indicating that linking failed, which isn't very helpful.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: No, it's just that without the extra verbose logging, when the test does fail it doesn't yield very useful information for things like linking failures.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: This is just attempting to faithfully duplicate the logic from the XCTest function here. It's expecting that the SwiftPMError is thrown, and that the error contains certain details taken from the XCTAssertThrowsCommandExecutionError.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: This test already had these two cases, one that checks stderr's contents. It didn't cover swiftbuild build system. This is just faithfully converting it to Swift Testing as-is.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: I've removed the verbose logging so that it doesn't fill up the logs. With linker errors, one will need a reproducing environment to be able to exfiltrate the linker arguments.

Tests/FunctionalTests/PluginTests.swift
cmcgee1024: I've removed the verbose logging so that it doesn't fill up the logs. With linker errors, one will need a reproducing environment to be able to exfiltrate the linker arguments.

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test macOS

PR #8799, opened 2025-06-10
Test: disable environment test in self hosted

bkhouri:
There is a test which runs in self hosted pipelines, which, when run in parallel, is prone to impacting other tests as it modifies the `PATH` environment variable.

Modify the test trait to also disable the tests when it's run in self hosted environment.


bkhouri:
@swift-ci test

bkhouri:
@swift-ci test macOS

PR #8717, opened 2025-05-23
Adding --verbose to swift build triggers a full rebuild instead of incremental

plemarquand:
Normalize any build parameters whose modifications do not need to cause a recompilation. For instance, building with or without `--verbose` should not cause a full rebuild.

Issue: #8690

PR #8803, opened 2025-06-11
Prevent unnecessary rebuilds when modifying swiftc arguments

plemarquand:
Many arguments for the compiler are marked as `doesNotAffectIncrementalBuild` in `swift-driver`. Use the argument hashing in swift-driver to hash the swift compiler flags and incorporate that in to the build hash signature used to determine if SwiftPM should do a full rebuild.

This prevents unnecssary rebuilds when adding or changing arguments like `-Xswiftc -diagnostic-style=llvm`. For a list of arguments that no longer affect rebuilds, search for `doesNotAffectIncrementalBuild` in https://github.com/swiftlang/swift-driver/blob/main/Sources/SwiftOptions/Options.swift.

This builds on the work introduced in #8717 which avoided unnecessary rebuilds when SwiftPM arguments changed.


plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

MahdiBM:
@plemarquand This is huge in certain situations. Does this also cover env vars, or only command line arguments?

plemarquand:
@MahdiBM This patch just covers `-Xswiftc` arguments. I dont think we can do something similar with env vars since we can't know what downstream tools (LLVM, swiftc, etc...) are going to use today or in future.

owenv (approved):
LGTM, but it would be nice to have a test for this if we can

plemarquand:
@owenv I agree. I'm going to wait for https://github.com/swiftlang/swift-package-manager/pull/8714 to land to avoid merge conflicts.

dschaefer2:
Great find. Thanks @plemarquand !

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci please test

plemarquand:
@owenv @bkhouri I've added a few tests. They verify that no rebuild has occurred by checking the mtime of built artifacts. Let me know if there is an issue with this approach.

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1125,7 +1125,96 @@ struct BuildCommandTestCases {
...
+                            .appending("ExecutableNew")
+                    )
+                case .swiftbuild:
+                    let triple = try UserToolchain.default.targetTriple.withoutVersion().tripleString
+                    return (
+                        .swiftbuild,
+                        try RelativePath(validating: ".build")
+                            .appending(triple)
+                            .appending("Products")
+                            .appending("Debug")
```
owenv: I'm pretty sure this path component will have a platform suffix everywhere except macOS, otherwise lgtm

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test windows

switch_branches.sh
bkhouri, line 1: Question: did we commit this shell script in error?

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1125,7 +1125,112 @@ struct BuildCommandTestCases {
...
-    
+    private static func buildSystemAndOutputLocation() throws -> [(BuildSystemProvider.Kind, Basics.RelativePath)] {
+        return try SupportedBuildSystemOnPlatform.map { buildSystem in
+            let triple = try UserToolchain.default.targetTriple.withoutVersion()
+            let debugFolder = triple.platformName() == "macosx" ? "Debug" : "Debug-\(triple.platformName() ?? "unknown")"
+            let base = try RelativePath(validating: ".build")
+            switch buildSystem {
+                case .xcode:
+                    let path = base
+                            .appending("apple")
```
bkhouri: FYI: we can use `buildSystem.binPathSuffixes(for: .debug)` to get part paths.

See https://github.com/swiftlang/swift-package-manager/blob/main/Sources/_InternalTestSupport/BuildSystemProvider%2BConfiguration.swift.   Some tests make use of this to generalize the output directory. For example, [this one](https://github.com/swiftlang/swift-package-manager/blob/86810fecf06461f9c9d354be82a95b8a73c318cb/Tests/CommandsTests/BuildCommandTests.swift#L309).

bkhouri (commented):
Nothing blocking, but the tests looks good to me.

bkhouri:
@swift-ci test self hosted

switch_branches.sh
plemarquand: Good catch, was debugging a different issue. I've removed it.

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

PR #8806, opened 2025-06-11
[Commands] Fixes a few issues with `swift package migrate` command

xedin:
### Motivation:

Fixes for the following problems:

  - Sometimes if the previous build was unsuccessful subsequent `swift package migrate` invocation fails with `error: did not compute a build plan yet`.
  - Implicitly generated targets (currently only test targets) are included in migration which results in extraneous warnings and failures (i.e. to update the manifest).

### Modifications:

- Update `Migrate` command's `createBuildSystem` API to avoid build manifest caching. That ensures that features in migration mode are never persistent and requires a fresh build plan which avoid issues with previous builds.
- Make it possible to mark a `Module` as `implicit` i.e. when it's a synthesized/discovered test target or a system module.
- Update `Migrate` command to filter `implicit` modules when no targets are specified by the user.

### Result:

`swift package migrate` is made more stable with fewer unactionable warnings and errors.

Resolves: rdar://152689053
Resolves: rdar://152687084

### DIFF
--- a/Sources/Commands/PackageCommands/Migrate.swift
+++ b/Sources/Commands/PackageCommands/Migrate.swift
@@ -114,8 +114,12 @@ extension SwiftPackageCommand {
             } else {
                 let graph = try await buildSystem.getPackageGraph()
                 for buildDescription in buildPlan.buildModules
-                    where graph.isRootPackage(buildDescription.package) && buildDescription.module.type != .plugin
+                    where graph.isRootPackage(buildDescription.package)
                 {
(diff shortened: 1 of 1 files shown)


bnbarham (approved):

xedin:
@swift-ci please test

jakepetroules (approved):

xedin:
@swift-ci please test

xedin:

xedin:

xedin:

PR #8807, opened 2025-06-11
Support using a Swift file as a build tool plugin executable

plemarquand:
A Swift file can be executed directly in the shell. Typically the file begins with a `#/usr/bin/swift` or an equivalent `xcrun` incantation, and then simply running `swift MyFile.swift` will automatically compile and execute it.

Build tool plugins often leverage Swift, but to do so they add two targets, a `.plugin` and an `.executableTarget`. The executable target builds a binary that the plugin tool then calls during a build.

Often this is more heavyweight than needed. By passing the path to a Swift file to the `executable` argument of the `.buildCommand` returned from your build tool plugin, you could bypass the need to create a separate executable target and just execute the file directly.

However, because SwiftPM executes build tool plugins within a sandbox the sandbox script needs to be ammended with the clang module cache temporary directory so that the swift compiler can write the modules it needs to execute the file.

Ammend the sandbox script to add write permissions to the root tmp directory that LLVM uses to write clang modules. With this addition you can now use a .swift file as a build tool plugin executable.

rdar://152874736

### DIFF
--- a/Fixtures/Miscellaneous/Plugins/SwiftFilePlugin/SimpleSwiftScript.swift
+++ b/Fixtures/Miscellaneous/Plugins/SwiftFilePlugin/SimpleSwiftScript.swift
@@ -1,0 +1,3 @@
+#!/usr/bin/swift
+
+print("Hello, Build Tool Plugin!")
--- a/Fixtures/Miscellaneous/Plugins/SwiftFilePlugin/SimpleSwiftScript.swift
+++ b/Fixtures/Miscellaneous/Plugins/SwiftFilePlugin/SimpleSwiftScript.swift
@@ -1,3 +1,3 @@
-#!/usr/bin/swift
+#!/usr/bin/env swift
 
 print("Hello, Build Tool Plugin!")
--- a/Tests/FunctionalTests/PluginTests.swift
+++ b/Tests/FunctionalTests/PluginTests.swift
@@ -1206,12 +1206,14 @@ final class PluginTests: XCTestCase {
         }
     }
 
+#if os(macOS)
     func testBuildToolPluginSwiftFileExecutable() async throws {
         try await fixture(name: "Miscellaneous/Plugins") { fixturePath in
             let (stdout, _) = try await executeSwiftBuild(fixturePath.appending("SwiftFilePlugin"), configuration: .Debug)
             XCTAssertTrue(stdout.contains("Hello, Build Tool Plugin!"), "stdout:\n\(stdout)")
         }
     }
+#endif
 
     func testTransitivePluginOnlyDependency() async throws {
         // Only run the test if the environment in which we're running actually supports Swift concurrency (which the plugin APIs require).
--- a/Tests/FunctionalTests/PluginTests.swift
+++ b/Tests/FunctionalTests/PluginTests.swift
@@ -1206,6 +1206,13 @@ final class PluginTests: XCTestCase {
         }
     }
 
+    func testBuildToolPluginSwiftFileExecutable() async throws {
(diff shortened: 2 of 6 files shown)


plemarquand:
@swift-ci please test

owenv (approved):

plemarquand:
@swift-ci please test

Fixtures/Miscellaneous/Plugins/SwiftFilePlugin/SimpleSwiftScript.swift
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/swift
```
MahdiBM: does this support pointing to any executable or is it locked to swift? just checking ... one might prefer to use e.g. bash sometimes.

plemarquand: You could use a bash script as a build tool plugin executable and I believe it would work today, without this patch, yes.

plemarquand:
@swift-ci please test macOS

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci please test

Fixtures/Miscellaneous/Plugins/SwiftFilePlugin/SimpleSwiftScript.swift
cmcgee1024: question: What happens if on the developer machine, or maybe a CI system somewhere if swift isn't installed in `/usr/bin`? For example, it can be a swiftly installed toolchain.

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -1206,6 +1206,15 @@ final class PluginTests: XCTestCase {
         }
     }
 
+#if os(macOS)
```
cmcgee1024, line 1209: question: Could this work on Linux too?

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -1206,6 +1206,15 @@ final class PluginTests: XCTestCase {
...
+    func testBuildToolPluginSwiftFileExecutable() async throws {
+        try await fixture(name: "Miscellaneous/Plugins") { fixturePath in
+            let (stdout, _) = try await executeSwiftBuild(fixturePath.appending("SwiftFilePlugin"), configuration: .Debug)
```
cmcgee1024: question: Does this work with the swiftbuild build system too? Can a test be added for it?

Fixtures/Miscellaneous/Plugins/SwiftFilePlugin/SimpleSwiftScript.swift
dschaefer2: Probably should be
`#!/usr/bin/env swift`

Tests/FunctionalTests/PluginTests.swift
plemarquand: I think this should already work on linux, as there is no sandbox there to contend with.

Tests/FunctionalTests/PluginTests.swift
plemarquand: Good call, I've expanded this test to try both build systems (will be parameterized when moved to swift-testing) and added the one line required to enable swift-build support.

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

PR #8812, opened 2025-06-12
[6.2] Fixes a few issues with `swift package {migrate, add-setting}` commands

xedin:
Cherry-pick of https://github.com/swiftlang/swift-package-manager/pull/8800, https://github.com/swiftlang/swift-package-manager/pull/8806

### Motivation:

Fixes for the following problems:

  - Sometimes if the previous build was unsuccessful subsequent `swift package migrate` invocation fails with `error: did not compute a build plan yet`.
  - Implicitly generated targets (currently only test targets) are included in migration which results in extraneous warnings and failures (i.e. to update the manifest).

The inconsistency in argument formatting came up as part of SE-0486 review. This is the first step in a direction to make command formatting more consistent.

- `swift package migrate`:
  - Rename `--targets` to `--target`
  - Change `--target` and `--to-feature` to accept space separated argument lists

- `swift package -add-setting`:
   - Switch to `.upToNextOption` formatting. Instead of `--swift A=B --swift C=D` switch to `--swift A=B C=D`.

### Modifications:

- Update `Migrate` command's `createBuildSystem` API to avoid build manifest caching. That ensures that features in migration mode are never persistent and requires a fresh build plan which avoid issues with previous builds.
- Make it possible to mark a `Module` as `implicit` i.e. when it's a synthesized/discovered test target or a system module.
- Update `Migrate` command to filter `implicit` modules when no targets are specified by the user.
- Update `MigrateOptions` to use `.upToNextOption` for targets and features.
- Update `SwiftPackageCommand`'s `_swiftSettings` option to use `.upToNextOption`

### Result:

`swift package migrate` is made more stable with fewer unactionable warnings and errors and the command line is now shorter and has consistent formatting.

Resolves: rdar://152687586
Resolves: rdar://152689053
Resolves: rdar://152687084

### DIFF
--- a/Sources/Build/BuildPlan/BuildPlan+Test.swift
+++ b/Sources/Build/BuildPlan/BuildPlan+Test.swift
@@ -102,7 +102,8 @@ extension BuildPlan {
                     dependencies: testProduct.underlying.modules.map { .module($0, conditions: []) },
                     packageAccess: true, // test target is allowed access to package decls by default
                     testDiscoverySrc: Sources(paths: discoveryPaths, root: discoveryDerivedDir),
-                    buildSettings: discoveryBuildSettings
+                    buildSettings: discoveryBuildSettings,
+                    implicit: true
                 )
                 let discoveryResolvedModule = ResolvedModule(
                     packageIdentity: testProduct.packageIdentity,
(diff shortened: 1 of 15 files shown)


xedin:
@swift-ci please test

bnbarham (approved):

PR #8818, opened 2025-06-13
build: avoid duplicating all the flags for FreeBSD

compnerd:
This simplifies the build logic and properly adds the linker search path to the build when building for FreeBSD.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,10 +53,9 @@ find_package(Foundation QUIET)
 find_package(SQLite3 REQUIRED)
 
 # Enable `package` modifier for the whole package.
+add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:-package-name;SwiftPM>")
 if(CMAKE_SYSTEM_NAME STREQUAL FreeBSD)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

CMakeLists.txt
```diff
@@ -53,10 +53,9 @@ find_package(Foundation QUIET)
 find_package(SQLite3 REQUIRED)
 
 # Enable `package` modifier for the whole package.
+add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:-package-name;SwiftPM>")
```
etcwilde, line 56: Do we want to `$<$<COMPILE_LANGUAGE:Swift>:SHELL:-package-name SwiftPM>` the generator expression in case someone somewhere else tries to throw in a `-package-name <blah>`? Don't want CMake de-duping either part of this.

etcwilde (approved):

CMakeLists.txt
compnerd: I think that is a good idea, but we should do that as a follow up (I will be renaming the package name, I can throw it into that change).

compnerd:

jakepetroules (approved):

compnerd:

compnerd:

compnerd:

PR #8842, opened 2025-06-19
SwiftBuildSystem: Expose public API for reuse by  SourceKit-LSP/BSP

owenv:
Reorganize some of the SwiftBuildSystem code so we can share the implementation of
- session creation
- writing PIF
- build request construction

There should be no functional change to SwiftPM

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -57,6 +57,26 @@ func withService(
     await service.close()
 }
 
+public func createSession(
+    service: SWBBuildService,
+    name: String,
+    toolchainPath: Basics.AbsolutePath,
+    packageManagerResourcesDirectory: Basics.AbsolutePath?
+) async throws-> (SWBBuildServiceSession, [SwiftBuildMessage.DiagnosticInfo]) {
+    // SWIFT_EXEC and SWIFT_EXEC_MANIFEST may need to be overridden in debug scenarios in order to pick up Open Source toolchains
+    let sessionResult = if toolchainPath.components.contains(where: { $0.hasSuffix(".xctoolchain") }) {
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test

dschaefer2 (approved):

owenv:
@swift-ci test Windows

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -57,44 +57,52 @@ func withService(
     await service.close()
 }
 
-func withSession(
+public func createSession(
```
bkhouri, line 60: issue (blocking): if we are exposing a public API, we need to add automated tests ensuring we don't break the "contract".

would you mind adding tests that verify each public API in isolation. This would provide an indication of whether the API makes sense to the caller.

This comment applies to all public API where it makes sense.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -700,6 +707,16 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
         }
     }
 
+    public func writePIF(buildParameters: BuildParameters) async throws {
```
bkhouri, line 710: suggestion (blocking): looking at the API in isolation without looking at the implementation, there is nothing that provide where the PIF is being written to.  As consumer of the API, how can we get the location of where the PIF was written?

The API should be update to either accept a path where to write the PIF or maybe a return either (1) the path where the PIF, or (2) the pif contents.

jakepetroules: The location should be opaque to the client (e.g. maybe in future it's not written to the filesystem at all), perhaps the API could be called something more general, like "constructPIF"?

PR #8843, opened 2025-06-19
Delete wrong doc comment for Package.name

ole:
Deletes a sentence from the doc comments for `Package.name` that wrongly talks about the possibility of `name` being `nil`.

### Motivation:

`Package.name` is a non-Optional property so it can never be `nil`, so [the current documentation for it](https://developer.apple.com/documentation/packagedescription/package/name) is factually wrong:

> The name of the Swift package.
>
> **Discussion**
>
> If the name of the package is `nil`, Swift Package Manager deduces the name of the package using its Git URL.

### Modifications:

- Deleted part of a doc comment.

### DIFF
--- a/Sources/PackageDescription/PackageDescription.swift
+++ b/Sources/PackageDescription/PackageDescription.swift
@@ -72,9 +72,6 @@
 /// package's manifest or losing access to existing packages.
 public final class Package {
     /// The name of the Swift package.
-    ///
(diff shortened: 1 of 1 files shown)


plemarquand (approved):

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test linux

PR #8852, opened 2025-06-19
[WIP] Fix for trait-guarded deps being included in resolution

bripeticca:
Trait-guarded dependencies were still being considered during dependency resolution, when they should be excluded if they aren't being used in any other scenario.

### Modifications:

Since we have the `--experimental-prune-unused-dependencies` feature behind an experimental flag, we'll now consider an alternate path that will prune trait-guarded package dependencies from the dependency graph **_if and only if_** said dependency is not used in any other unguarded context.

### Result:

Trait-guarded dependencies are excluded from dependency resolution.

### DIFF
--- a/Tests/FunctionalTests/TraitTests.swift
+++ b/Tests/FunctionalTests/TraitTests.swift
@@ -43,7 +43,7 @@ struct TraitTests {
             let (stdout, stderr) = try await executeSwiftRun(
                 fixturePath.appending("Example"),
                 "Example",
-                extraArgs: ["--experimental-prune-unused-dependencies"],
+//                extraArgs: ["--experimental-prune-unused-dependencies"],
                 buildSystem: buildSystem,
             )
             // We expect no warnings to be produced. Specifically no unused dependency warnings.
@@ -81,7 +81,7 @@ struct TraitTests {
             let (stdout, stderr) = try await executeSwiftRun(
                 fixturePath.appending("Example"),
                 "Example",
-                extraArgs: ["--traits", "default,Package9,Package10", "--experimental-prune-unused-dependencies"],
+                extraArgs: ["--traits", "default,Package9,Package10"/*, "--experimental-prune-unused-dependencies"*/],
                 buildSystem: buildSystem,
             )
             // We expect no warnings to be produced. Specifically no unused dependency warnings.
@@ -122,7 +122,7 @@ struct TraitTests {
             let (stdout, stderr) = try await executeSwiftRun(
                 fixturePath.appending("Example"),
                 "Example",
-                extraArgs: ["--traits", "default,Package9", "--experimental-prune-unused-dependencies"],
+                extraArgs: ["--traits", "default,Package9"/*, "--experimental-prune-unused-dependencies"*/],
                 buildSystem: buildSystem,
             )
             // We expect no warnings to be produced. Specifically no unused dependency warnings.
@@ -164,7 +164,7 @@ struct TraitTests {
                 extraArgs: [
                     "--traits",
                     "default,Package5,Package7,BuildCondition3",
-                    "--experimental-prune-unused-dependencies",
+                    //"--experimental-prune-unused-dependencies",
                 ],
                 buildSystem: buildSystem,
             )
@@ -205,7 +205,7 @@ struct TraitTests {
             let (stdout, stderr) = try await executeSwiftRun(
                 fixturePath.appending("Example"),
                 "Example",
-                extraArgs: ["--disable-default-traits", "--experimental-prune-unused-dependencies"],
+                extraArgs: ["--disable-default-traits"/*"--experimental-prune-unused-dependencies"*/],
                 buildSystem: buildSystem,
             )
             // We expect no warnings to be produced. Specifically no unused dependency warnings.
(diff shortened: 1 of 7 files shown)


bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

Sources/PackageModel/EnabledTraitsMap.swift
```diff
@@ -0,0 +1,40 @@
...
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// A wrapper for a dictionary that stores the transitively enabled traits for each package.
+public struct EnabledTraitsMap: ExpressibleByDictionaryLiteral {
```
FranzBusch, line 14: Does this need to be public?

FranzBusch (approved):
Amazing! So great to see unused dependencies getting pruned

Sources/PackageModel/EnabledTraitsMap.swift
bripeticca: I'll double check, it may not need to be

bripeticca: Ah yeah this does need to be public since it's used in public initializers in other packages.

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

Sources/Commands/PackageCommands/DumpCommands.swift
```diff
@@ -53,7 +53,7 @@ struct DumpSymbolGraph: AsyncSwiftCommand {
         // We turn build manifest caching off because we need the build plan.
         let buildSystem = try await swiftCommandState.createBuildSystem(
             // We are enabling all traits for dumping the symbol graph.
-            traitConfiguration: .init(enableAllTraits: true),
+            enableAllTraits: true,
```
bkhouri, line 56: question: what benefit does passing this argument have on the creation of a build system?  Should the `enableAllTraits` option be provided when building instead?

Sources/CoreCommands/SwiftCommandState.swift
```diff
@@ -292,6 +292,8 @@ public final class SwiftCommandState {
 
     package var preferredBuildConfiguration = BuildConfiguration.debug
 
+    public var traitConfiguration: TraitConfiguration
```
bkhouri: question: can this be a `let` instead of `var`?

Sources/PackageDescription/PackageDependency.swift
```diff
@@ -56,7 +56,7 @@ extension Package {
         public let kind: Kind
 
         /// The dependencies traits configuration.
-        @available(_PackageDescription, introduced: 999.0)
+        @available(_PackageDescription, introduced: 6.1)
```
bkhouri, line 59: praise: nice!

Tests/FunctionalTests/TraitTests.swift
```diff
@@ -405,13 +403,15 @@ struct TraitTests {
...
                     "--disable-default-traits",
-                    "--experimental-prune-unused-dependencies",
                 ],
                 buildSystem: buildSystem,
             )
             // We expect no warnings to be produced. Specifically no unused dependency warnings.
             let unusedDependencyRegex = try Regex("warning: '.*': dependency '.*' is not used by any target")
             #expect(!stderr.contains(unusedDependencyRegex))
+            if buildSystem == .swiftbuild {
+                print(stderr)
```
bkhouri: question: out of curiosity, why are we printing `stderr` here?  if it's meant to troubleshoot a failed expectation, we can add a comment to the `#expect(...)` or `#require(...)`

See the [expect(_:_:sourcelocation:) API docs](https://developer.apple.com/documentation/testing/expect(_:_:sourcelocation:)) for details.

bripeticca: Ah, this must've slipped by me. I was printing out stderr for debugging purposes. Thanks for catching this!

Sources/Commands/PackageCommands/DumpCommands.swift
bripeticca: This will affect how the `SwiftCommandState` fetches and stores the currently active workspace, which is determined by its method `SwiftCommandState.getActiveWorkspace`. When creating a build system, there are many calls to various `SwiftCommandState` methods that call `getActiveWorkspace`, and the workspace that is created in this method is stored in its `_workspace` property.

The workspace is only ever created once in this method, and every subsequent call to this method checks whether the `_workspace` property has a value. If so, it will return the workspace stored there.

The reason why we have `enableAllTraits` propagated to this call is because we have special edge case scenarios where we would initially create a `SwiftCommandState` with some trait configuration, but if we then make a call that wishes to dump the symbol graph (see `PluginDelegate.createSymbolGraphForPlugin`), it will return a workspace that has whichever trait configuration we've initialized with the `SwiftCommandState`, which isn't always what we'd want when requesting a dump of the symbol graph.

Sources/CoreCommands/SwiftCommandState.swift
bripeticca: Indeed it can!

bripeticca:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

PR #8853, opened 2025-06-20
[6.2] Async git repository opening (#8721)

plemarquand:
The `RepositoryProvider` protocol's `open` method is defined as
synchronous. Convert this to `async` to avoid concurrency workarounds in
clients that might be prone to deadlocks.

This involved porting `RepositoryManager.lookup` to structured
concurrency. The method has some concurrency constraints that are
preserved:

- There is a limit (`maxConcurrentOperations`) on how many simultaneous
lookups can be performed concurrently.
- If a lookup is requested and one is in flight for the same
`RepositorySpecifier`, the in flight request completes before the new
one is started.


plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test macos

PR #8854, opened 2025-06-20
Mark known issue as intermittent for plugin test

cmcgee1024:
This test will fail (or pass) intermittently depending on the Linux builder.
Mark it as intermittent so that it no longer blocks anyone.


cmcgee1024:
@swift-ci please test

Issue #8844, opened 2025-06-19
`swift run --quiet` is not quiet


simonbates:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

`swift run --quiet` outputs build messaging in addition to the output from the specified program.

I have created a minimal project to reproduce: <https://github.com/simonbates/SwiftRunQuietExample>

### Expected behavior

```
% swift run --quiet hello
hello, world
```

### Actual behavior

```
% swift run --quiet hello
Building for debugging...
[8/8] Applying hello
Build of product 'hello' complete! (0.60s)
hello, world
```

Additionally, I have noticed that the build messaging varies slightly depending if it's the first, second, or third time running:

```
% swift run --quiet hello
Building for debugging...
[8/8] Applying hello
Build of product 'hello' complete! (0.60s)
hello, world

% swift run --quiet hello
[1/1] Planning build
Building for debugging...
[1/1] Write swift-version--58304C5D6DBC2206.txt
Build of product 'hello' complete! (0.12s)
hello, world

% swift run --quiet hello
Building for debugging...
[1/1] Write swift-version--58304C5D6DBC2206.txt
Build of product 'hello' complete! (0.12s)
hello, world
```

### Steps to reproduce

1. Clone the example repository: <https://github.com/simonbates/SwiftRunQuietExample>
2. Run `swift run --quiet hello`

### Swift Package Manager version/commit hash

Swift Package Manager - Swift 6.1.0

### Swift & OS version (output of `swift --version ; uname -a`)

```
% swift --version
swift-driver version: 1.120.5 Apple Swift version 6.1.2 (swiftlang-6.1.2.1.2 clang-1700.0.13.5)
Target: arm64-apple-macosx15.0
```

dschaefer2:
Yeah, that's not very quiet...

Diggory:
The example project appears to perform as expected now...

```
Mac ➜   SwiftRunQuietExample git:(main) ✗ swift run --quiet hello
hello, world
```

plemarquand:
@Diggory you're right, this was fixed in https://github.com/swiftlang/swift-package-manager/pull/8858, closing

PR #8858, opened 2025-06-22
Fix --quiet option not working with swift run #8844

MaelRB:
Fixes an issue where the `--quiet` option had no effect when used with `swift run` and `swift build` command.

### Motivation:

When using the `--quiet` option with `swift run` only error messages should be logged.

### Modifications:

- Add `isQuiet` property to filter out messages based on the log level during build phases of the `native`, `xcode` and `swiftbuild` build system.
- Remove duplicated declaration of the isVerbose property from different targets.

### Result:

When the `--quiet` option is used with the `swift run` and `swift build` command only error messages are logged.

### DIFF
--- a/Sources/Basics/Observability.swift
+++ b/Sources/Basics/Observability.swift
@@ -443,6 +443,14 @@ public struct Diagnostic: Sendable, CustomStringConvertible {
         public var isBold: Bool {
             return true
         }
+
+        public var isVerbose: Bool {
+            self <= .info
+        }
+
+        public var isQuiet: Bool {
+            self >= .error
+        }
     }
 }
 
--- a/Sources/Build/LLBuildProgressTracker.swift
+++ b/Sources/Build/LLBuildProgressTracker.swift
@@ -241,7 +241,9 @@ final class LLBuildProgressTracker: LLBuildBuildSystemDelegate, SwiftCompilerOut
     }
 
     func commandStatusChanged(_ command: SPMLLBuild.Command, kind: CommandStatusKind) {
-        guard !self.logLevel.isVerbose else { return }
+        guard !self.logLevel.isVerbose,
+              !self.logLevel.isQuiet
+        else { return }
         guard command.shouldShowStatus else { return }
         guard !self.swiftParsers.keys.contains(command.name) else { return }
 
@@ -285,7 +287,7 @@ final class LLBuildProgressTracker: LLBuildBuildSystemDelegate, SwiftCompilerOut
 
             self.delegate?.buildSystem(self.buildSystem, didFinishCommand: BuildSystemCommand(command))
 
-            if !self.logLevel.isVerbose {
+            if !self.logLevel.isVerbose && !self.logLevel.isQuiet {
                 let targetName = self.swiftParsers[command.name]?.targetName
                 self.taskTracker.commandFinished(command, result: result, targetName: targetName)
                 self.updateProgress()
@@ -395,6 +397,7 @@ final class LLBuildProgressTracker: LLBuildBuildSystemDelegate, SwiftCompilerOut
 
     /// Invoked right before running an action taken before building.
     func preparationStepStarted(_ name: String) {
+        guard !self.logLevel.isQuiet else { return }
         self.queue.async {
             self.taskTracker.buildPreparationStepStarted(name)
             self.updateProgress()
@@ -404,6 +407,7 @@ final class LLBuildProgressTracker: LLBuildBuildSystemDelegate, SwiftCompilerOut
     /// Invoked when an action taken before building emits output.
     /// when verboseOnly is set to true, the output will only be printed in verbose logging mode
     func preparationStepHadOutput(_ name: String, output: String, verboseOnly: Bool) {
+        guard !logLevel.isQuiet else { return }
         self.queue.async {
             self.progressAnimation.clear()
             if !verboseOnly || self.logLevel.isVerbose {
@@ -416,6 +420,7 @@ final class LLBuildProgressTracker: LLBuildBuildSystemDelegate, SwiftCompilerOut
     /// Invoked right after running an action taken before building. The result
     /// indicates whether the action succeeded, failed, or was cancelled.
     func preparationStepFinished(_ name: String, result: CommandResult) {
+        guard !self.logLevel.isQuiet else { return }
         self.queue.async {
             self.taskTracker.buildPreparationStepFinished(name)
             self.updateProgress()
@@ -431,7 +436,7 @@ final class LLBuildProgressTracker: LLBuildBuildSystemDelegate, SwiftCompilerOut
                     self.outputStream.send("\(text)\n")
                     self.outputStream.flush()
                 }
-            } else {
+            } else if !self.logLevel.isQuiet {
                 self.taskTracker.swiftCompilerDidOutputMessage(message, targetName: parser.targetName)
                 self.updateProgress()
             }
@@ -466,6 +471,7 @@ final class LLBuildProgressTracker: LLBuildBuildSystemDelegate, SwiftCompilerOut
     }
 
     func buildStart(configuration: BuildConfiguration) {
+        guard !logLevel.isQuiet else { return }
         self.queue.sync {
             self.progressAnimation.clear()
             self.outputStream.send("Building for \(configuration == .debug ? "debugging" : "production")...\n")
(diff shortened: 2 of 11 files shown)


Sources/Build/LLBuildProgressTracker.swift
```diff
@@ -242,6 +242,7 @@ final class LLBuildProgressTracker: LLBuildBuildSystemDelegate, SwiftCompilerOut
 
     func commandStatusChanged(_ command: SPMLLBuild.Command, kind: CommandStatusKind) {
         guard !self.logLevel.isVerbose else { return }
+        guard !self.logLevel.isQuiet else { return }
```
plemarquand: nit: roll this up into the guard above

Sources/Basics/Observability.swift
```diff
@@ -443,6 +443,14 @@ public struct Diagnostic: Sendable, CustomStringConvertible {
         public var isBold: Bool {
             return true
         }
+
+        public var isVerbose: Bool {
+            self <= .info
+        }
+
+        public var isQuiet: Bool {
+            self == .error
```
plemarquand: nit: just for symmetry with `isVerbose`
```suggestion
            self >= .error
```

plemarquand (approved):
Looks good, thanks for this! A few little nits, but nothing major.

plemarquand:
@swift-ci please test

jakepetroules (approved):

MaelRB:
> Thanks for this. However, to ensure we do not regress, could I ask that you add a test for the for each SwiftPM executable, where appropriate?
> 
> 
> 
> Take a look in `Tests/CommandsTests/*CommandTests.swift`.

Yes, sure! Thanks for the hint. I think I have an idea on how to do that properly.

MaelRB:
@bkhouri I don't know how we should use fixtures folders. Should we have a unique folder per test or is it ok to reuse them ? Since test suites are serialized I went for the second option, there shouldn't be any concurrent access to fixtures folders causing flaky tests.

bkhouri:
> @bkhouri I don't know how we should use fixtures folders. Should we have a unique folder per test or is it ok to reuse them ? Since test suites are serialized I went for the second option, there shouldn't be any concurrent access to fixtures folders causing flaky tests.

@MaelRB : You can re-use fixtures were necessary :). When using the `fixture(name: ...)` function, it will copy the fixture to a temporary directory, so there is no "overlap".

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1330,6 +1330,103 @@ struct BuildCommandTestCases {
...
 
+    @Test(
+        .bug("https://github.com/swiftlang/swift-package-manager/issues/8844"),
+        arguments: SupportedBuildSystemOnPlatform
+    )
+    func swiftBuildQuietLogLevel(
+        buildSystem: BuildSystemProvider.Kind,
+    ) throws {
+        try fixture(name: "Miscellaneous/SwiftBuild") { fixturePath in
+            let mainFilePath = fixturePath.appending("main.swift")
```
bkhouri: suggestion (blocking): the function implementation can be related with the following, or equivalent.


```swift
    @Test(
         .bug("https://github.com/swiftlang/swift-package-manager/issues/8844"),
         arguments: SupportedBuildSystemOnPlatform
     )
     func swiftBuildQuietLogLevel(
     ) throws {
         // GIVEN we have a simple test package
         try fixture(name: "Miscellaneous/SwiftBuild") { fixturePath in
            //WHEN we build with the --quiet option
            let (stdout, stderr) = executeSwiftBuild(
                fixturePath,
                extraArgs: ["--quiet"],
                buildSystem: buildSystem,
            )

            // THEN we should not see any output in stderr
            #expect(stderr == "")
            // AND no content in stdout
            #expect(stdout == "")
        }
    }
```
an equivalent test can be added, but calls `executeSwiftRun` instead.

If we want to take it a step further, we can also build for the various build configurations.

```
diff --git a/Tests/CommandsTests/Foo.txt b/Tests/CommandsTests/Foo.txt
index 2120fc224..a9f178e42 100644
--- a/Tests/CommandsTests/Foo.txt
+++ b/Tests/CommandsTests/Foo.txt
@@ -1,9 +1,10 @@
     @Test(
          .bug("https://github.com/swiftlang/swift-package-manager/issues/8844"),
-         arguments: SupportedBuildSystemOnPlatform
+         arguments: SupportedBuildSystemOnPlatform, BuildConfiguration.allCases,
      )
      func swiftBuildQuietLogLevel(
+         configuration: BuildConfiguraiton,
      ) throws {
          // GIVEN we have a simple test package
          try fixture(name: "Miscellaneous/SwiftBuild") { fixturePath in
@@ -11,6 +12,7 @@
             let (stdout, stderr) = executeSwiftBuild(
                 fixturePath,
                 extraArgs: ["--quiet"],
+                configuraiton: configuraiton,
                 buildSystem: buildSystem,
             )
```

MaelRB:
I've noticed an inconsistency on which output stream is used by a build system to raise the error when the build failed with the `build` command. `SwiftBuild` uses `stderr` but `native` and `xcode` use `stdout`. Is it the wanted behavior ?

From the `swiftBuildQuietLogLevelWithError` method:
```swift
if buildSystem == .swiftbuild {
   // THEN we should see output in stderr
   #expect(stderr.isEmpty == false)
   // AND no content in stdout
   #expect(stdout.isEmpty)
} else {
   // THEN we should see content in stdout
   #expect(stdout.isEmpty == false)
   // AND no output in stderr
   #expect(stderr.isEmpty)
}
```

The `run` command is consistent though

plemarquand:
@swift-ci please test

bkhouri:
> I've noticed an inconsistency on which output stream is used by a build system to raise the error when the build failed with the `build` command. `SwiftBuild` uses `stderr` but `native` and `xcode` use `stdout`. Is it the wanted behavior ?

@MaelRB Please file an issue and we can looking into it :). Feel free to reference the test (and PR).

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1330,6 +1330,77 @@ struct BuildCommandTestCases {
...
+             let error = await #expect(throws: SwiftPMError.self) {
+                 try await executeSwiftBuild(
+                    fixturePath,
+                    configuration: .debug,
+                    extraArgs: ["--quiet"],
+                    buildSystem: buildSystem
+                 )
+             }
+
+             if case .executionFailure(underlying: _, stdout: let stdout, stderr: let stderr) = error {
```
bkhouri: issue (possibly-blocking): A more idiomatic way would be to do

```swift
            guard case SwiftPMError.executionFailure(_, let stdout, let stderr) = try #require(error) else {
                Issue.record("Incorrect error was raised.")
                return
            }
            if buildSystem == .swiftbuild {
                // expectation
            } else  {
                 // expectations
            }
````

NOTE: this comment applies to all similar instances

MaelRB: Looks better indeed, thanks for the tips !

bkhouri:
@swift-ci test

bkhouri:
@MaelRB: You ~may~ need to rebase (or merge `main`) into your branch to pick the change that address the build failures.

plemarquand:
@swift-ci please test

MaelRB:
@bkhouri @plemarquand Should I wrap test cases within a `withKnownIssue` block to pass the CI ? Most of the tests in `RunCommandTests` and `BuildCommandTests` do that

bkhouri:
> @bkhouri @plemarquand Should I wrap test cases within a `withKnownIssue` block to pass the CI ? Most of the tests in `RunCommandTests` and `BuildCommandTests` do that

@MaelRB : if they are related to SwiftBuild, I would say you can mark them `withKnownIssue`, and try to include a `.bug(...)` trait on the test referring to the change that "should" address it.

MaelRB:
> > @bkhouri @plemarquand Should I wrap test cases within a `withKnownIssue` block to pass the CI ? Most of the tests in `RunCommandTests` and `BuildCommandTests` do that
> 
> @MaelRB : if they are related to SwiftBuild, I would say you can mark them `withKnownIssue`, and try to include a `.bug(...)` trait on the test referring to the change that "should" address it.

I marked two of the tests `withKnownIssue`. I couldn't link them to any known `.bug(...)` though.

plemarquand:
@swift-ci please test

MaelRB:
@plemarquand I rebased and tests pass locally, could you run the CI please ?

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test windows

plemarquand:
@MaelRB Tests are all green, if there is nothing left to add I'd love to merge this. Thanks so much for the contribution!

MaelRB:
> @MaelRB Tests are all green, if there is nothing left to add I'd love to merge this. Thanks so much for the contribution!

Thanks @plemarquand, it's all good on my side !

PR #8859, opened 2025-06-23
build-using-self: Do not set unnecessary env var

bkhouri:
The SWIFTPM_CUSTOM_BIN_DIR environment variable was set as it was required by the Integration Tests when they were standalone Package. Since merging the Integration Tests Package with the SwiftPM package Tests, setting the environment variable is no longer necessary.

Update the `build-using-self` script to no longer set this environment variable.


bkhouri:
@swift-ci test self hosted

bkhouri:
@swift-ci test

PR #8860, opened 2025-06-23
Async repository fetching

plemarquand:
The RepositoryProvider protocol's fetch method is defined as synchronous. Convert this to async to avoid concurrency workarounds in clients that might be prone to deadlocks.


plemarquand:
@swift-ci please test

PR #8874, opened 2025-06-25
Fix test failures

cmcgee1024:
There are two tests failing on the main branch at desk because
a known issue doesn't occur. Revise the when criteria to capture
only the places where the known issue occurs.


cmcgee1024:
@swift-ci please test

PR #8888, opened 2025-07-02
Support swift package migrate with --build-system swiftbuild

owenv:
Similar to the API digester integration, pass along a build system delegate to capture serialized diagnostic paths from the build and pass them to the fix-it application infra.

Builds on the changes in https://github.com/swiftlang/swift-package-manager/pull/8857, only the second commit is new

### DIFF
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -396,9 +396,9 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
     }
 
     /// Perform a build using the given build description and subset.
-    public func build(subset: BuildSubset) async throws {
+    public func build(subset: BuildSubset) async throws -> BuildResult {
         guard !self.config.shouldSkipBuilding(for: .target) else {
-            return
+            return BuildResult(serializedDiagnosticPathsByTargetName: .failure(StringError("Building was skipped")))
         }
 
         let buildStartTime = DispatchTime.now()
@@ -422,7 +422,7 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
         // any errors up-front. Returns true if we should proceed with the build
         // or false if not. It will already have thrown any appropriate error.
         guard try await self.compilePlugins(in: subset) else {
-            return
+            return BuildResult(serializedDiagnosticPathsByTargetName: .failure(StringError("Plugin compilation failed")))
         }
 
         let configuration = self.config.configuration(for: .target)
@@ -452,6 +452,17 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
         )
         guard success else { throw Diagnostics.fatalError }
 
+        let serializedDiagnosticResult: Result<[String: [AbsolutePath]], Error>
+        var serializedDiagnosticPaths: [String: [AbsolutePath]] = [:]
+        do {
+            for module in try buildPlan.buildModules {
+                serializedDiagnosticPaths[module.module.name, default: []].append(contentsOf: module.diagnosticFiles)
+            }
+            serializedDiagnosticResult = .success(serializedDiagnosticPaths)
+        } catch {
+            serializedDiagnosticResult = .failure(error)
+        }
+
         // Create backwards-compatibility symlink to old build path.
         let oldBuildPath = self.config.dataPath(for: .target).parentDirectory.appending(
             component: configuration.dirname
@@ -463,7 +474,7 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
                     warning: "unable to delete \(oldBuildPath), skip creating symbolic link",
                     underlyingError: error
                 )
-                return
+                return BuildResult(serializedDiagnosticPathsByTargetName: serializedDiagnosticResult)
             }
         }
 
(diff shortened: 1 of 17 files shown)


owenv:
@swift-ci test

Sources/Commands/PackageCommands/Migrate.swift
```diff
@@ -110,42 +112,66 @@ extension SwiftPackageCommand {
...
-            var modules: [any ModuleBuildDescription] = []
-            if !targets.isEmpty {
-                for buildDescription in buildPlan.buildModules
-                    where targets.contains(buildDescription.module.name) {
-                    modules.append(buildDescription)
+            var summary = SwiftFixIt.Summary(numberOfFixItsApplied: 0, numberOfFilesChanged: 0)
+            let fixItDuration: Duration
+            let targetsToUpdateInManifest: [String]
+            if buildSystem.supportsSerializedDaignosticsCollectionViaDelegate {
+                let diagnosticPaths = delegate.serializedDiagnosticsPathsByTarget
```
xedin: When I was originally thinking about this future when we have to support two implementations I hoped that it would be possible to expose target -> diagnostics as part of the BuildSystem protocol and abstract the implementation so we don't have to do these kind of checks the command itself...

Sources/Build/BuildOperation.swift
```diff
@@ -200,6 +200,8 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
 
     public var hasIntegratedAPIDigesterSupport: Bool { false }
 
+    public var supportsSerializedDaignosticsCollectionViaDelegate: Bool { false }
```
AnthonyLatsis:
```suggestion
    public var supportsSerializedDiagnosticsCollectionViaDelegate: Bool { false }
```

owenv: ended up deleting this code altogether

Sources/Commands/PackageCommands/Migrate.swift
owenv: I reworked this a bit so that the serialized diagnostic paths are in a BuildResult returned by BuildSystem.build instead, it gets rid of the divergence here in Migrate.swift

owenv:
@swift-ci please test

Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
```diff
@@ -59,11 +60,20 @@ public protocol BuildSystem: Cancellable {
...
         try await build(subset: .allExcludingTests)
     }
 }
 
+public struct BuildResult {
+    package init(serializedDiagnosticPathsByTargetName: Result<[String: [AbsolutePath]], Error>) {
+        self.serializedDiagnosticPathsByTargetName = serializedDiagnosticPathsByTargetName
+    }
+    
+    public var serializedDiagnosticPathsByTargetName: Result<[String: [AbsolutePath]], Error>
```
xedin, line 74: I think there is a problem with using a "name" here as a string unless we can 100% make sure that it would distinguish between host and target builds of the same module (i.e. in BuildOperation we use `module.name` which won't make a distinction)...

Sources/Commands/PackageCommands/Migrate.swift
xedin: Nice, I like that approach very much! I left a comment about "name" but otherwise LGTM!

Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
xedin: I think we need a test that uses a target as a plugin dependency to make sure that we don't override the diagnostics.

owenv: I added another test here to ensure only one fix-it is applied even if the sources need to be built twice as a common dependency of host tools and target content. I think using the name here is in line with the fact that the list of targets may be user provided - names uniquely identify targets in the manifest model, just not once they've been lowered to targets building for a specific platform

owenv:
@swift-ci test

Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
xedin: My concern is that we do allow passing flags like -Xswiftc-host or whatever it's spelled so there could be a difference in fix-its produced when building for host and destination, if we use simply a name instead of an identifier that includes target information we'd just override diagnostic files in that map.

owenv: Right now they'd be concatenated instead of one overriding the other, so the entry corresponding to "Foo" includes the diagnostic paths for Foo when built for the host and those when built for the target, and we rely on SwiftFixIt to deduplicate them. I'm not sure we want to drop the host diagnostics entirely because that feels arbitrary imo.

owenv:
@swift-ci test Windows

Sources/SPMBuildCore/BuildSystem/BuildSystem.swift
xedin: Oh I see, yeah, that sounds okay, SwiftFixIt should be able to handle that!

AnthonyLatsis:
LGTM, thanks!

Issue #8890, opened 2025-07-02
Update macro template to support Swift Testing tests


bkhouri:
According to the SwiftPM code base, we are unable to use [Swift Testing to verify macros](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/Workspace/InitPackage.swift#L119-L122). However, the referenced GitHub issue (https://github.com/swiftlang/swift-syntax/issues/2400) is closed.

The [Swift Syntax 600 release notes](https://github.com/swiftlang/swift-syntax/blob/main/Release%20Notes/600.md) mentions `SwiftSyntaxMacrosGenericTestSupport`.  

Here is a snippet of the relevant Swift Syntax release notes

> `SwiftSyntaxMacrosGenericTestSupport`
> - Description: A version of the `SwiftSyntaxMacrosTestSupport` module that doesn't depend on `Foundation` or `XCTest` and can thus be used to write macro tests using `swift-testing`. Since swift-syntax can't depend on swift-testing (which would incur a circular dependency since swift-testing depends on swift-syntax), users need to manually specify a failure handler like the following, that fails the swift-testing test: `Issue.record("\($0.message)", fileID: $0.location.fileID, filePath: $0.location.filePath, line: $0.location.line, column: $0.location.column)`
> - Pull request: https://github.com/swiftlang/swift-syntax/pull/2647

The SwiftPM template for `swift package init --type macro` should be update to provide an idiomatic way to write a Swift Testing test to verify and validate a macro.

bkhouri:
This is resolved by #8897

PR #8897, opened 2025-07-05
Update macro template to support Swift Testing tests #8890

MaelRB:
Add support for macro project using Swift Testing in `InitPackage`.

### Motivation:

Since the introduction of `SwiftSyntaxMacrosGenericTestSupport`, Swift Testing should be supported to test macro

### Modifications:

- Update `writeMacroTestsFile` to support Swift Testing
- Use `MacroSpec` for both XCTest and Swift Testing
- Remove `swift-testing` dependency from the Package manifest when building macro template for Swift Testing
- Remove `unsupportedTestingLibraryForPackageType` error as it's unused
- Add unit tests for macro template generation

### Result:

Generated test file when using Swift Testing:

```swift
import SwiftSyntax
import SwiftSyntaxBuilder
import SwiftSyntaxMacros
import SwiftSyntaxMacroExpansion
import Testing
import SwiftSyntaxMacrosGenericTestSupport

// Macro implementations build for the host, so the corresponding module is not available when cross-compiling. Cross-compiled tests may still make use of the macro itself in end-to-end tests.
#if canImport(fooMacros)
import fooMacros

let testMacros: [String: MacroSpec] = [
    "stringify": MacroSpec(type: StringifyMacro.self),
]
#endif

struct fooTests {
    
    @Test
    func macro() {
#if canImport(fooMacros)
        assertMacroExpansion(
            """
            #stringify(a + b)
            """,
            expandedSource: """
            (a + b, "a + b")
            """,
            macroSpecs: testMacros
        ) {
            Issue.record(
                "\($0.message)",
                sourceLocation:
                    SourceLocation(
                        fileID: $0.location.fileID,
                        filePath: $0.location.filePath,
                        line: $0.location.line,
                        column: $0.location.column
                    )
            )
        }
#else
        Issue.record("macros are only supported when running tests for the host platform")
#endif
    }
    
    @Test
    func macroWithStringLiteral() {
#if canImport(fooMacros)
        assertMacroExpansion(
            #"""
            #stringify("Hello, \(name)")
            """#,
            expandedSource: #"""
            ("Hello, \(name)", #""Hello, \(name)""#)
            """#,
            macroSpecs: testMacros
        ) {
            Issue.record(
                "\($0.message)",
                sourceLocation:
                    SourceLocation(
                        fileID: $0.location.fileID,
                        filePath: $0.location.filePath,
                        line: $0.location.line,
                        column: $0.location.column
                    )
            )
        }
#else
        Issue.record("macros are only supported when running tests for the host platform")
#endif
    }
}

```

Fixes: #8890

### DIFF
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -116,11 +116,6 @@ public final class InitPackage {
         installedSwiftPMConfiguration: InstalledSwiftPMConfiguration,
         fileSystem: FileSystem
     ) throws {
-        if options.packageType == .macro && options.supportedTestingLibraries.contains(.swiftTesting) {
-            // FIXME: https://github.com/swiftlang/swift-syntax/issues/2400
-            throw InitError.unsupportedTestingLibraryForPackageType(.swiftTesting, .macro)
-        }
-
         self.options = options
         self.pkgname = name
         self.moduleName = name.spm_mangledToC99ExtendedIdentifier()
@@ -346,7 +341,6 @@ public final class InitPackage {
                                     dependencies: [
                                         "\(pkgname)Macros",
                                         .product(name: "SwiftSyntaxMacrosTestSupport", package: "swift-syntax"),
-                                        .product(name: "Testing", package: "swift-testing"),
                                     ]
                                 ),
                         """
@@ -726,15 +720,22 @@ public final class InitPackage {
             import SwiftSyntax
             import SwiftSyntaxBuilder
             import SwiftSyntaxMacros
-            import SwiftSyntaxMacrosTestSupport
+            import SwiftSyntaxMacroExpansion
 
             """##
 
         if options.supportedTestingLibraries.contains(.swiftTesting) {
-            content += "import Testing\n"
-        }
-        if options.supportedTestingLibraries.contains(.xctest) {
-            content += "import XCTest\n"
+            content += ##"""
+                import Testing
+                import SwiftSyntaxMacrosGenericTestSupport
+                
+                """##
+        } else if options.supportedTestingLibraries.contains(.xctest) {
+            content += ##"""
+                import XCTest
+                import SwiftSyntaxMacrosTestSupport
+                
+                """##
         }
 
         content += ##"""
@@ -743,8 +744,8 @@ public final class InitPackage {
             #if canImport(\##(moduleName)Macros)
             import \##(moduleName)Macros
 
-            let testMacros: [String: Macro.Type] = [
-                "stringify": StringifyMacro.self,
+            let testMacros: [String: MacroSpec] = [
+                "stringify": MacroSpec(type: StringifyMacro.self),
             ]
             #endif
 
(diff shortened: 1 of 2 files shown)


bkhouri:
This resolves #8890

vanvoorden:
Ahh… just found this one! I hope we can ship a solution to this soon.

FWIW it might be worth highlighting something to engineers reminding them this support would require swift-syntax 600:

```swift
#if canImport(SwiftSyntax600)
...
#else
// error? or warning?
#endif
```

It might be confusing to engineers if they create a macro target with this template and then try to back support swift-syntax 510.

Sources/Workspace/InitPackage.swift
```diff
@@ -726,15 +720,22 @@ public final class InitPackage {
             import SwiftSyntax
             import SwiftSyntaxBuilder
             import SwiftSyntaxMacros
-            import SwiftSyntaxMacrosTestSupport
+            import SwiftSyntaxMacroExpansion
```
vanvoorden, line 723: ~~Maybe we should also import `SwiftSyntaxMacrosGenericTestSupport`?~~

Ahh… nevermind… I see it below!

Sources/Workspace/InitPackage.swift
```diff
@@ -755,10 +756,68 @@ public final class InitPackage {
...
+                        Issue.record(
+                            "\($0.message)",
+                            sourceLocation:
+                                SourceLocation(
+                                    fileID: $0.location.fileID,
+                                    filePath: $0.location.filePath,
+                                    line: $0.location.line,
+                                    column: $0.location.column
+                                )
+                        )
```
vanvoorden, line 783: Style nit: Could this be refactored to an `extension` on `Issue`? Then we can put this work in one place instead of duplicated it across N different tests?

vanvoorden:
```swift
extension Issue {
  @discardableResult static func record(_ failure: TestFailureSpec) -> Self {
    Self.record(
      Comment(rawValue: failure.message),
      sourceLocation: SourceLocation(
        fileID: failure.location.fileID,
        filePath: failure.location.filePath,
        line: failure.location.line,
        column: failure.location.column
      )
    )
  }
}
```

cc @grynspan Was there anything blocking us from shipping that extension directly in `swift-testing`? Or is there another canonical way to bridge between `SwiftSyntaxMacrosGenericTestSupport.TestFailureSpec` and `Testing.Issue`?

vanvoorden: We could also choose to factor this to a free function:

```swift
func failureHander(_ failure: TestFailureSpec) {
  Issue.record(
    Comment(rawValue: failure.message),
    sourceLocation:
      SourceLocation(
        fileID: failure.location.fileID,
        filePath: failure.location.filePath,
        line: failure.location.line,
        column: failure.location.column
      )
  )
}
```

And then pass that `failureHander` to every test similar to how we pass `testMacros`.

grynspan: Swift Testing cannot directly link to swift-syntax since it doesn't exist in the toolchain at runtime. That means we cannot export any symbols that reference types from swift-syntax (except in our macro target which is only loaded/run at compile time.)

bkhouri:
we should work the the swift-syntax owners and the Testing Workgroup to ensure this aligns with the overall vision.

grynspan:
It's not how we _want_ to implement this feature. We'd _like_ to be able to write something vaguely like:

```swift
import Testing
import SwiftSyntax

#expect(givenTheseMacros: ["foobar": FoobarMacro.self, ...]) {
  let x = #foobar
} expandsTo: {
  let x = foobarImplementation(...)
}
```

(The exact naming/shape of the `#expect` overload isn't something we've designed out.)

PR #8898, opened 2025-07-05
Recognize visionOS and DriverKit triples correctly

jakepetroules:
This makes visionOS and DriverKit builds work correctly using the Swift Build backend.

### DIFF
--- a/Sources/Basics/Vendor/Triple+Platforms.swift
+++ b/Sources/Basics/Vendor/Triple+Platforms.swift
@@ -43,6 +43,12 @@ public enum DarwinPlatform: Hashable {
   /// watchOS, corresponding to the `watchos` OS name.
   case watchOS(EnvironmentWithoutCatalyst)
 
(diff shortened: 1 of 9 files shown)


jakepetroules:
@swift-ci test

Sources/Basics/Vendor/Triple+Platforms.swift
```diff
@@ -43,6 +43,12 @@ public enum DarwinPlatform: Hashable {
   /// watchOS, corresponding to the `watchos` OS name.
   case watchOS(EnvironmentWithoutCatalyst)
 
+  /// visionOS, corresponding to the `xros` and `visionos` OS names.
```
owenv, line 46: The triple implementation is vendor over from SwiftDriver so it's probably best to try to keep the two in sync

jakepetroules: Will take a look.

PR #8900, opened 2025-07-07
Include out-of-target sources

AkiSakurai:
Add support for out-of-target declared sources and headers in TargetSourcesBuilder for tools version vNext

### Motivation:

Since out-of-target resources are already supported, this change extends similar support to out-of-target declared sources and headers, enabling consistent handling of external source files in package loading for tools versions >= vNext.

### Modifications:

Updated TargetSourcesBuilder to process declaredSources, categorizing unhandled sources as headers or additional sources based on file extensions. Added logic to integrate these sources and headers for tools versions >= vNext.

### Result:

TargetSourcesBuilder now supports out-of-target declared sources and headers, allowing seamless inclusion of external source files in package loading for tools versions >= vNext.

### DIFF
--- a/Sources/PackageLoading/TargetSourcesBuilder.swift
+++ b/Sources/PackageLoading/TargetSourcesBuilder.swift
@@ -199,8 +199,23 @@ public struct TargetSourcesBuilder {
             additionalResources = []
         }
 
-        let headers = pathToRule.lazy.filter { $0.value == .header }.map { $0.key }.sorted()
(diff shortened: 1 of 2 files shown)


bkhouri:
There appears to be a lot of formatting changes, making it difficult to focus on the actual changes.  

Although having consistent formatting is something we plan on enforcing in the project in the future, the change currently makes it difficult to focus on the actual changes. 

Would it be possible to "remove" the formatting changes across the entire file?

PR #8815, opened 2025-06-12
Support String Catalog symbol generation by always treating String Catalogs as Sources

matthewseaman:
This change adds support for String Catalog symbol generation by ensuring that String Catalogs get added to the Sources build phase in the same way that Asset Catalogs do.

PR #8905, opened 2025-07-08
Cherry-pick Support String Catalog symbol generation by adding String Catalogs as both Sources and Resources (#8815)

dschaefer2:
Support String Catalog symbol generation by adding String Catalogs as both Sources and Resources (#8815)

    - **Explanation**:
    This change adds support for String Catalog symbol generation by
    ensuring that String Catalogs get added to the Sources build phase in
    the same way that Asset Catalogs do.

    Depends on: https://github.com/swiftlang/swift-build/pull/582
    - **Scope**:
      Packages that include String Catalogs.


dschaefer2:
@swift-ci please test

PR #8906, opened 2025-07-08
Cherry-pick Allow C targets to import the compatibility header generated for Swift libraries #8736

dschaefer2:
Cherry-pick #8736 to 6.2

Address two issues preventing a C source file to import the generated compatibility header:

Don't mark the module map generated for the compatibility header as requires objc. This triggers an error when importing it from a C source file. The compatibility header is already printed in a way where the Objective-C code is protected behind a language check. C clients can safely import it even if they may not see any content. Plus we're adding C content to the compatibility header with the official support for @cdecl that is independent of Objective-C.

Emit the generated compatibility header and module map in a new include directory under the temporary build directory of each Swift library. This directory is then used as a header search path for the dependent clang targets.

This replaces the previous strategy that relied only the generated module map to resolve imports. It was incompatible with C source files and non-modular textual includes. The use of a header search path preserves the support for module imports via the same module map and supports textual includes of the compatibility header.

### DIFF
--- a/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
@@ -910,15 +910,20 @@ public final class SwiftModuleBuildDescription {
         try self.fileSystem.writeFileContents(path, bytes: .init(encodingAsUTF8: content), atomically: true)
     }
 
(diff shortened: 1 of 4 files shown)


dschaefer2:
@swift-ci please test

jakepetroules (approved):

PR #8896, opened 2025-07-04
Disable check for unsafe flags.

dschaefer2:
As discussed in the Swift forums below, a number of projects require the use of "unsafe" flags. A common example is the use of absolute header include paths set via an environment variable that project consumers agree to use. As such, the flags can be safely used under those circumstances.

This change disables the one line that was doing the check.

We will further study the issue and come up with other mechanisms to ensure the challenges some flags have on quality of the package ecosystem are mitigated.

https://forums.swift.org/t/pitch-disable-checks-for-unsafe-flags-in-swiftpm/80698

PR #8914, opened 2025-07-09
Cherry-pick Disable check for unsafe flags. (#8896)

dschaefer2:
Cherry-pick #8896 

As discussed in the Swift forums below, a number of projects require the use of "unsafe" flags. A common example is the use of absolute header include paths set via an environment variable that project consumers agree to use. As such, the flags can be safely used under those circumstances.

This change disables the one line that was doing the check.

We will further study the issue and come up with other mechanisms to ensure the challenges some flags have on quality of the package ecosystem are mitigated.

https://forums.swift.org/t/pitch-disable-checks-for-unsafe-flags-in-swiftpm/80698

### DIFF
--- a/Sources/PackageLoading/PackageBuilder.swift
+++ b/Sources/PackageLoading/PackageBuilder.swift
@@ -2024,7 +2024,8 @@ extension Sequence {
 
 extension TargetDescription {
     fileprivate var usesUnsafeFlags: Bool {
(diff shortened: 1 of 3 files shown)


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

bripeticca (approved):

PR #8920, opened 2025-07-10
[6.2] Bump swift-syntax in template to 620.0.0-latest

rconnell9:
`swift package init --type macro` still points to swift-syntax version 6.0.0 although main has moved on to 6.2.0.

### Motivation:

`swift package init --type macro` is outdated

### Modifications:

Locally verified that `swift package init --type macro` uses the updated version.

### Result:

`swift package init --type macro` will output the latest version


rconnell9:
@swift-ci test

PR #8925, opened 2025-07-11
[6.2] AsyncProcess: simplify the code path to not depend on the TSC wrappers functions on Darwin

jakepetroules:


jakepetroules:
@swift-ci test

PR #8934, opened 2025-07-15
swift-package-migrate: Hide global options

AnthonyLatsis:
### Motivation:

Currently, options that are specific to `swift package migrate` are shown at the very bottom, following all the global options.

All anonymous option groups get coalesced into the default "OPTIONS" section, which always comes last. So there is no proper way to hoist subcommand options above all the base/global options other than giving the subcommand option group a title.

Using "options" as the title would give us 2 "OPTIONS" sections, whereas a custom title, such as "migrate options", would introduce yet another design inconsistency. Instead, address this by hiding global options, as most other package subcommands do.

### DIFF
--- a/Tests/CommandsTests/PackageCommandTests.swift
+++ b/Tests/CommandsTests/PackageCommandTests.swift
@@ -2106,6 +2106,15 @@ class PackageCommandTestCase: CommandsBuildProviderTestCase {
         }
     }
 
+    func testMigrateCommandHelp() async throws {
+        let (stdout, _) = try await self.execute(
+            ["migrate", "--help"],
+        )
+
+        // Global options are hidden.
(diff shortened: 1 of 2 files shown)


AnthonyLatsis:
@swift-ci please test

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -2106,6 +2106,15 @@ class PackageCommandTestCase: CommandsBuildProviderTestCase {
         }
     }
 
+    func testMigrateCommandHelp() async throws {
```
bkhouri, line 2109: praise: thank you for creating an automated tests.

bkhouri (approved):

PR #8940, opened 2025-07-16
[BuildPlan] Fix `traverseModules` to account for tests with direct ma…

xedin:
…cro dependencies

### Motivation:

Currently there is a difference in behavior between `BuildPlan.computeDestinations` and `BuildPlan.traverseModules`
where the latter always returns test targets as built for the "target". This is incorrect if i.e. targets have direct macro
dependencies.

### Modifications:

This change aligns `traverseModules` with `computeDestinations` in regard to test targets that are built for the host.
And let's test targets be discovered through an aggregate produce that also informs their destination.

### Result:

The fix helps sourcekit-lsp to correctly identify test targets.

Resolves: https://github.com/swiftlang/sourcekit-lsp/issues/2205

### DIFF
--- a/Sources/Build/BuildPlan/BuildPlan.swift
+++ b/Sources/Build/BuildPlan/BuildPlan.swift
@@ -1120,13 +1120,40 @@ extension BuildPlan {
             guard visited.insert(.package(package)).inserted else {
                 return []
             }
-            return package.modules.compactMap {
-                if case .test = $0.underlying.type,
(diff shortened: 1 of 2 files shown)


xedin:
@swift-ci please test

xedin:

bnbarham:
Thanks @xedin. I can confirm this fixes the issue 🙇

PR #8948, opened 2025-07-17
Cherry-pick Assure trait-guarded dependencies are not included in resolution; precompute enabled traits before resolution (#8852)

bripeticca:
Cherry pick of #8852 

Trait-guarded dependencies were still being considered during dependency
resolution, when they should be excluded if they aren't being used in
any other scenario. Additionally, enabled traits should be pre-computed
before entering the resolution stage in order to avoid possible race
conditions when navigating through the dependency graph nodes.

### Modifications:

Since we have the `--experimental-prune-unused-dependencies` feature
behind an experimental flag, we'll now consider an alternate path that
will prune trait-guarded package dependencies from the dependency graph
**_if and only if_** said dependency is not used in any other unguarded
context.

A dictionary wrapper titled `EnabledTraitsMap` is used to store the
enabled traits per package for the package graph and is stored as a
property within the `Workspace`, with some additional behaviour to
return a `["default"]` set of traits if the package has not yet been
stored in the dictionary, rather than returning `nil`.

Following this behaviour, when passing a set of traits to methods that
require them (e.g. for dependency computation, enabled trait
computation, etc.) we now require that it is not Optional, since the
checks done on a `nil` set of traits were redundant.

SwiftCommandState now also stores a `TraitConfiguration`, since we'll
want access to this across multiple `SwiftCommand`s, and it is
essentially a part of the state anyhow. `TraitOptions` is now included
in the `GlobalOptions` for `SwiftCommand`s to supplement this, so that
when a `SwiftCommandState` is created we will have access to the
user-passed enabled traits. These options, as entitled, are available
globally across all the swift package commands, so previous properties
that declared `TraitOptions` in these commands has been removed in
favour of using it straight from the `GlobalOptions`.

### Result:

Trait-guarded dependencies are excluded from dependency resolution, and
since traits are pre-computed there should no longer be an issue with
race conditions for traits in resolution as well.

### DIFF
--- a/Tests/FunctionalTests/TraitTests.swift
+++ b/Tests/FunctionalTests/TraitTests.swift
@@ -101,7 +101,7 @@ struct TraitTests {
             """)
         }
         } when: {
-            ProcessInfo.hostOperatingSystem == .windows && CiEnvironment.runningInSmokeTestPipeline || buildSystem == .swiftbuild
+            ProcessInfo.hostOperatingSystem == .windows && CiEnvironment.runningInSmokeTestPipeline || (buildSystem == .swiftbuild && [.windows, .linux].contains(ProcessInfo.hostOperatingSystem))
         }
     }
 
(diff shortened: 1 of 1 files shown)


bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

bripeticca:
@swift-ci please test windows

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test windows

PR #8951, opened 2025-07-17
Only enable unsafe flags if version 6.2 or newer.

dschaefer2:
As we disable checking for unsafe flags, make sure they are still checked when the swift-tools-version is earlier than 6.2. This ensures consumers of the package don't get unsafe flags errors when using older toolchains. They will need to upgrade their toolchains to use packages with unsafe flags.


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:

PR #8960, opened 2025-07-21
Cherry pick swiftCompilerTag and swift-syntax URL redirect

dschaefer2:
Cherry pick #8959

The swiftCompilerTag field has been recently added to the swiftc -print-target-info JSON we use to determine the swift compiler version for prebuilts. Use this field as the default value, falling back to extracting the swift compiler version from the compilerVersion field using a regex.

This gives us much better version numbers since it matches the tags for the toolchain.

Also add in the redirect url, https://github.com/apple/swift-syntax.git, as a supported option.


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

PR #8965, opened 2025-07-23
swift-package-migrate: Miscellaneous low-risk improvements and more tests

AnthonyLatsis:
### Modifications:

* Improve command-line option descriptions. Be clear that both options accept a comma-separated list, which is not at all obvious, and also state the default for `--target`.
* Add tests for feature resolution errors. This change also extracts feature resolution into a method and removes a superfluous colon from the 'unsupported feature' error message.
* Emit feature validation errors through the diagnostic emitter. These errors do not represent issues with parsing command-line arguments, so throwing a `ValidationError` is not suitable — we don't want to present the user with a command usage hint in these cases.
* Add more clarity to manifest update error message along with some tests for it.

### DIFF
--- a/Fixtures/SwiftMigrate/UpdateManifest/Package.updated.targets-A-B.swift
+++ b/Fixtures/SwiftMigrate/UpdateManifest/Package.updated.targets-A-B.swift
@@ -1,0 +1,28 @@
+// swift-tools-version:5.8
+
+import PackageDescription
+
+var swiftSettings: [SwiftSetting] = []
+
+let package = Package(
+    name: "WithErrors",
+    targets: [
+        .target(
+            name: "CannotFindSettings",
+            swiftSettings: swiftSettings
+        ),
+        .target(name: "A",swiftSettings: [
+    .enableUpcomingFeature("ExistentialAny"),
+    .enableUpcomingFeature("InferIsolatedConformances"),]),
+        .target(name: "B",swiftSettings: [
+    .enableUpcomingFeature("ExistentialAny"),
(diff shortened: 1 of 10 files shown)


AnthonyLatsis:
@swift-ci please test

Fixtures/SwiftMigrate/UpdateManifest/Package.updated.targets-A-B.swift
```diff
@@ -0,0 +1,28 @@
...
+
+let package = Package(
+    name: "WithErrors",
+    targets: [
+        .target(
+            name: "CannotFindSettings",
+            swiftSettings: swiftSettings
+        ),
+        .target(name: "A",swiftSettings: [
+    .enableUpcomingFeature("ExistentialAny"),
```
bkhouri, line 15: question: it's outside the scope of this PR, but it is possible for the `migrate` tool to properly indent this code? or it this reliant on a `.swiftformat` file, which we don't currently have in the repository

AnthonyLatsis: Yeah, I’ve noted this down. swift-syntax already has a basic formatting library (`SwiftBasicFormat`) that it uses to [format macro expansions](https://github.com/swiftlang/swift-syntax/blob/7b1fcd6ced8269aa030e042b2e9d8938174ea237/Sources/SwiftSyntaxMacroExpansion/MacroExpansion.swift#L516) for display, so we’re definitely in a good position to do a better job here.

https://github.com/swiftlang/swift-package-manager/issues/8966.

PR #8837, opened 2025-06-17
Extends support for arm64e to macOS & visionOS platforms 

pmattos:
Extends **arm64e** support for **macOS & visionOS** apps, besides the existing iOS support.

### Motivation:

SwiftPM currently only supports building **arm64e** packages for iOS.

In Xcode 26 we added support to building packages for arm64e targeting macOS & visionOS platforms too.

### Modifications:

Conditionally adds `"arm64e"` to the `ARCHS` build settings in the `PackagePIFBuilder` used to interface with Swift Build.

### Result:

PIF builder nows had the capability of builds packages for arm64e. Note, for now, that this is currently disabled in SwiftPM.

I validated this by building a simple macOS executable for arm64e:

```shell
$ swift package init --type executable --name HelloWorld
$ swift run swift-build --package-path HelloWorld --build-system swiftbuild
```
...and then I inspected the generated binary:
```shell
$ lipo -archs HelloWorld/.build/arm64-apple-macosx/Products/Debug/HelloWorld 
arm64e
```

PR #8969, opened 2025-07-25
Build fat packages when shouldPackagesBuildForARM64e is set

pmattos:
### Motivation:

I extended **arm64e** support for macOS & visionOS apps in my previous PR #8837. When enabled, that patch *only* produced the arm64e slice but... 

The App Store will now accept **arm64e** for macOS, iOS, and visionOS — arm64e is optional and off by default. But when you turn on arm64e support for packages (ie, using `shouldPackagesBuildForARM64e` PIF builder delegate method) we expect *both* slices in the binary now.

### Modifications:

Changes the PIF builder to update the `ARCHS` settings without overriding it completely, ensuring we get all required slices (ie, a so called fat binary).

### Result:

PIF builder nows has the capability of building packages for arm64 *and* arm64e. (Note, for now, that this is currently disabled in SwiftPM.)

I validated this by building a simple macOS executable for arm64e (and force enabling it in SwifPM):

```shell
$ swift package init --type executable --name HelloWorld
$ swift run swift-build --package-path HelloWorld --build-system swiftbuild --configuration release
```
...and then I inspected the generated binary:
```shell
$ lipo -archs HelloWorld/.build/arm64-apple-macosx/Products/Release/HelloWorld 
x86_64 arm64 arm64e
```

**Debug builds**. With `ONLY_ACTIVE_ARCH = "YES"` in debug builds, even though you specify multiple architectures in `ARCHS`, it will still only build for the active architecture during development. This is set to `"NO"` for release builds.

### DIFF
--- a/Sources/SwiftBuildSupport/PackagePIFBuilder.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFBuilder.swift
@@ -590,7 +590,7 @@ public final class PackagePIFBuilder {
                 default:
                     .init(from: arm64ePlatform)
                 }
-                settings.platformSpecificSettings[pifPlatform]![.ARCHS] = ["arm64e"]
+                settings.platformSpecificSettings[pifPlatform]![.ARCHS, default: []].append(contentsOf: ["arm64e"])
             }
         }
 


pmattos:
@swift-ci test

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
```diff
@@ -590,7 +590,7 @@ public final class PackagePIFBuilder {
                 default:
                     .init(from: arm64ePlatform)
                 }
-                settings.platformSpecificSettings[pifPlatform]![.ARCHS] = ["arm64e"]
+                settings.platformSpecificSettings[pifPlatform]![.ARCHS, default: []].append(contentsOf: ["arm64e"])
```
jakepetroules, line 593: This works because we set ARCHS to $(inherited) earlier in the scope, right?

pmattos: Exactly Jake! It ends up like this:
```swift
["$(inherited)", "arm64e"]
```
...which in turns gets evaluated to:
```swift
["x86_64", "arm64", "arm64e"]
```

dschaefer2 (approved):

jakepetroules (approved):

PR #8974, opened 2025-07-26
Cannot use Metal properly with Swift Packages

rconnell9:
Cannot use Metal properly with Swift Packages — even with Apple-recommended plugin workaround

The summary here is that when using xcrun to locate metal, xcodebuild is invoked which queries MobileAssets for the location of the metal toolchain. The query to MobileAssets requires the added entitlement.

Reference: https://github.com/swiftlang/swift-package-manager/issues/8930

### Modifications:

Add expected entitlement


dschaefer2:
@swift-ci please test

dschaefer2:
We probably need an integration test for this with a sample build plugin as recommended in the tech note.

rconnell9:
@swift-ci please test

PR #8992, opened 2025-08-05
Only fetch swiftc target info when needed.

dschaefer2:
We added fetching of the swift compiler version using swiftc -print-target-info in the init of UserToolchain so that it can be passed to the PrebuiltsManager to calculate the manifest and prebuilts zip file names.

However, there are cases where we pass in the triple for the current SDK which was intended to avoid the call to swiftc. We now call it all the time which causes an extra call to swiftc that isn't always needed.

To address this, we make the calculation of the target info lazy and move the call to fetch the swift version to the point we need it in the PrebuiltsManager. If prebuilts aren't required, the call is avoided.

To properly test this we need to run a test that calls out to swiftc. However, we don't have integration tests yet for prebuilts since they would take a long time to execute. I have instead manually tested this using a staged version of the prebuilts.

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -43,7 +43,10 @@ public final class UserToolchain: Toolchain {
     public let librarySearchPaths: [AbsolutePath]
 
     /// An array of paths to use with binaries produced by this toolchain at run time.
-    public let runtimeLibraryPaths: [AbsolutePath]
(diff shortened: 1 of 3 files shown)


dschaefer2:
@swift-ci please test

jakepetroules (approved):

dschaefer2:
@swift-ci please test macOS

dschaefer2:
@swift-ci please test macOS

dschaefer2:
@swift-ci please test macOS

PR #8994, opened 2025-08-05
Add build prebuilts support for any version of swift-syntax.

dschaefer2:
Removes the hardcoding of the contents of the prebuilt libraries.

Adds a --version argument to specify which versions of swift-syntax you want to build for. You can have multiple of these. We add a sample build.sh script to the folder to show an example, including finding the most recent prerelease of 602.

Also removes the hardcoding of the list of products that go into the library. Instead we ask the package for its list of library products and add them all.

Finally, removes the header files since SwiftPM now hooks up the include paths to the CModules from the source in the checkouts.

### DIFF
--- a/Sources/swift-build-prebuilts/BuildPrebuilts.swift
+++ b/Sources/swift-build-prebuilts/BuildPrebuilts.swift
@@ -35,88 +35,13 @@ struct Artifact: Codable {
     var swiftVersion: String?
 }
 
-// The master list of repos and their versions
-struct PrebuiltRepos: Codable {
-    let url: URL
-    let versions: [Version]
-
-    struct Version: Identifiable, Codable {
-        let tag: String
-        let manifest: Workspace.PrebuiltsManifest
-
-        var id: String { tag }
-    }
-}
-
-var prebuiltRepos: [PrebuiltRepos] = [
-    .init(
-        url: .init(string: "https://github.com/swiftlang/swift-syntax")!,
-        versions: [
-            .init(
-                tag:"600.0.1",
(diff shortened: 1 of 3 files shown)


dschaefer2:
@swift-ci please test

Sources/swift-build-prebuilts/BuildPrebuilts.swift
```diff
@@ -207,136 +131,110 @@ struct BuildPrebuilts: AsyncParsableCommand {
...
+        let srcModulesDir = buildDir.appending("Modules")
+        let prebuiltDir = stageDir.appending(id)
 
-        for repo in prebuiltRepos {
-            let repoDir = srcDir.appending(repo.url.lastPathComponent)
-            let scratchDir = repoDir.appending(".build")
-            let buildDir = scratchDir.appending("release")
-            let srcModulesDir = buildDir.appending("Modules")
-            let prebuiltDir = stageDir.appending(repo.url.lastPathComponent)
+        try await shell("git clone https://github.com/swiftlang/swift-syntax.git", cwd: srcDir)
```
jakepetroules, line 142: not new code but we should be avoiding `shell` where possible in order to avoid injections, especially where the command is not a string literal

jakepetroules (approved):

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test self-hosted windows

dschaefer2:
@swift-ci please test windows

dschaefer2:
Waiting to do a little more testing on all the platforms before pushing this.

PR #9001, opened 2025-08-06
Allow use of swiftbuild backend when multi-arch builds are explicitly requested

owenv:
Previously, these would fallback to --build-system Xcode, which isn't necessary since Swift Build can build universal binaries


owenv:
@swift-ci test

owenv:
Closes https://github.com/swiftlang/swift-package-manager/issues/8982

jakepetroules (approved):

PR #9005, opened 2025-08-08
Add a swift-format configuration file

### LINKED ISSUE #9397: [CI][Soundness] Enable support for Format

bkhouri:
Add a `.swift-format` file, format the code base, and enforce formatting in CI.

Depends on: https://github.com/swiftlang/swift-format/pull/1156

Fixes: #9397


bkhouri:
@swift-ci test

.swift-format
```diff
@@ -0,0 +1,11 @@
+{
+    "version": 1,
+    "lineLength": 10000,
+    "indentation": {
+        "spaces": 4
+    },
+    "lineBreakBeforeEachArgument": true,
```
owenv, line 7: This rule feels a little too opinionated to me, not sure if anyone else has an opinion

bkhouri: I think this option helps make the diff ever-so-slightly cleaner to read.

.swift-format
```diff
@@ -0,0 +1,11 @@
+{
+    "version": 1,
+    "lineLength": 10000,
+    "indentation": {
```
bkhouri, line 4: suggestion: I haven't read the documentation, but based on the setting name, we should consider `"multiElementCollectionTrailingCommas" : true,`.

.swift-format
```diff
@@ -0,0 +1,11 @@
...
+        "spaces": 4
+    },
+    "lineBreakBeforeEachArgument": true,
+    "rules": {
+    },
```
bkhouri, line 11: suggestion:  I haven't read the documentation ,but based on the setting name, we should consider the following, eventually:

```
  "rules" : {
    "AllPublicDeclarationsHaveDocumentation" : false,
  }
```

.swift-format
jakepetroules: That corresponds to:

```swift
[
    "a",
    "b"
]
```

vs

```swift
[
    "a",
    "b",
]
```

Support for the latter was added to Swift some time ago, the argument being that it helps reduce the SCM diff as new lines are added to the bottom.

I'd be fine with enabling that rule.

.swift-format
```diff
@@ -0,0 +1,11 @@
+{
+    "version": 1,
+    "lineLength": 10000,
```
jakepetroules, line 3: Agree with this; we should not enforce a line length.

.swift-format
jakepetroules: This would result in a pretty huge diff. While I agree it can enhance readability in some cases, I'm not sure we should make the jump right away. I'd omit this for now and maybe we should have a wider discussion first.

.swift-format
jakepetroules: Good to strive towards, but we should not add that immediately as I think for SwiftPM in particular there's a lot which is public but which maybe shouldn't be.

jakepetroules:
@bkhouri Can you open a similar PR for swift-build?

.swift-format
bkhouri: I want the latter :)

.swift-format
bkhouri: 100% agree :). Something to consider for the future.

bkhouri:
> @bkhouri Can you open a similar PR for swift-build?

@jakepetroules there is already a `.swift-format` file in `swift-build`.

Maybe we should merge this change once we have GitHub actions, where, in one PR, we enforce formatting rules and format the code base.

.swift-format
bkhouri: I also want

`[1, 3]`

to format to

```
[
    1,
    3,
]
```

.swift-format
bkhouri: I would like to get to the following

```
func foo(a: Int, b: String) {
}
```

to 

```
func foo(
    a: Int,
    b: String,
) {
}
```

bkhouri:
@swift-ci test

bkhouri:
@swift-ci test windows

bripeticca:
@swift-ci please test

.github/workflows/pull_request.yml
```diff
@@ -36,5 +36,5 @@ jobs:
       license_header_check_enabled: false
       unacceptable_language_check_enabled: false
       api_breakage_check_enabled: false
-      format_check_enabled: false
+      format_check_enabled: true
```
bkhouri, line 46: chore (blocking): We need to exclude the `Fixtures` directory from the formatting check since it contains code that does not compile.

bkhouri:
Waiting on resolution of https://github.com/swiftlang/swift-format/issues/870 before continuing this work to ensure at-desk formatting matches the CI formatting.

Package.swift
```diff
@@ -20,32 +20,36 @@ import PackageDescription
...
+    ]
 } else {
     swiftpmLinkSettings = []
     packageLibraryLinkSettings = []
 }
 
 // Common experimental flags to be added to all targets.
 let commonExperimentalFeatures: [SwiftSetting] = [
-    .enableExperimentalFeature("MemberImportVisibility"),
+    .enableExperimentalFeature("MemberImportVisibility")
```
bkhouri, line 44: chore (blocking): hmm.. it's disappointing the trailing comma was removed.

.github/workflows/pull_request.yml
bkhouri: This should now be feasible with https://github.com/swiftlang/swift-format/pull/1156.  We just need to wait for it to land in a main nightly toolchain.

Benchmarks/Benchmarks/PackageGraphBenchmarks/PackageGraphBenchmarks.swift
```diff
@@ -35,23 +34,26 @@ let benchmarks = {
...
+    {
         modulesGraphWidth = parsedValue
     } else {
         modulesGraphWidth = 150
     }
 
     let packagesGraphDepth: Int
     if let envVar = ProcessInfo.processInfo.environment["SWIFTPM_BENCHMARK_PACKAGES_GRAPH_DEPTH"],
-    let parsedValue = Int(envVar) {
+        let parsedValue = Int(envVar)
```
bkhouri, line 55: chore (blocking): I don't like this formatting change.

Fixtures/BinaryLibraries/Static/Package1/Package.swift
```diff
@@ -9,7 +9,7 @@ let package = Package(
             name: "Example",
             dependencies: [
                 "Simple",
-                "Wrapper"
+                "Wrapper",
```
bkhouri: I like that a trailing command was added here.

PR #9011, opened 2025-08-11
Add build prebuilts support for any version of swift-syntax. (#8994)

dschaefer2:
Removes the hardcoding of the contents of the prebuilt libraries.

Adds a --version argument to specify which versions of swift-syntax you want to build for. You can have multiple of these. We add a sample build.sh script to the folder to show an example, including finding the most recent prerelease of 602.

Also removes the hardcoding of the list of products that go into the library. Instead we ask the package for its list of library products and add them all.

Finally, removes the header files since SwiftPM now hooks up the include paths to the CModules from the source in the checkouts.

This is needed in 6.2 to support prebuilts generated with that version of the toolchain. I have tested this on all three host platforms including all Linux distros with both architectures and swift-syntax 600, 601, and the latest prerelease of 602

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -785,6 +785,9 @@ let package = Package(
                 .product(name: "ArgumentParser", package: "swift-argument-parser"),
                 "Basics",
                 "Workspace",
+            ],
+            exclude: [
+                "build.sh"
             ]
         ),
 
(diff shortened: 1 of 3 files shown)


jakepetroules (approved):

dschaefer2:
@swift-ci please test macOS

dschaefer2:
@swift-ci please test macOS

PR #9012, opened 2025-08-11
Tests: Migrate PackageCommandTests to Swift Testing and Augment

bkhouri:
Migrate the `PackageCommandTests` to Swift Testing and augment the suite to run against the Native and SwiftBuild build system, in addition to the `debug` and `release` build configurations.

Depends on: #9020
Relates to: #8997
issue: rdar://157669245

PR #9015, opened 2025-08-12
Fix for .when(traits:) condition not working for multiple traits

bripeticca:
The culprit here is the fact that we were essentially doing an AND boolean check across all the traits in the target dependency condition, when we should be doing an OR boolean check.

If at least one trait is enabled in the list of traits in the target dependency condition, then the target dependecy should be considered enabled as well.

### DIFF
--- a/Sources/PackageModel/Manifest/Manifest+Traits.swift
+++ b/Sources/PackageModel/Manifest/Manifest+Traits.swift
@@ -95,7 +95,7 @@ extension Manifest {
         _ parentPackage: PackageIdentifier? = nil
     ) throws {
         guard supportsTraits else {
-            if explicitlyEnabledTraits != ["default"] /*!explicitlyEnabledTraits.contains("default")*/ {
+            if explicitlyEnabledTraits != ["default"] {
                 throw TraitError.traitsNotSupported(
                     parent: parentPackage,
                     package: .init(self),
@@ -116,7 +116,7 @@ extension Manifest {
         let areDefaultsEnabled = enabledTraits.contains("default")
 
         // Ensure that disabling default traits is disallowed for packages that don't define any traits.
-        if !(explicitlyEnabledTraits == nil || areDefaultsEnabled) && !self.supportsTraits {
+        if !areDefaultsEnabled && !self.supportsTraits {
             // We throw an error when default traits are disabled for a package without any traits
             // This allows packages to initially move new API behind traits once.
             throw TraitError.traitsNotSupported(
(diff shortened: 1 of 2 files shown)


bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

FranzBusch (approved):

Sources/PackageModel/Manifest/Manifest+Traits.swift
```diff
@@ -449,15 +449,18 @@ extension Manifest {
...
             .init(stringLiteral: $0),
             enabledTraits,
-        ) }
+        ) })
 
         return traitsToEnable.isEmpty || isEnabled
     }
     /// Determines whether a given package dependency is used by this manifest given a set of enabled traits.
-    public func isPackageDependencyUsed(_ dependency: PackageDependency, enabledTraits: Set<String>/* = ["default"]*/) throws -> Bool {
+    public func isPackageDependencyUsed(_ dependency: PackageDependency, enabledTraits: Set<String>) throws -> Bool {
```
bkhouri, line 463: suggestion: can we write some automated tests that would validate this function in isolation.  This would allow us to validate various scenarios, include fault injection et al.   If we are unable to control the test the was we want, we can rework the function to make it testable.

Sources/PackageModel/Manifest/Manifest+Traits.swift
```diff
@@ -449,15 +449,18 @@ extension Manifest {
 
         let traitsToEnable = self.traitGuardedTargetDependencies(for: target)[dependency] ?? []
 
-        let isEnabled = try traitsToEnable.allSatisfy { try self.isTraitEnabled(
+        // Check if any of the traits guarding this dependency is enabled;
```
bkhouri, line 452: suggestion: can we write some automated tests that would validate this function in isolation. This would allow us to validate various scenarios, include fault injection et al. If we are unable to control the test the was we want, we can rework the function to make it testable.

Tests/FunctionalTests/TraitTests.swift
```diff
@@ -653,4 +653,56 @@ struct TraitTests {
...
+            isIntermittent: (ProcessInfo.hostOperatingSystem == .windows),
+        ) {
+        try await fixture(name: "Traits") { fixturePath in
+            // Test various combinations of traits that would
+            // enable the dependency on Package10Library2
+            let traitCombinations = ["ExtraTrait", "Package10", "ExtraTrait,Package10"]
+            // We expect no warnings to be produced. Specifically no unused dependency warnings.
+            let unusedDependencyRegex = try Regex("warning: '.*': dependency '.*' is not used by any target")
+
+            for traits in traitCombinations {
```
bkhouri: suggestion: Consider adding this as a test parameterization.

Have a look at #9012 that combines `BuildSystem` and `BuildConfiguration` into a single array, allow us to send a second array as argument.

Here's that [introduces the `getBuildData(...)` function](https://github.com/swiftlang/swift-package-manager/pull/9012/files#diff-eb4bb50a62166270974faffbfc1e5198dbc04c51fa2902be3bd114d5559a7e00R25) and [here's an application](https://github.com/swiftlang/swift-package-manager/pull/9012/files#diff-67d75918940c3c3422732811ec1d7efa9e203173cbce1dd182bf57d4ca1c2ae5R102)

bkhouri (commented):
Some comments for your consideration.

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

jakepetroules (approved):

bripeticca:
@swift-ci please test windows

bripeticca:

PR #9019, opened 2025-08-13
[6.1] Frequent crashes in LLBuildProgressTracker while building swift-java on linux and swift 6.1.2

rconnell9:
### Explanation:
(cherry picked from commit 16452e1305d5f635cc714146201ca9753bcfea3f)

Frequent crashes in LLBuildProgressTracker while building swift-java on linux and swift 6.1.2

This was a race condition because the 
Command is associated with a handle (pointer) that may be reclaimed when this function finishes.

### Scope:
Minimal, copy variables into locals before accessing them from async block.

### Issues:
https://github.com/swiftlang/swift-java/issues/314 almost all CI runs fail with this race condition due to the timing of the command being influenced by the gradle build in swift-java.

### Original PRs:
https://github.com/swiftlang/swift-package-manager/pull/8998

### Risk:
Minimal.  Change explicitly copies variables prior to usage in async block. This avoided a race condition that would manifest before as null pointer crashes when trying to access these variables.

Change has been tested in main and 6.2

### Testing:
Confirmed fixed in main and 6.2. Manually verified in always reproducing project.

### DIFF
--- a/Sources/Build/LLBuildProgressTracker.swift
+++ b/Sources/Build/LLBuildProgressTracker.swift
@@ -277,6 +277,9 @@ final class LLBuildProgressTracker: LLBuildBuildSystemDelegate, SwiftCompilerOut
         guard command.shouldShowStatus else { return }
         guard !self.swiftParsers.keys.contains(command.name) else { return }
 
+        let commandName = command.name
+        let commandDescription = command.description
+
         self.queue.async {
             if result == .cancelled {
                 self.cancelled = true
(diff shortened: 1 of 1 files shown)


ktoso (approved):
Thank you Lgtm! I confirmed this seems to resolve our issues when we tried a nightly toolchain with the fix.

Nitpick: Pr needs to follow the [template](https://github.com/swiftlang/.github/blob/main/PULL_REQUEST_TEMPLATE/release.md) a bit more

MaxDesiatov:
@swift-ci test

ktoso:
@airspeedswift approved, merged. Thank you for the fix!

Issue #9033, opened 2025-08-15
expand the information provided by `swift package show-dependencies` to include traits


heckj:
### Description

Noting this down as a feature request - and following up/[re-iterating a request in the Swift Forums](https://forums.swift.org/t/se-0450-package-traits/75598/52):

Right now it's hard to determine what features are exposed (and enabled) by dependencies you're using _other_ than inspecting the relevant Package.swift.

It would be really useful to be able to get this information from `swift package show-dependencies`. The suggestion in the Swift forums post was to extend the information returned, and offer a `--traits` option on the `show-dependencies` command to allow you to specify an additional trait that's exposed to you:

```
$ swift package show-dependencies --traits traitX
.
├── swift-argument-parser<https://github.com/apple/swift-argument-parser@1.3.0> (traits: default,traitX)
└── swift-system<https://github.com/apple/swift-system@1.3.2>
```

### Expected behavior

`show-dependencies` shows what traits are enabled for any dependencies that offer them, and have a default set enabled.
`show-dependencies` shows traits and dependencies included when you provide/activate a trait that the local package exposes.

### Actual behavior

No traits information is exposed

### Swift Package Manager version/commit hash

b6a3d955451044d16139d7831a616d213c06cdfd (`swift-DEVELOPMENT-SNAPSHOT-2025-08-14-a`)

### Swift & OS version (output of `swift --version && uname -a`)

```
swift-driver version: 1.127.11.2 Apple Swift version 6.2 (swiftlang-6.2.0.16.14 clang-1700.3.16.4)
Target: arm64-apple-macosx15.0
Darwin Joes-MacBook-Pro-3.local 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:40 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6041 arm64
```

PR #9034, opened 2025-08-15
adds traits to dependency output in show-dependencies --json and --text

heckj:
Adds traits to the output of the show-dependencies command for `--text` and `--json` output.

### Motivation:

Exposes the default traits, if any, that are enabled by the dependencies a package.

Resolves: #9033 

### Modifications:

Extended the logic in show-dependencies to add in the `enabledTraits` for a resolved package.

### Result:

The output of show-dependencies includes output akin to `(traits: FirstTrait, SecondTrait)` is the resolved package has traits and includes the traits that it has. If the package doesn't offer traits, no additional/new content is displayed in `--text`.
The JSON provided from `show-dependencies --json` includes a `traits` property, empty if there aren't any traits enabled.

### DIFF
--- a/Sources/Commands/Utilities/DependenciesSerializer.swift
+++ b/Sources/Commands/Utilities/DependenciesSerializer.swift
@@ -34,7 +34,14 @@ final class PlainTextDumper: DependenciesDumper {
 
                 let pkgVersion = package.manifest.version?.description ?? "unspecified"
 
-                stream.send("\(hanger)\(package.identity.description)<\(package.manifest.packageLocation)@\(pkgVersion)>\n")
+                let traitsEnabled: String
+                if let enabled = package.enabledTraits, !enabled.isEmpty {
+                    traitsEnabled = "(traits: \(package.enabledTraits?.joined(separator: ", ") ?? ""))"
+                } else {
+                    traitsEnabled = ""
+                }
+
+                stream.send("\(hanger)\(package.identity.description)<\(package.manifest.packageLocation)@\(pkgVersion)>\(traitsEnabled)\n")
 
                 if !package.dependencies.isEmpty {
                     let replacement = (index == packages.count - 1) ?  "    " : "│   "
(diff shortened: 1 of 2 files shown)


heckj:
@swift-ci please test

bkhouri (changes requested):
I know this is still in draft, but could we please add an automated test to ensure we don't regress?

heckj:
@bkhouri absolutely - I was keeping it draft until I could see what tests where running and add in something appropriate. Just added now, following the pattern of the other nearby "command tests"

bripeticca (approved):
this is great! thank you for doing this :)

bripeticca:
@swift-ci please test

Sources/Commands/Utilities/DependenciesSerializer.swift
```diff
@@ -130,6 +137,7 @@ final class JSONDumper: DependenciesDumper {
                 "url": .string(package.manifest.packageLocation),
                 "version": .string(package.manifest.version?.description ?? "unspecified"),
                 "path": .string(package.path.pathString),
+                "traits": .string(package.enabledTraits?.joined(separator: ", ") ?? ""),
```
jakepetroules: Stringifying this for the JSON output doesn't seem right when the source data is an array. How about this?


```suggestion
                "traits": .array(package.enabledTraits?.map { .string($0) } ?? []),
```

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -660,6 +660,30 @@ class PackageCommandTestCase: CommandsBuildProviderTestCase {
         }
     }
 
+    func testShowDependenciesWithTraits() async throws {
```
bkhouri: praise: thanks for creating a test.  though this will likely conflict with #9012, which migrate the test to Swift Testing and augments to run against the Native and SwiftBuild build system, in addition to running against the `debug` and `release` build configurations.

Unless it's urgent, could I ask that you wait to merge this change after #9012?

bkhouri (approved):

Tests/CommandsTests/PackageCommandTests.swift
heckj: More than happy to - nothing here is super critical, and I'd love to see how this is migrating to help me do the same elsewhere.

heckj:
@swift-ci please test

bripeticca:

PR #9036, opened 2025-08-17
Add SDK target triples to the output of the two bundle commands `swift sdk {list,configure --show-configuration bundle-id}`

finagolfin:
There was previously no way to query the target triples supported, so make sure they are explicitly listed on the CLI.


finagolfin:
@swift-ci test self hosted

finagolfin:
@swift-ci smoke test

finagolfin:
@swift-ci test linux

finagolfin:
@swift-ci test macos

finagolfin:
There doesn't seem to be a way to trigger the required non-Windows smoke tests alone? `swift-ci test` runs everything, while `swift-ci test self hosted` only runs the three self-hosted runs, but I can't seem to only run the remaining three smoke tests after that.

finagolfin:
@swift-ci test self hosted macos

finagolfin:
Alright, ready for review, as the self-hosted mac failure is spurious, since [it passed the first time](https://ci.swift.org/job/swift-package-manager-with-xcode-self-hosted-PR-osx-release-toolchain/1994/).

finagolfin:
@plemarquand, would you take a look?

Sources/SwiftSDKCommand/ListSwiftSDKs.swift
```diff
@@ -50,8 +50,13 @@ package struct ListSwiftSDKs: SwiftSDKSubcommand {
             return
         }
 
-        for artifactID in validBundles.sortedArtifactIDs {
-            print(artifactID)
+        let sdkBundles = validBundles.listSwiftSDKs(hostTriple: hostTriple)
+        for bundle in validBundles.sortedArtifactIDs {
+            print("SDK bundle: \(bundle)")
+            print("SDKs:")
+            for triple in sdkBundles[bundle]! {
```
jakepetroules: Can we avoid the force unwrap here?

finagolfin: I'll check. I am using the keys from the earlier method call to access these dictionary values, so I figured it was safe, but can add more checking.

finagolfin:
@swift-ci test self hosted linux

Sources/SwiftSDKCommand/ListSwiftSDKs.swift
finagolfin: Alright, replaced this with some checking to see if the bundle artifact ID is actually in the bundle dictionary and that there is at least one valid target triple. I wrote it this way because I want all the SDK output alphabetically ordered, so it first gets the dictionary of SDK triples, then gets the ordered set of SDK keys and iterates through the dictionary with those.

When the compiler complained that the key may not be valid and return an optional, I force-unwrapped it because I was simply iterating the dictionary with its own sorted keys. However, thinking about it some more, depending on whether bundle installs have any locking in place, it is possible for the list of bundles to change between getting the list of keys and then iterating over the dictionary and then rather than blowing up with inscrutable force-unwrap errors, this new code will simply output the minimal set of bundles that it finds.

Please let me know if you have any other comments.

finagolfin:
@swift-ci test linux

finagolfin:
@swift-ci test linux

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
```diff
@@ -150,4 +150,27 @@ extension [SwiftSDKBundle] {
     public var sortedArtifactIDs: [String] {
         self.flatMap(\.artifacts.keys).sorted()
     }
+
+    public func listSwiftSDKs(hostTriple: Triple) -> [String : [Triple]] {
```
MaxDesiatov, line 158: Is this meant to be adopted outside of SwiftPM? If not, it shouldn't be `public`. Also, please add a doc string to the function declaration to clarify if host triples or target triples are returned.

```suggestion
    package func listSwiftSDKs(hostTriple: Triple) -> [String : [Triple]] {
```

Sources/PackageModel/SwiftSDKs/SwiftSDKConfigurationStore.swift
```diff
@@ -202,6 +202,7 @@ public final class SwiftSDKConfigurationStore {
             }
 
             if showConfiguration {
+                print("\nSDK: \(targetTriple)")
```
MaxDesiatov: IMO this should be hidden under a flag. There are scripts in existence relying on each output line being a valid Swift SDK ID, and this change needlessly breaks this workflow without any warning or means for transition.

Sources/SwiftSDKCommand/ListSwiftSDKs.swift
```diff
@@ -50,8 +50,15 @@ package struct ListSwiftSDKs: SwiftSDKSubcommand {
             return
         }
 
-        for artifactID in validBundles.sortedArtifactIDs {
-            print(artifactID)
+        let sdkBundles = validBundles.listSwiftSDKs(hostTriple: hostTriple)
+        for bundle in validBundles.sortedArtifactIDs {
+            if let sdkTriples = sdkBundles[bundle], !sdkTriples.isEmpty {
+                print("SDK bundle: \(bundle)")
+                print("SDKs:")
```
MaxDesiatov: Ditto, this should be hidden behind a flag to preserve compatibility with existing scripts.

Sources/PackageModel/SwiftSDKs/SwiftSDKConfigurationStore.swift
MaxDesiatov: These are not SDKs, but Swift SDKs. These are two different concepts that have different implementation and different set of options etc. We should make the distinction clear to end users.

```suggestion
                print("\nSwift SDK: \(targetTriple)")
```

Sources/SwiftSDKCommand/Configuration/ShowConfiguration.swift
```diff
@@ -45,6 +45,7 @@ struct ShowConfiguration: ConfigurationSubcommand {
         _ swiftSDKsDirectory: AbsolutePath,
         _ observabilityScope: ObservabilityScope
     ) throws {
+        print("\nSDK: \(targetTriple)")
```
MaxDesiatov:
```suggestion
        print("\nSwift SDK: \(targetTriple)")
```

Sources/SwiftSDKCommand/ListSwiftSDKs.swift
MaxDesiatov:
```suggestion
                print("Swift SDK bundle: \(bundle)")
                print("Swift SDKs:")
```

MaxDesiatov (changes requested):
The existing output was made minimal on purpose to make it easily parsable with scripts. This change would break such scripts without any warning or means to transition to the new output format.

Additionally, it introduces confusion between different concepts of SDKs and Swift SDKs (e.g. we use different flags for these, `--sdk` vs `--swift-sdk`). The distinction should be clear in strings visible by end users, and ideally in the code as well.

Sources/SwiftSDKCommand/ListSwiftSDKs.swift
```diff
@@ -50,8 +50,15 @@ package struct ListSwiftSDKs: SwiftSDKSubcommand {
             return
         }
 
-        for artifactID in validBundles.sortedArtifactIDs {
-            print(artifactID)
+        let sdkBundles = validBundles.listSwiftSDKs(hostTriple: hostTriple)
+        for bundle in validBundles.sortedArtifactIDs {
+            if let sdkTriples = sdkBundles[bundle], !sdkTriples.isEmpty {
```
MaxDesiatov:
```suggestion
        let swiftSDKBundles = validBundles.listSwiftSDKs(hostTriple: hostTriple)
        for bundle in validBundles.sortedArtifactIDs {
            if let targetTriples = swiftSDKBundles[bundle], !targetTriples.isEmpty {
```

Sources/SwiftSDKCommand/ListSwiftSDKs.swift
```diff
@@ -50,8 +50,15 @@ package struct ListSwiftSDKs: SwiftSDKSubcommand {
...
+                print("SDK bundle: \(bundle)")
+                print("SDKs:")
+                for triple in sdkTriples {
```
MaxDesiatov:
```suggestion
                for triple in targetTriples {
```

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
finagolfin: I simply followed what the sorted list of `sortedArtifactIDs` was doing above by making it `public`, as I felt this was more useful than that, will add the doc.

Sources/PackageModel/SwiftSDKs/SwiftSDKConfigurationStore.swift
finagolfin: Hmm, are you sure there are many script consumers yet? I suspect that most users of these commands are on the command-line to simply show these SDKs, and this is very important info to everyone, that follows the general format of the existing output. I suspect the few script consumers will be very happy to have this important info and will quickly update their scripts.

Sources/SwiftSDKCommand/ListSwiftSDKs.swift
finagolfin: Same as above, this is such important info that users want, that nobody will complain that adding more info breaks a few scripts.

Sources/PackageModel/SwiftSDKs/SwiftSDKConfigurationStore.swift
finagolfin: Good point, will make that change.

finagolfin:
> The existing output was made minimal on purpose to make it easily parseable with scripts. This change would break such scripts without any warning or means to transition to the new output format.

That is great that you thought about and want to support such scripts, but the most important criterion is that users are given all the info they need. Is there any reason this crucial info of what target triples each SDK supports was never accessible on the CLI before this pull, just an oversight? Whatever the reason, I suspect both script and non-script users will be so happy to have this info finally that nobody will complain. 😃

Completely agree about the "Swift SDK" terminology, will make those improvements you asked for.

finagolfin:
@swift-ci test self hosted

finagolfin:
Made the terminology changes you asked for, Max, but disagree that we need to worry about breaking script consumers of these commands' output, as this SDK target info is so important that all scripts will happily update to use it. 😺

Sources/PackageModel/SwiftSDKs/SwiftSDKConfigurationStore.swift
```diff
@@ -202,6 +202,7 @@ public final class SwiftSDKConfigurationStore {
             }
 
             if showConfiguration {
+                print("\nswiftSDK: \(targetTriple)")
```
MaxDesiatov, line 205: Why is the key called `swiftSDK` if its value is target triple?

```suggestion
                print("\ntargetTriple: \(targetTriple)")
```

Sources/SwiftSDKCommand/Configuration/ShowConfiguration.swift
```diff
@@ -45,6 +45,7 @@ struct ShowConfiguration: ConfigurationSubcommand {
         _ swiftSDKsDirectory: AbsolutePath,
         _ observabilityScope: ObservabilityScope
     ) throws {
+        print("\nswiftSDK: \(targetTriple)")
```
MaxDesiatov, line 48: ditto

```suggestion
        print("\ntargetTriple: \(targetTriple)")
```

Sources/SwiftSDKCommand/ListSwiftSDKs.swift
```diff
@@ -50,8 +50,15 @@ package struct ListSwiftSDKs: SwiftSDKSubcommand {
             return
         }
 
-        for artifactID in validBundles.sortedArtifactIDs {
-            print(artifactID)
+        let swiftSDKBundles = validBundles.listSwiftSDKs(hostTriple: hostTriple)
+        for bundle in validBundles.sortedArtifactIDs {
+            if let targetTriples = swiftSDKBundles[bundle], !targetTriples.isEmpty {
+                print("Swift SDK bundle: \(bundle)")
+                print("Swift SDKs:")
```
MaxDesiatov, line 57:
```suggestion
                print("Target triples:")
```

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
MaxDesiatov: Existing code shouldn't be used as precedent for selecting access control for new code. There's a very high bar for introducing new public APIs in SwiftPM, as evolving it or removing in the future is a high maintenance cost. Unless we know of the exact use case and client adopting this new API, we should not introduce this as a public API in the first place.

MaxDesiatov:
> That is great that you thought about and want to support such scripts, but the most important criterion is that users are given all the info they need.

I agree with the general premise, but having control over output format, the amount/aspects of information in the output, and not breaking existing code relying on this format is just as important.

> Is there any reason this crucial info of what target triples each SDK supports was never accessible on the CLI before this pull, just an oversight?

The reason is that scripting and non-scripting output modes can be approached separately in separate PRs. I implemented the scripting mode in the original PR, and didn't get to non-scripting output mode yet. I personally never use `swift sdk list` outside of scripts, and haven't seen requests to prioritise non-scripting output mode prior to this PR.

> Whatever the reason, I suspect both script and non-script users will be so happy to have this info finally that nobody will complain. 😃

It's premature to claim so, especially as this PR as is breaks compatibility with existing code. The general philosophy of Swift is that breaking existing code and workflows without any deprecation/transition path is a bug when done outside of a major release. In theory we could consider breaking this in Swift 7.0, after a corresponding deprecation period that emits warnings and guides users towards a potential replacement.

In the meantime, I'd like to see this gated with `--verbose`, `--show-triples`, or whatever flag we find more appropriate.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
finagolfin: I feel that this is so generally useful that it should be a public API, do you disagree?

MaxDesiatov: SwiftPM has been overly generous with providing public APIs over the years, and given that it isn't semantically versioned none of these APIs could be removed or changed without breaking downstream clients, which in itself is very undesirable experience that it makes it very hard to change or refactor anything in the project.

Unless we have an exact client/project/package that we can point to that absolutely needs this API as `public` and doesn't have an alternative way of implementing it, nothing should be made `public` only because it's "generally useful".

Sources/PackageModel/SwiftSDKs/SwiftSDKConfigurationStore.swift
finagolfin: Good point, I chose to stick with the currently used terminology, as the command to actually use the SDK is `swift build --swift-sdk aarch64-swift-linux-musl`, as can be seen in [the static linux doc](https://www.swift.org/documentation/articles/static-linux-getting-started.html). Since @al45tair or others chose to use the triple as the selector for the SDKs without using the target triple terminology, eg in the help:
```
> ./swift-6.2-DEVELOPMENT-SNAPSHOT-2025-08-13-a-fedora39/usr/bin/swift build --help |ag "swift-sdk "
  --swift-sdk <swift-sdk> Filter for selecting a specific Swift SDK to build
```
I simply followed the language they used, so as not to confuse users.

However, I now see that the `sdk configure` help does mention target triples, so I can change this to match that.

Sources/SwiftSDKCommand/Configuration/ShowConfiguration.swift
finagolfin: Same as above.

Sources/SwiftSDKCommand/ListSwiftSDKs.swift
finagolfin: On this one I disagree, for the reasons given above, ie users are going to just use the `swift sdk list` command and plug that into `swift build --swift-sdk` most of the time, so using the same terminology is better.

Sources/PackageModel/SwiftSDKs/SwiftSDKConfigurationStore.swift
MaxDesiatov: Please do change this to triples, this is an opportunity to clean this up in preparation for [Swift SDK aliases](https://github.com/swiftlang/swift-package-manager/pull/8703).

Sources/SwiftSDKCommand/ListSwiftSDKs.swift
MaxDesiatov: Retrospectively, enabling use of triples as arguments for `--swift-sdk` was a mistake that led to confusing and inconsistent behavior. It's already not possible to use triples when targeting Wasm, as we have to ship multiple Swift SDKs with the same triple in the same bundle. We should not dig a deeper hole, but take this as an opportunity to slightly clean it up. In the future I expect vast majority of users to rely on [Swift SDK aliases](https://github.com/swiftlang/swift-package-manager/pull/8703) instead of Swift SDK IDs or triples.

finagolfin:
> not breaking existing code relying in this format is just as important.

I find it hard to believe that almost anybody is relying on this format yet, so better to break it and get it right now while still early.

> The reason is that scripting and non-scripting output modes can be approached separately in separate PRs. I implemented the scripting mode in the original PR, and didn't get to non-scripting output mode yet. I personally never use swift sdk list outside of scripts, and haven't seen requests to prioritise non-scripting output mode prior to this PR.

OK, but all that doesn't answer my question of why there isn't a single SwiftPM command yet that shows the target triples that are available in an installed Swift SDK bundle, whether for scripting output or non-scripted. That is critical info that currently has to be documented separately in doc like the static linux SDK instructions I linked above, because SwiftPM simply does not provide it.

> It's premature to claim so, especially as this PR as is breaks compatibility with existing code. The general philosophy of Swift is that breaking existing code and workflows without any deprecation/transition path is a bug when done outside of a major release. In theory we could consider breaking this in Swift 7.0, after a corresponding deprecation period that emits warnings and guides users towards a potential replacement.

I don't think anybody has ever claimed that such compatibility guarantees apply to unversioned CLI output like this, particularly for new commands like this that few people are using. Otherwise we'd almost never be able to change the output of many CLI commands.

> In the meantime, I'd like to see this gated with --verbose, --show-triples, or whatever flag we find more appropriate.

I disagree. I think the prior output is so inferior to the new output proposed in this pull that we should make this the new default, and relegate the old one to some extra flag.

I have felt that the current `swift sdk` output and the absence of a target triple list is seriously broken and a significant usability problem with this SDK bundles feature you added for some time, ever since I first started using it late last year. This is my attempt to fix that and I believe something like this should be the new default.

MaxDesiatov:
> I find it hard to believe that almost anybody is relying on this format yet better to break it and get it right now while still early.

This is not early by any measure. The commands and the output format have been present since Swift 5.8 released in March 2023, if we don't count development snapshots released prior to that. This is a stabilised feature that's way past the experimental incubation stage, where such breakage would be allowed. We had plenty of time for feedback and changes impacting default output workflows.

Now is a good opportunity for additive changes though, so having this new output format enabled with a flag in Swift 6.x timeline is perfectly fine, and I would wholeheartedly support that approach. If we see enough positive feedback arguing for this to become the default, next major release of Swift would be the opportunity to make the switch after appropriate deprecation period with a few releases that warn users about the change.

> OK, but all that doesn't answer my question of why there isn't a single SwiftPM command yet that shows the target triples that are available in an installed Swift SDK bundle, whether for scripting output or non-scripted. That is critical info that currently has to be documented separately in doc like the static linux SDK instructions I linked above, because SwiftPM simply does not provide it.

If it were critical, I'm sure we'd see reports on GitHub issues/PR or Swift Forums requesting this to be implemented, would you mind linking to those?

Either way, I'm not debating that this is needed. I'm grateful for the PR that adds this, but we must exercise care not to break established workflows.

> we'd almost never be able to change the output of many CLI commands.

It's important to distinguish change and breakage. What I'm arguing against is specifically breakage, while I do welcome changes for improvements that avoid breakage and respect existing users.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
finagolfin: Alright, will change it.

finagolfin:
> This is not early by any measure. The commands and the output format have been present since Swift 5.8 released in March 2023

You are right that this feature has been experimentally available for a couple years, I meant that it is still early in its adoption.

> Now is a good opportunity for additive changes though, so having this new output format enabled with a flag in Swift 6.x timeline is perfectly fine, and I would wholeheartedly support that approach. If we see enough positive feedback arguing for this to become the default, next major release of Swift would be the opportunity to make the switch after appropriate deprecation period with a few releases that warn users about the change.

Given how few people are likely scripting the current output, I think we should make listing SDK triples the default much sooner than that.

> If it were critical, I'm sure we'd see reports on GitHub issues/PR or Swift Forums requesting this to be implemented, would you mind linking to those?

As I'm sure you know from working at Apple, end users generally do not report usability problems like this. The people who usually post on the forums or github are power users who read the doc carefully or figure out things for themselves, so they are a small and unrepresentative sample.

Since you asked, I did check now though, [only 25 issues referencing "swift sdk" in this repo in those 2.5 years since you shipped it](https://github.com/swiftlang/swift-package-manager/issues?q=is%3Aissue%20%22swift%20sdk%22), mostly familiar names of Apple people and bundle implementors like me, more than a quarter from me. 😉 On the forum, [only five threads mention "swift sdk list" before this week](https://forums.swift.org/search?q=%22swift%20sdk%20list%22%20order%3Alatest) and [three mention "swift sdk configure", all by implementors](https://forums.swift.org/search?q=%22swift%20sdk%20configure%22%20order%3Alatest).

I think this data backs up what I'm saying that the forums/github are mostly for highly advanced users and that this feature is still early in its adoption. Not meant as a criticism, as I'm trying to get this feature much more used, just stating what I believe to be a fact.

> Either way, I'm not debating that this is needed. I'm grateful for the PR that adds this, but we must exercise care not to break established workflows.

> It's important to distinguish change and breakage. What I'm arguing against is specifically breakage, while I do welcome changes for improvements that avoid breakage and respect existing users.

Based on the limited usage of this feature so far, and likely almost no reliance by scripts, I think we should focus on usability and gaining adoption first, then we can worry about not breaking things for that larger userbase.

It doesn't look like you will agree on that approach though, so unless some other SwiftPM dev wants to step in and push you to make this change now, I will update this pull to gate this `swift sdk list` output change behind yet another flag.

Sources/SwiftSDKCommand/ListSwiftSDKs.swift
finagolfin: OK, if I'm gating this behind a flag anyway, I can change the label too.

PR #9052, opened 2025-08-20
Fix testDiagnostics usages in ModulesGraphTests

plemarquand:
The ModulesGraphTests have been converted to swift-testing, but they still use `testDiagnostics` which will call XCTFail instead of the failing with a swift-testing expectation.


bkhouri:
@swift-ci test

PR #9059, opened 2025-08-23
Emit "Found unhandled resource" to observability scope

jakepetroules:
Libraries should never call print. Emit this diagnostic message to the observability scope instead.


jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test windows

PR #9069, opened 2025-08-27
Use local dependencies in `build-using-self`

bnbarham:
Cross PR testing only works when using update-checkout and local dependencies. Update `build-using-self` to pass `SWIFTCI_USE_LOCAL_DEPS` so that we're not re-fetching (possibly incorrect) dependencies.


bnbarham:
@swift-ci please test

PR #9076, opened 2025-08-29
SE-0501: HTML coverage report

bkhouri:
Currently, `swift test` supports generating a JSON coverage report, which is great for ingesting into various systems. The JSON coverage report generation is processed by `llvm-cov` where SwiftPM is simply orchestrating the command line invocation. The JSON, however, is not very "human consumable friendly" while iterating at-desk.

This change adds an additional command line argument to swift test that allows specify whether to generate a JSON or HTML report. Selecting HTML will produce a `llvm-cov` HTML report, that is more easily consumable by a developer at desk. Similar to the JSON coverage report, Swift Package Manager is only responsible for constructing the command line argument that will be passed to `llvm-cov`, which will produce the HTML report.

Since `llvm-cov show` (which would produce the HTML) has numerous CLI arguments that allows configurations of the HTML report, this change supports `-Xcov` argument for specifying additional coverage report arguments.

See [SE-0501: HTML Coverage Report](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0501-swiftpm-html-coverage-report.md)

Depends on: https://github.com/apple/swift-argument-parser/pull/830
Depends on: #9162 
Depends on: #10146
Depends on: #10400
Depends on: #10434
Depends on: https://github.com/swiftlang/swift/pull/89403


bkhouri:
This PR is blocked on a Swift Argument Parser release which contains https://github.com/apple/swift-argument-parser/pull/830

FranzBusch:
@bkhouri What is the latest on this? The evolution has been accepted but the implementation is still pending

bkhouri:
> @bkhouri What is the latest on this? The evolution has been accepted but the implementation is still pending


@FranzBusch : This change required a feature in Swift Argument Parser 1.8.1, but that version requires being build with Swift 6.0.  Building the toolchain on linux currently failed as the base version in the images uses Swift 5.x.  So this PR is currently blocked on https://github.com/swiftlang/swift/pull/89403.

We need @shahmishal to update the Swift version used in the docker images used to build the Linux toolchain.

shahmishal:
>We need @shahmishal to update the Swift version used in the docker images used to build the Linux toolchain.

The Docker images were updated for `main` branch to 6.2.x version a few weeks ago.

bkhouri:
@swift-ci test

bkhouri:
@swift-ci test windows

Tests/CommandsTests/CoverageTests.swift
```diff
@@ -143,15 +160,327 @@ struct CoverageTests {
...
+                fixturePath,
+                configuration: configuration,
+                extraArgs: commonCoverageArgs + [
+                    "--enable-coverage"
+                ],
+                buildSystem: buildSystem
+            )
+
+            // Ensure all paths in the data exists.
+            // try withKnownIssue {
```
bkhouri: Chore (blocking): remove the commented code

bkhouri:
@swift-ci test

bkhouri:
@swift-ci test windows

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1155,7 +1345,90 @@ extension SwiftTestCommand {
...
+    do {
+        // Loop on the contents.
+        for (index, line) in arguments.enumerated() {
+            if !line.contains(outputDirectoryArgumentName) {
+                continue
+            }
+
+            if line == outputDirectoryArgumentName || line.hasSuffix(" \(outputDirectoryArgumentName)") {
+                // The argument value is on the next line
+                let value = "\(arguments[index + 1])"
```
plemarquand: (blocking) This crashes when the user omits an --output-dir argument, i.e.:
`swift test --enable-coverage --coverage-format html -Xcov --output-dir --show-coverage-path`

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1155,7 +1345,90 @@ extension SwiftTestCommand {
...
+                )
+
+            case .json:
+                outputPath = try await buildSystem.codeCovPath(for: productsBuildParameters).appending(component: rootManifest.displayName + ".json")
+        }
+        return outputPath
+    }
+}
+
+package func getHtmlCoverageOutputDir(
```
plemarquand: I don't think these methods should live in SwiftTestCommand.swift, instead move them to Sources/Commands/Coverage.

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1134,16 +1280,60 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
...
+        return String(decoding: data, as: UTF8.self)
+    }
+
+    func getCoveragePath(
+        _ swiftCommandState: SwiftCommandState,
+        format: CoverageFormat,
+    ) async throws -> AbsolutePath {
         let workspace = try swiftCommandState.getActiveWorkspace()
         let root = try swiftCommandState.getWorkspaceRoot()
         let rootManifests = try await workspace.loadRootManifests(
```
plemarquand: `getCoveragePath` loads the `rootManifests` as well as creates the build system, relatively expensive operations. Its is called once per coverage format supplied. 

It is called from two places, `processCodeCoverage`, where `rootManifests`/`buildSystem` is already being loaded, and in `getCodeCovOutputPaths` where we could load the rootManifests/buildSystem before calling `getCoveragePath` in a loop.

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -983,23 +1000,68 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
...
+            swiftCommandState.outputStream.send("Code coverage report:\n")
+            for (format, path) in coverageReportData {
+                swiftCommandState.outputStream.send("  - \(format.rawValue.uppercased()): \(path.pathString)\n")
+            }
+            swiftCommandState.outputStream.flush()
+        }
+
+        for format in formats {
+            let coverageReportOutput = try await self.getCoveragePath(swiftCommandState, format: format)
+            let extraArgs = self.options.coverageOptions.xcovArguments.getArguments(for: format)
```
plemarquand: is using `self.options.coverageOptions.xcovArguments` the intent here instead of the `xcovArguments` argument supplied to the function?

Sources/Commands/Coverage/CoverageOptions.swift
```diff
@@ -0,0 +1,154 @@
...
+
+    package var defaultValueDescription: String {
+        switch self {
+            case .json: "Produce a JSON coverage report by executing 'llvm-cov export'."
+            case .html: "Produce an HTML report by executing 'llvm-cov show'."
+        }
+    }
+}
+
+extension CoverageFormat: Comparable {
```
owenv, line 28: Can this conformance be synthesized?

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1155,7 +1345,90 @@ extension SwiftTestCommand {
...
+            // Let's parse via regular expression
+            if let match = line.wholeMatch(of: outputDirRegex) {
+                let (_, outputDir) = match.output
+                returnValue = try convertStringToAbsolutePath(outputDir)
+            }
+        }
+
+        return returnValue ?? defaultPath
+    } catch {
+        return defaultPath
```
plemarquand: Any error silently falls back to `defaultPath`, including unparsable paths.

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -983,23 +1000,68 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
...
+            }
+            swiftCommandState.outputStream.flush()
+        }
+
+        for format in formats {
+            let coverageReportOutput = try await self.getCoveragePath(swiftCommandState, format: format)
+            let extraArgs = self.options.coverageOptions.xcovArguments.getArguments(for: format)
+            switch format {
+                case .json:
+                    // let (productsBuildParameters, _) = try swiftCommandState.buildParametersForTest(options: self.options)
```
owenv: accidentally left in?

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1039,16 +1105,96 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
...
+            let output = try result.utf8Output() + result.utf8stderrOutput()
+            throw StringError("Unable to generate HTML code coverage report:\n \(output)")
+        }
+
+        // the output put can be updated via the command arg file
+        return outputPath
+    }
+
+    /// Gets all production source files from test targets and their dependencies.
+    private func getProductionSourceFiles(
```
owenv: Why do we call these production source files?

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1155,7 +1345,90 @@ extension SwiftTestCommand {
...
+    outputDirectoryArgumentName: String,
+    workspacePath: AbsolutePath,
+    defaultPath: AbsolutePath,
+) ->  AbsolutePath {
+    var returnValue : AbsolutePath? = nil
+    // let commandArg = "--output-dir"
+    // let lines = content.split(whereSeparator: \.isNewline)
+
+    let outputDir = Reference(String.self)
+    let outputDirRegex = Regex {
```
plemarquand: Parsing the --output-dir argument with regex isn't the right approach. I'd rather see this implemented as a method on XcodeArgumentCollection using more traditional techniques, something like:

```swift
func outputDirectory(
    for coverageFormat: CoverageFormat,
    relativeTo basePath: AbsolutePath
) throws -> AbsolutePath? {
    let arguments = self.getArguments(for: coverageFormat)
    let joinedPrefix = "\(Self.outputDirectoryArgument)="

    let values: [String] = try arguments.indices.compactMap { index in
        let argument = arguments[index]

        if argument.hasPrefix(joinedPrefix) {
            return String(argument.dropFirst(joinedPrefix.count))
        }

        guard argument == Self.outputDirectoryArgument else { return nil }
        guard arguments.indices.contains(index + 1) else {
            throw XcovArgumentError.missingOutputDirectoryValue
        }
        return arguments[index + 1]
    }

    guard let value = values.last else { return nil }
    guard !value.isEmpty else {
        throw XcovArgumentError.emptyOutputDirectoryValue
    }
    return try AbsolutePath(validating: value, relativeTo: basePath)
}
```

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1039,16 +1105,96 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
...
+    ) async throws -> [AbsolutePath] {
+        var sourceFiles = Set<AbsolutePath>()
+
+        // Get all modules from root packages that are not test modules
+        // These are the production modules that tests are covering
+        for package in packageGraph.rootPackages {
+            for module in package.modules {
+                // Include all non-test, non-plugin modules from root packages
+                if module.type != .test && module.type != .plugin {
+                    sourceFiles.formUnion(module.sources.paths)
```
owenv: I think this will exclude source files generated by plugins, is that OK?

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1039,16 +1105,96 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
...
+                }
+            }
+        }
+
+        // If no source files found from root packages, fall back to all reachable modules
```
owenv: When would we hit this fallback?

Sources/Commands/Coverage/CoverageFormatOutput.swift
```diff
@@ -0,0 +1,110 @@
...
+            return nil
+        }
+    }
+
+    /// Adds a key/value pair to the underlying dictionary.
+    /// - Parameters:
+    ///   - format: The coverage format key
+    ///   - path: The absolute path value
+    /// - Throws: `StringError` if the key already exists
+    package mutating func addFormat(_ format: CoverageFormat, path: AbsolutePath) throws {
```
plemarquand: Many of these methods seem to only be exercised by tests and can be removed.

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1155,7 +1345,90 @@ extension SwiftTestCommand {
...
+}
+
+package func getHtmlCoverageOutputDir(
+    from arguments: [String],
+    outputDirectoryArgumentName: String,
+    workspacePath: AbsolutePath,
+    defaultPath: AbsolutePath,
+) ->  AbsolutePath {
+    var returnValue : AbsolutePath? = nil
+    // let commandArg = "--output-dir"
```
owenv: left in accidentally?

Tests/CommandsTests/TestCommandTests.swift
```diff
@@ -28,7 +28,7 @@ import protocol ArgumentParser.AsyncParsableCommand
 import class TSCBasic.BufferedOutputByteStream
 
 @Suite(
-    .serialized,  // to limit the number of swift executable running.
+    // .serialized,  // to limit the number of swift executable running.
```
owenv: Should we delete this now?

Sources/Commands/Coverage/CoverageFormatOutput.swift
```diff
@@ -0,0 +1,110 @@
...
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import struct TSCBasic.StringError
+import struct Basics.AbsolutePath
+
+package struct CoverageFormatOutput: Encodable {
```
plemarquand, line 16: Actually, going by off of what is actually used, this whole struct can be reduced to just:

```swift
package struct CoverageFormatOutput: Encodable {
    private let paths: [CoverageFormat: AbsolutePath]

    package init(data: [CoverageFormat: AbsolutePath]) {
        self.paths = data
    }

    package func encode(to encoder: Encoder) throws {
        var container = encoder.container(keyedBy: DynamicCodingKey.self)
        for (format, path) in self.paths.sorted(by: { $0.key < $1.key }) {
            try container.encode(path.pathString, forKey: DynamicCodingKey(format))
        }
    }

    private struct DynamicCodingKey: CodingKey {
        let stringValue: String
        var intValue: Int? { nil }

        init(_ format: CoverageFormat) {
            self.stringValue = format.rawValue
        }

        init?(stringValue: String) {
            self.stringValue = stringValue
        }

        init?(intValue: Int) {
            return nil
        }
    }
}
```

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1039,16 +1105,96 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
...
+        } else {
+            try localFileSystem.createDirectory(outputPath, recursive: true)
+        }
+
+
+        var args = [
+            llvmCovPath.pathString,
+            "show",
+            "--project-title=\(title) Coverage Report",
+            "--instr-profile=\(profData.pathString)",
```
plemarquand, line 1163: This argument differs quite a bit from how it is set in `exportCodeCovAsJSON`, is that intentional?

`"-instr-profile=\(try await buildSystem.codeCovDataFile(for: productsBuildParameters))",`

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -983,23 +1000,68 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
...
+                    // Get all production source files from test targets
+                    let buildSystem = try await swiftCommandState.createBuildSystem()
+                    let packageGraph = try await buildSystem.getPackageGraph()
+
+                    let sourceFiles = try await getProductionSourceFiles(
+                        testProducts: testProducts,
+                        packageGraph: packageGraph,
+                    )
+                    for product in testProducts {
+                        let coveragaHtmlReportPath = try await generateHtmlCoverageReport(
```
plemarquand: This is creating one report per testProduct, but I think test coverage data should be merged together across all the products to create a single report instead of multiple. This is how VS Code does it.

plemarquand (changes requested):

Sources/Commands/SwiftTestCommand.swift
bkhouri: It's production code as we are trying to exclude test files.

Tests/CommandsTests/TestCommandTests.swift
bkhouri: Ack

Sources/Commands/SwiftTestCommand.swift
bkhouri: I will update

Sources/Commands/SwiftTestCommand.swift
bkhouri: nice catch.

Sources/Commands/Coverage/CoverageFormatOutput.swift
bkhouri: These must have been used in an earlier implementation. I'll remove them.

Sources/Commands/Coverage/CoverageFormatOutput.swift
bkhouri: thanks for this.

Sources/Commands/SwiftTestCommand.swift
bkhouri: yes. I'll remove.

Sources/Commands/SwiftTestCommand.swift
bkhouri: Yes. I'll remove.

Sources/Commands/Coverage/CoverageOptions.swift
bkhouri: If by `synthesized`, you mean `extension CoverageFormat: Comparable {}`, then it's not possible with the current definition.

```
/Users/bkhouri/Documents/git/public/swiftlang/swift-package-manager/Sources/Commands/Coverage/CoverageOptions.swift:28:1: error: type 'CoverageFormat' does not conform to protocol 'Comparable'
 14 | import struct Basics.Diagnostic
 15 | 
 16 | package enum CoverageFormat: String, ExpressibleByArgument, CaseIterable {
    |                              `- note: enum declares raw type 'String', preventing synthesized conformance of 'CoverageFormat' to 'Comparable'
 17 |     case html
 18 |     case json
    :
 26 | }
 27 | 
 28 | extension CoverageFormat: Comparable {}
    | |- error: type 'CoverageFormat' does not conform to protocol 'Comparable'
    | `- note: add stubs for conformance
 29 | //     package static func < (lhs: CoverageFormat, rhs: CoverageFormat) -> Bool {
 30 | //         return lhs.rawValue < rhs.rawValue
```

And we need the `CoverageFormat` to conform to `String` so Swift Argument Parser can convert a String to this value.

I will reorder the cases though.

Sources/Commands/SwiftTestCommand.swift
bkhouri: Nice catch.. I'll also add end-to-end tests to ensure we don't regress on this behaviour.

Sources/Commands/SwiftTestCommand.swift
bkhouri: Something with my implementation wasn't sitting well with me.  Your suggestion is a much better and cleaner implementation. I'll implement this suggestion.

Sources/Commands/SwiftTestCommand.swift
bkhouri: I refactored this.

Sources/Commands/SwiftTestCommand.swift
bkhouri: updated.

bkhouri:
@swift-ci test

bkhouri:
@swift-ci test windows

Sources/Commands/SwiftTestCommand.swift
bkhouri: Resolved.

Sources/Commands/SwiftTestCommand.swift
bkhouri: Both were pointing to the same file, I updated the JSON one to match the HTML.

Sources/Commands/SwiftTestCommand.swift
bkhouri: I made an update, but I still set the default path if the output directory cannot be computed. Though, a warning is now is emitted.

Sources/Commands/SwiftTestCommand.swift
bkhouri: I update this functions implementation.

Sources/Commands/SwiftTestCommand.swift
bkhouri: I update the function name to be more descriptive

bkhouri:
@swift-ci test

bkhouri:
@swift-ci test windows

bkhouri:

bkhouri:
@swift-ci test

bkhouri:
@swift-ci test windows

bkhouri:

bkhouri:

Sources/Commands/SwiftTestCommand.swift
bkhouri: My intent is not to execute the automated tests, but fail to generate the coverage report because of a user error, which is why I want to fall back to setting a default output. 

Alternatively, we can error out early if the output directory cannot be computer.

.github/workflows/automerge.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           - source_branch: release/6.4.x
             destination_branch: main
     name: Create PR to merge `${{ matrix.source_branch }}` into `${{ matrix.destination_branch }}`
-    uses: swiftlang/github-workflows/.github/workflows/create_automerge_pr.yml@0.0.13
```
owenv, line 21: Did a bad rebase pull this in?

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -982,24 +999,76 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
...
+
+                    // Get all production source files from test targets
+                    let buildSystem = try await swiftCommandState.createBuildSystem()
+                    let packageGraph = try await buildSystem.getPackageGraph()
+
+                    let sourceFiles = try await coverageReportSourceFiles(
+                        testProducts: testProducts,
+                        packageGraph: packageGraph,
+                    )
+                    let coveragaHtmlReportPath = try await generateHtmlCoverageReport(
```
owenv, line 1051: minor typo: "coveragaHtmlReportPath" -> "coverageHtmlReportPath"

owenv (approved):

plemarquand (approved):

PR #9032, opened 2025-08-15
Fix: Generate tests for executable/tool with `--enable-swift-testing`

JeffESchmitz:
Updated `InitPackage` to handle `.executable` and `.tool` types when generating test files with SwiftTesting. Added corresponding tests to verify correct test file content for these package types.

### Motivation:

The `swift package init` command did not generate a `Tests` directory for executable or tool packages when the `--enable-swift-testing` flag was used. This was caused by logic in `Sources/Workspace/InitPackage.swift` that incorrectly skipped test generation for these package types. This change fixes that behavior, ensuring that tests are created as expected.

### Modifications:

- In `Sources/Workspace/InitPackage.swift`, the `writeTests()` function was modified to proceed with test generation for `.executable` and `.tool` package types.
- In the same file, the `writeTestFileStubs()` function was updated to include `.executable` and `.tool` types, allowing the creation of standard library-style test files for them.

### Result:

After this change, running `swift package init --type executable --enable-swift-testing` or `swift package init --type tool --enable-swift-testing` correctly generates a `Tests` directory with sample test files, aligning the tool's behavior with its intended functionality.

#### Reproducing the Issue (on `main` branch):

1. Create a temporary directory: `mkdir -p /tmp/test-repro && cd /tmp/test-repro`
2. Run `swift package init --type executable --enable-swift-testing`.
3. Verify that no `Tests` directory is created by running `ls -F`.

#### Verifying the Fix (on this branch):

1. Build the package manager: `swift build`
2. Find the built executable: `find .build -name swift-package`
3. Create a temporary directory: `mkdir -p /tmp/test-fix && cd /tmp/test-fix`
4. Run the newly built package manager to init a project: `../.build/arm64-apple-macosx/debug/swift-package init --type executable --enable-swift-testing` (adjust path as needed).
5. Verify that a `Tests` directory is now created by running `ls -F`.

PR #8995, opened 2025-08-05
Package Init - Generate tests for .executable, .tool

plemarquand:
### Motivation:

Generate tests for the `.executable` and `.tool` template types when generating a new package with `swift package init`

### Modifications:

Add test targets to these types, and generate the templated test code. Also converts over the InitTests to swift-testing and uses parameterized tests to test the different types all produce tests as expected.

Issue: #8990

PR #9079, opened 2025-08-29
[Tests] Convert InitTests to Swift Testing

plemarquand:
Now that #9032 has landed, take the work from #8995 to convert the InitTests to Swift Testing and break it out in to its own PR.

### DIFF
--- a/Sources/_InternalTestSupport/SwiftTesting+Helpers.swift
+++ b/Sources/_InternalTestSupport/SwiftTesting+Helpers.swift
@@ -11,11 +11,17 @@
 import Basics
 import Testing
 
+// MARK: File System Helpers
+
+/// Verifies that a file exists at the specified path.
+///
+/// - Parameters:
+///   - path: The absolute path to check for file existence.
+///   - sourceLocation: The source location where the expectation is made.
 public func expectFileExists(
     at path: AbsolutePath,
     sourceLocation: SourceLocation = #_sourceLocation,
 ) {
-
     #expect(
         localFileSystem.exists(path),
         "Files '\(path)' does not exist.",
@@ -23,6 +29,12 @@ public func expectFileExists(
     )
 }
 
+/// Verifies that a file does not exist at the specified path.
+///
+/// - Parameters:
+///   - fixturePath: The absolute path to check for file non-existence.
+///   - comment: An optional comment to include in the failure message.
+///   - sourceLocation: The source location where the expectation is made.
 public func expectFileDoesNotExists(
     at fixturePath: AbsolutePath,
     _ comment: Comment? = nil,
@@ -41,6 +53,12 @@ public func expectFileDoesNotExists(
     )
 }
 
+/// Verifies that a file exists and is executable at the specified path.
+///
+/// - Parameters:
+///   - fixturePath: The absolute path to check for executable file existence.
+///   - comment: An optional comment to include in the failure message.
+///   - sourceLocation: The source location where the expectation is made.
 public func expectFileIsExecutable(
     at fixturePath: AbsolutePath,
     _ comment: Comment? = nil,
@@ -59,6 +77,11 @@ public func expectFileIsExecutable(
     )
 }
 
+/// Verifies that a directory exists at the specified path.
+///
+/// - Parameters:
+///   - path: The absolute path to check for directory existence.
+///   - sourceLocation: The source location where the expectation is made.
 public func expectDirectoryExists(
     at path: AbsolutePath,
     sourceLocation: SourceLocation = #_sourceLocation,
@@ -70,6 +93,11 @@ public func expectDirectoryExists(
     )
 }
 
+/// Verifies that a directory does not exist at the specified path.
+///
+/// - Parameters:
+///   - path: The absolute path to check for directory non-existence.
+///   - sourceLocation: The source location where the expectation is made.
 public func expectDirectoryDoesNotExist(
     at path: AbsolutePath,
     sourceLocation: SourceLocation = #_sourceLocation,
@@ -81,6 +109,15 @@ public func expectDirectoryDoesNotExist(
     )
 }
 
+// MARK: Error Helpers
+
+/// Verifies that an expression throws a `CommandExecutionError`.
+///
+/// - Parameters:
+///   - expression: The expression to evaluate.
+///   - message: An optional description of the failure.
+///   - sourceLocation: The source location where the expectation is made.
+///   - errorHandler: A closure that's called with the error if the expression throws.
 public func expectThrowsCommandExecutionError<T>(
     _ expression: @autoclosure () async throws -> T,
     _ message: @autoclosure () -> Comment = "",
@@ -99,7 +136,13 @@ public func expectThrowsCommandExecutionError<T>(
     }
 }
 
-/// An `async`-friendly replacement for `XCTAssertThrowsError`.
+/// An async-friendly replacement for `XCTAssertThrowsError` that verifies an expression throws an error.
+///
+/// - Parameters:
+///   - expression: The expression to evaluate that should throw an error.
+///   - message: An optional failure message to display if the expression doesn't throw.
+///   - sourceLocation: The source location where the expectation is made.
+///   - errorHandler: A closure that's called with the error if the expression throws.
 public func expectAsyncThrowsError<T>(
     _ expression: @autoclosure () async throws -> T,
     _ message: @autoclosure () -> Comment? = nil,
--- a/Tests/WorkspaceTests/InitTests.swift
+++ b/Tests/WorkspaceTests/InitTests.swift
@@ -14,25 +14,71 @@ import Basics
 import _InternalTestSupport
 import PackageModel
 import Workspace
-import XCTest
+import Testing
+import SPMBuildCore
 
-final class InitTests: XCTestCase {
+/// Tests for the `InitPackage` functionality, which creates new Swift packages with different configurations.
+struct InitTests {
+    /// The target triple for the current platform, used to locate build products.
+    static let targetTriple: Triple = {
+        do {
+            return try UserToolchain.default.targetTriple
+        } catch {
+            fatalError("Failed to determine target triple: \(error)")
+        }
+    }()
 
-    // MARK: TSCBasic package creation for each package type.
-    
-    func testInitPackageEmpty() throws {
-        try testWithTemporaryDirectory { tmpPath in
+    // MARK: - Helper Methods
+
+    /// Asserts that the package under test builds successfully.
+    public func expectBuilds(
+        _ path: AbsolutePath,
+        buildSystem: BuildSystemProvider.Kind,
+        configurations: Set<BuildConfiguration> = [.debug, .release],
+        extraArgs: [String] = [],
+        Xcc: [String] = [],
+        Xld: [String] = [],
+        Xswiftc: [String] = [],
+        env: Environment? = nil,
+        sourceLocation: SourceLocation = #_sourceLocation,
+    ) async {
+        for conf in configurations {
+            await #expect(throws: Never.self, sourceLocation: sourceLocation) {
+                try await executeSwiftBuild(
+                    path,
+                    configuration: conf,
+                    extraArgs: extraArgs,
+                    Xcc: Xcc,
+                    Xld: Xld,
+                    Xswiftc: Xswiftc,
+                    env: env,
+                    buildSystem: buildSystem
+                )
+            }
+        }
+    }
+
+    /// Creates a test package with the specified configuration and verifies its structure.
+    private func createAndVerifyPackage(
+        packageType: InitPackage.PackageType,
+        name: String = "Foo",
+        supportedTestingLibraries: Set<TestingLibrary> = [.xctest],
+        buildSystem: BuildSystemProvider.Kind? = nil,
+        customVerification: ((AbsolutePath, String) throws -> Void)? = nil
+    ) async throws {
+        return try await testWithTemporaryDirectory { tmpPath in
             let fs = localFileSystem
-            let path = tmpPath.appending("Foo")
-            let name = path.basename
+            let path = tmpPath.appending(name)
+            let packageName = path.basename
             try fs.createDirectory(path)
-            
+
             // Create the package
             let initPackage = try InitPackage(
-                name: name,
-                packageType: .empty,
+                name: packageName,
+                packageType: packageType,
+                supportedTestingLibraries: supportedTestingLibraries,
                 destinationPath: path,
-                fileSystem: localFileSystem
+                fileSystem: fs
             )
             var progressMessages = [String]()
             initPackage.progressReporter = { message in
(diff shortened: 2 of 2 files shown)


plemarquand:
@swift-ci test

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test

bripeticca:
@swift-ci please test windows

Sources/_InternalTestSupport/SwiftTesting+Helpers.swift
```diff
@@ -11,18 +11,30 @@
 import Basics
 import Testing
 
+// MARK: File System Helpers
+
+/// Verifies that a file exists at the specified path.
+///
+/// - Parameters:
+///   - path: The absolute path to check for file existence.
+///   - sourceLocation: The source location where the expectation is made.
```
bripeticca, line 20: praise: thank you for adding documentation to these files! this makes it a lot more readable.

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,454 +14,421 @@ import Basics
...
-                destinationPath: path,
-                fileSystem: localFileSystem
-            )
-            var progressMessages = [String]()
-            initPackage.progressReporter = { message in
-                progressMessages.append(message)
+import Testing
+import SPMBuildCore
+
+/// Tests for the `InitPackage` functionality, which creates new Swift packages with different configurations.
```
bripeticca, line 20: praise: another thank you for the comments/documentation here!

bripeticca (approved):

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

jakepetroules (approved):

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test

Sources/_InternalTestSupport/misc.swift
```diff
@@ -97,7 +97,15 @@ public func testWithTemporaryDirectory<Result>(
         .replacing(")", with: "")
         .replacing(".", with: "")
         .replacing(":", with: "_")
-    return try await withTemporaryDirectory(prefix: "spm-tests-\(cleanedFunction)") { tmpDirPath in
+
+    // Use shorter prefix on Windows to avoid MAX_PATH issues
+    #if os(Windows)
+    let prefix = "spm-\(abs(cleanedFunction.hashValue))"
```
bkhouri: suggestion: if we want to use short prefix on Windows, why don't we set `prefix = ""`?

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
+        buildSystem: BuildSystemProvider.Kind,
+        configurations: Set<BuildConfiguration> = [.debug, .release],
+        extraArgs: [String] = [],
+        Xcc: [String] = [],
+        Xld: [String] = [],
+        Xswiftc: [String] = [],
+        env: Environment? = nil,
+        sourceLocation: SourceLocation = #_sourceLocation,
+    ) async {
+        for conf in configurations {
```
bkhouri, line 47: suggestion: instead of providing `configurations` as a function argument, consider passing as a test parameter so it can be clearly tracked as it's own test. See https://github.com/swiftlang/swift-package-manager/blob/main/Tests/CommandsTests/PackageCommandTests.swift#L100-L112 as a sample test that add a cross matrix of build system and build configurations.  The name `SupportedBuildSystemOnAllPlatforms` is a bit misleading, but it currently is an array containing Native and Swift Build build systems.

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
-            // Not picky about the specific progress messages, just checking that we got some.
-            XCTAssertGreaterThan(progressMessages.count, 0)
-            
-            // Verify basic file system content that we expect in the package
+            #expect(progressMessages.count > 0, "Expected progress messages during package creation")
+
             let manifest = path.appending("Package.swift")
-            XCTAssertFileExists(manifest)
-            let manifestContents: String = try localFileSystem.readFileContents(manifest)
+            expectFileExists(at: manifest)
```
bkhouri, line 96: suggestion: consider changing this to a `try #require(...)`, which may mean we need to create an associated helper if one does not currently exist.  Using require makes more sense here since the following code requires it.

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
+        for packageType: InitPackage.PackageType,
+        at path: AbsolutePath,
+        name: String,
+        buildSystem: BuildSystemProvider.Kind
+    ) throws {
+        let expectedPath = path.appending(components: try buildSystem.binPath(for: BuildConfiguration.debug))
+
+        switch packageType {
+        case .library:
+            if buildSystem == .native {
```
bkhouri, line 129: suggestion: instead of using an `if` statement, consider using a `switch buildSystem` to ensure the tests are updated properly should entries to `BuildSystemProvider.Kind` be added or removed - this would ensure the tests are at least updated/looked at..

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
-            await XCTAssertBuilds(
-                path,
-                buildSystem: .native,
-            )
-            let triple = try UserToolchain.default.targetTriple
-            XCTAssertFileExists(path.appending(components: ".build", triple.platformBuildPathComponent, "debug", "Modules", "Foo.swiftmodule"))
+        case .executable, .tool:
+            expectFileExists(at: expectedPath.appending(executableName(name)))
+        case .empty, .buildToolPlugin, .commandPlugin, .macro:
+            // These types don't have specific build products to verify or are verified separately
```
bkhouri: question: do we want to support tests calling this?  this may provide a false positive.  Consider calling `Issue.record("Test implementation issue: \(packageType) don't have specific build products to verify.  Replace this function call with a different verification steps")` instead.

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
+    ) throws {
+        let testFile = path.appending("Tests").appending("\(name)Tests").appending("\(name)Tests.swift")
+        let testFileContents: String = try localFileSystem.readFileContents(testFile)
+
+        if hasSwiftTesting {
+            #expect(testFileContents.contains(#"import Testing"#))
+            #expect(testFileContents.contains(#"@Test func example() async throws"#))
+        } else {
+            #expect(!testFileContents.contains(#"import Testing"#))
+            #expect(!testFileContents.contains(#"@Test func example() async throws"#))
```
bkhouri: suggestion: can we also add `#expect(!testFileContent.contains("@Test")` in the event the template changes the test function name?

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
-            let triple = try UserToolchain.default.targetTriple
-            XCTAssertFileExists(path.appending(components: ".build", triple.platformBuildPathComponent, "debug", "Modules", "Foo.swiftmodule"))
-#endif
+        if hasXCTest {
+            #expect(testFileContents.contains(#"import XCTest"#))
+            #expect(testFileContents.contains("func testExample() throws"))
+
+            if hasSwiftTesting {
+                // When both are present, ensure XCTest content is properly formatted
+                #expect(testFileContents.contains("import XCTest"), "XCTest import should be present")
```
bkhouri: question: why does including SwiftTesting need to check for `import XCTest`?   This will be true based on line 157. and the `import Testing` is checked on line 149.

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
-            let name = path.basename
-            try fs.createDirectory(path)
+    /// Verifies plugin package contents.
+    private func verifyPluginPackage(
+        at path: AbsolutePath,
+        name: String,
+        isCommandPlugin: Bool
+    ) throws {
+        let manifest = path.appending("Package.swift")
+        expectFileExists(at: manifest)
```
bkhouri: suggestion: same as above, consider changing this to a `try #require(...)` call.

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
-            try initPackage.writePackageStructure()
+    /// Verifies the test file contents for a package.
+    private func verifyTestFileContents(
+        at path: AbsolutePath,
+        name: String,
+        hasSwiftTesting: Bool,
+        hasXCTest: Bool
+    ) throws {
+        let testFile = path.appending("Tests").appending("\(name)Tests").appending("\(name)Tests.swift")
+        let testFileContents: String = try localFileSystem.readFileContents(testFile)
```
bkhouri, line 150: suggestion: consider adding a `try #require(localFileSystem.exists(testFile))`

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
-                name: name,
-                packageType: .library,
-                supportedTestingLibraries: [.swiftTesting, .xctest],
-                destinationPath: path,
-                fileSystem: localFileSystem
-            )
-            try initPackage.writePackageStructure()
+        // Verify source file
+        let source = path.appending("Plugins", "\(name).swift")
+        expectFileExists(at: source)
```
bkhouri: suggestion: same as above, consider changing this to a try #require(...) call.

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
-            // Try building it
-            await XCTAssertBuilds(
-                path,
-                buildSystem: .native,
-            )
-            let triple = try UserToolchain.default.targetTriple
-            XCTAssertFileExists(path.appending(components: ".build", triple.platformBuildPathComponent, "debug", "Modules", "Foo.swiftmodule"))
-#endif
+        // Both plugin types should have Xcode extensions
+        #expect(sourceContents.contains("import XcodeProjectPlugin"))
```
bkhouri, line 200: question: is this valid on Linux and Windows as well?

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
-            // Create the package
-            let initPackage = try InitPackage(
-                name: name,
-                packageType: .tool,
-                supportedTestingLibraries: [.swiftTesting],
-                destinationPath: path,
-                fileSystem: localFileSystem
-            )
+    /// Tests creating an executable package named "main".
+    @Test(arguments: [BuildSystemProvider.Kind.native, .swiftbuild])
```
bkhouri: suggestion: consider replacing the value of `arguments` with  `SupportedBuildSystemOnAllPlatforms`.  This comment applies to all occurrences of an array of two items with `native` and `swiftbuild` build systems.

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
-            let testFileContents: String = try localFileSystem.readFileContents(testFile)
-            XCTAssertMatch(testFileContents, .contains(#"import Testing"#))
-            XCTAssertNoMatch(testFileContents, .contains(#"import XCTest"#))
-            XCTAssertMatch(testFileContents, .contains(#"@Test func example() async throws"#))
-            XCTAssertNoMatch(testFileContents, .contains("func testExample() throws"))
-        }
+    /// Tests creating packages with XCTest only.
+    @Test(arguments: [InitPackage.PackageType.library, .executable, .tool], [BuildSystemProvider.Kind.native, .swiftbuild])
+    func initPackageLibraryWithXCTestOnly(packageType: InitPackage.PackageType, buildSystem: BuildSystemProvider.Kind) async throws {
+        #if canImport(TestingDisabled)
```
bkhouri: suggestion: can we add a code comment that explains this compiler conditional and why it's needed?  Why does the `TestingDisable` module determine the build system under test?

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
+            let buildSys = buildSystem
+        #else
+            let buildSys: BuildSystemProvider.Kind? = nil
```
bkhouri: suggestion (possibly-blocking): emit a lot message stating we are overriding the build system specified as a test argument.  or maybe add the build system to the "failure messages" since it can be overridden based on the availability of `TestingDisabled`.  if this test fails, it may be confusing trying to reproduce as desk, especially if the test override the build system to `nil`, but the Swift Testing summary indicates we are testing build system X.

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
+            supportedTestingLibraries: [],
+            buildSystem: buildSys,
+            customVerification: { path, name in
+                let manifestContents: String = try localFileSystem.readFileContents(path.appending("Package.swift"))
+                #expect(!manifestContents.contains(#".testTarget"#))
+
+                expectDirectoryDoesNotExist(at: path.appending("Tests"))
+
+                #if canImport(TestingDisabled)
+                let expectedPath = path.appending(components: ".build", Self.targetTriple.platformBuildPathComponent, "debug", "Modules", "\(name).swiftmodule")
```
bkhouri: issue (blocking):  the `expectedPath` assume the native build system, which is not the same as the SwiftBuild build system.  it's also only looking at `debug` build configuration.  But doesn't this test also build `release` build configuration?   So why are we only looking for the `debug` swift module and there is no expectation on the `release` build configuration?

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
+            }
+        )
     }
 
-    // MARK: Special case testing
+    // MARK: - Special Case Tests
 
-    func testInitPackageNonc99Directory() async throws {
+    /// Tests creating a package in a directory with a non-C99 compliant name.
+    @Test(arguments: [BuildSystemProvider.Kind.native, .swiftbuild])
```
bkhouri: suggestion: consider passing the Build configuration as a test argument.  
```suggestion
    @Test(
        arguments: getBuildData(for: SupportedBuildSystemOnAllPlatforms),
    )
```

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
+    func initPackageNonc99Directory(buildSystem: BuildSystemProvider.Kind) async throws {
```
bkhouri: Providing the build configuration via test arguments, as suggestd above, will update the argument type as follows:
```suggestion
    func initPackageNonc99Directory(
        data: BuildData,
    ) async throws {
```

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
-            await XCTAssertBuilds(
-                packageRoot,
-                buildSystem: .native,
-            )
-            let triple = try UserToolchain.default.targetTriple
-            XCTAssertFileExists(packageRoot.appending(components: ".build", triple.platformBuildPathComponent, "debug", "Modules", "some_package.swiftmodule"))
+            await expectBuilds(packageRoot, buildSystem: buildSystem)
+
+            // Assert that the expected build products exist
+            let expectedPath = packageRoot.appending(components: try buildSystem.binPath(for: BuildConfiguration.debug))
```
bkhouri: issue (blocking): Since the test currently build both debug and release build configuration, we should also assert on the expected behaviour for the `release` build configuration.

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -14,458 +14,422 @@ import Basics
...
+
+            // Verify the module name is properly mangled
+            if buildSystem == .native {
```
bkhouri: suggestion: change this from an `if` to a `swift` so we get compilation time error should the build system kinds be updated. 

I used the following when I wanted to support a subset of build systems in a test 


```swift
switch buildSystem {
    case .native:
        #expect("native build expectations".count > 1)
    case .swiftbuild:
        #expect("swift build expectations".count > 1)
    case .xcode:
        Issue.record("Test expectations have not been implemented.")
}
```

Tests/WorkspaceTests/InitTests.swift
```diff
@@ -474,16 +438,15 @@ final class InitTests: XCTestCase {
                 fileSystem: localFileSystem
             )
             try initPackage.writePackageStructure()
-            
-            await XCTAssertBuilds(
-                packageRoot,
-                buildSystem: .native,
-            )
+
+            await expectBuilds(packageRoot, buildSystem: buildSystem)
```
bkhouri, line 425: suggestion: consider adding the build configuration as a test data

bkhouri (changes requested):
Thanks for migrating the Package Init Tests to Swift Testing.   I have some blocking comments, hence the "requests changes". I'm happy to have a discussion if you feel the comments are non-blocking.

plemarquand:
@swift-ci test

Tests/WorkspaceTests/InitTests.swift
plemarquand: This is for the case where you have both swift testing and XCTest explicitly enabled when you create the template

plemarquand:
@swift-ci test

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci please test windows

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test

Issue #8129, opened 2024-11-20
Add a way to attach a debugger to runs of `swift test`


daniel-grumberg:
### Description

Currently there is no existing workflow for attaching a debugger on the command line to an invocation of `swift test`. the workflow I use today is less than ideal, I run: `lldb -w -n swiftpm-testing-helper` for Swift Testing tests or `lldb -w -n xctest` for XCTest tests.

### Expected behavior

It would be nice to have a couple of couple of extra options to swift test to attach a debugger, maybe:
```
# Run tests normally
swift test

# Run tests under lldb
swift test --debugger

# Run tests with a debugger and pause initially (e.g. to set some breakpoints before the tests run)
swift test --debugger --pause-on-launch
```

This would lead to feature parity with `swift run --debugger`.

plemarquand:
This would be great. It should be made to function the same as `swift running --debugger`

PR #9085, opened 2025-09-02
Add LLDB debugger support to swift-test command

plemarquand:
Introduces interactive debugging capabilities to `swift test` via a `--debugger` flag enabling developers to failing tests directly within LLDB.

This adds debugging parity with `swift run --debugger` which allows users to debug their executables directly.

When launching lldb this implenentation uses `execv()` to replace the `swift-test` process with LLDB. This approach avoids process hierarchy complications and ensures LLDB has full terminal control for interactive debugging features.

When there is only one type of test the debugger creates a single LLDB target configured specifically for that framework. For XCTest, this means targeting the test bundle with xctest as the executable, while Swift Testing targets the test binary directly with appropriate command-line arguments.

When both testing frameworks have tests available we create multiple LLDB targets within a single debugging session. A Python script automatically switches targets as the executable exits, which lets users debug both types of tests spread across two executables in the same session. The Python script also maintains breakpoint persistence across target switches, allowing you to set a breakpoint for either test type no matter the active target.

Note that the CI smoke tests use an old version of LLDB that ships Swift 5.9.2. The `--debugger` feature relies on the version of LLDB that ships alongside swift-package-manager in the latest toolchain, and so many supporting tests are skipped when running in the smoke test jobs in CI.

Issue: #8129


plemarquand:
@swift-ci please test

pmattos (approved):

Tests/CommandsTests/TestCommandTests.swift
```diff
@@ -1172,4 +1172,263 @@ struct TestCommandTests {
...
+
+                #expect(
+                    stdout.contains("failbreak breakpoint set -s Testing -n \"failureBreakpoint()\""),
+                    "Expected Swift Testing failure breakpoint setup, got stdout: \(stdout), stderr: \(stderr)",
+                )
+            }
+        }
+
+        @Test(arguments: SupportedBuildSystemOnAllPlatforms)
+        func debuggerFlagWithBothTestingSuites(buildSystem: BuildSystemProvider.Kind) async throws {
```
pmattos, line 1899: **question:** Would there be an easy way to actually _interact_ with LLDB during tests (eg,`stdin`)? For instance, send it some commands and check back the expected results? Maybe you did it already... and I missed it :)

plemarquand: One of the reasons this is still in draft is I want to address that. I think I'd have to add supplemental commands to the file that gets passed to LLDB that contains its setup commands. That would let us set specific breakpoints in tests, though I still haven't figured out how to inspect the output since we replace the running process with `execv`.

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -216,6 +216,11 @@ struct TestCommandOptions: ParsableArguments {
           help: "Enable code coverage.")
     var enableCodeCoverage: Bool = false
 
+    /// Launch tests under LLDB debugger.
+    @Flag(name: .customLong("debugger"),
+          help: "Launch tests under LLDB debugger.")
+    var shouldLaunchInLLDB: Bool = false
```
pmattos, line 237: **nitpick:** The output from `swift run --help` phrases like this:
```
--debugger  Launch the executable in a debugger session.
```
Would be nice if both commands followed similar wording, no?

plemarquand:
@swift-ci please test

Tests/CommandsTests/TestCommandTests.swift
```diff
@@ -1172,4 +1177,268 @@ struct TestCommandTests {
...
+
+    @Suite
+    struct LLDBTests {
+        @Test(arguments: SupportedBuildSystemOnAllPlatforms)
+        func lldbWithParallelThrowsError(buildSystem: BuildSystemProvider.Kind) async throws {
+            let args = args(["--debugger", "--parallel"], for: buildSystem)
+            let command = try #require(SwiftTestCommand.parseAsRoot(args) as? SwiftTestCommand)
+            let (state, outputStream) = try commandState()
+
+            let error = await #expect(throws: ExitCode.self) {
```
bkhouri, line 1753: suggestion: can we replace this `#expect` and subsequent calls with a helper function similar to [`expectThrowsCommandExecutionError`](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/_InternalTestSupport/SwiftTesting%2BHelpers.swift#L84-L100) or [`expectAsyncThrowsError`](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/_InternalTestSupport/SwiftTesting%2BHelpers.swift#L102-L117)?

This comment applies to all similar instances of newly added tests.

Tests/CommandsTests/TestCommandTests.swift
```diff
@@ -1172,4 +1177,268 @@ struct TestCommandTests {
...
+                },
+                createPackagePath: false
+            )
+            return (state, outputStream)
+        }
+    }
+}
+
+fileprivate extension BuildConfiguration {
+    var args: [String] {
```
bkhouri: This code logic is duplicated here: https://github.com/swiftlang/swift-package-manager/blob/a8349375b1e32360f007eaedf641dd6cf3b3fa45/Sources/_InternalTestSupport/misc.swift#L525-L531

Can we refactor to have the argument logic is defined in a single location?

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

Tests/CommandsTests/TestCommandTests.swift
plemarquand: I don't think so, `expectThrowsCommandExecutionError` is for validating the command fails when run in a new process, and this more lightweight approach uses `command.run` in process.

`expectAsyncThrowsError` doesn't let you specify the error type and so requires a cast after the fact. In fact I don't think this method is required at all. There is a version of `expectThrowsCommandExecutionError` in APIDiffTests.swift that handles command execution errors in the idiomatic swift testing way.

I've replaced the implementation of `expectThrowsCommandExecutionError` in SwiftTesting+Helpers with the one from APIDiffTests, and removed the copy in APIDiffTests in https://github.com/swiftlang/swift-package-manager/pull/9125

Tests/CommandsTests/TestCommandTests.swift
plemarquand: Done

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test Linux

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

Sources/Commands/Utilities/TestingSupport.swift
```diff
@@ -276,6 +301,537 @@ enum TestingSupport {
...
+    /// This method is used when running Swift Package Manager's own tests to validate
+    /// debugger functionality. It launches LLDB without stdin attached, which causes
+    /// LLDB to execute its startup commands and then exit, allowing the test to capture
+    /// and validate the output.
+    ///
+    /// - Parameters:
+    ///   - lldbPath: Path to the LLDB executable
+    ///   - args: Command line arguments for LLDB
+    /// - Throws: Process execution errors
+    private func runLLDBForTesting(lldbPath: AbsolutePath, args: [String]) throws {
```
jakepetroules: This should be `async` in order to avoid using blocking overloads in `AsyncProcess`.

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci please test windows

Tests/CommandsTests/TestCommandTests.swift
plemarquand: I pushed this halfway there. There is a test in SwiftPM that will do a `run` in LLDB and validate the tests execute as expected, but we don't inject commands as that would require simulating interactivity which is much more difficult.

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:

plemarquand:

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

Sources/Commands/Utilities/TestingSupport.swift
```diff
@@ -418,6 +474,575 @@ extension ToolsVersion {
...
+    private func getModulePath(for target: DebuggableTestSession.Target) throws -> AbsolutePath {
+        switch target.kind {
+        case .xctest(let bundlePath):
+            guard buildParameters.triple.isDarwin() else {
+                return bundlePath
+            }
+            guard let name = bundlePath.components.last?.replacing(".xctest", with: "") else {
+                return bundlePath
+            }
+            let relativePath = try RelativePath(validating: "Contents/MacOS/\(name)")
```
owenv: I think if you always look at the binaryPath of the BuiltTestProduct it will compute this for you correctly. In general I've been trying to slowly refactor the testing code so it always passes around BuiltTestProducts instead of raw paths

Sources/Commands/Utilities/TestingSupport.swift
```diff
@@ -418,6 +474,575 @@ extension ToolsVersion {
...
+            return binaryPath
+        }
+    }
+
+    /// Creates a Python script that handles automatic target switching
+    private func createTargetSwitchingScript(in scratchDir: AbsolutePath) throws -> AbsolutePath {
+        let scriptPath = scratchDir.appending("target_switcher.py")
+
+        let pythonScript = """
+# target_switcher.py
```
owenv, line 982: It would be nice to break this out into a resource file, but that might take a bit of refactoring of the CMake build

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -223,6 +223,11 @@ struct TestCommandOptions: ParsableArguments {
           help: "Determines whether testing measures code coverage.")
     var enableCodeCoverage: Bool = false
 
+    /// Launch tests under LLDB debugger.
+    @Flag(name: .customLong("debugger"),
```
owenv, line 235: I expect it'd be uncontroversial but we should technically run an evolution proposal for non-experimental options like this

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -559,6 +575,143 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
...
+        swiftCommandState: SwiftCommandState
+    ) async throws {
+        guard !testProducts.isEmpty else {
+            throw DebuggerError.noTestProducts
+        }
+
+        let toolchain = try swiftCommandState.getTargetToolchain()
+
+        let xctestEnabled = options.testLibraryOptions.isEnabled(.xctest, swiftCommandState: swiftCommandState)
+        lazy var testEntryPointPath = testProducts.lazy.compactMap(\.testEntryPointPath).first
```
owenv: This might be copied from elsewhere but it seems odd that it's `lazy`

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

Sources/_InternalTestSupport/SwiftTesting+Helpers.swift
```diff
@@ -281,3 +283,28 @@ private func _expectThrowsCommandExecutionError<R, T>(
...
+///   - retryInterval: Time to wait between checks (default: 50 milliseconds)
+/// - Returns: True if the string was found within the timeout period
+public func waitForOutputStreamToContain(
+    _ outputStream: BufferedOutputByteStream,
+    _ needle: String,
+    timeout: TimeInterval = 3.0,
+    retryInterval: TimeInterval = 0.05
+) async throws -> Bool {
+    let startTime = Date()
+    while Date().timeIntervalSince(startTime) < timeout {
```
jakepetroules: Always use a monotonic clock for this rather than Date -- see https://developer.apple.com/documentation/swift/continuousclock

Sources/_InternalTestSupport/SwiftTesting+Helpers.swift
```diff
@@ -281,3 +283,28 @@ private func _expectThrowsCommandExecutionError<R, T>(
...
+        if outputStream.bytes.description.contains(needle) {
+            return true
+        }
+
+        try await Task.sleep(nanoseconds: UInt64(retryInterval * 1_000_000_000))
```
jakepetroules: Use `try await Task.sleep(for: retryInterval)`

Sources/_InternalTestSupport/SwiftTesting+Helpers.swift
```diff
@@ -281,3 +283,28 @@ private func _expectThrowsCommandExecutionError<R, T>(
...
+/// - Parameters:
+///   - outputStream: The output stream to check
+///   - needle: The string to search for in the output stream
+///   - timeout: Maximum time to wait for the string to appear (default: 3 seconds)
+///   - retryInterval: Time to wait between checks (default: 50 milliseconds)
+/// - Returns: True if the string was found within the timeout period
+public func waitForOutputStreamToContain(
+    _ outputStream: BufferedOutputByteStream,
+    _ needle: String,
+    timeout: TimeInterval = 3.0,
```
jakepetroules: Use `timeout: Duration = .seconds(3)`

Sources/_InternalTestSupport/SwiftTesting+Helpers.swift
```diff
@@ -281,3 +283,28 @@ private func _expectThrowsCommandExecutionError<R, T>(
...
+    retryInterval: TimeInterval = 0.05
```
jakepetroules: Use `retryInterval: Duration = .seconds(0.05)` or `retryInterval: Duration = .milliseconds(50)`

Sources/Commands/Utilities/TestingSupport.swift
```diff
@@ -418,6 +474,566 @@ extension ToolsVersion {
...
 
+/// A safe wrapper around TSCBasic.exec that resets signal masks and
+/// configures file descriptors before replacing the current process.
+///
+/// Dispatch disables almost all asynchronous signals on its worker threads.
+/// When called from async context, we must reset signal handlers and unblock
+/// all signals before exec so the child process starts with clean state.
+/// We also mark all non-standard file descriptors as close-on-exec to avoid
+/// leaking them to the new process.
+func safeExec(path: String, args: [String], observabilityScope: ObservabilityScope) throws -> Never {
```
jakepetroules, line 485: Is there a reason we absolutely must _replace_ the current process instead of using the normal subprocess spawning via AsyncProcess?

This is also something SwiftSubprocess doesn't support yet although it's been proposed: https://github.com/swiftlang/swift-subprocess/issues/251

plemarquand: Its not required, but it simplifies things quite a bit. If we spawned a subprocess we'd need to forward stdin/stdout/stderr, signals like CTRL+C as well as switching the terminal mode. 

Ultimately I just followed the existing pattern used by `swift run --debugger` which also calls `safeExec`.

If https://github.com/swiftlang/swift-subprocess/issues/251 is adopted I do think it would make sense to revisit this and adopt it in both commands.

Sources/Commands/SwiftTestCommand.swift
dschaefer2: Great point. It would also help highlight the advancement in this area.

plemarquand: Pitch is now available here https://forums.swift.org/t/pitch-debugging-swift-tests-with-lldb/86413

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test macos platform

plemarquand:

plemarquand:
@swift-ci test windows

owenv (approved):

PR #9096, opened 2025-09-05
[WIP] Emit output from SwiftBuild tasks

bripeticca:
WIP: an investigation into what logs/diagnostics are being emitted currently for SwiftBuild, and what it may be missing.

**TODO:**
- Determine which tasks should emit output
- Formalize verbosity levels for each type of task output, if necessary
- Investigate progress animation and the text output compared to native build system


jakepetroules:
One thing to keep in mind with verbosity is that we'll need to buffer certain events - for example you'll receive many events related to diagnostics and output for a single task, interleaved with other tasks, and so we need to buffer output and diagnostics until we see a "task ended" event, at which point we can emit content for that task to the log in the right order.

Right now it's just emitting events as they come in, leading to lots of confusing and out-of-order output.

bripeticca:
> One thing to keep in mind with verbosity is that we'll need to buffer certain events - for example you'll receive many events related to diagnostics and output for a single task, interleaved with other tasks, and so we need to buffer output and diagnostics until we see a "task ended" event, at which point we can emit content for that task to the log in the right order.
> 
> Right now it's just emitting events as they come in, leading to lots of confusing and out-of-order output.

Excellent point @jakepetroules, thanks for the heads up! I'll be sure to keep this in mind going forward :)

PR #9097, opened 2025-09-06
Use .app to detect Xcode toolchains rather than .xctoolchain

jakepetroules:
Open Source macOS toolchains also use the .xctoolchain suffix, so this isn't the right way to detect Xcode toolchains specifically.


jakepetroules:
@swift-ci test

dschaefer2:
I wonder if there are other .app's out there that include an OSS xctoolchain. Also, if you just have the Command Line Tools and no Xcode, you still get an Xcode toolchain, no? Where is it located? It seems you can look at the contents of the xctoolchain root plist file to get a better hint, maybe?

jakepetroules:
> I wonder if there are other .app's out there that include an OSS xctoolchain.

That's a really interesting point w.r.t. possible third party IDEs, although I'm not too concerned about that at this particular point in time compared to making today's expected workflows work correctly.

> Also, if you just have the Command Line Tools and no Xcode, you still get an Xcode toolchain, no? Where is it located? It seems you can look at the contents of the xctoolchain root plist file to get a better hint, maybe?

I don't think CLT will work at all with the current setup (we probably need a new enum case for it) and it hasn't been tested; we'll need to revisit it at some point.

jakepetroules:
@swift-ci test windows

jakepetroules:
@swift-ci test windows

PR #9112, opened 2025-09-09
Make sure we don't get backslashes in the Prebuilts Manifest

dschaefer2:
We are running into cases where, if the Windows prebuilts build runs last, it's the one producing the prebuilts manifest. Since we're using RelativePath to for the library include paths, it's resulting in Windows paths being used and that fails on the other platforms. This change makes sure we always use forward slashes which will work on all platforms.

Note to do this we're copying over the manifest types to the BuildPrebuilts script temporarily to so we don't need to make changes in 6.2.0. We'll revisit this for the next release where we probably should use per platform JSON files.

### DIFF
--- a/Sources/Workspace/Workspace+Prebuilts.swift
+++ b/Sources/Workspace/Workspace+Prebuilts.swift
@@ -54,7 +54,7 @@ extension Workspace {
             public let name: String
             public var products: [String]
             public var cModules: [String]?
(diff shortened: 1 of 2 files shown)


dschaefer2:
@swift-ci please test macOS

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

PR #9118, opened 2025-09-10
[Bootstrap] Add an extra argument to set a library path.

al45tair:
During Swift PR testing, we run the SwiftPM tests.  Unfortunately, the SwiftPM tests always try to link with the system version of the Swift runtime libraries.  This would be fine, except that Swift Testing has been built against the *new* runtimes and as a result has ended up referencing symbols from them that might not be present in the system runtime libraries.

To fix this, add an extra argument to allow us to add to the `DYLD_LIBRARY_PATH` setting SwiftPM uses to run its tests.

rdar://160273124

### DIFF
--- a/Utilities/bootstrap
+++ b/Utilities/bootstrap
@@ -168,6 +168,9 @@ def add_build_args(parser):
         action="store_true",
         help="whether to link to the llbuild framework")
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


al45tair:
Testing this solution on my failing PR.

PR #9127, opened 2025-09-11
[6.2] Make sure we don't get backslashes in the Prebuilts Manifest

dschaefer2:
Cherry pick of #9112 from main.

We are running into cases where, if the Windows prebuilts build runs last, it's the one producing the prebuilts manifest. Since we're using RelativePath to for the library include paths, it's resulting in Windows paths being used and that fails on the other platforms. This change makes sure we always use forward slashes which will work on all platforms. We use strings for the manifest and correct the slashes before writing them in the generator. We then convert them back to RelativePaths on the client side.


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

PR #9128, opened 2025-09-11
6.2.1 fix prebuilts slashes

dschaefer2:
Cherry pick of #8994 and #9127 to 6.2.1

8994 added the ability to build prebuilts and pass in the swift-syntax versions to build for. This was added to 6.2 after the 6.2.1 branch was created. We should make sure 6.2.1 has the same features in case we want to build the prebuilts from that branch. It also avoids a merge conflict with 9127.

9127 Fixes a bug where if Windows prebuilts built last, the include path to the C shims had backslashes in them instead of slashes. This forces them the other way. That unfortunately breaks the signature on the Windows client so this fixes the client to switch them back after signature verification.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -785,6 +785,9 @@ let package = Package(
                 .product(name: "ArgumentParser", package: "swift-argument-parser"),
                 "Basics",
                 "Workspace",
+            ],
+            exclude: [
+                "build.sh"
             ]
         ),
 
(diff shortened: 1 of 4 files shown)


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test linux

PR #9129, opened 2025-09-11
Expose a SwiftPM BSP interface based on Swift Build

owenv:
Add a `swift package bsp` command which builds on the Swift Build BSP interface to provide preparation + compiler args for packages as an out of process build server.


owenv:
This is very much a WIP right now, it just exposes the bare minimum functionality so we can discuss API design

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -767,8 +769,8 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
             buildProductsPath: ddPathPrefix + "/Products",
             buildIntermediatesPath: ddPathPrefix + "/Intermediates.noindex",
             pchPath: ddPathPrefix + "/PCH",
-            indexRegularBuildProductsPath: nil,
-            indexRegularBuildIntermediatesPath: nil,
+            indexRegularBuildProductsPath: ddPathPrefix + "/Index/Products",
+            indexRegularBuildIntermediatesPath: ddPathPrefix + "/Index/Intermediates.noindex",
```
bnbarham: The name is a little confusing here, but FWIW this is actually meant to be swapped for the index arena 😅 - the build products/intermediate paths would be the index arena and the "index regular" paths the regular build. It's also only used for adding an overlay to fallback to the build, which we don't necessarily need.

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
```diff
@@ -0,0 +1,192 @@
...
+
+    private func waitForBuildSystemUpdates(request: WorkspaceWaitForBuildSystemUpdatesRequest) async -> SWBBuildServerProtocol.VoidResponse {
+        await packageLoadingQueue.async {}.valuePropagatingCancellation
+        return VoidResponse()
+    }
+
+    public func scheduleRegeneratingBuildDescription() {
+        packageLoadingQueue.async { [buildSystem, session] in
+            do {
+                // TODO: should the PIF update API be exposed as API on SWBBuildServer somehow so we don't need to talk to the session directly?
```
bnbarham: The PIF path could be watched by the underlying server, which could then reload itself?

Sources/Commands/PackageCommands/BuildServer.swift
```diff
@@ -0,0 +1,77 @@
...
+        abstract: "Launch a build server for Swift Packages"
+    )
+
+    @OptionGroup(visibility: .hidden)
+    var globalOptions: GlobalOptions
+
+    func run(_ swiftCommandState: SwiftCommandState) async throws {
+        // Dup stdout and redirect the fd to stderr so that a careless print()
+        // will not break our connection stream.
+        let realStdout = dup(STDOUT_FILENO)
```
jakepetroules: We depend on Swift System, right? I suggest using FileDescriptor.duplicate cover API and not going straight down to POSIX.

IIRC there's also some oddities with the Windows versions of dup which System should be able to help smooth over.

Sources/Commands/PackageCommands/BuildServer.swift
```diff
@@ -0,0 +1,77 @@
...
+            protocol: bspRegistry,
+            inFD: FileHandle.standardInput,
+            outFD: realStdoutHandle,
+            inputMirrorFile: nil,
+            outputMirrorFile: nil
+        )
+
+        guard let buildSystem = try await swiftCommandState.createBuildSystem() as? SwiftBuildSystem else {
+            print("Build server requires --build-system swiftbuild")
+            Self.exit()
```
jakepetroules: This should exit with whatever error SwiftArgumentParser exits with when validate() fails

Sources/Commands/PackageCommands/BuildServer.swift
```diff
@@ -0,0 +1,77 @@
...
+        let server = try await SwiftPMBuildServer(packageRoot: packagePath, buildSystem: buildSystem, connectionToClient: clientConnection, exitHandler: {_ in Self.exit() })
+
+        clientConnection.start(
+            receiveHandler: server,
+            closeHandler: {
+                Self.exit()
+            }
+        )
+
+        // Park the main function by sleeping for 10 years.
```
jakepetroules: Cleaner solution?

```swift
await withCheckedContinuation { continuation in
    clientConnection.start(
       receiveHandler: server,
        closeHandler: {
            continuation.resume()
        }
    )
}
```

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
owenv: yeah, this is the approach I ended up taking

Sources/Commands/PackageCommands/BuildServer.swift
ahoppen: FWIW, we’re doing this in SourceKit-LSP and it hasn’t caused any issues so far.

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
```diff
@@ -0,0 +1,297 @@
...
+        switch request {
+        case let request as RequestAndReply<BuildShutdownRequest>:
+            await request.reply {
+                _ = try await connectionToUnderlyingBuildServer.send(request.params)
+                return await shutdown()
+            }
+        case let request as RequestAndReply<BuildTargetPrepareRequest>:
+            await request.reply {
+                for target in request.params.targets.filter({ $0.isSwiftPMBuildServerTargetID }) {
+                    // Error
```
ahoppen: This isn’t an error. SourceKit-LSP will eg. ask for the package manifest target to be prepared, it’s just that there’s nothing to do, so the BSP server can return immediately.

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
```diff
@@ -0,0 +1,297 @@
...
+            }
+        case let request as RequestAndReply<WorkspaceBuildTargetsRequest>:
+            await request.reply {
+                var targetsResponse = try await connectionToUnderlyingBuildServer.send(request.params)
+                targetsResponse.targets.append(await manifestTarget())
+                return targetsResponse
+            }
+        case let request as RequestAndReply<WorkspaceWaitForBuildSystemUpdatesRequest>:
+            await request.reply {
+                await waitForBuildSystemUpdates(request: request.params)
```
ahoppen, line 229: Do you also need to wait for updates of the underlying build server?

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
```diff
@@ -0,0 +1,297 @@
...
+    private func fileEventShouldTriggerPackageReload(event: FileEvent) -> Bool {
+        guard let fileURL = event.uri.fileURL else {
+            return false
+        }
+        switch event.type {
+        case .created, .deleted:
+            // TODO: this is overly conservative
+            return true
+        case .changed:
+            // TODO: check for changes to version specific manifests too
```
ahoppen: FWIW, I added support for version-specific package manifests in SourceKit-LSP’s built-in SwiftPM build server in https://github.com/swiftlang/sourcekit-lsp/pull/2259.

owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/Commands/PackageCommands/BuildServer.swift
jakepetroules: dup2's return value on Windows is always 0 on success rather than the fd as it is in POSIX. We're not using the return value here in a way that would hit that issue.

That said, strerror is also problematic due to thread safety and due to being deprecated on some environments... honestly, we should just use Swift System. It's cleaner and hides all of these platform nuances for you.

Sources/Commands/PackageCommands/BuildServer.swift
```diff
@@ -0,0 +1,78 @@
...
+        guard let packagePath = try swiftCommandState.getWorkspaceRoot().packages.first else {
+            throw ArgumentParser.ValidationError("unknown package")
+        }
+
+        let server = try await SwiftPMBuildServer(
+            packageRoot: packagePath,
+            buildSystem: buildSystem,
+            workspace: swiftCommandState.getActiveWorkspace(),
+            connectionToClient: clientConnection,
+            exitHandler: {_ in Self.exit() }
```
jakepetroules: I think Self.exit is from ArgumentParser and just calls exit(). We probably want to do something here to cause the closeHandler of the client connection to get called instead, so we shut down cleanly.

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
```diff
@@ -0,0 +1,331 @@
...
+        let underlyingInitializationResponse = try await connectionToUnderlyingBuildServer.send(request)
+        let underlyingSourceKitData = SourceKitInitializeBuildResponseData(fromLSPAny: underlyingInitializationResponse.data)
+        if underlyingSourceKitData?.watchers?.isEmpty == false {
+            logToClient(.warning, "Underlying build server reported unexpected file watchers")
+        }
+        state = .waitingForInitializedNotification
+        scheduleRegeneratingBuildDescription()
+        return InitializeBuildResponse(
+            displayName: "SwiftPM Build Server",
+            version: "",
```
jakepetroules: nit: could be the Swift version tag?

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

Tests/SwiftPMBuildServerTests/BuildServerTests.swift
```diff
@@ -0,0 +1,174 @@
...
+                    capabilities: .init(languageIds: [.swift, .c, .objective_c, .cpp, .objective_cpp])
+                )
+            )
+            connection.send(OnBuildInitializedNotification())
+            _ = try await connection.send(WorkspaceWaitForBuildSystemUpdatesRequest())
+            try await body(connection, notificationCollector, fixture)
+            _ = try await connection.send(BuildShutdownRequest())
+            connection.send(OnBuildExitNotification())
+            connection.close()
+            bspProcess.waitUntilExit()
```
jakepetroules: Please don't use waitUntilExit as it spins the run loop and is dangerous in Swift Concurrency. If you need to do work in between launch() and waitUntilExit() (which is why I assume you're not using AsyncProcess), then you can borrow this technique from Swift Build:

```swift
extension Process {
    /// Runs the process and returns a promise which is fulfilled when the process exits.
    ///
    /// - note: This method sets the process's termination handler, if one is set.
    /// - throws: Rethrows the error from ``Process/run`` if the task could not be launched.
    public func launch() throws -> Promise<Processes.ExitStatus, any Error> {
        let promise = Promise<Processes.ExitStatus, any Error>()

        terminationHandler = { process in
            do {
                promise.fulfill(with: try .init(process))
            } catch {
                promise.fail(throwing: error)
            }
        }

        do {
            try run()
        } catch {
            terminationHandler = nil

            // If `run` throws, the terminationHandler won't be called.
            throw error
        }

        return promise
    }
}
```

owenv: I updated this to use a regular continuation which I think should also do the job here

owenv:
@swift-ci test

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
```diff
@@ -0,0 +1,331 @@
...
+import SwiftBuild
+import SWBBuildService
+import Workspace
+import BuildServerProtocol
+import LanguageServerProtocol
+import LanguageServerProtocolTransport
+import ToolsProtocolsSwiftExtensions
+
+extension Connection {
+    public func send<Request: RequestType>(_ request: Request) async throws -> Request.Response {
```
ahoppen: This won’t cancel the request if the task is cancelled. I think we should just move https://github.com/swiftlang/sourcekit-lsp/blob/main/Sources/LanguageServerProtocolExtensions/Connection%2BSend.swift to swift-tools-protocols and use that.

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
```diff
@@ -0,0 +1,331 @@
...
+        case let notification as OnWatchedFilesDidChangeNotification:
+            // The underlying build server only receives updates via new PIF, so don't forward this notification.
+            for change in notification.changes {
+                if self.fileEventShouldTriggerPackageReload(event: change) {
+                    scheduleRegeneratingBuildDescription()
+                    return
+                }
+            }
+        case is OnBuildLogMessageNotification:
+            // If we receive a build log message notification, forward it on to the client
```
ahoppen, line 166: `OnBuildLogMessageNotification` is only sent from the server to the client according the the LSP spec, just bouncing it back to the client doesn’t really make sense. I’d just ignore it.

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
```diff
@@ -0,0 +1,331 @@
...
+            connectionToClient.send(notification)
+        default:
+            break
```
ahoppen: Do you want to log an error when receiving an unknown notification?

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
```diff
@@ -0,0 +1,331 @@
...
+        switch request {
+        case let request as RequestAndReply<BuildShutdownRequest>:
+            await request.reply {
+                _ = try await connectionToUnderlyingBuildServer.send(request.params)
+                return await shutdown()
+            }
+        case let request as RequestAndReply<BuildTargetPrepareRequest>:
+            await request.reply {
+                var underlyingRequest = request.params
+                underlyingRequest.targets.removeAll(where: { $0.isSwiftPMBuildServerTargetID })
```
ahoppen: Nit

```suggestion
                underlyingRequest.targets.removeAll(where: \.isSwiftPMBuildServerTargetID)
```

Same below

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
```diff
@@ -0,0 +1,331 @@
...
+                await waitForBuildSystemUpdates(request: request.params)
+            }
+        default:
+            await request.reply { throw ResponseError.methodNotFound(Request.method) }
+        }
+    }
+
+    private func initialize(request: InitializeBuildRequest) async throws -> InitializeBuildResponse {
+        if state != .waitingForInitializeRequest {
+            logToClient(.warning, "Received initialization request while the build server is \(state)")
```
ahoppen, line 238: Should these be log messages using `logger` so that they end up in the build server logs, eg. inside a sysdiagnose?

ahoppen (commented):
A few drive-by comments

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
owenv: This is here so that if we receive this notification from the Swift Build build server, we forward it to the SwiftPM build server client (LSP), so I think we still want it

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
owenv: For now I'm consistently sending everything to the client, but I'll probably do a pass over all this as a follow up and decide what should also/instead go to the system log

owenv:
@swift-ci test

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
owenv: yep, added

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
owenv: I copied the code for now, I'll open a PR to add it to tools-protocols later today and begin adopting that in the toolchain

Tests/SwiftPMBuildServerTests/BuildServerTests.swift
```diff
@@ -0,0 +1,178 @@
...
+                outFD: outPipe.fileHandleForWriting
+            )
+            let bspProcess = Process()
+            bspProcess.standardOutputPipe = inPipe
+            bspProcess.standardInput = outPipe
+            let execPath = SwiftPM.xctestBinaryPath(for: "swift-package").pathString
+            bspProcess.executableURL = URL(filePath: execPath)
+            bspProcess.arguments = ["experimental-build-server", "--build-system", "swiftbuild"]
+            bspProcess.currentDirectoryURL = URL(filePath: fixture.pathString)
+            bspProcess.launch() as Void
```
jakepetroules: launch() is effectively deprecated as it throws Objective-C exceptions instead of Swift errors; it's replaced by run.

```suggestion
            try bspProcess.run() as Void
```

Tests/SwiftPMBuildServerTests/BuildServerTests.swift
```diff
@@ -0,0 +1,178 @@
...
+                    rootUri: URI(URL(filePath: fixture.pathString)),
+                    capabilities: .init(languageIds: [.swift, .c, .objective_c, .cpp, .objective_cpp])
+                )
+            )
+            connection.send(OnBuildInitializedNotification())
+            _ = try await connection.send(WorkspaceWaitForBuildSystemUpdatesRequest())
+            try await body(connection, notificationCollector, fixture)
+            _ = try await connection.send(BuildShutdownRequest())
+            await withCheckedContinuation { continuation in
+                bspProcess.terminationHandler = { _ in
```
jakepetroules: This is too late, the terminationHandler needs to be installed before run/launch is called or you may miss the notification, leading to a hang.

You also need to make sure you uninstall the terminationHandler if run() throws.

I suggested my launch promise because it handles all of this for you and I expect we're going to want those abstractions in SwiftPM at some point or another anyways...

Tests/SwiftPMBuildServerTests/BuildServerTests.swift
jakepetroules: There are a few edge cases with that, commented inline.

Tests/SwiftPMBuildServerTests/BuildServerTests.swift
jakepetroules: Or, why not just use AsyncProcess? It looks like it has a launch() method and an async-compatible waitUntilExit() overload.

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
ahoppen: OK, I’m confused now. Isn’t this method handling notifications sent from the client (ie. SourceKit-LSP) to the server (ie. the SwiftPM BSP server)?

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
ahoppen: Great. Thank you. If you get a chance to also adopt it in SourceKit-LSP, that would be greatly appreciated.

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
owenv: It's handling both notifications from the underlying server to itself over one connection, and notifications from the client to itself on another, and they're both funneled to the same protocol requirement

Tests/SwiftPMBuildServerTests/BuildServerTests.swift
owenv: did this with run + async let instead. AsyncProcess makes the I/O kinda annoying

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

Sources/SwiftPMBuildServer/SwiftPMBuildServer.swift
ahoppen: Oh, interesting. I would encourage you to split these up (not in this PR, just in general). In SourceKit-LSP I invested a bit of time clearly documenting which methods receive data from the client and which ones send data to the client and it helped readability a lot IMO.

owenv:
@swift-ci test windows

Tests/SwiftPMBuildServerTests/BuildServerTests.swift
jakepetroules: Oh yeah, `async let`. That works, thanks!

PR #9130, opened 2025-09-11
Updates to get executables building and running on windows

daveinglis:
- the also changes the source modules on Windows to be a static lib since pre-linked .o is not supported.

depends on: https://github.com/swiftlang/swift-build/pull/795

### DIFF
--- a/Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
@@ -1000,18 +1000,16 @@ extension ProjectModel.BuildSettings {
     mutating func configureDynamicSettings(
         productName: String,
         targetName: String,
-        executableName: String,
         packageIdentity: PackageIdentity,
         packageName: String?,
         createDylibForDynamicProducts: Bool,
         installPath: String,
-        delegate: PackagePIFBuilder.BuildDelegate
+        delegate: PackagePIFBuilder.BuildDelegate,
     ) {
         self[.TARGET_NAME] = targetName
-        self[.PRODUCT_NAME] = createDylibForDynamicProducts ? productName : executableName
+        self[.PRODUCT_NAME] = productName
         self[.PRODUCT_MODULE_NAME] = productName
         self[.PRODUCT_BUNDLE_IDENTIFIER] = "\(packageIdentity).\(productName)".spm_mangledToBundleIdentifier()
-        self[.EXECUTABLE_NAME] = executableName
         self[.CLANG_ENABLE_MODULES] = "YES"
         self[.SWIFT_PACKAGE_NAME] = packageName ?? nil
 
--- a/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
@@ -237,31 +237,25 @@ extension PackagePIFProjectBuilder {
     ) throws -> (PackagePIFBuilder.ModuleOrProduct, resourceBundleName: String?) {
         precondition(sourceModule.isSourceModule)
 
-        let pifProductName: String
-        let executableName: String
         let productType: ProjectModel.Target.ProductType
 
         switch desiredModuleType {
         case .dynamicLibrary:
             // We are re-using this default for dynamic targets as well.
             if pifBuilder.createDylibForDynamicProducts {
-                pifProductName = "lib\(sourceModule.name).dylib"
-                executableName = pifProductName
                 productType = .dynamicLibrary
             } else {
-                pifProductName = sourceModule.name + ".framework"
-                executableName = sourceModule.name
                 productType = .framework
             }
 
         case .staticLibrary, .executable:
-            pifProductName = "\(sourceModule.name).o"
-            executableName = pifProductName
+            #if os(Windows) // Temporary until we get a new productType in swift-build
+            productType = .staticArchive
+            #else
             productType = .objectFile
+            #endif
 
         case .macro:
-            pifProductName = sourceModule.name
-            executableName = pifProductName
             productType = .hostBuildTool
         }
 
@@ -280,8 +274,8 @@ extension PackagePIFProjectBuilder {
             ProjectModel.Target(
                 id: sourceModule.pifTargetGUID(suffix: targetSuffix),
                 productType: productType,
-                name: "\(sourceModule.name)",
-                productName: pifProductName,
+                name: sourceModule.name,
+                productName: "$(EXECUTABLE_NAME)",
                 approvedByUser: approvedByUser
             )
         }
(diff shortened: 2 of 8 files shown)


Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
```diff
@@ -237,31 +237,25 @@ extension PackagePIFProjectBuilder {
...
             } else {
-                pifProductName = sourceModule.name + ".framework"
-                executableName = sourceModule.name
                 productType = .framework
             }
 
         case .staticLibrary, .executable:
-            pifProductName = "\(sourceModule.name).o"
-            executableName = pifProductName
+            #if os(Windows)
```
owenv: We'll need to vary this based on the platform being built for rather than the host platform so we don't break cross-compilation. There's not a great way to conditionalize productType based on platform in the PIF today, so this is where having a new "swiftPMTarget" product type to abstract away the platform specific differences might be useful

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
```diff
@@ -407,35 +401,22 @@ extension PackagePIFProjectBuilder {
...
             settings[.PRODUCT_NAME] = "$(TARGET_NAME)"
             settings[.PRODUCT_MODULE_NAME] = sourceModule.c99name
             settings[.PRODUCT_BUNDLE_IDENTIFIER] = "\(self.package.identity).\(sourceModule.name)"
                 .spm_mangledToBundleIdentifier()
-            settings[.EXECUTABLE_NAME] = executableName
             settings[.CLANG_ENABLE_MODULES] = "YES"
             settings[.GENERATE_PRELINK_OBJECT_FILE] = "NO"
             settings[.STRIP_INSTALLED_PRODUCT] = "NO"
 
-            // Macros build as executables, so they need slightly different
```
owenv, line 428: Why did this block need to be dropped?

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
daveinglis: as discussed offline, this is only temporary (I can add a comment here indicating such) but I think if we can get some basic windows functionally working  like building and running an executable which this change could get us there, with tests that now pass, we can then make the incremental changes like using a new product type and moving this kind of platform conditional into swift-build for that product type, and then have those tests ensure we don't regress with those change.

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
daveinglis: so correct me if I'm wrong but setting the MACH_O_TYPE should not be required to be set in the PIF as the product type definition in swift-build sets it for us, and with that said i feel CLANG_COVERAGE_MAPPING_LINKER_ARGS should be handled that same way

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
jakepetroules: How long is temporary - something on the order of a few weeks?

Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
```diff
@@ -1000,18 +1000,16 @@ extension ProjectModel.BuildSettings {
...
-        executableName: String,
         packageIdentity: PackageIdentity,
         packageName: String?,
         createDylibForDynamicProducts: Bool,
         installPath: String,
-        delegate: PackagePIFBuilder.BuildDelegate
+        delegate: PackagePIFBuilder.BuildDelegate,
     ) {
         self[.TARGET_NAME] = targetName
-        self[.PRODUCT_NAME] = createDylibForDynamicProducts ? productName : executableName
```
jakepetroules, line 1011: Will removing createDylibForDynamicProducts break the use case of xcodebuild building dylibs? We might be missing some tests there...

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -394,6 +394,9 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
                     do {
                         let workspaceInfo = try await session.workspaceInfo()
 
+                        #if !canImport(Darwin)
```
jakepetroules: Why non-Darwin only?

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
daveinglis: I hope so, will be working on the new product type next.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
daveinglis: I thought based on this https://github.com/swiftlang/swift-build/blob/978b1798c57a70b36b90ab07948fb8d5023b726a/SwiftBuild.docc/Development/build-debugging.md?plain=1#L5 that on Darwin these are control using user default settings, where this does not exist on non-darwin so I though I would just tie it to the verbosity flag of swift pm.

jakepetroules: UserDefaults exists on all platforms; we should not use the SwiftPM verbosity level for this.

daveinglis: I did not know that, how do you set these?

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
owenv: that's a good point, I wonder why this was put here originally...

and yeah, lets sink the CLANG_COVERAGE_MAPPING_LINKER_ARGS change into Swift Build

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: There isn't a convenience `defaults` command, but you can place a .plist file at a specific location and the settings will be picked up. That's the same way `defaults` works on macOS (or used to, in some cases it is a little more abstracted away from just a plist file on disk).

daveinglis:
@swift-ci please test

daveinglis:
@swift-ci please test

owenv (approved):
LGTM, once the CLANG_COVERAGE_MAPPING_LINKER_ARGS is applied at the Swift Build layer

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
daveinglis: removed

daveinglis:
@swift-ci please test

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
daveinglis: ya, I think @owenv was saying offline that there is some history here we just don't know about as to why this was like this.  I think we should try and move forward with this as it cleans things up and I imagine we should be able to address any fallout quickly if need be...

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

PR #9057, opened 2025-08-21
Fix pre-computation of traits when loading dependency manifests

bripeticca:
This is a tentative fix to address some issues that arose with trait-based dependency resolution on Linux and, found through development, a related issue found involving the `swift package show-dependencies` command.

Still under investigation at the moment, however this address the error that was thrown as a discrepancy in state when executing `swift package show-dependencies` if traits were guarding a dependency.

TODO:
* tests against show-dependencies with traits
* assert trait-based dependency resolution tests works on linux; create tests that reproduce previously encountered errors

PR #9136, opened 2025-09-15
[6.2] Cherry-pick: Fix pre-computation of traits when loading dependency manifests

bripeticca:
- **Explanation**:
Fixes some errors in trait computation that hadn't considered when default traits `["default"]` were being appended to the `EnabledTraitsMap`, which should only consider a list of flattened traits + *explicitly enabled traits* by a user/parent package. The inclusion of `default` in this dictionary was resulting in an inaccurate computation of which traits were enabled, since pre-computation serves to flatten the list of transitively enabled traits for future reference.
There were also cases where we were re-computing transitively enabled traits in areas where we had already filled out the `enabledTraitsMap`, so we now default to simply fetching the entry from the dictionary instead of computing the traits all over again.

- **Scope**:
Fixes trait-related computation in dependency resolution

- **Issues**:

- **Original PRs**:
#9057 
- **Risk**:
Low risk
- **Testing**:
Added new fixtures + unit tests to address the behaviour that was previously incorrect.
- **Reviewers**:
@dschaefer2

### DIFF
--- a/Fixtures/Traits/PackageConditionalDeps/Package.swift
+++ b/Fixtures/Traits/PackageConditionalDeps/Package.swift
@@ -1,0 +1,39 @@
+// swift-tools-version: 6.1
+
+import PackageDescription
+
+let package = Package(
+    name: "PackageConditionalDeps",
+    products: [
+        .library(
+            name: "PackageConditionalDeps",
+            targets: ["PackageConditionalDeps"]
+        ),
+    ],
+    traits: [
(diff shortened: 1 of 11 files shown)


bripeticca:
@swift-ci please test

rconnell9:
@swift-ci test windows

dschaefer2 (approved):

jakepetroules (approved):

PR #9139, opened 2025-09-15
[Swift Build] Default to the package's declared deployment target if none is explicitly specified

owenv:
Default to an unversioned triple if the user hasn't provided one so we don't raise the deployment target to a version the package may not support

closes https://github.com/swiftlang/swift-package-manager/issues/9095

### DIFF
--- a/Sources/SPMBuildCore/Plugins/DefaultPluginScriptRunner.swift
+++ b/Sources/SPMBuildCore/Plugins/DefaultPluginScriptRunner.swift
@@ -466,12 +466,19 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
 
         var env = Environment.current
 
+        // FIXME: This is largely a workaround for improper rpath setup on Linux. It should be
+        // removed once the Swift Build backend switches to use swiftc as the linker driver
+        // for targets with Swift sources. For now, limit the scope to non-macOS, so that
+        // plugins do not inadvertently use the toolchain stdlib instead of the OS stdlib
+        // when built with a Swift.org toolchain.
+        #if !os(macOS)
         // Update the environment for any runtime library paths that tools compiled
         // for the command plugin might require after they have been built.
         let runtimeLibPaths = self.toolchain.runtimeLibraryPaths
         for libPath in runtimeLibPaths {
             env.appendPath(key: .libraryPath, value: libPath.pathString)
         }
+        #endif
 
 #if os(Windows)
         let pluginLibraryPath = self.toolchain.swiftPMLibrariesLocation.pluginLibraryPath.pathString
--- a/Tests/CommandsTests/BuildCommandTests.swift
+++ b/Tests/CommandsTests/BuildCommandTests.swift
@@ -1459,8 +1459,11 @@ struct BuildCommandTestCases {
         let hostArch: String
         #if arch(arm64)
         hostArch = "arm64"
-        #else
+        #elseif arch(x86_64)
         hostArch = "x86_64"
+        #else
+        Issue.record("test is not supported on host arch")
+        return
         #endif
 
         // Unversioned triple - build should pass
@@ -1483,9 +1486,9 @@ struct BuildCommandTestCases {
                 )
         }
 
-        if buildSystem == .swiftbuild {
-            // Versioned triple with unsupported deployment target - build should fail
-            try await fixture(name: "Miscellaneous/RequiresOlderDeploymentTarget") { path in
+        // Versioned triple with unsupported deployment target - build should fail
+        try await withKnownIssue {
+            _ = try await fixture(name: "Miscellaneous/RequiresOlderDeploymentTarget") { path in
                 await #expect(throws: Error.self) {
                     try await executeSwiftBuild(
                         path,
@@ -1495,6 +1498,9 @@ struct BuildCommandTestCases {
                     )
                 }
             }
+        } when: {
+            // The native build system does not correctly pass the elevated deployment target
+            buildSystem != .swiftbuild
         }
     }
 }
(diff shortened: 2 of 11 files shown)


owenv:
@swift-ci test

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1385,6 +1385,63 @@ struct BuildCommandTestCases {
...
+                    path,
+                    buildSystem: buildSystem,
+                    throwIfCommandFails: true
+                )
+        }
+
+        let hostArch: String
+        #if arch(arm64)
+        hostArch = "arm64"
+        #else
```
jakepetroules: nit: else-if-x86_64 would be better

owenv:
@swift-ci test

owenv:
@swift-ci test

dschaefer2 (approved):

owenv:
Not sure how I broke the self-hosted plugin tests but those are reliably failing with this patch...

owenv:
@swift-ci test self-hosted

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test self hosted

GitHubismae (approved):

owenv:
@swift-ci test

owenv:
@swift-ci test macOS

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
Only remaining failure is a warning due to duplicate /usr/lib/swift rpaths, because both the driver and build system are adding one for the back deploy libs

owenv:
@swift-ci test

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test self hosted

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1396,6 +1396,66 @@ struct BuildCommandTestCases {
...
+        hostArch = "arm64"
+        #elseif arch(x86_64)
+        hostArch = "x86_64"
+        #else
+        Issue.record("test is not supported on host arch")
+        return
+        #endif
+
+        // Unversioned triple - build should pass
+        try await fixture(name: "Miscellaneous/RequiresOlderDeploymentTarget") { path in
```
bkhouri, line 1470: suggestion: can we make these as parameterized tests?  the test arguments can be the the `extraArgs`.  We can also add an ID to each argument to describe the test intent.

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1396,6 +1396,66 @@ struct BuildCommandTestCases {
...
+                try await executeSwiftBuild(
+                    path,
+                    extraArgs: ["--triple", "\(hostArch)-apple-macosx12.0"],
+                    buildSystem: buildSystem,
+                    throwIfCommandFails: true
+                )
+        }
+
+        if buildSystem == .swiftbuild {
```
bkhouri: question: why is this only applicable to SwiftBuild?  Should we verify and ensure the native build system also fails?

Sources/SPMBuildCore/Plugins/DefaultPluginScriptRunner.swift
```diff
@@ -466,12 +466,19 @@ public struct DefaultPluginScriptRunner: PluginScriptRunner, Cancellable {
 
         var env = Environment.current
 
+        // FIXME: This is largely a workaround for improper rpath setup on Linux. It should be
```
bkhouri, line 469: issue (possibly-blocking) the path setup on linux is fixed in #9218.  This PR depend on it so it provides the proper fix/solution.

Tests/CommandsTests/BuildCommandTests.swift
owenv: The old build system has incorrect behavior here, I replaced the check with a known issue

Tests/CommandsTests/BuildCommandTests.swift
owenv: IMO it's a bit easier to read written out like this since we're already parameterizing based on build system

Sources/SPMBuildCore/Plugins/DefaultPluginScriptRunner.swift
owenv: The stdlib rpaths are a different set from those added in https://github.com/swiftlang/swift-package-manager/pull/9218. We can likely drop these runtime paths entirely soon but I need to test that out in another patch first

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test self hosted

3405691582:
@owenv OpenBSD uses versioned triples for module imports, and so expects a versioned triple for `-target` arguments. This change causes some import errors since it supplies the unversioned triple to `-target` during the bootstrapping process.

jakepetroules:
@3405691582 this change is being partly reverted in #9271 for similar reasons on FreeBSD.

PR #9140, opened 2025-09-16
[Swift Build] Correct -parse-as-library behavior for executable targets

owenv:
Ensure we match the behavior of the native build system with various combination of targets with 1/2+ source files @main vs top level code, etc.

Depends on https://github.com/swiftlang/swift-build/pull/800
Closes https://github.com/swiftlang/swift-package-manager/issues/9098


owenv:
@swift-ci test

owenv:
@swift-ci please test

owenv:
@swift-ci test

PR #9141, opened 2025-09-16
[6.2.1] Cherry-pick: Fix traits .when(traits:) condition; pre-computation when loading dependency manifests

bripeticca:
- **Explanation**:
Addresses multiple traits issues:
     - Changes the check for trait-guarded target dependencies to determine whether the package dependency from which it derives is used in the parent package; prior to this change, the check was asserting that *all* traits in the `when` condition of a target dependency were enabled for the dependency to be considered unguarded, but now the check confirms that *at least one* trait in the condition is enabled.
    - Fixes some errors in trait computation that hadn't considered when default traits ["default"] were being appended to the EnabledTraitsMap, which should only consider a list of flattened traits + explicitly enabled traits by a user/parent package. The inclusion of default in this dictionary was resulting in an inaccurate computation of which traits were enabled, since pre-computation serves to flatten the list of transitively enabled traits for future reference.
There were also cases where we were re-computing transitively enabled traits in areas where we had already filled out the enabledTraitsMap, so we now default to simply fetching the entry from the dictionary instead of computing the traits all over again.
- **Scope**:
Package resolution stage; particularly affects pre-computation of transitively enabled traits when determining which target dependencies are/aren't guarded by traits. Also allows for multiple traits to be considered in a target dependency's `when` condition, any of which being enabled would then include said target dependency.
- **Issues**:
- **Original PRs**:
 #9015
 #9057 
- **Risk**:
Low risk; fixes trait-related issues. 
- **Testing**:
Fixtures added for end-to-end testing of the issue; regression tests added to multiple suites
- **Reviewers**:
@FranzBusch @jakepetroules @dschaefer2

### DIFF
--- a/Fixtures/Traits/Example/Package.swift
+++ b/Fixtures/Traits/Example/Package.swift
@@ -25,6 +25,7 @@ let package = Package(
         "BuildCondition1",
         "BuildCondition2",
         "BuildCondition3",
+        "ExtraTrait",
     ],
     dependencies: [
         .package(
@@ -101,6 +102,11 @@ let package = Package(
                     package: "Package10",
                     condition: .when(traits: ["Package10"])
                 ),
+                .product(
+                    name: "Package10Library2",
+                    package: "Package10",
+                    condition: .when(traits: ["Package10", "ExtraTrait"])
+                )
             ],
             swiftSettings: [
                 .define("DEFINE1", .when(traits: ["BuildCondition1"])),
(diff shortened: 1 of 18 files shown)


bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

PR #9143, opened 2025-09-17
Use SwiftBuild API to reliably compute built artifacts

owenv:
Depends on https://github.com/swiftlang/swift-build/pull/801

By querying the build system for the artifacts, we ensure we only report any we will actually produce in the current build. At the same time, pull in an alternative fix for https://github.com/swiftlang/swift-package-manager/pull/9121 which restores the original behavior of the native build system when considering products in dependency packages.

### DIFF
--- a/Sources/Build/BuildOperation.swift
+++ b/Sources/Build/BuildOperation.swift
@@ -461,11 +461,37 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
         let buildResultBuildPlan = buildOutputs.contains(.buildPlan) ? try buildPlan : nil
         let buildResultReplArgs = buildOutputs.contains(.replArguments) ? try buildPlan.createREPLArguments() : nil
 
+        let artifacts: [(String, PluginInvocationBuildResult.BuiltArtifact)]?
+        if buildOutputs.contains(.builtArtifacts) {
+            let builtProducts = try buildPlan.buildProducts
+            artifacts = try builtProducts.compactMap {
+                switch $0.product.type {
+                case .library(let kind):
+                    let artifactKind: PluginInvocationBuildResult.BuiltArtifact.Kind
+                    switch kind {
+                        case .dynamic: artifactKind = .dynamicLibrary
+                        case .static, .automatic: artifactKind = .staticLibrary
+                    }
+                    return try ($0.product.name, .init(
+                        path: $0.binaryPath.pathString,
+                        kind: artifactKind)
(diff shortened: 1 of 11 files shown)


Sources/Build/BuildOperation.swift
```diff
@@ -461,11 +461,32 @@ public final class BuildOperation: PackageStructureDelegate, SPMBuildCore.BuildS
...
 
+        let artifacts: [(String, PluginInvocationBuildResult.BuiltArtifact)]?
+        if buildOutputs.contains(.builtArtifacts) {
+            let builtProducts = try buildPlan.buildProducts
+            artifacts = try builtProducts.compactMap {
+                switch $0.product.type {
+                case .library(let kind):
+                    return try ($0.product.name, .init(
+                        path: $0.binaryPath.pathString,
+                        kind: (kind == .dynamic) ? .dynamicLibrary : .staticLibrary)
```
jakepetroules: Suggest using a switch instead of a ternary so we don't silently default new cases to staticLibrary in the future

owenv: fixed, I copied this around and didn't look it over carefully

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -585,7 +591,46 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
...
+                                    .staticLibrary
+                                case .dynamicLibrary:
+                                    .dynamicLibrary
+                                case .framework:
+                                    // We treat frameworks as dylibs here, but the plugin API should grow to accomodate more product types
+                                    .dynamicLibrary
+                                }
+                                var name = target.name
+                                // FIXME: We need a better way to map between SwiftPM target/product names and PIF target names
+                                if pifTargetName.hasSuffix("-product") {
```
jakepetroules, line 618: This seems like an accident waiting to happen :S

Should we have some more API surface here?

owenv: I think so, but I haven't decided what it should look like yet

cmcgee1024 (approved):

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test self hosted

Issue #9149, opened 2025-09-17
Tests in `swift package init --type executable --enable-swift-testing` have no test target


MaxDesiatov:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

Run `swift package init --type executable --enable-swift-testing`.

### Expected behavior

`Tests` directory that was created should have a corresponding test target in `Package.swift`

### Actual behavior

`Package.swift` has not corresponding test target.

### Swift Package Manager version/commit hash

`main`, also reproducible with `swift-DEVELOPMENT-SNAPSHOT-2025-09-14-a`

### Swift & OS version (output of `swift --version ; uname -a`)

```
> swift --version
Apple Swift version 6.3-dev (LLVM 73f0f4667f22bd1, Swift c987ea9c8e684c2)
Target: arm64-apple-macosx15.0

> swift build --version
Swift Package Manager - Swift 6.3.0-dev
```

hwii-florescent:
Hi! I'd like to work on this issue.

I'll start by reproducing the problem locally this week using `swift package init --type executable --enable-swift-testing`, then look into where the package init templates generate the `Package.swift` file. My current understanding is that the `Tests` directory is created, but the generated manifest is missing the matching test target.

Any pointers on where to start in the codebase would be appreciated!

plemarquand:
@hwii-florescent thanks for your interest in helping out! Unfortunately this one was fixed by https://github.com/swiftlang/swift-package-manager/pull/9152 but I forgot to close the issue out. If you encounter another issue you'd like to work on we'd be happy to point you in the right direction.

PR #9152, opened 2025-09-17
Add missing `testTarget` for package init swift-testing tests

plemarquand:
### Motivation:
When running `swift package init --type executable --enable-swift-testing` an example test is generated but the test target is not created in the manifest.

### Modifications:

Add the missing test target to the manifest when these types are initialized with swift-testing tests. This applies to the `tool` and `executable` types.

### Result:

Packages are initialized with a test target and running `swift test` runs the generated test successfully.

Issue: #9149

### DIFF
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -300,13 +300,36 @@ public final class InitPackage {
 
                 """
                 if packageType == .executable {
+                    let testTarget: String
+                    if !options.supportedTestingLibraries.isEmpty {
+                        testTarget = """
+                                .testTarget(
+                                    name: "\(pkgname)Tests",
+                                    dependencies: ["\(pkgname)"]
+                                ),
+                        """
+                    } else {
+                        testTarget = ""
+                    }
                     param += """
                             .executableTarget(
                                 name: "\(pkgname)"
                             ),
+                    \(testTarget)
                         ]
                     """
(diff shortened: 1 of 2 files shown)


plemarquand:
@swift-ci test

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

dschaefer2 (approved):

Sources/Workspace/InitPackage.swift
```diff
@@ -300,20 +300,44 @@ public final class InitPackage {
...
+                    if !options.supportedTestingLibraries.isEmpty {
+                        testTarget = """
+                                .testTarget(
+                                    name: "\(pkgname)Tests",
+                                    dependencies: ["\(pkgname)"]
+                                ),
+                        """
+                    } else {
+                        testTarget = ""
+                    }
```
dschaefer2, line 313: I think this is true now, but just checking that we support tests depending on  executables with all our supported platforms? I remember Windows being an issue but I think we fixed that.

plemarquand:
@swift-ci test windows

Sources/Workspace/InitPackage.swift
plemarquand: I believe this is true, yes

plemarquand:
@swift-ci test windows

PR #9173, opened 2025-09-23
Add support for traits in manifest code generation

omochi:
Add support for traits in manifest code generation

### Motivation:

Currently, manifest code generation does not support the trait feature. This results in incomplete or incorrect generated manifests whenever traits are involved. To ensure manifests accurately reflect the package definition, trait-related constructs need to be supported.

### Modifications:

This patch implements trait support in manifest code generation. Specifically, it adds support for three syntactic forms:

- Defining package traits
- Declaring trait conditions on package dependencies
- Declaring trait conditions on target dependencies

### Result:

Manifest code generation will correctly emit trait-related constructs. Packages that define traits, or that depend on other packages/targets conditionally based on traits, will now be represented accurately in the generated manifest.

### DIFF
--- a/Sources/PackageModel/Manifest/PackageDependencyDescription.swift
+++ b/Sources/PackageModel/Manifest/PackageDependencyDescription.swift
@@ -73,6 +73,11 @@ public enum PackageDependency: Equatable, Hashable, Sendable {
                 condition: condition
             )
         }
+
+        // represents `.defaults`
+        public var isDefaultsCase: Bool {
+            name == "default" && condition == nil
+        }
     }
 
     case fileSystem(FileSystem)
--- a/Sources/PackageModel/ManifestSourceGeneration.swift
+++ b/Sources/PackageModel/ManifestSourceGeneration.swift
@@ -199,27 +199,34 @@ fileprivate extension SourceCodeFragment {
         }
 
         if let traits = dependency.traits {
-            let traits = traits.sorted { a, b in
-                PackageDependency.Trait.precedes(a, b)
-            }
-            params.append(
-                SourceCodeFragment(
-                    key: "traits",
-                    subnodes: traits.map { SourceCodeFragment(from: $0) }
+            // If only `.defaults` is specified, do not output `traits:` .
+            // This is because `traits:` is not available in toolchains earlier than 6.1.
+            let isDefault = traits.count == 1 &&
+                traits.allSatisfy(\.isDefaultsCase)
+
+            if !isDefault {
+                let traits = traits.sorted { a, b in
+                    PackageDependency.Trait.precedes(a, b)
+                }
+                params.append(
+                    SourceCodeFragment(
+                        key: "traits",
+                        subnodes: traits.map { SourceCodeFragment(from: $0) }
+                    )
                 )
-            )
+            }
         }
 
         self.init(enum: "package", subnodes: params)
     }
 
     init(from trait: PackageDependency.Trait) {
-        guard let condition = trait.condition else {
-            if trait.name == "default" {
-                self.init(enum: "defaults")
-                return
-            }
+        if trait.isDefaultsCase {
+            self.init(enum: "defaults")
+            return
+        }
 
+        guard let condition = trait.condition else {
             self.init(string: trait.name)
             return
         }


omochi:
@swift-ci test

omochi:
@swift-ci test

Sources/PackageModel/ManifestSourceGeneration.swift
```diff
@@ -1023,6 +1129,47 @@ public struct SourceCodeFragment {
...
+        case (.some, .none): return false
+        case (.some(let a), .some(let b)):
+            return compareWrapped(a, b)
+        }
+    }
+}
+
+extension PackageDependency.Trait {
+    fileprivate static func precedes(_ a: PackageDependency.Trait, _ b: PackageDependency.Trait) -> Bool {
+        if a.name != b.name { return a.name < b.name }
```
bripeticca, line 1149: **suggestion**: I'm not sure we can ever have two traits of the same name in a single package (cc @FranzBusch is this a correct assumption?), so sorting traits that originate from the same package can probably just be done by using `traits.sorted(by: { $0.name < $1.name })` (this can similarly be done for the traits listed in the condition)

Sources/PackageModel/Manifest/PackageDependencyDescription.swift
```diff
@@ -73,6 +73,11 @@ public enum PackageDependency: Equatable, Hashable, Sendable {
                 condition: condition
             )
         }
+
+        // represents `.defaults`
+        public var isDefaultsCase: Bool {
```
FranzBusch, line 78: Does this need to be public?

Sources/PackageModel/ManifestSourceGeneration.swift
omochi: Indeed, if we unify the directives, sorting becomes simpler, which is nice, but...

At present, the following is valid:

```swift
dependencies: [
    .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "601.0.1", traits: [
        .trait(name: "Foo", condition: .when(traits: ["Blue"])),
        .trait(name: "Foo", condition: .when(traits: ["Green"])),
    ]),
],
```

It can be unified as follows:

```swift
dependencies: [
    .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "601.0.1", traits: [
        .trait(name: "Foo", condition: .when(traits: ["Blue", "Green"])),
    ]),
],
```

However, for example, if it is extended like this...

```swift
dependencies: [
    .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "601.0.1", traits: [
        .trait(name: "Foo", condition: .when(platforms: [.macOS], traits: ["Blue"])),
        .trait(name: "Foo", condition: .when(platforms: [.linux], traits: ["Green"])),
    ]),
],
```

This does not seem to be unifiable.

In the future, if it ever comes to this, we might overlook the necessary adjustments.
What should we do about that?

FranzBusch: There are two parts to traits.
1. The place where they are defined
2. The place where they are enabled for a given dependency

In the former, trait names can only be expressed once. In the latter, @omochi is right that it is totally valid to define an enabled trait multiple times with different conditions.

bripeticca: @omochi Thank you for the detailed examples! This helps a lot.

bripeticca:
@swift-ci please test windows

Sources/PackageModel/Manifest/PackageDependencyDescription.swift
omochi: Sorry for the lack of explanation.

The main purpose of this patch is not to make it `public`.

However, since `PackageModel.TraitDescription.isDefault` already exists and is `public`, I thought it would be natural to make `PackageModel.PackageDependency.Trait.isDefaultsCase` public as well.

In particular, for this part, the user-facing API uses `.defaults`, while the internal representation uses `"default"`, which do not match. This makes it easy to make mistakes during implementation, so I thought it would be helpful to provide an API.

omochi:
@swift-ci test windows

omochi:
@bripeticca @FranzBusch

The CI has passed. Could you please merge this?
Alternatively, since it hasn’t been approved yet, please review the code again if necessary.

bripeticca:
LGTM -- I'll let @FranzBusch have a look as well before merging.

omochi:
Thank you!

PR #9180, opened 2025-09-23
Remove animated progress bar from SwiftBuild output

bripeticca:
This change removes the progress bar from the output generated when we use the SwiftBuild build system in SwiftPM. The progress bar tears and affects the format of the emitted logs and diagnostics.


bripeticca:
@swift-ci please test

bripeticca:

bripeticca:

bripeticca:
@swift-ci please test macOS

bripeticca:

bripeticca:
@swift-ci please test macOS

bripeticca:
@swift-ci please test

bripeticca:
An update here: plucking out the progress reporting entirely was causing some tests to fail, since we check for particular strings in the output that we initially got from the animated progress bar. We should still report on the progress, so rather than using the progress bar we can default to the same tool that the native build system uses -- this grants us the ability to still emit the progress messages while not affecting the formatting in the same way as the progress bar.

bripeticca:
@swift-ci please test windows

PR #9188, opened 2025-09-25
Mirror what native build system does for enabling clang modules

daveinglis:
- the native builder only enables clang modules on Darwin, so we will need to do the same with swift-build.

### DIFF
--- a/Sources/SwiftBuildSupport/PIFBuilder.swift
+++ b/Sources/SwiftBuildSupport/PIFBuilder.swift
@@ -412,7 +412,8 @@ public final class PIFBuilder {
                 }
 
                 let packagePIFBuilderDelegate = PackagePIFBuilderDelegate(
-                    package: package
+                    package: package,
+                    buildParameters: buildParameters,
                 )
                 let packagePIFBuilder = PackagePIFBuilder(
                     modulesGraph: self.graph,
@@ -484,9 +485,11 @@ public final class PIFBuilder {
 
 fileprivate final class PackagePIFBuilderDelegate: PackagePIFBuilder.BuildDelegate {
     let package: ResolvedPackage
+    let buildParameters: BuildParameters
     
-    init(package: ResolvedPackage) {
+    init(package: ResolvedPackage, buildParameters: BuildParameters) {
         self.package = package
+        self.buildParameters = buildParameters
     }
     
     var isRootPackage: Bool {
@@ -522,7 +525,8 @@ fileprivate final class PackagePIFBuilderDelegate: PackagePIFBuilder.BuildDelega
     }
     
     func configureProjectBuildSettings(_ buildSettings: inout ProjectModel.BuildSettings) {
-        /* empty */
+        // This is parity to the native build, But we should investigate what it would take to get this working on platforms.
+        buildSettings[.CLANG_ENABLE_MODULES] = self.buildParameters.triple.isDarwin() ? "YES" : "NO"
     }
     
     func configureSourceModuleBuildSettings(sourceModule: ResolvedModule, settings: inout ProjectModel.BuildSettings) {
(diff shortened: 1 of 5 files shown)


daveinglis:
@swift-ci test

Sources/SwiftBuildSupport/PIFBuilder.swift
```diff
@@ -522,7 +525,8 @@ fileprivate final class PackagePIFBuilderDelegate: PackagePIFBuilder.BuildDelega
     }
     
     func configureProjectBuildSettings(_ buildSettings: inout ProjectModel.BuildSettings) {
-        /* empty */
+        // This is parity to the native build, But we should investigate what it would take to get this working on platforms.
```
owenv: Instead of checking the current platform in BuildParameters, we should use platform conditionals here so that the PIF remains platform-agnostic. The subscript on buildSettings has a `for` parameter you can use to add platform conditions

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
```diff
@@ -529,11 +529,6 @@ public final class PackagePIFBuilder {
         }
 
         settings[.USE_HEADERMAP] = "NO"
-        settings[.OTHER_SWIFT_FLAGS].lazilyInitializeAndMutate(initialValue: ["$(inherited)"]) { $0.append("-DXcode") }
```
owenv, line 532: I think we should address this separately and conditionalize it on a new field in the PIFBuilderParameters

Sources/SwiftBuildSupport/PIFBuilder.swift
jakepetroules: We should probably try at some point to avoid giving the PIF builder access to build parameters to help prevent this.

dschaefer2: Agreed. The package model will likely need enhancement so that we can properly express all the conditionals users will need, especially around platforms, but that's in the works.

pmattos: Agreed that using the *platform conditions* is okay for now:
```swift
buildSettings.platformSpecificSettings[.linux]![.FOO] = "bar"
```

pmattos: **question**: Why are we doing this in the PIF delegate implementation for SwiftPM? To be clear, we should reserve the delegate for SwiftPM *only* patches.

If this is a general patch (eg, Xcode, etc) we should hardcode it in `PackagePIFBuilder.addProjectBuildSettings` instead. For instance, Xcode has it's own independent `PackagePIFBuilder.BuildDelegate` implementation and, as such, won't inherit such configuration as is.

daveinglis: Unless I'm mistaken, it doesn't look like you can set single value settings using `buildSettings.platformSpecificSettings`  Looks like swif-build `ProjectModel.BuildSettings` would need to work to support this.

daveinglis:
blocked by https://github.com/swiftlang/swift-build/issues/827

daveinglis:
@swift-ci test

bkhouri:
@swift-ci test windows

daveinglis:
> This looks good to me, but because the new settings API handles $(inherited) for multivalued settings a little differently it may be worth comparing the PIF of some simple packages before and after to double-check there aren't any unexpected changes.

Ok, discovered the previous encoding filtered out multi-value settings with just [$(inherited)], which with this new change did cause a lot of these ending up in the PIF, so I update the helper to not initialize these and now the PIF look correct. (used swiftPM as the package to compare the PIF with before and after)

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

PR #9190, opened 2025-09-25
Compile all of the plugins at the start of a build with swiftbuild

### LINKED ISSUE #8977: PackageCommandSwiftBuildTest.testCommandPluginCompilationError does not raise a compilation error

### Description

The `PackageCommandSwiftBuildTest.testCommandPluginCompilationError` does as `swift build --build-system swiftbuild` against a package that has a Command Plugin, which has a compilation error.

However, the test does not throw an error when using the SwiftBuild backend, but it does when using the native build system.

### Expected behavior

Building the sample project using the SwiftBuild build system either fails in a similar manner as the Native build system.

### Actual behavior

No error is thrown.

### Steps to reproduce

1. Clone the repository
2. Run the following and expect a non-zero return code 
    ```
    swift run swift-build  --build-system swiftbuild --package-path Fixtures/Miscellaneous/Plugins/CommandPluginUnsuccessful                         
    ```

`5c57a39a3cecc865dfdecbbfae0f62cc13780e00`

Observed on MacOS and linux
```
❯ uname -a
Darwin BKs-Work-M1Pro-MacBook-Pro-3.local 25.0.0 Darwin Kernel Version 25.0.0: Sat Jul 12 00:30:16 PDT 2025; root:xnu_development-12377.0.154.0.2~25/DEVELOPMENT_ARM64_T6000 arm64
```
and
```
build-user@a73e2eb7c3de:/source$ uname -a
Linux a73e2eb7c3de 6.1.68 #1 SMP Fri Nov 22 18:05:17 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
```

cmcgee1024:
When building with the swiftbuild build system there can be a compilation problem with the
plugin.swift with one of the plugins. However, you won't discover that unless you run the
command plugin specifically. This is not the behaviour of the existing native build system
that does a plugin pass at the start of a build.

Implement a similar plugin pass with the swiftbuild system to catch problems with the plugin
code when building the packages.

Fixes #8977

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -31,6 +31,8 @@ import func TSCBasic.withTemporaryFile
 
 import enum TSCUtility.Diagnostics
 
+import var TSCBasic.stdoutStream
+
 import Foundation
 import SWBBuildService
 import SwiftBuild
@@ -342,11 +344,22 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
     }
 
     public func build(subset: BuildSubset, buildOutputs: [BuildOutput]) async throws -> BuildResult {
+        // If any plugins are part of the build set, compile them now to surface
+        // any errors up-front. Returns true if we should proceed with the build
+        // or false if not. It will already have thrown any appropriate error.
+        var result = BuildResult(
+            serializedDiagnosticPathsByTargetName: .failure(StringError("Building was skipped")),
+            replArguments: nil,
+        )
+
         guard !buildParameters.shouldSkipBuilding else {
-            return BuildResult(
-                serializedDiagnosticPathsByTargetName: .failure(StringError("Building was skipped")),
-                replArguments: nil,
-            )
+            result.serializedDiagnosticPathsByTargetName = .failure(StringError("Building was skipped"))
+            return result
+        }
+
+        guard try await self.compilePlugins(in: subset) else {
+            result.serializedDiagnosticPathsByTargetName = .failure(StringError("Plugin compilation failed"))
+            return result
         }
 
         try await writePIF(buildParameters: buildParameters)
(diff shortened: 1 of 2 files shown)


cmcgee1024:
@swift-ci please test

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -357,6 +370,145 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
...
+        // Figure out what, if any, plugin descriptions to compile, and whether
+        // to continue building after that based on the subset.
+        let graph = try await getPackageGraph()
+
+        /// Description for a plugin module. This is treated a bit differently from the
+        /// regular kinds of modules, and is not included in the LLBuild description.
+        /// But because the modules graph and build plan are not loaded for incremental
+        /// builds, this information is included in the BuildDescription, and the plugin
+        /// modules are compiled directly.
+        class PluginBuildDescription: Codable {
```
jakepetroules: nit: classes should be `final` by default if you're not using inheritance

cmcgee1024:
@swift-ci please test

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
cmcgee1024: This class is declared within the scope of this function, so I think that would limit the ability for anything external to extend it. Is it still worth declaring it final in this case?

cmcgee1024:
@swift-ci test Windows

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
dschaefer2: The next question then is, why is this a class and not a value type if the lifetime is so short.

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -6961,28 +6957,31 @@ struct PackageCommandTests {
...
-                        #expect(!stdout.contains("Building for \(data.config.buildFor)..."), "stderr: \(stderr)")
+                // Check that building just one of them just compiles that plugin and doesn't build anything else.
+                do {
+                    let (stdout, stderr) = try await executeSwiftBuild(
+                        packageDir,
+                        configuration: data.config,
+                        extraArgs: ["--target", "MyCommandPlugin"],
+                        buildSystem: data.buildSystem,
+                    )
+                    if data.buildSystem == .native {
```
bkhouri: suggestion: instead of an `if` block, can we use a `switch` block , and explicitly call out we don't expect any output from SwiftBuild?   e.g.:

```swift
switch data.buildSystem {
   case .native:
        #expect(!stdout.contains("Compiling plugin MyBuildToolPlugin"), "stderr: \(stderr)")
        #expect(stdout.contains("Compiling plugin MyCommandPlugin"), "stderr: \(stderr)")
    case .swiftbuild:
       // nothing specific
       break
     case .xcode:
         Issue.record("Test expected have not been considered")
}
```

cmcgee1024:
@swift-ci please test

cmcgee1024:
@swift-ci test Windows

cmcgee1024:
@swift-ci test Windows

bkhouri (approved):

PR #9191, opened 2025-09-26
Fix test output streaming on Windows

owenv:
The AsyncProcess implementation on Windows was effectively buffering all output until process exit. Vendor over some utilities from Swift Build to stream the output, and attempt to avoid the bug addressed in https://github.com/swiftlang/swift-package-manager/pull/8047#pullrequestreview-2373531833

### DIFF
--- a/Sources/Basics/Concurrency/AsyncProcess.swift
+++ b/Sources/Basics/Concurrency/AsyncProcess.swift
@@ -514,15 +514,18 @@ package final class AsyncProcess {
         if self.outputRedirection.redirectsOutput {
             let stdoutPipe = Pipe()
             let stderrPipe = Pipe()
+            let stdoutStream = DispatchFD(fileHandle: stdoutPipe.fileHandleForReading).dataStream()
+            let stderrStream = DispatchFD(fileHandle: stderrPipe.fileHandleForReading).dataStream()
 
             group.enter()
-            stdoutPipe.fileHandleForReading.readabilityHandler = { (fh: FileHandle) in
-                let data = (try? fh.read(upToCount: Int.max)) ?? Data()
-                if data.count == 0 {
-                    stdoutPipe.fileHandleForReading.readabilityHandler = nil
+            Task {
+                defer {
+                    print("--- finished consuming stdout ---")
                     group.leave()
-                } else {
-                    let contents = data.withUnsafeBytes { [UInt8]($0) }
+                }
+                print("--- started consuming stdout ---")
+                for try await data in stdoutStream {
+                    let contents = [UInt8](data)
                     self.outputRedirection.outputClosures?.stdoutClosure(contents)
                     stdoutLock.withLock {
                         stdout += contents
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

cmcgee1024 (approved):

dschaefer2 (commented):
Yay! Yes, the only proper way to read stdout/stderr on Windows is multi-threaded. I do wonder if creating dedicated threads for this instead of using the Task thread pool would be safer.

jakepetroules:
This is fine, it's all non-blocking through Dispatch.

owenv:
@swift-ci test

owenv:
@swift-ci test

bripeticca:
@swift-ci please test windows

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

Sources/Basics/Concurrency/AsyncProcess.swift
```diff
@@ -1354,3 +1362,51 @@ extension FileHandle: WritableByteStream {
...
+        }
+
+    /// Returns an async stream which reads bytes from the specified file descriptor. Unlike `FileHandle.bytes`, it does not block the caller.
+    @available(macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
+    public func dataStream() -> some AsyncSequence<DispatchData, any Error> {
+        AsyncThrowingStream<DispatchData, any Error> {
+            while !Task.isCancelled {
+                let chunk = try await readChunk(upToLength: 4096)
+                if chunk.isEmpty {
+                    return nil
```
rconnell9, line 1390: Should this be a continuation (e.g. cont.finish() here)?  I'm not sure what the implications are of returning nil are for clients of this stream.

owenv: returning nil here is intended to terminate the stream, this initializer doesn't use a continuation

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

ahoppen:
Just so you are aware, because we are using  `DispatchIO.read(offset:length:queue:)`, we are seeing one thread spinning on Windows in SourceKit-LSP (lengthy discussion [here](https://github.com/swiftlang/swift-corelibs-libdispatch/issues/820) and [here](https://forums.swift.org/t/sourcekit-lsp-cpu-spinning-and-libdispatch-complications/82472) and [here](https://github.com/swiftlang/sourcekit-lsp/pull/2315)).

You are using `read(fromFileDescriptor:maxLength:runningHandlerOn:handler:)`, which seems to take a different code path at first but I wouldn’t be surprised if it ends up with a very similar issue in the end. If you don’t, all the better.

PR #9194, opened 2025-09-26
fix: remove preceding `-Xcc` instead of the argument following `-Werror` or `-Wwarning`

loonatick-src:
When suppressing warnings in remote packages, `SwiftModuleBuildDescription.compileArguments` removes the flag following `-Werror` / `-Wwarning` instead of the preceding `-Xcc` (issue: https://github.com/swiftlang/swift-package-manager/issues/9192). This is a proposed fix to that issue by reversing the arguments before filtering.

### Motivation:
The problem is described in this issue: https://github.com/swiftlang/swift-package-manager/issues/9192 .
This was most likely introduced in this PR https://github.com/swiftlang/swift-package-manager/pull/8315 .
It looks like the originally submitted code was filtering on the reversed args, in which case the preceding `-Xcc` would have been removed. But the final iteration does not reverse the args.

### Modifications:
Instead of filtering with a closure with side-effects, initialize an empty array and while iterating over `args` append to new array/skip/skip+remove last as needed.
#### Stale
Reverse `args` before filtering, reverse again after filtering to recover original order. Use `Array.lazy` to avoid materializing intermediate arrays.
Edit: I'm not completely sure if lazy will play well with the side-effect of mutating the captured `removeNextArg`.

### Result:
Given `swift build -Xcc -Werror <other flags>`, remote packages should now build equivalently as `swift build -suppress-warnings <other flags>` and not throw the internal error.

### DIFF
--- a/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
+++ b/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
@@ -611,26 +611,21 @@ public final class SwiftModuleBuildDescription {
         // suppress warnings if the package is remote
         if self.package.isRemote {
             // suppress-warnings and the other warning control flags are mutually exclusive
-            var removeNextArg = false
-            args = Array(args.lazy.reversed().filter { arg in
-                if removeNextArg {
-                    removeNextArg = false
-                    return false
-                }
+            var filteredArgs = [String]()
+            for arg in args {
                 switch arg {
                 case "-warnings-as-errors", "-no-warnings-as-errors":
-                    return false
+                    // Skip this flag
+                    continue
                 case "-Wwarning", "-Werror":
-                    removeNextArg = true
-                    return false
+                    // Remove preceding -Xcc and skip this flag
+                    filteredArgs.removeLast()
                 default:
-                    return true
+                    filteredArgs.append(arg)
                 }
-            }.reversed())
-            guard !removeNextArg else {
-                throw InternalError("Unexpected '-Wwarning' or '-Werror' at the end of args")
             }
-            args += ["-suppress-warnings"]
+            args = filteredArgs
+            args.append("-suppress-warnings")
         }
 
         // Pass `-user-module-version` for versioned packages that aren't pre-releases.


Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
```diff
@@ -612,7 +612,7 @@ public final class SwiftModuleBuildDescription {
         if self.package.isRemote {
             // suppress-warnings and the other warning control flags are mutually exclusive
             var removeNextArg = false
-            args = args.filter { arg in
+            args = Array(args.lazy.reversed().filter { arg in
```
owenv: Instead of relying on side effects in a filter closure, do you mind rewriting this to just use a loop so it's easier to follow what it's doing?

loonatick-src: Done. Looks a lot easier to follow to me as well.

Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift
```diff
@@ -611,26 +611,21 @@ public final class SwiftModuleBuildDescription {
...
+            for arg in args {
                 switch arg {
                 case "-warnings-as-errors", "-no-warnings-as-errors":
-                    return false
+                    // Skip this flag
+                    continue
                 case "-Wwarning", "-Werror":
-                    removeNextArg = true
-                    return false
+                    // Remove preceding -Xcc and skip this flag
```
owenv, line 621: Rereading this code again, I think this was actually intended to filter out `-Werror SwiftWarningGroup` (which is also valid) rather than `-Xcc -Werror`. I think you'll need to examine the preceding arg to check if it's actually -Xcc and therefore safe to remove. If the arg is passed to swiftc directly, the following arg should probably continue to be removed. It would be nice to have some tests for this as well.

Stepping back, this makes me think the compiler should stop diagnosing conflicting flags here and just allow one to win

loonatick-src: > I think this was actually intended to filter out `-Werror SwiftWarningGroup`

Ah, I did not know that. Just found `-Werror` in `swiftc --help-hidden`.

So, if I understand correctly, the following cases can happen.
- `-Xcc -Werror`
- `-Xcc -Werror=<CppWarningGroup>`
- `-Werror <SwiftWarningGroup>`

and the corresponding `Wwarning` variants. Anything missing?

Edit: I suppose the complete list is in [this proposal](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0480-swiftpm-warning-control.md#remote-targets-behavior).

> For C/C++:
> - `-Wxxxx`
> - `-Wno-xxxx`
> - `-Werror`
> - `-Werror=xxxx`
> - `-Wno-error`
> -    `-Wno-error=xxxx`
> 
> For Swift:
> -    `-warnings-as-errors`
> -   `-no-warnings-as-errors`
> -   `-Wwarning xxxx`
> -    `-Werror xxxx`

loonatick-src: > the compiler should stop diagnosing conflicting flags here and just allow one to win

@owenv  If that would be the case, then `-suppress-warnings` would have to win out over the others for remote packages. But that might not be a good default for non-remote packages.

PR #8970, opened 2025-07-25
Use only enabled traits when loading package

plemarquand:
Both `loadPackage` and `loadRootPackage` specified all traits as enabled when loading a package. Narrow this to only the enabled traits so that the returned `Package` accurately reflects the enabled traits.

PR #9195, opened 2025-09-26
[6.2] Cherry Pick - Use only enabled traits when loading package (#8970)

plemarquand:
Both `loadPackage` and `loadRootPackage` specified all traits as enabled when loading a package. Narrow this to only the enabled traits so that the returned `Package` accurately reflects the enabled traits.


plemarquand:
@swift-ci test

rconnell9 (approved):

PR #9206, opened 2025-09-30
Fix macros that depend on libraries that also depend on swift-syntax

dschaefer2:
Previously the library would be built from source and then link together with the prebuilt to create the macro executable. That causes a crash in the macro.

This change extends what we build using the prebuilt to all targets when built for host. It will continue to build from source when targets are built for destination. This will include plugin tools as well. The separation of build output between host and destination (which is potentially a different platform) makes this work.

This is accomplished by recording in the BuildEnvironment whether the destination was host and the adding in an internal HostCondition that checks it in it's satisfy method.


dschaefer2:
@swift-ci please test

dschaefer2:
A better fix is presented here: https://github.com/swiftlang/swift-package-manager/pull/9602

PR #9207, opened 2025-09-30
[6.2] Cherry-pick: Remove animated progress bar from SwiftBuild output

bripeticca:
This change removes the progress bar from the output generated when we use the SwiftBuild build system in SwiftPM. The progress bar tears and affects the format of the emitted logs and diagnostics.


bripeticca:
@swift-ci please test

DougGregor:
Is this a permanent change or is it a temporary thing for 6.2? (I like the progress bar...)

bripeticca:
@DougGregor This will be temporary until we can find a better way to emit a progress bar/equivalent without it tearing 😄 There is currently an effort to get the emitted output from SwiftBuild to be more informative/complete. Some formatting and rendering improvements are also under consideration as well, and I've bucketed progress bar as a possible task for this!

owenv (approved):

PR #9162, opened 2025-09-20
Add SwiftBuild coverage support

bkhouri:
Ensure the SwiftBuild build system has feature parity with the Native build system as it relates to coverage.

Fixes: #9077
Fixes: #9197
Depends on: https://github.com/swiftlang/swift-build/pull/844
Issue: rdar://159461439

PR #9212, opened 2025-10-02
PIF: Ensure targets are created

### LINKED ISSUE #9138: Parity: `Could not find target named 'exec2'` failure when building sample package (SwiftBuild)
Running the `swift run swift-build --package-path Fixtures/Miscellaneous/MultipleExecutables --target exec2` fails with an unexpected error message.

```
❯ swift run swift-build --package-path Fixtures/Miscellaneous/MultipleExecutables --target exec2 --build-system swiftbuild
Building for debugging...
[1/1] Write swift-version--68B999B8113CC7E.txt
Build of product 'swift-build' complete! (0.35s)
error: Could not find target named 'exec2'
error: The operation couldn’t be completed. (TSCUtility.Diagnostics error 0.)
```

Observed on branch `main` (commit ID `ec2a626d905c74d71745912ac83ae55ddb179844`).

bkhouri:
Some PIF target creation were guarded behind a tool version.  Remove this guard to ensure modules can be build using `--target` with the Swift Build build system.

Fixes: #9138
Issue: rdar://160638539
Depends on: #9162

### DIFF
--- a/Sources/Basics/FileSystem/AbsolutePath.swift
+++ b/Sources/Basics/FileSystem/AbsolutePath.swift
@@ -53,14 +53,19 @@ public struct AbsolutePath: Hashable, Sendable {
     /// The input string will be normalized if needed, as described in the
     /// documentation for AbsolutePath.
     public init(validating pathString: String) throws {
-        self.underlying = try .init(validating: pathString)
+        self.underlying = try .init(
+            validating: pathString.trimmingCharacters(in: .whitespacesAndNewlines),
+        )
     }
 
     /// Initializes an AbsolutePath from a string that may be either absolute
     /// or relative; if relative, `basePath` is used as the anchor; if absolute,
     /// it is used as is, and in this case `basePath` is ignored.
     public init(validating pathString: String, relativeTo basePath: AbsolutePath) throws {
-        self.underlying = try .init(validating: pathString, relativeTo: basePath.underlying)
+        self.underlying = try .init(
+            validating: pathString.trimmingCharacters(in: .whitespacesAndNewlines),
+            relativeTo: basePath.underlying,
+        )
     }
 
     /// Initializes the AbsolutePath by concatenating a relative path to an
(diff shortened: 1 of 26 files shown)


bkhouri:
@swift-ci test

Sources/Basics/FileSystem/AbsolutePath.swift
```diff
@@ -53,14 +53,19 @@ public struct AbsolutePath: Hashable, Sendable {
     /// The input string will be normalized if needed, as described in the
     /// documentation for AbsolutePath.
     public init(validating pathString: String) throws {
-        self.underlying = try .init(validating: pathString)
+        self.underlying = try .init(
```
pmattos, line 56: **suggestion:** I would favor `TSCAbsolutePath(...)` instead of the harder to read `.init(...)`.

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
```diff
@@ -199,7 +199,6 @@ extension PackagePIFProjectBuilder {
     /// we also construct a testable version of said executable.
     mutating func makeTestableExecutableSourceModule(_ executableModule: PackageGraph.ResolvedModule) throws {
         precondition(executableModule.type == .executable)
-        guard self.package.manifest.toolsVersion >= .v5_5 else { return }
```
pmattos, line 202: **issue:** For reference, this is the comment on the original PIF builder (which I preserved in the `func` doc comment above):
```swift
// If we're building an executable and the tools version is new enough, 
// we also construct a testable version of the executable.
if manifest.toolsVersion >= .v5_5 { ... }
```
So maybe Swift 5.5 somehow made this possible? Have you tested this in toolchains older than 5.5?

bkhouri: The SwiftBuild integration in SwiftPM is occurring on branch `main`, and I don't expect we will support swift 5.x

owenv: Packages below tools version 5.5 shouldn't generate this target representing the testable version of an executable since that feature isn't available. The tools version of a package is independent of the version of the tools themselves. I think the underlying issue here is likely that `--target` is not discovering the regular variant of the target, which is created in all tools versions. My best guess is that maybe that target got subsumed by a product target at the PIF layer, and the lookup code doesn't account for that.

Issue #9067, opened 2025-08-27
[SPI https://github.com/123flo321/pogoprotos-swift ]: swift build crashes when building package with 'swiftbuild' backend


kcieplak:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

swift build `--buildsystem=swiftbuild` crashes when building the https://github.com/123flo321/pogoprotos-swift package.

```
debug: PIF:     Added resource file '/workspace/spi-workdir/6D3BABBC-AC0B-4EE6-811F-B45117E08D65/pogoprotos-swift/.build-swiftbuild/checkouts/swift-protobuf/Sources/SwiftProtobufPluginLibrary/PrivacyInfo.xcprivacy'
debug: PIF:   Added linked dependency on target 'PACKAGE-TARGET:SwiftProtobuf'
debug: PIF: Created target 'PACKAGE-TARGET:SwiftProtobufPlugin' of type hostBuildTool and name 'SwiftProtobufPlugin'
debug: PIF:   Added dependency on target 'PACKAGE-TARGET:SwiftProtobuf'
debug: PIF:   Added dependency on target 'PACKAGE-TARGET:SwiftProtobufPluginLibrary'
SwiftBuildSupport/PackagePIFBuilder+Helpers.swift:518: Fatal error: Unexpectedly found nil while unwrapping an Optional value

*** Signal 4: Backtracing from 0x782dfb72dfbf... done ***

*** Program crashed: Illegal instruction at 0x0000782dfb72dfbf ***

Platform: x86_64 Linux (Debian GNU/Linux 12 (bookworm))

Thread 0 "swift-build":

  0  0x0000782df9d4f301 <unknown> in libc.so.6
...

Thread 1 "DispatchWorker" crashed:

  0                  0x0000782dfb72dfbf closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) + 623 in libswiftCore.so
  1 [ra]             0x0000782dfb8a2758 _assertionFailure(_:_:file:line:flags:) + 327 in libswiftCore.so
[...]

PR #9217, opened 2025-10-03
Fix crash in PIF generation

kcieplak:
Fix a crash when unwrapping an optional 'mainModule' of a module. In the case where a plugin has a dependency on a 'binaryTarget' there is no main module.

This only presents itself when you have defined the binaryTarget as both a product and target.

* Update method that determines plugin dependencies to handle binaryTarget types explicitly.
* Add a PIF generation test that captures this behaviour.

### DIFF
--- a/Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
@@ -523,8 +523,8 @@ extension PackageGraph.ResolvedModule {
             // triple will be '.tools' while the target we want to depend on will have a build triple of '.destination'.
             // See for more details:
             // https://github.com/swiftlang/swift-package-manager/commit/b22168ec41061ddfa3438f314a08ac7a776bef7a.
-            return mainModuleProduct.mainModule!.packageIdentity == self.packageIdentity &&
-                mainModuleProduct.mainModule!.name == self.name
+            return mainModule.packageIdentity == self.packageIdentity &&
+                mainModule.name == self.name
             // Intentionally ignore the build triple!
         }
     }
--- a/Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
@@ -513,6 +513,12 @@ extension PackageGraph.ResolvedModule {
 
     func productRepresentingDependencyOfBuildPlugin(in mainModuleProducts: [ResolvedProduct]) -> ResolvedProduct? {
         mainModuleProducts.only { (mainModuleProduct: ResolvedProduct) -> Bool in
+            // Handle binary-only executable products that don't have a main module, i.e. binaryTarget
+            guard let mainModule = mainModuleProduct.mainModule else {
+                return mainModuleProduct.type == .executable &&
+                    mainModuleProduct.modules.only?.type == .binary &&
+                    mainModuleProduct.modules.only?.name == self.name
+            }
             // NOTE: We can't use the 'id' here as we need to explicitly ignore the build triple because our build
             // triple will be '.tools' while the target we want to depend on will have a build triple of '.destination'.
             // See for more details:
--- a/Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
+++ b/Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
@@ -135,4 +135,32 @@ struct PIFBuilderTests {
             #expect(releaseConfig.settings.platformSpecificSettings[.windows]?[.SWIFT_ACTIVE_COMPILATION_CONDITIONS] == ["$(inherited)"])
         }
     }
+
+    @Test func pluginWithBinaryTargetDependency() async throws {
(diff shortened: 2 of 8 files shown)


kcieplak:
@swift-ci test

Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
```diff
@@ -513,6 +513,11 @@ extension PackageGraph.ResolvedModule {
 
     func productRepresentingDependencyOfBuildPlugin(in mainModuleProducts: [ResolvedProduct]) -> ResolvedProduct? {
         mainModuleProducts.only { (mainModuleProduct: ResolvedProduct) -> Bool in
+            // Handle binary-only executable products that don't have a main module, i.e. binaryTarget
+            guard let mainModule = mainModuleProduct.mainModule else {
+                return mainModuleProduct.type == .executable &&
+                    mainModuleProduct.modules.only?.type == .binary
```
pmattos: **issue:** I think we need to add this 3rd check to be safe: 
```swift
mainModuleProduct.modules.only?.name == self.name
```

Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
```diff
@@ -513,6 +513,11 @@ extension PackageGraph.ResolvedModule {
...
+            }
             // NOTE: We can't use the 'id' here as we need to explicitly ignore the build triple because our build
             // triple will be '.tools' while the target we want to depend on will have a build triple of '.destination'.
             // See for more details:
```
pmattos, line 524: **todo:** Below we can replace `mainModuleProduct.mainModule!` with just `mainModule`.

Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
```diff
@@ -135,4 +135,32 @@ struct PIFBuilderTests {
             #expect(releaseConfig.settings.platformSpecificSettings[.windows]?[.SWIFT_ACTIVE_COMPILATION_CONDITIONS] == ["$(inherited)"])
         }
     }
+
+    @Test func pluginWithBinaryTargetDependency() async throws {
```
pmattos, line 139: **praise:** Great to see more PIF tests in OSS SPM ☺️

kcieplak:
Fixes #9067

kcieplak:
@swift-ci test

Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
kcieplak: Done

pmattos:
**praise:** Matthew just confirmed this fixes **rdar://159087121** in Xcode too :)

pmattos:
I've a PR up to validate this patch within Xcode... will report back the results :)

kcieplak:
@swift-ci test

ci.swift.org was 504 for a while

kcieplak:
@swift-ci test

kcieplak:
@swift-ci test

kcieplak:
@swift-ci test

kcieplak:
@swift-ci test windows

kcieplak:

Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
bkhouri: I agree with @pmattos.  We should expand the PIF test coverage. 


Although this is a great test, it's a `large` test (ie: end-to-end). Can we investigate writing tests lower in the test pyramid, possibly testing `productRepresentingDependencyOfBuildPlugin` in isolation?

Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
```diff
@@ -135,4 +135,32 @@ struct PIFBuilderTests {
...
+            // Verify the plugin target exists
+            let pluginTarget = try project.target(named: "MyPlugin")
+            #expect(pluginTarget.common.name == "MyPlugin")
+
+            // Verify the executable target that uses the plugin exists
+            let executableTarget = try project.target(named: "MyPluginExe")
+            #expect(executableTarget.common.name == "MyPluginExe")
+
+            // Verify no errors were emitted during PIF generation
+            let errors = observabilitySystem.diagnostics.filter { $0.severity == .error }
```
bkhouri, line 155: question: do we expect warnings to occur?  If no, should we also ensure no warnings exist?  if the answer is "we don't care about warning", is there benefits adding a comment indicating so?

bkhouri:
Looks a Swift Testing test was not terminated correctly 

```
08:37:52  [1464/1464] Testing SourceControlTests.RepositoryManagerTests/testCancel
14:46:24  Cancelling nested steps due to timeout
14:46:24  Sending interrupt signal to process
14:46:44  After 20s process did not stop
```

Re-triggering the build

pmattos:
> I've a PR up to validate this patch within Xcode... will report back the results :)

All good in Xcode too 🚀

kcieplak:
@swift-ci test

kcieplak:
@swift-ci test windows

kcieplak:
@swift-ci test macOS

PR #9223, opened 2025-10-06
Rename runCommandPlugins methods to a more accurate name

cmcgee1024:

### Motivation:

This method is very confusing because it suggests that command plugins will be run during build, which is not the case. Command plugins always run at the user's request using `swift package plugin <name>` outside of a build.

Instead, this method should be renamed to reflect that it is running commands that the plugin has generated when it was invoked from a collection of results.

Rename the methods in both the SwiftBuildSupport and Build components and update the documentation.


cmcgee1024:
@swift-ci please test

owenv (approved):

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test Windows

PR #9178, opened 2025-09-23
Disable optimization to work around a compiler bug

Steelskin:
Temporarily disable optimizations for some functions.

### Motivation:

There is a bug in the Swift rebranch that prevents some functions from being optimized in a way that the assembler can handle. This temporarily disables optimizations for these functions until a permanent solution is implemented.

Bug: swiftlang/llvm-project#11377

### Modifications:

Disable optimizations on Windows for problematic functions.

### Result:

The project builds for the Swift rebranch.

PR #9225, opened 2025-10-07
Revert "Disable some optimizations on Windows"

Steelskin:
The issue has been fixed in the compiler so the workaround is no longer necessary.

Bug: swiftlang/llvm-project#11377

### Motivation:

The compiler bug has been fixed.

### Modifications:

Reverts #9178.

### Result:

The Windows arm64 build still works.

### DIFF
--- a/Sources/PackageLoading/ManifestLoader.swift
+++ b/Sources/PackageLoading/ManifestLoader.swift
@@ -202,15 +202,15 @@ extension ManifestLoaderProtocol {
     ) async throws -> Manifest {
         // find the manifest path and parse it's tools-version
         let manifestPath = try ManifestLoader.findManifest(
-            packagePath: packagePath, 
-            fileSystem: fileSystem, 
+            packagePath: packagePath,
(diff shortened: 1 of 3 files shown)


Steelskin:
@swift-ci please test

Sources/PackageLoading/ManifestLoader.swift
```diff
@@ -676,14 +677,10 @@ public final class ManifestLoader: ManifestLoaderProtocol {
 
             return result
         }
+
```
owenv: Looks like there are a handful of accidental white space diffs which would be good to clean up

owenv (approved):

Sources/PackageLoading/ManifestLoader.swift
Steelskin: I reverted the reverted whitespace changes.

Steelskin:
@swift-ci please test

Steelskin:

PR #9228, opened 2025-10-08
Cleanup some unnecessary build setting

daveinglis:
- fix adding -lc++ to OTHER_LDFLAGS to match native build system.
- remove OTHER_LDRFLAGS setting


daveinglis:
@swift-ci test

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
```diff
@@ -488,14 +488,16 @@ extension PackagePIFProjectBuilder {
...
-            indent: 1,
-            "Added '\(impartedSettings[.OTHER_LDFLAGS]!)' to imparted OTHER_LDFLAGS"
-        )
-
+        if sourceModule.isCxx {
+            for platform in ProjectModel.BuildSettings.Platform.allCases {
+                // darwin & freebsd
+                if [.macOS, .macCatalyst, .iOS, .watchOS, .tvOS, .xrOS, .driverKit, .freebsd].contains(platform) {
+                    settings[.OTHER_LDFLAGS, platform] = ["-lc++", "$(inherited)"]
+                } else if platform != .windows { // all others except windows
```
owenv: I wonder if we should warn or emit a debug log if we encounter a totally unknown platform here so we know we need to go fix it up

owenv:
It might be nice to see if we can put in a test in SwiftBuildSupportTests/PIFBuilderTests for this

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
daveinglis: I made the new test fail of it finds a unknown platform.

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

daveinglis:
@swift-ci test linux

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
```diff
@@ -552,7 +552,6 @@ public final class PackagePIFBuilder {
         // We currently deliberately do not support Swift ObjC interface headers.
         settings[.SWIFT_INSTALL_OBJC_HEADER] = "NO"
         settings[.SWIFT_OBJC_INTERFACE_HEADER_NAME] = ""
-        settings[.OTHER_LDRFLAGS] = []
```
jakepetroules, line 555: Do we know why this was set? Could be a default value that this is explicitly wanting to clear out.

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
```diff
@@ -484,14 +488,16 @@ extension PackagePIFProjectBuilder {
...
-        impartedSettings[.OTHER_LDFLAGS] = (sourceModule.isCxx ? ["-lc++"] : []) + ["$(inherited)"]
-        impartedSettings[.OTHER_LDRFLAGS] = []
-        log(
-            .debug,
-            indent: 1,
-            "Added '\(impartedSettings[.OTHER_LDFLAGS]!)' to imparted OTHER_LDFLAGS"
-        )
-
+        if sourceModule.isCxx {
+            for platform in ProjectModel.BuildSettings.Platform.allCases {
```
jakepetroules, line 492: On Windows, clang uses the MSVC C++ runtime, _not_ libc++, right?

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
daveinglis: I think it came over from the old PIF builder, not sure the history here.

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
daveinglis: looks like it, -lc++ was not found, and it look like the linker is adding the MSVC libraries as c++ packages seem to be building now

daveinglis:
something weird happened with the PR where I could not update it anymore... here (https://github.com/swiftlang/swift-package-manager/pull/9234) is the new one and will close this one

Issue #9220, opened 2025-10-04
SwiftPM selects wrong architecture from multi-arch Swift SDK when using --swift-sdk and --triple


krzyzanowskim:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

When building with a multi-architecture Swift SDK using both `--swift-sdk` and `--triple`, SwiftPM nondeterministically selects the wrong architecture variant from the SDK bundle, ignoring the target triple specified via `--triple`.

The SDK bundle contains both **x86_64** and **aarch64** variants, and even though `--triple aarch64-swift-linux-musl` is specified, SwiftPM sometimes selects the **x86_64** variant instead, causing build failures.

### Expected behavior

SwiftPM should select the SDK variant matching the `--triple` argument (aarch64) from the multi-arch SDK bundle.

### Actual behavior

SwiftPM nondeterministically selects between x86_64 and aarch64 variants

### Steps to reproduce

1. Install a multi-architecture Swift SDK (SDK contains both x86_64-swift-linux-musl and aarch64-swift-linux-musl)
  \$ `swift sdk list`
  \# Output: swift-6.2-RELEASE_static-linux-0.0.1
2. Create a minimal test package:
$ `mkdir -p /tmp/swiftpm-bug-repro && cd /tmp/swiftpm-bug-repro`
$ `swift package init --type executable --name TestApp`
3. Build for aarch64 Linux:
$ `swift build --swift-sdk swift-6.2-RELEASE_static-linux-0.0.1 --triple aarch64-swift-linux-musl -Xswiftc -static-executable`
4. Observe the warning and error:
**warning**: multiple Swift SDKs match ID `swift-6.2-RELEASE_static-linux-0.0.1` and host triple arm64-apple-macosx26.0
**error**: could not find module '_Concurrency' for target 'aarch64-swift-linux-musl';
found: x86_64-swift-linux-musl, at: .../x86_64/usr/lib/swift_static/linux-static/_Concurrency.swiftmodule
5. Run multiple times to observe nondeterministic behavior (sometimes succeeds, sometimes fails with wrong architecture)

### Swift & OS version (output of `swift --version ; uname -a`)

Swift version: Apple Swift version 6.2 (swift-6.2-RELEASE)
[...]

Issue #7330, opened 2024-02-11
`--triple` option doesn't take installed Swift SDKs into account


MaxDesiatov:
### Description

It's confusing that the existing `--triple` option does look at target triples of installed SDKs and seems to be completely detached from the Swift SDK infrastructure. We should be able to look at the list of installed SDKs and their supported target triples and select the SDK based on that. Probably for multiple Swift SDKs supporting the same target triple we should fail with an error message.

### Expected behavior

`swift build --triple wasm32-unknown-wasi` picks up installed Swift SDK that supports triple.

### Actual behavior

`swift build --triple wasm32-unknown-wasi` doesn't find the installed Swift SDK.

### Steps to reproduce

1. Install a Swift SDK that supports `wasm32-unknown-wasi` target triple
2. Change current directory to the root directory of a package that you'd like to build.
3. Run `swift build --triple wasm32-unknown-wasi`

### Swift Package Manager version/commit hash

Swift Package Manager - Swift 5.11.0-dev

### Swift & OS version (output of `swift --version ; uname -a`)

Swift version 5.11-dev (LLVM 87ace14daa2139a, Swift 8bb684628593220)
Target: aarch64-unknown-linux-gnu

kateinoigakukun:
I think we can filter non-compatible SDKs that contain swiftmodules compiled by different compiler version than the host one and don't contain compiler tools. If we apply such filter, we can get single SDK for most senario

PR #9229, opened 2025-10-08
Finally wire up the `sdk configure` settings to compilation

### LINKED ISSUE #8958: swift sdk configure settings are not used for cross-compilation

### Description

The configured values specified as part of `swift sdk configure` are not being used when building with `swift build --swift-sdk …`. For example, if I install the static Linux SDK and then configure it with invalid values (non-existent folders) for the various options, and then try to build a project, the build should fail. Instead, it appears the values from the SDK's `swift-sdk.json` are used regardless of the attempted configuration overrides.

```
$ swift sdk list
swift-6.1.2-RELEASE_static-linux-0.0.1

$ swift sdk configure swift-6.1.2-RELEASE_static-linux-0.0.1 aarch64-swift-linux-musl --sdk-root-path /bogus --swift-resources-path /bogus --swift-static-resources-path /bogus --include-search-path /bogus --library-search-path /bogus --toolset-path /bogus
info: These properties of Swift SDK `swift-6.1.2-RELEASE_static-linux-0.0.1` for target triple `aarch64-swift-linux-musl` were successfully updated: sdkRootPath, swiftResourcesPath, swiftStaticResourcesPath, includeSearchPath, librarySearchPath, toolsetPath.

$ swift sdk configure swift-6.1.2-RELEASE_static-linux-0.0.1 aarch64-swift-linux-musl --show-configuration
sdkRootPath: /bogus
swiftResourcesPath: /bogus
swiftStaticResourcesPath: /Users/marc/Library/org.swift.swiftpm/swift-sdks/swift-6.1.2-RELEASE_static-linux-0.0.1.artifactbundle/swift-6.1.2-RELEASE_static-linux-0.0.1/swift-linux-musl/musl-1.2.5.sdk/aarch64/usr/lib/swift_static
includeSearchPaths: ["/bogus"]
librarySearchPaths: ["/bogus"]
toolsetPaths: ["/bogus"]

$ swift build --swift-sdk aarch64-swift-linux-musl -v
warning: 'testing-sample': /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/bin/swift-frontend -frontend -c -primary-file /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/Package.swift -target arm64-apple-macosx13.0 -Xllvm -aarch64-use-tbi -enable-objc-interop -sdk /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.5.sdk -I /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/macosx/testing -I /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/lib -I /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/pm/ManifestAPI -F /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/Library/Frameworks -F /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/Library/PrivateFrameworks -vfsoverlay /var/folders/zl/wkdjv4s1271fbm6w0plzknkh0000gn/T/com.apple.shortcuts.mac-helper/TemporaryDirectory.hTh7HN/vfs.yaml -swift-version 6 -package-description-version 6.1.0 -empty-abi-descriptor -plugin-path /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/host/plugins/testing -resource-dir /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift -module-name main -in-process-plugin-server-path /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/host/libSwiftInProcPluginServer.dylib -plugin-path /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/host/plugins -plugin-path /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/local/lib/swift/host/plugins -target-sdk-version 15.5 -target-sdk-name macosx15.5 -external-plugin-path '/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/lib/swift/host/plugins#/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/bin/swift-plugin-server' -external-plugin-path '/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/local/lib/swift/host/plugins#/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/bin/swift-plugin-server' -o /var/folders/zl/wkdjv4s1271fbm6w0plzknkh0000gn/T/com.apple.shortcuts.mac-helper/TemporaryDirectory.s92t4n/Package-1.o
/Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/bin/clang /var/folders/zl/wkdjv4s1271fbm6w0plzknkh0000gn/T/com.apple.shortcuts.mac-helper/TemporaryDirectory.s92t4n/Package-1.o -F /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/Library/Frameworks -F /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/Library/PrivateFrameworks --sysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.5.sdk --target=arm64-apple-macosx13.0 /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/macosx/libswiftCompatibilityPacks.a -L /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/macosx -L /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.5.sdk/usr/lib/swift -L /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/pm/ManifestAPI -L /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/macosx/testing -L /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/lib -lPackageDescription -Xlinker -rpath -Xlinker /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/pm/ManifestAPI -o /var/folders/zl/wkdjv4s1271fbm6w0plzknkh0000gn/T/com.apple.shortcuts.mac-helper/TemporaryDirectory.vqMGon/testing-sample-manifest
Apple Swift version 6.1.2 (swift-6.1.2-RELEASE)
Planning build
Building for debugging...
Write auxiliary file /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/swift-version-D4C7CC622E9D0AC.txt
/Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/bin/swiftc -module-name LibraryDemo -emit-dependencies -emit-module -emit-module-path /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/Modules/LibraryDemo.swiftmodule -output-file-map /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/LibraryDemo.build/output-file-map.json -parse-as-library -incremental -c @/opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/LibraryDemo.build/sources -I /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/Modules -target aarch64-swift-linux-musl -v -enable-batch-mode -index-store-path /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/index/store -Onone -enable-testing -j10 -DSWIFT_PACKAGE -DDEBUG -module-cache-path /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/ModuleCache -resource-dir /Users/marc/Library/org.swift.swiftpm/swift-sdks/swift-6.1.2-RELEASE_static-linux-0.0.1.artifactbundle/swift-6.1.2-RELEASE_static-linux-0.0.1/swift-linux-musl/musl-1.2.5.sdk/aarch64/usr/lib/swift_static -parseable-output -parse-as-library -color-diagnostics -swift-version 6 -static-executable -static-stdlib -tools-directory /Users/marc/Library/org.swift.swiftpm/swift-sdks/swift-6.1.2-RELEASE_static-linux-0.0.1.artifactbundle/swift-6.1.2-RELEASE_static-linux-0.0.1/swift-linux-musl/swift.xctoolchain/usr/bin -sdk /Users/marc/Library/org.swift.swiftpm/swift-sdks/swift-6.1.2-RELEASE_static-linux-0.0.1.artifactbundle/swift-6.1.2-RELEASE_static-linux-0.0.1/swift-linux-musl/musl-1.2.5.sdk/aarch64 -g -Xcc --sysroot -Xcc /Users/marc/Library/org.swift.swiftpm/swift-sdks/swift-6.1.2-RELEASE_static-linux-0.0.1.artifactbundle/swift-6.1.2-RELEASE_static-linux-0.0.1/swift-linux-musl/musl-1.2.5.sdk/aarch64 -Xcc -g -package-name testing_sample -Xcc -fno-omit-frame-pointer
Apple Swift version 6.1.2 (swift-6.1.2-RELEASE)
Target: aarch64-swift-linux-musl
/Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/bin/swift-frontend -frontend -emit-module -experimental-skip-non-inlinable-function-bodies-without-types /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/Sources/LibraryDemo/LibraryDemo.swift -target aarch64-swift-linux-musl -Xllvm -aarch64-use-tbi -disable-objc-interop -sdk /Users/marc/Library/org.swift.swiftpm/swift-sdks/swift-6.1.2-RELEASE_static-linux-0.0.1.artifactbundle/swift-6.1.2-RELEASE_static-linux-0.0.1/swift-linux-musl/musl-1.2.5.sdk/aarch64 -I /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/Modules -color-diagnostics -enable-testing -g -debug-info-format=dwarf -dwarf-version=4 -module-cache-path /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/ModuleCache -swift-version 6 -Onone -D SWIFT_PACKAGE -D DEBUG -empty-abi-descriptor -resource-dir /Users/marc/Library/org.swift.swiftpm/swift-sdks/swift-6.1.2-RELEASE_static-linux-0.0.1.artifactbundle/swift-6.1.2-RELEASE_static-linux-0.0.1/swift-linux-musl/musl-1.2.5.sdk/aarch64/usr/lib/swift_static -use-static-resource-dir -enable-anonymous-context-mangled-names -file-compilation-dir /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample -Xcc --sysroot -Xcc /Users/marc/Library/org.swift.swiftpm/swift-sdks/swift-6.1.2-RELEASE_static-linux-0.0.1.artifactbundle/swift-6.1.2-RELEASE_static-linux-0.0.1/swift-linux-musl/musl-1.2.5.sdk/aarch64 -Xcc -g -Xcc -fno-omit-frame-pointer -module-name LibraryDemo -package-name testing_sample -in-process-plugin-server-path /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/host/libSwiftInProcPluginServer.dylib -plugin-path /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/lib/swift/host/plugins -plugin-path /Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/local/lib/swift/host/plugins -emit-module-doc-path /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/Modules/LibraryDemo.swiftdoc -emit-module-source-info-path /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/Modules/LibraryDemo.swiftsourceinfo -emit-dependencies-path /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/LibraryDemo.build/LibraryDemo.emit-module.d -parse-as-library -o /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/Modules/LibraryDemo.swiftmodule
/Users/marc/Library/Developer/Toolchains/swift-6.1.2-RELEASE.xctoolchain/usr/bin/swift-frontend -modulewrap /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/Modules/LibraryDemo.swiftmodule -target aarch64-swift-linux-musl -o /opt/src/github/swift-android-sdk/swift-android-samples/testing-sample/.build/aarch64-swift-linux-musl/debug/Modules/LibraryDemo.o
Build complete! (0.37s)
```

If the configured values were being used, the build command would not include the default value from `swift-sdk.json` (`-Xcc --sysroot -Xcc /Users/marc/Library/org.swift.swiftpm/swift-sdks/swift-6.1.2-RELEASE_static-linux-0.0.1.artifactbundle/swift-6.1.2-RELEASE_static-linux-0.0.1/swift-linux-musl/musl-1.2.5.sdk/aarch64`), but instead would reference `/bogus`, and the build command would fail.

This is irrespective of the recent fixes around `swift sdk configure` in  https://github.com/swiftlang/swift-package-manager/pull/8687 and https://github.com/swiftlang/swift-package-manager/pull/8856.

### Expected behavior

The values specified as part of `swift sdk configure` should be used when compiling with a Swift SDK. Specifying invalid values (e.g., paths that do not exist) for parameters like `sdk-root-path` should cause compilation to fail.

### Actual behavior

The values for the various SDK properties seem to be ignored when 

### Steps to reproduce

1. `swift sdk install https://download.swift.org/swift-6.1.2-release/static-sdk/swift-6.1.2-RELEASE/swift-6.1.2-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz --checksum df0b40b9b582598e7e3d70c82ab503fd6fbfdff71fd17e7f1ab37115a0665b3b`
2. `swift sdk configure swift-6.1.2-RELEASE_static-linux-0.0.1 aarch64-swift-linux-musl --sdk-root-path /bogus --swift-resources-path /bogus --swift-static-resources-path /bogus --include-search-path /bogus --library-search-path /bogus --toolset-path /bogus`
3. `swift sdk configure swift-6.1.2-RELEASE_static-linux-0.0.1 aarch64-swift-linux-musl --show-configuration`
4. `mkdir SampleLibrary`
5. `cd SampleLibrary`
6. `swift package init`
7. `swift build --swift-sdk aarch64-swift-linux-musl`
8. Build should fail due to invalid paths being used, but it succeeds.

swift-6.1.2-RELEASE

Apple Swift version 6.1.2 (swift-6.1.2-RELEASE)
Target: arm64-apple-macosx15.0
Darwin zap.local 24.5.0 Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:49 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6000 arm64

finagolfin:
This will require more discussion, but this attempts to fix #8958, will have more info soon.

Commit message: This only affects the native build system currently, with swift-build requiring more changes.


Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
```diff
@@ -761,6 +761,50 @@ public struct SwiftSDK: Equatable {
...
+                               targetTriple: targetTriple)
+                            {
+                                matchSDK = readSDK
+                            }
+                        }
+                        if multipleConfigMatches {
+                observabilityScope.emit(
+                    warning: """
+                        Multiple SDKs have been configured with this triple \(swiftSDKSelector), so
+                            dropping back to values from original bundle.
```
finagolfin: The reason I had to add all this logic is because one can specify `--swift-sdk triple` instead of a Swift bundle name, what is termed an artifact ID in the code, and since that triple may be supplied by multiple installed Swift SDKs, the existing logic simply picks the first SDK it finds and warns that it found multiple matching SDKs. Since this local config override then doesn't know which SDK bundle was chosen, as the bundle artifact ID picked is not passed back by these internal methods, this logic has to again search for the bundle chosen and can't use the local config override if multiple bundles matched.

Instead, I would like to make two independent changes to these SDK selection methods:

1. Only return a result if there is a unique matching SDK bundle ID or triple, error if not. If there is a unique match, pass back both the bundle artifact ID and the SDK, so that this local config override can then be uniquely checked.
2. Do not allow passing in triples to `--swift-sdk`, use the `--triple` flag instead. The use of that flag is currently broken with SDK bundles, as reported in #7330 and #9220,  so this change would also fix those issues.

@MaxDesiatov, let me know what you think of such proposed fixes, which this pull does not implement, as this pull only uses existing SwiftPM methods.

finagolfin: @dschaefer2, [not getting any response from Max](https://github.com/swiftlang/swift-package-manager/issues/8958#issuecomment-3383069579), who implemented this subcommand, [doesn't seem like he's active in this repo anymore](https://github.com/swiftlang/swift-package-manager/commits?author=MaxDesiatov). Is there anybody else I can discuss these proposed changes with or should I just go ahead and make them?

MaxDesiatov: I can have a look when this is ready for review and not marked as a work in progress draft.

MaxDesiatov: (modulo any other work I and other people are doing, sometimes lead time for a review can be much longer than 2 days because of that)

finagolfin: @MaxDesiatov, as I explained in detail starting two weeks ago, it is only marked as a draft because I think this overall approach isn't great and that we should change more of the underlying and user-facing behavior.

Let me know when you get a chance to think about those deeper changes I've proposed.

MaxDesiatov: In accordance with semantic versioning, these changes must be made in a backward-compatible manner with an appropriate deprecation period. I'm fine with these changes in principle, as long as we don't break anything for existing users, notify about the desired migration path, and give them time to migrate before the next major release.

finagolfin: Alright, I will make the changes, let's see. One key question I asked a couple weeks ago still stands, can I substantially alter a [seemingly non-public extension method like `selectSwiftSDK(selector:hostTriple:observabilityScope)`](https://github.com/swiftlang/swift-package-manager/blob/4a6660ac95f4b2f88e1006f84576b0ca16343eb9/Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift#L84) for both its behavior and declaration, or would I have to create a new method and deprecate the old one?

As for backwards compatibility and migration paths, the current behavior is buggy, non-deterministic, and incomplete for these two features, so pretty sure that making it strict and work right won't be "compatible" 😉, but we can try for the parts that were working.

MaxDesiatov: IIRC none of the Swift SDK code in SwiftPM is `public`, so that's not an API and hence not something that has to follow SemVer. I'm referring to CLI commands, and their arguments, options, and flags.

finagolfin:
@dschaefer and @jakepetroules, got delayed by my CI and release work this week, so I will put up an expanded prototype of this pull this weekend, along the lines I discussed above, and once we hammer out the final behavior we want, I will shape it to that spec and add tests in the coming week.

Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
```diff
@@ -786,7 +786,22 @@ public struct SwiftSDK: Equatable {
...
+                        hostTimeTriple: hostTriple,
+                        swiftSDKBundleStore: store
+                    )
+                    if let configSDK = try configurationStore.readConfiguration(
+                        sdkID: ID,
+                        targetTriple: swiftSDK.targetTriple!)
+                    {
+                        swiftSDK = configSDK
+                    }
+                } catch { /* do nothing to override if a custom SDK config is not found */ }
```
finagolfin: Since the above `selectBundle()` will get the default config from the installed bundles, simply do nothing if this logic to get the results of the user's local `sdk configure` config fails for some reason.

Maybe better warnings could be added in this block instead.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
```diff
@@ -67,7 +67,7 @@ extension [SwiftSDKBundle] {
                         }
                     }
 
-                    return variant.swiftSDKs.first { $0.targetTriple == targetTriple }
+                    return variant.swiftSDKs.first { $0.targetTriple?.tripleString == targetTriple.tripleString }
```
finagolfin, line 68: This was causing arbitrary Android versions to match, so it would just pick the first Android triple version that it would find, eg specifying `aarch64-unknown-linux-android21` would match `aarch64-unknown-linux-android28`, even though the former is not even in the Android SDK! 😄

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
```diff
@@ -79,72 +79,35 @@ extension [SwiftSDKBundle] {
...
-                warning:
-                """
-                multiple Swift SDKs match the query `\(selector)` and host triple \(
-                    hostTriple.tripleString
-                ), selected one at \(matchedByID.path.appending(matchedByID.variant.metadata.path))
-                """
-            )
-        }
-
-        return matchedByID?.swiftSDK ?? matchedByTriple?.swiftSDK
```
finagolfin: While it's nice that Max tried to give the user more leeway here, this is all highly indeterminate, so I just started erroring on all these cases instead, see `SwiftSDKBundleStore.swift`.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift
```diff
@@ -136,17 +166,41 @@ public final class SwiftSDKBundleStore {
...
+        if let triple = targetTriple {
+            guard var selectedSwiftSDK = validBundles.selectSwiftSDK(
+                id: selector,
+                hostTriple: hostTriple,
+                targetTriple: triple
+            ) else {
+                throw Error.noMatchingSwiftSDKWithTriple(selector: selector, hostTriple: hostTriple, targetTriple: triple)
+            }
+            selectedSwiftSDK.applyPathCLIOptions()
+            return (selector, selectedSwiftSDK)
```
finagolfin, line 179: This finally implements a working `--triple` flag with Swift SDKs, no problems in my testing.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift
```diff
@@ -136,17 +166,41 @@ public final class SwiftSDKBundleStore {
...
+        }
+
+        var selectedSwiftSDK: SwiftSDK
+        var id: String
+        let SDKs = validBundles.selectSwiftSDK(matching: selector, hostTriple: hostTriple)
+        if SDKs.tripleMatches.count > 1 {
+            throw Error.multipleSDKMatchesForTriple(selector: selector, hostTriple: hostTriple, matches: Array(SDKs.tripleMatches.keys))
+        } else if SDKs.idMatches.count > 1 {
+            throw Error.multipleSDKMatchesForID(selector: selector, hostTriple: hostTriple)
```
finagolfin, line 188: The previous code would warn and pick a random SDK from the matches instead, which caused indeterminate behavior.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift
```diff
@@ -136,17 +166,41 @@ public final class SwiftSDKBundleStore {
...
+        } else if SDKs.idMatches.count == 1 {
+            guard SDKs.tripleMatches.count == 0 else {
+              throw Error.matchingBothToSwiftSDK(selector: selector, hostTriple: hostTriple)
```
finagolfin, line 191: This is fairly unlikely, ie that the selector matches both a triple and the name of an SDK, but I put it in just because it _is_ possible.

finagolfin:
This fixes a handful of bugs and keeps the same basic interface that Swift SDKs have had for years now, but makes the `--triple` option actually work and errors in uncertain situations, like if multiple SDKs vend the same triple, rather than the previous behavior of just picking one of those matching SDKs, warning about it, and moving ahead!

After thinking about it more, I think this is the better approach, as it will keep `--swift-sdk triple` working if you know what you're doing. We should update all doc to tell users to explicitly pass in `--swift-sdk SDK-name --triple triple` instead, then we can deprecate the `--swift-sdk triple` option over time.

Of course, my initial work to fix the `swift sdk configure` command is also included, as that local config was never actually passed into compilation, as detailed in #8958. Apply this pull and follow [the instructions in the doc to build SwiftPM](https://github.com/swiftlang/swift/blob/68036ac35d0e50d3a0c4f9dca8ed1d2e310c070b/docs/HowToGuides/GettingStarted.md#the-actual-build) to try these changes out, as I just did on macOS.

I have not bothered running this through the tests or adding more yet, as I want to get this direction agreed on by reviewers first. Instead, I built the toolchain and tried a bunch of these commands, until nothing failed anymore like they were before.

Once the overall direction here is agreed on, I will add the last finish of updating the remaining doc comments and fixing and adding tests.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
owenv: I agree matching triples with different versions is obviously wrong, I do wonder if we should allow a user-specified unversioned triple to match a versioned triple in the SDK if there is only one. Or potentially flag a 'near miss' as a descriptive error.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift
owenv: Personally I'm very much in favor of this behavior change but it would be good to ensure we have consensus and give others a bit of time to chime in

owenv (commented):
Thank you! I like this general direction, will give others a chance to review as well and make sure there aren't any high-level objections.

I do think it may be worth breaking out the `sdk configure` and Swift SDK selection changes into separate PRs, both to make them easier to review and because I think we may need to do a bit of work to get `sdk configure` working with the new build system backend (that is currently reading SDKs directly from disk and may need to handle the merging the in-memory customizations differently)

jakepetroules:
Agree with everything @owenv said. Also very strongly in favor of erroring in any ambiguous situations.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
jakepetroules: >  I do wonder if we should allow a user-specified unversioned triple to match a versioned triple in the SDK if there is only one

Agree with that, as FreeBSD SDKs will probably hit this scenario.

jakepetroules: I'm a little confused though, as your patch retains the "first" behavior... don't you need to change this to error if there's >1?

Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift
```diff
@@ -136,17 +166,41 @@ public final class SwiftSDKBundleStore {
...
+            }
+            id = selector
+            selectedSwiftSDK = SDKs.idMatches[0]
+        } else if SDKs.tripleMatches.count == 1 {
+            id = SDKs.tripleMatches.first!.key
```
jakepetroules: prefer `[0]`, I hate to see force unwraps

Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift
jakepetroules: Definitely should error out instead of picking something at random.

finagolfin:
Since this pull no longer removes the `--swift-sdk triple` option, [as I originally proposed months ago](https://github.com/swiftlang/swift-package-manager/pull/9229#discussion_r2447111876), can we get this into a 6.3 patch release also? If so, should we [aim for 6.3.1](https://forums.swift.org/t/development-open-for-swift-6-3-1-for-linux-and-windows/85715), or a subsequent one that includes macOS? It won't be good if it's fixed for some platforms but not others, but I think we should get this into a 6.3 release, as this is a definite UX problem for our expanding SDK consumers, and these are clear bugs I'm fixing.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift
dschaefer2: Or better yet `else if let id = SDKs.tripleMatches.first?.key`

dschaefer2:
I like the direction of this and agree we should review this change with the community. We need to get more in the habit of doing that.

As whether this qualifies for 6.3.x, I worry about making behaviour changes like this in a maintenance release. But if it fixes a defect for Android, we should consider it.

finagolfin:
> As whether this qualifies for 6.3.x, I worry about making behaviour changes like this in a maintenance release. But if it fixes a defect for Android, we should consider it.

These are all bugs with the `swift sdk` feature that have been around for years, nothing here is specific to Android (even the Android triple versioning I mentioned above was likely only one way in which that triple checking was broken, though that method was only used by the broken `swift sdk configure` before this pull).

For example, this fixes #9220, which was reported with the static linux SDK, but affects all multi-target SDK bundles.

I'll break this off into separate pulls, as Owen suggested, because some of these commits are just bugfixes, which should not be gated on how long we take to decide changing the behavior of the `--swift-sdk` flag.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift
jakepetroules: No, that would drop the `count == 1` check, which is important.

finagolfin: Hmm, actually Doug is right, we already checked if there was more than one match, so his approach would work, 👍 simply another way of checking if a match exists.

dschaefer2: In the past if I need it to be one and only one we do have an extension somewhere called `only` that returns nil if there are none or more than one.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
finagolfin: Note that this `selectSwiftSDK(id: hostTriple: targetTriple)` "method was only used by the broken `swift sdk configure` before this pull," as I mentioned below, so it was essentially unused until I started using it for `--triple <triple>` checking in this pull and fixed `swift sdk configure` so it _can_ be used.

> I do wonder if we should allow a user-specified unversioned triple to match a versioned triple in the SDK if there is only one. Or potentially flag a 'near miss' as a descriptive error.

I'd rather not add these new features alongside these bug fixes. If you'd like those search features added, please file an issue and we can get them in later. This change merely gets `--triple <triple>` to work and [exactly matches the triple checking we've long had](https://github.com/swiftlang/swift-package-manager/pull/9229/changes#diff-c15b78e0314b27acd57daf9c85e41778ed1cb107b685bdaa6e7a552d4dce4791R103) with `--swift-sdk <triple>`, ie no new triple-matching behavior.

> your patch retains the "first" behavior... don't you need to change this to error if there's >1?

I thought about that too and kept it the way it was written, because I figured it is unlikely for SDK bundlers to make that mistake of specifying the same target triple multiple times and that the existing bundle verification probably already catches that issue.

However, I just tried manually editing the Android bundle to repeat some triples, and there is no such verification, either before or after installing: it just grabs the first matching triple it finds from multiple candidates in a bundle's JSON config. 🤨

The question now is how to address this unlikely potential problem in SDK bundles: should we fix it when [verifying the SDK before installing](https://github.com/swiftlang/swift-package-manager/blob/69f851d48eacdfc19c2cb94eb61e4c0858713474/Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift#L317), as [we do for the name of the SDK itself](https://github.com/swiftlang/swift-package-manager/blob/69f851d48eacdfc19c2cb94eb61e4c0858713474/Sources/PackageModel/SwiftSDKs/SwiftSDKBundleStore.swift#L324), or put in some error-checking here when searching instead?

owenv:
@finagolfin sorry for the lack of comment the past couple weeks. I'd like to keep making progress on this, even if it means subsetting out a few parts to make incremental progress. 

Based on the discussion so far, if you're willing to split out the changes to SwiftSDKBundle.swift and SwiftSDKBundleStore.swift which fix the SDK selection ambiguity issues, I'm ready to approve that change now and try to get it landed ASAP in at least `main` before 6.4 branches. As we can work on that maybe we can continue deciding whether we should attempt to back port it to an earlier release branch, and keep discussing some of the other changes in this PR.

finagolfin:
No problem, @owenv, I figured we were just keeping this pull up to "review this change with the community," as Doug said before. I [just addressed some of your and Jake's comments on triple matching](https://github.com/swiftlang/swift-package-manager/pull/9229#discussion_r3069256878) a couple days ago as part of that review. While I will make a few more tweaks as needed and finally add tests, this pull is mostly done.

I see three parts to this pull, as can easily be seen by going through each commit individually:

1. the fix to get `--triple <triple>` working
2. the changes to error on multiple SDK matches, rather than just randomly picking one of the matches
3. the fixes to get `swift sdk configure` finally working

I'm ready to split out 1., but is it okay to keep the latter two combined here, because [I changed the return signature of `selectBundle()` in 2.](https://github.com/swiftlang/swift-package-manager/pull/9229/changes/5859ade5be71d542d8a5696fa268cbde7ac27b83#diff-c15b78e0314b27acd57daf9c85e41778ed1cb107b685bdaa6e7a552d4dce4791R86) so I can use it in 3., plus the latter is pretty small anyway? If you really want them split conceptually, I can separate this into three pulls along the above lines, so just let me know if I should split the latter two also.

owenv:
Sure, how about we split out just 1 for now and I'll approve as an obvious bugfix assuming no last-minute objections.

I would really like to get 2) in as well as I think it's a nice UX improvement. Maybe that's worth a quick forums post to satisfy concerns about running this by the community first? If folks do end up thinking that merits a proposal cycle, I'd be happy to help push that process forward if it's a concern. If the consensus is we don't need one, I think we go ahead and land it.

I don't have a strong opinion on 3), I need to set aside some time to review it properly. I don't see any obvious issue with your patch

finagolfin:
Alright, I will split out 1. and add tests for it, plus post on the forums about 2. As you said above, I have not tried this pull with the new `swift-build` default, only the previous `llbuild` backend, so I will make sure that works too.

I'd like to get all these fixes in as soon as possible, including into a 6.3 patch release, so just let me know what else is needed as I submit the final versions of these fixes.

jakepetroules:
I'll add my input that (2) seems like such an obvious bug fix I can't see how there would possibly be any controversy about it. That said, let's get that forums post up ASAP so we can gauge that, and then hopefully move forward with landing the fix very soon afterwards. It would be a shame for any of these fixes to miss 6.4.

As far as Swift Build support for (3), this might need new API, because it looks like the `swift sdk configure` creates a sidecar configuration directory with JSON blobs that are meant to overlay a single triple within a single SDK... not sure how we'll want to represent that. Today Swift Build just accepts a file path to the swift-sdk.json file. I don't know if the structure of the configuration/ directory is considered an implementation detail or was actually specced as part of the original Swift SDKs proposal, but if it's a mere implementation detail we could make things easier by defining it to shadow the entire swift-sdk.json and then we could just pass the path to that one into Swift Build without API changes. Otherwise we might need a map of triples to overlay paths as an additional property in the build request alongside the main swift SDK JSON path.

finagolfin:
I just went ahead and spun off 2. as #9937.

finagolfin:
Spun off 1. as #9955, plus I'll see if I can get #7330, ie the `--triple` flag alone, working with SDK bundles. Once I have some new tests and these pulls passing CI, we can start getting them in, assuming we don't need to wait any more for external review feedback.

> I don't know if the structure of the configuration/ directory is considered an implementation detail or was actually specced as part of the original Swift SDKs proposal, but if it's a mere implementation detail we could make things easier by defining it to shadow the entire swift-sdk.json and then we could just pass the path to that one into Swift Build without API changes.

Implementation detail, and that's basically all the last commit here does, in a different way perhaps, to get `swift sdk configure` working finally.

finagolfin:
Not ready yet, but just rebasing and checking linux CI

@swift-ci test linux

finagolfin:
Tested this more locally, added a warning for when the local config cannot be read, and crimped in some tests. This worked with the native build system in my local testing and all tests passed on macOS, but `swift-build` appears to ignore these locally configured settings from the target SDK. Since that will need to be fixed separately in the `swift-build` repo anyway, after discussing precisely what our configuration approach will be, as currently [`swift-build` even ignores the toolset _in_ the Android SDK bundle](https://github.com/swiftlang/swift-build/blob/8e0514c4f9c02180836ad5596eae4abfe4febd2d/Sources/SWBAndroidPlatform/Plugin.swift#L177), pushed that fix pending further discussion.

finagolfin:
@ken0nek, please review, since you saw the first bug too.

@swift-ci test

Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
```diff
@@ -786,7 +786,28 @@ public struct SwiftSDK: Equatable {
...
+                    let configurationStore = try SwiftSDKConfigurationStore(
+                        hostTimeTriple: hostTriple,
+                        swiftSDKBundleStore: store
+                    )
+                    if let configSDK = try configurationStore.readConfiguration(
+                        sdkID: ID,
+                        targetTriple: swiftSDK.targetTriple!)
+                    {
+                        swiftSDK = configSDK
+                    }
```
ken0nek: Could `SwiftSDKConfigurationStore` grow a sibling method that returns just the persisted overlay (a `PathsConfiguration?`), so we can merge it into the SDK we already selected, instead of taking a re-selected full `SwiftSDK`? Roughly:

```swift
if let targetTriple = swiftSDK.targetTriple,
   let overlay = try configurationStore.readPathsOverlay(sdkID: id, targetTriple: targetTriple) {
    swiftSDK.pathsConfiguration.merge(with: overlay)
}
```

The existing `readConfiguration` would stay as-is for `swift sdk show` (`ConfigurationSubcommand.swift:67`) and the self-call in `configure()`. The new method would just sit alongside it for the compile path.

Reasoning:
1. `readConfiguration` re-walks `allValidBundles` and re-runs `selectSwiftSDK`, but we already have the bundle from `selectBundle` two lines up.
2. That re-selection passes `hostTriple: nil`. Per `SwiftSDKBundle.swift:56-76`, nil skips the host filter and returns the first variant's matching target, which for multi-host bundles can differ from the host-resolved variant we already have.
3. `readConfiguration` returns the SDK whether or not a configuration JSON exists, so `swiftSDK = configSDK` runs for every installed bundle, not just users who ran `swift sdk configure`. An overlay returning `nil` when there's no file gives a clean opt-in signal.
4. The configure subcommand stores path overrides, and `readConfiguration` merges only `pathsConfiguration` from the persisted JSON, so replacing the whole SDK reaches further than the configure data warrants.

finagolfin: Good catch on 2., I don't think we have any multi-host SDKs yet, but that was a bug in waiting. I think we just tighten that up to a required parameter, as it is for the other `selectSwiftSDK(selector:hostTriple)` overload.

I don't think your efficiency concerns in 1. and 3. matter too much, as most people won't have many bundles installed, but 4. _is_ a good reason to implement what you suggest, as it minimizes what's modified, lessening the risk of bugs like 2.

I'll add that and update this pull.

finagolfin:
In looking at uses of `readConfiguration()` currently, I see it is only used two places, which should also be updated to use this new method Ken suggested instead. One is in the old `sdk configuration` command, which [has been marked as deprecated for two years](https://github.com/swiftlang/swift-package-manager/pull/7535/changes#diff-e0c5e27d53cd80f6fc5b9ccf4ac1133267ba1369e7cd949c7f197e8581f7d188R21) now, the other is when invoking `swift sdk configure`.

@jakepetroules, what do you think about two further changes in this pull?

- Remove all `sdk configuration` code finally.
- Remove all use of `readConfiguration()`, replacing it with a new `readPathsConfiguration()` method that both this pull and `sdk configure` use. I can keep `readConfiguration()` itself around if wanted, or mark it as deprecated.

These changes represent zero risk, because while these SDK configuration commands would update and modify internal config files, those files have never been used for SDK compilation, until this pull.

jakepetroules:
Sounds reasonable I suppose.

finagolfin:
I modified `readConfiguration()` to pass in an `inout SwiftSDK` instead, and just modify that passed-in SDK if there's a config file. Been testing this all out locally, works well to configure, but found that the `--library-search-path` and `--toolset-path` flags still don't affect compilation, while [the other four do](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/sdkconfigure). I'll see if I can track down those bugs and update this pull soon.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
```diff
@@ -53,18 +53,16 @@ extension [SwiftSDKBundle] {
...
                     continue
                 }
 
                 for variant in variants {
-                    if let hostTriple {
-                        guard variant.isSupporting(hostTriple: hostTriple) else {
-                            continue
-                        }
+                    guard variant.isSupporting(hostTriple: hostTriple) else {
+                        continue
```
finagolfin, line 65: Fixed the issue that @ken0nek found

Sources/PackageModel/SwiftSDKs/SwiftSDKConfigurationStore.swift
```diff
@@ -85,56 +85,31 @@ public final class SwiftSDKConfigurationStore {
...
-        return []
-    }
-
     public func readConfiguration(
         sdkID: String,
-        targetTriple: Triple
-    ) throws -> SwiftSDK? {
+        sdk: inout SwiftSDK
+    ) throws -> SwiftSDK {
+        let targetTriple = sdk.targetTriple!
```
finagolfin, line 92: @ken0nek, passing in the mutable SDK here let's me avoid loading all the bundles below, without having to pass the configuration around and merge it each time. Let me know what you think.

jakepetroules (approved):

Sources/PackageModel/SwiftSDKs/SwiftSDKConfigurationStore.swift
```diff
@@ -173,33 +148,34 @@ public final class SwiftSDKConfigurationStore {
...
-            let validBundles = try self.swiftSDKs(for: sdkID)
-            targetTriples = validBundles.compactMap(\.targetTriple)
-            if targetTriples.isEmpty {
-                throw SwiftSDKError.swiftSDKNotFound(
-                    artifactID: sdkID,
-                    hostTriple: hostTriple,
-                    targetTriple: nil
-                )
-            }
+            sdks = bundles.selectSwiftSDK(matching: sdkID, hostTriple: hostTriple).idMatches
```
finagolfin, line 165: @marcprux, refactored this code to operate on SwiftSDKs and to use the existing `selectSwiftSDK()` methods, rather than your new private `swiftSDKs()` method that wasn't checking the host and would have hit [the same bug Ken raised above for future multi-host SDKs](https://github.com/swiftlang/swift-package-manager/pull/9229#discussion_r3149543976).

finagolfin:
Updated with the changes discussed: this works well in my local testing, except for the aforementioned broken library search and toolset path integration, and passed all tests on macOS.

I will look for fixes for those last two bugs and add some tests, and this should be ready to go by Wednesday morning PDT.

@swift-ci test

finagolfin:
Found the two fixes: #10008 has the library search path fix, which is completely unrelated to this `sdk configure` command so it affects the config file that comes with the SDK bundle too, and the last commit here integrates the toolset paths. Also removed tests for the now-removed `sdk configuration` subcommand, and modified some tests for the `sdk configure` refactoring of `readConfiguration()`.

Just need to work out a test for the last commit next, then this can go in.

finagolfin:
The toolset path fix I had put up in 9f884aa3e broke a single SDK bundle test case, so I will spin that off in a separate pull.

This passes all tests for me locally, @owenv, let me know if you want to review too.

@swift-ci test

finagolfin:
@swift-ci test windows

PR #9232, opened 2025-10-08
Default to bootstrapping using Swift Build

owenv:
I believe we've resolved most or all the blockers here, let's see what's left


owenv:
@swift-ci test

owenv:
@swift-ci test

PR #9234, opened 2025-10-09
Cleanup some unnecessary build setting

daveinglis:
- fix adding -lc++ to OTHER_LDFLAGS to match native build system.
- remove OTHER_LDRFLAGS setting

### DIFF
--- a/Sources/SwiftBuildSupport/PackagePIFBuilder.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFBuilder.swift
@@ -552,6 +552,9 @@ public final class PackagePIFBuilder {
         // We currently deliberately do not support Swift ObjC interface headers.
         settings[.SWIFT_INSTALL_OBJC_HEADER] = "NO"
         settings[.SWIFT_OBJC_INTERFACE_HEADER_NAME] = ""
+        
+        // rdar://47937899 (Don't try to link frameworks to object files) 
+        //  - looks like this defaults to OTHER_LDFLAGS (via xcspec) which can result in linking frameworks to mh_objects which is unwanted.
         settings[.OTHER_LDRFLAGS] = []
 
         // Packages use the SwiftPM workspace's cache directory as a compiler working directory to maximize module
--- a/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
@@ -484,14 +484,16 @@ extension PackagePIFProjectBuilder {
         if enableDuplicateLinkageCulling {
             impartedSettings[.LD_WARN_DUPLICATE_LIBRARIES] = "NO"
         }
-        impartedSettings[.OTHER_LDFLAGS] = (sourceModule.isCxx ? ["-lc++"] : []) + ["$(inherited)"]
-        impartedSettings[.OTHER_LDRFLAGS] = []
-        log(
-            .debug,
-            indent: 1,
-            "Added '\(impartedSettings[.OTHER_LDFLAGS]!)' to imparted OTHER_LDFLAGS"
-        )
-
+        if sourceModule.isCxx {
+            for platform in ProjectModel.BuildSettings.Platform.allCases {
+                // darwin & freebsd
+                if [.macOS, .macCatalyst, .iOS, .watchOS, .tvOS, .xrOS, .driverKit, .freebsd].contains(platform) {
+                    settings[.OTHER_LDFLAGS, platform] = ["-lc++", "$(inherited)"]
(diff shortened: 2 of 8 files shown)


daveinglis:
@swift-ci test

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
```diff
@@ -552,7 +552,6 @@ public final class PackagePIFBuilder {
         // We currently deliberately do not support Swift ObjC interface headers.
         settings[.SWIFT_INSTALL_OBJC_HEADER] = "NO"
         settings[.SWIFT_OBJC_INTERFACE_HEADER_NAME] = ""
-        settings[.OTHER_LDRFLAGS] = []
```
jakepetroules, line 555: This is set because OTHER_LDRFLAGS defaults to OTHER_LDFLAGS, and this might've been on purpose if we wanted to set linker flags _without_ those applying to `ld -r` invocations (e.g. when producing mh_objects).

Let's double check whether we actually mean to remove this, and if not, document via a code comment why it's empty.

jakepetroules: Looks like the original purpose was to avoid trying to link frameworks when producing object files as part of `ld -r`.

daveinglis:
@swift-ci test

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
daveinglis: updated

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

daveinglis:
@jakepetroules I need you to review and remove you requested change if you are ok with this.

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
```diff
@@ -484,14 +484,16 @@ extension PackagePIFProjectBuilder {
...
-        impartedSettings[.OTHER_LDRFLAGS] = []
-        log(
-            .debug,
-            indent: 1,
-            "Added '\(impartedSettings[.OTHER_LDFLAGS]!)' to imparted OTHER_LDFLAGS"
-        )
-
+        if sourceModule.isCxx {
+            for platform in ProjectModel.BuildSettings.Platform.allCases {
+                // darwin & freebsd
```
jakepetroules, line 489: Instead of if-else, can you use a switch?

```swift
switch platform {
case .macOS, .macCatalyst, .iOS, .watchOS, .tvOS, .xrOS, .driverKit, .freebsd:
    settings[.OTHER_LDFLAGS, platform] = ["-lc++", "$(inherited)"]
case .android, .linux, .wasi, .openbsd:
    settings[.OTHER_LDFLAGS, platform] = ["-lstdc++", "$(inherited)"]
case .windows:
    break // no flags needed
}
```

That will benefit from exhaustivity checking and the compiler will force us to implement the right flags when we add a new platform (like NetBSD or QNX for example).

Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
```diff
@@ -135,6 +148,28 @@ struct PIFBuilderTests {
...
+    @Test func platformCCLibrary() async throws {
+        try await withGeneratedPIF(fromFixture: "PIFBuilder/CCPackage") { pif, observabilitySystem in
+            let releaseConfig = try pif.workspace
+                .project(named: "CCPackage")
+                .target(id: "PACKAGE-TARGET:CCTarget")
+                .buildConfig(named: "Release")
+
+            for platform in ProjectModel.BuildSettings.Platform.allCases {
+                let ld_flags = releaseConfig.settings[.OTHER_LDFLAGS, platform]
+                if [.macOS, .macCatalyst, .iOS, .watchOS, .tvOS, .xrOS, .driverKit, .freebsd].contains(platform) {
```
jakepetroules, line 160: I'd use a switch here as well

jakepetroules (approved):
Approved but I think there's still a couple improvements you could make (see inline comments)

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
daveinglis: https://github.com/swiftlang/swift-package-manager/pull/9267

Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
daveinglis: https://github.com/swiftlang/swift-package-manager/pull/9267

Issue #8653, opened 2025-05-10
Do no link intermediate swift objects for embedded


rauhul:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

SwiftPM should not link intermediate object files when building for embedded swift. Supersedes: #7612

dschaefer2:
There is discussion about what swift modules should build to. I think we are leaning towards static libraries as apposed to the prelinking that swiftbuild does now. That and the Windows linker can't relink which is why we don't have swiftbuild for Windows yet.

PR #9246, opened 2025-10-14
Stop using mergeable-symbols in embedded

rauhul:
This flag is now a no-op and isn't needed after recent changes to the embedded swift linkage model.

Fixes: #8653


MaxDesiatov:
@swift-ci test

PR #9259, opened 2025-10-17
Fix building frameworks

daveinglis:
- fix pif target product name for frameworks
- add new experimental flag to swift build that will build dylibs as frameworks so we can have a tests that ensure we don't regress.


daveinglis:
@swift-ci test

owenv (approved):

jakepetroules (approved):

daveinglis:
@swift-ci test macOS

Issue #8928, opened 2025-07-11
`swift test` output is buffered (rather than streamed) for swift-testing tests on Windows


jmschonfeld:
### Description

When running `swift test` on Windows in a VS Tools CMD Prompt or PowerShell, the results of each test are not output as the tests run. Instead, the prompt outputs text indicating that XCTests (no tests) have started/finished, then no output occurs for a few seconds while tests run, and when tests complete the results of all tests are printed at the same time.

### Reproduction

1. Checkout swift-foundation
2. Run `swift test`

### Expected behavior

I'd expect that (like on linux/macOS) each test would print its result in near-realtime as the test completes rather than waiting for all tests to complete before printing results

### Environment

```
Swift version 6.2-dev (LLVM 4197ac1672a278c, Swift acbdfef4f4d71b1)
Target: aarch64-unknown-windows-msvc
Build config: +assertions
```

```
→ Testing Library Version: 6.3-dev (e63d542c8248599)
→ Target Platform: aarch64-unknown-windows-msvc
```

grynspan:
Duplicate of a known SPM issue I believe.

dschaefer2:
Definitely a huge sore point right now as we try to debug test failures on Windows in CI.

grynspan:
So what I'd _like_ to do here is rip out the current piping/forwarding/whatevering logic in SwiftPM and just forward stdout/stderr normally. As far as I'm aware, the only time we need to intercept output is when constructing an XML report from XCTest with `--parallel`, and if @plemarquand and I have our way, we won't even need it for that.

PR #9260, opened 2025-10-17
Adopt swift-subprocess when running unit tests

owenv:
Alternative to https://github.com/swiftlang/swift-package-manager/pull/9191

Begin adopting swift-subprocess, beginning with test execution, in an effort to standardize on one implementation for process output streaming across platforms.

This required a bit of additional refactoring of the parallel test runner to make it async

Depends on https://github.com/swiftlang/swift/pull/84947
Fixes: #8928

### DIFF
--- a/Sources/Commands/SwiftTestCommand.swift
+++ b/Sources/Commands/SwiftTestCommand.swift
@@ -11,6 +11,12 @@
 //===----------------------------------------------------------------------===//
 
 import ArgumentParser
+import Subprocess
+#if canImport(System)
+import System
+#else
+import SystemPackage
+#endif
 
 @_spi(SwiftPMInternal)
 import Basics
@@ -342,7 +348,7 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
                         observabilityScope: swiftCommandState.observabilityScope
                     )
 
-                    testResults = try runner.run(tests)
+                    testResults = try await runner.run(tests)
                     result = runner.ranSuccessfully ? .success : .failure
                 }
 
@@ -538,7 +544,7 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
         )
 
         // Finally, run the tests.
-        return runner.test(outputHandler: {
+        return await runner.test(outputHandler: {
             // command's result output goes on stdout
             // ie "swift test" should output to stdout
             print($0, terminator: "")
@@ -837,7 +843,7 @@ extension SwiftTestCommand {
                     )
 
                     // Finally, run the tests.
-                    let result = runner.test(outputHandler: {
+                    let result = await runner.test(outputHandler: {
                         // command's result output goes on stdout
                         // ie "swift test" should output to stdout
                         print($0, terminator: "")
@@ -911,7 +917,7 @@ final class TestRunner {
     // The toolchain to use.
     private let toolchain: UserToolchain
 
-    private let testEnv: Environment
+    private let testEnv: Basics.Environment
 
     /// ObservabilityScope  to emit diagnostics.
     private let observabilityScope: ObservabilityScope
@@ -945,7 +951,7 @@ final class TestRunner {
         additionalArguments: [String],
         cancellator: Cancellator,
         toolchain: UserToolchain,
-        testEnv: Environment,
+        testEnv: Basics.Environment,
         observabilityScope: ObservabilityScope,
         library: TestingLibrary
     ) {
@@ -974,10 +980,10 @@ final class TestRunner {
 
     /// Executes and returns execution status. Prints test output on standard streams if requested
     /// - Returns: Result of spawning and running the test process, and the output stream result
-    func test(outputHandler: @escaping (String) -> Void) -> Result {
+    func test(outputHandler: @escaping (String) -> Void) async -> Result {
         var results = [Result]()
         for path in self.bundlePaths {
-            let testSuccess = self.test(at: path, outputHandler: outputHandler)
+            let testSuccess = await self.test(at: path, outputHandler: outputHandler)
             results.append(testSuccess)
         }
         return results.reduce()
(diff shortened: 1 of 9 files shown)


Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1021,33 +1027,36 @@ final class TestRunner {
...
+                env[.init(rawValue: key.rawValue)!] = value
             }
-            let outputRedirection = AsyncProcess.OutputRedirection.stream(
-                stdout: outputHandler,
-                stderr: outputHandler
-            )
-            let process = AsyncProcess(arguments: try args(forTestAt: path), environment: self.testEnv, outputRedirection: outputRedirection)
-            guard let terminationKey = self.cancellator.register(process) else {
-                return .failure // terminating
+            let processConfig = Subprocess.Configuration(.path(.init(args[0])), arguments: .init(Array(args.dropFirst())), environment: .custom(env))
```
jakepetroules: Can we introduce a convenience extension on Subprocess.Configuration that takes an array as an input? This pattern seems likely to appear elsewhere.

```swift
extension Subprocess.Configuration {
    init(commandLine: [String], environment: Environment) throws {
        guard let arg0 = commandLine.first else {
            throw ItsEmpty()
        }
        self.init(.path(arg0), .init(Array(commandLine.dropFirst())), environment: environment)
    }
}
```

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1021,33 +1027,36 @@ final class TestRunner {
...
+            let result = try await Subprocess.run(processConfig, input: .none, error: .standardOutput) { execution, outputSequence in
+                let token = cancellator.register(name: "Test Execution", handler: { _ in
+                    await execution.teardown(using: [.gracefulShutDown(allowedDurationToNextStep: .seconds(5))])
```
jakepetroules: I think this is unsafe. Because deregister is synchronous, it may return before the cancellation block has finished executing, meaning teardown could be called when the process is already dead. It is NOT safe to use execution once the body block has returned, as the pid may have already been recycled or the pidfd may have become invalid. This is one reason why Execution would probably ideally be ~Escapable...

Subprocess.run already response to Swift Concurrency task cancellation, so you'd want something like this:

```swift
extension Cancellator {
    func run<T: Sendable>(_ block: @escaping @Sendable () async throws -> T) async throws -> T {
        let task = Task { try await block() }
        let token = register(name: "Test Execution", handler: { _ in task.cancel() })
        defer {
            token.map { deregister($0) }
        }
        return try await task.value
    }
}

try await cancellator.run {
    try await Subprocess.run(...)
}
```

That said, do you even need the cancellator, or is cooperative cancellation on its own, enough?

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1190,77 +1168,75 @@ final class ParallelTestRunner {
...
-                    var output = ""
-                    let outputLock = NSLock()
-                    let start = DispatchTime.now()
-                    let result = testRunner.test(outputHandler: { _output in outputLock.withLock{ output += _output }})
-                    let duration = start.distance(to: .now())
-                    if result == .failure {
-                        self.ranSuccessfully = false
+        var pendingTests: [UnitTest] = tests
+        var processedTests: [TestResult] = []
+        observabilityScope.emit(error: "wefhfehjef")
```
jakepetroules: remove debugging code

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1190,77 +1168,75 @@ final class ParallelTestRunner {
...
+                    bundlePaths: [test.productPath],
+                    additionalArguments: additionalArguments,
+                    cancellator: self.cancellator,
+                    toolchain: self.toolchain,
+                    testEnv: testEnv,
+                    observabilityScope: self.observabilityScope,
+                    library: .xctest // swift-testing does not use ParallelTestRunner
+                )
+                var output = ""
+                let outputLock = NSLock()
```
jakepetroules: nit but also kinda seriously: I'd either use an AsyncStream or _at least_ use a LockedValue<T> or similar abstraction.

jakepetroules (changes requested):

Sources/Commands/SwiftTestCommand.swift
owenv: fixed. I think we do need the explicit cancellation here to ensure tests can be interrupted interactively

Sources/Commands/SwiftTestCommand.swift
owenv: I copied this in the refactor, but looking at it more closely this lock seems unnecessary, I deleted it

owenv:
Depends on https://github.com/swiftlang/swift-subprocess/pull/201 for the CMake build

owenv:
> Can we refactor the SwiftTestCommand so we can write 'small' or 'medium' tests, that is, we test the modified functions/code in isolation, as opposed to relying on 'large' (ie: end-to-end tests)?

This PR isn intended to be NFC so I've been relying on the existing tests so far

owenv:
@swift-ci test

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1021,33 +1027,27 @@ final class TestRunner {
...
-                stdout: outputHandler,
-                stderr: outputHandler
-            )
-            let process = AsyncProcess(arguments: try args(forTestAt: path), environment: self.testEnv, outputRedirection: outputRedirection)
-            guard let terminationKey = self.cancellator.register(process) else {
-                return .failure // terminating
+            let args = try args(forTestAt: path)
+            let processConfig = try Subprocess.Configuration(commandLine: args, environment: .init(self.testEnv))
+            let status = try await cancellator.run(name: "Test Execution") {
+                try await Subprocess.run(processConfig, input: .none, error: .standardOutput) { execution, outputSequence in
```
jakepetroules: Shouldn't this be `error: .combineWithOutput`? (see swiftlang/swift-subprocess#180)

owenv: yeah, that's what I was looking for

bkhouri:
> > Can we refactor the SwiftTestCommand so we can write 'small' or 'medium' tests, that is, we test the modified functions/code in isolation, as opposed to relying on 'large' (ie: end-to-end tests)?
> 
> This PR isn intended to be NFC so I've been relying on the existing tests so far

I understand we are relying on existing tests, but we really should use the Test Pyramid to validate the code changes instead of relying on all the tests.  Although the code changes may be exercises via existing tests, they tests to no guarantee we will exercise all the code paths.  As such, we should try to write tests at different layers of the [Test Pyramid](https://martinfowler.com/articles/practical-test-pyramid.html).

owenv:
@swift-ci test

owenv:
@swift-ci test

dschaefer2 (approved):
I assume swift-subprocess has tests for itself.

PR #9261, opened 2025-10-17
[6.2]Fix crash in macros caused by mixing prebuilts and source.

dschaefer2:
There is an issue when a macro that uses prebuilts uses a library that also depends on the swift-syntax library. This is commonly used to share utility code between macros.

However, we currently can't tell whether the library will only be built for the host so it can not take advantage of the prebuilts. When a macro which is using prebuilts uses that library, we are seeing crashes in the macro caused when the library and the swift-syntax it uses is built for debug which is then linked with the macro which uses the swift-syntax prebuilts which are built for release.

For now, this change turns off the prebuilts for macros that use such a library to avoid the crash. We'll investigate making the use of prebuilts conditional on the build settings in a future release.

To be able to implement this change, we need to be able to detect the downstream swift-syntax dependency. We move the insertion of the settings for the prebuilts from Module creation time until after the ResolvedModuleBuilder is created so that we can make the changes there before the final ResolvedModules are created.

### DIFF
--- a/Tests/BuildTests/BuildPlanTests.swift
+++ b/Tests/BuildTests/BuildPlanTests.swift
@@ -4657,16 +4657,48 @@ class BuildPlanTestCase: BuildSystemProviderTestCase {
             emptyFiles: [
                 "/MyPackage/Sources/MyMacroMacros/MyMacroMacros.swift",
                 "/MyPackage/Sources/MyMacros/MyMacros.swift",
-                "/MyPackage/Sources/MyMacroTests/MyMacroTests.swift"
+                "/MyPackage/Sources/MyMacroTests/MyMacroTests.swift",
+                "/swift-syntax/Sources/SwiftSyntaxMacrosTestSupport/SwiftSyntaxMacrosTestSupport.swift",
+                "/swift-syntax/Sources/SwiftSyntaxMacros/SwiftSyntaxMacros.swift",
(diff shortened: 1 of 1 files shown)


dschaefer2:
Addressing https://github.com/swiftlang/swift-package-manager/issues/9196

Sources/PackageGraph/ModulesGraph+Loading.swift
```diff
@@ -19,6 +19,7 @@ import Foundation
 import func TSCBasic.bestMatch
 import func TSCBasic.findCycle
 import struct TSCBasic.KeyedPair
+import func TSCBasic.topologicalSort
```
dschaefer2: In the end I couldn't use this since the ResolvedModuleBuilder.Dependencies aren't Hashable. I've removed this locally. But probably should think a bit more about how the dependencies check is done.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

Sources/PackageGraph/ModulesGraph+Loading.swift
dschaefer2: That's been updated now.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

PR #9264, opened 2025-10-20
Fix crash in macros caused by mixing prebuilts and source. (#9261)

dschaefer2:
There is an issue when a macro that uses prebuilts uses a library that also depends on the swift-syntax library. This is commonly used to share utility code between macros.

However, we currently can't tell whether the library will only be built for the host so it can not take advantage of the prebuilts. When a macro which is using prebuilts uses that library, we are seeing crashes in the macro caused when the library and the swift-syntax it uses is built for debug which is then linked with the macro which uses the swift-syntax prebuilts which are built for release.

For now, this change turns off the prebuilts for macros that use such a library to avoid the crash. We'll investigate making the use of prebuilts conditional on the build settings in a future release.

To be able to implement this change, we need to be able to detect the downstream swift-syntax dependency. We move the insertion of the settings for the prebuilts from Module creation time until after the ResolvedModuleBuilder is created so that we can make the changes there before the final ResolvedModules are created.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/PackageModel/Module/PrebuiltLibrary.swift
+++ b/Sources/PackageModel/Module/PrebuiltLibrary.swift
@@ -1,9 +1,14 @@
+//===----------------------------------------------------------------------===//
 //
-//  PrebuiltLibrary.swift
-//  SwiftPM
+// This source file is part of the Swift open source project
 //
-//  Created by Doug Schaefer on 2025-10-16.
+// Copyright (c) 1994-2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
 //
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
 
 import Basics
 


dschaefer2:
@swift-ci please test

Sources/PackageGraph/ModulesGraph+Loading.swift
```diff
@@ -800,6 +781,91 @@ private func createResolvedPackages(
         }
     }
 
+    // Prebuilts
+    for packageBuilder in packageBuilders {
+        for moduleBuilder in packageBuilder.modules {
```
owenv: Traversing the transitive dependencies of every module separately could get expensive in densely connected package graphs. Could this do a DFS traversal of the graph to mark all the modules ineligible for prebuilts in a single pass?

Sources/PackageModel/Module/PrebuiltLibrary.swift
```diff
@@ -0,0 +1,51 @@
+//
```
jakepetroules, line 2: Needs the standard Apache header

dschaefer2: Done.

dschaefer2:
A better fix is presented in https://github.com/swiftlang/swift-package-manager/pull/9602

PR #9243, opened 2025-10-10
Testing: Run all test SPM executable actions in a operation queue

kcieplak:
Add an operation queue around execution of swiftPM process execution.

**swift-testing** based test will run all tests in parallel at the same time, causing each invocation to spawn a process which will then spawn many other subprocesses.  This is compounded by tests that use arguments, as each combination will create another test that will run in parallel.

These leads to a process bomb consuming all CPU and causes massive I/O blocking.

Add a limit to how many swiftPM top level processes can run at the same time during testing.  This is similar behaviour to XCtest based testing with the **--num-workers** option limiting the number of occurrent tests.

PR #9266, opened 2025-10-21
[6.2] Testing: Run all test SPM executable actions in a operation queue (#9243)

kcieplak:
Cherry pick of b16bb8fcb8675c95dca051a8c88a0a8b73d895ca

Add an operation queue around execution of swiftPM process execution.

    **swift-testing** based test will run all tests in parallel at the same
    time, causing each invocation to spawn a process which will then spawn
    many other subprocesses. This is compounded by tests that use arguments,
    as each combination will create another test that will run in parallel.

    These leads to a process bomb consuming all CPU and causes massive I/O
    blocking.

    Add a limit to how many swiftPM top level processes can run at the same
    time during testing. This is similar behaviour to XCtest based testing
    with the **--num-workers** option limiting the number of occurrent
    tests.

### DIFF
--- a/Sources/_InternalTestSupport/SwiftPMProduct.swift
+++ b/Sources/_InternalTestSupport/SwiftPMProduct.swift
@@ -18,6 +18,9 @@ import struct Basics.AsyncProcessResult
 
 import enum TSCBasic.ProcessEnv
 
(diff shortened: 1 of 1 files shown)


kcieplak:
@swift-ci test

kcieplak:
@swift-ci test windows

kcieplak:
@swift-ci test windows

jakepetroules (approved):

kcieplak:

PR #9267, opened 2025-10-21
Use a switch instead of if/else if

daveinglis:
Cleanup from previous PR


daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

PR #9271, opened 2025-10-22
Use versioned host triples for non-Darwin platforms

jakepetroules:
Only Darwin platforms have the property of defaulting to the manifest-specified version in the target triple. For other platforms, revert to the previous behavior of including the version.

### DIFF
--- a/Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
+++ b/Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
@@ -173,8 +173,13 @@ public struct BuildParameters: Encodable {
         testingParameters: Testing = .init(),
         apiDigesterMode: APIDigesterMode? = nil
     ) throws {
-        // Default to the unversioned triple if none is provided so that we defer to the package's requested deployment target.
-        let triple = try triple ?? .getUnversionedHostTriple(usingSwiftCompiler: toolchain.swiftCompilerPath)
+        // Default to the unversioned triple if none is provided so that we defer to the package's requested deployment target, for Darwin platforms. For other platforms, continue to include the version since those don't have the concept of a package-specified version, and the version is meaningful for some platforms including Android and FreeBSD.
+        let triple = try triple ?? {
+            let hostTriple = try Triple.getHostTriple(
+                    usingSwiftCompiler: toolchain.swiftCompilerPath)
+            return hostTriple.versionedTriple.isDarwin() ? hostTriple.unversionedTriple : hostTriple.versionedTriple
+        }()
+
         self.debuggingParameters = debuggingParameters ?? .init(
             triple: triple,
             shouldEnableDebuggingEntitlement: configuration == .debug,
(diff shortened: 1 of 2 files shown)


jakepetroules:
@swift-ci test

Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
```diff
@@ -174,7 +174,12 @@ public struct BuildParameters: Encodable {
         apiDigesterMode: APIDigesterMode? = nil
     ) throws {
         // Default to the unversioned triple if none is provided so that we defer to the package's requested deployment target.
-        let triple = try triple ?? .getUnversionedHostTriple(usingSwiftCompiler: toolchain.swiftCompilerPath)
+        let triple = try triple ?? {
+            let hostTriple = try Triple.getHostTriple(
+                    usingSwiftCompiler: toolchain.swiftCompilerPath)
+            return hostTriple.versionedTriple.isDarwin() ? hostTriple.unversionedTriple : hostTriple.versionedTriple
```
owenv, line 180: Would be good to leave a comment explaining why we're doing this

owenv (approved):

Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
jakepetroules: Yes, good point.

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test windows

jakepetroules:
@dschaefer2 Not really, this is more so related to the API surface we vend in Package.swift, nothing the compiler is or should be aware of.

jakepetroules:
@swift-ci test windows

jakepetroules:
@swift-ci test windows

jakepetroules:
@swift-ci test windows

jakepetroules:
@swift-ci test windows

3405691582:
I think this may still be insufficient for OpenBSD, possibly because FreeBSD and other platforms don't use versioned imports, but OpenBSD still does. Just `Sources/PackageModel/UserToolchain.swift` needs to change. I'll send a pr when I can, or feel free to beat me to it.

dschaefer2:
At some point soon, I want to go through all the hardcoding of platform specifics and see if we can do this with plugins associated with a formalization of Platform in the target graph model so individual platforms can specify what behaviour they need.

etcwilde:
> I think this may still be insufficient for OpenBSD, possibly because FreeBSD and other platforms don't use versioned imports, but OpenBSD still does. Just Sources/PackageModel/UserToolchain.swift needs to change. I'll send a pr when I can, or feel free to beat me to it.

We're still seeing the issue bootstrapping on FreeBSD as well: `error: compiling for freebsd 0, but module 'Foundation' has a minimum deployment target of freebsd 14.3`

PR #9275, opened 2025-10-23
Add detailed logging on AsyncProcess to help diagnose hangs

bc-lee:
### Motivation:

See https://github.com/swiftlang/swift-package-manager/issues/9274 for context.
The existing logging is insufficient to diagnose hangs in AsyncProcess, so adding more detailed logging around process start, output, and termination should help identify where things are going wrong.

### Modifications:

- Added `SWIFTPM_DEBUG_PROCESS_IO` environment variable to enable low-level I/O thread debugging without interfering with normal observability output
- Added `processTrackingID` (8-character UUID prefix) to uniquely identify each process instance in logs for correlation across multiple concurrent processes
- Added `logProcessIODebug()` helper that writes directly to stderr when debug mode is enabled, bypassing all handlers to capture issues even when the observability system has problems
- Instrumented all critical synchronization points in the I/O thread lifecycle

### Result:

After this change, AsyncProcess will log detailed information about process lifecycle events, including when processes are started, when output is received, and when processes terminate. This should provide better visibility into where hangs may be occurring during dependency resolution.

Users experiencing hangs can now run:
```
SWIFTPM_DEBUG_PROCESS_IO=1 swift package resolve 2>&1 | tee debug.log
```

### DIFF
--- a/Sources/Basics/Concurrency/AsyncProcess.swift
+++ b/Sources/Basics/Concurrency/AsyncProcess.swift
@@ -159,6 +159,22 @@ extension DispatchQueue {
     )
 }
 
+// MARK: - Process Logging
+
+/// Enable detailed process I/O thread debugging by setting SWIFTPM_DEBUG_PROCESS_IO=1
+/// This is a low-level debugging flag for diagnosing process hangs and defunct processes.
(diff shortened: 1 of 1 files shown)


bripeticca:
@swift-ci please test

PR #9295, opened 2025-10-27
Align PIF generation for user-authored modulemaps with native build system

owenv:
Swift Build was relying on hand-authored module maps being discovered in include directories, while the old build system always passed them using -fmodule-map-file as well. This isn't ideal, but packages like Foundation rely on this to avoid clashes with modules that ship in the toolchain, so for now we align the behavior for compatibility with existing code.

Closes https://github.com/swiftlang/swift-package-manager/issues/9290


owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

dschaefer2:
Nice, this looks like it will help the plug-in tool generated module map as well (once we discover its path properly in TargetSourcesBuider). If someone wants to generate an apinote, they will need to make sure the module map is right next to it.

dschaefer2 (approved):

owenv:
@swift-ci test Windows

PR #9296, opened 2025-10-28
Detect system GNU ld script pretending to be system libraries

daniel-grumberg:
When a system library is shimmed with a GNU ld script, pull in the required libraries when auditing binary artifacts

### Motivation:

Basic C runtime libraries appear to sometimes be replaced by a GNU ld script that pulls it some other libraries as a backwards compatibility mechanism on certain Linux distros. This makes it difficult for the binary auditing tool to find all required libraries. The only previously known case of this was libgcc_s so we used to just special case for it, but now that Ubuntu does this for splitting up libm, a more general solution is required.

### Modifications:

Add basic support for detecting such GNU ld script and parsing out libraries they pull in. The subset of features supported is consistent with how we have seen libraries being repackaged.

### Result:

Binary auditing tool will be more resilient.

### DIFF
--- a/Sources/BinarySymbols/ClangHostDefaultObjectsDetector.swift
+++ b/Sources/BinarySymbols/ClangHostDefaultObjectsDetector.swift
@@ -14,7 +14,8 @@ import Foundation
 import protocol TSCBasic.WritableByteStream
 
 package func detectDefaultObjects(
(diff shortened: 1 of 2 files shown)


daniel-grumberg:
@swift-ci please test

PR #9300, opened 2025-10-28
Resolve traits on top level manifests explicitly

plemarquand:
If a package is contained in a wrapper project it will have a top level manifest with no root packages. In this case, we weren't resolving default traits on the dependencies within this top level manifest.


plemarquand:
@swift-ci please test

bripeticca:
I think there are some traits-related test failures from this change - I'll look into it since I believe SwiftPM is making some assumptions about the `PackageGraphRootInput` in those tests as well and I'd like to make sure we can account for it in the future 😄

plemarquand:
@swift-ci test

bripeticca:
@swift-ci please test windows

PR #9301, opened 2025-10-28
Fix artifactPath used for binary targets in PIF

owenv:
The path of a binary module is always the root, whereas the artifactPath points to the artifact bundle / xcframework. Ensure we use the latter when generating file references in PIF so that we can link binary artifacts properly.


owenv:
@swift-ci test

owenv:
Closes https://github.com/orgs/swiftlang/projects/26/views/1?layout=board&pane=issue&itemId=135817219&issue=swiftlang%7Cswift-package-manager%7C9103

PR #9303, opened 2025-10-29
Disable ONLY_ACTIVE_ARCH when --arch is used to request more than one explicitly

owenv:
This ensures that even in a debug build, we build a universal binary on macOS if it's requested.

Closes https://github.com/swiftlang/swift-package-manager/issues/8982 by resolving the remaining issue building swiftly with the Swift Build backend

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -842,7 +842,7 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
             sdkVariant: sdkVariant,
             targetArchitecture: buildParameters.triple.archName,
             supportedArchitectures: [],
-            disableOnlyActiveArch: false
+            disableOnlyActiveArch: (buildParameters.architectures?.count ?? 1) > 1
         )
     }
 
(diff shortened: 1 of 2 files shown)


owenv:
@swift-ci test

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -842,7 +842,7 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
             sdkVariant: sdkVariant,
             targetArchitecture: buildParameters.triple.archName,
             supportedArchitectures: [],
-            disableOnlyActiveArch: false
+            disableOnlyActiveArch: (buildParameters.architectures?.count ?? 1) > 1
```
jakepetroules, line 845: Should we just always disable this?

jakepetroules (approved):

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
owenv: I think if you do a debug build and don't explicitly specify any target/arch we probably shouldn't build universal

PR #9308, opened 2025-10-30
OpenBSD uses versioned triples for imports.

3405691582:
Without this change, the build will fail to find modules since the import path on OpenBSD currently uses versioned triples.

### Motivation:

After #9271, bootstrap failures still occurred on OpenBSD.

### Modifications:

Used the versioned triple as the default, if the unversioned triple was OpenBSD.

### Result:

swiftpm bootstraps properly on OpenBSD.

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -741,7 +741,10 @@ public final class UserToolchain: Toolchain {
             // targetInfo from the compiler
             let targetInfo = try customTargetInfo ?? Self.getTargetInfo(swiftCompiler: swiftCompilers.compile)
             self._targetInfo = targetInfo
-            triple = try swiftSDK.targetTriple ?? Self.getHostTriple(targetInfo: targetInfo, versioned: false)
+            triple = try Self.getHostTriple(targetInfo: targetInfo, versioned: false)
+            if !triple.isDarwin() {
+                triple = try Self.getHostTriple(targetInfo: targetInfo, versioned: true)
+            }
         }
 
         // Change the triple to the specified arch if there's exactly one of them.


3405691582:
@swift-ci please test.

Sources/PackageModel/UserToolchain.swift
```diff
@@ -741,7 +741,11 @@ public final class UserToolchain: Toolchain {
             // targetInfo from the compiler
             let targetInfo = try customTargetInfo ?? Self.getTargetInfo(swiftCompiler: swiftCompilers.compile)
             self._targetInfo = targetInfo
-            triple = try swiftSDK.targetTriple ?? Self.getHostTriple(targetInfo: targetInfo, versioned: false)
+            var defaultTriple = try Self.getHostTriple(targetInfo: targetInfo, versioned: false)
+            if defaultTriple.isOpenBSD() {
```
owenv: This triple was always versioned before my PR to fix the deployment target issues on Darwin, so we may want to just make this an is-non-darwin check to preserve behavior, wdyt?

Sources/PackageModel/UserToolchain.swift
```diff
@@ -741,7 +741,11 @@ public final class UserToolchain: Toolchain {
...
+                defaultTriple = try Self.getHostTriple(targetInfo: targetInfo, versioned: true)
+            }
+            triple = swiftSDK.targetTriple ?? defaultTriple
```
owenv: This was already the case before your patch, but swiftSDK.targetTriple should always be nil if we're in this else branch

etcwilde:
FreeBSD also hits this in CI and I suspect once version numbers are enabled, Android will too.
I am currently successfully working with this diff on my local FreeBSD at the moment:
```
diff --git a/Sources/PackageModel/UserToolchain.swift b/Sources/PackageModel/UserToolchain.swift
index 86149e2d9..9a20812bb 100644
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -741,7 +741,10 @@ public final class UserToolchain: Toolchain {
             // targetInfo from the compiler
             let targetInfo = try customTargetInfo ?? Self.getTargetInfo(swiftCompiler: swiftCompilers.compile)
             self._targetInfo = targetInfo
-            triple = try swiftSDK.targetTriple ?? Self.getHostTriple(targetInfo: targetInfo, versioned: false)
+            triple = try Self.getHostTriple(targetInfo: targetInfo, versioned: true)
+            if triple.isDarwin() {
+                triple = try Self.getHostTriple(targetInfo: targetInfo, versioned: false)
+            }
         }
```

3405691582:
Nice, that's basically what I have with the suggested review comments with the reversed sense.

Sources/PackageModel/UserToolchain.swift
3405691582: Done.

Sources/PackageModel/UserToolchain.swift
3405691582: Makes sense; I've elided this.

3405691582:
@swift-ci please test.

jakepetroules (approved):
Thanks!

jakepetroules:
@swift-ci test windows

Sources/PackageModel/UserToolchain.swift
```diff
@@ -741,7 +741,10 @@ public final class UserToolchain: Toolchain {
             // targetInfo from the compiler
             let targetInfo = try customTargetInfo ?? Self.getTargetInfo(swiftCompiler: swiftCompilers.compile)
             self._targetInfo = targetInfo
-            triple = try swiftSDK.targetTriple ?? Self.getHostTriple(targetInfo: targetInfo, versioned: false)
```
3405691582, line 744: If I understand correctly, `swiftSDK.targetTriple` is nil in this branch (see l737), so there should be no intended behavioral change.

jakepetroules: Looks like you're right, thanks for following up on that!

PR #9309, opened 2025-10-30
Add some extra conditionals for swift-build on OpenBSD.

3405691582:
Despite the toolchain path being specified swift-build is still not initializing a toolchain properly. I have my suspicions, but I believe they can be addressed on the swift-build side, and this change on the swiftpm side is nonetheless correct.

### Motivation:

When attempting to use swift-build, errors are raised.

### Modifications:

Provide the correct platform conditionals to suppress these errors.

### Result:

swift-build is not properly working, but I believe these changes will still be necessary.

### DIFF
--- a/Sources/Basics/Process.swift
+++ b/Sources/Basics/Process.swift
@@ -16,6 +16,7 @@ public enum OperatingSystem: Hashable, Sendable {
     case linux
     case android
     case freebsd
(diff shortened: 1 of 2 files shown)


3405691582:
@swift-ci please test.

owenv:
thanks!

PR #9317, opened 2025-11-02
Feature/cmake consumer

zamderax:
Added cmake support 

### Motivation:

When creating Swift Package for consumption, you can add a git submodule written in C or C++ for example: https://github.com/libsdl-org/SDL . A downstream app that declares your Swift package will clone the repo of the package _and recursively get the submodules which is great!_. However, if the C/C++ project uses Cmake to build, then SwiftPM won't understand how to link everything.

### Modifications:

* Added a Build Plan to detect a Cmake project, invoked Cmake builds, and allowed the project to turn into a CLang Module after building. 
* Made it into `.systemLibrary` that uses the Clang module and which leverages 
* Added experimental support for Swift SDKs with triples 

### Result:

Best read: https://github.com/zamderax/swift-package-manager/blob/feature/cmake-consumer/Documentation/CMakeIntegration.md#cross-compilation-and-swift-sdks

### DIFF
--- a/Documentation/CMakeIntegration.md
+++ b/Documentation/CMakeIntegration.md
@@ -9,6 +9,9 @@ This guide shows you how to create a Swift Package that wraps a CMake-based C/C+
 * [Configuration Reference](#configuration-reference)
 * [Module Map Modes](#module-map-modes)
 * [Advanced Topics](#advanced-topics)
+  * [Cross-Compilation and Swift SDKs](#cross-compilation-and-swift-sdks)
+  * [Textual Headers](#textual-headers)
+  * [Build Caching](#build-caching)
 * [Troubleshooting](#troubleshooting)
 * [Examples](#examples)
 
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@zamderax I have no opinion on this PR specifically, but in general before creating a PR to a release branch, you have to get one merged into `main` first.

dschaefer2:
I'm actually working on a vision document that will cover integrating external build systems with SwiftPM that will be more encompassing and introduce new package manifest syntax and build plugins to handle them. You should be able to add an integration like this without having SwiftPM needing to know the details. I'll be sharing that and looking for feedback in the next few days.

Also we're about to deprecate the old native build system, i.e. BuildPlan, and switch to swiftbuild. This integration will need to be handled quite differently.

rconnell9:
@zamderax , can you target this change to the _main_ branch?  As Max mentioned we generally don't allow changes to land in a release branch prior to main.

PR #9335, opened 2025-11-05
Command availability annotations

heckj:
Adds some availability detail to the command line tools recently added to the core docs

### Motivation:

Saw comments about swift package migrate and swift package show-traits being in the docs, but not available in released toolchains.

### Modifications:

Added availability annotations to the documents

### Result:

CLI documentation pages will show availability of when commands are available for versions of the Swift toolchain


heckj:
@swift-ci please test

PR #9354, opened 2025-11-10
Ensure resource bundles aren't marked as executable

plemarquand:
This patch reverts a single line from
https://github.com/swiftlang/swift-package-manager/commit/a112ab817dafb40f3d95daaa31b034f7c766fa8e#diff-b1845cad8f64e37500b06818ff71629726c14cf5a8d34add6ef3a9d38236cfe5

The executable name should be deliberately omitted from the build settings to avoid having the bundle be marked as executable with the CFBundleExecutable in its Info.plist.

### DIFF
--- a/Fixtures/Miscellaneous/TIF/.gitignore
+++ b/Fixtures/Miscellaneous/TIF/.gitignore
@@ -1,5 +1,0 @@
-.DS_Store
-/.build
-/Packages
-/*.xcodeproj
-xcuserdata/
--- a/Sources/SwiftBuildSupport/PackagePIFProjectBuilder.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFProjectBuilder.swift
@@ -211,6 +211,8 @@ struct PackagePIFProjectBuilder {
         settings[.PRODUCT_MODULE_NAME] = bundleName
         settings[.PRODUCT_BUNDLE_IDENTIFIER] = "\(self.package.identity).\(module.name).resources"
             .spm_mangledToBundleIdentifier()
+        // Resource bundles are not executable. Setting the name to an empty string will
+        // omit the CFBundleExecutable key from the Info.plist.
         settings[.EXECUTABLE_NAME] = ""
         settings[.GENERATE_INFOPLIST_FILE] = "YES"
         settings[.PACKAGE_RESOURCE_TARGET_KIND] = "resource"


plemarquand:
@swift-ci please test

Fixtures/Miscellaneous/TIF/.gitignore
```diff
@@ -0,0 +1,5 @@
+.DS_Store
```
owenv: nit: fixtures can usually drop the gitignore

Sources/SwiftBuildSupport/PackagePIFProjectBuilder.swift
```diff
@@ -211,6 +211,7 @@ struct PackagePIFProjectBuilder {
         settings[.PRODUCT_MODULE_NAME] = bundleName
         settings[.PRODUCT_BUNDLE_IDENTIFIER] = "\(self.package.identity).\(module.name).resources"
             .spm_mangledToBundleIdentifier()
+        settings[.EXECUTABLE_NAME] = ""
```
owenv, line 216: Can you add a comment describing why this needs to be set empty?

plemarquand: Good call, done

plemarquand:
@swift-ci please test

plemarquand:
@swift-ci test windows

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1760,6 +1760,45 @@ struct BuildCommandTestCases {
...
+            let result = try await build(
+                [],
+                packagePath: fixturePath,
+                configuration: data.config,
+                cleanAfterward: false,
+                buildSystem: buildSystem,
+            )
+            try #require(result.binContents.contains("TIF_TIF.bundle"))
+            let contentsDir = result.binPath.appending("TIF_TIF.bundle", "Contents")
+            let bundleResourceDir = contentsDir.appending("Resources")
```
bkhouri, line 1786: Question: is there value is validating this path exists so we can have "better" message in the event it does not?

Fixtures/Miscellaneous/TIF/Package.swift
bkhouri, line 1: Suggest: can we rename "TIF" with something more descriptive?

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1760,6 +1760,45 @@ struct BuildCommandTestCases {
...
+
+            let bundleResources = try localFileSystem.getDirectoryContents(bundleResourceDir)
+            #expect(bundleResources.contains("some.txt"))
+            #expect(bundleResources.contains("Assets.car"))
+            #expect(bundleResources.contains("SomeAlert.nib"))
+
+            // Check that the Info.plist of the resource bundle looks reasonable.  In particular, it shouldn't have a CFBundleExecutable key, since it's a codeless bundle.
+            let infoPlistPath = contentsDir.appending("Info.plist")
```
bkhouri, line 1794: Suggestion: SwiftPM has no control over the contents of the info.plist file. As a result, changes to the SwiftBuild service can impact this test. Can we move the intent of this test to SwiftBuild?

plemarquand: SwiftPM does have control depending on what build settings it passes down to swift-build.

This PR is not test only, it reverts a line of code that changed the structure of the Info.plist in resource bundles.

plemarquand:
@swift-ci test windows

Sources/SwiftBuildSupport/PackagePIFProjectBuilder.swift
```diff
@@ -211,6 +211,9 @@ struct PackagePIFProjectBuilder {
         settings[.PRODUCT_MODULE_NAME] = bundleName
         settings[.PRODUCT_BUNDLE_IDENTIFIER] = "\(self.package.identity).\(module.name).resources"
             .spm_mangledToBundleIdentifier()
+        // Resource bundles are not executable. Setting the name to an empty string will
```
jakepetroules, line 214: CFBundleExecutable doesn't mean the binary is executable per se. It's present if the bundle _has_ an executable (which could be a dylib).

Was this found to cause an actual problem somewhere, or just speculative?

jakepetroules: This was mistakenly removed in https://github.com/swiftlang/swift-package-manager/commit/a112ab817dafb40f3d95daaa31b034f7c766fa8e#diff-203ff12a972b403e46d402150fce4826027b92cc374a006e6108bb523c91cb6dL214

PR #9356, opened 2025-11-10
Make lazy properties in UserToolchain thread safe

plemarquand:
`runtimeLibraryPaths`, `swiftCompilerVersion` and `targetInfo` were all marked as lazy on the UserToolchain class which could lead to race conditions if callers on multiple threads attempted to initialize them simultaneously.

Use a `ThreadSafeBox` to guard access while initialization occurs, preventing crashes.

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -42,29 +42,17 @@ public final class UserToolchain: Toolchain {
     /// An array of paths to search for libraries at link time.
     public let librarySearchPaths: [AbsolutePath]
 
-    /// Lock for thread-safe access to lazy properties (recursive to handle property dependencies)
-    private let lazyPropertiesLock = NSRecursiveLock()
-
-    /// Cached runtime library paths
-    private var _runtimeLibraryPaths: [AbsolutePath]? = nil
+    /// Thread-safe cached runtime library paths
+    private let _runtimeLibraryPaths = ThreadSafeBox<[AbsolutePath]>()
 
     /// An array of paths to use with binaries produced by this toolchain at run time.
     public var runtimeLibraryPaths: [AbsolutePath] {
-        lazyPropertiesLock.lock()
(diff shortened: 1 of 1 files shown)


plemarquand:
@swift-ci please test

Sources/PackageModel/UserToolchain.swift
```diff
@@ -42,11 +42,30 @@ public final class UserToolchain: Toolchain {
     /// An array of paths to search for libraries at link time.
     public let librarySearchPaths: [AbsolutePath]
 
+    /// Lock for thread-safe access to lazy properties (recursive to handle property dependencies)
+    private let lazyPropertiesLock = NSRecursiveLock()
```
jakepetroules: Directly using NSLock/NSRecursiveLock is not the recommended pattern in modern Swift, instead you'll want to use an abstraction like Swift Build's LockedValue. In SwiftPM you can use ThreadSafeBox from TSCBasic.

If you want to use a single lock for multiple values to avoid a proliferation of fine-grained locks, you can define a struct to hold the values, and then guard access to the overall struct through the wrapper, rather than using a separate lock for each value.

Later, once we can increase our deployment target to macOS 15, we can replace LockedValue/ThreadSafeBox with [Mutex](https://developer.apple.com/documentation/synchronization/mutex) from the Swift Standard Library.

plemarquand:
@swift-ci please test

Sources/PackageModel/UserToolchain.swift
plemarquand: Thanks, I figured these could be replaced with Mutex once we're ready to adopt but didn't know about ThreadSafeBox's used in SwiftPM. I've updated the implementation accordingly.

jakepetroules (approved):

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci please test linux

PR #9358, opened 2025-11-11
Warn when SWIFT_EXEC or SWIFT_EXEC_MANIFEST point to invalid paths.

rjmansfield:
Previously when SWIFT_EXEC or SWIFT_EXEC_MANIFEST were set to non-existent or non-executable paths, swiftpm would silently ignore them and fall back to the default Swift compiler. This caused confusion during development when typos or incorrect paths were used.

### DIFF
--- a/Sources/CoreCommands/SwiftCommandState.swift
+++ b/Sources/CoreCommands/SwiftCommandState.swift
@@ -1041,6 +1041,7 @@ public final class SwiftCommandState {
                 swiftSDK: swiftSDK,
                 environment: self.environment,
                 customTargetInfo: targetInfo,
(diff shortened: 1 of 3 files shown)


rjmansfield:
@swift-ci test

jakepetroules (changes requested):
We want to avoid writing to standard output streams in library code. Can you plumb an ObservabilityScope down to this method instead?

rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

jakepetroules (approved):

jakepetroules:
@swift-ci test

PR #9362, opened 2025-11-11
Capture code snippet from diagnostic compiler output

bripeticca:
This PR refactors diagnostic handling in the Swift build system by introducing a dedicated message handler and per-task output buffering to properly parse and emit compiler diagnostics individually.

  ### **Key Changes**

  **SwiftBuildSystemMessageHandler**
  - Introduced a new dedicated handler class to process `SwiftBuildMessage` events from the build operation
  - Moved message handling logic out of inline nested functions for better organization and testability
  - Maintains build state, progress animation, and diagnostic processing in a single cohesive component

  **Per-Task Data Buffering**
  - Added `taskDataBuffer` struct in `BuildState` to capture compiler output per task signature
  - New `TaskDataBuffer` struct allows for using `LocationContext` or `LocationContext2` as a subscript key to fetch the appropriate data buffer for a task, defaulting to the global buffer if no associated task or target can be identified.
  - Task output is accumulated in the buffer as `.output` messages arrive
  - Buffer contents are processed when tasks complete, ensuring all output is captured before parsing
  - Failed tasks with no useful or apparent message will be demoted to an info log level to avoid creating too much noise on the output.
  
  **Per-Task Diagnostic Buffering**
  - Added `diagnosticBuffer` property to the `BuildState` to track diagnostics to emit once we receive a `taskComplete` event
  - A check is done to ascertain whether the diagnostic info we receive is a global/target diagnostic, and if so we emit the diagnostic immediately; all other diagnostics are accumulated in the buffer to be emitted once the associated task is completed.
  
  **EmittedTasks**
  - Helper struct for the message handler to track which task's messages have already been emitted
  - Handles both taskIDs as well as taskSignatures
  
  ### **Test Suite**
  
  **SwiftBuildSystemMessageHandlerTests**
  - New test suite created to assert that the diagnostic output is formatted and emitted as expected.
  - Uses the initializers for the nested `SwiftBuildMessage` info structs that are exposed for testing purposes only

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystemMessageHandler.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystemMessageHandler.swift
@@ -24,6 +24,7 @@ import protocol TSCBasic.OutputByteStream
 /// Handler for SwiftBuildMessage events sent by the SWBBuildOperation.
 public final class SwiftBuildSystemMessageHandler {
     private let observabilityScope: ObservabilityScope
+    private let outputStream: OutputByteStream
     private let logLevel: Basics.Diagnostic.Severity
     private var buildState: BuildState = .init()
     private let enableBacktraces: Bool
@@ -59,6 +60,7 @@ public final class SwiftBuildSystemMessageHandler {
     )
     {
         self.observabilityScope = observabilityScope
+        self.outputStream = outputStream
         self.logLevel = logLevel
         self.progressAnimation = ProgressAnimation.ninja(
             stream: outputStream,
@@ -119,7 +121,12 @@ public final class SwiftBuildSystemMessageHandler {
     ) throws {
         // Begin by emitting the text received by the task started event.
         if let started = self.buildState.startedInfo(for: startedInfo) {
-            observabilityScope.emit(info: started)
+            // Determine where to emit depending on the verbosity level.
+            if self.logLevel.isVerbose {
+                self.outputStream.send(started.description + "\n")
+            } else {
+                observabilityScope.emit(info: started)
+            }
         }
 
         guard info.result == .success else {
@@ -312,10 +319,10 @@ extension SwiftBuildSystemMessageHandler {
 
             // Track relevant task info to emit to user.
             let output = if let cmdLineDisplayStr = task.commandLineDisplayString {
-                    cmdLineDisplayStr
-                } else {
-                    task.executionDescription
-                }
+                "\(task.executionDescription)\n\(cmdLineDisplayStr)"
+            } else {
+                task.executionDescription
+            }
             taskDataBuffer[task] = output
         }
 
--- a/Tests/CommandsTests/PackageCommandTests.swift
+++ b/Tests/CommandsTests/PackageCommandTests.swift
@@ -5768,6 +5768,7 @@ struct PackageCommandTests {
                     let filteredStderr = stderr.components(separatedBy: "\n")
                         .filter { !$0.contains("Unable to locate libSwiftScan") }
                         .filter { !($0.contains("warning: ") && $0.contains("unable to find libclang")) }
+                        .filter { !$0.contains("Build description")}
                         .joined(separator: "\n")
                     #expect(filteredStderr == isEmpty)
                 }
@@ -5786,6 +5787,7 @@ struct PackageCommandTests {
                     let filteredStderr = stderr.components(separatedBy: "\n")
                         .filter { !$0.contains("Unable to locate libSwiftScan") }
                         .filter { !($0.contains("warning: ") && $0.contains("unable to find libclang")) }
+                        .filter { !$0.contains("Build description")}
                         .joined(separator: "\n")
                     #expect(filteredStderr == isEmpty)
                 }
--- a/Tests/FunctionalTests/PluginTests.swift
+++ b/Tests/FunctionalTests/PluginTests.swift
@@ -1717,7 +1717,7 @@ struct PluginTests {
                     #expect(stdout.contains("Build of product 'MyLocalTool' complete!"), "stdout:\n\(stdout)")
                 case .swiftbuild:
                     #expect(stdout.contains("MySourceGenBuildTool-product"), "stdout:\n\(stdout)\nstderr:\n\(stderr)")
-                    #expect(stderr.contains("Creating foo.swift from foo.dat"), "stdout:\n\(stdout)\nstderr:\n\(stderr)")
+                    #expect(stdout.contains("Creating foo.swift from foo.dat"), "stdout:\n\(stdout)\nstderr:\n\(stderr)")
                     #expect(stdout.contains("Build complete!"), "stdout:\n\(stdout)\nstderr:\n\(stderr)")
                 case .xcode:
                     Issue.record("Test expected have not been considered")


Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +184,174 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+                "\(progressInfo.message)"
+            }
+            progressAnimation.update(step: step, total: 100, text: message)
+            self.delegate?.buildSystem(self.buildSystem, didUpdateTaskProgress: message)
+        case .diagnostic(let info):
+            if info.appendToOutputStream {
+                emitInfoAsDiagnostic(info: info)
+            }
+        case .output(let info):
+            let parsedOutput = String(decoding: info.data, as: UTF8.self)
```
owenv: One consideration here: I don't think output messages are guaranteed to be split at character boundaries. So for example, we could emit two output messages, with the last byte of one message being the first byte of a 2-byte character and the second message having the second byte of that character. And then this String(decoding:) call might mangle the character.

I think it's also possible we receive something like `(output message) -> (progress update message) -> (output message)` which could split the output at arbitrary points.

Maybe when we receive an output message, we should append to a per-task buffer, then only decode it as a string and emit the output when the taskComplete message arrives?

bripeticca: Good point, I've found this to be the case with some light local testing as well -- I'll give the per-task buffer a shot!

bripeticca:
@swift-ci please test

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,388 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+        }
+
+        mutating func started(target: SwiftBuild.SwiftBuildMessage.TargetStartedInfo) throws {
+            if targetsByID[target.targetID] != nil {
+                throw Diagnostics.fatalError
+            }
+            targetsByID[target.targetID] = target
+        }
+
+        mutating func target(for task: SwiftBuild.SwiftBuildMessage.TaskStartedInfo) throws -> SwiftBuild.SwiftBuildMessage.TargetStartedInfo? {
```
owenv: Does this need to be mutating?

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,388 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+            // Grab the taskID to append to buffer-per-task storage
+            guard let taskSignature = info.locationContext2.taskSignature else {
+                // If we cannot find the task signature from the locationContext2,
+                // use deprecated locationContext instead.
+                if let taskID = info.locationContext.taskID,
+                    let taskSignature = buildState.taskSignature(for: taskID) {
+                    buildState.appendToBuffer(taskSignature, data: info.data)
+                }
+
+                return
```
owenv: If we're unable to get a taskID, this may be target/global output, so I think we should print it the best we can instead of returning and silently dropping it

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,388 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+            }
+            let targetInfo = try buildState.target(for: info)
+            buildSystem.delegate?.buildSystem(buildSystem, willStartCommand: BuildSystemCommand(info, targetInfo: targetInfo))
+            buildSystem.delegate?.buildSystem(buildSystem, didStartCommand: BuildSystemCommand(info, targetInfo: targetInfo))
+        case .taskComplete(let info):
+            let startedInfo = try buildState.completed(task: info)
+
+            // If we've captured the compiler output with formatted diagnostics, emit them.
+//            if let buffer = buildState.dataBuffer(for: startedInfo) {
+            emitDiagnosticCompilerOutput(startedInfo)
```
owenv: This could be something other than a compile task which emits output with different formatting than we expect

owenv (commented):
Thie generally lgtm but I have some concerns about the regex-based parsing when we emit the textual compiler output.

1. Perf - It's important this is fast so that it doesn't block the end of the build if a command produces huge quantities of output. It's hard to say if this will be a real issue without some testing
2. We're re-parsing information which we're already getting from the compiler in structured form. I see the appeal of not reporting a diagnostic twice if multiple compile jobs report it though

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,388 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+            case "note": return .info
+            case "remark": return .debug
+            default: return .info
+            }
+        }
+    }
+
+    /// Split compiler output into individual diagnostic segments
+    /// Format: /path/to/file.swift:line:column: severity: message
+    private func splitIntoDiagnostics(_ output: String) -> [ParsedDiagnostic] {
```
jakepetroules: I don't think we should be doing _any_ output parsing at this level, Swift Build already has an entire subsystem dedicated to doing this.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,388 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+        self.progressAnimation = ProgressAnimation.ninja(
+            stream: self.outputStream,
+            verbose: self.logLevel.isVerbose
+        )
+    }
+
+    struct BuildState {
+        private var targetsByID: [Int: SwiftBuild.SwiftBuildMessage.TargetStartedInfo] = [:]
+        private var activeTasks: [Int: SwiftBuild.SwiftBuildMessage.TaskStartedInfo] = [:]
+        private var taskBuffer: [String: Data] = [:]
```
jakepetroules: Instead of a string as the key, can you use a `SwiftBuildMessage.LocationContext`?

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: Also suggested using a locationContext as the key, which might help. 

It might be about time to clean up these deprecated APIs, too. I recall there was something missing from the new ones which blocked total removal.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,388 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+//            }
+
+            if info.result != .success {
+                self.observabilityScope.emit(severity: .error, message: "\(startedInfo.ruleInfo) failed with a nonzero exit code. Command line: \(startedInfo.commandLineDisplayString ?? "<no command line>")")
+            }
+            let targetInfo = try buildState.target(for: startedInfo)
+            buildSystem.delegate?.buildSystem(buildSystem, didFinishCommand: BuildSystemCommand(startedInfo, targetInfo: targetInfo))
+            if let targetName = targetInfo?.targetName {
+                serializedDiagnosticPathsByTargetName[targetName, default: []].append(contentsOf: startedInfo.serializedDiagnosticsPaths.compactMap {
+                    try? Basics.AbsolutePath(validating: $0.pathString)
```
jakepetroules: Just let this throw, if this isn't an absolute path something has gone very wrong.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,388 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+        case .taskStarted(let info):
+            try buildState.started(task: info)
+
+            if let commandLineDisplay = info.commandLineDisplayString {
+                self.observabilityScope.emit(info: "\(info.executionDescription)\n\(commandLineDisplay)")
+            } else {
+                self.observabilityScope.emit(info: "\(info.executionDescription)")
+            }
+
+            if self.logLevel.isVerbose {
```
jakepetroules: We should not write any output on task started events and instead always defer its emission to task completed events. The reason is that multiple tasks can start in parallel, and the outputStream content emitted when the taskStarted and taskEnded events are emitted may not be reasonably ordered at all. so you could have:

- task 1 header
- task 2 output
- task 1 header
- task 2 output

...which will be really confusing since task 2's output will look like it belongs to task 1, and so on.

The observabilityScope stuff might be OK, provided that doesn't _also_ end up in the textual output stream and is only for API clients.

jakepetroules:
> 2\. We're re-parsing information which we're already getting from the compiler in structured form. I see the appeal of not reporting a diagnostic twice if multiple compile jobs report it though

I also pointed this out inline, though I'm not sure why the re-parsing relates to deduplication? We should be able to deduplicate diagnostics whether we parse them again or use the existing ones.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: @jakepetroules  I would agree, however the events we're receiving from SwiftBuild accumulates **every** diagnostic message into a singular data buffer -- when emitting this as-is, it's possible that we'd be coupling some info-level diagnostics with error-level diagnostics with no way to separate the severity. Splitting the string on a per-diagnostic basis is the only way to achieve this in this way.

The observability scope that we use to emit these diagnostics will capture the entire string blob, and we have to decide with what severity to emit the entire message. I'm not sure if there's an alternative path here that would give us the same ergonomics on the user front, but IMO it's preferred to separate each diagnostic and ensure that we're emitting them with the appropriate severity, rather than sending the entire string of possibly many diagnostics with varying severities.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: It does not, will modify this

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: Definitely. There are some guards in `emitDiagnosticCompilerOutput` that checks whether there is an existing data buffer that we've logged from the received `SwiftBuildMessage.OutputInfo` case so if this is anything else it will not be emitted.

Is it preferred to have this check be more explicit re: asserting this before calling this method?

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: Thanks for the detailed explanation! I'm in agreement -- this is some legacy implementation that I moved from the inline method to this class, so it's possible that this kind of behaviour is sprinkled elsewhere in `emitEvent`. I'll keep an eye on that :)

bripeticca:
> though I'm not sure why the re-parsing relates to deduplication

(@jakepetroules @owenv -- tagging for visibility, GitHub notifications can be weird)

Re-parsing doesn't affect the de-duplication -- when tracking the data buffer per task, I also track whether we've emitted the associated output for a given task (using its signature) and guard against this before emitting for that task again. We only go down this path (emitting for a given task) once we've received the task completed event.

I mentioned this inline as well but for visibility: the re-parsing just addresses the fact that for a given task signature, we have an accumulated data buffer that contains *all* possible diagnostic messages coming from the compiler. I find that the user ergonomics aren't great when simply emitting the entire string blob through the observability scope, since these diagnostics can have varying severities and we'll have to decide up-front which severity to choose to emit the *entire* string of *all* diagnostics.

I do also maintain a list of the `DiagnosticInfo` that we omit in favour of emitting the `OutputInfo` containing the same diagnostic messages, but with the plus of having the pre-formatted code snippet (the `DiagnosticInfo` is missing the pre-formatted code snippet but contains enough information to recreate it ourselves, and it was suggested that we instead fall back to using the `OutputInfo` for that reason).

Perhaps some more discussion is needed here. 😄

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: Let me try to clarify: Swift Build _already_ parses the individual diagnostic messages into structured data objects independently of the singular output buffer, which is for the textual output of the tool. So Swift Build is already doing the equivalent of what splitIntoDiagnostics is doing, and those messages are given to you in the `diagnostics` message. They include rich information including severity, line number, and so on.

Similar to what I mentioned elsewhere about the taskStarted/taskEnded events, you want to capture those diagnostics' association to the specific task (e.g. unprocessedDiagnostics should be a dictionary rather than an array, and belongs in BuildState rather than in the top level object), and then replay them at task completion time, without attempting to do your own parsing.

bripeticca:
@swift-ci please test

owenv:
The failing tests seem to be command plugins that do things like this
```
// Check if the build log contains "-enable-testing" flag
        let isTestable = result.logText.contains("-enable-testing")
        if isTestable != shouldTestable {
            fatalError("Testability mismatch: expected \(shouldTestable), but got \(isTestable):\n\(result.logText)")
        }
```

Not sure if we're failing to print the command line in the log, or the observabilityscope isn't adding it to logText

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,422 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+        private var taskIDToSignature: [Int: String] = [:]
+        var collectedBacktraceFrames = SWBBuildOperationCollectedBacktraceFrames()
+
+        /// Rich model to store data buffers for a given `SwiftBuildMessage.LocationContext` or
+        /// a `SwiftBuildMessage.LocationContext2`.
+        struct TaskDataBuffer {
+            private var taskSignatureBuffer: [String: Data] = [:]
+            private var taskIDBuffer: [Int: Data] = [:]
+            private var targetIDBuffer: [Int: Data] = [:]
+            private var globalBuffer: Data = Data()
```
owenv: It doesn't look like we currently print this anywhere. Emitting it at the end of the build might work to start, but we might also consider printing up to each newline we see since some "build preparation" messages get emitted w/ no task and target

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,422 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+                    try Basics.AbsolutePath(validating: $0.pathString)
+                })
+            }
+        case .targetStarted(let info):
+            try buildState.started(target: info)
+        case .backtraceFrame(let info):
+            if buildSystem.enableTaskBacktraces {
+                buildState.collectedBacktraceFrames.add(frame: info)
+            }
+        case .planningOperationStarted, .planningOperationCompleted, .reportBuildDescription, .reportPathMap, .preparedForIndex, .buildStarted, .preparationComplete, .targetUpToDate, .targetComplete, .taskUpToDate:
```
owenv: Should we emit any buffered target messages upon receiving targetComplete here?

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
owenv: I realized that while we emit global **diagnostics** fairly often, the build system doesn't emit a lot of global raw output, so this may actually be ok as-is for now. I'm going to look at the build system to see if we can make changes to guarantee that raw output is always attached to a task

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
owenv: Same comment as https://github.com/swiftlang/swift-package-manager/pull/9362#discussion_r2583217630, this may be ok as-is

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: After our conversation earlier I checked this as well and the BuildOperationConsoleOutputEmitted constructors for non-task output _are never called_ (outside of a test). The only possible way output could be non-task-associated at all is through its Decodable initializer, and the sole place that's called is in the IPC decoding over the wire, so no chance because no such objects are constructed in the first place.

I think we should change the API to guarantee this, as global or target-attached console output doesn't make any sense anyways.

For now @bripeticca can just ignore it (meaning, let's eliminate the field from this PR).

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: Yup, not used.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,422 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+
+        func taskSignature(for id: Int) -> String? {
+            if let signature = taskIDToSignature[id] {
+                return signature
+            }
+            return nil
+        }
+
+        mutating func appendToBuffer(_ info: SwiftBuildMessage.OutputInfo) {
+            // Attempt to key by taskSignature; at times this may not be possible,
```
jakepetroules: The reason this taskID vs taskSignature situation exists was because we started to transition from IDs (which change on every build operation) to signatures (which are stable across build operations), but didn't complete it, and also ran into performance issues (106579386) and turned them off for console output at least... I don't remember why and maybe we can re-enable them at some point, but we should definitely clean this up (not that it should block this PR though).

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,422 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+                unprocessedDiagnostics.append(info)
+            }
+        case .output(let info):
+            // Append to buffer-per-task storage
+            buildState.appendToBuffer(info)
+        case .taskStarted(let info):
+            try buildState.started(task: info)
+
+            if let commandLineDisplay = info.commandLineDisplayString {
+                self.observabilityScope.emit(info: "\(info.executionDescription)\n\(commandLineDisplay)")
```
jakepetroules: These log emissions still need to be deferred to the task complete event, like the diagnostics and output.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,422 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+}
+
+/// Handler for SwiftBuildMessage events sent by the SWBBuildOperation.
+public final class SwiftBuildSystemMessageHandler {
+    private let observabilityScope: ObservabilityScope
+    private let logLevel: Basics.Diagnostic.Severity
+    private var buildState: BuildState = .init()
+
+    let progressAnimation: ProgressAnimationProtocol
+    var serializedDiagnosticPathsByTargetName: [String: [Basics.AbsolutePath]] = [:]
```
jakepetroules: This needs to be keyed by targetID, not target name. Target names are not globally unique.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: It would be good to remove targets from the in-progress map when hitting targetComplete though.

Sources/Basics/Observability.swift
```diff
@@ -160,6 +172,8 @@ public final class ObservabilityScope: DiagnosticsEmitterProtocol, Sendable, Cus
 
 public protocol DiagnosticsHandler: Sendable {
     func handleDiagnostic(scope: ObservabilityScope, diagnostic: Diagnostic)
+
+    func printToOutput(message: String)
```
jakepetroules: Call this `func print(_ output: String, verbose: Bool)` instead? Since we already have a viable implementation.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,422 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+            } else {
+                "\(progressInfo.message)"
+            }
+            progressAnimation.update(step: step, total: 100, text: message)
+            buildSystem.delegate?.buildSystem(buildSystem, didUpdateTaskProgress: message)
+        case .diagnostic(let info):
+            if info.appendToOutputStream {
+                emitInfoAsDiagnostic(info: info)
+            } else {
+                unprocessedDiagnostics.append(info)
```
jakepetroules: I think we don't need to save these at all. By definition these are the diagnostics which were parsed from the textual output and therefore the textual output will contain them.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,422 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+            let message = if let targetName = progressInfo.targetName {
+                "\(targetName) \(progressInfo.message)"
+            } else {
+                "\(progressInfo.message)"
+            }
+            progressAnimation.update(step: step, total: 100, text: message)
+            buildSystem.delegate?.buildSystem(buildSystem, didUpdateTaskProgress: message)
+        case .diagnostic(let info):
+            if info.appendToOutputStream {
+                emitInfoAsDiagnostic(info: info)
```
jakepetroules: We need to capture task-level diagnostics and defer emission of those until the taskCompleted event.

Global diagnostics and target diagnostics can be emitted immediately.

Sources/Basics/Observability.swift
bkhouri: issue: the function name is a bit misleading. Where if the output going to? to a file, stdout, stderr, somewhere else? can we please update the function name to make this more explicit?

bkhouri: I don't think it's the `DiagnosticHandlers` responsibility to print the diagnostic here.  We should instead have a "LoggingHandler" which would know where to emit the "message", whether it's to stdout, stderr, to a file, or some other place.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,422 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
     }
 }
 
+/// Convenience extensions to extract taskID and targetID from the LocationContext.
+extension SwiftBuildMessage.LocationContext {
```
bkhouri: question: Should these extension belong in `SwiftBuild`?

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,422 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+        switch self {
+        case .task(_, let id), .target(let id):
+            return id
+        case .global, .globalTask:
+            return nil
+        }
+    }
+}
+
+/// Handler for SwiftBuildMessage events sent by the SWBBuildOperation.
```
bkhouri: suggestion: can we move this class to it's own source file?

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -174,6 +174,422 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
...
+
+    let progressAnimation: ProgressAnimationProtocol
+    var serializedDiagnosticPathsByTargetName: [String: [Basics.AbsolutePath]] = [:]
+
+    /// Tracks the diagnostics that we have not yet emitted.
+    private var unprocessedDiagnostics: [SwiftBuildMessage.DiagnosticInfo] = []
+    /// Tracks the task IDs for failed tasks.
+    private var failedTasks: [Int] = []
+    /// Tracks the tasks by their signature for which we have already emitted output.
+    private var tasksEmitted: Set<String> = []
```
bkhouri: question (possibly-blocking): what is the difference between `tasksEmitted` and `taskIDsEmitted`?  If they are tracking the same thing, can we remove one?  If we need both, could you please provide an explanation?

bkhouri (dismissed):
This is great work, but we should also include automated tests as part of this change to ensure we won't regress the behaviour.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: Thanks for the clarification!

Sources/Basics/Observability.swift
bripeticca: This function is meant to be a temporary workaround given that we are emitting diagnostics from the compiler output accumulated as one string blob, and observability scope usually demands a severity along with this despite the fact that these diagnostics could very possibly have differing severities.

It's a little hacky but I'm hoping that future discussion surrounding how we'd like to architect logging will improve this :)

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: They certainly could, though in this context I only use them within this file. There are other ways in which we extract the `taskID` and `targetID` from `LocationContext`, however the ergonomics surrounding a `if case let <...>` is not my favourite 😄 

If folks prefer it, I can post a PR on the swiftbuild end.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: Definitely!

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: Many `taskID`s can map to a singular `taskSignature`, and so I initially thought it could be helpful to track both separately. However, I think simply tracking by `taskSignature` may be sufficient here, so I removed `taskIDsEmitted` in the latest change.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: I added a diagnostic buffer to handle these cases; it will track whether a `diagnostic` event is related to a `taskID`, and if so it will append it to the buffer and wait for the associated `taskComplete` event to emit.

If this is a global/target diagnostic, then I emit it as-is here.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: Sounds good -- removed this property in the latest change.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: I'll have to do a little bit of bookkeeping here, since this ultimately gets passed into a `BuildResult`, which expects this to be keyed by target name. 

Initially, I did attempt to change the mapping that `BuildResult` holds to also use a `targetID`, but this doesn't easily scale in other components that use this model (`BuildOperation`, for example, tries to create this map using the build plan's modules which only has a target name accessible).

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: > Many taskIDs can map to a singular taskSignature

I think that shouldn't be true within a single build operation. Are you seeing that to not be the case?

Sources/SwiftBuildSupport/SwiftBuildSystemMessageHandler.swift
```diff
@@ -0,0 +1,533 @@
...
+        case .task(_, let id), .target(let id):
+            return id
+        case .global, .globalTask:
+            return nil
+        }
+    }
+
+    var isGlobal: Bool {
+        switch self {
+        case .global, .globalTask:
```
jakepetroules: Move .globalTask to the false case -- where this is used, we need to defer emission of diagnostics for any kind of task, whether that's a target-attached task or a global task.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: I have seen this behaviour locally, usually between two tasks that have an execution description of `Emit Swift module (arm64)` and `Compile <filename>.swift (arm64)` respectively; they'll share a task signature but have differing task IDs.

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: That sounds like a bug. The signatures are supposed to be globally unique.

cc @owenv

jakepetroules: I suspect we should avoid using the signatures at all right now, and use _only_ the numeric task IDs for safety.

Sources/SwiftBuildSupport/SwiftBuildSystemMessageHandler.swift
```diff
@@ -0,0 +1,539 @@
...
+
+/// Handler for SwiftBuildMessage events sent by the SWBBuildOperation.
+public final class SwiftBuildSystemMessageHandler {
+    private let observabilityScope: ObservabilityScope
+    private let logLevel: Basics.Diagnostic.Severity
+    private var buildState: BuildState = .init()
+
+    let progressAnimation: ProgressAnimationProtocol
+    var serializedDiagnosticPathsByTargetID: [Int: [Basics.AbsolutePath]] = [:]
+    var serializedDiagnosticPathsByTargetName: [String: [Basics.AbsolutePath]] {
```
jakepetroules, line 41: This is going to end up mixing diagnostics from potentially completely unrelated targets.

I see the existing BuildResult structure is forcing you into providing this, and it's only used by the Migrate command that was added recently. That should be refactored to use a globally unique identifier for the target instead (`ResolvedModule.ID`?), but that can be done separately from this PR.

For now I would add a FIXME comment that effectively repeats what I said above, though.

PhantomInTheWire (commented):
hey @bripeticca just a quick one do you think your changes might fix https://github.com/swiftlang/swift-package-manager/issues/8877?

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: In this case, these two tasks also produce the same output (particularly the diagnostics) which helped to reduce duplicate logs. I'll try some alternative approaches locally but I may have to track both ID and signature for the time being.

bripeticca:
@PhantomInTheWire I'll take a look and see if this addresses the issue, thanks for the heads up!

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

bripeticca:
Errors seem unrelated to this change, re-triggering tests

@swift-ci please test linux

bripeticca:
@swift-ci please test macOS

bripeticca:
@swift-ci please test linux

bripeticca:
@PhantomInTheWire It looks like this change will definitely affect possible solutions for swiftlang/swift-package-manager#8877 , but I am skeptical that it will resolve it entirely. If you're interested in taking a look at it still, I would highly recommend doing so once this PR is merged 😄

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
bripeticca: I created an issue to investigate this in a separate PR: #9485

PhantomInTheWire:
sure i'll take a look once this is merged, @bripeticca

jakepetroules:
@swift-ci test

jakepetroules (approved):

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci test windows

bripeticca:
@swift-ci test

bripeticca:
@swift-ci test windows

Tests/SwiftBuildSupportTests/SwiftBuildSystemMessageHandlerTests.swift
```diff
@@ -0,0 +1,355 @@
...
+
+        #expect(observability.hasErrorDiagnostics)
+
+        try expectDiagnostics(observability.diagnostics) { result in
+            result.check(diagnostic: "Simple diagnostic", severity: .error)
+        }
+    }
+
+    @Test
+    func testManyDiagnosticsReported() throws {
```
bkhouri, line 86: praise: When it comes to "arrays", I tend to want to validate the empty, 1, 2 and many items.   It's great to see the 0, 1 and many diagnostics tests added.  Could I ask  to cover case where 2 diagnostics are emitted?

Tests/SwiftBuildSupportTests/SwiftBuildSystemMessageHandlerTests.swift
```diff
@@ -0,0 +1,355 @@
...
+
+        for event in events {
+            _ = try messageHandler.emitEvent(event)
+        }
+
+        #expect(observability.hasErrorDiagnostics)
+
+        try expectDiagnostics(observability.diagnostics) { result in
+            result.check(diagnostic: "Simple diagnostic", severity: .error)
+            result.check(diagnostic: "Another diagnostic", severity: .debug)
```
bkhouri, line 129: question: maybe I'm missing something, but is this diagnosing a `.note` kind? if so, why do we expect it to be `debug`?

Tests/SwiftBuildSupportTests/SwiftBuildSystemMessageHandlerTests.swift
```diff
@@ -0,0 +1,355 @@
...
+
+        let anotherWarningDiagnosticString: String = "[warning]: Another warning diagnostic\n"
+        let anotherWarningOutputInfo: SwiftBuildMessage = .outputInfo(
+            data: data(anotherWarningDiagnosticString),
+            locationContext: .task(taskID: 3, targetID: 1),
+            locationContext2: .init(targetID: 1, taskSignature: "warning-diagnostic")
+        )
+
+        let events: [SwiftBuildMessage] = [
+            .taskStartedInfo(taskID: 1, taskSignature: "simple-diagnostic"),
```
bkhouri, line 168: question: do events need to start and end with the  `.taskStartedInfo(...)` and `.taskCompleteInfo(...)` for these tests?

Tests/SwiftBuildSupportTests/SwiftBuildSystemMessageHandlerTests.swift
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    @Test
+    func testCompilerOutputDiagnosticsWithoutDuplicatedLogging() throws {
+        let (messageHandler, outputStream, observability) = createMessageHandler()
+
+        let simpleDiagnosticString: String = "[error]: Simple diagnostic\n"
+        let simpleOutputInfo: SwiftBuildMessage = .outputInfo(
+            data: data(simpleDiagnosticString),
+            locationContext: .task(taskID: 1, targetID: 1),
```
bkhouri, line 142: question: do the values of `taskID` and `targetID` need to match between all test `SwiftBuildMessage` that are in the `events` array?  If so, consider storing the values in a `let` and using said variable name in various places to make the test more readable.

Tests/SwiftBuildSupportTests/SwiftBuildSystemMessageHandlerTests.swift
```diff
@@ -0,0 +1,355 @@
...
+        for event in events {
+            _ = try messageHandler.emitEvent(event)
+        }
+
+        let outputText = outputStream.bytes.description
+        #expect(outputText.contains("error"))
+    }
+
+    @Test
+    func testDiagnosticOutputWhenOnlyWarnings() throws {
```
bkhouri, line 212: suggestion: should we add additional test that verify other diagnostics?  e.g.: only `.note`, `.debug` and all other possible diagnostic kind?

Tests/SwiftBuildSupportTests/SwiftBuildSystemMessageHandlerTests.swift
```diff
@@ -0,0 +1,355 @@
...
+                taskSignature: taskSignature,
+                result: result,
+                signalled: signalled,
+                metrics: metrics
+            )
+        )
+    }
+
+    /// SwiftBuildMessage.DiagnosticInfo
+    package static func diagnosticInfo(
```
bkhouri, line 287: suggestion: rename this to `diagnostic` or `diagnosticData` as `Info` may give a false sense that were are creating an `info` level diagnostics message/event

Tests/BuildTests/BuildSystemDelegateTests.swift
```diff
@@ -47,8 +47,8 @@ struct BuildSystemDelegateTests {
             case .swiftbuild:
                 let searchPathRegex = try Regex("warning:(.*)Search path 'foobar' not found")
                 #expect(
-                    stderr.contains(searchPathRegex),
-                    "log didn't contain expected linker diagnostics. stderr: '\(stdout)",
+                    stderr.contains("ld: warning: search path 'foobar' not found"),
+                    "log didn't contain expected linker diagnostics. stderr: '\(stderr)",
                 )
                 #expect(
                     !stdout.contains(searchPathRegex),
```
bkhouri, line 54: issue (possibly-blocking): this should match the string from `stderr` expectation?

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -4536,7 +4535,8 @@ struct PackageCommandTests {
                 #expect(stdout.contains("Build complete!"))
 
                 // We expect a warning about `library.bar` but not about `library.foo`.
-                #expect(!stderr.contains(RelativePath("Sources/MyLibrary/library.foo").pathString))
+                let libraryFooPath = RelativePath("Sources/MyLibrary/library.foo").pathString
+                #expect(!stderr.components(separatedBy: "\n").contains { $0.contains("warning: ") && $0.contains(libraryFooPath) })
                 if data.buildSystem == .native {
```
bkhouri, line 4540: suggestion: It's not related to your change, but since you are modifying this file, can we convert the `if data.buildSystem == .native` to a `switch` call, by explicitly adding a `case` call for each value (ie: do not use `default`)?  this will ensure we handle "new build systems" should we ever add one in the future.

Tests/FunctionalTests/PluginTests.swift
```diff
@@ -1717,7 +1717,7 @@ struct PluginTests {
                     #expect(stdout.contains("Build of product 'MyLocalTool' complete!"), "stdout:\n\(stdout)")
                 case .swiftbuild:
                     #expect(stdout.contains("MySourceGenBuildTool-product"), "stdout:\n\(stdout)\nstderr:\n\(stderr)")
-                    #expect(stderr.contains("Creating foo.swift from foo.dat"), "stdout:\n\(stdout)\nstderr:\n\(stderr)")
+                    #expect(stdout.contains("Creating foo.swift from foo.dat"), "stdout:\n\(stdout)\nstderr:\n\(stderr)")
```
bkhouri, line 1720: praise: this is great to see!

Tests/SwiftBuildSupportTests/SwiftBuildSystemMessageHandlerTests.swift
bripeticca: Yes -- the way the message handler works is that it will track when a task is completed to emit the diagnostics/output associated with that task. A task started event is needed before we handle a task completed event, otherwise the handler will throw an error since shouldn't be possible.

Tests/SwiftBuildSupportTests/SwiftBuildSystemMessageHandlerTests.swift
bripeticca: They would need to match if I'm trying to mock out a stream of events for which I want to track its progress per ID. I'll refactor this in a follow-up PR!

PR #9368, opened 2025-11-12
Link Swift Testing documentation in its test template

jerryjrchen:
Link Swift Testing documentation in its test template

### Motivation:

This hopefully makes it easier for new users to find relevant documentation for our test framework!

It also matches the existing pattern where XCTest documentation is linked in its respective test template.

### Modifications:

- Add a link to Swift Testing documentation in the test template.
- Trims some trailing whitespace. Realistically, I don't have much control over this since my editor is respecting the project's .editorconfig's `trim_trailing_whitespace = true` directive. I can manually put them back if that would be better.

### Result:

Default Swift Testing template will include a comment linking to the developer docs.

Testing it locally:
```
~/scraps/autogen/xx.ZtJsA ❯ /Users/jerrychen/workspace/opensource/swift-package-manager/.build/arm64-apple-macosx/debug/swift-package init
Creating library package: xx.ZtJsA
Creating Package.swift
Creating .gitignore
Creating Sources
Creating Sources/xx.ZtJsA/xx_ZtJsA.swift
Creating Tests/
Creating Tests/xx.ZtJsATests/
Creating Tests/xx.ZtJsATests/xx_ZtJsATests.swift
~/scraps/autogen/xx.ZtJsA ❯ /bin/cat Tests/xx.ZtJsATests/xx_ZtJsATests.swift
import Testing
@testable import xx_ZtJsA

@Test func example() async throws {
    // Write your test here and use APIs like `#expect(...)` to check expected conditions.
    // Swift Testing Documentation
    // https://developer.apple.com/documentation/testing
}
```

### DIFF
--- a/Sources/Workspace/InitPackage.swift
+++ b/Sources/Workspace/InitPackage.swift
@@ -549,7 +549,7 @@ public final class InitPackage {
                     func createBuildCommand(for inputPath: URL, in outputDirectoryPath: URL, with generatorToolPath: URL) -> Command? {
                         // Skip any file that doesn't have the extension we're looking for (replace this with the actual one).
                         guard inputPath.pathExtension == "my-input-suffix" else { return .none }
-                        
+
                         // Return a command that will run during the build to generate the output file.
                         let inputName = inputPath.lastPathComponent
                         let outputName = inputPath.deletingPathExtension().lastPathComponent + ".swift"
@@ -644,7 +644,7 @@ public final class InitPackage {
             content = """
             // The Swift Programming Language
             // https://docs.swift.org/swift-book
-            // 
+            //
             // Swift Argument Parser
             // https://swiftpackageindex.com/apple/swift-argument-parser/documentation
 
@@ -731,6 +731,8 @@ public final class InitPackage {
 
                 @Test func example() async throws {
                     // Write your test here and use APIs like `#expect(...)` to check expected conditions.
+                    // Swift Testing Documentation
+                    // https://swiftpackageindex.com/swiftlang/swift-testing/documentation
                 }
 
                 """


grynspan:
Is there an equivalent link on swift.org we could use?

jerryjrchen:
> Is there an equivalent link on swift.org we could use?

Closest I could find is https://www.swift.org/documentation/core-libraries/#swift-testing, which transitively links to the Swift Testing GitHub repo.

Sources/Workspace/InitPackage.swift
```diff
@@ -731,6 +731,8 @@ public final class InitPackage {
 
                 @Test func example() async throws {
                     // Write your test here and use APIs like `#expect(...)` to check expected conditions.
+                    // Swift Testing Documentation
+                    // https://developer.apple.com/documentation/testing
```
bkhouri: issue (blocking): I love this and completely support it.  Though,I have been _disappointed_ (for a lack of a better word) previously while inspecting the Apple XCTest documentation when targeting non-Apple devices as the XCTest API is not identical on macOS and Linux.  Although I suspect this is not the case with Swift Testing, my experience has lead me to not have confidence that the [developer.apple.com](http://developer.apple.com/) documentation will be valid on non-apple platforms.

Can the link point to the [Swift.org](http://swift.org/) Swift Testing API. If that documentation is not published, I would say updating the template is gating upon that.

bkhouri (changes requested):

Sources/Workspace/InitPackage.swift
grynspan: > Though,I have been _disappointed_ (for a lack of a better word) previously while inspecting the Apple XCTest documentation when targeting non-Apple devices as the XCTest API is not identical on macOS and Linux.

Feel free to open a PR, although I have no idea _where_ you'd do so.

grynspan: > Can the link point to the [Swift.org](http://swift.org/) Swift Testing API. If that documentation is not published, I would say updating the template is gating upon that.

Could we link to swiftpackageindex.com? There's a copy of the documentation hosted there.

jerryjrchen: Updated to link to Swift Package Index

grynspan:
Works for me™.

bkhouri:
> Looks good to me. the only downside is https://swiftpackageindex.com/swiftlang/swift-testing/documentation currently displays "An unknown error occurred"
> 
> We should reach out to the Swift Package Index folks to determine what is happening

This issue was intermittent.  it's working now.

bkhouri:
@swift-ci test

jerryjrchen:
@swift-ci test

bkhouri:
@swift-ci test windows

PR #9378, opened 2025-11-14
Improve reliability of online/offline detection when fetching dependencies

plemarquand:
When fetching dependencies on Darwin we relied on SCNetworkReachability to determine if the user is offline when a dependency fetch fails. This check considered `SCNetworkReachabilityFlags.transientConnection` to be considered "offline", however this state still has connectivity.

If the user had the dependency cached, this would result in the cached local dependency being used instead of a fetch of the latest dependency. Ultimately this meant very occasionally the user may end up with a stale dependency when they had all the requirements met to ge the latest one.

Move off of the SCNetworkReachability APIs as they're not really recommended for determining online status, and simply check if the fetch failed with a "Could not resolve host" error. This is what we actually want to check for anyway. Even if the host is down but the connection is fine the host being down is what is relevant to whether or not we should try and use the cached dependency.

### DIFF
--- a/Sources/SourceControl/RepositoryManager.swift
+++ b/Sources/SourceControl/RepositoryManager.swift
@@ -46,7 +46,7 @@ public class RepositoryManager: Cancellable {
     // Limits how many concurrent operations can be performed at once.
     private let asyncOperationQueue: AsyncOperationQueue
 
-    private var emitNoConnectivityWarning = ThreadSafeBox<Bool>(true)
+    private var emitNoConnectivityWarning = ThreadSafeBox<[String: Bool]>([:])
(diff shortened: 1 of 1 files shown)


plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

rconnell9 (approved):

Issue #8378, opened 2025-03-17
Add support for enable/disable-get-task-allow-entitlement with swiftbuild build system


cmcgee1024:
The test case `BuildCommandSwiftBuildTests.testGetTaskAllowEntitlement` is currently skipped because the test doesn't work with the swiftbuild build system. The test is in place to verify that a specific entitlement `com.apple.security.get-task-allow` is `codesigned` into an executable or not based on factors, such as the build configuration (release or debug), and also certain hidden flags `--enable/disable-get-task-allow-entitlement` that override the configuration. The entitlement itself, like all entitlements, is macOS specific and it grants extra debugging capabilities with the executable. https://developer.apple.com/documentation/security/resolving-common-notarization-issues#Avoid-the-get-task-allow-entitlement

This test case needs to be adjusted to not rely so heavily on parsing the build output. Instead, it should verify the executables after each build to see if the entitlement is present, or not. It can use something like the following command and check the output for the get-task-allow entitlement:

```
codesign -d --entitlements - <path_to_executable>
--
[Dict]
        [Key] com.apple.security.get-task-allow
        [Value]
                [Bool] true
```

Also, the SwiftBuildSupport will need to pass along the command-line option override, if present to swift-build as a build setting: `ENTITLEMENTS_DONT_REMOVE_GET_TASK_ALLOW` for the test to succeed in some of the cases.

jakepetroules:
See `checkEntitlements` in https://github.com/swiftlang/swift-build/blob/main/Sources/SwiftBuildTestSupport/CoreQualificationTester.swift for an implementation of reading entitlements (which also handles a bunch of edge cases with the simulator).

svmkr-dev:
Hello, I've been looking into this to see if maybe I could resolve it.

There's one thing which confused me in the code. In `SWBProvisioningTaskInputs` and `CoreQualificationTester` there's this concept of "simulated entitlements". Could you explain or link some resource which explains what are simulated enitlements and how are they different from signed entitlements?

Additionally, `CoreQualificationTester` code suggests both entitlements sets should be same, but currently in [`PlanningOperationDelegate`](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/SwiftBuildSupport/SwiftBuildSystem.swift#L141) used in `SwiftBuildSystem` if simulator SDK is detected, `get-task-allow` entitlement (so iOS version of entitlement) is added but only to signed entitlements, not simulated, which adds to confusion.

EDIT:
[...]

Issue #9078, opened 2025-08-29
`noWarningFromRemoteDependencies` tests in `MiscellaneousTests` fails against Swift Build


bkhouri:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

While working on #9039, which augmented the `MiscellaneousTests` to run against native and Swift Build build system, in addition to the `debug` and `release` build configuration, test `noWarningFromRemoteDependencies ` failed when run against `swiftbuild` build system. 

This issue is open to track the resolution of the test.

Search for the GitHub issue in the repository to track the impacted tests

### Expected behavior

All tests pass without a `withKnownIssue`

### Actual behavior

The `noWarningFromRemoteDependencies(data:)` test fails.
```
Test noWarningFromRemoteDependencies(data:) recorded an issue with 1 argument data → BuildData(buildSystem: SPMBuildCore.BuildSystemProvider.Kind.swiftbuild, config: PackageModel.BuildConfiguration.release) at MiscellaneousTests.swift:1154:13: Expectation failed: ((stdout + stderr) → "Building for production...
14:19:05  0%: Computing dependencies
14:19:05  0%: Computing dependencies
14:19:05  0%: Provisioning 4 / 7
14:19:05  0%: Constructing description
14:19:05  0%: Pre-planning 1 / 125
14:19:05  0%: Planning deferred tasks
14:19:05  15%: dep2 10 / 60
14:19:05  49%: dep2 31 / 62
14:19:05  54%: dep2 35 / 64
14:19:05  75%: 49 / 64
14:19:05  83%: app-product 57 / 68
14:19:05  96%: app 66 / 68
14:19:05  100%: 
14:19:05  Build complete! (3.51 secs)
14:19:05  Fetching /tmp/Miscellaneous_DependenciesWarnings.h7bmF5/dep2
14:19:05  Fetching /tmp/Miscellaneous_DependenciesWarnings.h7bmF5/dep1
14:19:05  Fetched /tmp/Miscellaneous_DependenciesWarnings.h7bmF5/dep2 from cache (0.03s)
14:19:05  Fetched /tmp/Miscellaneous_DependenciesWarnings.h7bmF5/dep1 from cache (0.03s)
[...]

Issue #8511, opened 2025-04-16
Tests: some `AbsolutePath` and `RelativePath` test fail on Windows


bkhouri:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

There a `AbsolutePath` and `RelativePath` test failures on Windows.  These may be causing other tests to fails.

### Expected behavior

All `AbsolutePath` and `RelativePath` tests pass on Windows.

### Actual behavior

Currently some `AbsolutePath` and `RelativePath` tests fails.  These tests were updated in #8450 

### Steps to reproduce

1. Clone #8450 or `git pull` the branch once #8450  is merged
2. Remove all `withKnownIssue` blocks in the tests
3. `swift test --filter PathTests --filter AbsolutePathTests --filter RelativePathTests`

### Swift Package Manager version/commit hash

based on #8450

### Swift & OS version (output of `swift --version ; uname -a`)

Windows 11 Enterprise

```
PS C:\Users\bkhouri\Documents\git\swift-package-manager> swift --version
Swift version 6.2-dev (LLVM d9449c924457b3a, Swift b57f8a9750f4b7b)
Target: x86_64-unknown-windows-msvc
Build config: +assertions
```

PR #9380, opened 2025-11-15
Implement enable/disable-get-task-allow for swift build (#8378)

svmkr-dev:
Implement enable/disable-get-task-allow for swift build backend

### Motivation:

Resolve issue #8378 

### Modifications:

Rewrite of BuildCommandTest.getTaskAllowedEntitlement to check for entitlements using `codesign`. Also use build type as test parameter to shorten test code slightly.
In `SwiftBuildSystem` `PlanningOperationsDelegate` receives argument used to decide whether add macOS `get-task-allow` entitlement when Swift Build requests provisioning data. When we want to add entitlement we need to use at least ad-hoc signing so code also ensures it is done. It is only added on macOS.
Added settings override for swift-build instructing to not remove get-task-allow when we want to have it.

Also fixed wording in warning produced when user uses get-task-allow command line options on unsupported platforms.

### Result:

When building with swift-build backend:
* debug executable for macOS will have macOS get-task-allow entitlement
* any executable for macOS built with option `--enable-get-task-allow-entitlement` will have macOS get-task-allow entitlement

### DIFF
--- a/Tests/CommandsTests/BuildCommandTests.swift
+++ b/Tests/CommandsTests/BuildCommandTests.swift
@@ -1300,6 +1300,9 @@ struct BuildCommandTestCases {
     }
 
     @Test(
+        // Windows builds of ExecutableNew using swiftbuild can fail because of problem with handling long paths which
+        // is root cause of linked issue
+        .issue("https://github.com/swiftlang/swift-package-manager/issues/9420", relationship: .defect),
         .tags(
             .Feature.CommandLineArguments.DisableGetTaskAllowEntitlement,
             .Feature.CommandLineArguments.EnableGetTaskAllowEntitlement,
@@ -1390,34 +1393,38 @@ struct BuildCommandTestCases {
 
             verify(entitlements: entitlements, getTaskAllow: .forbidden)
             #else
-            var buildResult = try await build(
-                ["-v"],
-                packagePath: fixturePath,
-                configuration: buildConfiguration,
-                buildSystem: buildSystem
-            )
+            try await withKnownIssue(isIntermittent: true) {
+                var buildResult = try await build(
+                    ["-v"],
+                    packagePath: fixturePath,
+                    configuration: buildConfiguration,
+                    buildSystem: buildSystem
+                )
 
-            #expect(!buildResult.stdout.contains("codesign --force --sign - --entitlements"))
+                #expect(!buildResult.stdout.contains("codesign --force --sign - --entitlements"))
 
-            buildResult = try await build(
-                ["--disable-get-task-allow-entitlement", "-v"],
-                packagePath: fixturePath,
-                configuration: buildConfiguration,
-                buildSystem: buildSystem
-            )
+                buildResult = try await build(
+                    ["--disable-get-task-allow-entitlement", "-v"],
+                    packagePath: fixturePath,
+                    configuration: buildConfiguration,
+                    buildSystem: buildSystem
+                )
 
-            #expect(!buildResult.stdout.contains("codesign --force --sign - --entitlements"))
-            #expect(buildResult.stderr.contains(SwiftCommandState.entitlementsMacOSWarning))
+                #expect(!buildResult.stdout.contains("codesign --force --sign - --entitlements"))
+                #expect(buildResult.stderr.contains(SwiftCommandState.entitlementsMacOSWarning))
 
-            buildResult = try await build(
-                ["--enable-get-task-allow-entitlement", "-v"],
-                packagePath: fixturePath,
-                configuration: buildConfiguration,
-                buildSystem: buildSystem
-            )
+                buildResult = try await build(
+                    ["--enable-get-task-allow-entitlement", "-v"],
+                    packagePath: fixturePath,
+                    configuration: buildConfiguration,
+                    buildSystem: buildSystem
+                )
 
-            #expect(!buildResult.stdout.contains("codesign --force --sign - --entitlements"))
-            #expect(buildResult.stderr.contains(SwiftCommandState.entitlementsMacOSWarning))
+                #expect(!buildResult.stdout.contains("codesign --force --sign - --entitlements"))
+                #expect(buildResult.stderr.contains(SwiftCommandState.entitlementsMacOSWarning))
+            } when: {
+                buildSystem == .swiftbuild && ProcessInfo.hostOperatingSystem == .windows
+            }
             #endif
         }
     }


Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -132,6 +147,16 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
                 ).merging(provisioningSourceData.projectEntitlements ?? [:], uniquingKeysWith: { _, new in new })
                 : [:]
 
+            var additionalEntitlements: [String: SWBPropertyListItem] = [:]
+
+            if provisioningSourceData.sdkRoot.contains("simulator") {
+                additionalEntitlements["get-task-allow"] = .plBool(true)
+            }
```
svmkr-dev: I’ve left this here to not introduce some regression by accident, but it feels like it maybe should be handled by my code. 
I’m confused though, because on one hand there’s comment in `swift-build` source code (https://github.com/swiftlang/swift-build/blob/51fa0aee68406cf54fe25c0bb553678315d87113/Sources/SWBTaskExecution/TaskActions/ProcessProductEntitlementsTaskAction.swift#L259) stating when building for simulator, signed entitlements should only contain `get-task-allow`, but on the other hand when checking what entitlements Xcode adds to products it produces for xcode targets (so “app” or “framework” managed by xcode projects, not SwiftPM) I’ve only seen `get-task-allow` in signed entitlements of application binary built for actual iphone (tested on macOS 26, Xcode 26.1). 
But I don’t think SwiftPM can build executables for iOS running on actual device, because as far as I know such executables need provisioning profile and SwiftPM doesn’t allow specifying one (I think?).
Due to that I’m not sure what exact rules are about adding this entitlement. I’d appreciate advice about this code, whether it can stay like this or some changes are necessary, maybe new test needs to be added which builds for iphonesimulator?

svmkr-dev:
With this code `BuildCommandsTestCase.doesNotRebuildWithFlags(data:flags:)` test fails for swift build backend and debug configuration. I think it’s because even though swift build does not rebuild the binary, it is still modified by codesign process executed by swift build. I guess easy “fix” could be passing `—disable-get-task-allow-entitlement` in that test, but it doesn’t seem elegant/right.
I’d appreciate advice about what to do with this test.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -112,13 +112,28 @@ func withSession(
 }
 
 private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sendable {
+    private let shouldEnableMacOsDebuggingEntitlement: Bool
```
owenv: capitalization nit:
```suggestion
    private let shouldEnableMacOSDebuggingEntitlement: Bool
```

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -112,13 +112,28 @@ func withSession(
...
+        self.shouldEnableMacOsDebuggingEntitlement = shouldEnableMacOsDebuggingEntitlement
+    }
+
     public func provisioningTaskInputs(
         targetGUID: String,
         provisioningSourceData: SWBProvisioningTaskInputsSourceData
     ) async -> SWBProvisioningTaskInputs {
-        let identity = provisioningSourceData.signingCertificateIdentifier
+        // if we need to add debug entitlement we have to do codesigning, so we need to ensure at least ad-hoc signing
+        let identity = if provisioningSourceData.signingCertificateIdentifier
```
owenv: Instead of falling back to "-" here, I wonder if this would be a little easier to follow if we added "CODE_SIGN_IDENTITY=-" as an override in makeBuildParameters. @jakepetroules do you have any opinions?

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -1054,7 +1082,9 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
     private static func constructDebuggingSettingsOverrides(from parameters: BuildParameters.Debugging) -> [String: String] {
         var settings: [String: String] = [:]
         // TODO: debugInfoFormat: https://github.com/swiftlang/swift-build/issues/560
-        // TODO: shouldEnableDebuggingEntitlement: Enable/Disable get-task-allow
+        if parameters.shouldEnableDebuggingEntitlement {
+            settings["ENTITLEMENTS_DONT_REMOVE_GET_TASK_ALLOW"] = "YES"
```
owenv: This might be a good place to put the CODE_SIGN_IDENTITY override if we move it out of the delegate method

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1280,7 +1282,9 @@ struct BuildCommandTestCases {
     }
 
     @Test(
-        .SWBINTTODO("Test failed because swiftbuild doesn't output precis codesign commands. Once swift run works with swiftbuild the test can be investigated."),
+        .SWBINTTODO(
```
owenv: We're working towards phasing out the "--build-system xcode" option because it's functionality is a strict subset of "--build-system swiftbuild", so I think we should probably drop this TODO and stop attempting to run this test in that mode

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -112,13 +112,28 @@ func withSession(
...
+        {
+            "-"
+        } else {
+            provisioningSourceData.signingCertificateIdentifier
+        }
+
         if identity == "-" {
-            let signedEntitlements = provisioningSourceData.entitlementsDestination == "Signature"
+            let signedEntitlements = provisioningSourceData
+                .entitlementsDestination == "Signature" && provisioningSourceData.sdkRoot.contains("iphoneos")
```
owenv: I believe the second part of this condition should be "not macOS" rather than "is iOS", maybe @jakepetroules can confirm

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
owenv: It may be ok to drop this, but I need to investigate a bit. This code was initially just the bare minimum needed to bring up the new backend so it may not all be strictly correct

owenv (commented):
Thanks for the PR! This change looks like a good improvement to me, but I'd like to double check a few details and give other folks a chance to review as well, which might take a few days. In the meantime I left a handful of minor comments.

> With this code BuildCommandsTestCase.doesNotRebuildWithFlags(data:flags:) test fails for swift build backend and debug configuration. I think it’s because even though swift build does not rebuild the binary, it is still modified by codesign process executed by swift build. I guess easy “fix” could be passing —disable-get-task-allow-entitlement in that test, but it doesn’t seem elegant/right.
I’d appreciate advice about what to do with this test.

Looks like you exposed a case where the task's signature was pulling in some irrelevant settings overrides and invalidating it too often. With your patch layered on top of https://github.com/swiftlang/swift-build/pull/920, I'm now seeing the test pass

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: Should just be `shouldEnableDebuggingEntitlement`, since this applies to all Apple platforms, not just macOS.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: I think this shouldn't check for platform at all.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -132,6 +147,16 @@ private final class PlanningOperationDelegate: SWBPlanningOperationDelegate, Sen
                 ).merging(provisioningSourceData.projectEntitlements ?? [:], uniquingKeysWith: { _, new in new })
                 : [:]
 
+            var additionalEntitlements: [String: SWBPropertyListItem] = [:]
+
+            if provisioningSourceData.sdkRoot.contains("simulator") {
+                additionalEntitlements["get-task-allow"] = .plBool(true)
+            }
+
+            if shouldEnableMacOsDebuggingEntitlement {
```
jakepetroules: I think it should be `com.apple.security.get-task-allow` if macOS or simulator, else `get-task-allow`.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: Instead of turning this on, we should probably make sure `DEPLOYMENT_POSTPROCESSING=YES` is set in PIF generation for release builds.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -726,7 +751,10 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
 
                     let operation = try await session.createBuildOperation(
                         request: request,
-                        delegate: PlanningOperationDelegate(),
+                        delegate: PlanningOperationDelegate(shouldEnableMacOsDebuggingEntitlement: self.buildParameters
+                            .triple.darwinPlatform == .macOS && self.buildParameters.debuggingParameters
```
jakepetroules: Shouldn't need to check for macOS here.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: I think we shouldn't need to provide an override at all. On macOS, CODE_SIGN_IDENTITY defaults to ad-hoc signing if no DEVELOPMENT_TEAM is set, which is already a good default for the SwiftPM CLI.

I don't remember off the top of my head what signingCertificateIdentifier being empty means, but if we need to handle that case you could just adjust the branch below to do `if identity == "-" || identity.isEmpty` instead of overriding the identity.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: From the looks of it, this code was copied from Swift Build's `swbuild` testing CLI, which I think I added for testing purposes as a minimal and incomplete replica of what Xcode does. So there's not much risk of regression because nothing is really using this.

> But I don’t think SwiftPM can build executables for iOS running on actual device, because as far as I know such executables need provisioning profile and SwiftPM doesn’t allow specifying one (I think?).

Technically builds can pick a suitable one by default, but the bigger problem is that SwiftPM presently provides no way to define an app bundle target, which is the only deployable product type (bare executables will not work).

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
svmkr-dev: so just always add platform specific application-identifier entitlement, right? the only "issue" (if it can be called that at all) i see is that I don't think native build system adds application-identifier entitlement on macOS

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
svmkr-dev: when using swiftbuild it seems identity is empty, hence me changing it to ad-hoc: otherwise binary is just signed ad-hoc by linker

`if identity == "-" || identity.isEmpty` seems simple enough so I can go with that if there is no special meaning to signingCertificateIdentifier being empty

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
svmkr-dev: so something like `DEPLOYMENT_POSTPROCESSING=YES` in `PackagePIFBuilder.addProjectBuildSettings()` for release builds and `DEPLOYMENT_POSTPROCESSING=NO` here to ensure release builds with --enable-get-task-allow get the entitlement?

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1289,75 +1293,112 @@ struct BuildCommandTestCases {
...
+        arguments: getBuildData(for: SupportedBuildSystemOnPlatform),
     )
     func getTaskAllowEntitlement(
-        buildSystem: BuildSystemProvider.Kind,
+        data: BuildData,
     ) async throws {
-        try await withKnownIssue(isIntermittent: (ProcessInfo.hostOperatingSystem == .linux)) {
+        let buildSystem = data.buildSystem
+        let buildConfiguration = data.config
+        try await withKnownIssue(isIntermittent: ProcessInfo.hostOperatingSystem == .linux) {
```
bkhouri: This test does not appear to do anything on linux or Windows.  if that is intended, consider removing the `try await withKnownIssue` and add the `.requireHostOS(.macOS)` (or something similar) trait to the test.

svmkr-dev: I'll be removing `withKnownIssue` due to comment [on line R1285](https://github.com/swiftlang/swift-package-manager/pull/9380/files#r2530028600), but by "does not appear to do anything" do you mean that its current behaviour on Linux and Windows is not useful enough to warrant running it?

Currently it's checking for warning message in output when using `enable-` or `disable-get-task-allow` options and ensuring no `codesign` call message is in output log (though build would fail on Linux or Windows if such call occurred when I think about it), which is the same test behaviour as currently on main.
I can confirm it does check for that when running it in dev container or docker image (though this test fails at first fixture build command for swiftbuild because of linker error about `-pie and static` being not compatible, but it doesn't look related to my changes), though I do recognize that it doesn't check for much and may not be useful test at all on Linux or Windows, so I just want to confirm that this is what you mean.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: Yes, although you can skip the `DEPLOYMENT_POSTPROCESSING=NO` bit as it's off by default.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
jakepetroules: Yes and I think that should be fine.

Tests/CommandsTests/BuildCommandTests.swift
svmkr-dev: @bkhouri could you clarify please

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
svmkr-dev: Missed that earlier, but it seems `DEPLOYMENT_POSTPROCESSING` is only declared as macro in swift-build, whereas `PackagePIFBuilder.addProjectBuildSettings()` operates using enums from `ProjectModel.BuildSettings` declared in `swift-build`.
Should I open PR to add `DEPLOYMENT_POSTPROCESSING` to `ProjectModel.BuildSettings.SingleValueSetting` in `swift-build` repo or do you know a good place in SwiftPM to set this setting up?

jakepetroules: Yes, it's standard practice to expose settings in `ProjectModel.BuildSettings.SingleValueSetting` on-demand as they become needed.

owenv:
@swift-ci test

svmkr-dev:
@owenv is there anything i could help with that would speed up accepting this PR?

(also sorry, it seems i have unintentionally removed CI logs from this PR by force-pushing this branch after rebasing it on main - i've had some problems with building this PR in devcontainer today which were resolved by this rebase)

owenv:
@svmkr-dev sorry, I lost track of this, let's run CI again and see if we can get it landed. Feel free to ping me again in 2-3 days if you don't hear back again.

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
CI should be in a better state now
@swift-ci test

owenv:
Seeing some test failures on linux due to link failures
`[2026-01-22T06:12:48.976Z]     /usr/bin/ld.gold: fatal error: -pie and -static are incompatible`
Which is surprising as this should effectively be a no-op there

svmkr-dev:
I remember seeing that linking error in (I think?) one of previous CI logs.
Right now from what i see most errors on linux platforms are from swift-build build system trying to codesign things when swift-build SwiftPM command runs.

I've tried to look into this, but trying to run debugger on swift-build SwiftPM command in linux environments available to me is nightmare. When trying to run it from vscode in devcontainer breakpoints deeper into code don't even register, and if they do, debugger fails to navigate (step-in for example) through code, or resuming code makes swift-build SwiftPM command process kind of... hang? stop executing?. Just running command in debug session if it doesn't hit any breakpoints is ok in sense that command executes as expected (so with codesign error).
Trying to run that command in lldb session in shell within docker container or WSL results in debug session stopping due to signals (SIGSTOP), or just process being unresponsive/stopped or something if I pass those signals back to process and lldb-server losing connection to it. Not sure if i have to refine debugger settings on signal handling to make it work or do something else. It seems that swift-build build system is running in `inProcessStatic` mode, so I'd expect breakpoints to just work, but I also saw swift-build SwiftPM command spawning child process with same command and arguments (so forking i guess), so I probably have yet to understand how connection with swift-build build system works. Maybe I should wait to attach to that child process as if it was `SWBBuildServiceBundle` process.

Possibly self-inflicted debugging issues aside, using some simpler means I found out that when running swift-build SwiftPM command which executes in CI, when swift-build build system constructs codesign tasks (`CodeSign.swift` file in swift-build repo) `CODE_SIGNING_ALLOWED` macro evaluates to true even on Linux. By playing around with `SettingsTests.signingDefaultsByProductType` test from swift-build repo I'm guessing it's because `ProductTypes.xcspec` in `SWBUniversalPlatform` in swift-build build system sets it up and it's never overriden on non-Apple platforms.
After looking at my code here, it seems it was not a problem till now because empty signing identiy disabled code signing  previously (swit-build build system received empty `SWBProvisioningTaskInputs` in that case) and after changes if code signing identiy is empty `SwiftBuildSystemPlanningOperationDelegate` sends ad-hoc signing data.

I went back to my original idea of overriding identity if it's empty and we need to add debugging entitlements and it seems to fix that signing problem on Linux. It works well with `shouldEnableDebuggingEntitlements` because right now it's only set to true if we are on macOS. I've considered proposed approach of using `CODE_SIGNING_IDENTITY` override in `makeBuildParameters` but I don't see how I can verify that identity I want to override is empty, so I feel my approach is safer.

Maybe swift-build build system could benefit for some platform checks on whether to codesign.

svmkr-dev:
@owenv pinging to remind

owenv:
@swift-ci please test

svmkr-dev:
Writing this when Test macOS Platform (smoke test) and Test Windows Platform are still pending

Amazon Linux 2 Platform (Self hosted), Linux platform self hosted and Linux Platform smoke test seem to be failing due to invalid option passed to strip in some test runs.
Windows also seems to have strip problem, but different one: binary is not found in some test runs (it looks like these are different tests than ones failing on linux).
More worrying is that `getTaskAllowEntitlement` test which this PR touched is failing with swift build, but reason is bizarre: it seems like when trying to link sample project being built during test execution, `ExecutableNew.LinkFileList` is missing during the second build (which would be with `--disable-get-task-allow` flag) even though it is being logged in the same build that writing that file occurred.

macOS platform self hosted looks like CI bugging out, because it couldn't even clone project.

To me it doesn't look like these failures are related to my changes, but if there's anything I should or could do here please let me know.

svmkr-dev:
@owenv is there anything I can do on my side?

owenv:
The logs got garbage collected so I'm going to kick off CI again here. It's possible the right path forward might be to go define CODE_SIGNING_ALLOWED to NO either in the build request or the Linux/Windows product types, but hopefully this will give us some more details first

@swift-ci test linux

owenv:
@swift-ci test self hosted

svmkr-dev:
Is that `CODE_SIGNING_ALLOWED` to ensure in the future builds won't fail if someone uses code signing identity on unsupported platform?

I'm guessing these failures with `strip` command on linux are because [StripSymbols.xcspec](https://github.com/swiftlang/swift-build/blob/f49864e7aac2143bafb7c4dc37635bdf52aab8da/Sources/SWBUniversalPlatform/Specs/StripSymbols.xcspec) defines `-T` option which is specific for Swift language and does not appear in linux `strip` implementation. It probably appears right now because of `DEPLOYMENT_POSTPROCESSING` on release builds which this PR adds per Jake Petroules suggestion.

owenv:
> I'm guessing these failures with strip command on linux are because [StripSymbols.xcspec](https://github.com/swiftlang/swift-build/blob/f49864e7aac2143bafb7c4dc37635bdf52aab8da/Sources/SWBUniversalPlatform/Specs/StripSymbols.xcspec) defines -T option which is specific for Swift language and does not appear in linux strip implementation. It probably appears right now because of DEPLOYMENT_POSTPROCESSING on release builds which this PR adds per Jake Petroules suggestion.

Yeah, that looks correct - we should update Swift Build to avoid passing this option when targeting platforms with toolchains that don't support it. I'll take a quick look

owenv:
Running tests on https://github.com/swiftlang/swift-build/pull/1063 which I believe will help

owenv:
Merged the change in the other repo, let's try this again

@swift-ci test

svmkr-dev:
thanks @owenv !
it looks like you'll have to rerun ci for these four failures, it seems like CI itself broke and no tests were run

bkhouri:
@swift-ci smoke test

bkhouri:
@swift-ci test

svmkr-dev:
Looking at windows results, failures are
- most failures are due to `strip` not being present
- there are two failures of `getTaskAllow` test where during build of sample project linkFileList file could not be found when linking (??)
- `resourcesOutsideOfTargetCanBeIncluded` fails due to permissions error when writing a file
- `swiftRunQuietLogLevel` just... fails

I'm not sure what should be done about these.

For macOS, there are 6 recorded issues, all in `getTaskAllow` test.
This test performs 3 builds of some sample project:
1. build without entitlement option
2. build with `--enable-get-task-allow`
3. build with `--disable-get-task-allow`
For debug config we expect builds 1 and 2 to contain entitlement, for release config we expect only second build to contain entitlement.

Issues are raised in following cases:
- native build system with release config, first build
- native build system with release config, third build
- native build system with debug config, third build
- swiftbuild with release config, first build
- swiftbuild with release config, third build
- swiftbuild with debug config, third build

Raised issue is codesign return code 1, which IIRC means signature verification failed.
It looks like test fails only in cases were we do not perform additional codesign. AFAIK all binaries are signed by default by macOS linker, so it kind of looks like something went wrong with default signing mechanism, regardless of build system used in test.
I've tried running this locally with main-snapshot toolchain or 6.2-snapshot toolchain but I could not reproduce these failures.

@owenv do you have any tips on how I could reproduce these failures locally or how should I proceed?

svmkr-dev:
@owenv ping

owenv:
On Windows, I think we should turn the post processing strip task off for the time being since we can't rely on strip.exe being available.

RE: the macOS failures, as a perf optimization the build system passes -no_adhoc_codesign to the linker if it will be signing it itself later. Given the additional overrides here, it's possible that's being applied incorrectly. This is also a case where the x86 and arm64 behavior may differ (CI is primarily x86 hardware at the moment), so that could be a factor in reproducing. Although I'd expect to see less strict validation on x86 CI rather than more

svmkr-dev:
> On Windows, I think we should turn the post processing strip task off for the time being since we can't rely on strip.exe being available.

Ok, I'll take care of that.

> RE: the macOS failures, as a perf optimization the build system passes -no_adhoc_codesign to the linker if it will be signing it itself later. Given the additional overrides here, it's possible that's being applied incorrectly. This is also a case where the x86 and arm64 behavior may differ (CI is primarily x86 hardware at the moment), so that could be a factor in reproducing. Although I'd expect to see less strict validation on x86 CI rather than more

Thanks! CI being on x86 sounds like the thing that's causing it. I don't have x86 macOS hardware to check it, but here's what I think is happening:

Because build is on x86, linker is not doing adhoc signature - only arm macOS requires signature to run a binary. I can't find source files for current ld linker used, but that's ld64 behaviour according to its code (https://github.com/apple-oss-distributions/ld64/blob/1a4389663d65d6630e4b3e31ace2a86b6183b452/src/ld/Options.cpp#L6202-L6208) so I'm guessing new linker is same. Codesign returns 1 because it couldn't find any signature (and prints "code object is not signed at all" to stderr).

I'll remove expectations checking for codesign exit status, since they are superfluous and improve entitlement checking in that test, since get-task-allow is actually what we want to focus on there.

svmkr-dev:
@owenv would be great if you could take a look at my last commit just to be sure all code is reviewed, maybe CI run can be launched then to confirm all is good.

owenv:
I think this is on the right track, I'll take a closer look sometime in the next couple days - thanks for your patience working through this. Let's kick off tests in the meantime

owenv:
@swift-ci test

owenv:
Last Windows failure is 
```
[2026-02-18T07:50:38.252Z]     error: FileSystemError(kind: TSCBasic.FileSystemError.Kind.noEntry, path: Optional(<AbsolutePath:"C:\Users\ContainerAdministrator\AppData\Local\Temp\ValidLayouts_SingleModule_ExecutableNew.OYFcBZ\ValidLayouts_SingleModule_ExecutableNew\.build\out\Intermediates.noindex\ExecutableNew.build\Debug-windows\ExecutableNew.build\Objects-normal\x86_64\ExecutableNew.LinkFileList">))
```

I _think_ this is an unrelated failure related to handling of long paths on Windows, in which case we'd just want to mark the test as `withKnownIssue(intermittent: true)` on a Windows host OS, since @daveinglis is in the process of fixing up many of these issues

svmkr-dev:
> Last Windows failure is
> 
> ```
> [2026-02-18T07:50:38.252Z]     error: FileSystemError(kind: TSCBasic.FileSystemError.Kind.noEntry, path: Optional(<AbsolutePath:"C:\Users\ContainerAdministrator\AppData\Local\Temp\ValidLayouts_SingleModule_ExecutableNew.OYFcBZ\ValidLayouts_SingleModule_ExecutableNew\.build\out\Intermediates.noindex\ExecutableNew.build\Debug-windows\ExecutableNew.build\Objects-normal\x86_64\ExecutableNew.LinkFileList">))
> ```
> 
> I _think_ this is an unrelated failure related to handling of long paths on Windows, in which case we'd just want to mark the test as `withKnownIssue(intermittent: true)` on a Windows host OS, since @daveinglis is in the process of fixing up many of these issues

Yep, I think you are right. It seems my username is short enough to not trigger it, but when I added custom `dir` argument to `withTemporaryDirectory` call within `fixture` function to provide slightly longer path as temporary directory I was able to get the same error as CI gets.

I was thinking about linking an issue to this test to ensure `withKnownIssue` gets removed in future. Do you know any issue number I could use for that? I found #8511 about Path tests on Windows, but I'm not sure if fixing it will result in error here disappearing. I also found #9420, this one shares root cause I think.

daveinglis:
> 
> I was thinking about linking an issue to this test to ensure `withKnownIssue` gets removed in future. Do you know any issue number I could use for that? I found #8511 about Path tests on Windows, but I'm not sure if fixing it will result in error here disappearing. I also found #9420, this one shares root cause I think.

#9420 is would the an issues that shares the same root cause, that said, https://github.com/swiftlang/swift-package-manager/pull/9744 may also fix this as there is a small change that shortens the paths to fixture data.

svmkr-dev:
@owenv pinging to remind

owenv:
@swift-ci test

svmkr-dev:
macOS Platform (smoke test) timed out, hence reported failure.

All other failed runs have `noWarningFromRemoteDependencies` test failing as reported in #9078 (issue mentions `withKnownIssue` in that test, but I don't see it on main (https://github.com/swiftlang/swift-package-manager/blob/main/Tests/FunctionalTests/MiscellaneousTests.swift#L1193))

Additionaly, Linux Platform (self hosted) has `commandPluginBuildingCallbacks` test failing for `swiftbuild` and Linux Platform (smoke test) has `taskBacktraces` failing.

owenv:
Those should all be fixed or skipped now

@swift-ci please test

svmkr-dev:
Thank you for review and help with PR! Glad I could contribute.

PR #2889, opened 2020-08-26
PackageLoading: introduce `-Xmanifest`

compnerd:
This introduces a new `-Xmanifest` flag that will allow the user to pass
additional flags to the manifest build invocation.  This is particularly
important for the case of Windows, where the split SDK/toolchain model
requires the additional flags such as `-sdk` or `-resource-dir` to be
passed to the manifest build system.

PR #9390, opened 2025-11-17
Pass build-tools flags for macro compilation

drodriguez:
Similar in spirit to #5966, but where that one was about plugins, this one is about Swift macros.

We reuse the existing `-Xbuild-tools-swiftc` because that was what was designed to be used for (any usage of `swiftc` used for the build tools, which in the case of macros, need to build for the building machine, to not the target machine).

Motivation:
----------

For most cases, compiling against the host toolchain does not need extra arguments, but mixing the arguments for the target with those for the host toolchain can create problems if the target is different enough to the host. In the case of Swift macros, any argument passed to `-Xswiftc` and others was being forwarded to the Swift macro compilation, which is not correct, since the macros need to compile and execute in the host machine, not the target machine.

The original intention of #5966 was that all host compilation used the "host" flags, and the original spelling proposed in #5966 (`-Xhost`) was chosen because of this, but the PR review settled in `-Xbuild-tools-swiftc`, and I think macro compilation did not end up using those flags because it did not convey the actual intention of those flags.

Modifications:
--------------

This creates a parallel variable to the existing `pluginSwiftCFlags` named `hostSwiftCFlags` which picks up the `buildToolsSwiftCFlags` that come from the command line `-Xbuild-tools-swiftc` flags.

With those, it builds a `hostBuildFlags` with mostly empty flags, except for the `swiftCompilerFlags` taking its value from `hostSwiftCFlags`. The rest of the flags are empty because there is not equivalent `-Xbuild-tools-cc`, `-Xbuild-tools-cxx`, to take their values from, and they have never been needed.

When building the `SwiftCommandState`, depending on the `destination`, one set of flags or the other will be chosen, allowing macros to avoid the `-Xswiftc` flags intended for the target.

Result:
-------

With these modifications the `-Xswiftc` flags of the target in complicated setups are not propagated to the macro compilation, while the ones from `-Xbuild-tools-swiftc` are, allowing issuing `swift build` that just works when host and target are very different.

### DIFF
--- a/Sources/CoreCommands/Options.swift
+++ b/Sources/CoreCommands/Options.swift
@@ -444,6 +444,10 @@ public struct BuildOptions: ParsableArguments {
         self._buildToolsSwiftCFlags
     }
 
+    var hostSwiftCFlags: [String] {
+        self._buildToolsSwiftCFlags
+    }
+
     public var buildFlags: BuildFlags {
         BuildFlags(
             cCompilerFlags: self.cCompilerFlags,
@@ -451,6 +455,16 @@ public struct BuildOptions: ParsableArguments {
             swiftCompilerFlags: self.swiftCompilerFlags,
             linkerFlags: self.linkerFlags,
             xcbuildFlags: self.xcbuildFlags
+        )
+    }
+
+    public var hostBuildFlags: BuildFlags {
+        BuildFlags(
+            cCompilerFlags: [],
+            cxxCompilerFlags: [],
+            swiftCompilerFlags: self.hostSwiftCFlags,
+            linkerFlags: [],
+            xcbuildFlags: []
         )
     }
 
--- a/Fixtures/Macros/MinimalMacroPackage/Sources/MacroImpl/StringifyMacro.swift
+++ b/Fixtures/Macros/MinimalMacroPackage/Sources/MacroImpl/StringifyMacro.swift
@@ -1,5 +1,9 @@
 import Foundation
 
+#if SHOULD_NOT_BE_SET
+#error("SHOULD_NOT_BE_SET was passed to macro compilation, but should only affect the target build.")
+#endif
+
 @main
 struct MacroPlugin {
     static func main() throws {
@@ -38,9 +42,14 @@ struct MacroPlugin {
                     try writeMessage(responseData, to: FileHandle.standardOutput)
                 }
             } else if json.keys.contains("expandFreestandingMacro") {
+                #if USE_CUSTOM_EXPANSION
+                let expandedSource = "\"custom_expanded\""
+                #else
+                let expandedSource = "\"expanded\""
+                #endif
                 let response: [String: Any] = [
                     "expandMacroResult": [
-                        "expandedSource": "\"expanded\"",
+                        "expandedSource": expandedSource,
                         "diagnostics": []
                     ]
                 ]
--- a/Sources/CoreCommands/SwiftCommandState.swift
+++ b/Sources/CoreCommands/SwiftCommandState.swift
@@ -926,13 +926,20 @@ public final class SwiftCommandState {
             case (true, true): .noLazy
             }
 
+        let flags = switch destination {
+        case .host:
+            options.build.hostBuildFlags
+        case .target:
+            options.build.buildFlags
+        }
+
         return try BuildParameters(
             destination: destination,
             dataPath: dataPath,
             configuration: self.options.build.configuration ?? self.preferredBuildConfiguration,
             toolchain: toolchain,
             triple: triple,
-            flags: options.build.buildFlags,
+            flags: flags,
             buildSystemKind: options.build.buildSystem,
             pkgConfigDirectories: options.locations.pkgConfigDirectories,
             architectures: options.build.architectures,
--- a/Tests/FunctionalTests/MacroTests.swift
+++ b/Tests/FunctionalTests/MacroTests.swift
@@ -66,4 +66,45 @@ struct MacroTests {
             #expect(stdout.contains("Build complete!"), "stdout:\n\(stdout)")
         }
     }
+
+    @Test(
+        .tags(
+            Tag.Feature.Command.Build,
+            Tag.Feature.CommandLineArguments.XbuildToolsSwiftc,
+        ),
+        arguments: [BuildSystemProvider.Kind.native],
+    )
+    func macroCanBeAffectedByXBuildToolsSwiftcParameters(
+        buildSystem: BuildSystemProvider.Kind,
+    ) async throws {
+        try await fixture(name: "Macros/MinimalMacroPackage") { fixturePath in
+            let (stdout, _) = try await executeSwiftRun(
+                fixturePath,
+                "MacroClient",
+                extraArgs: ["-Xbuild-tools-swiftc", "-DUSE_CUSTOM_EXPANSION"],
+                buildSystem: buildSystem,
+            )
+            #expect(stdout.contains("Macro result: custom_expanded"), "stdout:\n\(stdout)")
+        }
+    }
+
+    @Test(
+        .tags(
+            Tag.Feature.Command.Build,
+            Tag.Feature.CommandLineArguments.XbuildToolsSwiftc,
+        ),
+        arguments: [BuildSystemProvider.Kind.native],
+    )
+    func macroIsNotAffectedByXswiftcParameters(
+        buildSystem: BuildSystemProvider.Kind,
+    ) async throws {
+        try await fixture(name: "Macros/MinimalMacroPackage") { fixturePath in
+            let (stdout, _) = try await executeSwiftBuild(
+                fixturePath,
+                Xswiftc: ["-DSHOULD_NOT_BE_SET"],
+                buildSystem: buildSystem,
+            )
+            #expect(stdout.contains("Build complete!"), "stdout:\n\(stdout)")
+        }
+    }
 }


drodriguez:
@swift-ci please smoke test

drodriguez:
Not sure why the testing for macOS/Linux was not started (I might not have permissions for those?). The self hosted Windows seems to be failing for everyone for unrelated reasons (`swift-tools-protocols` as a dependency requiring SwiftPM 6.2?)

drodriguez:
Friendly ping.

Had someone had the opportunity to look at this and can someone start the testing that I don't see to be able to start?

owenv:
@swift-ci test

finagolfin:
Thanks, [some more info on the forum here](https://forums.swift.org/t/re-examining-meaning-of-global-options-in-cross-compilation-contexts/72583).

drodriguez:
@swift-ci please smoke test

drodriguez:
Added those tests that were requested. At the moment it only seems to work with the native build system, not with swiftbuild. Need to find where those arguments are passed to swiftbuild, because they are passed incorrectly.

drodriguez:
I am afraid this might need to stay as a "native" build system feature for the time being.

In https://github.com/swiftlang/swift-package-manager/blob/a40388b92a9b7582d0a5375a7a67de3a2819ce16/Sources/CoreCommands/BuildSystemSupport.swift#L112-L146 one can see that `SwiftBuildSystem` completely ignores `toolsBuildParameters`. I have looked inside that type to see what it would take, and it might be possible to follow the pattern that `BuildSystem` used, but there are points in which it is not clear to me how things are supposed to work (I have not worked with swift-build at all). It seems to be that the PIF manifest and the invocation of swift-build might not be prepared for having two different compilation environments (cross-compiling). I am not sure if that's part of the plans, or the system was not designed to support cross-compilation (and so features like that should be gated to the native build system only).

drodriguez:

drodriguez:

drodriguez:

drodriguez:

finagolfin:
@xedin, perhaps you'd like to chime in, since you also addressed this issue in the linked forum post.

drodriguez:
@owenv you had looked at this PR in the past. Is there some feedback about the added test or the code (which is unchanged). Should I ping someone else?

finagolfin:
Pinging @rauhul and @ktoso also, as both showed interest in this feature in that forum thread and may be able to help get this in.

Sources/CoreCommands/Options.swift
```diff
@@ -454,6 +458,16 @@ public struct BuildOptions: ParsableArguments {
         )
     }
 
+    public var hostBuildFlags: BuildFlags {
+        BuildFlags(
+            cCompilerFlags: [],
```
owenv, line 463: With this change, -Xcc no longer has an impact on build tools, but there's also no alternative for users to move to. This might not break many workflows in practice, but I do wonder if we should consider adding host versions of the c/cxx/linker flags as well

owenv:
@swift-ci test

drodriguez:
> I do want to make sure the semantics we define for this flag make sense with both backends though. The main relevant difference is that iff the host and target platform are the same, and a target is a dependency of e.g. both a plugin and a library product, `--build-system swiftbuild` will only build the target once, whereas `--build-system native` would build it twice (when cross-compiling, --build-system swiftbuild will build it twice, once for the host and once for the destination, but within a single build operation). In this specific case, I think we should decide if it makes the most sense to:
> 
> 1. Pass only -Xswiftc flags
> 2. Pass only -Xhost-swiftc flags
> 3. Pass both sets of flags
> 4. Emit a warning/error that -Xhost-swiftc is not recommended when not cross-compiling
> 
> I think I lean toward option 3 or 4, but I'd be interested in hearing your thoughts.
> 
> Aside from the higher level design questions, the code itself lgtm.

Let me recap the history of these flags, and why they are useful for cross-compilation, and how I have been using them even when not cross compiling:

Before the introduction of plugins there was a `-Xmanifest` flag (#2889). This was used to pass "additional" flags to the manifest compilation. The rest of the flags (`-Xswiftc`, `-Xcc`, etc.) were "additionally" passed to the rest of the compilations. In my experience, if I wanted to configure something complicated, even if not cross-compiling, I passed `-Xmanifest` and `-Xswiftc` with both, knowing that each one will apply to one of the environments.

In #5966 I followed the same ideas present in `-Xmanifest` for the plugins and because in my experience one should not need different arguments for the manifest compilation than for the plugin, I proposed `-Xhost` (deprecating `-Xmanifest`). The model was still the same: additional flags to pass to the compilation when the manifest or plugins are compiled. The rest of the compilation uses `-Xswiftc`. During the review it was proposed to change the spelling to `-Xbuild-tools-swiftc`, to allow the system to grow to `-Xbuild-tools-cc` or similar if needed (I had never needed that).

In this PR I proposing to extend the reach of `-Xbuild-tools-swiftc` to also affect macros, which was the original idea discussed in #5966, but when the macro implementation landed, this piece was not implemented. Again, the difference between the flags is very clear: `-Xbuild-tools-swiftc` apply to manifest/plugins/macros, `-Xswiftc` applies to the other compilations (target).

---

In my mind, when cross-compiling, the arguments passed to `-Xswiftc` should never reach the manifest/plugins/macros. I am building binaries for the host machine, not for the target. In the case of not cross-compiling, my mental model is still the same to keep things easy: `-Xswiftc` does not get passed to manifest/plugin/macros. If one needs a `-Xswiftc` for some reason, and the manifest/plugin/macros also need it, you pass it twice: once as `-Xswiftc`, once as `-Xbuild-tools-swiftc`. This keep cross-compiling and not-cross-compiling worlds parallel.

It is true what you say that in the case of not-cross-compiling a dependency of the target will be build twice. I think the only way of avoiding the double build is to compare final actions "commands" and determine if they are the same. For the case of not passing any `-Xbuild-tools-swiftc` or `-Xswiftc`, the results are always the same. In the case of passing any of the two, both will have to be compared to avoid the double build. If they differ by a single argument, you will have to double build.

---

So in my mind Option 2 is the only possible option to keep compatible with the current behaviour and mental model that has existed since years ago.

Option 1: the tests that I implemented use `-Xbuild-tools-swiftc -DUSE_CREATING` to change the behaviour of the code depending if a plugin or normal code was being compiled. If one doesn't pass `-Xbuild-tools-swiftc`, the code will always compile as normal code. There is no way of customizing the manifest/plugins/macros compilation.

Option 3: the negative of the previous might happen: if someone expects `-Xswiftc -DUSE_GENERATING` to only apply to normal code, they will end up using it in manifest/plugins/macros. This option also works completely different if we are cross-compiling than if we are not-cross-compiling? That would be a harder mental model (IMO).

Option 4: It will not allow cases like the one used in the tests and described in Option 1. While those tests are artificial, customizing the manifest with `-D` is not uncommon, and those `-D` are not interesting at all for the normal code.

Sources/CoreCommands/Options.swift
drodriguez: I never had the need to pass `-Xcc` to the manifest, plugins or macros. I imagine that there's cases for plugins or macros, but I have not found them. In #5966 the idea behind introducing `-Xbuild-tools-swiftc` is that we could introduce `-Xbuild-tools-cc` was needed to cover those cases.

owenv:
> It is true what you say that in the case of not-cross-compiling a dependency of the target will be build twice. I think the only way of avoiding the double build is to compare final actions "commands" and determine if they are the same. For the case of not passing any -Xbuild-tools-swiftc or -Xswiftc, the results are always the same. In the case of passing any of the two, both will have to be compared to avoid the double build. If they differ by a single argument, you will have to double build.

This is where we run into some technical difficulties today, because the new build system backend currently doesn't support building the same target twice _for the same platform_ within the same build. So, for example, if a package uses a macro that depends on swift-syntax to build a command line tool which also depends on swift-syntax, and we build for macOS on a macOS host, we must pick a single set of compiler flags to build swift-syntax with. It's possible to lift this limitation, but fairly nontrivial.

Macros are fundamentally a bit different from manifests and plugin scripts in this regard because they unlike the dependencies of macros and plugin build tools, they never build for both host and target. As a result I'm not fully convinced the flags should always be applied the same way, even though that does make sense for some use cases.

dschaefer2:
I may have missed this, but do we have an idea of what flags developers want to pass into macro builds?

drodriguez:
> I may have missed this, but do we have an idea of what flags developers want to pass into macro builds?

I cannot go into details and I have to simplify, but we perform a "Linux x86-64 to Linux arm64" cross-compilation. Swift cannot have a compiler resource directory with both architectures in Linux (no Universal Mach-O equivalent in ELF). We need the macros to be compiled against the x86-64 resource directory with the x86-64 flags in order to execute in the x86-64 compiler, but we need to target the arm64 resource directory, with the arm64 flags (the sysroot specially).

(The case of Android cross-compilation is probably easier to think about, because it is similar to iOS: you need the macros to be executed in the host compiler, that is macOS arm64, but you want the final code to be compiled for the target, iOS arm64).

dschaefer2:
> Swift cannot have a compiler resource directory with both architectures in Linux

That feels like a problem we'll need to solve for a few platforms. e.g. the Pi Pico 2 has both ARM and RISCV and it would be great to have a single SDK that supports both.

drodriguez:
> > Swift cannot have a compiler resource directory with both architectures in Linux
> 
> That feels like a problem we'll need to solve for a few platforms. e.g. the Pi Pico 2 has both ARM and RISCV and it would be great to have a single SDK that supports both.

As I said. I cannot share the details, but the architecture is not the only thing that changes. We need two very different set of flags for building for the host and building for the target. The only way at the moment is using `-Xbuild-tools-swiftc` to customize the compilation of the manifest/plugins/macros to the host, while still using `-Xswiftc` to customize the target accordingly. The Swift compiler resource directory is just one of the differences we have to overcome, others are things that cannot be fixed in the Swift toolchain.

dschaefer2:
> We need the macros to be compiled against the x86-64 resource directory with the x86-64 flags in order to execute in the x86-64 compiler, but we need to target the arm64 resource directory, with the arm64 flags (the sysroot specially).

Sorry, I'm a bit confused here. Are you saying the host is x86-64 and the cross-compile target is arm64. The macros are built for host because that's the architecture the compiler is. The only other host targets would be plugin tools, i.e. executables that get added by plugin tools to generate code for the rest of the build.

Everything else would build for the cross-compile target and would use a Swift SDK with its resource dir set to the arm64 bits, or does that not work in your environment?

dschaefer2:
Yeah, from your descriptions so far, it really looks like you want a Swift SDK with a toolset.json file to manage your target builds and let the host builds just use the defaults. The command line args are an older way of doing this and was replaced by Swift SDKs.

drodriguez:
> Sorry, I'm a bit confused here. Are you saying the host is x86-64 and the cross-compile target is arm64. The macros are built for host because that's the architecture the compiler is. The only other host targets would be plugin tools, i.e. executables that get added by plugin tools to generate code for the rest of the build.

Minor detail: the manifest is also a host target (but I think that's not the point of your comment).

> Everything else would build for the cross-compile target and would use a Swift SDK with its resource dir set to the arm64 bits, or does that not work in your environment?

Yes. That's how it works.

> Yeah, from your descriptions so far, it really looks like you want a Swift SDK with a toolset.json file to manage your target builds and let the host builds just use the defaults. The command line args are an older way of doing this and was replaced by Swift SDKs.

We have been doing this for a while. When we started `destination.json` was not a thing. We actually use `destination.json` so we are not normally using `-Xswiftc` directly in the command line anymore, but we are still using `-Xbuild-tools-swiftc` because we have not find a way of avoiding it in the command line. `--toolset toolset.json` doesn't seem that solves a lot for our use case if I understand its purpose correctly (only provides paths to the compiler, linker, etc?)

Edit: the host builds also need many flags to work. In our setup, the sysroot is not in `/` and the compilers are not in `/usr/bin` (think something like Nix, but crazier).

dschaefer2:
The current best spec for Swift SDK's is here:

https://github.com/swiftlang/swift-evolution/blob/main/proposals/0387-cross-compilation-destinations.md

It includes setting resource directories (for static and non-static) and the sysroot in a `swift-sdk.json` file.

We should really be putting effort in making sure we can use a Swift SDK for the host platform as well (and that may even be possible now though I've never seen it done). As we advance the capabilities of the Swift build system to support many platforms, e.g. building a distributed system in a single call to `swift build`, it's going to be impossible to pass the command line arguments to the build system correctly. That's why we created Swift SDKs so you don't have to do that.

drodriguez:
I can investigate if we can use those new specs in our use case, but if you suspect that the host platform might not be supported it might not be an option for my use case at this point. I need to configure the host platform with several dozen flags for things to work (don't ask why, but there's not a single sysroot, but several "sysroot" and our build system selects the necessary ones, which do not live in a single constant path). I can probably hack my way using the other sets of flags, at the risk of being broken in every upgrade of SPM.

But can we rewind a little and look at this PR from a different perspective? Isn't the current code passing the same flags to the target and to the host compilations? Is that correct? For manifest and plugins the code passes the `-Xbuild-tools-swiftc` ones, but it fails to do so for compiling the macro plugins. Even if I manage to use those new `swift-sdk.json` files, wouldn't that still be a bug in the legacy implementation?

dschaefer2:
I'll follow @owenv's lead on this. He's much more the expert on the build system. I was just curious what your use case was and whether there are other architecture solutions to it.

We just need to keep in mind that we are in the middle of the transition from the old native build system to the new SwiftBuild build system. Anything we add at this point needs to be supported in the new one.

owenv:
> But can we rewind a little and look at this PR from a different perspective? Isn't the current code passing the same flags to the target and to the host compilations? Is that correct? For manifest and plugins the code passes the -Xbuild-tools-swiftc ones, but it fails to do so for compiling the macro plugins. Even if I manage to use those new swift-sdk.json files, wouldn't that still be a bug in the legacy implementation?

Conceptually manifests and plugin scripts are a little different from plugin build tool executables and macro executables in that the former two are single files that are compiled and run in order to plan the build, while the latter two are first-class targets in the build graph which build and run while executing the build. To many users I don't think that distinction matters, although it could if for example we decided to start sandboxing manifests by building them for wasm, in which case they'd no longer be targeting the same platform as e.g. macros.

I hate to create more process, but I think it might be beneficial to run a review of a (small) evolution proposal here to fully enumerate the backwards and forwards compatibility concerns and clearly document the behavior to prevent future confusion in this area. What I'm thinking of is basically a table which for each of the relevant flags (-Xswiftc, -Xbuild-tools-swiftc, any others which might be needed...) lists:
1. Which compiles it is applied to (out of targets building for destination, targets building for host, manifest/plugin scripts)
2. Expected behavior when a user runs a build where some combination of host/target/manifest platforms overlap, keeping in mind the limitations of the build systems we have today and how we want to lift those in the future

If you'd like, we could also bring this up with the build and packaging workgroup to get some more opinions on next steps here - I'll be announcing the 3/12 meeting on the forums shortly and we could put this on the agenda if you're interested.

drodriguez:
> Conceptually manifests and plugin scripts are a little different from plugin build tool executables and macro executables in that the former two are single files that are compiled and run in order to plan the build, while the latter two are first-class targets in the build graph which build and run while executing the build. To many users I don't think that distinction matters, although it could if for example we decided to start sandboxing manifests by building them for wasm, in which case they'd no longer be targeting the same platform as e.g. macros.

That is fair. My proposal is intended to solve a problem we have with the current system. If some day manifests and plugin scripts compile into wasm, I imagine the `-Xbuild-tools-swiftc` might not be the best name. One will need to stop passing those flags into the manifest and plugin scripts, while the flags for build tools keep being passed to plugin build tool executables and macro executables. But this is also the same if macro executables end up being compiled into wasm for portability and run into some sandbox. Maybe I should have fight a little bit more for `-Xhost` back then, but I didn't.

I also think trying to fix this today for the problem we have today is not a bad thing, even if in the future it might not be the best solution. This kinds of flags are probably used rarely, so deprecating and moving into a new system should still be possible.

> I hate to create more process, but I think it might be beneficial to run a review of a (small) evolution proposal here to fully enumerate the backwards and forwards compatibility concerns and clearly document the behavior to prevent future confusion in this area. What I'm thinking of is basically a table which for each of the relevant flags (-Xswiftc, -Xbuild-tools-swiftc, any others which might be needed...) lists:
> 
> 1. Which compiles it is applied to (out of targets building for destination, targets building for host, manifest/plugin scripts)
> 2. Expected behavior when a user runs a build where some combination of host/target/manifest platforms overlap, keeping in mind the limitations of the build systems we have today and how we want to lift those in the future

"here" meaning in this PR or somewhere else? I can try to write something, but it was not on my head trying to define future direction, but fixing a bug (IMO) which was not following the direction we talk about in #5966 (but I cannot find the part of discussion that I thought happened about macros). The proposal in https://forums.swift.org/t/re-examining-meaning-of-global-options-in-cross-compilation-contexts/72583 seems quite well thought, even if it did ignore the existing facilities (`-Xbuild-tools-swiftc` existed at the moment of posting for more than a year), and I don't agree with the general consensus that Option 1 is better (but I don't see a problem with it, I just don't prefer it). It also doesn't cover more cases than "host" and "target", like this PR tries, not cases in which there's a third environment involved or one wants to pass different options to different target kinds (but in my travels I have not found a case that in which I need to provide a different set of flags to the manifest and to Swift macros).

I am fine, if you all decide to go with the post proposal, and I can try to help the best I can, but since the proposal is probably not going to be implemented tomorrow, and since existing flags like `-Xmanifest` and `-Xbuild-tools-swiftc` will need to be migrated to any proposal, I wonder if this could be accepted, knowing that its usages will need to be migrated to any new system.

> If you'd like, we could also bring this up with the build and packaging workgroup to get some more opinions on next steps here - I'll be announcing the 3/12 meeting on the forums shortly and we could put this on the agenda if you're interested.

Happy to attend if time zones allow. I have looked in the forums for meeting announcements like the C++ Interop group does (https://forums.swift.org/t/c-interoperability-workgroup-upcoming-meetings/63016), but I haven't found one for the Package Manager, but I am probably looking in the wrong place.

drodriguez:
Converting this to a draft to not give the idea that it is up for review. In the meeting it was decided that a Swift Evolution process will be necessary for this change, since other parties might be interesting to pitch in, and future directions can be discussed better.

PR #9416, opened 2025-11-25
Include linker generated symbol in default defined set

daniel-grumberg:
It seems like `_GLOBAL_OFFSET_TABLE_` is defined by the linker without being present in any object file on x86_64.

### DIFF
--- a/Sources/BinarySymbols/ReferencedSymbols.swift
+++ b/Sources/BinarySymbols/ReferencedSymbols.swift
@@ -15,7 +15,26 @@ package struct ReferencedSymbols {
     package private(set) var undefined: Set<String>
 
     package init() {
-        self.defined = []
+        // Some symbols are defined by linker directly by convention and need to be assumed defined.
+        // The list below was pulled from
+        // https://github.com/llvm/llvm-project/blob/177e38286cd61a7b5a968636e1f147f128dd25a2/lld/ELF/Config.h#L632
+        self.defined = [
+            "__bss_start",
+            "_etext",
+            "etext",
+            "_edata",
+            "edata",
+            "_end",
+            "end",
+            "_GLOBAL_OFFSET_TABLE_",
+            "_gp",
+            "_gp_disp",
+            "__gnu_local_gp",
+            "__global_pointer$",
+            "__rela_iplt_start",
+            "__rela_iplt_end",
+            "_TLS_MODULE_BASE",
+        ]
         self.undefined = []
     }
 
--- a/Tests/BinarySymbolsTests/LLVMObjdumpSymbolProviderTests.swift
+++ b/Tests/BinarySymbolsTests/LLVMObjdumpSymbolProviderTests.swift
@@ -34,8 +34,7 @@ struct LLVMObjdumpSymbolProviderTests {
             DYNAMIC SYMBOL TABLE:
             """
         )
-
-        #expect(output.defined.isEmpty)
+        #expect(output.defined == ReferencedSymbols().defined)
         #expect(output.undefined.isEmpty)
     }
 
@@ -51,7 +50,7 @@ struct LLVMObjdumpSymbolProviderTests {
     func detectsUndefinedSymbol() throws {
         let output = try getSymbols("0000000000000000         *UND*  0000000000000000 calloc")
 
-        #expect(output.defined.isEmpty)
+        #expect(output.defined == ReferencedSymbols().defined)
         #expect(output.undefined.contains("calloc"))
     }
 


daniel-grumberg:
@swift-ci please test

Sources/BinarySymbols/ReferencedSymbols.swift
```diff
@@ -13,7 +13,7 @@ package struct ReferencedSymbols {
     package private(set) var undefined: Set<String>
 
     package init() {
-        self.defined = []
+        self.defined = ["_GLOBAL_OFFSET_TABLE_"]
```
FranzBusch: Can we leave a comment here why we include this?

daniel-grumberg: I have since found a bit more information about this. The PR has been updated to point to lld's source code (which provides a list of automatically defined symbols) as well as using the full list to avoid issues in the future hopefully.

daniel-grumberg:
@swift-ci please test

daniel-grumberg:
Errors here seem unrelated...

FranzBusch (approved):

jakepetroules:
What problem is this change solving?

This list will also be different on different platforms, compilers, architectures, etc.; how are we accounting for that?

FranzBusch:
The problem that I ran into is that I build a static library in a `quay.io/pypa/manylinux_2_28_x86_64` image and then ran the audit tool in `swiftlang/swift:nightly-main`. This resulted in the following error:

> Invalid artifact binary /__w/swift-temporal-sdk/swift-temporal-sdk/temporal.artifactbundle/temporal/libTemporal-linux-x86_64.a, found undefined symbols:
- _GLOBAL_OFFSET_TABLE_

The GTO is not a problematic symbol here since it will get resolved at link time.

daniel-grumberg:
Generally there are few well defined symbols that are expected to be inserted in the final binary by the linker, like `_GLOBAL_OFFSET_TABLE` and a few others. It's unfortunate that the list needs to be hardcoded but there isn't another way to derive that information. This wouldn't change based on compilers, just based on platform but the tool is only supported on Linux currently for which this list is exhaustive.

FranzBusch:
@swift-ci please test

FranzBusch:
@swift-ci please test

FranzBusch:
@swift-ci please test

FranzBusch:
@swift-ci please test

daniel-grumberg:
@swift-ci please test

daniel-grumberg:
@swift-ci please test

daniel-grumberg:
@swift-ci please test

FranzBusch:
@swift-ci please test

Tests/BinarySymbolsTests/LLVMObjdumpSymbolProviderTests.swift
```diff
@@ -35,7 +35,6 @@ struct LLVMObjdumpSymbolProviderTests {
             """
         )
 
-        #expect(output.defined.isEmpty)
```
bkhouri, line 38: issue: it seems this expectation has changed.  Can we update the expectation to reflect the expected behaviour?

Tests/BinarySymbolsTests/LLVMObjdumpSymbolProviderTests.swift
```diff
@@ -51,7 +50,6 @@ struct LLVMObjdumpSymbolProviderTests {
     func detectsUndefinedSymbol() throws {
         let output = try getSymbols("0000000000000000         *UND*  0000000000000000 calloc")
 
-        #expect(output.defined.isEmpty)
```
bkhouri, line 54: issue: it seems this expectation has changed. Can we update the expectation to reflect the expected behaviour?

daniel-grumberg: There is now a hardcoded initial set of symbols that are automatically defined by the linker. It would be weird to hardcode the list in here. The test doesn't check anything specific about the definition list, but rather makes sure that the undefined symbol is picked up, so I thought just removing it would make the test less brittle.

bkhouri: Sure, but there is a change in Sources/BinarySymbols/ReferencedSymbols.swift which does not contain a test that verified the defined is not empty. 

Also, has/should this been/be tested with the SwiftBuild build system?

daniel-grumberg: It seemed pointless to test that a hardcoded list remained in place. If you feel strongly I am happy to add it. This is only used by a custom command and doesn't actually participate in any building so I think there is no point in testing with SwiftBuild.

bkhouri: I am adamant about having an expectations checked asthe "defined" is expected to no longer be empty per https://github.com/swiftlang/swift-package-manager/pull/9416/files.  We should ensure the test validates this and to ensure we do does not regress.

Also, do we need to ensure the same behavioural changes apply to the SwiftBuild build system?  I wouldn't want developers to rely on this functionality, only to have it regress when SwiftPM changes to use SwiftBuild S the default - whenever that may occur.  So we may need to modify the SwiftPM's interaction with SwiftBuild, or maybe even the SwiftBuild codebase to implement the same functionality 

The SwiftBuild functionality can be done in a follow-up PR, if applicable.

FranzBusch: > Also, do we need to ensure the same behavioural changes apply to the SwiftBuild build system? I wouldn't want developers to rely on this functionality, only to have it regress when SwiftPM changes to use SwiftBuild S the default - whenever that may occur. So we may need to modify the SwiftPM's interaction with SwiftBuild, or maybe even the SwiftBuild codebase to implement the same functionality

I agree with Daniel that this doesn't require SwiftBuild changes. It is entirely decoupled from the build system.

daniel-grumberg: I have added checks that validate a similar property. @bkhouri Can you provide a follow on review.

daniel-grumberg:
@swift-ci please test

bkhouri:
I'm happy with the change but would like to hear back from @jakepetroules to see if he's ok with the response to https://github.com/swiftlang/swift-package-manager/pull/9416#issuecomment-3599125805

PR #9424, opened 2025-11-26
Add an experimental interface for custom test libraries.

grynspan:
This PR adds command-line experimental support for custom test libraries that don't sit on top of XCTest or Swift Testing (irony inbound). Developers can specify additional libraries to invoke using the `--experimental-testing-library` argument.

In a separate PR, I will teach Swift Testing how to read the `--experimental-testing-library` argument (forwarded as `--testing-library` once per, er, library.) A developer can write a small shim function that gets added to the `__swift5_tests` section as a special test content record and which, when called, invokes its corresponding testing library and translates output to the "language" Swift Testing speaks.

The above description is light on details because we haven't designed out the mechanisms in question. In particular, we don't have a stable configuration/input mechanism, so for the moment I'm just forwarding command-line arguments. We also need to teach some part of the stack how to collate results from multiple libraries to produce a single unified test report in various formats. This PR is just one step.

You can use the new argument like so:

```sh
swift test --experimental-testing-library xctest --experimental-testing-library pytest --experimental-testing-library junit [...]
```

The above command would try to run XCTest, pytest, and JUnit tests hosted in the current Swift package. As of right now, it will run XCTest and then `fatalError()` on pytest and JUnit (until I can plumb things through Swift Testing.)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -529,7 +531,7 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
         let runnerPaths: [AbsolutePath] = switch library {
         case .xctest:
             testProducts.map(\.bundlePath)
-        case .swiftTesting:
+        default:
```
bkhouri, line 534: suggestion: instead of using `default`, can we explicitly call out the individual case so that when a new entry is added, the developer is "forced" to make a conscious decision about the implementation?

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1006,7 +1010,7 @@ final class TestRunner {
                     throw TestError.xcodeNotInstalled
                 }
                 args += [xctestPath.pathString]
-            case .swiftTesting:
+            default:
```
bkhouri, line 1013: suggestion: instead of using `default`, can we explicitly call out the individual case so that when a new entry is added, the developer is "forced" to make a conscious decision about the implementation?

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -1018,10 +1022,17 @@ final class TestRunner {
     #endif
         }
 
-        if library == .swiftTesting {
-            // HACK: tell the test bundle/executable that we want to run Swift Testing, not XCTest.
-            // XCTest doesn't understand this argument (yet), so don't pass it there.
-            args += ["--testing-library", "swift-testing"]
+        var addLibraryArgument = true
+#if os(macOS)
```
bkhouri, line 1026: question: Maybe I'm mistaken, but this seems like a behavioural difference.  Do we really want to add the library argument on non-macOS platforms when they execute XCTest?

Sources/CoreCommands/Options.swift
```diff
@@ -675,24 +677,36 @@ public struct TestLibraryOptions: ParsableArguments {
           help: .private)
     public var explicitlyEnableExperimentalSwiftTestingLibrarySupport: Bool?
 
+    /// Experimental listing of arbitrary testing libraries by name (rather than
+    /// just toggling between XCTest and Swift Testing).
+    @Option(name: .customLong("experimental-testing-library"),
+            help: .private)
+    public var explicitlyEnabledLibraries: [TestingLibrary] = []
```
bkhouri, line 684: question: have we considered make this a type `Set`?

Sources/CoreCommands/Options.swift
```diff
@@ -705,6 +719,35 @@ public struct TestLibraryOptions: ParsableArguments {
...
+        if isEnabled(.swiftTesting, swiftCommandState: swiftCommandState) {
+            result.append(.swiftTesting)
+        }
+
+        // Include all explicitly enabled libraries (other than XCTest and Swift
+        // Testing which we have special-cased.) Since this is an opt-in list
+        // rather than a toggle, there's no need to repeatedly call isEnabled().
+        // `foundLibraries` is just used to dedup inputs.
+        var foundLibraries: Set<TestingLibrary> = [.xctest, .swiftTesting]
+        for library in explicitlyEnabledLibraries where !foundLibraries.contains(library) {
```
bkhouri, line 744: suggestion: the `foundLibraries` wouldn't to be needed if `explicitlyEnabledLibraries` was type `Set`.

bkhouri:
Since changes were made to the Native build system, we need to ensure the experimental change works with the SwiftBuild build system integration.

Sources/Commands/SwiftTestCommand.swift
grynspan: It's an open set of possible values, we have no idea what they could be.

grynspan: Oh I guess we can do `case .swiftTesting, case .other` you mean?

Sources/Commands/SwiftTestCommand.swift
grynspan: Good point.

Sources/CoreCommands/Options.swift
grynspan: The order developers specify libraries in should be respected.

Sources/CoreCommands/Options.swift
grynspan: We want to preserve ordering (note that for now XCTest and Swift Testing always come first to preserve compatibility, but I will probably come back and change that later if/when we productize this experimental feature.)

Sources/Commands/SwiftTestCommand.swift
bkhouri: > Oh I guess we can do case .swiftTesting, case .other you mean?

Yes, that's what I was referring to.

grynspan:
Rebase blocked on #10158 landing.

dschaefer2:
I'm curious why Swift Testing would be involved running these other test libraries. Why wouldn't SwiftPM invoke them directly?

grynspan:
To be blunt: because this is not SwiftPM's job, it's part of Swift Testing's mandate to act as infrastructure for other testing libraries that want to integrate with our tools.

More practically speaking, we want to get SwiftPM out of the business of doing any real work with tests. Instead, it should be responsible for invoking swift-build, then something like swift-test-harness that does all the actual work to run tests.

And from a technical aspect, in order for SwiftPM to discover these libraries and run their tests, it would need to know _how_ to do that. Since these libraries may not be written in Swift or even a compiled language, there's no guarantee that a Package.swift file would exist to declare a dependency on them, and SwiftPM would need to run some amount of code from the built target in order to even know these libraries exist in the first place.

Sources/Commands/SwiftTestCommand.swift
grynspan: Not necessary in this context.

grynspan:
This branch is too stale to salvage, mostly due to added debugger support in recent versions of SwiftPM.

PR #9434, opened 2025-12-01
Provide the downloadable metal toolchain to swiftbuild

rconnell9:
Provide the downloadable metal toolchain to swiftbuild using EXTERNAL_TOOLCHAINS_DIR env variable

### Motivation:

https://github.com/swiftlang/swift-package-manager/issues/9202 [Parity] error: unable to spawn process 'metal' (No such file or directory) 

Requires CI to include Metal toolchain in order to enable Metal tests:
https://github.com/swiftlang/swift-package-manager/issues/9443

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -1261,7 +1261,7 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
     package func createLongLivedSession(name: String) async throws -> LongLivedBuildServiceSession {
         let service = try await SWBBuildService(connectionMode: .inProcessStatic(swiftbuildServiceEntryPoint))
         do {
-            let (session, diagnostics) = try await createSession(service: service, name: name, toolchainPath: buildParameters.toolchain.toolchainDir, packageManagerResourcesDirectory: packageManagerResourcesDirectory)
+            let (session, diagnostics) = try await createSession(service: service, name: name, toolchain: buildParameters.toolchain, packageManagerResourcesDirectory: packageManagerResourcesDirectory)
             let teardownHandler = {
                 try await session.close()
                 await service.close()


Tests/BuildMetalTests/BuildMetalTests.swift
```diff
@@ -0,0 +1,69 @@
...
+            
+            // Check that default.metallib exists
+            let metallibPath = binPath.appending(components:["MyRenderer_MyRenderer.bundle", "Contents", "Resources", "default.metallib"])
+            #expect(
+                localFileSystem.exists(metallibPath),
+                "Expected default.metallib to exist at \(metallibPath)"
+            )
+
+            // Verify we can load the metal library
+            let device = MTLCreateSystemDefaultDevice()!
```
jakepetroules: This will need to be ifdef'ed because this is an Apple-only API, or use a lighter weight way of validating the installation. I'm also not sure if calling this API is a good idea because virtualized environments have had issues in that area IIRC.

rconnell9:
@swift-ci test

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -885,6 +892,14 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
             settings["SWIFT_EXEC"] = buildParameters.toolchain.swiftCompilerPath.pathString
         }
 
+        if let metalToolchainId = buildParameters.toolchain.metalToolchainId {
```
jakepetroules: I suggest setting `settings["TOOLCHAINS"]` in exactly one place, since the successive appending/prepending might get confusing over time. I was about to say that you missed $(inherited), but then I saw TOOLCHAINS gets set above, but then looked again and realized it _is_ an issue in the case where there isn't an existing toolchain override... case in point. So how about something like this:


```swift
let toolchainID = try await session.lookupToolchain(at: buildParameters.toolchain.toolchainDir.pathString)
...

let overrideToolchains = [buildParameters.toolchain.metalToolchainId, toolchainID].compactMap { $0 }
if !overrideToolchains.isEmpty {
    settings["TOOLCHAINS"] = (overrideToolchains + ["$(inherited)"]).joined(separator: " ")
}
```

That way you can see at a glance that $(inherited) is properly handled.

Tests/BuildMetalTests/BuildMetalTests.swift
```diff
@@ -0,0 +1,70 @@
...
```
jakepetroules: Here we should use `#require` so the test records an error instead of crashing, if we fail to create the system Metal device.

```suggestion
            let device = try #require(MTLCreateSystemDefaultDevice())
```

jakepetroules (changes requested):
Looks great! Just one had minor point of feedback about the correctness of handling $(inherited) in constructing the build settings.

rconnell9:
@swift-ci test

rconnell9:
@swift-ci test

Package.swift
```diff
@@ -991,6 +991,13 @@ let package = Package(
             name: "SwiftBuildSupportTests",
             dependencies: ["SwiftBuildSupport", "_InternalTestSupport", "_InternalBuildTestSupport"]
         ),
+        .testTarget(
```
owenv, line 1020: I think we should consider putting the new test in SwiftBuildSupportTests instead of creating a whole new test target since this is a fairly specific test

owenv (approved):

Sources/PackageModel/UserToolchain.swift
```diff
@@ -1041,6 +1045,55 @@ public final class UserToolchain: Toolchain {
...
+        }
+
+        guard let metalPath = try? AbsolutePath(validating: output) else {
+            return nil
+        }
+
+        guard let toolchainPath: AbsolutePath = {
+            var currentPath = metalPath
+            while currentPath != currentPath.parentDirectory {
+                if currentPath.basename == "Metal.xctoolchain" {
```
jakepetroules, line 1099: maybe we should just check that the suffix is .xctoolchain... worried that the name could change in the future, include a version number, etc.

rconnell9: There is an issue there in that when you do not have metal installed you may get something like this where it points you to the ghost binary:
% xcrun --find metal                        
../Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/metal

I'm inclined to review the mechanism more holistically (likely get back to xcodebuild -showComponent) sooner than we may have anticipated, but is sufficient for now.  Thoughts?

jakepetroules: To clarify, my suggestion was:

```suggestion
                if currentPath.basename.ends(with: ".xctoolchain") {
```

I'm not sure what you're referring to by ghost binary - if metal doesn't exist, xcrun should just exit with non-zero.

jakepetroules: I see what you mean now.

rconnell9:
@swift-ci test

rconnell9:

rconnell9:
@swift-ci test

Tests/BuildMetalTests/BuildMetalTests.swift
```diff
@@ -0,0 +1,71 @@
...
+#if os(macOS)
+import Metal
+#endif
+
+@Suite
+struct BuildMetalTests {
+
+#if os(macOS)
+    @Test(
+        .disabled("Require downloadable Metal toolchain"),
```
bkhouri, line 26: suggestion (non-blocking): is it possible to detect the location of the metal toolchain and enable the test if it's detected?

rconnell9:
@swift-ci test macOS

bkhouri:
@swift-ci test windows

rconnell9:
@swift-ci test

rconnell9:
@swift-ci test

Package.swift
rconnell9: I'll include a follow-up

rconnell9:
@swift-ci test

rconnell9:
@swift-ci test

rconnell9:
@swift-ci test linux

rconnell9:
@swift-ci test

rconnell9:
@swift-ci Please Test Amazon Linux 2

rconnell9:
@swift-ci test linux

PR #9444, opened 2025-12-04
Handle invalid source control URLs in registry identity lookup

pepicrft:
## Summary

Add validation for source control URLs at multiple levels to handle malformed URLs gracefully.

## Problem

When macOS `git-credential-osxkeychain` fails to find credentials, it outputs error messages that can get concatenated with the original URL:

```
https://github.com/owner/repo.git': failed looking up identity for https://github.com/owner/repo.git
```

These malformed URLs were being sent to the registry server, causing errors. This was observed in the [Tuist registry](https://tuist.dev) where we added server-side validation to return HTTP 400 for invalid URLs (see [tuist/tuist#8829](https://github.com/tuist/tuist/pull/8829)).

### 1. Add `isValid` property to `SourceControlURL`

```swift
public var isValid: Bool {
    // URLs with whitespace are invalid
    guard !self.urlString.contains(where: \.isWhitespace) else {
        return false
    }
    // Check for standard URL format or SSH-style git URLs
    ...
}
```

Valid source control URLs:
- Don't contain whitespace (error messages contain spaces)
- Are parseable as standard URLs with a host, OR
- Match SSH-style git format (`user@host:path`)

### 2. Early validation in `mapRegistryIdentity`

```swift
guard url.isValid else {
    throw RegistryError.invalidSourceControlURL(url)
}
```

This validates URLs before making any HTTP requests to the registry.

### 3. Handle HTTP 400 from registry servers

```swift
case 400:
    throw RegistryError.invalidSourceControlURL(scmURL)
```

If a URL passes client validation but the server still rejects it with 400 Bad Request, we throw the same specific error.

## Changes

- `Sources/Basics/SourceControlURL.swift`: Added `isValid` computed property
- `Sources/Workspace/Workspace+Registry.swift`: Added early URL validation
- `Sources/PackageRegistry/RegistryClient.swift`: 
  - Added `invalidSourceControlURL` error case
  - Handle 400 response in `lookupIdentities`
- `Tests/BasicsTests/SourceControlURLTests.swift`: Added tests for URL validation
- `Tests/PackageRegistryTests/RegistryClientTests.swift`: Added test for 400 handling

## Test Plan

- [x] `SourceControlURLTests` - Tests for `isValid` property (4 tests)
- [x] `LookupIdentities` - Tests for registry client including 400 handling (6 tests)
- [x] All tests pass (10 tests total)
- [x] Build succeeds

## Related

- Server-side fix in Tuist registry: [tuist/tuist#8829](https://github.com/tuist/tuist/pull/8829)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/PackageRegistry/RegistryClient.swift
+++ b/Sources/PackageRegistry/RegistryClient.swift
@@ -1066,6 +1066,11 @@ public final class RegistryClient: AsyncCancellable {
             )
             observabilityScope.emit(debug: "matched identities for \(scmURL): \(packageIdentities)")
             return Set(packageIdentities.identifiers.map(PackageIdentity.plain))
+        case 400:
+            // 400 indicates the server rejected the URL as invalid.
+            // This can happen when malformed URLs (e.g., containing git credential error messages)
+            // are passed to the registry.
+            throw RegistryError.invalidSourceControlURL(scmURL)
         case 404:
             // 404 is valid, no identities mapped
             return []
@@ -1073,7 +1078,7 @@ public final class RegistryClient: AsyncCancellable {
             throw RegistryError.failedIdentityLookup(
                 registry: registry,
                 scmURL: scmURL,
-                error: self.unexpectedStatusError(response, expectedStatus: [200, 404])
+                error: self.unexpectedStatusError(response, expectedStatus: [200, 400, 404])
             )
         }
     }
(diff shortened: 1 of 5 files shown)


Sources/PackageRegistry/RegistryClient.swift
```diff
@@ -1017,6 +1017,19 @@ public final class RegistryClient: AsyncCancellable {
         timeout: DispatchTimeInterval? = .none,
         observabilityScope: ObservabilityScope
     ) async throws -> Set<PackageIdentity> {
+        // Validate that the source control URL is well-formed before making the request.
+        // This catches malformed URLs that may contain error messages (e.g., from git credential helpers).
+        // Git credential errors often produce messages like:
+        // "'https://github.com/owner/repo.git': failed looking up identity for ..."
+        // which can get concatenated with the URL.
+        guard let url = scmURL.url,
```
fortmarek: wouldn't a better fix be to catch the errors when forming the `scmURL` over handling it here?

pepicrft: Yeah, I'm adjusting that. I also added some code to handle 400 coming from the server.

PR #9475, opened 2025-12-08
Fixup licence header in DisableSigpipe.swift

plemarquand:
It is causing the licence check CI job to fail.


plemarquand:
@swift-ci test

PR #9477, opened 2025-12-08
Refactor ThreadSafeBox to use non-optional type

plemarquand:
Change the underlying storage type from `Value?` to storing `Value`, with optional-specific operations moved to constrained extensions.

This improves type safety by eliminating unnecessary optionality for non-optional types.

This stronger type constraint makes it more difficult for the struct to be misused. For instance in the current implementation the extension method `increment()` will fail to increment if the user forgot to initialize the `ThreadSafeBox` with a value. Now a value must be set initially before increment is called.

Also this patch prevents a race condition in the `memoize` methods where the call to the memoized `body` was not guarded by a lock which allowed multiple threads to call memoize at once and produce inconsistent results.

Finally, fix the subscript setter so it works properly with `structs`, as the `if var value` and then value set will set the value on a copy, not on the `self.underlying` struct. Now that the type is no longer optional this unwrapping isn't necessary and we can set directly on the value.


plemarquand:
@swift-ci test

Sources/Basics/Concurrency/ThreadSafeBox.swift
```diff
@@ -12,137 +12,302 @@
...
-            } else {
+            existingValue + value
+        }
+    }
+}
+
+extension ThreadSafeBox: @unchecked Sendable where Value: Sendable {}
+
+/// Thread-safe value boxing structure that provides synchronized asynchronous memoization of a wrapped value.
+public final class AsyncMemoizableThreadSafeBox<Value: Sendable>: @unchecked Sendable {
```
jakepetroules: We have an existing implementation for this; let's use it: https://github.com/swiftlang/swift-build/blob/main/Sources/SWBUtil/AsyncSingleValueCache.swift

plemarquand: This was to replace the async version of `memoize` that was removed from `ThreadSafeBox`, in order to avoid instantiating the `AsyncCoordinator` in every case.

`AsyncCache` does look like it can replace this use case, so I'll remove this and switch to it.

jakepetroules: AsyncSingleValueCache in this case, not AsyncCache (I edited my post after you already saw it)

plemarquand: Looks like there was already `AsyncThrowingValueMemoizer` in the SwiftPM codebase. Importing `AsyncSingleValueCache` required importing SWBUtil, which isn't really available from SwiftPM as its configured today.

plemarquand:
@swift-ci test

Sources/Basics/Concurrency/ThreadSafeBox.swift
jakepetroules: Awesome, even better :)

And yeah we shouldn't use modules other than SwiftBuild in SwiftPM at this time.

plemarquand:
@swift-ci test macOS Platform

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test windows

plemarquand:
@swift-ci test linux

plemarquand:
@swift-ci test linux

PR #9480, opened 2025-12-08
Use the correct module name for executable targets combined with an executable product with a different name

owenv:
If an executable target is combined with an executable product at the PIF layer, it must retain the target's module name, so that tests which consume it can import the right module name.

### DIFF
--- a/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
@@ -112,7 +112,8 @@ extension PackagePIFProjectBuilder {
         settings[.TARGET_NAME] = product.name
         settings[.PACKAGE_RESOURCE_TARGET_KIND] = "regular"
         settings[.PRODUCT_NAME] = "$(TARGET_NAME)"
-        settings[.PRODUCT_MODULE_NAME] = product.c99name
+        // We must use the main module name here instead of the product name, because they're not guranteed to be the same, and the users may have authored e.g. tests which rely on an executable's module name.
+        settings[.PRODUCT_MODULE_NAME] = mainModule.c99name
         settings[.PRODUCT_BUNDLE_IDENTIFIER] = "\(self.package.identity).\(product.name)"
             .spm_mangledToBundleIdentifier()
         settings[.SWIFT_PACKAGE_NAME] = mainModule.packageName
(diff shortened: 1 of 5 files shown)


owenv:
Closes https://github.com/swiftlang/swift-package-manager/issues/9468

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test windows

owenv:
@swift-ci test self hosted

Tests/CommandsTests/TestCommandTests.swift
```diff
@@ -317,6 +317,31 @@ struct TestCommandTests {
...
+        .tags(
+            .Feature.TargetType.Executable,
+        ),
+        arguments: SupportedBuildSystemOnAllPlatforms, BuildConfiguration.allCases,
+    )
+    func testableExecutableWithDifferentlyNamedExecutableProduct(
+        buildSystem: BuildSystemProvider.Kind,
+        configuration: BuildConfiguration,
+    ) async throws {
+        try await withKnownIssue(isIntermittent: true) {
```
bkhouri, line 331: issue (blocking): can we add a `.bug()` trait referencing the GitHub issue that track the Windows intermittent failure?

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
```diff
@@ -112,7 +112,8 @@ extension PackagePIFProjectBuilder {
         settings[.TARGET_NAME] = product.name
         settings[.PACKAGE_RESOURCE_TARGET_KIND] = "regular"
         settings[.PRODUCT_NAME] = "$(TARGET_NAME)"
-        settings[.PRODUCT_MODULE_NAME] = product.c99name
+        // We must use the main module name here instead of the product name, because they're not guranteed to be the same, and the users may have authored e.g. tests which rely on an executable's module name.
+        settings[.PRODUCT_MODULE_NAME] = mainModule.c99name
```
bkhouri, line 116: suggestion: can we add a PIFBuilder Test that will validate we don't regress this behaviour?

Tests/CommandsTests/TestCommandTests.swift
owenv: This is what IssueWindowsLongPath is for, no? It's marked intermittent so the ongoing work to resolve those issues doesn't trigger "known issue was not recorded" failures

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
owenv: It's really important in this case that we ensure packages compile instead of just producing valid PIF, which is why I added the full text fixture above. Otherwise we might generate valid PIF which then fails to compile the user's sources.

bripeticca (approved):

PR #9482, opened 2025-12-09
Add Dependabot configuration for GitHub Actions

shahmishal:


jakepetroules:
This function of this will be to prompt us whenever new versions of the GitHub Actions workflows we use, are available?

Do we need a similar PR for swift-build?

shahmishal:
>This function of this will be to prompt us whenever new versions of the GitHub Actions workflows we use, are available?
> Do we need a similar PR for swift-build?

Yes, after I get this merged I will create one for swift-build. I wanted to make sure it's working as expected first.

PR #9483, opened 2025-12-09
Mark build server tests as flaky while debugging build description loading issue

owenv:
I've seen these tests fail to load a build description in CI twice. Expand the issue criteria while I debug the root cause.


owenv:
@swift-ci test

owenv:
@swift-ci test self hosted

jakepetroules (approved):

PR #9484, opened 2025-12-09
Expose PackageDependency creation methods

plemarquand:
The versions without the `traits` parameter were public for clients to use, but the version with the `traits` parameter were marked as `package`. Expose them as `public` just like the versions without traits.


plemarquand:
@swift-ci test

plemarquand:
@swift-ci test

PR #9486, opened 2025-12-09
Ensure testable variants of targets depend on the primary variant's resource bundle

owenv:
This is required in order to ensure that package resource accessors are correctly generated for the testable variants

Closes https://github.com/swiftlang/swift-package-manager/issues/9459
Depends on https://github.com/swiftlang/swift-build/pull/967


owenv:
@swift-ci test

PR #9490, opened 2025-12-10
Ensure call to unsafe_await is not called from Swifts concurrency system

rconnell9:
Ensure call to unsafe_await is not called from Swifts concurrency system

### Motivation:

Defensive check in response to:
https://github.com/swiftlang/swift-package-manager/issues/9441

### Modifications:

Ensure call to unsafe_await is not called from Swift's concurrency system (an asynchronous context).

### DIFF
--- a/Sources/Basics/Concurrency/ConcurrencyHelpers.swift
+++ b/Sources/Basics/Concurrency/ConcurrencyHelpers.swift
@@ -27,6 +27,12 @@ public enum Concurrency {
 
 @available(*, noasync, message: "This method blocks the current thread indefinitely. Calling it from the concurrency pool can cause deadlocks")
 public func unsafe_await<T: Sendable>(_ body: @Sendable @escaping () async -> T) -> T {
+    withUnsafeCurrentTask { task in
+        if task != nil {
+            fatalError("This function must not be invoked from the Swift Concurrency thread pool.")
+        }
+    }
+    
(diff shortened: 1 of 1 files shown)


rconnell9:
@swift-ci test

Sources/Basics/Concurrency/ConcurrencyHelpers.swift
```diff
@@ -27,6 +27,12 @@ public enum Concurrency {
 
 @available(*, noasync, message: "This method blocks the current thread indefinitely. Calling it from the concurrency pool can cause deadlocks")
 public func unsafe_await<T: Sendable>(_ body: @Sendable @escaping () async -> T) -> T {
+    withUnsafeCurrentTask { task in
+        if task != nil {
+            fatalError("This function must not be invoked from the Swift Concurrency thread pool.")
```
bkhouri, line 32: question: will the fatal error contains the function reference?  if not, consider replading `function` with `\(#function)` and maybe even include `\(#file)` somewhere in the message .

jakepetroules: Good call, but only if passing in #function as a default parameter argument, otherwise it will be constant, which isn't that useful.

I think backtraces give you most of the value anyways.

rconnell9:
@swift-ci test

rconnell9:
@swift-ci test

PR #9497, opened 2025-12-11
Removing setting of SWIFTPM_TESTS_MODULECACHE during tests as it is no longer required

rconnell9:
Removing setting of SWIFTPM_TESTS_MODULECACHE during tests as it is no longer required


rconnell9:
@swift-ci test

owenv:
This is here for performance rather than correctness, so we should be careful removing it doesn't introduce a regression

PR #9500, opened 2025-12-12
Don't rely on current working directory in BasicTests - use packagePath

rconnell9:
Don't rely on current working directory in BasicTests - use packagePath

### Motivation:

Encountering race conditions with concurrent tests making use of current working directory.  Removing usages.


rconnell9:
@swift-ci test

rconnell9:
@swift-ci test windows

PR #9505, opened 2025-12-13
Remove a couple areas where tests where changing cwd (fixes AL2 crashes)

daveinglis:
Fix crashes on AL2 where tests where changing current working directory


daveinglis:
@swift-ci test

PR #9514, opened 2025-12-15
Do not set a fixture prefix

bkhouri:
Ensure `fixture` prefix is an empty string, and update the tests to reflect the new behaviour.

### DIFF
--- a/Sources/_InternalTestSupport/misc.swift
+++ b/Sources/_InternalTestSupport/misc.swift
@@ -125,7 +125,7 @@ public func testWithTemporaryDirectory<Result>(
         let copyName = fixtureSubpath.components.joined(separator: "_")
 
         // Create a temporary directory for the duration of the block.
-        return try withTemporaryDirectory(prefix: copyName) { tmpDirPath in
+        return try withTemporaryDirectory(prefix: "") { tmpDirPath in
 
             defer {
                 // Unblock and remove the tmp dir on deinit.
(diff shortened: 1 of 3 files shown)


bkhouri:
@swift-ci test

bkhouri:
@swift-ci test

bkhouri:
@swift-ci test windows

bkhouri:
@swift-ci test

bkhouri:
@swift-ci test windows

Sources/_InternalTestSupport/misc.swift
```diff
@@ -167,7 +167,7 @@ public enum TestError: Error {
         let copyName = fixtureSubpath.components.joined(separator: "_")
```
daveinglis, line 167: You could just change this to use .last! that way you get a much shorted but still relevant directory name for the test (good for debugging) instead of just `TemporaryDirectory.<xxxxxxx>`

daveinglis:
note: shortening the paths may cause a bunch of tests that don't have `isIntermittent: true` to fail

bkhouri:

Sources/_InternalTestSupport/misc.swift
bkhouri: the sample package can always be retrieved by inspecting the test.

bkhouri:
> note: shortening the paths may cause a bunch of tests that don't have `isIntermittent: true` to fail

Yep.. I'll update those tests accordingly.

daveinglis:
> > note: shortening the paths may cause a bunch of tests that don't have `isIntermittent: true` to fail
> 
> Yep.. I'll update those tests accordingly.

That change is essentially this https://github.com/swiftlang/swift-package-manager/pull/9268

bkhouri:
@swift-ci test windows

bkhouri:
@swift-ci test

PR #9515, opened 2025-12-15
Add a required job to the workflow

bkhouri:
Add a job that simply echo's a string and depends on all jobs that are to be marked "required".


bkhouri:
@swift-ci test

bkhouri:
Closing PR for now.

PR #9516, opened 2025-12-15
backtrace

jakepetroules:
- also mark withKnownIssues tests as isIntermittent: true so that as we fix things in dependent packages (like swift-build) we don't break swiftPM CI

- serializing some test that change cwd to hopefully fix the AL2 crashes where the CWD has been deleted and calls to get CWD returns nil

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Tests/CommandsTests/BuildCommandTests.swift
+++ b/Tests/CommandsTests/BuildCommandTests.swift
@@ -782,7 +782,7 @@ struct BuildCommandTestCases {
     ) async throws {
         let buildSystem = data.buildSystem
         try await fixture(name: "Miscellaneous/ParseableInterfaces") { fixturePath in
-            try await withKnownIssue(isIntermittent: ProcessInfo.hostOperatingSystem == .windows) {
+            try await withKnownIssue(isIntermittent: true) {
                 let result = try await build(
                     ["--enable-parseable-module-interfaces"],
                     packagePath: fixturePath,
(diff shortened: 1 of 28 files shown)


jakepetroules:

jakepetroules:
@swift-ci test windows

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1148,24 +1148,31 @@ struct BuildCommandTestCases {
...
-                configuration: .release,
-                buildSystem: buildSystem,
-            )
-            if result.stdout.contains(
-                "remark: Incremental compilation has been disabled: it is not compatible with whole module optimization"
-            ) {
-                #expect(result.stdout.contains("optimization\n"))
-                #expect(!result.stdout.contains("optimization["))
-                #expect(!result.stdout.contains("optimizationremark"))
+         try await withKnownIssue(
```
bkhouri, line 1151: issue: this is a regression.  the test was passing on Windows, and not it no longer is.  do we know what caused this?  Can we add a test in SwiftBuild to prevent this regression?

Tests/CommandsTests/PackageCommandTests.swift
```diff
@@ -3090,7 +3088,7 @@ struct PackageCommandTests {
                 }
             }
         } when: {
-            ProcessInfo.isHostAmazonLinux2()
+            ProcessInfo.isHostAmazonLinux2() //rdar://134238535
```
bkhouri, line 3091: do we have a GitHub issue instead of a radar link?

Also, can we add said issue as a `.bug(...)` trait?

Tests/CommandsTests/TestCommandTests.swift
```diff
@@ -1012,10 +1011,10 @@ struct TestCommandTests {
         buildSystem: BuildSystemProvider.Kind,
         configuration: BuildConfiguration,
     ) async throws {
-        try await withKnownIssue("Failes to find test executable") {
+        try await withKnownIssue("Failes to find test executable, or getting error: module 'Simple' was not compiled for testing, onMacOS", isIntermittent: true) {
```
bkhouri, line 1014: issue: The `withKnownIssues` where conditionalized based on configuration and swiftBuild, but we are know unconditionally expecting an intermittent failure.  

This removes the visibility into which conditions passes, and which are actually intermittent.  Can we re-add the `when:` closure associated with the test so we have better understanding of what needs to be fixed when it's time to remove the `withKnownIssues`

Tests/FunctionalTests/MiscellaneousTests.swift
```diff
@@ -877,9 +877,8 @@ final class MiscellaneousTestCase: XCTestCase {
         try await fixtureXCTest(name: "Miscellaneous/RootPackageWithConditionals") { path in
             _ = try await executeSwiftBuild(
                 path,
-                extraArgs: ["--build-system=swiftbuild"],
                 env: ["SWIFT_DRIVER_SWIFTSCAN_LIB" : "/this/is/a/bad/path"],
-                buildSystem: .native,
+                buildSystem: .swiftbuild,
```
bkhouri, line 881: issue: in order to ensure we don't regress on `native` build systems, can we add that test back in?

Tests/FunctionalTests/ResourcesTests.swift
```diff
@@ -17,13 +17,6 @@ import XCTest
 
 final class ResourcesTests: XCTestCase {
     func testSimpleResources() async throws {
-        try XCTSkipOnWindows(
```
bkhouri, line 20: praise: glad to see more tests being enabled.

Tests/PackageGraphTests/ModulesGraphTests.swift
```diff
@@ -114,6 +114,7 @@ struct ModulesGraphTests {
 
     @Test(
         .IssueWindowsPathTestsFailures,
+        .disabled("Requires swift-testing bug fix https://github.com/swiftlang/swift-testing/pull/1441"),
```
bkhouri, line 117: question: Do we really need to disabled this test?  Can we instead guard this based on compiler version?  For example, see the `requireSwift6_2` trait

https://github.com/swiftlang/swift-package-manager/blob/main/Sources/_InternalTestSupport/SwiftTesting%2BTraitConditional.swift#L200-L208

Tests/PackageLoadingTests/PD_6_2_LoadingTests.swift
```diff
@@ -72,21 +72,25 @@ struct PackageDescription6_2LoadingTests {
...
-                    toolchain: try! UserToolchain.default
-                ),
-                observabilityScope: observability.topScope
-            )
-        try expectDiagnostics(validationDiagnostics) { results in
-            results.checkIsEmpty()
-        }
-        try expectDiagnostics(observability.diagnostics) { results in
-            results.checkIsEmpty()
+        try await withKnownIssue("https://github.com/swiftlang/swift-package-manager/issues/8543: there are compilation errors on Windows", isIntermittent: true) {
```
bkhouri, line 75: suggestion: can we apply a `.issue("https://github.com/swiftlang/swift-package-manager/issues/8543", relationship: .defect)` trait on the test?

Tests/WorkspaceTests/RegistryPackageContainerTests.swift
```diff
@@ -24,11 +24,6 @@ import XCTest
 import struct TSCUtility.Version
 
 final class RegistryPackageContainerTests: XCTestCase {
-
-    override func setUpWithError() throws {
```
bkhouri, line 28: praise:  great to see this.

Tests/WorkspaceTests/SourceControlPackageContainerTests.swift
```diff
@@ -125,10 +125,6 @@ private let v1Range: VersionSetSpecifier = .range("1.0.0" ..< "2.0.0")
 
 final class SourceControlPackageContainerTests: XCTestCase {
     func testVprefixVersions() async throws {
-        try XCTSkipOnWindows(because: """
```
bkhouri, line 128: praise:  great to see this.

PR #9408, opened 2025-11-21
Change AsyncProcess stdout/err reads to 4096 instead of Int.max

daveinglis:
- this should allow us to consume available in the Pipe as its filled and flushed but the running process.

rdar://155634107

PR #9539, opened 2025-12-23
Revert "Change AsyncProcess stdout/err reads to 4096 instead of Int.max (#9408)"

meg-gupta:
I am reverting this since it seems like a probable cause for timeouts on sourcekit-lsp tests on windows. Example: https://ci-external.swift.org/job/swift-PR-windows/49333/consoleText


meg-gupta:
@swift-ci test

PR #9540, opened 2025-12-24
Add conceptual documentation for dependency mirrors

### LINKED ISSUE #8867: Add documentation for dependency mirrors
### Description

Currently, the only documentation available for mirrors is seen on the [evolution proposal page](https://github.com/FranzBusch/swift-evolution/blob/3aa2bc4db8a42d560199e47bfcec38c7a23d15d2/proposals/0219-package-manager-dependency-mirroring.md). This feature should also be documented in our DocC catalogue.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

_No response_

_No response_

_No response_

Viditgupta-official:
This PR adds a conceptual documentation page explaining Swift Package Manager dependency mirrors.

It describes:
- What dependency mirrors are
- When they are useful
- How they are configured
- Their scope and safety considerations

It also links to SE-0219 for full design details.

Fixes #8867

### DIFF
--- a/Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
+++ b/Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
@@ -7,7 +7,8 @@ Dependency mirrors allow Swift Package Manager to fetch a package dependency fro
 
 You commonly use dependency mirrors when working in corporate or restricted network environments. They let you redirect dependencies to internal mirrors, or control where dependencies come from without modifying existing package manifests.
 
-Dependency mirrors are configured locally and apply only to the top-level package being built. Mirror configuration is stored outside the package manifest and is not shared when a package is published or checked into version control. Dependency mirrors apply to all versions of a dependency identity and can’t be scoped to individual versions.
+You configure dependency mirrors locally, and they apply only to the top-level package you're building. Swift Package Manager stores mirror configuration outside the package manifest and doesn't share it when you publish a package or check it into version control.
+Dependency mirrors apply to all versions of a dependency identity and can't be scoped to individual versions.
 
 During dependency resolution, Swift Package Manager transparently rewrites dependency source locations based on the configured mirrors, without modifying the package manifest.
 When you configure a mirror for a dependency, Swift Package Manager treats it as authoritative and doesn't fall back to the original source location.


Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,34 @@
+# Dependency Mirrors
+
+Dependency mirrors allow Swift Package Manager to fetch a dependency from an alternate location without modifying the package manifest. This can be useful in environments where dependencies need to be fetched from internal mirrors, cached repositories, or alternate hosting locations.
```
heckj: The lines right after the H1 header in DocC are considered the abstract and shown with every reference to the page. This ideally should be a single, short sentence that works with the article title. I'd recommend reworking this to build this information into a combination of the title and abstract such that the combination provides the detail about dependency mirrors and how to set them up.

Ideally this article shouldn't be just that this concept exists, but how and when you'd use it. To provide a little background, take a look at the articles into which this fits - the ones referenced from `AddDependency.md` and how those titles and abstracts look:

https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/addingdependencies

Image

Structurally for the article, I'd recommend adding in an `## Overview` and then adding the top level detail that you want to use to introduce dependency mirrors that wasn't covered by the title and abstract under that. If you don't put it in explicitly, DocC infers that one exists after the sentence/paragraph right after the `H1` header.

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,34 @@
+# Dependency Mirrors
```
heckj: When you add an article into an existing DocC catalog, if you don't put in where it's referenced, it'll be included at the top level page under "Articles" as a general grouping. You should add in a `<doc:DependencyMirrors>` reference into one of the top-level markdown files that provides the organization for the articles (we call this "curation" - picking where it should go)

Since this is about working with dependencies, the current place where such detail is coordinated is "AddDependencies.md", which is referenced from "Documentation.md". The `AddDependencies.md` acts as a sort of coordination space for information about the various things you can do with dependencies within SwiftPM. I might suggest adding this into the set of  topics at the bottom of that markdown file (the bits under `## Topics`)

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,34 @@
...
+
+Dependency mirrors allow Swift Package Manager to fetch a dependency from an alternate location without modifying the package manifest. This can be useful in environments where dependencies need to be fetched from internal mirrors, cached repositories, or alternate hosting locations.
+
+Dependency mirrors are configured locally and affect only the top-level package being built.
+
+## When to use dependency mirrors
```
heckj: Structural feedback: Use ### instead of ## for section headers within an article. You can absolutely use H2's here, but it'll put the content sort of weirdly promoted up with Overview (and Topics, if you are using the page as a general page that references deeper content)

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Scope and safety considerations
+
+Dependency mirrors apply only to the top-level package being built. Dependencies cannot define or override mirror configurations for downstream packages.
+
+Because mirror configuration is local and external to the package manifest, it is not shared when a package is published or checked into version control.
+
+## Learn more
+
+For the complete design and behavior details, see the Swift Evolution proposal [SE-0219: Package Manager Dependency Mirroring](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0219-package-manager-dependency-mirroring.md).
```
heckj: My personal opinion is that leaning on the evolution proposals for more detail isn't a good idea and shouldn't be encouraged. In the general case, the implementations can sometimes change, and in some cases you need to read a lot more than one proposal to get the full scope of understanding. I'd personally suggest leaving this out, and if there's something missing from the article that is well described in the proposal, copy it into the documentation.

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,34 @@
...
+
+Dependency mirrors are configured using Swift Package Manager commands. These commands allow setting, querying, and removing mirror configurations for package dependencies:
+
+- ``swift package config set-mirror``
+- ``swift package config get-mirror``
+- ``swift package config unset-mirror``
+
+Mirror configuration is stored locally and is not part of the package manifest.
+
+## Scope and safety considerations
```
heckj: Because this is so short, and has the real detail - I'd recommend removing this section altogether and moving the sentence detail of this into the Overview. The first of these sentences is already up there, so I'd group these together into a paragraph within the overview of the content.

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,34 @@
+# Dependency Mirrors
+
+Dependency mirrors allow Swift Package Manager to fetch a dependency from an alternate location without modifying the package manifest. This can be useful in environments where dependencies need to be fetched from internal mirrors, cached repositories, or alternate hosting locations.
+
+Dependency mirrors are configured locally and affect only the top-level package being built.
+
+## When to use dependency mirrors
+
+Dependency mirrors are commonly used in the following scenarios:
```
heckj: Rather than having a section here, I'd recommend this be part of the overview - providing some detail as to when and why you use this feature is perfect for that section.

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,34 @@
...
+
+- Accessing dependencies from internal or corporate-hosted mirrors
+- Working in environments with restricted network access
+- Redirecting dependency URLs without modifying existing package manifests
+
+## Configuring dependency mirrors
+
+Dependency mirrors are configured using Swift Package Manager commands. These commands allow setting, querying, and removing mirror configurations for package dependencies:
```
heckj: I know your PR says "add conceptual" documentation, but I think this could be far more beneficial as a "how do you use it" article, in which case I'd suggest building out section into the core of the article, and provide a detail of how you'd do this and what to expect when you do. 

In it's current form, this doc is saying "hey, this is possible", but it's leaving the reader hanging for anything more for HOW to do it, which - if they got here - they're likely either trying to understand if they should do it, or they're trying to enable it for themselves.

Viditgupta-official:
Thank you for the detailed and thoughtful review — I really appreciate you taking the time to go through this. I agree with the points you raised around structure and alignment with the rest of the documentation. I’ll revise the article to refine the abstract so it better matches the DocC style used elsewhere, restructure the content to align more closely with existing conceptual documentation, and expand on the practical aspects by clarifying when and how dependency mirrors are typically configured. I’ll also add the appropriate curation so the article appears in the correct place within the documentation hierarchy.

I’ll preview the changes locally using the DocC plugin before pushing an updated version. Thanks again for the guidance — it’s very helpful.

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
Viditgupta-official: Thanks for the suggestion — I’ve removed the reference to the Swift Evolution proposal and incorporated the relevant behavioral details directly into the article instead.

Viditgupta-official:
I’ve updated the article to fold usage and scope details into the overview, adjusted section structure to follow DocC conventions, and expanded the configuration section to clearly describe how dependency mirrors are used in practice, including the relevant SwiftPM commands and when they take effect during dependency resolution.

I’ve also removed reliance on Swift Evolution links and curated the article under the existing dependency documentation. Please let me know if you’d like the usage flow expanded further or split into a separate article.

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,31 @@
...
+Dependency mirrors are configured using Swift Package Manager’s local configuration commands. These commands allow you to map a package dependency identity or source location to an alternate location without modifying the package manifest. Commands such as <doc:SwiftBuild>, <doc:SwiftRun>, and <doc:PackageUpdate> trigger dependency resolution.
+
+
+To configure a mirror, you register a mapping between the original dependency source and the mirror location. Once configured, Swift Package Manager automatically applies this mapping during dependency resolution.
+
+For example, if a package depends on a repository hosted at a public Git URL, you can configure Swift Package Manager to fetch that dependency from an internal mirror instead. After setting the mirror, subsequent dependency resolution operations will use the mirrored location transparently, without requiring changes to the package’s dependency declarations.
+
+Mirror configuration affects only the top-level package being built. Once a mirror is set, Swift Package Manager treats it as authoritative and does not fall back to the original source location.
+
+```bash
```
heckj, line 35: These commands don't work quite this directly, at least based on my experimentation with Swift 6.2. All three appear (in Swift 6.2) to require the parameter `--original`, and in the first case it also requires the parameter `--mirror` before the URL.

I'm not familiar with this specific functionality, but I'm not sure how you might inspect a local configuration _without_ knowing the original URL. Honestly, it seems really odd that even if there aren't any mirrors set, that `swift package config get-mirror` doesn't just report "nothing set" versus requiring a `--original` parameter.

Digging into the source at `Sources/Commands/PackageCommands/Config.swift`, it looks like that's intentional.

But while I was in there, the it looks like there's both a local mirrors configuration file and a shared mirrors configuration file that goes into reading and setting these values, which isn't reflected here at all. It appears to be optional in the Workspace (Sources/Workspace/Workspace+Configuration), so maybe it's a leftover that was just never cleaned out as SwiftPM evolved, but let's figure out if this content is even baseline accurate - if there's an expectation of shared vs. local configurations, how you set them, and how those work. When I was digging in the source for the same, I saw a few references to implied values from Environment variables that look like the specific locations could be controlled or overridden.

Before we progress here, I'd suggest we dig up what we think is happening (which means reading/greping through the SwiftPM source) and then checking with the team that puts together SwiftPM about what's expected and intended here. There's a deep enough legacy, with different folks at different times over the past 10 years working on this, that the details might be pretty well buried. I suspect a Swift Forums thread might help to get answers when we're ready to ask questions about intent and validating what we've found.

Would you be willing to take a pass through the source, dig around on this idea of local vs. shared mirrors, and see if there are references elsewhere to this that we should be including here? And then maybe making your own mirror an explicitly trying this out for some repository to verify how the commands work with explicit examples?

Viditgupta-official: I dug into Config.swift, and both SetMirror and GetMirror explicitly require --original. GetMirror is implemented purely as a lookup (mirrors.mirror(for:)) and does not enumerate or report global state. If no mirror exists, it prints “not found” and exits with failure, even when no mirrors are configured at all. This appears intentional rather than a UX bug, since mirrors are keyed strictly by original identity and there’s no API to list them. Local vs shared scope is resolved implicitly via getMirrorsConfig, not exposed at the CLI layer.

Viditgupta-official:
I’ve updated the Dependency Mirrors documentation to reflect the actual key-based mirror model used by SwiftPM. The text now clarifies that mirror configuration is indexed by the original dependency identity, that get-mirror is a lookup requiring --original, and that there’s no global enumeration of mirrors. It also documents the authoritative behavior and the implicit handling of configuration scope, all aligned with the current SwiftPM implementation.

Viditgupta-official:
@heckj I’ve pushed the requested changes addressing the feedback above. Whenever you get a chance, I’d appreciate a quick look. Thanks!

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,44 @@
+# Dependency Mirrors
```
heckj:
```suggestion
# Configuring Dependency Mirrors
```

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,44 @@
...
+
+Dependency mirrors are commonly used when working in corporate or restricted network environments, when redirecting dependencies to internal mirrors, or when controlling where dependencies are sourced from without modifying existing package manifests.
+
+Dependency mirrors are configured locally and apply only to the top-level package being built. Mirror configuration is stored outside the package manifest and is not shared when a package is published or checked into version control. Dependency mirrors apply to all versions of a dependency identity and can’t be scoped to individual versions.
+
+During dependency resolution, Swift Package Manager transparently rewrites dependency source locations based on the configured mirrors, without modifying the package manifest. When a mirror is configured for a dependency, it is treated as authoritative and Swift Package Manager does not fall back to the original source location. Swift Package Manager resolves dependency versions according to the rules described in <doc:ResolvingPackageVersions>.
+
+
+### Configuring dependency mirrors
```
heckj:
```suggestion
### Configure dependency mirrors
```

(looking at other content in this space, section headers are more often imperative phrases than gerund phrase. We use gerund phrase (****-ing something) when organizing API references, but not as much in articles, where we typically want the section headers to read a bit like an outline and be skim-able)

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,44 @@
...
+Dependency mirrors are configured using Swift Package Manager’s local configuration commands. These commands allow you to map a package dependency identity or source location to an alternate location without modifying the package manifest. Commands such as <doc:SwiftBuild>, <doc:SwiftRun>, and <doc:PackageUpdate> trigger dependency resolution.
+
+
+To configure a mirror, you register a mapping between the original dependency source and the mirror location. Once configured, Swift Package Manager automatically applies this mapping during dependency resolution.
+
+For example, if a package depends on a repository hosted at a public Git URL, you can configure Swift Package Manager to fetch that dependency from an internal mirror instead. After setting the mirror, subsequent dependency resolution operations will use the mirrored location transparently, without requiring changes to the package’s dependency declarations.
+
+Mirror configuration affects only the top-level package being built. Once a mirror is set, Swift Package Manager treats it as authoritative and does not fall back to the original source location.
+
+### Inspecting a mirror configuration
```
heckj:
```suggestion
### Inspect a mirror configuration
```

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,44 @@
...
+dependency identity or URL:
+
+```bash
+swift package config set-mirror --original <original-url-or-identity> --mirror <mirror-url-or-identity>
+swift package config get-mirror --original <original-url-or-identity>
+swift package config unset-mirror --original <original-url-or-identity>
+```
+
+Mirror configuration is stored locally and is not part of the package manifest.
+### Configuration scope
```
heckj:
```suggestion
### Understand configuration scope
```

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,44 @@
+# Dependency Mirrors
+
+Dependency mirrors let Swift Package Manager fetch dependencies from alternate locations without modifying the package manifest.
+
+## Overview
+Dependency mirrors allow Swift Package Manager to fetch a package dependency from an alternate location without modifying the package manifest. This is useful in environments where dependencies must be fetched from internal mirrors, cached repositories, or alternate hosting locations. For more information about declaring dependencies, see <doc:AddingDependencies>.
+
+Dependency mirrors are commonly used when working in corporate or restricted network environments, when redirecting dependencies to internal mirrors, or when controlling where dependencies are sourced from without modifying existing package manifests.
```
heckj:
```suggestion
You commonly use dependency mirrors when working in corporate or restricted network environments. They let you redirect dependencies to internal mirrors, or control where dependencies come from without modifying existing package manifests.
```

(switching to active voice and breaking up the sentence a bit to make it more readable)

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,44 @@
...
+
+Dependency mirrors are configured locally and apply only to the top-level package being built. Mirror configuration is stored outside the package manifest and is not shared when a package is published or checked into version control. Dependency mirrors apply to all versions of a dependency identity and can’t be scoped to individual versions.
```
heckj:
```suggestion
You configure dependency mirrors locally, and they apply only to the top-level package you're building. Swift Package Manager stores mirror configuration outside the package manifest and doesn't share it when you publish a package or check it into version control.
Dependency mirrors apply to all versions of a dependency identity and can't be scoped to individual versions.
```

(reworking to active voice)

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,44 @@
...
+
+During dependency resolution, Swift Package Manager transparently rewrites dependency source locations based on the configured mirrors, without modifying the package manifest. When a mirror is configured for a dependency, it is treated as authoritative and Swift Package Manager does not fall back to the original source location. Swift Package Manager resolves dependency versions according to the rules described in <doc:ResolvingPackageVersions>.
```
heckj:
```suggestion
During dependency resolution, Swift Package Manager transparently rewrites dependency source locations based on the configured mirrors, without modifying the package manifest.
When you configure a mirror for a dependency, Swift Package Manager treats it as authoritative and doesn't fall back to the original source location.
Swift Package Manager resolves dependency versions according to the rules described in <doc:ResolvingPackageVersions>.
```

(reworking to active voice)

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,44 @@
...
+
+
+### Configuring dependency mirrors
+Dependency mirrors are configured using Swift Package Manager’s local configuration commands. These commands allow you to map a package dependency identity or source location to an alternate location without modifying the package manifest. Commands such as <doc:SwiftBuild>, <doc:SwiftRun>, and <doc:PackageUpdate> trigger dependency resolution.
```
heckj:
```suggestion
You configure dependency mirrors using Swift Package Manager's local configuration commands.
These commands let you map a package dependency identity or source location to an alternate location without modifying the package manifest.
Commands such as <doc:SwiftBuild>, <doc:SwiftRun>, and <doc:PackageUpdate> trigger dependency resolution.
```

(reworking to active voice)

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,44 @@
...
+
+
+To configure a mirror, you register a mapping between the original dependency source and the mirror location. Once configured, Swift Package Manager automatically applies this mapping during dependency resolution.
+
+For example, if a package depends on a repository hosted at a public Git URL, you can configure Swift Package Manager to fetch that dependency from an internal mirror instead. After setting the mirror, subsequent dependency resolution operations will use the mirrored location transparently, without requiring changes to the package’s dependency declarations.
+
+Mirror configuration affects only the top-level package being built. Once a mirror is set, Swift Package Manager treats it as authoritative and does not fall back to the original source location.
```
heckj:
```suggestion
Mirror configuration affects only the top-level package you're building.
Once you set a mirror, Swift Package Manager treats it as authoritative and doesn't fall back to the original source location.
```

(reworking to active voice)

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,44 @@
...
+Use `swift package config get-mirror` to look up the mirror for a specific
+dependency identity or URL:
+
+```bash
+swift package config set-mirror --original <original-url-or-identity> --mirror <mirror-url-or-identity>
+swift package config get-mirror --original <original-url-or-identity>
+swift package config unset-mirror --original <original-url-or-identity>
+```
+
+Mirror configuration is stored locally and is not part of the package manifest.
```
heckj:
```suggestion
Swift Package Manager stores mirror configuration locally; it's not part of the package manifest.

```

(reworking to active voice, adding a blank line before the section header that comes after)

Sources/PackageManagerDocs/Documentation.docc/Package/DependencyMirrors.md
```diff
@@ -0,0 +1,44 @@
...
+### Configuration scope
+
+Mirror configuration is persisted by Swift Package Manager and may be
+applied from different configuration scopes (such as workspace-local or
+shared configuration). The specific scope used is determined implicitly
+by SwiftPM based on the command context and environment.
```
heckj:
```suggestion
Swift Package Manager persists mirror configuration and may apply it from different configuration scopes (such as workspace-local or shared configuration).
It automatically determines the specific scope to use based on the command context and environment.
```

(rework to active voice)

Viditgupta-official:
@heckj Thanks for the review! I’ve incorporated your suggestions to align the headers with imperative phrasing and revised the text to use active voice and second-person language.

Happy to make any further tweaks if needed, and I’ll wait for sign-off from the SwiftPM team.

Viditgupta-official:
Hi @bripeticca , I noticed you’ve reviewed similar docs changes before, so tagging you here.
Would appreciate a review when convenient—thanks!

bripeticca:
@swift-ci please test

Viditgupta-official:
Hi @bripeticca @heckj! I see that the branch now has merge conflicts due to recent changes on `main`.

Just confirming — would you like me to rebase and resolve the conflicts on my side, or is there any preferred approach you’d recommend?

Happy to update the PR accordingly. Thanks!

Issue #8469, opened 2025-04-09
Documentation Request: How to read other plugin's outputs?


ed-irl:
Related to #6221. Is it possible at all to read another plugin's outputs from a target dependency? If so, what is the correct way to do so? [The documentation](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/Plugins.md) seems to imply this is possible, and perhaps even a common scenario, because it describes a system where input and output tracking between plugins happens and is useful for build optimization:

> Build commands are preferred over prebuild commands when the paths of all of the inputs and outputs are known before the command runs, since they allow the build system to more efficiently decide when they should be run. This is actually quite common. Examples include source translation tools that generate one output file (with a predictable name) for each input file, or other cases where the plugin can control the names of the outputs without having to first run the tool. In this case the build system can run the command only when some of the outputs are missing or when the inputs have changed since the last time the command ran.

`SourceModule` targets expose a `pluginGeneratedSources` property but this appears to be empty for at least `buildCommand` targets. The docs for this property are unclear:

> Paths of any sources generated by other plugins that have been applied to the given target before the plugin currently being executed.
>
> Note: Plugins are applied in order of declaration in the package manifest. Generated files are vended to the target the current plugin is being applied to, but not necessarily to other targets in the package graph.

In this case, what does "not necessarily" mean? It seems to imply that under some circumstances other package graph targets may expose generated sources, but it is unclear when this happens, if it does at all.

ed-irl:
Another observation on this topic: in my project even running two plugins within the same target (in this case the swift protobuf plugin and my own plugin), `pluginGeneratedSources` is not populated; it would be helpful to have clearer documentation of when it is populated.

PR #9543, opened 2025-12-25
Add conceptual documentation for SwiftPM plugins

Viditgupta-official:
This PR adds a conceptual overview of Swift Package Manager plugins.

It explains:
- The two kinds of plugins supported by SwiftPM (command plugins and build tool plugins)
- When and how each type is invoked
- How plugin outputs are handled by the build system
- Visibility and limitations of generated outputs between plugins

This documentation clarifies common points of confusion around plugin execution and output visibility, and addresses the questions raised in #8469.

### DIFF
--- a/Sources/PackageManagerDocs/Documentation.docc/Package/Plugins.md
+++ b/Sources/PackageManagerDocs/Documentation.docc/Package/Plugins.md
@@ -1,0 +1,24 @@
+# Plugins
+
+Swift Package Manager supports plugins that can extend package functionality.
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


heckj:
@swift-ci please test

heckj (changes requested):
Hey Vidit,

Thanks for providing this - did you by chance see the content in Plugins.md, one directory above this?

I think rather than dropping in another article with the same name, which DocC will have some trouble with (it views everything in the catalog as a single namespace, and sort of ignores any folder/directory structure within it), you'd be better revising what's in the existing content to help answer the questions in #8469

PR #9547, opened 2025-12-28
Add functional test for swift test --sanitize --filter failure

### LINKED ISSUE #9546: swift test --filter with --sanitize fails: "Sanitizer load violates platform policy"

### Description

When running swift test with both --filter and --sanitize flags, the test fails to execute because the sanitizer runtime library cannot be loaded into the xctest subprocess.

  Environment:
  - Swift: 6.2.3 (swiftlang-6.2.3.3.21)
  - Xcode: 26.2
  - macOS: 26.1 (Build 25B78)
  - Architecture: arm64

Steps to Reproduce:

  1. Create or use any Swift package with tests
  2. Run: swift test --sanitize address --filter <TestName>

### Expected behavior

Tests matching the filter run with AddressSanitizer enabled.

### Actual behavior

Build succeeds, but test execution fails with:

  Library not loaded: '@rpath/libclang_rt.asan_osx_dynamic.dylib'
  ...
  code signature in '/Applications/Xcode-26.2.0.app/.../libclang_rt.asan_osx_dynamic.dylib'
  not valid for use in process: Sanitizer load violates platform policy

  Workaround:

  Running without --filter works correctly:
  swift test --sanitize address  # ✓ works
  swift test --sanitize address --filter Foo  # ✗ fails

  Analysis:

  When --filter is used, Swift PM spawns xctest as a subprocess to run filtered tests. The sanitizer runtime's code signature appears to be rejected by macOS platform policy when loaded into this subprocess, but works when tests run directly without the filter mechanism.

### Steps to reproduce

_No response_

_No response_

swift-driver version: 1.127.14.1 Apple Swift version 6.2.3 (swiftlang-6.2.3.3.21 clang-1700.6.3.2)
Darwin Alex-M3-Air.local 25.1.0 Darwin Kernel Version 25.1.0: Mon Oct 20 19:26:04 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T8122 arm64

Viditgupta-official:
Add functional test covering swift test --sanitize --filter failure

Motivation:

When running swift test on macOS with both --sanitize address and --filter enabled, test execution fails with the error:

Sanitizer load violates platform policy.

Running either flag independently works as expected, but the combination consistently fails. This behavior is currently undocumented and untested. Adding coverage ensures the issue is reproducible and prevents accidental regressions while the underlying cause is investigated.

Modifications:

Added a new functional test in TestDiscoveryTests

Used SwiftPM’s functional test harness to invoke swift test

Passed --sanitize address and --filter together

Asserted on the observed platform policy violation message

Guarded the test to macOS

Result:

The sanitizer + filter failure is now captured by SwiftPM’s test suite, making the behavior explicit and providing a stable foundation for future fixes or refactors.
Fixes #9546

### DIFF
--- a/Tests/FunctionalTests/TestDiscoveryTests.swift
+++ b/Tests/FunctionalTests/TestDiscoveryTests.swift
@@ -325,7 +325,7 @@ func testWithSanitizeAndFilterFailsWithPlatformPolicyViolation(
 ) async throws {
 #if os(macOS)
 try await withKnownIssue(
-    "Fails due to swiftpm-xctest-helper (Xcode tool) violating platform sanitizer policy"
+    "Fails due to swiftpm-xctest-helper (Xcode tool) violating platform sanitizer policy (#9546, rdar://168234231)"
 ) {
     try await fixture(name: "Miscellaneous/TestDiscovery/Simple") { fixturePath in
         let (_, stderr) = try await executeSwiftTest(


Viditgupta-official:
This PR intentionally only adds coverage for the sanitizer + filter behavior.
I plan to investigate the root cause of the sanitizer runtime loading failure in a follow-up change once this test is in place.

Viditgupta-official:
Note: This test documents a known failure that appears to originate in the swiftpm-xctest-helper (an Xcode tool), not SwiftPM itself.

The test is intentionally added as coverage to prevent regressions and to make the behavior explicit while the underlying issue is tracked separately (via Apple Feedback).

Tests/FunctionalTests/TestDiscoveryTests.swift
```diff
@@ -319,4 +319,26 @@ struct TestDiscoveryTests {
             buildSystem == .swiftbuild && ProcessInfo.hostOperatingSystem == .windows
         }
     }
+	@Test(arguments: SupportedBuildSystemOnAllPlatforms)
+func testWithSanitizeAndFilterFailsWithPlatformPolicyViolation(
+    _ buildSystem: BuildSystemProvider.Kind
+) async throws {
+#if os(macOS)
+try await withKnownIssue(
+    "Fails due to swiftpm-xctest-helper (Xcode tool) violating platform sanitizer policy"
```
stmontgomery: You can also note the associated tracking bugs
```suggestion
    "Fails due to swiftpm-xctest-helper (Xcode tool) violating platform sanitizer policy (#9546, rdar://168234231)"
```

Viditgupta-official: That makes sense! I'll include the commit.

Tests/FunctionalTests/TestDiscoveryTests.swift
```diff
@@ -319,4 +319,26 @@ struct TestDiscoveryTests {
...
+            fixturePath,
+            extraArgs: ["--sanitize", "address", "--filter", "testExample"],
+            buildSystem: buildSystem
+        )
+
+        #expect(stderr.contains("Sanitizer load violates platform policy"))
+    }
+}
+#else
+throw SkipTest("Sanitizer + filter behavior is macOS-specific")
```
stmontgomery, line 341: I'd generally recommend an `.enabled(if:)` for expressing that a particular test is only relevant on certain platforms

Tests/FunctionalTests/TestDiscoveryTests.swift
```diff
@@ -319,4 +319,26 @@ struct TestDiscoveryTests {
             buildSystem == .swiftbuild && ProcessInfo.hostOperatingSystem == .windows
         }
     }
+	@Test(arguments: SupportedBuildSystemOnAllPlatforms)
+func testWithSanitizeAndFilterFailsWithPlatformPolicyViolation(
+    _ buildSystem: BuildSystemProvider.Kind
+) async throws {
+#if os(macOS)
+try await withKnownIssue(
```
stmontgomery, line 327: nit: indentation looks off; I'd expect the body of this whole function to be indented one level

Tests/FunctionalTests/TestDiscoveryTests.swift
```diff
@@ -319,4 +319,26 @@ struct TestDiscoveryTests {
...
+    "Fails due to swiftpm-xctest-helper (Xcode tool) violating platform sanitizer policy (#9546, rdar://168234231)"
```
stmontgomery, line 328: This is the part I commented on/suggested, and that part LGTM! Thanks

PR #9554, opened 2026-01-02
Traits update

heckj:
Updating the traits documentation to touch on missed pieces and revise to conform a little better to developer content guidelines.

### Motivation:

- making it clear that traits are additive across a build tree when requested
- touching on expected constraints when using common trait names that may impact multiple dependencies.

### Modifications:

Grammar revisions and added sentences for clarity.

### DIFF
--- a/Sources/PackageManagerDocs/Documentation.docc/Dependencies/PackageTraits.md
+++ b/Sources/PackageManagerDocs/Documentation.docc/Dependencies/PackageTraits.md
@@ -8,7 +8,7 @@ Swift packages before version 6.1 offer a non-configurable API surface for each 
 Starting with Swift 6.1, packages may offer traits, which express a configurable API surface for a package.
 
 Use traits to enable additional API beyond the core API of the package.
-For example, a trait may enable an experimental API, optional extended functionality that requires additional dependencies, or functionality that isn't critical that you may want to enable only in specific circumstances.
+For example, a trait may enable an experimental API, optional extended functionality that requires additional dependencies, or functionality you may want to enable only in specific circumstances.
 
 Traits that you specify when building a package only activate within the package you're building.
 If your package wants to use a trait in a dependent package, it needs to encode the traits it needs in its dependencies.
@@ -18,15 +18,16 @@ If your package wants to use a trait in a dependent package, it needs to encode 
 Within the package that defines a trait, the trait expresses conditional compilation.
 Swift Package Manager exposes enabled traits as conditional blocks (for example, `#if YourTrait`) that you can use to conditionally enable imports or different compilation paths in code.
 
-Trait names are namespaced within the package that hosts them - a trait name in one package has no impact on any other package.
+Trait names are namespaced within the package that hosts them.
+A trait name in one package has no impact on any other package.
 Trait names must be [valid Swift identifiers](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/lexicalstructure#Identifiers) with the addition of the characters of `-` and `+`.
 Don't use the trait names `default` or `defaults` (regardless of capitalization).
-These values aren't allowed in order to avoid confusion with the default traits that a package defines.
+These values aren't allowed to avoid confusion with the default traits that a package defines.
 
 ### Declare traits
 
-Create a trait to define a discrete amount of additional functionality, and define it in the [traits](https://docs.swift.org/swiftpm/documentation/packagedescription/package/traits) property of the package manifest.
-Use [`.default(enabledTraits:)`](https://docs.swift.org/swiftpm/documentation/packagedescription/trait/default(enabledtraits:)) to provide the set of traits that the package uses as a default.
+Create a trait to define additional functionality, and define it in the [traits](https://docs.swift.org/swiftpm/documentation/packagedescription/package/traits) property of the package manifest.
+Use [`.default(enabledTraits:)`](https://docs.swift.org/swiftpm/documentation/packagedescription/trait/default(enabledtraits:)) to provide the set of traits that the package uses by default.
 If you don't define a default set of traits to enable, Swift Package Manager enables no traits by default.
 
 The following example illustrates a single trait, `FeatureA`, that is enabled by default:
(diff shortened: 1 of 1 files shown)


heckj:
@swift-ci please test

Sources/PackageManagerDocs/Documentation.docc/Dependencies/PackageTraits.md
```diff
@@ -1,30 +1,32 @@
...
 Swift packages prior to Swift 6.1 offered a non-configurable API surface for each version.
-With Swift 6.1, packages may offer traits, which express a configurable API surface for a package.
+Starting with Swift 6.1, packages may offer traits, which express a configurable API surface for a package.
 
 Use traits to enable additional API beyond the core API of the package.
-For example, a trait may enable an experimental API, optional extended functionality that requires additional dependencies, or functionality that isn't critical that a developer may want to enable only in specific circumstances.
+For example, a trait may enable an experimental API, optional extended functionality that requires additional dependencies, or functionality that isn't critical that you may want to enable only in specific circumstances.
+
+When Swift Package Manager builds with a requested trait, it does so across the package dependency tree, propagating that request for traits throughout the dependencies.
+Traits affect all targets being built in the package as well as any dependencies, not just targets within a package.
```
MaxDesiatov: I don't think this one's true as written. Traits live in per-package namespaces. If root package `Foo` has trait `T` enabled, and wants trait `T` enabled in dependency `Bar`, it has to convert (namespacing explicitly here for illustration purposes) `Foo.T` to `Bar.T` explicitly in `Package.swift`. @FranzBusch, @bripeticca  do I understand this correctly?

Sources/PackageManagerDocs/Documentation.docc/Dependencies/PackageTraits.md
```diff
@@ -1,30 +1,32 @@
...
 
 > Note: Traits should never "remove" or disable public API when a trait is enabled.
 
-Within the package, traits express conditional compilation, and may be used to declare additional dependencies that are enabled when that trait is active.
+Within the package, traits express conditional compilation, and you can use them to declare additional dependencies that are enabled when that trait is active.
+Swift Package Manager exposes enabled traits as conditional blocks (for example, `#if YourTrait`) that you can use to conditionally enable imports or different compilation paths in code.
 
-Traits are identified by their names, which are name-spaced within the package that hosts them.
+Trait names are name-spaced within the package that hosts them. <!-- what's this mean, exactly - what does this enable or impact? pending question to Bri -->
```
MaxDesiatov: Covered in the comment above, it means that trait `T` declared in multiple packages are actually different unrelated traits.

Sources/PackageManagerDocs/Documentation.docc/Dependencies/PackageTraits.md
dschaefer2: I'll let Bri or Franz respond for sure but I believe Max to be correct here. I will comment that Traits are highly misunderstood. We need to be very crisp in the documentation. I'll take a closer look when I get back from my break next week.

heckj: Yeah, that's why I kept this in Draft - I want to make sure this is dead on correct and nailed down.

My confusion came from what you invoke a build with traits, you're giving it a set of strings to represent the traits - if two packages happen to offer traits that use the same name, are both enabled?

MaxDesiatov: > My confusion came from what you invoke a build with traits, you're giving it a set of strings to represent the traits - if two packages happen to offer traits that use the same name, are both enabled?

For two different packages the answer in general is "no" (unless you wire up traits with same names between packages explicitly in `Package.swift`), but in which package namespace they live depends on which is a root and which is a dependency.

The set of strings represents traits in the top-level (root) package namespace. 

Only traits in the root package are enabled by CLI options, the dependency needs them passed down explicitly. If the dependency is transitive, they need to be passed down explicitly multiple times at each dependency level in each `Package.swift`.

dschaefer2: The key thing I keep coming back to is that traits are a way to specify feature optionality for a given package. We don't support features that span multiple packages. As Max mentioned, a package may request a trait to be enabled in the dependency on that package but that's actually unrelated to the traits in that package. (That said, we should be allowing dependencies and their traits setting to be conditional on a package's traits as well).

bripeticca: Essentially what @MaxDesiatov mentioned, specifying traits to enable through the CLI will only affect the enabled traits of the root package. There isn't currently a way to specify enabled traits for dependencies through the CLI, but one can configure enabled traits of a dependency in the root package's manifest.

Sources/PackageManagerDocs/Documentation.docc/Dependencies/PackageTraits.md
```diff
@@ -1,30 +1,32 @@
...
 
-Traits are identified by their names, which are name-spaced within the package that hosts them.
+Trait names are namespaced within the package that hosts them - a trait name in one package has no impact on any other package.
 Trait names must be [valid swift identifiers](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/lexicalstructure#Identifiers) with the addition of the characters of `-` and `+`.
-The trait names `default` and `defaults` (regardless of any capitalization) aren't allowed to avoid confusion with the default traits that a package defines.
-
-Enabled traits are exposed as conditional blocks (for example, `#if YourTrait`) that you can use to conditionally enable imports or different compilation paths in code.
-
-## Overview
+Don't use the trait names `default` or `defaults` (regardless of capitalization) to avoid confusion with the default traits that a package defines.
```
FranzBusch: This sounds less strict than before. We do throw an error if one tries to use those names.

Sources/PackageManagerDocs/Documentation.docc/Dependencies/PackageTraits.md
```diff
@@ -46,18 +48,21 @@ traits: [
...
 ```
 
-The traits enabled by default for the example above is `FeatureA`.
+For the example above, the default trait is `FeatureA`.
 
 > Note: Changing the default set of traits for your package is a major semantic version change if it removes API surface.
 > Adding additional traits is not a major version change.
 
-#### Mutually Exclusive Traits
+Swift Package Manager treats traits as purely additive, and enables traits across all packages within the build graph.
```
FranzBusch:
```suggestion
Swift Package Manager treats traits as purely additive, and unifies enabled traits across all packages within the build graph.
```

heckj:
@swift-ci please test

heckj:

heckj:
@swift-ci please test swiftbuild

PR #9561, opened 2026-01-06
Pin swift-tools-protocol to specific revision

bkhouri:
Pin swift-tools-protocols to a specific revision to avoid issues with upstream changes.


bkhouri:
@swift-ci test

bkhouri:

PR #9602, opened 2026-01-15
Expand prebuilts to plugins, and ensure prebuilts don't leak to cross targets

dschaefer2:
Problems have been occuring when modules that use prebuilts and modules that use those dependencies from source end up linked together. This particularly happens with plugins and macros that share code gen libraries. We also have issues in cross compilation where a library that uses prebuilts ends up also being compiled for a cross target where we don't have prebuilts available.

With this change, we add a decoration to the PackageModelBuilders that detect when a prebuilt is being used modules in the dependency tree of host tools and only host tools. If a macro, macro test, or plugin is trying to use a module that is used by an explicit product in the root package that could be built cross, we disable prebuilts entirely.

It is only after that is determined do we adjust the modules to remove the dependency on the prebuilt source and add in the build settings to use the prebuilt binaries. In SwiftBuild we also set these targets to be HOST_PLATFORM only. Plugins are already marked that way.

This still isn't ideal. We need to figure out how to filter use of prebuilts by platform, i.e. sdk and architecture, and allow modules to be built in both modes in a given build.

### DIFF
--- a/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
@@ -125,9 +125,6 @@ extension PackagePIFProjectBuilder {
 
         // Any dependencies of plugin targets need to be built for the host.
         buildSettings[.SUPPORTED_PLATFORMS] = ["$(HOST_PLATFORM)"]
-        if pluginModule.platformConstraint == .host, let archs = PrebuiltsPlatform.hostPlatform?.arch.archs {
-            buildSettings[.ARCHS] = archs
-        }
 
         self.project[keyPath: pluginTargetKeyPath].common.addBuildConfig { id in
             BuildConfig(id: id, name: "Debug", settings: buildSettings)
@@ -368,9 +365,6 @@ extension PackagePIFProjectBuilder {
 
         if sourceModule.platformConstraint == .host {
             settings[.SUPPORTED_PLATFORMS] = ["$(HOST_PLATFORM)"]
-            if let archs = PrebuiltsPlatform.hostPlatform?.arch.archs {
-                settings[.ARCHS] = archs
-            }
         }
         if shouldGenerateBundleAccessor {
             settings[.GENERATE_RESOURCE_ACCESSORS] = "YES"
--- a/Sources/PackageModel/PrebuiltLibrary.swift
+++ b/Sources/PackageModel/PrebuiltLibrary.swift
@@ -75,7 +75,7 @@ public enum PrebuiltsPlatform: String, Codable, CaseIterable {
     case debian_12_aarch64
     case debian_12_x86_64
 
-    public enum Arch: String {
+    public enum Arch: String, CaseIterable {
         case x86_64
         case aarch64
     }
--- a/Sources/SwiftBuildSupport/PackagePIFBuilder.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFBuilder.swift
@@ -725,14 +725,3 @@ extension PackagePIFBuilder.LinkedPackageBinary {
         }
     }
 }
-
-extension PrebuiltsPlatform.Arch {
-    public var archs: [String] {
-        switch self {
-        case .aarch64:
-            return ["arm64"]
-        case .x86_64:
-            return ["x86_64"]
-        }
-    }
-}
--- a/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
@@ -121,9 +121,6 @@ extension PackagePIFProjectBuilder {
 
         if product.platformConstraint == .host {
             settings[.SUPPORTED_PLATFORMS] = [ "$(HOST_PLATFORM)" ]
-            if let archs = PrebuiltsPlatform.hostPlatform?.arch.archs {
-                settings[.ARCHS] = archs
-            }
         }
 
         if mainModule.type == .test {
@@ -690,9 +687,6 @@ extension PackagePIFProjectBuilder {
 
         if product.platformConstraint == .host {
             settings[.SUPPORTED_PLATFORMS] = ["$(HOST_PLATFORM)"]
-            if let archs = PrebuiltsPlatform.hostPlatform?.arch.archs {
-                settings[.ARCHS] = archs
-            }
         }
 
         // Add other build settings when we're building an actual dylib.
@@ -955,9 +949,6 @@ extension PackagePIFProjectBuilder {
 
         if pluginProduct.platformConstraint == .host {
             buildSettings[.SUPPORTED_PLATFORMS] = ["$(HOST_PLATFORM)"]
-            if let archs = PrebuiltsPlatform.hostPlatform?.arch.archs {
-                buildSettings[.ARCHS] = archs
-            }
         }
 
         self.project[keyPath: pluginTargetKeyPath].common.addBuildConfig { id in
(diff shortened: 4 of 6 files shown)


dschaefer2:
Addresses https://github.com/swiftlang/swift-package-manager/issues/9331 amongst others.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test

Sources/PackageGraph/ModulesGraph+Loading.swift
```diff
@@ -901,9 +882,212 @@ private func createResolvedPackages(
...
+    packageBuilders: [ResolvedPackageBuilder],
+    prebuilts: [PackageIdentity: [String: PrebuiltLibrary]],
+) -> [ResolvedPackageBuilder] {
+    // Skip this if there are no prebuilts. Modules are unconstrained by default.
+    guard !prebuilts.isEmpty else {
+        return []
+    }
+
+    // First decorate the platform constraints from explicit products in the root packages
+    for packageBuilder in packageBuilders where packageBuilder.package.manifest.packageKind.isRoot {
```
owenv, line 903: Intuitively I think the recursion in these traversals is unlikely to overflow the stack, but we may want to add a helper to GraphAlgorithms.swift for iterative breadth first traversal. Doesn't need to be in this PR IMO

Sources/PackageGraph/ModulesGraph+Loading.swift
```diff
@@ -901,9 +882,212 @@ private func createResolvedPackages(
...
+                        return isHostOnly(childModule)
+                    case .product(let childProduct, conditions: _):
+                        return childProduct.moduleBuilders.allSatisfy {
+                            isHostOnly($0)
+                        }
+                    }
+                }
+            }
+
+            return isHostOnly(moduleBuilder)
```
owenv: It looks like isHostOnly here will traverse the entire graph reachable from the starting node, and it's being called for every node, which has the potential to blow up if the results aren't memoized

Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
```diff
@@ -635,6 +639,18 @@ extension PackageGraph.ResolvedModule {
...
+                    singleValueSetting = nil
+                    multipleValueSetting = .OTHER_SWIFT_FLAGS
+                    values = settingAssignment.values.flatMap { ["-I", $0] }
+                case .PREBUILT_LIBRARY_PATHS:
+                    singleValueSetting = nil
+                    multipleValueSetting = .LIBRARY_SEARCH_PATHS
+                    values = settingAssignment.values
+                case .PREBUILT_LIBRARIES:
+                    singleValueSetting = nil
+                    multipleValueSetting = .OTHER_LDFLAGS
```
owenv, line 652: Ideally I think we'd introduce a new setting other than OTHER_LDFLAGS, e.g. ADDITIONAL_LIBRARIES, which preserves meaning of the flags and can potentially do some deduplication. This is probably ok for now as long as it's not overriding other assignments to OTHER_LDFLAGS in the PIF

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
```diff
@@ -362,6 +366,12 @@ extension PackagePIFProjectBuilder {
         // Configure the target-wide build settings. The details depend on the kind of product we're building.
         var settings: BuildSettings = self.package.underlying.packageBaseBuildSettings
 
+        if sourceModule.platformConstraint == .host {
+            settings[.SUPPORTED_PLATFORMS] = ["$(HOST_PLATFORM)"]
```
owenv, line 367: Macros should get this from the product type, did you need to add this for plugin tools? If so, we should look at sinking it into swift-build

owenv: It's also repeated in a few places, would be nice to see if there's a good way to refactor this to avoid that

Sources/PackageGraph/ModulesGraph+Loading.swift
dschaefer2: Sounds good.

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
dschaefer2: I'll double check but yeah, the plugin tools ended up getting built for the run destination as well. It could be because they are getting sucked into the All* targets. If I pull them out of that would it may be cleaner. I'll check that as well.

Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
dschaefer2: Agreed. We're using OTHER_LDFLAGS in a few places. That should all be cleanedup. I tried LINK_LIBRARIES hoping to get deduplication but that didn't seem to happen.

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
dschaefer2: In the end, we'll need to recognize when a prebuilt is available for a given platform and arch build of a target and use it. And when there isn't, build that dependency from source as we normally do. It's really supposed to be a target level cache. As we generalize prebuilts, we should rethink how we do this to be more like a cache.

Sources/PackageGraph/ModulesGraph+Loading.swift
dschaefer2: Part of the reason to put this code here is that it's done after the cycle checker to make sure we don't get stuck in an infinite loop walking the graph. It's interesting that `findCycle` in GraphAlgorithms is also a DFS with an FIXME to use a stack. We should definitely solve both of these at the same time.

Sources/PackageGraph/ModulesGraph+Loading.swift
dschaefer2: Yes, I definitely want to bring this closer to O(n).

dschaefer2: Actually the guard `moduleBuilder.isHostOnly` ensures we're only checking the dependencies of macros, macro tests, and plugins. It had me confused so it's probably a bad name since the function `isHostOnly` is checking the transitive dependencies of those which is different. That makes it much less than O(n2) but still not super optimal. 

I'll give it some thought. There's definitely potential performance issues here. At least everything is a reference type in the Builders graph which makes traversing it somewhat cheap.

dschaefer2: OK, was able to work this into the colouring algorithm. Should be a bit more efficient. Update coming.

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
dschaefer2: For now though, I needed to add this to all the libraries used by plugins and macros that use prebuilts since they are host only and only for the arch of the host as well (i.e. not universal). I'll take a look at that and see if I can make them universal and see if that helps.

dschaefer2: OK, was able to produce universal binary prebuilts for macOS so we don't need the ARCHS build setting. We still need HOST_PLATFORM set on targets that use the prebuilts to ensure they don't try to build against them when building for cross.

PR #9608, opened 2026-01-16
Reuse Collections utilities from SWBUtil instead of duplicating the c…

pmattos:
### Motivation:

Reduce code duplication in the PIF builder (within the `SwiftBuildSupport` module).

### Modifications:

Some `Collection` *extensions* in `SwiftBuildSupport/PackagePIFBuilder+Helpers.swift` were also available in `SWBUtil` too (from Swift Build) — so let’s use that instead! Some provided even better/faster implementation :)

### Result:

Less code duplication in the `SwiftBuildSupport` module.

Fixes rdar://168274021.

### DIFF
--- a/Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
@@ -58,15 +58,12 @@ import func PackageLoading.pkgConfigArgs
 import SPMBuildCore
 
 import enum SwiftBuild.ProjectModel
+import SWBUtil
 
 // MARK: - PIF GUID Helpers
 
 public enum TargetSuffix: String {
     case testable, dynamic
-
-    func hasSuffix(id: GUID) -> Bool {
-        id.value.hasSuffix("-\(self.rawValue)")
-    }
 }
 
 extension TargetSuffix? {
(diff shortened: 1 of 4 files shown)


pmattos:
@swift-ci test

Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
```diff
@@ -1226,25 +1224,9 @@ extension Collection {
         !self.isEmpty
     }
```
pmattos, line 1226: These 3 helpers are now coming from SWBUtil.

Sources/SwiftBuildSupport/PackagePIFBuilder+Helpers.swift
```diff
@@ -58,15 +58,12 @@ import func PackageLoading.pkgConfigArgs
 import SPMBuildCore
 
 import enum SwiftBuild.ProjectModel
+import SWBUtil
```
jakepetroules, line 61: We have traditionally not allowed use of this library outside of Swift Build as it's not part of the API surface. Given the special relationship between SB and SPM I wonder if it's time to relax that a bit _for SPM only_. In theory it might cause some problems though. @owenv what do you think?

pmattos: Ooops, I wasn't aware of that constraint! Let me check with @owenv then...

owenv: I think we should continue to avoid exposing SWBUtil to clients if we can, as it's intentionally been kept an implementation detail of the SwiftBuild module in the past so it wasn't exposed as ABI. If there's a strong need, maybe we can reexport the parts that are needed via the SwiftBuild target?

Issue #9603, opened 2026-01-15
spm permits invalid package names


3405691582:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

```
$ swift package init --name 1
```

This generates a test that contains `@testable import _`, which is invalid, a target in Package.swift named "1" but cannot be imported since it is also invalid. This causes `swift build` to fail.

### Expected behavior

spm should enforce name correctness at package initialization time.

### Actual behavior

spm generates targets and tests with invalid identifiers.

### Steps to reproduce

swift package init --name 1
swift test

### Swift & OS version (output of `swift --version ; uname -a`)

Verified this on Linux with Swift 6.2.3, as well as OpenBSD with Swift 6.2.0.

PR #9610, opened 2026-01-16
Fix invalid module names when package names mangle to single underscore

plemarquand:
### Motivation:

Previously, InitPackage.swift used the original package name for both the package name and target names. This caused build failures when the original name would be mangled to "_" by the Swift compiler, as "_" is not a valid Swift module or struct identifier.

### Modifications:

Use the moduleName for targets and dependencies, and if it mangles to "_", replace it with "__" to ensure that targets and structs still have valid Swift identifiers.

Issue: #9603


plemarquand:
@swift-ci test

PR #9611, opened 2026-01-17
Fix test failures in URLSessionHTTPClientTests

rconnell9:
The MockURLProtocol makes use of static dictionaries for expected
observers. The keys are composed of url and method. Multiple tests
were using the same imaginary URL, which was causing collisions.

Update the tests so that they use more unique URLs to avoid
collisions.


rconnell9:
@swift-ci test

PR #9298, opened 2025-10-28
Add Git LFS support to Swift Package Manager

wcharysz:
**Motivation:**

Swift Package Manager currently lacks native support for Git Large File Storage (LFS), which limits its usability in projects that rely on large binary assets or versioned media files. This change aims to improve compatibility with repositories using Git LFS, enabling smoother dependency resolution and package integration.

**Modifications:**

Integrated Git LFS handling into the package resolution workflow.
Ensured LFS-tracked files are properly fetched and available during package checkout.
Added tests to validate LFS file presence and integrity.

**Result:**

SwiftPM can now correctly handle packages that use Git LFS, improving support for a broader range of repositories and workflows involving large assets.

PR #9625, opened 2026-01-21
Add automatic Git LFS support

### LINKED ISSUE #8233: Swift package manager is not working with 'git lfs' in Xcode.

robmaceachern:
> [!NOTE]
> Disclaimer: This was primarily implemented with Claude Code but I iterated on it to reach its current state. The original plan/prompt doc is below, though the implementation in this branch has diverged from that plan.

[GIT_LFS_IMPLEMENTATION_PLAN.md](https://github.com/user-attachments/files/24771943/GIT_LFS_IMPLEMENTATION_PLAN.md)

## Summary

Resolves #8233: SPM fails to resolve packages using Git LFS.

**Root cause:** SwiftPM caches source dependencies as bare mirrors. LFS smudge filters do not run in those bare repositories, so working copies can end up with LFS pointers or missing objects unless the cache and checkout paths handle LFS explicitly.

**This branch now:**
1. Detects LFS usage by checking `.gitattributes` for `filter=lfs`.
2. Runs `git lfs fetch` after mirror clone/fetch to hydrate the bare repository cache.
3. Runs `git lfs pull` from `checkout(tag:)` / `checkout(revision:)` to materialize LFS files in working copies.
4. Repairs the upgrade path for existing shared checkouts by hydrating the alternate object-store bare cache and retrying the pull when needed.
5. Respects `GIT_LFS_SKIP_SMUDGE`.
6. Preserves `CancellationError` and narrows the fallback behavior around LFS detection.

## Changes

| File | Changes |
|------|---------|
| `Sources/SourceControl/GitRepository.swift` | Add LFS helpers, cache the git-lfs probe, move LFS materialization into checkout, serialize bare-cache LFS fetch under the repo lock, repair the upgrade path for existing shared checkouts, and tighten cancellation/error handling |
| `Tests/SourceControlTests/GitRepositoryTests.swift` | Replace the vacuous provider test with a real LFS-backed setup and add regression coverage for provider cache hydration, upgrade-path repair, `GIT_LFS_SKIP_SMUDGE`, cache invalidation across revisions, and end-to-end fetch/pull |

### Auto-detect vs Opt-in Flag

This PR uses **auto-detection** rather than requiring a flag like `--enable-experimental-git-lfs`:

- **Zero configuration:** Users with git-lfs installed get LFS support automatically.
- **Clear errors:** If git-lfs is not installed but required, users get actionable install instructions.
- **No breaking changes:** Repositories without LFS are unaffected.

### Why `git lfs fetch` Does NOT Use `--all`

Using `git lfs fetch --all` fetches LFS objects for all refs and commits in history. That can fail with 404s when historical LFS objects have been pruned from the server, which is common for large repositories.

By omitting `--all`, SwiftPM only fetches objects needed for the current/default ref, which is sufficient for checkout and avoids failures on repositories with pruned LFS history.

```swift
// DO NOT add --all here - causes 404 errors on repos with pruned LFS history
return try self.run(["-C", path.pathString, "lfs", "fetch"])
```

### Upgrade Path

Existing caches/checkouts do **not** require a full purge/re-clone. On the next `checkout(tag:)` / `checkout(revision:)`, SwiftPM now attempts LFS materialization for the working copy. If the working copy is backed by an older alternate object-store bare cache that never fetched LFS objects, SwiftPM hydrates that bare cache first and then retries the pull.

### Setup

```bash
# Create test environment
mkdir -p /tmp/lfs-test-env && cd /tmp/lfs-test-env

# Create LFS-enabled package
mkdir lfs-package && cd lfs-package
git init
# If needed on a clean machine:
# git config user.name "Your Name"
# git config user.email "you@example.com"
git lfs install --local
echo '*.bin filter=lfs diff=lfs merge=lfs -text' > .gitattributes
cat > Package.swift << 'PKGEOF'
// swift-tools-version:5.9
import PackageDescription
let package = Package(name: "LFSPackage", targets: [.target(name: "LFSPackage")])
PKGEOF
mkdir -p Sources/LFSPackage && echo "public struct LFSPackage {}" > Sources/LFSPackage/LFSPackage.swift
head -c 1024 /dev/urandom > data.bin
git add . && git commit -m "Initial commit with LFS"
git tag 1.0.0
cd ..

# Create bare repo (simulates remote)
git clone --bare lfs-package lfs-package.git

# Push LFS objects into the bare repo as well
cd lfs-package
git remote add bare /tmp/lfs-test-env/lfs-package.git
git lfs push --all bare
cd ..

# Create consumer package
mkdir consumer && cd consumer
cat > Package.swift << 'PKGEOF'
// swift-tools-version:5.9
import PackageDescription
let package = Package(
    name: "Consumer",
    dependencies: [.package(url: "file:///tmp/lfs-test-env/lfs-package.git", from: "1.0.0")],
    targets: [.executableTarget(name: "Consumer", dependencies: ["LFSPackage"])]
)
PKGEOF
mkdir -p Sources/Consumer && echo "import LFSPackage; print(\"Hello\")" > Sources/Consumer/main.swift
```

### Test: LFS files are fetched automatically

```bash
cd /tmp/lfs-test-env/consumer
rm -rf .build Package.resolved
swift package resolve
```

**Expected:** Success. Verify the LFS file is actual binary data:

```bash
file .build/checkouts/lfs-package/data.bin
# Should output: data (not "ASCII text" which would indicate an LFS pointer)
```

## Test Coverage

The PR now includes coverage for:

- LFS detection via `.gitattributes`
- Detection caching and cache invalidation across revisions
- Error message formatting
- Provider/bare-cache LFS hydration with the original source repo removed
- Upgrade-path repair for pre-LFS shared checkouts
- `GIT_LFS_SKIP_SMUDGE`
- End-to-end `fetch` + `checkout` / `pull` behavior

The LFS integration tests require `git-lfs` to be accessible to SwiftPM's git binary and skip otherwise.

## Validation

Locally revalidated with:

- `swift test --filter GitRepositoryTests`
- the manual testing flow above (with the corrected local bare-remote LFS push step)

## Related

- Issue: #8233
- Related PR: #8235 (similar approach, used opt-in flag)
- Related PR: #9298 (automatic integration, uses `--all`)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -834,6 +834,8 @@ public final class GitRepository: Repository, WorkingCheckout {
                 let alternateRepository = GitRepository(git: self.git, path: alternateRepositoryPath, isWorkingRepo: false)
                 try alternateRepository.fetchLFSIfNecessary()
                 try self.git.pullLFS(at: self.path)
+            } catch let error as CancellationError {
+                throw error
             } catch {
                 throw GitLFSError.pullFailed(underlyingError: error)
             }


Sources/SourceControl/GitRepository.swift
```diff
@@ -677,11 +746,68 @@ public final class GitRepository: Repository, WorkingCheckout {
...
+
+        if try self.hasLFSTrackedFiles() {
+            do {
+                try self.git.pullLFS(at: self.path)
+            } catch let error as GitLFSError {
+                throw error
+            } catch {
+                throw GitLFSError.pullFailed(underlyingError: error)
+            }
+        }
```
robmaceachern, line 842: I am wondering if it would make sense to disable the pull if desired. Either as explicit flag or respecting `ProcessEnv.vars["GIT_LFS_SKIP_SMUDGE"]`.

robmaceachern:
@plemarquand Do you think there's any interest in iterating on this?

Sources/SourceControl/GitRepository.swift
```diff
@@ -78,6 +78,48 @@ private struct GitShellHelper {
...
         }
     }
+
+    // MARK: - LFS Support
+
+    /// Checks if git-lfs is installed
+    /// - Returns: true if git-lfs is available
+    func isGitLFSInstalled() -> Bool {
+        do {
+            _ = try self.run(["lfs", "version"])
```
plemarquand: if `self.run` throws a `CancellationError` we should rethrow it instead of returning false.

Sources/SourceControl/GitRepository.swift
```diff
@@ -201,6 +243,18 @@ public struct GitRepositoryProvider: RepositoryProvider, Cancellable {
             ["--mirror"],
             progress: progressHandler
         )
+
+        // Auto-detect and fetch LFS objects if repository uses LFS
+        let repo = GitRepository(git: self.git, path: path, isWorkingRepo: false)
+        if try repo.hasLFSTrackedFiles() {
+            do {
+                try self.git.fetchLFS(at: path)
```
plemarquand: Same here, we should do a straight rethrow of CancellationError as well

Sources/SourceControl/GitRepository.swift
```diff
@@ -677,11 +746,68 @@ public final class GitRepository: Repository, WorkingCheckout {
...
+                guard !trimmed.isEmpty, !trimmed.hasPrefix("#") else {
+                    return false
+                }
+                // Gitattributes format: pattern attr1 attr2 ...
+                // Check for filter=lfs as a distinct whitespace-separated token
+                return trimmed.split(whereSeparator: \.isWhitespace).contains { $0 == "filter=lfs" }
+            }
+        } catch {
+            // .gitattributes doesn't exist or can't be read - no LFS
+            hasLFS = false
```
plemarquand: Same here, we should do a straight rethrow of CancellationError as well. Actually we should rethrow everything except `GitRepositoryError` I think; that would be the only case where `hasLFS` is false

Sources/SourceControl/GitRepository.swift
```diff
@@ -677,11 +746,68 @@ public final class GitRepository: Repository, WorkingCheckout {
         return try !self.isBare()
     }
 
+    /// Checks if the repository uses Git LFS by examining .gitattributes
+    /// - Returns: true if the repository has LFS-tracked files
+    func hasLFSTrackedFiles() throws -> Bool {
+        if let cached = cachedHasLFS.get() {
```
plemarquand: This should use the same `.memoize` pattern as `readTree`, `readBlob`, `open`, etc... Using `get/put` creates a race condition where `hasLFSTrackedFiles` can be called multiple times before `put` sets the value.

Sources/SourceControl/GitRepository.swift
```diff
@@ -677,11 +746,68 @@ public final class GitRepository: Repository, WorkingCheckout {
...
+        return hasLFS
+    }
+
+    /// Clears the cached LFS detection result
+    func clearLFSCache() {
+        cachedHasLFS.clear()
+    }
+
     private func updateSubmoduleAndCleanIfNecessary() throws {
         if self.cachedHasSubmodules.get(default: false) || localFileSystem.exists(self.path.appending(".gitmodules")) {
```
plemarquand: I know this isn't from your patch but could you fix this up an use `.memoize` here as well?

Sources/SourceControl/GitRepository.swift
```diff
@@ -572,6 +627,20 @@ public final class GitRepository: Repository, WorkingCheckout {
                 progress: progress
             )
             self.cachedTags.clear()
+            self.cachedHasLFS.clear()  // Clear LFS cache on fetch
+        }
+
+        // Auto-detect and fetch LFS objects (for bare repos)
+        if !self.isWorkingRepo {
```
plemarquand: I think this should happen within the `withLock` above. As unlikely it is that fetch will be called twice from two different threads we still want to guard against pulling LFS files multiple times at the same time.

Sources/SourceControl/GitRepository.swift
plemarquand: I think we should respect `GIT_LFS_SKIP_SMUDGE`. I would expect setting that to disable LFS operations.

Sources/SourceControl/GitRepository.swift
plemarquand: Should we be caching this result? This result is always going to be the same for the duration of the package resolution, and shelling out to git is expensive.

Tests/SourceControlTests/GitRepositoryTests.swift
```diff
@@ -930,4 +932,224 @@ class GitRepositoryTests: XCTestCase {
...
+        let fetchError = GitLFSError.fetchFailed(underlyingError: NSError(domain: "test", code: 1, userInfo: [NSLocalizedDescriptionKey: "fetch failed"]))
+        XCTAssertTrue(fetchError.description.contains("Git LFS fetch failed"))
+
+        // Test pullFailed error
+        let pullError = GitLFSError.pullFailed(underlyingError: NSError(domain: "test", code: 1, userInfo: [NSLocalizedDescriptionKey: "pull failed"]))
+        XCTAssertTrue(pullError.description.contains("Git LFS pull failed"))
+    }
+
+    /// Test GitRepositoryProvider with auto-detected LFS
+    func testGitRepositoryProviderWithLFS() async throws {
```
plemarquand, line 1124: How does this test work? There is no .gitattributes file generated w/ lfs support, and from what I can tell there isn't another way to enable the support.

Tests/SourceControlTests/GitRepositoryTests.swift
```diff
@@ -930,4 +932,224 @@ class GitRepositoryTests: XCTestCase {
             XCTAssertFalse(try repositoryManager.isValidDirectory(packageDir, for: RepositorySpecifier(url: SourceControlURL(packageDir.pathString.appending(".git")))))
         }
     }
+
+    // MARK: - Git LFS Tests
```
plemarquand, line 1017: How do we handle the update path, where users have a repo checked out already, they update SPM to this version with LFS support. Are LFS files now pulled? Do users need to fully purge their cache and re-clone?

Sources/SourceControl/GitRepository.swift
```diff
@@ -1181,6 +1307,34 @@ private enum GitInterfaceError: Swift.Error {
     case fatalError
 }
 
+/// Errors related to Git LFS operations
+public enum GitLFSError: Error, CustomStringConvertible {
```
plemarquand: nit: This can be `package` instead of `public`

Sources/SourceControl/GitRepository.swift
```diff
@@ -677,11 +746,68 @@ public final class GitRepository: Repository, WorkingCheckout {
...
+    }
+
     private func updateSubmoduleAndCleanIfNecessary() throws {
         if self.cachedHasSubmodules.get(default: false) || localFileSystem.exists(self.path.appending(".gitmodules")) {
             self.cachedHasSubmodules.put(true)
             try self.updateSubmoduleAndCleanNotOnQueue()
         }
+
+        // Pull LFS files after submodule update
+        try self.pullLFSIfNecessary()
```
plemarquand: Can you move this to `checkout(tag:)` and `checkout(revision:)`? LFS is unrelated to submodules and cleaning, so I wouldn't expect this method to pull LFS files

plemarquand:
@swift-ci test

Sources/SourceControl/GitRepository.swift
robmaceachern: Added `GIT_LFS_SKIP_SMUDGE` handling in GitShellHelper, so both `git lfs fetch` and `git lfs pull` now short-circuit when that env var is set. I kept it centralized there so the skip applies consistently to both the bare-cache and working-copy paths.

Sources/SourceControl/GitRepository.swift
```diff
@@ -677,13 +768,78 @@ public final class GitRepository: Repository, WorkingCheckout {
...
+                guard case .pullFailed = error,
+                      let alternateRepositoryPath = self.alternateObjectStoreRepositoryPath()
+                else {
+                    throw error
+                }
+
+                let alternateRepository = GitRepository(git: self.git, path: alternateRepositoryPath, isWorkingRepo: false)
+                try alternateRepository.fetchLFSIfNecessary()
+                try self.git.pullLFS(at: self.path)
+            } catch {
```
plemarquand, line 839: Catch and rethrow CancellationError here instead of wrapping it.

Sources/SourceControl/GitRepository.swift
robmaceachern: Fixed. `run()` now rethrows `CancellationError`, and the git-lfs probe was changed to a throwing, memoized helper so cancellation is no longer collapsed into `false`.

Sources/SourceControl/GitRepository.swift
robmaceachern: Should be fixed now.

Tests/SourceControlTests/GitRepositoryTests.swift
robmaceachern: I updated the checkout path so existing pre-LFS caches/checkouts are repaired on the next checkout rather than requiring a full purge/re-clone. `checkout(tag:)` / `checkout(revision:)` now run LFS materialization, and if `git lfs pull` fails because the checkout is backed by an older alternate object-store bare cache that never fetched LFS objects, SwiftPM hydrates that bare cache first and retries the pull.

`testGitLFSUpgradePathPullsObjectsIntoExistingCheckout()` covers that upgrade path explicitly. It creates the pre-LFS state by cloning a bare mirror, creating a shared checkout from it, and checking out the tag with `GIT_LFS_SKIP_SMUDGE=1` so the working copy contains an LFS pointer. It then calls `workingCopy.checkout(revision:)` using the new code and verifies that the same checkout is repaired in place to the real binary data.

At least that's the plan haha.

Tests/SourceControlTests/GitRepositoryTests.swift
robmaceachern: Yeah I think the previous test was no working as expected. I replaced it with an actual LFS-backed setup: it creates a repo with LFS tracking enabled, fetches through the provider into the bare cache, deletes the original source repo, then creates a working copy and verifies the binary still materializes from the cache.

Sources/SourceControl/GitRepository.swift
robmaceachern: Fixed in https://github.com/swiftlang/swift-package-manager/pull/9625/changes/e88329d44f835d07fcd0cc76ecf51b4bb9f3607f

Sources/SourceControl/GitRepository.swift
robmaceachern: Addressed in [9676fc8fb](https://github.com/swiftlang/swift-package-manager/pull/9625/commits/9676fc8fb3c1355e42ec124dd406679f26d379c7)

Sources/SourceControl/GitRepository.swift
robmaceachern: Addressed in [9676fc8fb](https://github.com/swiftlang/swift-package-manager/pull/9625/commits/9676fc8fb3c1355e42ec124dd406679f26d379c7): cachedHasLFS now uses `ThreadSafeBox.memoize`, so the detection path no longer has the `get`/`put` race.

Sources/SourceControl/GitRepository.swift
robmaceachern: Addressed in [9676fc8fb](https://github.com/swiftlang/swift-package-manager/pull/9625/commits/9676fc8fb3c1355e42ec124dd406679f26d379c7): I switched the submodule cache to `memoizeOptional` so it follows the same thread-safe memoization pattern.

Double check this one though.

Sources/SourceControl/GitRepository.swift
robmaceachern: Addressed in [9676fc8fb](https://github.com/swiftlang/swift-package-manager/pull/9625/commits/9676fc8fb3c1355e42ec124dd406679f26d379c7): the bare-repo LFS fetch now runs inside the existing `fetch()` lock by calling `fetchLFSIfNecessaryWithoutLock()` from within `withLock`.

Sources/SourceControl/GitRepository.swift
robmaceachern: Addressed in [9676fc8fb](https://github.com/swiftlang/swift-package-manager/pull/9625/commits/9676fc8fb3c1355e42ec124dd406679f26d379c7): `isGitLFSInstalled()` is now memoized, so we only shell out to `git lfs version` once per helper instance.

Sources/SourceControl/GitRepository.swift
robmaceachern: Addressed in [9676fc8fb](https://github.com/swiftlang/swift-package-manager/pull/9625/commits/9676fc8fb3c1355e42ec124dd406679f26d379c7): `GitLFSError` is now `package` instead of `public`

Sources/SourceControl/GitRepository.swift
robmaceachern: Addressed in [9676fc8fb](https://github.com/swiftlang/swift-package-manager/pull/9625/commits/9676fc8fb3c1355e42ec124dd406679f26d379c7): LFS materialization was moved out of the submodule/clean helper and now runs from `checkout(tag:)` / `checkout(revision:)`.

Tests/SourceControlTests/GitRepositoryTests.swift
```diff
@@ -33,6 +35,87 @@ class GitRepositoryTests: XCTestCase {
...
+        environment["GIT_ALLOW_PROTOCOL"] = "file"
+        Git.environmentBlock = environment
+
+        do {
+            try await AsyncProcess.checkNonZeroExit(
+                args: Git.tool, "lfs", "version",
+                environment: .init(Git.environmentBlock)
+            )
+        } catch {
+            throw XCTSkip("git-lfs not accessible to SPM's git binary, skipping test")
```
robmaceachern, line 50: This makes me nervous but maybe it is the right way to approach this type of test?

robmaceachern:
Thanks for all the great feedback @plemarquand! It should all be addressed now but lmk if there are any other tweaks you'd like to see.

plemarquand:
@swift-ci test

Tests/SourceControlTests/GitRepositoryTests.swift
plemarquand: @shahmishal is git-lfs installed in CI? If not is it possible to install it on one of the node types?

plemarquand:
@swift-ci test windows

robmaceachern:
My pleasure! I don't have access to merge this myself so I'll leave it in your hands @plemarquand

PR #9631, opened 2026-01-22
[6.3] Set isDevelopment flag to false in order to drop the trailing -dev from the version output

rconnell9:
[6.3] Set isDevelopment flag to false in order to drop the trailing -dev from the version output


rconnell9:
@swift-ci test

PR #9634, opened 2026-01-22
Fix race condition in PIFBuilderTests when running tests concurrently

pmattos:
### Motivation:
Fix a race condition when running PIF tests in parallel. The crash was caused by a *concurrent* mutation in a shared, global `Dictionary` in `PackageModel.SwiftSDK`:

```swift
private static var _sdkPlatformFrameworkPath: [DarwinPlatform: PlatformPaths] = [:]
```

### Modifications:

Replace the plain `Dictionary` above with our safer `ThreadSafeKeyValueStore` instead. This seemed as a more sensible alternative than promoting the whole `SwiftSDK` class to a (singleton) `actor`, resulting in clients having to handle an async APIs, etc. 

This PR also reduces the logging noise in `PIFBuilderTests.swift`.

### Result:

Tests can now safely run concurrently providing a nice 5x speed up for execution time. This works fine now:

```
$ swift test --filter SwiftBuildSupport
Test run started.
…
Test run with 21 tests in 6 suites passed after 4.5 seconds.
```

…and works okay in Xcode too ;)

Fixes rdar://168686020.

### DIFF
--- a/Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
+++ b/Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
@@ -669,45 +669,45 @@ public struct SwiftSDK: Equatable {
         for darwinPlatform: DarwinPlatform,
         environment: Environment = .current
     ) throws -> PlatformPaths {
-        if let path = _sdkPlatformFrameworkPath[darwinPlatform] {
-            return path
+        let sdkPlatformFrameworkPath = try _sdkPlatformFrameworkPathCache.memoize(darwinPlatform) {
+            // Compute the platform path.
+            let platformPath = try environment[
+                EnvironmentKey("SWIFTPM_PLATFORM_PATH_\(darwinPlatform.xcrunName)")
+            ] ?? AsyncProcess.checkNonZeroExit(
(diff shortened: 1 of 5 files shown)


Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
```diff
@@ -669,45 +669,45 @@ public struct SwiftSDK: Equatable {
...
+                components: "Developer", "Library", "PrivateFrameworks"
+            )
 
-        let sdkPlatformFrameworkPath = PlatformPaths(
-            buildTimeFrameworkSearchPaths: [frameworksPath /* omit privateFrameworksPath */],
-            buildTimeLibrarySearchPaths: [librariesPath],
-            runtimeFrameworkSearchPaths: [frameworksPath, privateFrameworksPath],
-            runtimeLibrarySearchPaths: [librariesPath]
-        )
-        _sdkPlatformFrameworkPath[darwinPlatform] = sdkPlatformFrameworkPath
```
pmattos, line 705: This *mutation* was the crash source.

plemarquand:
@swift-ci test

pmattos:
@swift-ci test macOS

pmattos:
@swift-ci test Linux

jakepetroules (approved):

PR #9635, opened 2026-01-22
[6.3] Fix pif linked package metadata 6.3

plemarquand:
### Motivation:

When using the new PIF builder (i.e., `--build-system swiftbuild`), the returned `[PackagePIFBuilder.LinkedPackageBinary]` objects were lacking package dependencies information. 

For instance, given this simple sample package:
```swift
Package(
    name: "TestLinking",
    products: [
        .library(
            name: "TestLinking",
            type: .static,
            targets: ["TestLinking"]
        )
    ],
    dependencies: [
        .package(url: "https://github.com/apple/swift-collections.git", branch: "1.1.1")
    ],
    targets: [
        .target(
            name: "TestLinking",
            dependencies: [
                .product(name: "Collections", package: "swift-collections")
            ]
        )
    ]
)
```

This would be the *expected* build graph (by traversing the returned `[PackagePIFBuilder.ModuleOrProduct]` array):

```
TestLinking 
  TestLinking 
    TestLinking.swift 
    Collections 
      Collections
        _RopeModule
          InternalCollectionsUtilities 
        OrderedCollections 
          InternalCollectionsUtilities 
        HeapModule
          InternalCollectionsUtilities 
        HashTreeCollections 
          InternalCollectionsUtilities 
        DequeModule 
          InternalCollectionsUtilities 
        BitCollections 
          InternalCollectionsUtilities
```

...but this is what we would be getting instead (ie, note that all `swift-collections` dependencies are missing from the graph):

```
TestLinking
  TestLinking
    TestLinking.swift
```

#### Why this issue wasn't breaking the package build too?
Good question! The sample package built just fine with the new PIF builder (i.e., `--build-system swiftbuild`) even though `PackagePIFBuilder.LinkedPackageBinary` encoded the wrong information, as outlined above. This happened because the build is driven *exclusively* by the PIF, which is correct.

The broken `PackagePIFBuilder.LinkedPackageBinary` objects are only a side effect of computing the PIF and don't effect the build at all. The bug would only manifest in clients who called `PackagePIFBuilder` directly and then inspected the resulting build graph from such auxiliary metadata (e.g., Xcode).

### Modifications:

The *root cause* was how we initialized `LinkedPackageBinary` objects during the PIF construction. For a given package *dependency*, we were incorrectly reusing the *client* package as the package name (i.e., it should be the package name where the dependency is coming *from* instead).

This PR also replaces `packageName: String` with `PackageIdentity`  for tracking packages in `LinkedPackageBinary` objects, ensuring unique IDs for packages (and exposing a more typesafe API too).

### Result:

This fixes rdar://154408350 (*New SwiftPM PIF packages builder breaks previews in Xcode*).

Cherry-pick of https://github.com/swiftlang/swift-package-manager/pull/9436


seavmeng776644-crypto:
iOS 16

PR #9637, opened 2026-01-23
Partially revert #9480 (Use the correct module name for executable targets combined with an executable product with a different name)

plemarquand:
If a package has two products that depend on the same main module target, the PIF builder will use the same name for both executable product artifacts and cause multiple products with the same name to be emitted.

Add a test to verify this regression, and go back to using the product name instead of mainModule name.


plemarquand:
@swift-ci test

PR #9646, opened 2026-01-26
Include a note after Swift Testing tests finish if any XCTests failed

stmontgomery:
This adds a note in the console output after Swift Testing tests finish if one or more XCTests failed, so the user knows to look earlier in the log output for those details.

Fixes rdar://168311253

### Motivation:

Swift Package Manager runs XCTests followed by Swift Testing tests, and they are separate subprocess invocations. Each test framework prints its own output to the console and each has a summary at the end, which includes the total number of failures. Since Swift Testing finishes second (assuming both frameworks are enabled), its summary always appears at the bottom and it can potentially be misleading to a user if there were one or more XCTest failures but zero Swift Testing failures because they may incorrectly believe zero tests failed overall.

Long-term, we are planning to unify the console output between these testing frameworks so that they present a consolidated summary. However, that remains an ambition that is still being planned and will likely take time to arrive.

Another motivator for this is that soon, I anticipate we'll land an enhancement in Swift Testing which will expand its failure summary to span more lines and make it easier for users to locate which Swift Testing tests failed. (For details, see https://github.com/swiftlang/swift-testing/pull/1420.) This will be helpful, but also has the potential to exacerbate the pre-existing confusion situation if (say) a mixture of Swift Testing and XCTests failed.

### Modifications:

Modify the `swift test` command such that if Swift Testing is enabled, it will keep track of whether any XCTests failed before running Swift Testing, and if any did, emit a note at the very end (after Swift Testing finishes) indicating that to the user.

### Result:

When relevant, the new note described above will be included in the console output.

### DIFF
--- a/Sources/Commands/SwiftTestCommand.swift
+++ b/Sources/Commands/SwiftTestCommand.swift
@@ -392,7 +392,9 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
             // emit a message informing the user so they aren't misled and know to look elsewhere for
             // those details.
             if anyXCTestFailed {
-                swiftCommandState.observabilityScope.print(Self.xctestFailedNote, verbose: false)
+                // In theory this could, or should, use `observabilityScope.print(_:verbose:)`,
+                // but that causes tests which check for this output to fail in CI.
+                print(Self.xctestFailedNote)
             }
         }
 
--- a/Tests/CommandsTests/TestCommandTests.swift
+++ b/Tests/CommandsTests/TestCommandTests.swift
@@ -834,7 +834,7 @@ struct TestCommandTests {
                 buildSystem: .native,
                 throwIfCommandFails: false,
             )
-            #expect(stderr.contains(SwiftTestCommand.xctestFailedNote) == arg.expectNote, "stdout: \(stdout), stderr: \(stderr)")
+            #expect(stdout.contains(SwiftTestCommand.xctestFailedNote) == arg.expectNote, "stdout: \(stdout), stderr: \(stderr)")
         }
     }
 


stmontgomery:
@swift-ci test

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -377,6 +380,13 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
                     debug: "Skipping automatic Swift Testing invocation because a test entry point path is present: \(testEntryPointPath)"
                 )
             }
+
+            // After running Swift Testing tests, if we determined that any XCTests failed earlier,
+            // emit a message informing the user so they aren't misled and know to look elsewhere for
+            // those details.
+            if anyXCTestFailed {
+                print("Note: One or more XCTests failed, see logging above for details.")
```
plemarquand: This should use `swiftCommandState.observabilityScope` over `print`

Tests/CommandsTests/TestCommandTests.swift
```diff
@@ -734,6 +734,104 @@ struct TestCommandTests {
...
+            .init(
+                fixturePath: "Miscellaneous/TestSingleFailureSwiftTesting",
+                expectNote: false,
+            ),
+            .init(
+                fixturePath: "Miscellaneous/TestSingleFailureXCTest",
+                isXCTestEnabled: false,
+                expectNote: false,
+            ),
+            .init(
```
plemarquand, line 802: nit: missing the very silly case testing `--disable-swift-testing --disable-xctest`

Sources/Commands/SwiftTestCommand.swift
stmontgomery: Thanks, I half suspected this. I originally tried `emit(info: ...)` on `swiftCommandState.observabilityScope` but that did not result in any visible output. However, I now see there is a `print(_:verbose:)` on that type so I assume that's what I should use.

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -377,6 +380,13 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
                     debug: "Skipping automatic Swift Testing invocation because a test entry point path is present: \(testEntryPointPath)"
                 )
             }
+
+            // After running Swift Testing tests, if we determined that any XCTests failed earlier,
+            // emit a message informing the user so they aren't misled and know to look elsewhere for
+            // those details.
+            if anyXCTestFailed {
```
grynspan, line 394: I think we only want to include this output if Swift Testing actually ran.

stmontgomery: Yeah, this is guarded by `if options.testLibraryOptions.isEnabled(.swiftTesting, ...)` so that's happening already. One of the accompanying test cases (the one which specifies `isSwiftTestingEnabled: false`) validates that.

grynspan: Did you push a change? GitHub was showing this line aligned with the outer scope. 🫠

stmontgomery: It's already guarded the way I'm describing

Sources/Commands/SwiftTestCommand.swift
stmontgomery: Made that change

Tests/CommandsTests/TestCommandTests.swift
stmontgomery: Added a case for this

stmontgomery:
@swift-ci test

Sources/Commands/SwiftTestCommand.swift
```diff
@@ -377,6 +380,13 @@ public struct SwiftTestCommand: AsyncSwiftCommand {
...
+                swiftCommandState.observabilityScope.print("Note: One or more XCTests failed, see logging above for details.", verbose: false)
```
bkhouri: suggestion: can we store the output string in a variable, and then reference this variable in the test?  This would make the test more robust to changes to this string.

stmontgomery:
@swift-ci test Windows Platform

stmontgomery:
@swift-ci test macOS

stmontgomery:
@swift-ci test

Sources/Commands/SwiftTestCommand.swift
stmontgomery: Done!

stmontgomery:
@swift-ci test Windows

stmontgomery:
@swift-ci test Windows

bkhouri (approved):

bkhouri:

stmontgomery:

stmontgomery:

stmontgomery:
@swift-ci test

stmontgomery:
@swift-ci test

stmontgomery:
@swift-ci test

stmontgomery:
@swift-ci test

stmontgomery:
@swift-ci test Windows Platform

PR #9658, opened 2026-01-28
Fix Swift SDK triple handling

### LINKED ISSUE #9220: SwiftPM selects wrong architecture from multi-arch Swift SDK when using --swift-sdk and --triple

stackotter:
Resolves https://github.com/swiftlang/swift-package-manager/issues/9220, https://github.com/swiftlang/swift-package-manager/issues/7973, and others.

### Motivation:

When a user has two SDK bundles installed that have at least two supported target triples in common, then it's impossible to disambiguate the SDKs via the command line. That issue is tracked by https://github.com/swiftlang/swift-package-manager/issues/7973.

### Modifications:

I've solved the issue by updating the SDK resolution logic to respect `--triple` when present (which fixes #9220). I've simply extended the search logic to require the candidate SDKs' target triples to match the provided triple to be considered.

### Result:

These changes should only affect SwiftPM behaviour when both `--triple` and `--swift-sdk` are present. Currently, SwiftPM ignores `--triple` when `--swift-sdk` is present. After the changes, it respects `--triple` in a way that I find intuitive, and others in the linked issues seem to find intuitive as well.

---

The PR currently targets `release/6.3` because it's late at night and `main` wasn't compiling on my machine. Before undrafting I will rebase onto whichever branch is considered suitable. Please let me know which branch I should target with this PR.

### DIFF
--- a/Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
+++ b/Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
@@ -79,11 +79,13 @@ extension [SwiftSDKBundle] {
     /// - Parameters:
     ///   - selector: either an artifact ID or target triple to filter with.
     ///   - hostTriple: triple of the host building with these Swift SDKs.
+    ///   - targetTriple: target triple for compilation.
     ///   - observabilityScope: observability scope to log warnings about multiple matches.
     /// - Returns: ``SwiftSDK`` value matching `query` either by artifact ID or target triple, `nil` if none found.
     func selectSwiftSDK(
         matching selector: String,
         hostTriple: Triple,
+        targetTriple: Triple?,
         observabilityScope: ObservabilityScope
     ) -> SwiftSDK? {
         var matchedByID: (path: AbsolutePath, variant: SwiftSDKBundle.Variant, swiftSDK: SwiftSDK)?
@@ -95,7 +97,9 @@ extension [SwiftSDKBundle] {
                     guard variant.isSupporting(hostTriple: hostTriple) else { continue }
 
                     for swiftSDK in variant.swiftSDKs {
-                        if artifactID == selector {
+                        if artifactID == selector,
+                           targetTriple == nil || swiftSDK.targetTriple?.triple == targetTriple?.triple
+                        {
                             if let matchedByID {
                                 observabilityScope.emit(
                                     warning:
@@ -104,6 +108,8 @@ extension [SwiftSDKBundle] {
                                         hostTriple.tripleString
                                     ), selected one at \(
                                         matchedByID.path.appending(matchedByID.variant.metadata.path)
+                                    ) with target triple \(
+                                        matchedByID.swiftSDK.targetTriple?.description ?? "<unspecified>"
                                     )
                                     """
                                 )
@@ -126,6 +132,15 @@ extension [SwiftSDKBundle] {
                                 )
                             } else {
                                 matchedByTriple = (bundle.path, variant, swiftSDK)
+                                if targetTriple != nil {
+                                    observabilityScope.emit(
+                                        warning:
+                                        """
+                                        the provided --swift-sdk is a target triple, but --triple was \
+                                        also provided; ignoring --triple
+                                        """
+                                    )
+                                }
                             }
                         }
                     }
--- a/Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
+++ b/Sources/PackageModel/SwiftSDKs/SwiftSDK.swift
@@ -773,13 +773,13 @@ public struct SwiftSDK: Equatable {
             } else {
                 throw SwiftSDKError.noSwiftSDKDecoded(customDestination)
             }
-        } else if let targetTriple = customCompileTriple,
-                  let targetSwiftSDK = SwiftSDK.defaultSwiftSDK(for: targetTriple, hostSDK: hostSwiftSDK)
-        {
-            swiftSDK = targetSwiftSDK
         } else if let swiftSDKSelector {
             do {
-                swiftSDK = try store.selectBundle(matching: swiftSDKSelector, hostTriple: hostTriple)
+                swiftSDK = try store.selectBundle(
+                    matching: swiftSDKSelector,
+                    hostTriple: hostTriple,
+                    targetTriple: customCompileTriple
+                )
             } catch {
                 // If a user-installed bundle for the selector doesn't exist, check if the
                 // selector is recognized as a default SDK.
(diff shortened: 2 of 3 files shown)


finagolfin:
I don't like this because I'd like to change that selection logic more substantially, will let you know.

stackotter:
What way would you like to do selection differently? Do you mean resolving to a different SDK than my code currently does? Or having an SDK identifier format that allows specifying the triple without the triple argument?

As far as I can see, this does resolve the ambiguity issues, and respecting triple this way seems like it makes sense no matter whether or not we add triple disambiguation to sdk identifiers themselves.

But I totally could’ve overlooked something. I’m not intimately familiar with the SDK system.

finagolfin:
[See the planned changes I wrote up earlier](https://github.com/swiftlang/swift-package-manager/pull/9229#discussion_r2447111876).

stackotter:
Ah ok thanks for the link. From my understanding, it seems like the main differences between your planned approach and my implementation are that you will throw an error on duplicates instead of emitting a warning, you will deprecate the use of triples as sdk selectors (supposedly just with a warning for now?), and something regarding local config overrides. I agree with the first two and am unaware of the latter but can familiarise myself with it.

I would happily look into implementing those in my PR if it will be some time before you get back to yours. Please let me know if that would be helpful or not, because if my PR is going to sit here indefinitely regardless of whether or not I adopt your planned changes then I will just wait until yours is ready.

dschaefer2:
There's some major architectural work for SDKs I'm trying to put together. But in the end, you should be able to select an SDK and a triple that that SDK supports. As we've seen, there's too much magic happening with the Swift SDK selection.

dschaefer2:
This is next on my list. I'll be doing it on main though and we can look at back porting it to 6.3.x

dschaefer2:
FWIW, I believe the main flow should be to specify the SDK and the triple. You can skip the triple if there's only one.

I'm not sure why we allowed you to specify the triple as the SDK. We can support that for backwards compatibility if there is only one SDK that supports that triple. If it's ambiguous we'd have to error out.

finagolfin:
@dschaefer2, I'm finally looking at this myself too. You mentioned "major architectural work for SDKs" before, whereas I was just planning on modifying or replacing the 2-3 `PackageModel` methods I listed in #9229. Let me know if your approach differs much and we can figure out which to use.

dschaefer2:
The main change I want to make sure we leave a path for is for SDKs to be delivered in packages and for them to contribute to platform conditionals. There'd I would want to make things conditional on the SDK being used for the build and the triple.

I also want to see if we can simplify the id's for SDKs. It looks like the current format is <swiftCompilerTag>_<platformName>. Can we just specify `--swift-sdk android --triple x86_64-unknown-linux-android28` for example. Though in this case that triple is specific to that SDK so we could get away with the current practice of specifying `--swift-sdk triple` though that seems very wrong.

dschaefer2:
We use the swiftCompilerTag to pick which prebuilt to use. We should be able to use it for Swift SDKs as well.

finagolfin:
Alright, I'll slap up my pull tomorrow morning and we can compare notes. 👍

dschaefer2:
You're way ahead of me so will follow you're lead :)

jakepetroules:
Thanks @finagolfin, looking forward to it! Will be glad to have the nondeterministic matching gone :)

finagolfin:
I got delayed by my Android CI work, but this pull has now reached the top of my stack, so will update soon.

Sources/PackageModel/SwiftSDKs/SwiftSDKBundle.swift
```diff
@@ -104,6 +108,8 @@ extension [SwiftSDKBundle] {
                                         hostTriple.tripleString
                                     ), selected one at \(
                                         matchedByID.path.appending(matchedByID.variant.metadata.path)
+                                    ) with target triple \(
+                                        matchedByID.swiftSDK.targetTriple?.description ?? "<unspecified>"
```
finagolfin, line 112: This is all indeterminate, including your checking for the triple above, see my updated #9229. Try that pull out instead and let me know how it works for you.

Sorry for pushing this simpler pull off for so long, but we really need to fix this properly.

finagolfin:
> We use the swiftCompilerTag to pick which prebuilt to use. We should be able to use it for Swift SDKs as well.

I took a look at that and while we can get the compiler tag that way, there is currently no way to get the SDK tag. Maybe it's buried in the swiftmodules somewhere and we can fish it out, but we haven't done that yet, if it's even possible.

stackotter:
> I took a look at that and while we can get the compiler tag that way, there is currently no way to get the SDK tag. Maybe it's buried in the swiftmodules somewhere and we can fish it out, but we haven't done that yet, if it's even possible.

In Swift Bundler I’ve got it locating valid toolchain+sdk pairs, but the only way I’ve found to do so remotely reliably is to parse out the Swift compiler version from the metadata comments at the top of swiftinterface files (also available in some of the other binary serialised files). It feels a bit hacky for SwiftPM to do that though (Swift Bundler just has to do the best it can given the circumstances, but SwiftPM of course has less of an excuse; being as integrated with the compiler and everything else as it is)

finagolfin:
Please try out #9955, and close this if that works instead.

PR #9662, opened 2026-01-29
[🍒] Include linker generated symbol in default defined set (#9416)

FranzBusch:
Cherry-pick of #9416


FranzBusch:
@swift-ci test

PR #9665, opened 2026-01-30
Treat package root top dependencies as exporting for prebuilts

dschaefer2:
The package dependencies in the PackageGraphRoot are also exporting to the outside world. That means they can be built for destination (non-host). Factor that into the leaky product algorithm for prebuilts to ensure we don't force libraries exported out that way to build for host only.


dschaefer2:
@swift-ci please test windows

PR #9668, opened 2026-01-30
[9654] Add helpful hint for 'swift package lint' command

### LINKED ISSUE #9654: Error: unknown or missing subcommand 'swift-lint'
### Description

Seems to be happening in the latest version, been using the embedded lint tool for a while now but this is happening recently `6.2.3`. Format still works great.

### Reproduction

```sh
mac@Macs-MacBook-Air Developer % swift lint
error: unknown or missing subcommand 'swift-lint'
```

### Expected behavior

Swift lint should print the help text.

### Environment

swiftc -version

Apple Swift version 6.2.3 (swift-6.2.3-RELEASE)
Build config: +assertions

### Additional information

_No response_

rickhohler:
### Motivation:

Resolves #9654.
Users attempting to run `swift lint` (expecting functionality similar to `swift test`) encounter a generic "unknown subcommand" error. The intended command is now part of `swift-format`. This change intercepts the `lint` command to provide a helpful hint, guiding the user to the correct invocation.

### Modifications:

1.  Created `Sources/Commands/PackageCommands/Lint.swift`: A new `ParsableCommand` that is hidden from the help menu. It catches the `lint` subcommand and throws a `CleanExit` with a specific error message:
    > error: unknown subcommand 'lint'; did you mean 'swift format lint'?
2.  Registered the `Lint` subcommand in `SwiftPackageCommand.swift` to enable this interception.
3.  Added a regression test `testLintSubcommand` in `PackageCommandTests.swift` to verify the error message is printed correctly to stderr/stdout.

### Result:

When a user runs `swift package lint` (or `swift lint` if the driver delegates), they will see a clear error message suggesting `swift format lint` instead of a generic "unknown command" error. The command exits with a non-zero status code, preserving the failure behavior but improving the diagnostics.

### DIFF
--- a/Sources/Commands/PackageCommands/Lint.swift
+++ b/Sources/Commands/PackageCommands/Lint.swift
@@ -1,0 +1,29 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2014-2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import CoreCommands
+
+struct Lint: ParsableCommand {
+    static let configuration = CommandConfiguration(
+        abstract: "Lint the package (deprecated).",
+        shouldDisplay: false
+    )
+
+    // We don't need any arguments because we just want to catch the command itself
+    // and fail immediately with the hint.
+
+    func run() throws {
+        // Use CleanExit.message to exit cleanly with an error message but without a stack trace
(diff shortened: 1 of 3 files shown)


plemarquand:
@swift-ci test

plemarquand (commented):
I think this is a good improvement. 

My only thought is maybe we should simply alias "swift lint" to "swift package lint" with a warning message instead. 

I'll let other contributors chime in since I'm not sure if there is a policy on aliasing like this, as it could easily get out of hand trying to cover all possible options.

bkhouri (changes requested):
I like the drive of wanting to provide an actionable error message. However, I feel the better solution is modifying the https://github.com/swiftlang/swift-format repository to produce a `swift-lint` executable, that would perform the real "lint" action.

See https://github.com/swiftlang/swift-format/issues/1152, which "duplicates" #9654 to the swift-format repository.

rickhohler:
Thanks @bkhouri for the suggestion! I've opened a PR in `swift-format` to implement `swift-lint` there: https://github.com/swiftlang/swift-format/pull/1153. I'll close this one in favor of that implementation.

PR #8845, opened 2025-06-19
Activate swift build support for command plugins

cmcgee1024:
Be consistent in the detection of the build system settings so that both the
build system and the build parameters are using the same setting. Use the
build system setting given after the command plugin name as the highest
precedence, followed by the one (if any) provided after the package subcommand.

When building a specific target with the swiftbuild build system, set the configured
target in the build request to be the one that isn't a dynamic target.

Tag the `TraitTest.packagePluginGetSymbolGraph_enablesAllTraits` with a known
issue when using the swiftbuild build system since the symbol graph extraction
is not yet working there.

Add the toolchain's runtime environment to the environment used when running
command plugins so that the tools have the necessary runtime dependencies when
the plugin decides to run them on platforms like Linux.

PR #9669, opened 2026-01-31
Filter out testable target variants when building macros with swiftbuild

Ryu0118:
This PR fixes building macros with `--target` in swiftbuild.

### Motivation:

`swift build --target <macro_name> --build-system swiftbuild` fails with `Found multiple targets named` error. 
The PIF builder creates both a macro target and its `.testable` variant, and both match by name.
The existing code only filtered out `.dynamic` variants (#8845), but not `.testable` variants.

### Modifications:

Extend the target filter to exclude both `.dynamic` and `.testable` variants.

### Result:

`swift build --target <macro_name> --build-system swiftbuild` works correctly.

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -578,8 +578,13 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
                         let workspaceInfo = try await session.workspaceInfo()
 
                         configuredTargets = try [pifTargetName].map { targetName in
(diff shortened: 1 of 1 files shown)


owenv (approved):
thanks - I think we could refactor this to avoid relying on suffix checks to map the user-facing targets to PIF targets by having the PIFBuilder maintain a mapping between the two, but in the meantime this is a good fix

owenv:
@swift-ci please test

owenv:
@swift-ci please test linux

owenv:
@swift-ci please test Linux

PR #8405, opened 2025-03-24
Introduces the new PIF builder for SwiftBuild

pmattos:
Introduces the new PIF builder for packages in `SwiftBuildSupport`.

### Motivation:

This continues the work started by [PR 8271](https://github.com/swiftlang/swift-package-manager/pull/8271), introducing now our new *PIF builder* for packages. As expected, this will happen in the new `SwiftBuildSupport` target, introduced to support the just released **Swift Build** project. 

### Modifications:

This 1st PR *only* includes the new code in the corresponding `SwiftBuildSupport/PIFPackageBuilder*.swift` source files. The code isn't actually being used (or unit tested) just yet.

The code still uses the older `SWBProjectModel.PIF` API to build the PIF data model. In an upcoming PR, I will be migrating this over to the newer `SwiftBuild.ProjectModel` API, and then finally replacing the original code to actually use the PIF builder to send package build information to Swift Build.

PS. I did run `SwiftFormat` in all new code, as indicated by the contributors guide.

### Result:

This marks the start of our *package PIF builder* unification (i.e., between Xcode and SwiftPM).

Tracked by rdar://147767772.

PR #9670, opened 2026-01-31
Remove function computing legacy pif package product name

Ryu0118:
This PR removes the unused `computePackageProductFrameworkName` function.

### Motivation:
This function was introduced in #8405 but the only call site was removed in #9130.

### Modifications:
Remove the unused `computePackageProductFrameworkName` function.


plemarquand:
@swift-ci test

PR #9672, opened 2026-02-01
Cleanup prebuilts for MacOS, make manifest per platform.

dschaefer2:
Prebuilts for the Xcode toolchain need to take into account the MacOS SDK they are built against in case the same compiler version is used for multiple SDK releases. Load up the SDKSettings.json file in the MacOS SDK referenced in the Workspace's host toolchain to fetch the Canonical Name and add it to the manifest and zip file names when building and downloading the files.

Also clean up and produce signed manifests per platform instead of one shared one. The shared one was hard to update properly since CI builds prebuilts for the platforms in parallel. Allows us to clean up a lot code.

Remove cModules from the Library structure since that isn't used any more. Generally, clean things up since the version constraints ensures that prebuilts only work in the toolchain SwiftPM is distributed. No backwards compatibility is needed.

### DIFF
--- a/Sources/Workspace/Workspace+Prebuilts.swift
+++ b/Sources/Workspace/Workspace+Prebuilts.swift
@@ -228,7 +228,7 @@ extension Workspace {
 
         func prebuiltName(workspace: Workspace) throws -> String {
             if let customSwiftCompilerVersion {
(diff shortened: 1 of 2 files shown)


dschaefer2:
@swift-ci please test

PR #9679, opened 2026-02-03
Fix failed prebuilts test

dschaefer2:
Fixed the compile error but now the test is failing.


dschaefer2:
@swift-ci please test

PR #9681, opened 2026-02-03
[6.3] Reduce restrictions on Code Gen of C artifacts

dschaefer2:
The Swan project which provides a Swift interface for the Dawn webgpu library. It is suffering from restrictions in SwiftPM that many others have run into. In particular, they'd like to generate API notes for the Dawn interface as described in the dawn.json file. To support that we need to also generate a modulemap and header file co-located with the API notes. We then need to add that to the public header path for the module.

To differentiate these generated headers from other header files that are provide, the public files must be in an a directory called "include" in the plugin output directory for that target.

Also to support Dawn on Windows, we need to copy DLLs to support the webgpu_dawn library that are shipped with the binary artifact into the build output directory so that generated executables will run properly.

This is just a stop gap and presents reasonable defaults. But in the future we should make these actions configurable from the package manifest and plugins. In the mean time consider these features experimental.

### DIFF
--- a/Tests/BuildTests/CGenPluginsBuildPlanTests.swift
+++ b/Tests/BuildTests/CGenPluginsBuildPlanTests.swift
@@ -31,6 +31,14 @@ import Build
         case swiftModule
     }
 
+    let pluginOutputDir: Basics.AbsolutePath = "/PluginOut/outputs/mypkg/MyModule/destination/MyPlugin"
+    var pluginIncludeDir: Basics.AbsolutePath { pluginOutputDir.appending("include") }
+    var pluginModuleMapFile: Basics.AbsolutePath { pluginIncludeDir.appending("module.modulemap") }
+    var pluginModuleMapArg: String { "-fmodule-map-file=\(pluginModuleMapFile.pathString)" }
+    var pluginAPINotesFile: Basics.AbsolutePath { pluginIncludeDir.appending("Gened.apinotes")}
+    var pluginHeaderFile: Basics.AbsolutePath { pluginIncludeDir.appending("Gened.h")}
+    var pluginSourceFile: Basics.AbsolutePath { pluginOutputDir.appending("Gened.c") }
+
     func setup(
         kind: Kind = .cModule,
         gened: [RelativePath],
@@ -262,10 +270,10 @@ import Build
 
         let warnings = observability.warnings.map(\.message)
         let messages: [String] = [
-            "Only C modules support plugin generated C header files: /PluginOut/outputs/mypkg/MyModule/destination/MyPlugin/include/Gened.h",
-            "Only C modules support plugin generated module map files: /PluginOut/outputs/mypkg/MyModule/destination/MyPlugin/include/module.modulemap",
-            "Only C modules support plugin generated API notes files: /PluginOut/outputs/mypkg/MyModule/destination/MyPlugin/include/Gened.apinotes",
-            "Only C modules support plugin generated C source files: /PluginOut/outputs/mypkg/MyModule/destination/MyPlugin/Gened.c",
+            "Only C modules support plugin generated C header files: \(pluginHeaderFile.pathString)",
+            "Only C modules support plugin generated module map files: \(pluginModuleMapFile.pathString)",
+            "Only C modules support plugin generated API notes files: \(pluginAPINotesFile.pathString)",
+            "Only C modules support plugin generated C source files: \(pluginSourceFile.pathString)",
         ]
 
         #expect(warnings.count == messages.count)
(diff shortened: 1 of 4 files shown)


dschaefer2:
@swift-ci please test

Sources/Build/BuildPlan/BuildPlan+Swift.swift
```diff
@@ -74,6 +77,13 @@ extension BuildPlan {
 
                         swiftTarget.libraryBinaryPaths.insert(library.libraryPath)
                     }
+                    if swiftTarget.buildParameters.triple.isWindows() {
+                        // Look for DLLs that need to be copied over with the library
+                        let executables = try self.parseExecutableArtifactsArchive(for: target, triple: swiftTarget.buildParameters.triple)
```
owenv, line 82: dlls should be treated as library artifacts, not executables. Rather than inserting a special case here I think we need to figure out why they're not being classified correctly

Sources/PackageLoading/TargetSourcesBuilder.swift
```diff
@@ -560,26 +566,89 @@ public struct TargetSourcesBuilder {
             switch rule {
             case .compile:
                 if absPath.extension == "swift" {
-                    sources.append(absPath)
+                    files.sources.append(absPath)
+                } else if toolsVersion >= .v6_3 {
```
owenv, line 570: This should use the special 999 tools version until it's gone through swift-evolution

Sources/SPMBuildCore/Plugins/PluginInvocation.swift
```diff
@@ -619,41 +619,53 @@ extension ModulesGraph {
...
+                defaultLocalization: target.defaultLocalization,
+                module: target.underlying,
+                observabilityScope: observabilityScope
+            )
+            generatedFiles.add(files)
+            if !files.headers.isEmpty {
+                // Capture the public include directory if there were header files generated there
+                // Hardcoding as the default for now
+                let publicDir = result.pluginOutputDirectory.appending(ClangModule.defaultPublicHeadersComponent)
+                if files.headers.contains(where: { $0.isDescendantOfOrEqual(to: publicDir) }) {
```
owenv, line 640: Header search paths aren't recursive, so I think this check is too broad

Sources/SPMBuildCore/Plugins/PluginInvocation.swift
```diff
@@ -619,41 +619,53 @@ extension ModulesGraph {
...
+                module: target.underlying,
+                observabilityScope: observabilityScope
+            )
+            generatedFiles.add(files)
+            for header in files.headers {
+                let outputDirs = result.outputDirectories.filter { header.isDescendantOfOrEqual(to: $0) }
+                for dir in outputDirs {
+                    // If the output dir ends in the default public header dir, make it public (yuck)
+                    if dir.basename == ClangModule.defaultPublicHeadersComponent {
+                        generatedFiles.publicHeaderPaths.append(dir)
```
owenv, line 662: This is potentially going to expose unrelated generated content in the header search paths. I'd recommend copying generated headers into an install location and searching that at compile time. This also becomes more important if we want to allow building package products for binary distribution with their header content in the future

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
```diff
@@ -396,37 +401,51 @@ extension PackagePIFProjectBuilder {
...
-            case .umbrellaHeader(let path):
-                log(.debug, "\(package.name).\(sourceModule.name) generated umbrella header")
-                moduleMapFileContents = """
-                module \(sourceModule.c99name) {
-                umbrella header "\(path.escapedPathString)"
-                export *
+            if let generatedModuleMapPath = generatedFiles.moduleMaps.first {
+                // Warn about ignored generated module maps if more than one
+                if generatedFiles.moduleMaps.count > 1 {
+                    let ignoredFiles = generatedFiles.moduleMaps.filter({ $0 != generatedModuleMapPath })
```
owenv, line 407: I think we should have manifest API for this instead of making an arbitrary decision here.

Sources/SwiftBuildSupport/PackagePIFProjectBuilder.swift
```diff
@@ -408,19 +408,32 @@ struct PackagePIFProjectBuilder {
...
+                    forModule: module,
+                    toolsVersion: self.package.manifest.toolsVersion
+                )
+
+                generatedFiles.add(files)
+                if !files.headers.isEmpty {
+                    // Capture the public include directory if there were header files generated there
+                    // Hardcoding as the default for now
+                    let publicDir = command.pluginOutputDir.appending(ClangModule.defaultPublicHeadersComponent)
+                    if files.headers.contains(where: { $0.isDescendantOfOrEqual(to: publicDir) }) {
```
owenv, line 423: Same comment as above about search paths not being recursive

Sources/SPMBuildCore/Plugins/PluginInvocation.swift
```diff
@@ -619,41 +619,53 @@ extension ModulesGraph {
...
+                toolsVersion: toolsVersion,
+                additionalFileRules: additionalFileRules,
+                defaultLocalization: target.defaultLocalization,
+                module: target.underlying,
+                observabilityScope: observabilityScope
+            )
+            generatedFiles.add(files)
+            if !files.headers.isEmpty {
+                // Capture the public include directory if there were header files generated there
+                // Hardcoding as the default for now
```
owenv, line 638: I think silently dropping headers generated outside this path is going to be very confusing in practice. It would be preferable to expose plugin API which allows the plugin itself to impose search paths on downstream compilation

PR #9682, opened 2026-02-04
Fix pkg-config path lookup on non-macOS platforms when using Swift Build

owenv:
Ensure we don't attempt to pass a homebrew prefix on linux, where it will cause us to fail the tool lookup.

Closes https://github.com/swiftlang/swift-package-manager/issues/9630


owenv:
@swift-ci test

PR #9683, opened 2026-02-04
Fix `OutputHandler.print` method API

bripeticca:
- Modify the method signature to instead use a `condition` in the form of an enum (`OutputCondition`) to clarify that this parameter serves to check whether the output should be printed depending on if the `logLevel` is verbose.
- Fix areas where this method is called

### DIFF
--- a/Sources/Basics/Observability.swift
+++ b/Sources/Basics/Observability.swift
@@ -44,7 +44,7 @@ public class ObservabilitySystem {
     public convenience init(
         _ handler: @escaping @Sendable (ObservabilityScope, Diagnostic) -> Void,
         outputStream: (any WritableByteStream)? = nil,
-        logLevel: Diagnostic.Severity
+        logLevel: Diagnostic.Severity = .error,
     ) {
         self.init(
             SingleDiagnosticsHandler(


bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci test windows

Sources/_InternalTestSupport/Observability.swift
```diff
@@ -88,7 +88,7 @@ public struct TestingObservability {
             self.diagnostics.append(diagnostic)
         }
 
-        func print(_ output: String, verbose: Bool) {
+        func print(_ output: String, condition: OutputCondition) {
             if verbose {
```
bkhouri: issue (blocking): we should perform a `switch condition` here.

Sources/Basics/Observability.swift
```diff
@@ -60,7 +60,7 @@ public class ObservabilitySystem {
             self.underlying(scope, diagnostic)
         }
 
-        func print(_ output: String, verbose: Bool) {
+        func print(_ output: String, condition: OutputCondition) {
             outputStream?.write(output)
```
bkhouri: issue (blocking): should we check `condition` to determine if we should write to the outputStream?

Sources/CoreCommands/SwiftCommandObservabilityHandler.swift
```diff
@@ -117,12 +117,12 @@ public struct SwiftCommandObservabilityHandler: ObservabilityHandlerProvider {
         }
 
         // for raw output reporting
-        func print(_ output: String, verbose: Bool) {
+        func print(_ output: String, condition: OutputCondition) {
             self.queue.async(group: self.sync) {
-                guard verbose || self.logLevel.isVerbose else {
-                    return
+                if (condition == .onlyWhenVerbose && self.logLevel.isVerbose) ||
```
bkhouri: suggestion: If we are to add a new `case` to the `OutputCondition` enum, we may forget to handle the condition.  Instead of an `if` block, can we convert this to a `switch` to ensure we handle new `case`s added to the `OutputCondition` enum?

Tests/BasicsTests/ObservabilitySystemTests.swift
```diff
@@ -310,10 +310,8 @@ struct ObservabilitySystemTest {
             self._diagnostics.append(diagnostic)
         }
 
-        func print(_ output: String, verbose: Bool) {
-            if verbose {
-                Swift.print(output)
-            }
+        func print(_ output: String, condition: OutputCondition) {
+            Swift.print(output)
```
bkhouri: question: why aren't we taking the `condition` value into consideration here?

bripeticca: Usually we'd follow up the `condition` check with a guard against whether the log level is verbose or not, but this struct doesn't have one available. I can add it and then do a `switch` against `condition`.

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test windows

bripeticca:
@swift-ci test

bripeticca:
@swift-ci test

bripeticca:
@swift-ci test windows

rconnell9:
@swift-ci test

bripeticca:
@swift-ci test

bripeticca:
@swift-ci test

bripeticca:
@swift-ci test windows

bripeticca:
@swift-ci test windows

bripeticca:
@swift-ci test linux

PR #9693, opened 2026-02-07
SwiftCompilerFeature: Do not require upcoming features to carry a language mode

AnthonyLatsis:
Newer upcoming features are tied to a future language mode that does not exist yet, so don't expect it to always be present. In due time, this change will enable us to stop printing a made up language mode for these features in the compiler without breaking SwiftPM.


xedin (approved):

AnthonyLatsis:
@swift-ci please test

PR #9697, opened 2026-02-09
Fix containsAtMain to detect @main after */ on same line

pranav-singh1:
Fix containsAtMain handling of @main following block comments

Motivation

containsAtMain fails to detect @main when a block comment ends mid-line and the annotation appears after the closing */. In this case, the remainder of the line is not scanned, causing valid @main declarations to be missed.

This affects otherwise valid Swift sources that place @main after a block comment terminator on the same line.

Modifications

Updated containsAtMain to continue scanning the remainder of a line after a block comment ends.

Added a regression test covering @main appearing on the same line as a block comment terminator.

Result

@main is now correctly detected when it appears after */ on the same line, with no change to existing behavior in other cases.

### DIFF
--- a/Sources/SPMBuildCore/MainAttrDetection.swift
+++ b/Sources/SPMBuildCore/MainAttrDetection.swift
@@ -28,8 +28,15 @@ package func containsAtMain(fileSystem: FileSystem, path: AbsolutePath) throws -
         if line.hasPrefix("/*") {
             multilineComment = true
         }
-        if line.hasSuffix("*/") {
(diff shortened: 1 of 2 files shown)


plemarquand:
@swift-ci test

grynspan:
Naïvely… can we make this use swift-syntax?

bkhouri:
@pranav-singh1 : Thanks for this fix.  Could I ask that you resolve the merge conflict?  Then we can re-trigger the CI builds.

PR #9709, opened 2026-02-11
Make sure Windows DLLs are copied over with their LIBs.

dschaefer2:
On Windows, static and dynamic libraries are similar in that you link in the LIB file. If it's for a DLL, it also needs to be copied over to the build output directory so the resulting executable can find it in the same directory without having to mess with PATH. Since artifact bundles don't have that concept, we reuse "executable" type and check if it's a Windows target and a ".dll" file extension.

Note this is Native build system only.

### DIFF
--- a/Sources/SPMBuildCore/BinaryTarget+Extensions.swift
+++ b/Sources/SPMBuildCore/BinaryTarget+Extensions.swift
@@ -39,6 +39,18 @@ public struct ExecutableInfo: Equatable {
     public let supportedTriples: [Triple]
 }
 
+/// Information about a Windows DLL from a binary dependency.
+public struct WindowsDLLInfo: Equatable {
+    /// The tool name
+    public let name: String
+
+    /// The path to the executable.
+    public let dllPath: Basics.AbsolutePath
+
+    /// Supported triples, e.g. `x86_64-apple-macosx`
+    public let supportedTriples: [Triple]
+}
+
 extension BinaryModule {
     public func parseXCFrameworks(for triple: Triple, fileSystem: FileSystem) throws -> [LibraryInfo] {
         // At the moment we return at most a single library.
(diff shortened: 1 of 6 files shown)


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

Tests/BuildTests/LLBuildManifestBuilderTests.swift
```diff
@@ -206,4 +206,136 @@ struct LLBuildManifestBuilderTests {
...
+            ]
+        )
+
+        try fs.writeFileContents("/MyPkg/my.artifactbundle/info.json", string: """
+            {
+              "schemaVersion": "1.0",
+              "artifacts": {
+                "MyBinaryLib": {
+                  "version": "1",
+                  "type": "staticLibrary",
```
owenv, line 224: A .lib corresponding to a DLL really isn't a static library, and a static library really isn't an executable so this feels like we're misusing the artifact bundle layout. I think it would be best to look at introducing a new artifact type with separate fields for the library interface and implementation, both to make the intent clearer, and to diagnose accidental misuse of a particular artifact type better.

dschaefer2: For sure. This is really to unblock people now trying to package DLLs for Windows in artifact bundles and are doing so by specifying them as static libraries, which is also working but adds them to the link line which is also not what we want.

dschaefer2: I could change the tag and mark it as "experimentalDLL" or something like that.

dschaefer2: Actually, I'll do that. Make it clear it's only temporary.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

Sources/SPMBuildCore/BinaryTarget+Extensions.swift
```diff
@@ -39,6 +39,18 @@ public struct ExecutableInfo: Equatable {
...
 
+/// Information about a Windows DLL from a binary dependency.
+public struct WindowsDLLInfo: Equatable {
+    /// The tool name
+    public let name: String
+
+    /// The path to the executable.
+    public let dllPath: Basics.AbsolutePath
+
+    /// Supported triples, e.g. `x86_64-apple-macosx`
```
jakepetroules, line 50: nit: Apple triple in a Windows-specific type :)

jakepetroules (approved):

PR #9713, opened 2026-02-12
Make sure Windows DLLs are copied over with their LIBs. (cherry-pick #9709)

dschaefer2:
On Windows, static and dynamic libraries are similar in that you link in the LIB file. If it's for a DLL, it also needs to be copied over to the build output directory so the resulting executable can find it in the same directory without having to mess with PATH. Since artifact bundles don't have that concept, we reuse "executable" type and check if it's a Windows target and a ".dll" file extension.

Note this is Native build system only.


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test macOS

owenv (approved):

PR #9730, opened 2026-02-12
Experimental C code generation from build tool plugins.

dschaefer2:
Add the ability to generated C source, header, module maps, and API notes from build tool plugins. Public headers as well as the module maps and API notes must be generated into a directory called "include" in the plugin output directory for a target. This will be configurable in the final implementation of this feature in upcoming releases.

To enable this feature, adjust the swift-tools-version in the Package.swift that contains the target the plugin is generating for as follows:

// swift-tools-version: 6.3;(experimentalCGen)

This is an experimental method for specifying experimental features to be used by that package.

This feature when enabled works with both native and swiftbuild build systems.


dschaefer2:
@swift-ci please test

Sources/PackageLoading/ToolsVersionParser.swift
```diff
@@ -358,6 +358,20 @@ public struct ToolsVersionParser {
         /// - Note: For a misspelt Swift tools version specification `"// swift-too1s-version:5.3"`, the first `"1"` is considered as the first character of the version specifier, and so `"1s-version:5.3"` is taken as the version specifier.
         let versionSpecifier = specificationSnippetFromLabelToLineTerminator[startIndexOfVersionSpecifier..<indexOfVersionSpecifierTerminator]
 
+        /// Look for experimental features. They are space separated values contained in parenthases right after the ";", e.g. "// swift-tools-version: 6.3;(experimentalCGen)"
```
owenv, line 361: I can't think of anything off the top of my head, but do you know of any tools outside SwiftPM which might do their own parsing of this line?

dschaefer2: I do not. It's a weird syntactic element which turned out pretty handy but does seem to be there for anyone to add metadata there. But people could put anything there so they would have to deal with that anyway like I did having the parenthesis right after the semicolon and ignoring flags we don't know.

dschaefer2: BTW, the 999 trick didn't work. The development mode flag lets it be larger than the current swiftpm version, otherwise it fails. Luckily someone put in the ';' trick or otherwise, putting anything else on the line would cause it to fail to parse.

dschaefer2:
@swift-ci please test windows

PR #9732, opened 2026-02-13
Fix containsAtMain to handle @main on same line as block comment end

### LINKED ISSUE #9685: `containsAtMain` does not handle `@main` on the same line as a multi-line end comment
The `containsAtMain` function call does not correctly return `true` with the following input 

```swift
                /*
                This is a multi-line comment
                /* @main
                struct MyApp {
                    static func main() {
                        print("Hello, World!")
                    }
                }
```

Search for the GitHub issue in the swift-package-manager repository to located an impacted automated test

william-laverty:
## Summary

Fixes #9685

The `containsAtMain` function failed to detect `@main` when it appeared on the same line as a block comment end (`*/ @main`).

## Changes

**`Sources/SPMBuildCore/MainAttrDetection.swift`:**
- Replaced line-level `hasPrefix`/`hasSuffix` comment checks with a scanning approach that tracks `/*` and `*/` boundaries within a single line
- After encountering `*/`, the scanner continues processing the remainder of the line, allowing `@main` to be detected

**`Tests/SPMBuildCoreTests/MainAttrDetectionTests.swift`:**
- Corrected the existing regression test data (`/* @main` → `*/ @main`) to match the actual bug scenario
- Removed `withKnownIssue` wrapper since the fix resolves the defect
- Added a new test case for nested comment opening (`/* @main` inside an existing block comment, expected: `false`)

## Testing

All existing tests continue to pass with the new implementation. The previously-failing regression test now passes.

### DIFF
--- a/Sources/SPMBuildCore/MainAttrDetection.swift
+++ b/Sources/SPMBuildCore/MainAttrDetection.swift
@@ -25,17 +25,31 @@ package func containsAtMain(fileSystem: FileSystem, path: AbsolutePath) throws -
         if line.hasPrefix("//") {
             continue
         }
-        if line.hasPrefix("/*") {
-            multilineComment = true
-        }
-        if line.hasSuffix("*/") {
-            multilineComment = false
-        }
-        if multilineComment {
-            continue
-        }
-        if line.hasPrefix("@main") {
-            return true
+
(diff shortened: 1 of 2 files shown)


william-laverty:
Closing — no longer contributing to swiftlang repos. Thanks!

PR #9733, opened 2026-02-14
Simple cmake build script for SwiftPM

dschaefer2:
To help test CMake changes, added a simple bash script that builds all the dependencies from source out of the checkouts directory in the scratch path and then build SwiftPM. This is using whatever CMake and Ninja you have on your path, which should be fine.

I needed to swap swift-crypto and swift-certificates to make this work.


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

jakepetroules (commented):
Instead of a bash script, we should do this as a build plugin the same way it was done for Swift Build: https://github.com/swiftlang/swift-build/blob/main/Plugins/cmake-smoke-test/cmake-smoke-test.swift

dschaefer2:
> Instead of a bash script, we should do this as a build plugin the same way it was done for Swift Build: https://github.com/swiftlang/swift-build/blob/main/Plugins/cmake-smoke-test/cmake-smoke-test.swift

Ah, that makes sense. I'll dig more.

dschaefer2:
Closing this. We'll do one like SwiftBuild has done.

PR #9753, opened 2026-02-19
Some command-line parity fixes when using swift-build

daveinglis:
- support -debug-info-format option
- support --omit-frame-pointers option

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -972,11 +972,6 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
             }
         }
 
-        if buildParameters.driverParameters.codesizeProfileEnabled {
-            // dSYM generation is required to attribute code size to source locations
-            settings["DEBUG_INFORMATION_FORMAT"] = "dwarf-with-dsym"
-        }
-
         // Optionally also set the list of architectures to build for.
         if let architectures = buildParameters.architectures, !architectures.isEmpty {
             settings["ARCHS"] = architectures.joined(separator: " ")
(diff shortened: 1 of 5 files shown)


owenv (approved):
LGTM, it would be good to have a really basic PIF generation test for these. Especially the debugInfoFormat one, to ensure `-c release` and `-c debug` translate to the right defaults

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -1104,11 +1104,28 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
 
     private static func constructDebuggingSettingsOverrides(from parameters: BuildParameters.Debugging) -> [String: String] {
         var settings: [String: String] = [:]
-        // TODO: debugInfoFormat: https://github.com/swiftlang/swift-build/issues/560
         if parameters.shouldEnableDebuggingEntitlement {
             settings["DEPLOYMENT_POSTPROCESSING"] = "NO"
         }
-        // TODO: omitFramePointer: https://github.com/swiftlang/swift-build/issues/561
+        // Set DEBUG_INFORMATION_FORMAT based on debugInfoFormat
+        switch parameters.debugInfoFormat {
```
bkhouri, line 1108: issue: can we add an automated test that verifies the correct build suggesting are sent to SwiftBuild?

daveinglis: done

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -997,6 +992,11 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
         try settings.merge(Self.constructTestingSettingsOverrides(from: buildParameters.testingParameters), uniquingKeysWith: reportConflict)
         try settings.merge(Self.constructAPIDigesterSettingsOverrides(from: buildParameters.apiDigesterMode), uniquingKeysWith: reportConflict)
 
+        if buildParameters.driverParameters.codesizeProfileEnabled {
```
jakepetroules, line 1026: What happens if `parameters.debugInfoFormat` is dwarf or codeview? We should either error out, or warn users that incompatible options were specified and which values we decided to choose.

daveinglis: the default is always `.dwarf`, I wonder if it should support `debugInfoFormat`  as an optional so that we can tell if it was set by the user and also allow for build system level default to be different (like maybe on windows it should be codeview?)

jakepetroules: > support debugInfoFormat as an optional

Yes, that is a good idea, let's do that.

That said, I wasn't talking about what the defaults are, I was saying: what happens if the user explicitly specifies codesize profiling is enabled but then specifies debug info format as codeview or dwarf? Those options are incompatible as each option will try to set DEBUG_INFORMATION_FORMAT to a different value, and which one "wins" is ill-defined.

daveinglis: So my thinking was to add a warning message stating the debug format was overriden due to the code size profiling being enabled, but the optional change would be needed so that I only emit it if the user also ask for a incompatible debug format.

jakepetroules: That sounds good

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

daveinglis:
@swift-ci test macos

daveinglis:
@swift-ci test windows

daveinglis:
@swift-ci test windows

daveinglis:
@swift-ci test windows

PR #9754, opened 2026-02-19
Don't set OBJC_ARC build setting

daveinglis:
- build system already knows correct defaults for each supported
  platform

closes: https://github.com/swiftlang/swift-package-manager/issues/9750


owenv:
This should really be restricted to Apple platforms only, and the build system already has this:
```
 // Add ObjC ARC support for Apple platforms.
            if let project, project.isPackage, variant.llvmTargetTripleVendor == "apple" {
                sdkTable.push(BuiltinMacros.CLANG_ENABLE_OBJC_ARC, literal: true)
            }
```

So I think we should drop use of this setting in PIF generation altogether

daveinglis:
> This should really be restricted to Apple platforms only, and the build system already has this:
> 
> ```
>  // Add ObjC ARC support for Apple platforms.
>             if let project, project.isPackage, variant.llvmTargetTripleVendor == "apple" {
>                 sdkTable.push(BuiltinMacros.CLANG_ENABLE_OBJC_ARC, literal: true)
>             }
> ```
> 
> So I think we should drop use of this setting in PIF generation altogether

Happy to remove this all together... update in comming

owenv (approved):

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

daveinglis:
@swift-ci test windows

PR #9759, opened 2026-02-23
Manifest parsing

DougGregor:
Introduces a new package manifest loader that parses declarative manifests using [swift-syntax](https://github.com/swiftlang/swift-syntax) and evaluates the syntax tree directly.

### Motivation:

The current package manifest loader calls the compiler to build the package manifest into a program, then runs that program and parses the JSON output describing the package. This allows arbitrary code execution in the package manifest, which is useful, but can be somewhat slow due to the cost of all of these steps. Many package manifests are are simple and fit into a declarative style: for those, we can parse and directly evaluate the package manifests, producing the same results in-process more quickly. Doing so can improve the performance of the Swift package manager, especially on first fetch of a package manifest when we don't have the caches warmed up.

### Modifications:

This pull request introduces a new manifest loader, `ParsingManifestLoader`, that conforms to the existing `ManifestLoaderProtocol`. The `ParsingManifestLoader` uses [swift-syntax](https://github.com/swiftlang/swift-syntax) to parse the manifest directly, and then walks the resulting syntax tree to find the package declaration, its targets, dependencies, and so on. For anything in the syntax tree that the parsing manifest loader does not recognize, it records a "limitation" associated with that syntax tree node. The presence of any limitations means that the result of the parsing manifest loader should not be trusted, in which case we can fall back to the existing manifest loader. One can request that the limitations, if found, are printed to the terminal using the experimental flag `--experimental-show-manifest-parser-limitations`. This can be used to determine why a particular package manifest isn't handled by the parsing manifest loader. We could use this information to expand the capabilities of the parsing manifest loader in the future, so that more manifests would benefit from it as a fast path.

The parsing manifest loader takes advantage of the `SwiftIfConfig` library to handle `#if` in the package manifest. It queries the compiler frontend to determine the configuration of the compiler that would be used to build the manifest so we get the same results from `#if`. Additionally, the parsing manifest loader maintains a static version of the `Context` that is available to the package manifest, so it can substitute (e.g.) the package directory, Git information, or environment information into the package manifest.

The large amount of churn in the package loading tests are because most of the tests have been switched from just using the existing manifest loader to iterating over both loaders (using `forEachManifestLoader`). This ensures that both loaders pass the same set of tests. Additionally, we check that the resulting manifests from both loaders matches exactly (by encoding them into JSON) to ensure that the results are the same. This way, we can keep the manifest loaders in sync for the subset of package manifests that are meant to work.

This also adds an experimental command-line option, `--experimental-manifest-processing-mode`, that enables the parsing manifest loader. It must be provided with one of the following options:
* `only-parsed`: only use the parsing loader. This is not correct, because manifests can require execution, but can be used for testing.
* `only-executed`: only use the executing loader, which is the normal one. This is the default.
* `parsed-with-fallback`: try using the parsing loader. If it encounters limitations, fall back to the executing loader without failing. This option is what can provide build-time performance improvements when the parsing manifest loader takes over.
* `crosscheck`: try both the parsing and the executing loaders and compare the results to ensure they match. If the parsing loader encounters limitations, no further checking will be done. On success, it will print the relative times. If the results differ, it will print the JSON form of the manifest from each parser.

### Result:

The parsing manifest loader is significantly faster than the existing manifest loader (when it misses the cache), about 2-3 orders of magnitude, with package manifests parsing in ~1ms. In its current form, it handles ~87% of the package manifests in the Swift Package Index as declarative manifests. The remaining package manifests encounter limitations, so SwiftPM can silently fall back to the executing manifest loader. There are no cross-checking failures in the packages tests in the package index.

This pull request does not change the behavior of SwiftPM. The parsing manifest loader is kept experimental, and can be enabled by the  `--experimental-manifest-processing-mode` command-line option. If we decide that we want to enable it, a subsequent pull request could change the default to `parsed-with-fallback`, which provides the performance win for declarative manifests (that ~87% mentioned above) without breaking manifests that need to be executable.

### DIFF
--- a/Sources/CMakeLists.txt
+++ b/Sources/CMakeLists.txt
@@ -8,6 +8,11 @@
 
 add_compile_definitions(USE_IMPL_ONLY_IMPORTS)
 
+# The CMake build does not support the parsing manifest loader due to
+# the bootstrap process intermingling host runtimes with the built
+# runtimes.
+add_compile_definitions(DISABLE_PARSING_MANIFEST_LOADER)
+
 add_subdirectory(_AsyncFileSystem)
 add_subdirectory(Basics)
 add_subdirectory(BinarySymbols)
--- a/Sources/PackageLoading/ManifestParsing/ChainedParsingManifestLoader.swift
+++ b/Sources/PackageLoading/ManifestParsing/ChainedParsingManifestLoader.swift
@@ -1,0 +1,230 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2014-2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Basics
+import Dispatch
+import Foundation
+import PackageModel
+import SourceControl
+
+public import SwiftDiagnostics
+import SwiftParser
+import SwiftParserDiagnostics
+import SwiftSyntax
+
+import struct TSCBasic.ByteString
+
+import struct TSCUtility.Version
+
+/// Manifest loader that chains together a parsing manifest loader (which
+/// parses the manifest source directly) and an another manifest loader
+/// (e.g., that executes the manifest). It can use the parsing manifest
+/// loader when that succeeds, or cross-check the results of the parsing
+/// manifest loader against the other manifest loader to verify that they
+/// produce the same results.
+public final class ChainedParsingManifestLoader: ManifestLoaderProtocol {
+    let parsingLoader: ParsingManifestLoader
+    let executingLoader: any ManifestLoaderProtocol
+
+    /// Whether to show the limitations that prevent us from using the
+    /// results of the parsing loader.
+    let showLimitations: Bool
+
+    /// Whether to cross-check the results of the two loaders. Otherwise,
+    /// the results will be taken from the parsing loader if it succeeds,
+    /// and the executing loader otherwise.
+    let crosscheck: Bool
+
+    public init(
+        parsingLoader: ParsingManifestLoader,
+        executingLoader: any ManifestLoaderProtocol,
+        showLimitations: Bool,
+        crosscheck: Bool
+    ) {
+        self.parsingLoader = parsingLoader
+        self.executingLoader = executingLoader
+        self.showLimitations = showLimitations
+        self.crosscheck = crosscheck
+    }
+
+    public func load(
+        manifestPath: AbsolutePath,
+        manifestToolsVersion: ToolsVersion,
+        packageIdentity: PackageIdentity,
+        packageKind: PackageReference.Kind,
+        packageLocation: String,
+        packageVersion: (version: Version?, revision: String?)?,
+        identityResolver: IdentityResolver,
+        dependencyMapper: DependencyMapper,
+        fileSystem: FileSystem,
+        observabilityScope: ObservabilityScope,
+        delegateQueue: DispatchQueue
+    ) async throws -> Manifest {
+        if crosscheck {
+            // When cross-checking, we're doing timings. Pre-load the file
+            // to warm the filesystem cache rather than charging it to
+            // a particular loader.
+            let manifestContents: ByteString
+            do {
+                manifestContents = try fileSystem.readFileContents(manifestPath)
+            } catch {
+                throw ManifestParserError.inaccessibleManifest(path: manifestPath, reason: String(describing: error))
+            }
+            _ = manifestContents
+        }
+
+        // Parse the manifest directly with the parsing loader.
+        let parsedManifest: Manifest?
+        let parsingStartTime = DispatchTime.now()
+        let parsingEndTime: DispatchTime
+        do {
+            let manifest = try parsingLoader.load(
+                manifestPath: manifestPath,
+                manifestToolsVersion: manifestToolsVersion,
+                packageIdentity: packageIdentity,
+                packageKind: packageKind,
+                packageLocation: packageLocation,
+                packageVersion: packageVersion,
+                identityResolver: identityResolver,
+                dependencyMapper: dependencyMapper,
+                fileSystem: fileSystem,
+                observabilityScope: observabilityScope,
+                delegateQueue: delegateQueue
+            )
+
+            // We successfully parsed the manifest. If we aren't
+            // cross-checking the results, we're done.
+            if !crosscheck {
+                return manifest
+            }
+
+            parsedManifest = manifest
+            parsingEndTime = DispatchTime.now()
+        } catch {
+            guard case .limitations(let limitations) = error else {
+                throw error
+            }
+
+            // We hit a limitation of the parsed approach, which can either
+            // be a missing feature or something executable in the manifest.
+            parsingEndTime = DispatchTime.now()
+
+            if showLimitations {
+                print("Manifest parser encountered \(limitations.count) limitations that prevent its use for '\(manifestPath.pathString)'")
+
+                let formatter = DiagnosticsFormatter()
+                let filename = manifestPath.pathString
+                let locationConverter = SourceLocationConverter(
+                    fileName: filename,
+                    tree: limitations[0].syntax.root
+                )
+                for limitation in limitations {
+                    let diagLoc = locationConverter.location(
+                        for: limitation.syntax.position
+                    )
+                    let prefix = "\(filename):\(diagLoc.line):\(diagLoc.column):"
+                    let message = formatter.formattedMessage(limitation)
+
+                    let source = formatter.annotatedSource(
+                        tree: limitation.syntax.root,
+                        diags: [limitation.asDiagnostic()]
+                    )
+
+                    print(
+                        "\(prefix) \(message)\n\(source)"
+                    )
+                }
+            }
+            parsedManifest = nil
+        }
+
+        // Use the executing loader to process the manifest.
+        let executingStartTime = DispatchTime.now()
+        let executedManifest = try await executingLoader.load(
+            manifestPath: manifestPath,
+            manifestToolsVersion: manifestToolsVersion,
+            packageIdentity: packageIdentity,
+            packageKind: packageKind,
+            packageLocation: packageLocation,
+            packageVersion: packageVersion,
+            identityResolver: identityResolver,
+            dependencyMapper: dependencyMapper,
+            fileSystem: fileSystem,
+            observabilityScope: observabilityScope,
+            delegateQueue: delegateQueue
+        )
+        let executingEndTime = DispatchTime.now()
+
+        let parsingDuration = parsingStartTime.distance(to: parsingEndTime)
+        let executingDuration = executingStartTime.distance(to: executingEndTime)
+
+        // If we have a parsed manifest, it means that we want to
+        // cross-check the results. Do so now.
+        if let parsedManifest {
+            precondition(crosscheck)
+
+            let parsedJSON = try parsedManifest.toJSON()
+            let executedJSON = try executedManifest.toJSON()
+            guard parsedJSON == executedJSON else {
+                print("""
+                    Manifest loading cross-check failed for '\(manifestPath.pathString)':
+                      - Parsing took \(parsingDuration)
+                      - Executing took \(executingDuration)
+                    """
+                )
+                throw ChainedParsingError.manifestMismatch(
+                    manifestPath: manifestPath.pathString,
+                    parsed: parsedJSON,
+                    executed: executedJSON
+                )
+            }
+
+            print("""
+                Manifest loading cross-check succeeded for '\(manifestPath.pathString)':
+                  - Parsing took \(parsingDuration)
+                  - Executing took \(executingDuration)
+                """
+            )
+            return parsedManifest
+        } else {
+            print("""
+                Manifest loading encountered limitations for '\(manifestPath.pathString)':
+                  - Parsing took \(parsingDuration)
+                  - Executing took \(executingDuration)
+                """
+            )
+        }
+
+        return executedManifest
+    }
+
+    public func resetCache(observabilityScope: Basics.ObservabilityScope) async {
+        await parsingLoader.resetCache(observabilityScope: observabilityScope)
+        await executingLoader.resetCache(observabilityScope: observabilityScope)
+    }
+
+    public func purgeCache(observabilityScope: Basics.ObservabilityScope) async {
+        await parsingLoader.purgeCache(observabilityScope: observabilityScope)
+        await executingLoader.purgeCache(observabilityScope: observabilityScope)
+    }
+
+    enum ChainedParsingError: Error, CustomStringConvertible {
+        case manifestMismatch(manifestPath: String, parsed: String, executed: String)
+
+        var description: String {
+            switch self {
+            case .manifestMismatch(manifestPath: let manifestPath, parsed: let parsed, executed: let expected):
+                "The manifest produced by parsing '\(manifestPath)' does not match the one produced by executing: \(parsed) != \(expected)"
+            }
+        }
+    }
+}
--- a/Sources/PackageLoading/ManifestParsing/ChainedParsingManifestLoader.swift
+++ b/Sources/PackageLoading/ManifestParsing/ChainedParsingManifestLoader.swift
@@ -10,6 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if !DISABLE_PARSING_MANIFEST_LOADER
 import Basics
 import Dispatch
 import Foundation
@@ -228,3 +229,4 @@ public final class ChainedParsingManifestLoader: ManifestLoaderProtocol {
         }
     }
 }
+#endif
--- a/Sources/PackageLoading/ManifestParsing/ManifestParserError.swift
+++ b/Sources/PackageLoading/ManifestParsing/ManifestParserError.swift
@@ -1,0 +1,52 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2014-2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Basics
+import Foundation
+import SwiftDiagnostics
+
+/// An error that can be produced when parsing a manifest directly.
+public enum ManifestParserError: Error {
+    /// The manifest parser encountered known limitations and cannot produce
+    /// a complete manifest.
+    case limitations([ManifestParseLimitation])
+
+    /// The parser encountered syntactic errors when parsing the manifest.
+    case syntaxErrors([SwiftDiagnostics.Diagnostic])
+
+    /// The manifest file could not be loaded.
+    case inaccessibleManifest(path: AbsolutePath, reason: String)
+
+    /// The manifest file is missing a package name.
+    case missingPackageName
+}
+
+extension ManifestParserError: CustomStringConvertible {
+    public var description: String {
+        switch self {
+        case .limitations(let limitations):
+            return "Manifest parser encountered limitations\n" + limitations.map(\.description).joined(separator: "\n")
+        case .syntaxErrors:
+            return "Syntax errors in manifest"
+        case .inaccessibleManifest(path: let path, reason: let reason):
+            return "Could not read package manifest at \(path): \(reason)"
+        case .missingPackageName:
+            return "Could not find the package name"
+        }
+    }
+}
+
+extension ManifestParserError: LocalizedError {
+    public var errorDescription: String? {
+        description
+    }
+}
--- a/Sources/PackageLoading/ManifestParsing/ManifestParserError.swift
+++ b/Sources/PackageLoading/ManifestParsing/ManifestParserError.swift
@@ -10,6 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if !DISABLE_PARSING_MANIFEST_LOADER
 import Basics
 import Foundation
 import SwiftDiagnostics
@@ -50,3 +51,4 @@ extension ManifestParserError: LocalizedError {
         description
     }
 }
+#endif
--- a/Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
+++ b/Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
@@ -1,0 +1,3290 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2014-2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Basics
+import Dispatch
+import Foundation
+import PackageModel
+import SourceControl
+
+public import SwiftDiagnostics
+import SwiftIfConfig
+import SwiftOperators
+import SwiftParser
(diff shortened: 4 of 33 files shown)


dschaefer2:
Fascinating :). How different is this from letting `swift-frontend --interpret` process the manifest?

DougGregor:
> Fascinating :). How different is this from letting `swift-frontend --interpret` process the manifest?

It's a complete reimplementation as a parse + a walk over the syntax tree within the process.

`swift-frontend --interpret` is parsing + type checking + code generation through SIL and IR + JIT'ing + JSON serialization + JSON deserialization in a separate process.

The fact that it's a complete reimplementation means that it's a duplicate path to maintain, so it's only worthwhile if it (1) covers a significant fraction of the manifests in the wild and (2) is significantly faster in general.

Sources/PackageModel/Manifest/Manifest.swift
```diff
@@ -664,8 +664,15 @@ extension Manifest: Encodable {
...
+        try container.encode(self.traits.sorted(by: { $0.name < $1.name }), forKey: .traits)
         try container.encode(self.platforms, forKey: .platforms)
         try container.encode(self.packageKind, forKey: .packageKind)
     }
+
+    public func toJSON() -> String {
+        let encoder = JSONEncoder()
+        encoder.outputFormatting = [ .prettyPrinted, .sortedKeys ]
+        let data = try! encoder.encode(self)
+        return String(data: data, encoding: .utf8)!
```
ankushkushwaha, line 676: Shall we avoid force unwrapping here?

DougGregor: Yeah, sure

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
Ah, tests are failing because the underlying compiler is Swift 6.2, which didn't support the `-print-static-build-config` option needed by the parsing manifest loader. The tests need to be robust against this for now.

dschaefer2:
> Ah, tests are failing because the underlying compiler is Swift 6.2, which didn't support the `-print-static-build-config` option needed by the parsing manifest loader. The tests need to be robust against this for now.

Ah, yes. We should be moving to 6.3 when it's released.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please smoke test

DougGregor:
@swift-ci please smoke test

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3063 @@
...
+/// manifest execution from succeeding, for example syntax errors in the
+/// manifest. The client can choose to report these errors directly rather
+/// than continue to manifest parsing.
+public final class ParsingManifestLoader: ManifestLoaderProtocol {
+    let pruneDependencies: Bool
+    let config: StaticBuildConfiguration
+    let environment: [String: String]?
+
+    public init(
+        toolchain: UserToolchain,
```
plemarquand, line 67: Can we supply a `StaticBuildConfiguration` instead of a `UserToolchain`? It would be nice to be able to run the ParsingManifestLoader without the need to have/create a UserToolchain.

Sources/PackageLoading/ManifestParsing/ManifestParserError.swift
```diff
@@ -0,0 +1,52 @@
...
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Basics
+import Foundation
+import SwiftDiagnostics
+
+/// An error that can be produced when parsing a manifest directly.
+public enum ManifestParserError: Error {
```
plemarquand, line 19: nit: This is the only folder in the project that has a space in the name. For consistency can you rename to `ManifestParsing`?

Sources/PackageModel/EnabledTrait.swift
```diff
@@ -564,8 +564,8 @@ extension String: EnabledTraitConvertible {
 // MARK: - Collection + EnabledTrait
 
 extension Collection where Element == EnabledTrait {
-    public var names: Set<String> {
-        Set<String>(self.map(\.name))
+    public var names: [String] {
```
plemarquand: The [equality comparison](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/PackageModel/EnabledTrait.swift#L460) for `EnabledTraits` compares one `names` against another. Now that this is `[String]` instead of `Set<String>` that comparison becomes order dependent. I'm not positive this will cause an issue in SPM or other clients, but it could be potentially subtle bug.

Sources/PackageLoading/Manifest Parsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3063 @@
...
+                    }
+                } else {
+                    limitations.append(.unsupportedExpression(argument.expression, expected: "known library type"))
+                }
+            } else {
+                // For Apple product types, additional labels are handled below.
+                // On non-Apple builds any unrecognized label is a limitation.
+                #if ENABLE_APPLE_PRODUCT_TYPES
+                if label == "bundleIdentifier" {
+                    bundleIdentifier = argument.expression.asStringLiteralValueWithContextEvaluation(contextModel: contextModel)
```
plemarquand: I'm not sure if I'm missing something but when I build with `ENABLE_APPLE_PRODUCT_TYPES=1 swift build` I get `error: value of type 'ExprSyntax' has no member 'asStringLiteralValueWithContextEvaluation'`

Sources/PackageLoading/Manifest Parsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3063 @@
...
+        var bonjourServiceTypes: [String]?
+        var condition: ProductSetting.IOSAppInfo.DeviceFamilyCondition?
+        
+        for argument in arguments {
+            let label = argument.label?.text
+            
+            if label == "purposeString" {
+                purposeString = argument.expression.asStringLiteralValueWithContextEvaluation(contextModel: contextModel)
+            } else if label == "bonjourServiceTypes" {
+                bonjourServiceTypes = argument.expression.asStringArray()
```
plemarquand: `ENABLE_APPLE_PRODUCT_TYPES=1 swift build` produces `error: missing argument for parameter 'in' in call` for me here.

Sources/PackageLoading/Manifest Parsing/ManifestParseLimitation.swift
```diff
@@ -0,0 +1,136 @@
...
+import SwiftSyntax
+
+/// A description of a limitation of the manifest parser, such as an
+/// unrecognized syntax node or some kind of dynamic expression that the
+/// manifest parser does not understand.
+///
+/// Every limitation will have a corresponding syntax node and can be
+/// treated as a `DiagnosticMessage` so that it can be shown to the user.
+public enum ManifestParseLimitation {
+    /// An nexpected syntax node of any kind.
```
plemarquand: nit: 
```suggestion
    /// An unexpected syntax node of any kind.
```

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
DougGregor: Yes, of course!

Sources/PackageModel/EnabledTrait.swift
DougGregor: Hmm. I haven't seen this be an issue, but I can fix it.

Sources/PackageLoading/Manifest Parsing/ParsingManifestLoader.swift
DougGregor: No, this is me forgetting to build with ENABLE_APPLE_PRODUCT_TYPES after refactoring stuff. I'll fix

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please smoke test

DougGregor:
@swift-ci please smoke test

DougGregor:
@swift-ci please test

DougGregor:
Rerunning tests because I forgot what broke.

@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please smoke test

DougGregor:
@swift-ci please test macOS

DougGregor:
@swift-ci please test Linux

DougGregor:
@swift-ci bootstrap

DougGregor:
@swift-ci test bootstrap

0xpablo:
Hi there, thanks a lot for this. We tested this out in my company where we have ~200 local packages in a monorepo + ~100 remote packages and went from a bit over 800s to around 300s on a clean package resolution. Next test I want to make is trying out moving everything remote to a registry which I expect will cut that down significantly.
 
Half of the wins for us come from the fact that the `GitInformation` is now lazy and we don't use it, and this was very very expensive even for local packages. And the rest from the faster parsing. 
I can run more serious tests next week if they are helpful. I ran some experiments as I opened [a PR](https://github.com/swiftlang/swift-package-manager/pull/9794) a few days ago to avoid expensive redundant `GitInformation` calls and then found this and tested these changes out and they are super promising 👏

DougGregor:
> Hi there, thanks a lot for this. We tested this out in my company where we have ~200 local packages in a monorepo + ~100 remote packages and went from a bit over 800s to around 300s on a clean package resolution.

@0xpablo that is amazing, thank you for trying this out!

DougGregor:
Rebased and collapsed a bunch of unnecessary history. The primary semantic change from before the rebase is that this now sits on top of the package manifest nondeterminism fix split out into https://github.com/swiftlang/swift-package-manager/pull/9860. That makes this pull request implement *only* the parsing manifest loader and its tie-ins, without other semantic changes to the existing package manifest format or how it is processed.

DougGregor:
@swift-ci please test

DougGregor:

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please smoke test

DougGregor:
Did a full review of the manifest loader, fixing up any remaining places where we were failing to check arguments

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please clean smoke test Linux

DougGregor:
@swift-ci please clean smoke test macOS

DougGregor:
@swift-ci please clean smoke test Windows

DougGregor:
@swift-ci please smoke test

DougGregor:
@swift-ci please test

DougGregor:

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:

DougGregor:
@swift-ci please test

DougGregor:

DougGregor:

DougGregor:

DougGregor:
https://github.com/swiftlang/swift-syntax/pull/3311

@swift-ci please test Windows

DougGregor:
https://github.com/swiftlang/swift-syntax/pull/3311

@swift-ci please test

DougGregor:
https://github.com/swiftlang/swift-syntax/pull/3311

@swift-ci please test Windows

DougGregor:
https://github.com/swiftlang/swift-syntax/pull/3311

@swift-ci please test Windows

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:

DougGregor:
@swift-ci please test

DougGregor:

DougGregor:
@swift-ci please test

DougGregor:

DougGregor:
@swift-ci please test

DougGregor:

DougGregor:
@swift-ci please test

DougGregor:

DougGregor:

DougGregor:
@swift-ci test self hosted amazon linux 2

DougGregor:
@swift-ci test self hosted amazon

DougGregor:
@swift-ci test self hosted

DougGregor:
@swift-ci test self hosted amazon

DougGregor:
https://github.com/swiftlang/sourcekit-lsp/pull/2596

@swift-ci please test Windows

DougGregor:
https://github.com/swiftlang/sourcekit-lsp/pull/2596

@swift-ci please test Windows

DougGregor:
@swift-ci test self hosted Linux

DougGregor:
@swift-ci test self hosted

DougGregor:
https://github.com/swiftlang/sourcekit-lsp/pull/2596

@swift-ci please test Windows

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test self hosted Windows

Sources/PackageLoading/ManifestParsing/ChainedParsingManifestLoader.swift
```diff
@@ -0,0 +1,232 @@
...
+import SwiftParser
+import SwiftParserDiagnostics
+import SwiftSyntax
+
+import struct TSCBasic.ByteString
+
+import struct TSCUtility.Version
+
+/// Manifest loader that chains together a parsing manifest loader (which
+/// parses the manifest source directly) and an another manifest loader
```
owenv, line 30:
```suggestion
/// parses the manifest source directly) and another manifest loader
```

Sources/PackageLoading/ManifestParsing/ChainedParsingManifestLoader.swift
```diff
@@ -0,0 +1,232 @@
...
+        packageLocation: String,
+        packageVersion: (version: Version?, revision: String?)?,
+        identityResolver: IdentityResolver,
+        dependencyMapper: DependencyMapper,
+        fileSystem: FileSystem,
+        observabilityScope: ObservabilityScope,
+        delegateQueue: DispatchQueue
+    ) async throws -> Manifest {
+        if crosscheck {
+            // When cross-checking, we're doing timings. Pre-load the file
```
owenv, line 74: Is this a meaningful speedup in practice? There's no guarantee it will stay in the cache when loading a few dozen manifests in parallel

Sources/PackageLoading/ManifestParsing/ChainedParsingManifestLoader.swift
```diff
@@ -0,0 +1,232 @@
...
+                    )
+                    let prefix = "\(filename):\(diagLoc.line):\(diagLoc.column):"
+                    let message = formatter.formattedMessage(limitation)
+
+                    let source = formatter.annotatedSource(
+                        tree: limitation.syntax.root,
+                        diags: [limitation.asDiagnostic()]
+                    )
+
+                    print(
```
owenv, line 143: Can this be emitted through the observability scope instead of sent to stdout?

Sources/PackageLoading/ManifestParsing/ChainedParsingManifestLoader.swift
```diff
@@ -0,0 +1,232 @@
...
+
+        // If we have a parsed manifest, it means that we want to
+        // cross-check the results. Do so now.
+        if let parsedManifest {
+            precondition(crosscheck)
+
+            let parsedJSON = try parsedManifest.toJSON()
+            let executedJSON = try executedManifest.toJSON()
+            guard parsedJSON == executedJSON else {
+                print("""
```
owenv, line 179: Same note about using the observabilityscope

Sources/PackageLoading/ManifestParsing/ChainedParsingManifestLoader.swift
```diff
@@ -0,0 +1,232 @@
...
+        await executingLoader.purgeCache(observabilityScope: observabilityScope)
+    }
+
+    enum ChainedParsingError: Error, CustomStringConvertible {
+        case manifestMismatch(manifestPath: String, parsed: String, executed: String)
+
+        var description: String {
+            switch self {
+            case .manifestMismatch(manifestPath: let manifestPath, parsed: let parsed, executed: let expected):
+                "The manifest produced by parsing '\(manifestPath)' does not match the one produced by executing: \(parsed) != \(expected)"
```
owenv, line 227: nit:
```suggestion
                "The manifest produced by parsing '\(manifestPath)' does not match the one produced by executing: '\(parsed)' != '\(expected)'"
```

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3292 @@
...
+    /// Targets
+    var targets: [TargetDescription] = []
+
+    var pkgConfig: String?
+
+    /// Swift language versions.
+    var swiftLanguageVersions: [SwiftLanguageVersion]?
+
+    /// Package dependencies.
+    ///
```
owenv, line 340:
```suggestion
```

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3292 @@
...
+              importNode.path.count == 1,
+              let moduleName = pathComponent.name.identifier
+        else {
+            limitations.append(.unsupportedImportForm(importNode))
+            return .skipChildren
+        }
+
+        // Check for module names we understand.
+        switch moduleName.name {
+        case "PackageDescription", "Foundation", "CompilerPluginSupport":
```
owenv, line 434: Maybe worth adding AppleProductTypes here as well

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3292 @@
...
+
+        return .skipChildren
+    }
+
+
+    /// Check whether import declarations match known modules. Otherwise, we
+    /// can't reason about what the manifest is doing.
+    override func visit(_ importNode: ImportDeclSyntax) -> SyntaxVisitorContinueKind {
+        // Match import declaration with a single path component.
+        guard importNode.attributes.isEmpty, importNode.modifiers.isEmpty,
```
owenv, line 423: Before getting here should we also verify that extraManifestFlags aren't customizing any search paths?

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3292 @@
...
+            // Unhandled argument.
+            limitations.append(.unsupportedArgument(argument, callee: "Package"))
+        }
+    }
+    
+    /// Parse a target declaration like .target(name: "foo", dependencies: [...])
+    private func parseTarget(_ expr: ExprSyntax) -> TargetDescription? {
+        guard let functionCall = expr.as(FunctionCallExprSyntax.self),
+              let memberAccess = functionCall.calledExpression.as(MemberAccessExprSyntax.self),
+              memberAccess.base == nil, // Leading dot syntax
```
owenv, line 696: It feels a little odd not to allow qualifying the lookup here, even if it's very rare in practice

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3292 @@
...
+                    limitations.append(.unsupportedExpression(argument.expression, expected: "publicHeadersPath string"))
+                }
+            } else if label == "pkgConfig" {
+                if let value = argument.expression.asStringLiteralValue() {
+                    pkgConfig = value
+                } else {
+                    limitations.append(.unsupportedExpression(argument.expression, expected: "pkgConfig string"))
+                }
+            } else if label == "providers" {
+                if let providersArray = argument.expression.as(ArrayExprSyntax.self) {
```
owenv, line 799: A few of these items are somewhat nontrivial to parse and can appear at the package or target scope, might be nice to share the code between them

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3292 @@
...
+    
+    /// Parse a target dependency like "dep1", .target(name: "dep2"), or .product(name: "dep3", package: "Pkg")
+    private func parseTargetDependency(_ expr: ExprSyntax) -> TargetDescription.Dependency? {
+        // Case 1: String literal dependency (e.g., "dep1")
+        if let depName = expr.asStringLiteralValue(in: contextModel) {
+            return .byName(name: depName, condition: nil)
+        }
+        
+        // Case 2: .target(name: ...) or .product(name: ..., package: ...)
+        guard let (methodName, arguments) = expr.asMemberAccessCall() else {
```
owenv, line 899: Unlike the case above, here we're not checking the base of the lookup at all

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3292 @@
...
+            return nil
+        } else {
+            limitations.append(.unsupportedExpression(expr, expected: ".target or .product dependency"))
+            return nil
+        }
+    }
+    
+    /// Parse a package condition like .when(platforms: [.macOS, .linux])
+    private func parsePackageCondition(_ expr: ExprSyntax) -> PackageConditionDescription? {
+        guard let (methodName, arguments) = expr.asMemberAccessCall(),
```
owenv, line 981: Also skipping the base of the lookup here

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3292 @@
...
+        var aliases: [String: String] = [:]
+        
+        // Check if it's a dictionary with elements
+        switch dictExpr.content {
+        case .colon:
+            // Empty dictionary [:]
+            return nil
+        case .elements(let elements):
+            // Dictionary with key-value pairs
+            for element in elements {
```
owenv, line 1055: I think we should diagnose duplicate keys here so we're not silently overwriting earlier ones with later ones

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3292 @@
...
+            return nil
+        }
+
+        // Parse product arguments
+        var name: String?
+        var targets: [String] = []
+        var productType: ProductType?
+        var settings: [ProductSetting] = []
+
+        #if ENABLE_APPLE_PRODUCT_TYPES
```
owenv, line 1791: We should either address the comment above about allowing `import AppleProductTypes` to appear, or drop this parsing code

Sources/PackageLoading/ManifestParsing/ParsingManifestLoader.swift
```diff
@@ -0,0 +1,3292 @@
...
+    fileprivate func asSwiftLanguageVersionArray() -> [SwiftLanguageVersion]? {
+        guard let arrayExpr = self.as(ArrayExprSyntax.self) else {
+            return nil
+        }
+        
+        var result: [SwiftLanguageVersion] = []
+        for element in arrayExpr.elements {
+            let expr = element.expression
+            
+            // Try to parse as member access (e.g., .v3, .v4, .v4_2)
```
owenv, line 3153: There's some duplication here with parseSwiftLanguageVersion above

Sources/CMakeLists.txt
```diff
@@ -8,6 +8,11 @@
 
 add_compile_definitions(USE_IMPL_ONLY_IMPORTS)
 
+# The CMake build does not support the parsing manifest loader due to
+# the bootstrap process intermingling host runtimes with the built
+# runtimes.
+add_compile_definitions(DISABLE_PARSING_MANIFEST_LOADER)
```
owenv, line 14: Currently the CMake build is used to produce the final installable product on Windows, so I think we'll need to solve this one way or another

dschaefer2: I agree with this. If we're doing this, we need to be all in. I need to understand the issue more.

PR #9767, opened 2026-02-24
Add --enable-codesize-profile option to emit SIL, LLVM IR, and optimization records for code size profiling.

rjmansfield:
This option is a convenience option to emit supplementary output such as Swift IL, LLVM IR and optimization records, so that they can be analyzed for code size.

Adds --codesize-profile-output-dir to control the output directory.
Handles both native and swiftbuild backends. 
Handles per-file vs module-level output placement for single-threaded and multi-threaded WMO.

### DIFF
--- a/Sources/Commands/SwiftBuildCommand.swift
+++ b/Sources/Commands/SwiftBuildCommand.swift
@@ -176,6 +176,22 @@ public struct SwiftBuildCommand: AsyncSwiftCommand {
             toolsBuildParameters.printPIFManifestGraphviz = true
         }
 
+        if swiftCommandState.options.build.enableCodesizeProfile {
+            var driverParameters = productsBuildParameters.driverParameters
+            driverParameters.codesizeProfileEnabled = true
+            driverParameters.emitSILFiles = true
+            driverParameters.emitIRFiles = true
+            driverParameters.emitOptimizationRecord = true
+
+            if let outputDir = swiftCommandState.options.build.codesizeProfileOutputDirectory {
+                let outputPath = try AbsolutePath(validating: outputDir, relativeTo: swiftCommandState.originalWorkingDirectory)
+                driverParameters.silOutputDirectory = outputPath
+                driverParameters.irOutputDirectory = outputPath
+                driverParameters.optimizationRecordDirectory = outputPath
+            }
+            productsBuildParameters.driverParameters = driverParameters
+        }
+
         do {
             try await build(
                 swiftCommandState,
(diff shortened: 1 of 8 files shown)


rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

Sources/Commands/SwiftBuildCommand.swift
```diff
@@ -120,6 +120,14 @@ struct BuildCommandOptions: ParsableArguments {
 
     @OptionGroup(title: "SBOM")
     var sbom: SBOMOptions
+
+    /// Enable code size profiling by emitting SIL, LLVM IR, and optimization records.
+    @Flag(name: .customLong("enable-codesize-profile"), help: "Generate SIL, LLVM IR, and optimization record files for code size profiling.")
```
rconnell9: Is the generation of SIL, LLVM IR records intended to be used exclusively for code side-profiling?  Just wondering about the naming here, whether it should be along the lines of "generate-optimization-records".

Sources/Commands/SwiftBuildCommand.swift
```diff
@@ -120,6 +120,14 @@ struct BuildCommandOptions: ParsableArguments {
 
     @OptionGroup(title: "SBOM")
     var sbom: SBOMOptions
+
+    /// Enable code size profiling by emitting SIL, LLVM IR, and optimization records.
```
owenv: We've been trying to encourage running swift-evolution proposals for new command line options without an `-experimental` prefix. I don't think this would be controversial, but maybe we could do something lightweight or land these in experimental form initially

owenv (approved):

Sources/Commands/SwiftBuildCommand.swift
rjmansfield: Well there are already compiler options to dump IR, SIL, opt records. This is intended as a convenience option that dumps enough intermediate output the code size profiling tool to process as well as enabling debug info so the tool can do size to source mapping.

Sources/Commands/SwiftBuildCommand.swift
rjmansfield: If that's the case, I'll add the -experimental prefix.

rconnell9 (approved):

rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci please test

rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test

rjmansfield:
@swift-ci test Windows

PR #9771, opened 2026-02-25
Fix playgrounds projects using new PIF builder

plemarquand:
The new PIF builder sets `LINKER_DRIVER="auto"` at the project level. For Swift Playgrounds projects this causes `clang` to be selected for the `__preview.dylib` link step. `clang` does not add the correct SDK library search paths for the `macabi` target triple, resulting in the error `"Library 'System' not found"`. 

Remove the setting for application packages so Swift Build uses its default linker driver selection.


plemarquand:
@swift-ci test

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
```diff
@@ -596,8 +596,19 @@ public final class PackagePIFBuilder {
...
+        // unless this is an application product package (e.g. Swift
+        // Playgrounds). With LINKER_DRIVER="auto", clang may be
+        // selected for the debug dylib link step but does not add
+        // the correct SDK library search paths for playgrounds projects.
+        let hasApplicationProduct = self.package.products.contains { product in
+            guard product.type == .executable else { return false }
+            return self.delegate.customProductType(forExecutable: product.underlying) == .application
+        }
+
+        if !hasApplicationProduct {
```
owenv, line 621: Removing auto here should cause the build system to default to using clang, so there may be some other unexpected interaction here we should root-cause

plemarquand:
@swift-ci test Linux

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
```diff
@@ -596,8 +596,19 @@ public final class PackagePIFBuilder {
...
         settings[.COMPILER_WORKING_DIRECTORY] = "$(WORKSPACE_DIR)"
 
-        // Defer to the build system for linker driver selection.
-        settings[.LINKER_DRIVER] = "auto"
+        // Defer to the build system for linker driver selection,
+        // unless this is an application product package (e.g. Swift
+        // Playgrounds). With LINKER_DRIVER="auto", clang may be
+        // selected for the debug dylib link step but does not add
+        // the correct SDK library search paths for playgrounds projects.
+        let hasApplicationProduct = self.package.products.contains { product in
```
pmattos, line 616: **Suggestion**. Maybe move this to a helper method in `PackagePIFBuilder+Helpers.swift` instead?
```swift 
extension PackageGraph.ResolvedPackage { 
    func hasApplicationProduct(delegate: PackagePIFBuilder.BuildDelegate) -> Bool {
        ...
    }
}
```

pmattos (approved):

plemarquand:
@swift-ci please test

jakepetroules (changes requested):
Let's dig into this a little more first, because with System missing I suspect we're actually missing a sysroot flag or something more obscure and it would be best to fix the root cause.

jakepetroules:
Owen's raised swiftlang/swift-build#1234 to fix the root cause.

PR #9777, opened 2026-02-25
[Swift Build] filter warning control flags from the build of remote packages

owenv:
Two changes to match SE-0480 more closely:
- Don't suppress warnings for non-root local packages
- Filter warning control flags for remote packages so they do not conflict with -suppress-warnings


owenv:
@swift-ci test

owenv:
@swift-ci test Windows

PR #9778, opened 2026-02-26
[Build] Fix --static-swift-stdlib flag being ignored on FreeBSD

### LINKED ISSUE #9072: Static linking when cross compiling to FreeBSD fails due to SwiftPM looking for `static-stdlib-args.lnk` in the wrong location

### Description

It currently isn't possible to do a static link of the Swift standard library when using a swift SDK for cross compilation:

```
$ xcrun --toolchain "Swift Development Snapshot" -- swift build --swift-sdk swift-freebsd14.3 -c release -Xswiftc -static-stdlib
[1/1] Planning build
Building for production...
SwiftDriver/GenericUnixToolchain+LinkerSupport.swift:277: Fatal error: /Users/user/Library/Developer/Toolchains/swift-6.2-DEVELOPMENT-SNAPSHOT-2025-08-21-a.xctoolchain/usr/lib/swift_static/freebsd/static-stdlib-args.lnk not found
[3/4] Linking swift-freebsd-test
```

The problem is that SwiftPM is looking for the file `static-stdlib-args.lnk` inside of the toolchain, when it should be looking inside of the SDK. For example, in my case the file exists inside of the FreeBSD SDK at this path:

```
swift-freebsd14.3.artifactbundle/swift-freebsd14.3/aarch64-unknown-freebsd14.3/usr/lib/swift_static/freebsd/static-stdlib-args.lnk
```

If I create a symbolic link to that file inside of the toolchain, then the compilation and static linking works as expected:

```
$ pwd
/Users/user/Library/Developer/Toolchains/swift-6.2-DEVELOPMENT-SNAPSHOT-2025-08-21-a.xctoolchain/usr/lib/swift_static/freebsd
$ ln -s /Users/user/Scratch/sysroot/swift-freebsd14.3.artifactbundle/swift-freebsd14.3/aarch64-unknown-freebsd14.3/usr/lib/swift_static/freebsd/static-stdlib-args.lnk static-stdlib-args.lnk
$ xcrun --toolchain "Swift Development Snapshot" -- swift build --swift-sdk swift-freebsd14.3 -c release -Xswiftc -static-stdlib
[1/1] Planning build
Building for production...
[4/4] Linking swift-freebsd-test
Build complete! (0.40s)
$ ls -l .build/aarch64-unknown-freebsd/release/swift-freebsd-test
-rwxr-xr-x  1 user  staff  8063072 Aug 27 13:43 .build/aarch64-unknown-freebsd/release/swift-freebsd-test
```

SwiftPM therefore should be fixed so that this file and other required resources are retrieved from the provided SDK, and not from the toolchain.

```
$ swift --version
Apple Swift version 6.2-dev (LLVM bc9310bd0a6ef8e, Swift a57432aae96a36b)
Build config: +assertions
```

ramonasuncion:
Fixes https://github.com/swiftlang/swift-package-manager/issues/9779 
Fixes #9072

Should OpenBSD be included there? I managed to trace it after looking at how `-static-swift-stdlib` works, and it looks like it’s working.

### DIFF
--- a/Sources/Build/BuildPlan/BuildPlan.swift
+++ b/Sources/Build/BuildPlan/BuildPlan.swift
@@ -1363,7 +1363,7 @@ func generateResourceInfoPlist(
 
 extension Basics.Triple {
     var isSupportingStaticStdlib: Bool {
-        isLinux() || arch == .wasm32
+        isLinux() || isFreeBSD() || arch == .wasm32
     }
 }
 


ramonasuncion:
😶 Now that's a lot of reviewers.

ramonasuncion:
@swift-ci test

jakepetroules:
> Should OpenBSD be included there? I managed to trace it after looking at how -static-swift-stdlib works, and it looks like it’s working.

I actually think allow-listing is the wrong approach, what we should do is just exclude Apple triples (where static linking is not permitted due to Swift being an OS component), and allow everything else.

@compnerd is currently working on supporting static linking on Windows; not sure if he has an opinion to block it temporarily or not until it's fully ready.

ramonasuncion:
> > Should OpenBSD be included there? I managed to trace it after looking at how -static-swift-stdlib works, and it looks like it’s working.
> 
> I actually think allow-listing is the wrong approach, what we should do is just exclude Apple triples (where static linking is not permitted due to Swift being an OS component), and allow everything else.
> 
> @compnerd is currently working on supporting static linking on Windows; not sure if he has an opinion to block it temporarily or not until it's fully ready.

Ah, ok. And, @etcwilde did suggest I should look at the CMakefile. Only after making this PR I see now that the `stdlib/public/Resources/` directory has linux, linux-static, wasi (`static-stdlib-args.lnk`) but no freebsd.

But, I'll keep looking into this and put this on hold until @compnerd fixes this or one of the other maintainers takes over.

Edit: I thought I found an easy solutions. 😅

etcwilde:
> Ah, ok. And, @etcwilde did suggest I should look at the CMakefile. Only after making this PR I see now that the stdlib/public/Resources/ directory has linux, linux-static, wasi (static-stdlib-args.lnk) but no freebsd.

Yep, we'll need a separate `static-stdlib-args.lnk` file for FreeBSD because of the dependency on execinfo.

Though that may not actually need that much CMake work since it looks like things are mostly set up to copy the file from the `lib_dir` anyway:
```cmake
    set(lib_dir "${SWIFT_SDK_${sdk}_LIB_SUBDIR}")
    set(static_binary_lnk_src "${SWIFT_SOURCE_DIR}/stdlib/public/Resources/${lib_dir}/static-executable-args.lnk")

    # Generate the static-executable-args.lnk file used for ELF systems (eg linux)
    set(linkfile "${lib_dir}/static-executable-args.lnk")
```
https://github.com/swiftlang/swift/blob/e7c3bf853574278b2a6b5561d56cd3b94a3005d6/stdlib/public/runtime/CMakeLists.txt#L126-L127

Sources/Build/BuildPlan/BuildPlan.swift
```diff
@@ -1363,7 +1363,7 @@ func generateResourceInfoPlist(
 
 extension Basics.Triple {
     var isSupportingStaticStdlib: Bool {
-        isLinux() || arch == .wasm32
+        isLinux() || isFreeBSD() || arch == .wasm32
```
bkhouri, line 1366: This change only impacts the native build system, which is deprecated on branch `main`.  The default build system in `main` is set to Swift Build build system, which should handle the `--static-swift-stdlib` command line argument, which should work if Swift on FreeBSB has proper static swift stdlib support.

See the [[6.3] Preview the Swift Build System Integration > Key improvements and differences ](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/swiftbuildpreview#Key-improvements-and-differences) section on the documentation that mentioned this.

PR #9782, opened 2026-02-26
[PIF Builder] Add conditional linker settings test

thePianoKid:
### Motivation:

Xcode's current PIF builder does not respect conditional linker settings. This new version of the PIF builder does respect those settings. I added a little test to avoid regressions. 

### Modifications: 

No new features were added. I added a single test that reads the flags after a `debug` and `release` build. I setup the test fixture so that there should be flags in the `debug` build, but not the `release` build. 

### Result:

This test will fail if conditional linker settings are not being respected, preventing regressions.

### DIFF
--- a/Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
+++ b/Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
@@ -345,16 +345,13 @@ struct PIFBuilderTests {
             let ldflags = targetConfig.settings[.OTHER_LDFLAGS]
             switch configuration {
             case .debug:
+               let debugFlags = try #require(ldflags, "Debug config requires OTHER_LDFLAGS")
                 #expect(
-                    ldflags?.contains("-Xlinker") == true && ldflags?.contains("-interposable") == true,
-                    "Debug config should contain -Xlinker and -interposable, but got: \(String(describing: ldflags))"
+                    debugFlags.contains("-Xlinker") && debugFlags.contains("-interposable"),
+                    "Debug config missing required flags: \(debugFlags)"
                 )
             case .release:
-                let hasInterposable = ldflags?.contains("-interposable") ?? false
-                #expect(
-                    !hasInterposable,
-                    "Release config should not contain -interposable, but got: \(String(describing: ldflags))"
-                )
+                #expect(ldflags == nil, "Release config should not have debug flags, but got \(ldflags)")
             }
         }
     }


Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
```diff
@@ -331,6 +331,33 @@ struct PIFBuilderTests {
...
+        try await withGeneratedPIF(fromFixture: "PIFBuilder/ConditionalBuildSettings") { pif, observabilitySystem in
+            let errors = observabilitySystem.diagnostics.filter { $0.severity == .error }
+            #expect(errors.isEmpty, "Expected no errors during PIF generation, but got: \(errors)")
+
+            let targetConfig = try pif.workspace
+                .project(named: "ConditionalBuildSettings")
+                .target(id: "PACKAGE-TARGET:ConditionalBuildSettings")
+                .buildConfig(named: configuration)
+
+            let ldflags = targetConfig.settings[.OTHER_LDFLAGS]
```
bkhouri: I suggest changing to use the `#require` API

```suggestion
            let ldflags = try #require(targetConfig.settings[.OTHER_LDFLAGS], "OTHER_LDFLAGs was not expected to be nil")
```

This has two benefits:
- Ensures the value is not an Optional
- stops test execution is the value is `nil`

bkhouri (commented):
Thank you very much for adding this test to ensure we don't regress. I have a comment on a way to improve the test.

Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
thePianoKid: There is actually a valid case where `lsflags` is `nil`. When `configuration` is set to `.release`, the conditional linker settings in the test fixture are not applied, so `ldflags` is `nil`. 

I'm going to tweak your suggested commit so I only check for a nil value in the `.debug` case.

thePianoKid: My tweaked commit is https://github.com/swiftlang/swift-package-manager/pull/9782/commits/8261d78e67479507860e3df18a90b24dd10cfe34

bkhouri:
@swift-ci test

thePianoKid:
GitHub Actions is down... that seems to be the reason for the CI failures. I will kick off another build once GH Actions is green again.

thePianoKid:
@swift-ci test

thePianoKid:
@swift-ci test Windows

PR #9785, opened 2026-03-02
Remove blocking wait calls in RepositoryManagerTests

owenv:
I'm seeing flakiness in these tests when run from the swift-build repo. I think it may be because the Swift Concurrency thread pool is getting blocked by the wait calls in these tests. Either way, this prevents them from timing out when I set LIBDISPATCH_COOPERATIVE_POOL_STRICT=1, so I think it's worth it.

### DIFF
--- a/Tests/SourceControlTests/RepositoryManagerTests.swift
+++ b/Tests/SourceControlTests/RepositoryManagerTests.swift
@@ -58,7 +58,7 @@ final class RepositoryManagerTests: XCTestCase {
                 XCTAssertDirectoryExists(checkoutPath)
                 XCTAssertFileExists(checkoutPath.appending("README.txt"))
 
-                try delegate.wait(timeout: .now() + 2)
+                try await delegate.wait(timeout: .now() + 2)
                 XCTAssertEqual(delegate.willFetch.map { $0.repository }, [dummyRepo])
                 XCTAssertEqual(delegate.didFetch.map { $0.repository }, [dummyRepo])
             }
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

Tests/SourceControlTests/RepositoryManagerTests.swift
```diff
@@ -874,12 +880,16 @@ fileprivate class DummyRepositoryManagerDelegate: RepositoryManager.Delegate, @u
         self._didUpdate = .init()
     }
 
-    public func wait(timeout: DispatchTime) throws {
-        switch self.group.wait(timeout: timeout) {
-        case .success:
-            return
-        case .timedOut:
-            throw StringError("timeout")
+    public func wait(timeout: DispatchTime) async throws {
```
jakepetroules, line 881: The correct way to implement non-blocking wait on DispatchGroup is using `notify`:

```swift
public func wait() async {
    await withCheckedContinuation { continuation in
        group.notify {
            continuation.resume()
        }
    }
}
```

Tests/SourceControlTests/RepositoryManagerTests.swift
```diff
@@ -518,7 +518,13 @@ final class RepositoryManagerTests: XCTestCase {
                     self.outstandingGroup.enter()
                     defer { self.outstandingGroup.leave() }
                     print("\(repository) waiting to be cancelled")
-                    XCTAssertEqual(.success, self.terminatedGroup.wait(timeout: .now() + 5), "timeout waiting on terminated signal")
+                    await withCheckedContinuation { continuation in
```
jakepetroules, line 521: The correct way to implement non-blocking wait on DispatchGroup is using `notify`:

```swift
public func wait() async {
    await withCheckedContinuation { continuation in
        group.notify {
            continuation.resume()
        }
    }
}
```

Tests/SourceControlTests/RepositoryManagerTests.swift
owenv: Main reason I didn't do this is that it drops the timeout

jakepetroules: I don't think the timeout is important, the only reason that existed was because the synchronous overloads force it to be provided.

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test Windows

Issue #9756, opened 2026-02-21
UserToolchain needs to be refactored to stop passing raw flags to Swift Build


jakepetroules:
UserToolchain aggregates flags from a number of places including Swift SDKs, into its extraFlags property, and several others.

These are incorporated into BuildParameters, which are then added to Swift Build build requests as raw flags using overriding build parameters like OTHER_CFLAGS, OTHER_SWIFT_FLAGS, and so on.

However, this approach is problematic for Swift Build, as passing raw flags interferes with Swift Build's build settings based approach to forming the underlying tool command lines. For example, the --sysroot flag is duplicated with conflicting values on some platforms.

We need to refactor UserToolchain so that all of the flags it assembles are ignored when using the Swift Build backend, since the Swift SDK and toolset properties are already handled internally by it. This task will also involve taking inventory on exactly which flags are coming through this mechanism and ensuring an appropriate alternative is selected in each case.

dschaefer2:
Speaking of Swift SDKs specifically, would it be better for Swift Build if they placed their settings in the SDKSettings file for the SDK instead of the freeform-ish toolset.json?

jakepetroules:
SDKSettings is for Apple-based SDKs (and the "built-in" SDKs on Windows, because that predated our stewardship of the build system). Its schema is defined by Apple, not subject to Swift Evolution, and subject to change at any time without notice. And a good chunk of it is actually quite freeform already. So we should not bring it to Swift SDKs.

Swift SDKs already have some structured properties for a handful of metadata values in toolset.json and swift-sdk.json, we can extend those where needed.

dschaefer2:
But it is what Swift Build speaks, no? Toolset.json isn't good enough. I'd hate to have to invent something that already exists.

dschaefer2:
I'm not saying Swift SDK creators would write the SDKSettings file. My intention is to explore SDKs in Packages and these settings would come from the build settings model we already have for targets. Swift Build would generate these when it creates the SDK target in the package.

jakepetroules:
Sounds like a cool idea (the package-defined SDKs), but that's quite a ways off and I think is off topic for this issue.

[...]

PR #9787, opened 2026-03-02
Track the source of swift BuildFlags and filter them when redundant

owenv:
The motivation behind this change was that when using the Swift Build backend, we added the Swift Testing search paths twice. First, SwiftPM inserted them via OTHER_SWIFT_FLAGS ahead of the built products dir search paths, and then Swift Build added them internally in the correct order. This broke the build of packages which build Swift Testing from source, because the macro implementation and module got mismatched.

This change adds a new BuildFlag type which tracks both the spelling of a flag and its source (default search path, Swift SDK, etc.). We can incrementally adopt this mechanism over time and use it to filter out some flags which Swift Build will compute itself based on platform/SDK/run destination. For now, I've only adopted it for Swift flags to keep the change relatively small.

### DIFF
--- a/Sources/PackageModel/BuildFlags.swift
+++ b/Sources/PackageModel/BuildFlags.swift
@@ -10,6 +10,38 @@
 //
 //===----------------------------------------------------------------------===//
 
+public struct BuildFlag: Hashable, Sendable, Encodable {
+    public var value: String
+
+    /// Describes the origin of this flag, for example if it was sourced from a Swift SDK, or added as a builtin option by SwiftPM.
+    public var source: Source?
+
+    public init(value: String, source: Source? = nil) {
+        self.value = value
+        self.source = source
+    }
+
+    public enum Source: Sendable, Hashable, Codable {
+        case defaultSwiftTestingSearchPath
+        case swiftSDK
+        case commandLineOptions
+    }
+}
+
+extension [BuildFlag] {
+    public var rawFlags: [String] {
+        return self.map(\.value)
+    }
+}
+
+extension [String] {
+    public func constructBuildFlags(source: BuildFlag.Source?) -> [BuildFlag] {
+        return self.map {
+            BuildFlag(value: $0, source: source)
+        }
+    }
+}
+
 /// Build-tool independent flags.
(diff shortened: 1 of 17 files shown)


Sources/PackageModel/BuildFlags.swift
```diff
@@ -10,6 +10,43 @@
 //
 //===----------------------------------------------------------------------===//
 
+public struct BuildFlag: ExpressibleByStringLiteral, Hashable, Sendable, Codable {
+    public var value: String
+
+    /// Describes the origin of this flag, for example if it was sourced from a Swift SDK, or added as a builtin option by SwiftPM.
+    public var source: Source?
```
jakepetroules, line 17: Why can source be `nil`?

jakepetroules:
This makes partial progress on #9756

Sources/PackageModel/BuildFlags.swift
owenv: It makes incremental adoption simpler if we don't have to tag everything with a source in one PR

owenv:
@swift-ci test

owenv:
- [ ]  BuildFlags is directly decoded from a Swift SDK, so I need to fix that

owenv:
@swift-ci test

owenv:
@swift-ci test windows

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test self hosted

owenv:
Latest push:
- drops ExpressibleByStringLiteral from BuildFlag because it was causing some bugs to creep into the tests
- Adds a deprecated BuildFlags initializer to avoid having to synchronize changes with sourcekit-lsp

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test windows

owenv:

owenv:

owenv:
@swift-ci smoke test linux

owenv:
@swift-ci test linux

owenv:
@swift-ci test linux

PR #9790, opened 2026-03-04
[Swift Build] Ensure we respect the 'copy' resource rule

owenv:
Depends on https://github.com/swiftlang/swift-build/pull/1158

Ensure we don't apply build rules to resources the package explicitly requests we copy unchanged, to align with native build system behavior


Tests/FunctionalTests/ResourcesTests.swift
```diff
@@ -61,6 +61,30 @@ struct ResourcesTests{
...
+    @Test(
+        .IssueWindowsRelativePathAssert,
+        .IssueWindowsPathTestsFailures,
+        .tags(
+            .Feature.Command.Run,
+        ),
+    )
+    func swiftBuildResourceRules() async throws {
+        let config = BuildConfiguration.debug
+        try await withKnownIssue(isIntermittent: true) {
```
bkhouri: Does this actually fail in the windows self hosted pipeline?  If not, can we remove the `withKnownIssue`?

Tests/FunctionalTests/ResourcesTests.swift
```diff
@@ -92,8 +116,6 @@ struct ResourcesTests{
 
     @Test(
         .requireHostOS(.macOS),  // originally macOS only
-        // .skipHostOS(.linux), // currently failing on Ubuntu
```
bkhouri, line 95: praise: thanks for cleaning this up.

Tests/FunctionalTests/ResourcesTests.swift
owenv: I believe this does hit path length limits but I can double check in the latest run

owenv:
@swift-ci test

owenv:
@swift-ci test

Tests/FunctionalTests/ResourcesTests.swift
owenv: I wasn't thinking, this test actually needs to require a macOS host OS since it runs actool

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test Windows self hosted

owenv:

owenv:
@swift-ci test Windows

PR #9794, opened 2026-03-06
PackageLoading: cache manifest git context per repository

0xpablo:
Cache manifest git context per repository to speed up package resolution in monorepos with many local packages.

### Motivation:

Manifest evaluation currently fetches git information from each package directory independently. In monorepos with many local packages this repeats the same git operations across directories in the same repository.

### Modifications:

This change introduces `ManifestGitInformationCacheActor` to cache repository identity and git information across manifest evaluations:

- detects repository roots via .git directories and .git pointer files (worktree-aware)
- caches git information once per repository and reuses it for sibling package directories
- keeps per-directory memoization and clears cache state in `resetCache`
- routes manifest context git-information lookups through the shared cache actor

### Result:

No user-visible behavior changes are expected.
This change only deduplicates `GitInformation` lookups per repository during a single resolution.
If anything, resolution becomes more consistent: previously, two packages in the same repository could observe different GitInformation values if the repository changed between separate lookups. With this change, they share a single cached snapshot per repository for that resolution.

Here's the benchmark with and without this patch for our monorepo with ~200 local packages:

| Metric | Baseline (main) | Patched | Improvement |
|---|---:|---:|---:|
| Dependencies resolved in | 546.01s | 411.39s | -134.62s (-24.66%) |
| `git status -s` calls | 313 | 110 | -203 (-64.86%) |

### DIFF
--- a/Sources/PackageLoading/CMakeLists.txt
+++ b/Sources/PackageLoading/CMakeLists.txt
@@ -9,6 +9,7 @@
 add_library(PackageLoading
   ContextModel.swift
   Diagnostics.swift
+  ManifestGitInformationCache.swift
   ManifestJSONParser.swift
   ManifestLoader.swift
   ManifestLoader+Validation.swift
--- a/Sources/PackageLoading/ManifestGitInformationCache.swift
+++ b/Sources/PackageLoading/ManifestGitInformationCache.swift
@@ -1,0 +1,245 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2014-2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Basics
+import Dispatch
+import Foundation
+import SourceControl
+
+import struct TSCBasic.ByteString
+
+// FIXME: Use Synchronization.Mutex when available in SwiftPM's supported deployment targets.
+private final class Mutex<Value>: @unchecked Sendable {
+    var lock: NSLock
+    var value: Value
+
+    init(value: Value) {
+        self.lock = .init()
+        self.value = value
+    }
+
+    func withLock<Result>(_ body: (inout Value) -> Result) -> Result {
+        self.lock.lock()
+        defer { self.lock.unlock() }
+        return body(&self.value)
+    }
+}
+
+/// Caches repository identity and git context used by manifest loading.
+package final class ManifestGitInformationCache: @unchecked Sendable {
+    private enum ScopeKey: Hashable {
+        case repository(AbsolutePath)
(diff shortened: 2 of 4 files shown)


0xpablo:
I came across #9759 after opening this PR. I'm guessing we'd need to extract the cache to apply it to the new parser, as I understand git the overhead would be the same in the new parser.
I'll experiment creating a branch from Doug's and extracting the cache for reuse.

EDIT:
I tried looked into Doug's manifest parsing implementation in more detail and did some benchmarks and it already fetches git info only when it's used, so it probably is not worth to have any cache in the parsing path. 

As for the executor path, I'm going to benchmark alternative scenarios:
- Actor cache with and without a monorepo setup (to make sure it doesn't negatively impact perf on the non-monorepo scenario)
- An alt synchronous cache that does not serialize git operation which we might want to consider if the actor cache hurts performance

In any case, given #9759 mostly mitigates the issue, if we expect parsing to become the default and don't expect much usage of the executor path, it might also make sense to consider not shipping any of these caches. Happy to stick with what SPM maintainers decide.

0xpablo:
Follow-up after benchmarking a few alternatives on `main`.

I first looked more closely at the parser path in #9759. It already evaluates `Context.gitInformation` lazily and only fetches git info when it is actually used in the manifest, so I do not think it is worth adding a cache there.

I then benchmarked three executor-path variants with manifest DB caching disabled (`--manifest-cache none`) to avoid cross-run contamination:

- `none`: current `main`
- `actor`: the original actor-based per-repository cache from this PR
- `sync`: a synchronous cache with shared state plus per-repository in-flight coordination, so unrelated repositories can still perform git work concurrently

I tested two scenarios:

- a mixed monorepo graph
- a synthetic graph with 60 local packages, each in its own git repository

### Mixed monorepo graph

| Variant | Resolve time | `git status -s` calls |
| --- | ---: | ---: |
| `none` | `812.82s` | `313` |
| `actor` | `498.07s` | `109` |
| `sync` | `406.44s` | `109` |

So both cache implementations remove the same redundant local git work, but the synchronous version is still materially faster than the actor version (`406.44s` vs `498.07s`).

### Many distinct repositories

This graph does not have any same-repository reuse to exploit, so it is mainly useful to see whether serializing git work hurts.

| Variant | Run 1 | Run 2 | Avg |
| --- | ---: | ---: | ---: |
| `none` | `33.16s` | `28.73s` | `30.95s` |
| `actor` | `28.62s` | `32.16s` | `30.39s` |
| `sync` | `23.43s` | `28.99s` | `26.21s` |

All three variants issued the same number of git calls in this setup:

- `61` `git status -s` calls

### Conclusion

Based on these numbers:

- I do not think adding a parser-path cache is worthwhile for #9759.
- If we want to keep an executor-path cache for the period where executed manifests are still common, the actor cache is probably not the best shape.
- The synchronous cache preserved the same monorepo deduplication benefit, but also performed better than the actor cache in both scenarios.

I will update this PR to use the faster cache implementation, but as mentioned before, with no plans to add it to the parer path from #9759 as runtime there is roughly the same so happy to close this if we expect the executor path usage to be minimal in the future.

EDIT: Pushed the updated implementation

PR #9801, opened 2026-03-11
Refactor more properties to use the new BuildFlags type

jakepetroules:
This brings us further along on not passing redundant and incompatible flags to Swift Build (#9756).

There is only one remaining unannotated flag, in Sources/Build/LLBuildCommands.swift, which appears to be specific only to the legacy build system. The remaining ones are in a deprecated constructor not used within SwiftPM itself, and in tests.

### DIFF
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -581,11 +581,11 @@ public final class UserToolchain: Toolchain {
         swiftSDK: SwiftSDK,
         environment: Environment,
         fileSystem: any FileSystem
-    ) throws -> [String] {
-        var swiftCompilerFlags = swiftSDK.toolset.knownTools[.swiftCompiler]?.extraCLIOptions ?? []
+    ) throws -> [BuildFlag] {
+        var swiftCompilerFlags = (swiftSDK.toolset.knownTools[.swiftCompiler]?.extraCLIOptions ?? []).map { BuildFlag(value: $0, source: .toolset)}
 
         if let linker = swiftSDK.toolset.knownTools[.linker]?.path {
-            swiftCompilerFlags += ["-ld-path=\(linker)"]
+            swiftCompilerFlags += [BuildFlag(value: "-ld-path=\(linker)", source: .toolset)]
         }
 
         guard let sdkDir = swiftSDK.pathsConfiguration.sdkRootPath else {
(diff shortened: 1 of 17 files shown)


jakepetroules:
@swift-ci test

Sources/PackageModel/UserToolchain.swift
```diff
@@ -659,7 +667,7 @@ public final class UserToolchain: Toolchain {
                                 validating: "usr/lib/swift/windows/\(triple.archName)",
                                 relativeTo: XCTestInstallation
                             ).pathString,
-                        ]
+                        ].map { BuildFlag(value: $0, source: .defaultWindowsSettings) }
```
jakepetroules, line 670: As far as I can tell, all these testing search paths are handled correctly already.

Sources/PackageModel/UserToolchain.swift
```diff
@@ -901,22 +911,22 @@ public final class UserToolchain: Toolchain {
                             "-D_DLL",
                             "-Xclang",
                             "--dependent-lib=msvcrtd",
-                        ]
+                        ].constructBuildFlags(source: .defaultWindowsSettings)
```
jakepetroules, line 914: These Windows flags aren't fully handled, PR here: https://github.com/swiftlang/swift-build/pull/1179

jakepetroules:
@swift-ci test

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -1349,3 +1312,33 @@ fileprivate extension Triple {
...
+                // as it can break the intended search path ordering, for example, if the user is building
+                // Swift Testing as a package dependency.
+                return false
+            case .defaultWindowsSettings:
+                // Swift Build computes these internally.
+                return false
+            case .swiftSDK, .toolset:
+                // Swift Build loads Swift SDK and toolset flags internally, and may introspect them to override build
+                // settings. Don't duplicate them here.
+                return false
```
owenv, line 1337: We don't load toolsets passed to SwiftPM without a Swift SDK yet, so these might need to be split up for now

jakepetroules: In that case, note there is one occurrence of .swiftSDK that I had changed to .toolset... https://github.com/swiftlang/swift-package-manager/pull/9801/changes#diff-a434883777befb0b6ef138f44a098a3d63d3b9dc59b521e16e2c5b3a6163117fL254

jakepetroules: Ah, that's dead code, so it doesn't even matter.

jakepetroules: OK, I stopped filtering out the .toolset flags in SwiftBuildSystem

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

owenv (approved):

jakepetroules:

PR #9805, opened 2026-03-12
[Swift Build] Don't set EXECUTABLE_PREFIX when building dylib targets as frameworks

owenv:
Otherwise we can end up with a framework that looks like: X.framework/libX

### DIFF
--- a/Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
+++ b/Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
@@ -29,12 +29,13 @@ extension PIFBuilderParameters {
     static func constructDefaultParametersForTesting(
         temporaryDirectory: Basics.AbsolutePath,
         addLocalRpaths: Bool,
+        shouldCreateDylibForDynamicProducts: Bool = false,
         pluginScriptRunner: PluginScriptRunner? = nil
     ) throws -> Self {
         try self.init(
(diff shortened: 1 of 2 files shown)


owenv:
@swift-ci test

bkhouri:
@swift-ci test

bkhouri:
@swift-ci test windows

owenv:
@swift-ci smoke test Linux

owenv:
@swift-ci test linux

Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
```diff
@@ -55,6 +56,7 @@ extension PIFBuilderParameters {
 fileprivate func withGeneratedPIF(
     fromFixture fixtureName: String,
     addLocalRpaths: Bool = true,
+    shouldCreateDylibForDynamicProducts: Bool = true,
```
pmattos, line 59: **suggestion**. Given the default is `false` in `constructDefaultParametersForTesting` shouldn't we stick to `false` here as well?

owenv: hmm, yeah we might want to just generally drop a lot of the default values for this parameter, it makes it easier to make mistakes. I'll probably look at that as a follow-up

PR #9806, opened 2026-03-12
[Swift Build] Handle FREEBSD_DEPLOYMENT_TARGET

jakepetroules:
This will ensure the triple version is handled correctly by Swift Build.

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -1275,6 +1275,8 @@ fileprivate extension Triple {
             return "WATCHOS_DEPLOYMENT_TARGET"
         case (_, .android):
             return "ANDROID_DEPLOYMENT_TARGET"
+        case (.freebsd, _):
+            return "FREEBSD_DEPLOYMENT_TARGET"
         default:
             return nil
         }


jakepetroules:
@swift-ci test

owenv (approved):

jakepetroules:
In the fullness of time I think it'd be better if we could just have Swift Build extract this from the triple it receives in the Build Request, so clients don't have to understand triple parsing or know about the specific deployment target variables.

plemarquand (approved):

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -1275,6 +1275,8 @@ fileprivate extension Triple {
             return "WATCHOS_DEPLOYMENT_TARGET"
         case (_, .android):
             return "ANDROID_DEPLOYMENT_TARGET"
+        case (.freebsd, _):
+            return "FREEBSD_DEPLOYMENT_TARGET"
         default:
```
bkhouri, line 1280: Question: is there value is removing the default to to ensure SwiftPM explicitly handles all the cases?

jakepetroules: Not really, these enums are massive sets, and deployment target variables are bespoke and added sparingly. As I said in my other comment, I'd rather remove this code entirely and handle it at the swift build layer where we might be able to automate the whole thing.

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

PR #9807, opened 2026-03-12
[Swift Build] Test runner executables should be exempted from --disable-local-rpath

owenv:
Test runner executables are not currently designed to be relocatable and must be able to load the shared library in the same directory containing the actual tests. I think it makes sense to exempt them from --disable-local-rpath for now.

### DIFF
--- a/Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
+++ b/Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
@@ -682,6 +682,23 @@ struct PIFBuilderTests {
         }
     }
 
+    @Test(arguments: [true, false])
+    func testRunnerAlwaysHasLocalRpath(addLocalRpaths: Bool) async throws {
+        try await withGeneratedPIF(fromFixture: "PIFBuilder/Simple", addLocalRpaths: addLocalRpaths) { pif, observabilitySystem in
+            #expect(observabilitySystem.diagnostics.filter { $0.severity == .error }.isEmpty)
+            for configuration in BuildConfiguration.allCases {
+                let config = try pif.workspace
+                    .project(named: "Simple")
+                    .target(named: "SimpleTests-test-runner")
+                    .buildConfig(named: configuration)
+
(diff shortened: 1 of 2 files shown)


owenv:
@swift-ci test

Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
```diff
@@ -682,6 +682,23 @@ struct PIFBuilderTests {
         }
     }
 
+    @Test func testRunnerAlwaysHasLocalRpath() async throws {
+        for addLocalRpaths in [true, false] {
```
bkhouri: Add this and a test parameter to take full advantage of test parallelism

Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
```diff
@@ -682,6 +682,23 @@ struct PIFBuilderTests {
...
+            try await withGeneratedPIF(fromFixture: "PIFBuilder/Simple", addLocalRpaths: addLocalRpaths) { pif, observabilitySystem in
+                #expect(observabilitySystem.diagnostics.filter { $0.severity == .error }.isEmpty)
+                for configuration in BuildConfiguration.allCases {
+                    let config = try pif.workspace
+                        .project(named: "Simple")
+                        .target(named: "SimpleTests-test-runner")
+                        .buildConfig(named: configuration)
+
+                    // Test runners should have an rpath even if addLocalRpaths=false
+                    #expect(config.settings[.LD_RUNPATH_SEARCH_PATHS] == ["$(inherited)", "$(RPATH_ORIGIN)"])
```
bkhouri: Suggestion: consider adding the build configuration in the #expect() comment to make test failures more explicit

owenv:
@swift-ci test

owenv:
@swift-ci test Windows

jakepetroules:
I think ideally we'd just fix `swift test` to set up the environment properly, rather than add an exception.

PR #9809, opened 2026-03-13
Workaround test crash when compiling against 6.2 corelibs-foundation

owenv:
These tests sometimes crash when compiled against a foundation without the fix in https://github.com/swiftlang/swift-corelibs-foundation/pull/5269. Only run them when compiled with a 6.3 or later compiler


owenv:
@swift-ci test

owenv:
@swift-ci test self hosted macOS

owenv:
@swift-ci test macOS self hosted

jakepetroules (approved):

PR #9810, opened 2026-03-13
Add library paths so that test runner executables are able to find the corresponding test libraries

owenv:
... even if local rpaths were disabled in the build.

Alternative to https://github.com/swiftlang/swift-package-manager/pull/9807 which doesn't special case the rpaths of test runners


owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test Linux

PR #9811, opened 2026-03-13
[Swift Build] Allow clients to customize framework bundle identifier and executable name

pmattos:
### Motivation ###

Provide hooks to support customizing the *bundle identifier* and *executable name* for a given framework built with the new `--build-system swiftbuild ` option.

### Changes ###

Introduce `customBundleIdentifierPrefix(forFramework:)` on the PIF builder delegate so clients can prepend a prefix (e.g. `"com.foo.bar."`) to the bundle identifier of framework products. Also renames `customExecutableName(product:)` to `customExecutableName(forFramework:)` to clarify it only applies to frameworks.

Both overrides are applied in `configureDynamicSettings` for framework builds *only*.

Fixes rdar://172491270.


pmattos:
@swift-ci test

PR #9812, opened 2026-03-15
Add opt-in `exactLiteral` version requirements

bc-lee:
Proposal: [swiftlang/swift-evolution#3188](https://github.com/swiftlang/swift-evolution/pull/3188) 

### Motivation:

SwiftPM currently ignores SemVer build metadata when matching exact version
requirements. This change adds an opt-in way to match a full version identifier
literally, including build metadata.

### Modifications:

- Add `exactLiteral` requirement APIs for source-control and registry
  dependencies.
- Preserve the new requirement through manifest and package models.
- Update resolution to honor literal matches for metadata-distinct versions.

### Result:
Packages can now write requirements such as `.exactLiteral("1.0.0+debug")`
without changing existing `.exact(...)` or range semantics.

### DIFF
--- a/Sources/Basics/Version+Extensions.swift
+++ b/Sources/Basics/Version+Extensions.swift
@@ -23,4 +23,12 @@ extension Version {
             try? self.init(versionString: tag, usesLenientParsing: true)
         }
     }
+
+    package func literalEqual(to other: Version) -> Bool {
+        self.major == other.major &&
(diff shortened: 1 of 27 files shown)


bc-lee:
Since the proposal has been rejected, I’m closing this PR. Instead, I’m working on fixing the actual behavior of `.exact(...)` in #10410.

PR #9814, opened 2026-03-16
Set the plugin generated module map path for SwiftBuild

dschaefer2:
SwiftBuild wasn't properly tracking the module map file when generated by a plugin. This sets the MODULEMAP_PATH so it knows where it is.


dschaefer2:
@swift-ci please test

PR #9815, opened 2026-03-16
Fix toolchain initialization for Windows

jakepetroules:
There, developer dir != toolchain dir, so we were incorrectly applying CC and SWIFT_EXEC overrides.

Also remove a LIBRARY_SEARCH_PATHS; this stems from #8271 and the reason was unclear but Swift Build should already be computing this.


jakepetroules:
@swift-ci test

owenv (approved):

PR #9819, opened 2026-03-17
Allow using experimental-build-server with a user-controlled build dir

owenv:
- Add a flag to allow the build server to opt out of acquiring the workspace lock, so that manually initiated builds are not blocked by a long-lived build server which isn't being used for preparation
- Fix derivedDataPath when constructing the Swift Build build request
- Update some tests which were out of date

SourceKit-LSP defaults to background indexing with an independent build arena, but it also supports using the products of user-initiated builds. These changes are intended to support this mode when --build-system swiftbuild becomes the default


owenv:
@swift-ci test

owenv:

PR #9827, opened 2026-03-18
[Swift Build] Do not materialize static archive projects when using -multiroot-data-file

owenv:
This maintains behavioral compatibility with the native build system, the comment has some more detail


owenv:
@swift-ci test

Issue #9823, opened 2026-03-17
Swift Package Repository regression with version specific Package.swift files


cuberoot:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

I am getting an error when trying to resolve packages from a Swift Package Repository (Atifactory). In particular, the error occurs when the packages from the repository has a version specific Package.swift file (e.g. Package@swift-5.8.swift).

I am using a recent 6.3 snapshot.

I have tracked this problem down to [this PR commit](https://github.com/swiftlang/swift-package-manager/pull/9774/changes#diff-6e5a2a399d8a377f261cac16a7a00ba1b45c8e4a39f7680e75c21730cf5e1cfa). The addition of the `experimentalFeatures` property to the Codable ToolsVersion struct makes it a mandatory field when parsing. I think that the problem occurs because ToolsVersion is [used here](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/PackageFingerprint/FilePackageFingerprintStorage.swift#L201).

### Expected behavior

I should be able to place any Swift package in my Swift Package Repository and use it from my Swift packages without errors.

### Actual behavior

When I attempt to use a package from my Swift Package Repository that has a version specific Package.swift file, then I get an error like:
```
Computing version for weichsel.zipfoundation
[...]

PR #9828, opened 2026-03-19
Fix ToolsVersion Codable backward compatibility 

shrudge:
Fix `ToolsVersion` Codable backward compatibility by safely decoding `experimentalFeatures`.
**issue** :  #9823

### Motivation:

Older cached registry fingerprints (such as those from a Swift 5.8 toolchain) currently fail to decode in recent snapshots, throwing a `keyNotFound` error for `experimentalFeatures`. it occurs because the recent addition of the `experimentalFeatures` property caused the auto-synthesized `Decodable` conformance to expect the key to always be present, breaking backward compatibility with existing cache files. 

### Modifications:

* Added a custom `init(from decoder: Decoder)` to the `ToolsVersion` struct.
* Implemented `decodeIfPresent` for the `experimentalFeatures` property, defaulting it to an empty set (`[]`) if the key is missing from the JSON payload.
* Added unit tests to `ToolsVersionTests` to verify that JSON missing the `experimentalFeatures` key decodes successfully and that modern JSON containing the key round-trips correctly.

### Result:

Swift Package Manager successfully decodes older cached version fingerprints that lack the `experimentalFeatures` key. This restores backward compatibility for existing user caches without affecting how new fingerprints are encoded going forward :)

### DIFF
--- a/Sources/PackageModel/ToolsVersion.swift
+++ b/Sources/PackageModel/ToolsVersion.swift
@@ -90,15 +90,15 @@ public struct ToolsVersion: Equatable, Hashable, Codable, Sendable {
     public enum ExperimentalFeature: String, Sendable, Codable {
         case experimentalCGen
     }
-    public let experimentalFeatures: Set<ExperimentalFeature>
+    public let experimentalFeatures: Set<ExperimentalFeature>?
 
     /// Helpers for experimental
     public var experimentalCGen: Bool {
-        self >= .v6_3 && experimentalFeatures.contains(.experimentalCGen)
+        self >= .v6_3 && experimentalFeatures?.contains(.experimentalCGen) == true
     }
 
     /// Create an instance of tools version from a given string.
-    public init?(string: String, experimentalFeatures: Set<ExperimentalFeature> = []) {
+    public init?(string: String, experimentalFeatures: Set<ExperimentalFeature>) {
         guard let match = ToolsVersion.toolsVersionRegex.firstMatch(
             in: string, options: [], range: NSRange(location: 0, length: string.count)) else {
             return nil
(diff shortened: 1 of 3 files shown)


owenv:
cc @dschaefer2

dschaefer2:
I was just going to make the field optional...

shrudge:
@dschaefer2 that makes total sense and is much cleaner I'll just push an update to make the field optional and remove the custom decoder.

shrudge:
@dschaefer2 I removed the custom decoder and made `experimentalFeatures` optional. also updated the tests, and everything is passing locally. Let me know if it looks good to go

dschaefer2:
I also want to see why there are no registry tests for this and check what else is missing in that area.

shrudge:
@dschaefer2 That makes total sense, It definitely seems like a blind spot. I can draft a test that mocks a registry release metadata payload intentionally omitting the `experimentalFeatures` key. We can then verify that `RegistryClient` decodes the ToolsVersion fingerprint successfully without throwing. lmk if you want me to add that in this PR.

dschaefer2:
No, let me take a look at that whole area first.

dschaefer2:
@swift-ci please test

dschaefer2:
People are running into this now. We'll worry about registry tests in a subsequent update.

dschaefer2:
@swift-ci please test Windows

shrudge:
@dschaefer2 
Could you please re-run the macOS smoke test? 
the Jenkins timed out :/

shrudge:
@dschaefer2 looks like the recent CI run failed during dependency resolution, and I haven't made any changes to the package dependencies in this PR, it seems like there might be a temporary version mismatch between some of the upstream core packages right now.

I guess we're waiting on an upstream sync for this. Just lemme know if I need to rebase on main once the toolchain pins are updated.

dschaefer2:
Those failures are non-blocking but I want to take a quick look and see why they're happening.

dschaefer2:
The actions aren't merging in the latest main. Builds fine for me locally when I do that. I'm going to push this forward anyway.

Thanks for the help!

PR #9829, opened 2026-03-19
Install Swift Module for applications

rbenegal:
For cases such as Swift Playground applications, the Swift module is required to enable features such as SwiftUI previews.

### Motivation:

Ensure that previews work correctly for SwiftUI applications.

### Modifications:

- Do not set the `SWIFT_INSTALL_MODULE` setting to `NO` for product types of applications
- Use the `product` name instead of `mainModule` name for applications

### Result:

The .swiftmodule folder is now present for applications.

### Testing:
Confirmed that Previews now work correctly for Swift Playground applications instead of giving errors.

### DIFF
--- a/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
@@ -117,9 +117,16 @@ extension PackagePIFProjectBuilder {
         settings[.TARGET_TEMP_DIR_SUFFIX] = "-p"
         settings[.PACKAGE_RESOURCE_TARGET_KIND] = "regular"
         settings[.PRODUCT_NAME] = "$(TARGET_NAME)"
-        // We must use the main module name here instead of the product name, because they're not guranteed to be the same, and the users may have authored e.g. tests which rely on an executable's module name.
-        settings[.PRODUCT_MODULE_NAME] = mainModule.c99name
-        if product.type == .executable {
+        if pifProductType == .application {
+            // Use product name for applications to enable features such as SwiftUI Previews.
+            settings[.PRODUCT_MODULE_NAME] = product.c99name
+        } else {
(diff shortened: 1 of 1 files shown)


Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
```diff
@@ -117,9 +117,16 @@ extension PackagePIFProjectBuilder {
         settings[.TARGET_TEMP_DIR_SUFFIX] = "-p"
         settings[.PACKAGE_RESOURCE_TARGET_KIND] = "regular"
         settings[.PRODUCT_NAME] = "$(TARGET_NAME)"
-        // We must use the main module name here instead of the product name, because they're not guranteed to be the same, and the users may have authored e.g. tests which rely on an executable's module name.
-        settings[.PRODUCT_MODULE_NAME] = mainModule.c99name
-        if product.type == .executable {
+        if pifProductType == .application {
+            // Use product name for applications to enable features such as SwiftUI Previews.
```
bkhouri, line 121: Can we add an automated tests that verifies the function in isolation. That is that the PRODUCT_MODULE_NAME is set correctly in the PIC?

rbenegal: Thanks @bkhouri ! I'm not sure if this is the right fix, will let @plemarquand take a look at this at well, possibly can be rolled up into https://github.com/swiftlang/swift-package-manager/pull/9771 also, since both will be required to fix this I think. Will definitely add a test here (will probably have to rely ok mocks and such) if we just want to submit this separately.

bkhouri: No need to rely on mocks.  have a look at https://github.com/swiftlang/swift-package-manager/blob/main/Tests/SwiftBuildSupportTests/PIFBuilderTests.swift for example on how to write a test that will verifies this behaviour.

Issue #9821, opened 2026-03-17
Condition platform ".custom("none")" inside .unsafeFlags not works


purpln:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

I use `#if os(none)` to detect WASM platform inside Swift.
Fully working Swift Embedded WASM example:
```swift
#if os(none)
@_extern(c, "__main_argc_argv")
private func main(_ argc: Int32, _ argv: UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>) -> Int32

@_cdecl("_start")
func entry() {
    let pointer = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 1)
    pointer.initialize(to: nil)
    _ = main(0, pointer)
}
#endif
```
So I tried to apply specific arguments related to the WASM platform.
```swift
linkerSettings: [.unsafeFlags([...], .when(platforms: [.custom("none")]))]
```
My guess is that the 'none' platform is missing or not handled properly.
This problem might also be true for the Embedded only platform.

### Expected behavior

cSettings, cxxSettings, swiftSettings or linkerSettings with platform 'none' to be applied for Embedded WASM platform as it works inside Swift

### Actual behavior

SwiftPM ignores unsafeFlags with '.custom("none")' platform

### Steps to reproduce

```swift
let package = Package(name: "Application", products: [
    .executable(name: "Application", targets: ["Application"]),
], targets: [
    .executableTarget(name: "Application", linkerSettings: [
        .unsafeFlags(["-Xclang-linker", "-nostdlib"], .when(platforms: [.custom("none")])),
    ]),
])
```
swift build --triple wasm32-unknown-none-wasm -Xswiftc -enable-experimental-feature -Xswiftc Embedded

### Swift & OS version (output of `swift --version ; uname -a`)

Apple Swift version 6.2.3 (swift-6.2.3-RELEASE)

shrudge:
Hii @purpln, thanks for the great reproduction steps about the 'none' platform not being handled properly. Just opened a PR here: #9830
The issue was that BuildParameters.currentPlatform was falling through to .linux by default (kinda biased). I updated the mapping so that noneOS triples correctly map to .custom("none").

PR #9830, opened 2026-03-19
Fix BuildEnvironment platform mapping for triples with noneOS

shrudge:
Fix Condition platform ".custom("none")" inside .unsafeFlags.  so custom platform conditions evaluate correctly.
issue : #9821

### Motivation:

Currently, when building for Embedded Swift with a triple like `wasm32-unknown-none-wasm` (where the OS is `none`), passing specific linker settings conditionally using `.when(platforms: [.custom("none")])` fails. This happens because the `BuildParameters.currentPlatform` mapping defaults to `.linux` for any unrecognized OS, causing the platform condition to evaluate to false and the flags to be ignored. This resolves Issue #9821.

### Modifications:

* Added an `isNoneOS()` helper to `Triple` in `Triple+Basics.swift`.
* Updated `currentPlatform` in `BuildParameters.swift` to explicitly check for `isNoneOS()`. If true, it returns `.custom(name: self.triple.osNameUnversioned, oldestSupportedVersion: .unknown)`.
* Added `PlatformConditionTests.swift` to verify that a `BuildEnvironment` created from a `noneOS` triple correctly satisfies `.custom("none")` conditions and does not satisfy `.linux` conditions.

### Result:

Swift Package Manager correctly maps `noneOS` triples to a custom "none" platform, allowing developers targeting Embedded Swift and bare-metal environments to successfully apply conditional `.unsafeFlags` and linker settings in their `Package.swift` manifests.

### DIFF
--- a/Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
+++ b/Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
@@ -123,8 +123,10 @@ public struct BuildParameters: Encodable {
             return .freebsd
         } else if self.triple.isNoneOS() {
             return .custom(name: self.triple.osNameUnversioned, oldestSupportedVersion: .unknown)
-        } else {
+        } else if self.triple.isLinux() {
             return .linux
+        } else {
+            return .custom(name: "unknown", oldestSupportedVersion: .unknown)
         }
     }
 


Sources/SPMBuildCore/BuildParameters/BuildParameters.swift
```diff
@@ -121,6 +121,8 @@ public struct BuildParameters: Encodable {
             return .openbsd
         } else if self.triple.isFreeBSD() {
             return .freebsd
+        } else if self.triple.isNoneOS() {
+            return .custom(name: self.triple.osNameUnversioned, oldestSupportedVersion: .unknown)
         } else {
             return .linux
```
jakepetroules, line 127: Ugh, this should not default to Linux. Would you mind adding an explicit if-check for Linux, and make the else case `.custom(name: "unknown", oldestSupportedVersion: .unknown)`?

shrudge:
@jakepetroules absolutely, I've pushed the changes...

jakepetroules:
@swift-ci test

shrudge:
@jakepetroules seems like some infra issues, Could you please re-trigger the CI jobs when you get a chance?

jakepetroules:
@swift-ci test macOS

PR #9831, opened 2026-03-19
Fixes windows linking against a .dynamic product

daveinglis:
When linking against a dll on windows, the source module(s) for the module(s) that that are in the dll (product) was still being built with -static, this resulted in undefined symbols at link time, this change mimics native build behavior where the products immediate deps (but not there transitive deps) are what is built without -static as these module(s) are considered the exported interface for the dll.

close: https://github.com/swiftlang/swift-package-manager/issues/9421

### DIFF
--- a/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
@@ -512,7 +512,7 @@ extension PackagePIFProjectBuilder {
 
         // Apply target-specific build settings defined in the manifest.
         let allBuildSettings = mainModule.computeAllBuildSettings(observabilityScope: pifBuilder.observabilityScope, forRemotePackage: pifBuilder.delegate.isRemote)
-        
+
         // Apply settings using the convenience methods
         allBuildSettings.apply(to: &debugSettings, for: .debug)
         allBuildSettings.apply(to: &releaseSettings, for: .release)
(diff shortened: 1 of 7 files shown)


daveinglis:
This fixes building packages like https://github.com/brightdigit/PackageDSL.git

daveinglis:
@swift-ci test

Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Products.swift
```diff
@@ -592,7 +592,14 @@ extension PackagePIFProjectBuilder {
 
         // Also create a dynamic product for use by development-time features such as Previews and Swift Playgrounds.
         // If all targets this product is comprised of are binaries, we should *not* create a dynamic variant.
-        if libraryType == .automatic && libraryProduct.hasSourceTargets {
+        // Check if the product is macOS only by verifying all modules target only macOS.
```
owenv: I'm not sure this condition is correct, it should maybe be an option in PIFBuilderParameters or based on the host OS. Is it required alongside the other changes in this PR?

daveinglis: ya, this is actually required, without this we end up building **a lot** of swift modules without -static, which results in too many exported symbols when linking

owenv: Discussed offline that we probably want this to be something clients of the PIFBuilder can opt into, and the SwiftPM CLI opts out of, rather than inferring it based on platform

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test

daveinglis:
@swift-ci test windows

daveinglis:
@swift-ci test macOS

Issue #9800, opened 2026-03-10
`Context.gitInformation?.currentCommit` evaluates to `nil` if `.git` repository is added after `swift package resolve`


ahoppen:
Consider the attached package that produces an executable which just prints the current commit hash. When the `.git` directory is only added after running `swift package resolve`, the package manifest does not seem to get re-evaluates, resulting in a commit hash of `1` to be output.

To reproduce it, run `docker build --progress plain  .` and note the version that is printed by the last step. When copying `.git` before `swift package resolve`, the correct commit hash is printed.

[Test.zip](https://github.com/user-attachments/files/25871258/Test.zip)

shrudge:
@ahoppen, thanks for the excellent reproduction steps and the zip folder! Having the exact Dockerfile setup made tracking this down super eazzy.

I was able to reproduce the stale cache locally and just opened a PR for the fix here: #9832

The root cause was actually in how the cache checksum is generated. `ManifestLoader.CacheKey` was hashing the manifest contents, tools version, environment, etc., but it completely ignored whether the `.git `folder actually existed. Because of that, adding the folder later didn't change the cache key, so swiftPM never knew it needed to ig re-evaluate.

I updated the CacheKey SHA256 computation to include a simple boolean check for the .git directory's existence. Now, initializing or copying over a .git folder changes the hash and busts the cache perfectly.
im attaching a screenshot of the local test run successfully picking up the new commit hash `(1e1117c...)`after the fix! Let me know if the PR needs any more tweaks :) 

<img width="551" height="429" alt="Image" src="https://github.com/user-attachments/assets/838b28b9-1ca1-44ab-9986-535b6634e524" />

bkhouri:
Fixed by #9832

PR #9832, opened 2026-03-20
Fix manifest cache invalidation when accessing gitInformation

shrudge:
Fix manifest cache invalidation when accessing `Context.gitInformation`.
Issue : #9800

### Motivation:

Currently, `Context.gitInformation?.currentCommit` evaluates to `nil` if the `.git` directory is added *after* an initial `swift package resolve`. This occurs because the manifest `CacheKey` checksum does not include the state of the `.git` directory. If a manifest is evaluated before the `.git` directory exists, that evaluation is cached. Subsequent builds use this stale cache even after the `.git` directory is created, resulting in missing or incorrect git information. This resolves Issue #9800.

### Modifications:

* Updated `ManifestLoader.CacheKey` to check for the existence of the package's `.git` directory (`gitDirectoryExists`).
* Included the `gitDirectoryExists` boolean in the `computeSHA256Checksum` function so that the cache key directly depends on the directory's presence.
* Added `testCacheInvalidationOnGitDirectoryCreation` to `ManifestLoaderCacheTests.swift` to verify that creating a git repository after an initial cached load correctly invalidates the cache and forces a re-evaluation.

### Result:

The manifest cache will now correctly invalidate when a `.git` folder is initialized or copied into the directory after the initial dependency resolution, allowing `Context.gitInformation` to fetch the correct commit hash. 

Image : 
Image

### DIFF
--- a/Sources/PackageLoading/ManifestLoader.swift
+++ b/Sources/PackageLoading/ManifestLoader.swift
@@ -1032,6 +1032,8 @@ extension ManifestLoader {
               fileSystem: FileSystem
         ) throws {
             let manifestContents = try fileSystem.readFileContents(manifestPath).contents
+            let gitDirectoryPath = manifestPath.parentDirectory.appending(".git")
+            let gitDirectoryExists = fileSystem.exists(gitDirectoryPath)
             let sha256Checksum = try Self.computeSHA256Checksum(
                 packageIdentity: packageIdentity,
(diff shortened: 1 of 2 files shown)


bripeticca:
@swift-ci test

rconnell9:
@swift-ci test windows

bripeticca (approved):

PR #9835, opened 2026-03-22
[Swift Build] Only build for the active arch by default in release builds

owenv:
Previously, we built universal binaries when targeting Apple platforms. This is arguably better behavior, but in practice it causes compatibility issues in toolchain bootstrapping. Unconditionally set ONLY_ACTIVE_ARCH = YES for now, but we should reconsider this in the future.


owenv:
@swift-ci test

Issue #9833, opened 2026-03-20
Error Reporting: Improve messaging when Git executable isn't found


etcwilde:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

The error messaging when SwiftPM fails to fetch a dependency is fairly vague, which can make it especially hard to figure out what is going on in a CI environment.

> ```
>  Fetching https://github.com/swiftlang/swift-syntax.git
>  warning: 'swift-syntax': skipping cache due to an error: Failed to clone repository https://github.com/swiftlang/swift-syntax.git:
>  error: Failed to clone repository https://github.com/swiftlang/swift-syntax.git:
> ```

In this instance, I figured out that it was because SwiftPM wasn't finding the git executable, but this was after having to verify that we had network access and write access to the working directory.

### Expected behavior

A bit more information about why the dependency failed to clone. Some categories to start with might include network issues, missing write permissions, the disk is full, or that SwiftPM simply couldn't find the `git` executable.

### Actual behavior

The same "Failed to clone repository" error when there are network issues, file system issues, and platform configuration issues.

### Steps to reproduce

1. `docker run --rm -it -v $PWD:/mnt/host swiftlang/swift:nightly-debian12`
2. `mkdir -p /tmp/repro && cd /tmp/repro`
4. `git clone https://github.com/swiftlang/swift-testing.git`
5. `apt remove git`
6. `swift build`

```
Fetching https://github.com/swiftlang/swift-syntax.git
warning: 'swift-syntax': skipping cache due to an error: Failed to clone repository https://github.com/swiftlang/swift-syntax.git:
error: Failed to clone repository https://github.com/swiftlang/swift-syntax.git:
```

### Swift Package Manager version/commit hash

Swift version 6.4-dev (LLVM a3655ee8d8c4d74, Swift d13cbbfd336f246)

### Swift & OS version (output of `swift --version ; uname -a`)

Ubuntu 24.04
Debian 12
FreeBSD 14.3

All on top of main

```
Swift version 6.4-dev (LLVM a3655ee8d8c4d74, Swift d13cbbfd336f246)
Target: aarch64-unknown-linux-gnu
Build config: +assertions
```

dschaefer2:
[...]

PR #9836, opened 2026-03-22
Improve error reporting when git executable fails or is missing

shrudge:
Resolves Issue #9833.

### Motivation:
When SwiftPM fails to fetch a dependency because the `git` executable is missing from the system, it previously output a vague error with a trailing colon: `error: Failed to clone repository [URL]:` (found by @etcwilde). This occurred because the underlying process spawn error was not being captured, resulting in an empty string being appended to the console output. 

### Modifications:
* Updated `GitShellError` and `GitCloneError` in `GitRepository.swift` to fall back to `result.errorOutput` (which captures missing executable errors) if standard output is empty. Added explicit handling for the `missingExecutableProgram` error to provide a clear, actionable description.
* Refactored the `FetchError` message formatting in `Workspace.swift` to remove the hardcoded trailing colon. It now only appends `:\n[error description]` if a valid underlying error description actually exists.
* Added `testMissingGitExecutableError` to `GitRepositoryTests` to verify the underlying error mapping.
* Added `testFetchMissingGitExecutableMessage` to `WorkspaceTests` to verify the exact console output formatting.

### Result:
When `git` is missing, or when other systemic clone failures occur, SwiftPM will now surface the actual underlying system error rather than swallowing it, making debugging significantly easier in minimal CI environments like Docker.

---
*Local test run showing successful cache-busting and explicit error handling:*

Untitled

### DIFF
--- a/Sources/SourceControl/GitRepository.swift
+++ b/Sources/SourceControl/GitRepository.swift
@@ -1202,11 +1202,8 @@ package struct GitShellError: Error, CustomStringConvertible {
     let result: AsyncProcessResult
 
     public var description: String {
-        let stdout = (try? self.result.utf8Output()) ?? ""
-        let stderr = (try? self.result.utf8stderrOutput()) ?? ""
-        let output = (stdout + stderr).spm_chomp()
         let command = self.result.arguments.joined(separator: " ")
-        return "Git command '\(command)' failed: \(output)"
+        return formatErrorDescription(result: self.result, message: "Git command '\(command)' failed")
     }
 }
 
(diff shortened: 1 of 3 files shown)


pepicrft (approved):

Sources/SourceControl/GitRepository.swift
```diff
@@ -1259,10 +1278,20 @@ public struct GitCloneError: Error, CustomStringConvertible, DiagnosticLocationP
     }
 
     public var description: String {
-        let stdout = (try? self.result.utf8Output()) ?? ""
-        let stderr = (try? self.result.utf8stderrOutput()) ?? ""
-        let output = (stdout + stderr).spm_chomp().spm_multilineIndent(count: 4)
-        return "\(self.message):\n\(output)"
+        var output = ""
+        switch (self.result.output, self.result.stderrOutput) {
```
pepicrft: There seems to be some duplication across the two getters. I'd consider encapsulating that in a function that you can reuse.

shrudge: on it!

shrudge: I've extracted the shared formatting logic into a fileprivate helper function to DRY things up. Lemme know if it looks good.

build.log
bkhouri, line 1: chore (blocking): can we remove this file?

bkhouri (changes requested):
I thought I had reviewed this, but noticed my comments were still pending.  Sorry for the delay.

Due to this delay, there are merge conflicts that will also need to be resolved.

PR #9837, opened 2026-03-23
Fix definition of --install-only.

3405691582:
### Motivation:

The flag was apparently not behaving as intended. 

### Modifications:

It is likely the flag variable when used was not initialized properly. Instead, use the dest attribute on the flag.

### Result:

Setting flag true behaves as intended.


3405691582:
@swift-ci please test.

3405691582:
Let me double-check this behavior, this might not be necessary. I'll convert back to ready once done.

Issue #9745, opened 2026-02-18
[Parity][macOS] SwiftBuild backend does not code sign on macOS.


bkhouri:
The `getTaskAllowEntitlement` tests is currently marked as `withKnownIssue` when executing with the `swiftbuild` build backend.

The Swift Build build system looks to currently have code signing support, but SwiftPM does not currently wire it through.

shrudge:
Hi! @bkhouri  Im testing a fix by passing `CODE_SIGN_IDENTITY = "-"` and `ENTITLEMENTS_DONT_REMOVE_GET_TASK_ALLOW = "YES"` to `.swiftbuild` in `SwiftBuildSystem.swift.` I'm verifying the unmuted tests locally now and will open a PR shortlyyy

shrudge:
<img width="1756" height="226" alt="Image" src="https://github.com/user-attachments/assets/bde3e01d-c527-41e6-9e63-b7bd7387dca4" />

opening a PR

svmkr-dev:
This issue is duplicate of #8378 and can be safely closed, since #9380 implements it.
Link to this issue placed at `getTaskAllowEntitlement` tests was removed in #9862.

Issue #9420, opened 2025-11-26
[Parity] [Impact 1.59%:1.94%] [Windows] error: FileSystemError(kind: TSCBasic.FileSystemError.Kind.noEntry


kcieplak:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

https://github.com/apple/swift-collections-benchmark.git

Windows

The following package does not build using the 'swifbuild' backend but does with 'native'

```
swift build `--buildsystem=swiftbuild` fails to build package
```

```
info: Discovering Swift tasks after 'Compiling test_entry_point.swift'
info: Wrapping Swift module CollectionsBenchmarkTests_test_runner
info: Discovering Swift tasks after 'Wrapping Swift module CollectionsBenchmarkTests_test_runner'
info: Compile CollectionsBenchmarkTests-test-runner (aarch64)
    cd S:\\
[...]

PR #9838, opened 2026-03-23
Wire up code signing for swiftbuild backend on macOS

shrudge:
Resolves Issue #9745.

### Motivation:
The `getTaskAllowEntitlement` test is currently failing (and marked as a known issue) when using the `.swiftbuild` backend. While the native build system correctly signs debug executables on macOS, `SwiftPM` was not passing the necessary code-signing overrides down to the `SwiftBuild` configuration.

### Modifications:
* Updated `constructDebuggingSettingsOverrides` in `SwiftBuildSystem.swift` to pass `CODE_SIGN_IDENTITY = "-"` and `ENTITLEMENTS_DONT_REMOVE_GET_TASK_ALLOW = "YES"` when `shouldEnableDebuggingEntitlement` is true. 
* Removed the `.issue` wrapper for `getTaskAllowEntitlement` in `BuildCommandTests.swift` since the backend now correctly applies the entitlement.

### Result:
The `.swiftbuild` backend now correctly performs ad-hoc code signing and preserves the `get-task-allow` entitlement for debug builds on macOS, matching the behavior of the native build system. The `getTaskAllowEntitlement` test now passes for both backends.

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -1121,7 +1121,6 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
         var settings: [String: String] = [:]
         if parameters.shouldEnableDebuggingEntitlement {
             settings["DEPLOYMENT_POSTPROCESSING"] = "NO"
-            settings["CODE_SIGN_IDENTITY"] = "-"
         }
         // Set DEBUG_INFORMATION_FORMAT based on debugInfoFormat
         switch parameters.debugInfoFormat {
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -1108,6 +1108,7 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
         var settings: [String: String] = [:]
         if parameters.shouldEnableDebuggingEntitlement {
             settings["DEPLOYMENT_POSTPROCESSING"] = "NO"
+            settings["CODE_SIGN_IDENTITY"] = "-"
         }
         // Set DEBUG_INFORMATION_FORMAT based on debugInfoFormat
         switch parameters.debugInfoFormat {
--- a/Tests/CommandsTests/BuildCommandTests.swift
+++ b/Tests/CommandsTests/BuildCommandTests.swift
@@ -1276,10 +1276,9 @@ struct BuildCommandTestCases {
     }
 
     @Test(
+        .issue("https://github.com/swiftlang/swift-package-manager/issues/9420", relationship: .defect),
         // Windows builds of ExecutableNew using swiftbuild can fail because of problem with handling long paths which
         // is root cause of linked issue
-        .issue("https://github.com/swiftlang/swift-package-manager/issues/9420", relationship: .defect),
-        .issue("https://github.com/swiftlang/swift-package-manager/issues/9745", relationship: .defect),
         .tags(
             .Feature.CommandLineArguments.DisableGetTaskAllowEntitlement,
             .Feature.CommandLineArguments.EnableGetTaskAllowEntitlement,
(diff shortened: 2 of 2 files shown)


Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -1108,6 +1108,10 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
         var settings: [String: String] = [:]
         if parameters.shouldEnableDebuggingEntitlement {
             settings["DEPLOYMENT_POSTPROCESSING"] = "NO"
+            settings["ENTITLEMENTS_DONT_REMOVE_GET_TASK_ALLOW"] = "YES"
```
owenv: Isn't this a no-op?

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1278,8 +1278,6 @@ struct BuildCommandTestCases {
     @Test(
         // Windows builds of ExecutableNew using swiftbuild can fail because of problem with handling long paths which
         // is root cause of linked issue
-        .issue("https://github.com/swiftlang/swift-package-manager/issues/9420", relationship: .defect),
```
owenv, line 1281: This test was already passing before the changes in this PR, this issue is specifically related to long-path handling on windows

owenv (changes requested):

Tests/CommandsTests/BuildCommandTests.swift
shrudge: I removed the `ENTITLEMENTS` no-op and restored the #9420 skip tag for Windows. Thanks for the review

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -1119,6 +1108,7 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
         var settings: [String: String] = [:]
         if parameters.shouldEnableDebuggingEntitlement {
             settings["DEPLOYMENT_POSTPROCESSING"] = "NO"
+            settings["CODE_SIGN_IDENTITY"] = "-"
```
svmkr-dev: issue: this is already handled

This is handled in `SwiftBuildSystemPlanningOperationDelegate`: https://github.com/swiftlang/swift-package-manager/blob/be73cfd93a5c17237709da2957d1201ab8074a2b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift#L162-L166
so it's not needed unless we want to change Delegate implementation.

There was unresolved discussion in #9380 on whether to use `CODE_SIGN_IDENTITY` override, or that `if` block in `SwiftBuildSystemPlanningOperationDelegate`. I went with what is currently implemented because I felt it was safer: when setting `CODE_SIGN_IDENTITY` override we can't check if identity we try to override is empty.

Tests/CommandsTests/BuildCommandTests.swift
```diff
@@ -1276,10 +1276,9 @@ struct BuildCommandTestCases {
     }
 
     @Test(
+        .issue("https://github.com/swiftlang/swift-package-manager/issues/9420", relationship: .defect),
```
svmkr-dev: nitpick: comments below refer to this issue so I feel like it should really stay as it was: first comments, then issue link.

svmkr-dev (changes requested):
#9745 was actually closed by #9380.
`.issue` link on test exists probably because I missed it being introduced on main branch and it just cleanly merged with no conflicts. This PR should only remove that issue link in my opinion.
While at it we could also close #8378, since #9380 was merged.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
shrudge: Ah great catch thanks man I js saw the .issue wrapper still lingering on main and completely missed #9380.
I even read your PR thread long before coming to this, my bad.

I will revert my changes to SwiftBuildSystem.swift and scope this PR down to just cleaning up the stale test wrapper and yes also the comment order too.

svmkr-dev (approved):

bkhouri:
@swift-ci test

svmkr-dev:
This PR can be safely closed without merging since #9862 was merged and it introduced the change this PR does.

PR #9839, opened 2026-03-23
[Swift Build] Pass -Xswiftc flags to swiftc when used as the linker driver

owenv:
This resolves an issue where swift-driver's build script passed `-Xswiftc -no-toolchain-stdlib-rpath` when building for installation in the toolchain, which had no effect because the flag was only passed to the compile task.


owenv:
@swift-ci test

owenv:
@swift-ci test self hosted macOS

owenv:
@swift-ci test

owenv:
merged in https://github.com/swiftlang/swift-package-manager/pull/9661

PR #9841, opened 2026-03-23
Add a single public symbol to the library template so it builds without warnings

owenv:
If a library target is composed of objects which have no symbols, it will emit a warning when linked as a static archive on macOS. Add a single placeholder public function to the library template so it builds without warnings out of the box.


owenv:
@swift-ci test

jakepetroules (approved):

owenv:
@swift-ci test self hosted

owenv:
merged in https://github.com/swiftlang/swift-package-manager/pull/9661

PR #9843, opened 2026-03-23
Fix extraction of Swift SDKs on Windows

dschaefer2:
First, make sure we're using the Windows tar to extract archives on Windows. It's based on libarchive so also handles zip files. Move hardcoded use of the ZipArchiver to the UniversalArchiver so the platform logic can kick in.

Second fix the args of the extract command to properly extract the Swift SDK tarballs.

Also SandboxTests was calling tar directly which breaks when running the tests in bash.

### DIFF
--- a/Tests/WorkspaceTests/WorkspaceTests.swift
+++ b/Tests/WorkspaceTests/WorkspaceTests.swift
@@ -8101,11 +8101,14 @@ final class WorkspaceTests: XCTestCase {
         // Checks an unsupported extension.
         do {
             let unknownPath = sandbox.appending("unknown")
-            try XCTAssertThrowsError(
+            XCTAssertThrowsError(
                 try binaryArtifactsManager.checksum(forBinaryArtifactAt: unknownPath),
                 "error expected"
             ) { error in
-                let stringError = try XCTUnwrap(error as? StringError)
+                guard let stringError = error as? StringError else {
+                    XCTFail("not a string error")
+                    return
+                }
                 // What the file types are is platform specific
                 XCTAssert(stringError.description.hasPrefix("unexpected file type"))
             }


dschaefer2:
@swift-ci please test

Sources/Basics/Archiver/TarArchiver.swift
```diff
@@ -48,6 +65,7 @@ public struct TarArchiver: Archiver {
         to destinationPath: AbsolutePath,
         completion: @escaping @Sendable (Result<Void, Error>) -> Void
     ) {
+        print("Running", tarCommand)
```
dschaefer2: I'll remove this in the next update. I want to get through this CI run first.

Sources/Basics/Archiver/TarArchiver.swift
```diff
@@ -37,7 +45,16 @@ public struct TarArchiver: Archiver {
         self.cancellator = cancellator ?? Cancellator(observabilityScope: .none)
 
         #if os(Windows)
-        self.tarCommand = "tar.exe"
+        var tarPath: PWSTR?
+        defer { CoTaskMemFree(tarPath) }
+        let hr = withUnsafePointer(to: FOLDERID_System) { id in
+            SHGetKnownFolderPath(id, DWORD(KF_FLAG_DEFAULT.rawValue), nil, &tarPath)
```
jakepetroules: Can you please refactor this into a free function? Also, you can parameterize the folder ID so it's generalizable. See https://github.com/swiftlang/swift-build/blob/main/Sources/SWBWindowsPlatform/KnownFolders.swift

dschaefer2: Sure. Should I just clone the swift-build one? This probably should be somewhere where tools can share, but for now...

jakepetroules: Yeah I'd just copy the swift-build one for now.

dschaefer2:
The real fix was actually to stop hardcoding the compression format when extracting with tar so that the Windows tar command would handle zip files properly.

dschaefer2:
Our future selves should merge the Zip and Tar archivers so that we standardize on libarchive where we can.

jakepetroules:
Where are you finding .zip based Swift SDKs, btw? All of our official ones are .tar.gz.

dschaefer2:
Ah, true. You are correct, I ran into this running the tests on Windows and making sure we're running the right tar. The registry archives and prebuilts are zips. We'll always have a mix.

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test windows

PR #9846, opened 2026-03-24
Remove testSDKTracking test which didn't actually run an incremental build

owenv:
This test only ran a single clean build, so it wasn't actually testing incrementally, and that's better to test at the build system layer instead. Also, /tmp / /private/tmp symlink resolution issues were causing this to fail in swift-build GitHub actions CI. Drop this test entirely since it isn't very useful


owenv:
@swift-ci test

owenv:
@swift-ci test Windows

owenv:
@swift-ci test windows

owenv:
@bkhouri setting SDKROOT in this test is effectively a no-op because the test is only running a single build, so nothing is actually observing the effect of changing the variable

Issue #9703, opened 2026-02-10
Malformed JSON via NSJSONSerializationErrorIndex


irgordon:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

Failed to parse target info (malformed(json: \"error: permissionDenied\", underlyingError: Error Domain=NSCocoaErrorDomain Code=3840 \"Invalid value around line 1, column 0.\" UserInfo={NSDebugDescription=Invalid value around line 1, column 0., NSJSONSerializationErrorIndex=0})).\nRaw compiler output: error: permissionDenied"

### Steps to reproduce

Antigravity Sandboxed Run

### Swift & OS version (output of `swift --version ; uname -a`)

swift-driver version: 1.127.15 Apple Swift version 6.2.4 (swiftlang-6.2.4.1.4 clang-1700.6.4.2)
Target: arm64-apple-macosx26.0
Darwin mac.local 25.2.0 Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:56 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6041 arm64

shrudge:
<img width="1024" height="259" alt="Image" src="https://github.com/user-attachments/assets/142025bc-8d8a-45d7-a217-2f2ee66d15d5" />

The catch block currently tries to parse the plain-text compiler error as JSON and leaks the NSJSONSerialization error when it fails.

thanks @irgordon for reporting this will put up a PR to intercept this in the catch block and surface the raw compiler output cleanly instead...

heckj:
(my bad, sorry - was starting to comment about a similar error - hit the wrong buttons)

PR #9847, opened 2026-03-25
Fix plain-text compiler error handling for -print-target-info (#9703)

shrudge:
Fix plain-text compiler error handling for -print-target-info (#9703)

Motivation:

When swiftc fails and outputs plain text instead of JSON, SwiftPM tries to parse it anyway. This masks the real compiler error and throws a confusing InternalError (usually a JSON decoding or NSCocoaErrorDomain crash).

Modifications:

Error Handling (Triple+Basics.swift, UserToolchain.swift): Added a check to see if the trimmed compiler output starts with { or [. If not, it bails out of JSON parsing and throws a clean StringError containing the raw text.

Tests (TripleTests.swift): Added testTargetInfoPlainTextError using a mocked fake-swiftc script to verify the plain-text bypass works.

Result:

Plain-text compiler errors are now printed directly to the terminal instead of causing a JSON decoding crash.

Manual Verification:
Created a temporary DummyPackage with a failing mocked compiler to verify the clean terminal output:

image

### DIFF
--- a/Sources/Basics/Triple+Basics.swift
+++ b/Sources/Basics/Triple+Basics.swift
@@ -9,7 +9,7 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 //===----------------------------------------------------------------------===//
-
+import Foundation
 import TSCUtility
 import enum TSCBasic.JSON
 
@@ -112,6 +112,12 @@ extension Triple {
         do {
             parsedTargetInfo = try JSON(string: compilerOutput)
         } catch {
+            let trimmedOutput = compilerOutput.trimmingCharacters(in: .whitespacesAndNewlines)
+            if !trimmedOutput.hasPrefix("{") && !trimmedOutput.hasPrefix("[") {
+                // It's plain text, throw cleanly without the JSON error
+                throw StringError("Failed to parse target info: \(trimmedOutput)")
+            }
+
             throw InternalError(
                 "Failed to parse target info (\(error.interpolationDescription)).\nRaw compiler output: \(compilerOutput)"
             )
--- a/Sources/PackageModel/UserToolchain.swift
+++ b/Sources/PackageModel/UserToolchain.swift
@@ -217,8 +217,14 @@ public final class UserToolchain: Toolchain {
         do {
             return try JSON(string: compilerOutput)
         } catch {
+            let trimmedOutput = compilerOutput.trimmingCharacters(in: .whitespacesAndNewlines)
+            if !trimmedOutput.hasPrefix("{") && !trimmedOutput.hasPrefix("[") {
+                // It's plain text, throw cleanly without the JSON error
+                throw StringError("Failed to parse target info: \(trimmedOutput)")
+            }
+
             throw InternalError(
-                "Failed to parse target info (\(error.interpolationDescription)).\nCompiler exited with staus \(result.exitStatus).\nRaw compiler stdout: \(compilerOutput)\nRaw compiler stderr: \(compilerStderr)"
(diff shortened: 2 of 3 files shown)


finagolfin:
Thanks, this seems worthwhile, as some of us have hit this bug before. @jakepetroules, I see you modified some of this code before and are in the list of reviewers, mind taking a look at this small pull?

Sources/SwiftBuildSupport/PackagePIFBuilder.swift
```diff
@@ -615,7 +615,7 @@ public final class PackagePIFBuilder {
         settings[.LINKER_DRIVER] = "auto"
 
         // Don't emit warnings when a target exports no symbols (e.g. the empty library template)
-        settings[.LIBTOOL_NO_WARNING_FOR_NO_SYMBOLS] = "YES"
+        settings[single: "LIBTOOL_NO_WARNING_FOR_NO_SYMBOLS"] = "YES"
```
jakepetroules: This seems like an unrelated change.

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -1117,6 +1117,7 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
 
     private static func constructDebuggingSettingsOverrides(from parameters: BuildParameters.Debugging) -> [String: String] {
         var settings: [String: String] = [:]
+        settings["CODE_SIGN_IDENTITY"] = "-"
```
jakepetroules: This seems like an unrelated change.

jakepetroules:
> Thanks, this seems worthwhile, as some of us have hit this bug before. @jakepetroules, I see you modified some of this code before and are in the list of reviewers, mind taking a look at this small pull?

I'm a global code owner so I'm _always_ in the list of reviewers :)

dschaefer2:
I'd be happy with this if we split out the build setting changes into a separate PR. As Jake mentions, those seem unrelated and they need to be analyzed with a different set of eyes.

shrudge:
@jakepetroules @dschaefer2 Reverted those, pushing them in another PR, thanks :)

Regarding those other two changes: I originally added them because I hit a missing enum error for `LIBTOOL_NO_WARNING_FOR_NO_SYMBOLS` and some friction with ad-hoc signing, and I thought the fixes might be helpful upstream. However, if forcing `CODE_SIGN_IDENTITY = "-"` or using the string subscript is too aggressive or breaks other workflows, I am completely fine just keeping them as local workarounds on my build.

Let me know if you still want me to open that separate PR for further discussion, or if I should just drop it entirely

finagolfin:
@swift-ci test

jakepetroules (approved):

finagolfin:
Hmm, the self-hosted Windows job alone is failing with a _bunch_ of strange errors, but why are we even [still building with such an old prebuilt nightly 6.2 compiler](https://ci-external.swift.org/job/pr-swiftpm-windows-self-hosted/4733/consoleText)?
```
cmd /s /c "swift --version & swift test --version & cd C:\source\swiftpm & python Utilities/build-using-self" 
[2026-04-15T11:18:11.716Z] Swift version 6.2-dev (LLVM 8ba594305dd7286, Swift e68723f3bab80a3)
[2026-04-15T11:18:11.716Z] Target: x86_64-unknown-windows-msvc
[2026-04-15T11:18:11.716Z] Build config: +assertions
[2026-04-15T11:18:11.716Z] Swift Package Manager - Swift 6.2.0
```

@bkhouri, do you know? We should get that Windows CI job updated.

finagolfin:
@swift-ci test windows

finagolfin:
@swift-ci test windows

finagolfin:
@swift-ci test windows

PR #9857, opened 2026-03-26
Default interop mode to "complete" for swift-tools-version >= 6.4

jerryjrchen:
Set environment variable SWIFT_TESTING_XCTEST_INTEROP_MODE=complete at runtime if the user has not already provided an override.

More details in the evolution proposal: https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0021-targeted-interoperability-swift-testing-and-xctest.md

### Motivation:

This opts in new projects (and existing projects that bump tools version) to the complete interop mode.

### Modifications:

* Update `constructTestEnvironment` to take in a desired interop mode

* Only set interop mode if the user hasn't already provided an override

* Tools version prior to 6.4 doesn't modify default interop mode

* Testing: e2e testing for the different scenarios

### Result:

Note we don't bother retrieving interop mode for getTestSuites since it's not relevant there.

Resolves rdar://173410320

### DIFF
--- a/Sources/Commands/Utilities/TestingSupport.swift
+++ b/Sources/Commands/Utilities/TestingSupport.swift
@@ -212,7 +212,8 @@ enum TestingSupport {
             ).productsBuildParameters,
             sanitizers: sanitizers,
             library: .swiftTesting,
-            testProductPaths: [path]
+            testProductPaths: [path],
+            interopMode: nil // Interop not required when listing tests
         )
 
         var args: [String]


Sources/Commands/Utilities/TestingSupport.swift
```diff
@@ -286,6 +308,16 @@ enum TestingSupport {
     }
 }
 
+extension ToolsVersion {
+    /// Default interop mode for this tools version.
+    ///
+    /// If provided, this value should be set for the
+    /// `SWIFT_TESTING_XCTEST_INTEROP_MODE` environment variable.
+    var defaultInteropMode: String? {
+        if self >= .v6_4 { "complete" } else { nil }
```
stmontgomery:
```suggestion
        self >= .v6_4 ? "complete" : nil
```

Sources/CoreCommands/SwiftCommandState.swift
```diff
@@ -823,6 +823,21 @@ public final class SwiftCommandState {
...
+    /// - Throws: If an error occurs when trying to resolve workspace details.
+    /// - Returns: The current tools version, nil if no manifests are found.
+    public func getToolsVersion() async throws -> ToolsVersion? {
+        let workspace = try self.getActiveWorkspace()
+        let root = try self.getWorkspaceRoot()
+        let rootManifests = try await workspace.loadRootManifests(
+            packages: root.packages,
+            observabilityScope: self.observabilityScope
+        )
+        return rootManifests.values.first.map { $0.toolsVersion }
```
dschaefer2: What if there is more than one root package?

jerryjrchen: Hmm, it seems to me that `constructTestEnvironment` gets all the test products without grouping by package. Ideally (?) we could go group those together by package and figure out the tools version for each so we run each package with the appropriate default interop mode. Seems like a lot of effort and code restructuring for like not that much reward to me though.

If the multiple root packages case is rare, I think a "good enough" fallback behaviour would just take the lowest toolsVersion out of all the packages. Does that sound reasonable?

jerryjrchen:
@swift-ci Please test

jerryjrchen:
@swift-ci Please test

jerryjrchen:
@swift-ci Please test Windows

jerryjrchen:
@swift-ci Please test

jerryjrchen:
@swift-ci Please test

jerryjrchen:
@swift-ci Please test Windows

PR #9859, opened 2026-03-27
[Package description] Sort traits when serializing the package description

DougGregor:
Various uses of traits in the package manifest are expressed via the Set type. When serializing the package manifest, this results in nondeterministic output because set iteration ordering is itself nondeterministic. Write the serialization logic for these types explicitly and sort these sets as part of that serialization, making the results deterministic.


DougGregor:
@swift-ci please test

DougGregor:
https://github.com/swiftlang/swift-package-manager/pull/9860 is a better approach to this

PR #9860, opened 2026-03-27
[Package description & model] Sort sets when serializing a manifest

DougGregor:
Various uses of traits in the package manifest are expressed via the Set type. When serializing the package manifest, this results in nondeterministic output because set iteration ordering is itself nondeterministic. Eliminate this nondeterminism in two places:
* `PackageDescription`: Use arrays within the serialization layer and have clients building the serialization structures that sort the nondeterministic inputs at that point. This makes the resulting manifest deterministic while not changing the package manifest API. It affects the JSON that comes back from executing a package manifest.
* `PackageModel`: There is no separate serialization layer. So, explicitly implement `encode(to:)` for those types that do have sets in them to sort when encoding. This affects the output of `dump-package`.

### DIFF
--- a/Sources/PackageModel/Manifest/Manifest.swift
+++ b/Sources/PackageModel/Manifest/Manifest.swift
@@ -664,7 +664,7 @@ extension Manifest: Encodable {
         try container.encode(self.dependencies, forKey: .dependencies)
         try container.encode(self.products, forKey: .products)
         try container.encode(self.targets, forKey: .targets)
-        try container.encode(self.traits, forKey: .traits)
+        try container.encode(self.traits.sorted { $0.name < $1.name }, forKey: .traits)
         try container.encode(self.platforms, forKey: .platforms)
         try container.encode(self.packageKind, forKey: .packageKind)
     }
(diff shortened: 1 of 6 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:

DougGregor:
@swift-ci please test Windows

dschaefer2:
Can we leave it as a Set other than the Serialization boundary? It really is a Set, we just want to make sure that it's deterministic across package resolution.

DougGregor:
> Can we leave it as a Set other than the Serialization boundary? It really is a Set, we just want to make sure that it's deterministic across package resolution.

The `Manifest` in `PackageModel` has the same nondeterminism issue due to its sets, which will affect the `dump-package` command. We can handle this some other way, for example by writing custom `encode(to:)` for everything in the package model that involves sets, but this seemed cleaner. Thoughts?

DougGregor:

DougGregor:
@swift-ci please test

DougGregor:

DougGregor:
@swift-ci please test Windows

DougGregor:

dschaefer2:

dschaefer2:
Sounds like the Foundation time zone issue test has been disabled this morning.

DougGregor:

DougGregor:

PR #9865, opened 2026-03-30
[6.3 cherry-pick] Fix ToolsVersion Codable backward compatibility

dschaefer2:
Cherry pick #9828 to release/6.3.

Fix `ToolsVersion` Codable backward compatibility by safely decoding `experimentalFeatures`.
**issue** :  #9823

### Motivation:

Older cached registry fingerprints (such as those from a Swift 5.8 toolchain) currently fail to decode in recent snapshots, throwing a `keyNotFound` error for `experimentalFeatures`. it occurs because the recent addition of the `experimentalFeatures` property caused the auto-synthesized `Decodable` conformance to expect the key to always be present, breaking backward compatibility with existing cache files.

### Modifications:

* Added a custom `init(from decoder: Decoder)` to the `ToolsVersion` struct.
* Implemented `decodeIfPresent` for the `experimentalFeatures` property, defaulting it to an empty set (`[]`) if the key is missing from the JSON payload.
* Added unit tests to `ToolsVersionTests` to verify that JSON missing the `experimentalFeatures` key decodes successfully and that modern JSON containing the key round-trips correctly.

### Result:

Swift Package Manager successfully decodes older cached version fingerprints that lack the `experimentalFeatures` key. This restores backward compatibility for existing user caches without affecting how new fingerprints are encoded going forward :)

### DIFF
--- a/Sources/PackageModel/ToolsVersion.swift
+++ b/Sources/PackageModel/ToolsVersion.swift
@@ -90,15 +90,15 @@ public struct ToolsVersion: Equatable, Hashable, Codable, Sendable {
     public enum ExperimentalFeature: String, Sendable, Codable {
         case experimentalCGen
     }
-    public let experimentalFeatures: Set<ExperimentalFeature>
+    public let experimentalFeatures: Set<ExperimentalFeature>?
 
     /// Helpers for experimental
     public var experimentalCGen: Bool {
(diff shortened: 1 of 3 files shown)


dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:
@swift-ci please test windows

PR #9868, opened 2026-04-01
[Experiment] Optionally use a package for the Swift standard library

DougGregor:
### Motivation:
The Swift standard library is generally pre-built in the toolchain or in a separate SDK for one specific target triple, built from the CMake sources in the Swift compiler repository. Building for a different target triple generally requires creating or using a separate SDK.

### Modifications:
There is an experimental package manifest that builds the Swift standard library and related modules (like `_Builtin_float` and `Synchronization`) from sources. Introduce a new command-line flag to SwiftPM, `--experimental-standard-library-package`, that uses that experimental package manifest to build the standard library. It does so by injecting a few things into the build graph:

* A dependency on the "swift" package that contains the Swift standard library, which is currently pointed at my own Swift fork and its branch containing the experimental package.
* For each non-host target, a dependency on the "Swift" product from that package, which is the standard library.
* For each non-host target, the Swift compiler flags `-nostdimport` and `-nostdlibimport`, which prevent the compiler from looking into the default toolchain or SDK directories for the standard library and related modules.

### Result:

When using this flag, a "normal" package that builds for Embedded Swift can be built entirely from sources, independent of the prebuilt standard library modules, for any target triple.

### DIFF
--- a/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
+++ b/Sources/SwiftBuildSupport/PackagePIFProjectBuilder+Modules.swift
@@ -453,6 +453,15 @@ extension PackagePIFProjectBuilder {
             }
         }
 
+        /// If we're building the Swift standard library module from a package,
+        /// build everything that depends on it without reference to the
+        /// pre-built standard library in the toolchain or SDK.
+        if sourceModule.packageName == "swift" && sourceModule.name == "Swift" {
+            impartedSettings[.OTHER_SWIFT_FLAGS, default: ["$(inherited)"]]
+                .append(contentsOf: ["-nostdimport", "-nostdlibimport"])
+        }
+
+
         if desiredModuleType == .dynamicLibrary {
             settings.configureDynamicSettings(
                 product: nil,


DougGregor:
This is totally absolutely experimental. It only works with the native build system, and has no tests beyond what I hacked together at desk. But it does work with a minimal package project where I run:

```
swift build --experimental-standard-library-package --toolset toolset.json -c release --build-system native --triple spir64-none-none-eabi
```

using Swift 6.4 (from `main`) and `toolset.json` that contains:

```
{
    "schemaVersion": "1.0",
    "swiftCompiler": {
        "extraCLIOptions": [
            "-enable-experimental-feature", "Embedded",
        ]
    }
}
```

DougGregor:
@swift-ci please test

DougGregor:
For reference, the Swift standard library package this is referencing is at https://github.com/swiftlang/swift/pull/86287

jakepetroules:
This is pretty cool. I see this is currently specific to Embedded Swift but could this scheme eventually be extended to allow a user to cross-compile a statically linked binary for a platform Swift has never supported before, like NetBSD for example? (assuming the stdlib itself didn't happen to need changes for said platform)

> It only works with the native build system

We've already switched the default build system to Swift Build in main. If you want to land this, I think it realistically needs to gain Swift Build support.

DougGregor:
> This is pretty cool. I see this is currently specific to Embedded Swift but could this scheme eventually be extended to allow a user to cross-compile a statically linked binary for a platform Swift has never supported before, like NetBSD for example? (assuming the stdlib itself didn't happen to need changes for said platform)

It can be extended for non-Embedded Swift. There are some SwiftPM limitations that make it a little harder (e.g., the inability to have conditionally excluded/included files, the inability to mix C++ and Swift in a target), but we might be able to power through them.

> We've already switched the default build system to Swift Build in main. If you want to land this, I think it realistically needs to gain Swift Build support.

Of course. I started this before the default switched over. It's likely a trivial fix in the PIF builder.

jakepetroules:
> > This is pretty cool. I see this is currently specific to Embedded Swift but could this scheme eventually be extended to allow a user to cross-compile a statically linked binary for a platform Swift has never supported before, like NetBSD for example? (assuming the stdlib itself didn't happen to need changes for said platform)
> 
> It can be extended for non-Embedded Swift. There are some SwiftPM limitations that make it a little harder (e.g., the inability to have conditionally excluded/included files, the inability to mix C++ and Swift in a target), but we might be able to power through them.

Nice. We've been talking about removing the mixed sources limitation since Swift Build enables us to do so. Conditionally included/excluded files seems like an obvious addition as well, but will need an evolution proposal of course.

dschaefer2:
Working on the mixed language when using SwiftBuild now.

Lots of evolution proposals on my list. As Doug G suggests, we just need to power through them.

PR #9870, opened 2026-04-01
feat: Source Archives and shallow clones

vsarunas:
Download source archives from GitHub instead of performing full git clones for semantic version ranged dependencies.

Sub-PRs:
1. https://github.com/swiftlang/swift-package-manager/pull/9909
2. https://github.com/swiftlang/swift-package-manager/pull/9910
3. https://github.com/swiftlang/swift-package-manager/pull/9911
4. https://github.com/swiftlang/swift-package-manager/pull/9912
5. https://github.com/swiftlang/swift-package-manager/pull/9942

### Motivation:

SPM currently clones the full git history (`git clone --mirror`) for every dependency.
For projects with many dependencies, this results in long resolution times and large `.build/` directories due to pulling full version history.

This PR takes a different approach that sidesteps git cloning entirely for the common case: downloading ZIP source archives directly from GitHub CDN.

This PR provides typically 2x faster resolves and 3x smaller disk footprint for GitHub based repos.

See the [forum post](https://forums.swift.org/t/swiftpm-2x-faster-resolves-3x-smaller-disk-footprint/85750) for extended motivation, benchmarks methodology, and edge case discussion.

### Modifications:

Opt-in via `--experimental-source-archive-downloads` or `SWIFTPM_USE_SOURCE_ARCHIVES=1`.

**Fetch strategy:**

| Dependency type | Fetch method |
|---|---|
| Semver version, supported host, no submodules | ZIP archive download |
| Semver version, supported host, has submodules | Shallow clone (`--depth 1 --recurse-submodules`) |
| Branch or revision pin | Standard git clone |
| SSH URL or unsupported host | Standard git clone |
| Any failure in archive/shallow path | Fallback to standard git clone |

**How it works:**

1. `git ls-remote --tags` - discover versions (same as today, zero API quota)
6. HTTP GET `Package.swift` from `raw.githubusercontent.com` - check tools-version compatibility
7. HEAD probes for manifest variants (`Package@swift-X.Y.swift`) - only 0.9% of packages use these
8. HTTP GET `.gitmodules` - determines ZIP vs shallow clone
9. Download ZIP archive or `git clone --depth 1 --recurse-submodules`
10. On any failure - fallback to `git clone --mirror`, per-dependency

All metadata cached permanently by commit SHA. Downloads go through a two-tier cache (shared global - workspace-local) and are prefetched in parallel.

The existing git path reads manifests by forking `git ls-tree` + `git cat-file` subprocesses per version against the local bare mirror - sequential process forks that add up during resolution. The source archive path replaces these with HTTP GETs to CDN endpoints, avoiding per-version process fork overhead and enabling concurrent manifest fetching.

**Extensibility:**

Host-specific behavior is behind the `SourceArchiveProvider` protocol - two methods (`archiveURL(for:)` and `rawFileURL(for:sha:)`) plus a `cacheKey`. Adding GitLab or Bitbucket support means implementing one type conforming to this protocol.

**Authentication:**

Private repos work if HTTPS auth is configured. The `GitHubTokenAuthorizationProvider` wraps SPM's existing `AuthorizationProvider` (netrc, keychain, credential helpers) and falls back to `GITHUB_TOKEN` / `GH_TOKEN` environment variables for GitHub hosts (`github.com`, `raw.githubusercontent.com`, `codeload.github.com`). SSH-only repos are detected early via a validation probe and fall back to git.

**Cache directory structure:**

```
.build/
  source-archives/{identity}/{version}/     <- extracted source tree
  shallow-clones/{identity}/                <- shallow clone with submodules

~/.cache/org.swift.swiftpm/
  source-archives/
    downloads/{identity}/{version}/         <- shared extracted source (not ZIP)
    metadata/{owner}/{repo}/{sha}/          <- manifest + submodule flag (immutable by SHA)
      metadata.json
      Package.swift
      Package@swift-6.0.swift              <- variant, if found
    shallow-clones/{identity}/{version}/    <- shared shallow clones
```

ZIP is deleted after extraction. Cache hits are directory copies. `swift package purge-cache` clears all new cache directories.

**Key design decisions:**

- **Tags not SHAs for archive URLs** - GitHub generates different ZIP bytes for `/zip/{sha}` vs `/zip/refs/tags/{tag}` even for the same commit. The file contents are identical, but the top-level directory name differs (e.g., `swift-log-1.6.3` vs `swift-log-3d8596ed08bd13520157f0355e35caed215ffbfa`), which changes the ZIP entry headers and thus the checksum. Using tags consistently avoids spurious TOFU checksum mismatches.
- **`codeload.github.com` directly** - avoids the 302 redirect from `github.com/.../archive/...` that crashes Foundation's `URLSession` on Linux (`_HTTPURLProtocol.configureEasyHandle` fails during `didCompleteRedirectCallback` in `libFoundationNetworking.so`).
- **Annotated tag peeling** - `git ls-remote` returns both the tag object SHA and the dereferenced commit SHA (`^{}`); parser prefers the peeled SHA.
- **Validation probe on container creation** - fetches tools version for 3 recent versions to detect HTTPS issues early (e.g., `insteadOf` rewrites, SSH-only auth).
- **`Package.resolved` unchanged** - revision SHAs from `git ls-remote` are identical to git checkout.
- **ZIP determinism** - GitHub does not guarantee byte-stable archives forever. In January 2023, a Git upgrade changed the compression library used by `git archive`, silently altering checksums for every repository - identical source, different bytes. It was [rolled back](https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/) after breaking Bazel, Homebrew, and Nix, and GitHub now [commits to](https://github.blog/open-source/git/update-on-the-future-stability-of-source-code-archives-and-hashes/) at least one year of stability with six months' notice before changes. If it happens again, TOFU checksums stored on one machine become invalid. The fix is `swift package purge-cache` to accept new checksums. A future central registry would eliminate this entirely by serving its own immutable archives.

**Test coverage:**

**68 tests** | **87.6% line coverage (1,280/1,461)**

| File | Coverage | Lines |
|------|----------|-------|
| `Basics/SourceArchiveMetadataCache.swift` | **100.0%** | 65/65 |
| `Basics/SourceArchiveProvider.swift` | **100.0%** | 3/3 |
| `Workspace/SourceArchiveDownloader.swift` | **99.4%** | 162/163 |
| `Basics/SourceArchiveResolver.swift` | **93.3%** | 222/238 |
| `Workspace/PackageContainer/SourceArchivePackageContainer.swift` | **93.3%** | 377/404 |
| `Basics/GitHubSourceArchiveProvider.swift` | **90.1%** | 73/81 |
| `Workspace/Workspace+SourceArchives.swift` | **74.6%** | 378/507 |

<details>
<summary>What's not covered</summary>

**`Workspace+SourceArchives.swift` (74.6%)** - largest gap:
- `removeSourceArchive` / `removeShallowClone` - cleanup paths
- Shallow clone with shared cache path (copy-from-cache and copy-to-cache branches)
- `prefetchSourceArchives` - the full prefetch loop including cancellation handling
- `canUseSourceArchive` guard checks
- `downloadSourceArchive` - the real download-and-extract path using `SourceArchiveDownloader` (vs the mock path tested in integration tests)

**`SourceArchiveResolver.swift` (93.3%)**:
- `defaultGitTagsProvider` - the real `git ls-remote` fallback (tests inject a mock provider)
- Some defensive guards in tag parsing (empty SHA, non-hex SHA, non-`refs/tags/` prefixes)
- `probeManifestVariant` when no candidates exist

**`SourceArchivePackageContainer.swift` (93.3%)**:
- `shouldInvalidatePinnedVersions`, `description` - trivial properties
- `isToolsVersionCompatible` - async compatibility check
- Tag preference logic when a new tag has more dot-components than an existing one
- `prefetchManifestContents` early-return when all SHAs are cached
- Git container fallback for revision/unversioned queries

**`GitHubSourceArchiveProvider.swift` (90.1%)**:
- `archiveURL` / `rawFileURL` - the real URL construction (tests use the protocol, not GitHub-specific URLs)
- `preconditionFailure` branches for malformed URLs
- Empty owner/repo guard in `parseGitHubURL`

</details>

### Result:

Zero changes to `Package.swift`, `Package.resolved`, or any public API. When enabled:

**Cold resolve** (all caches wiped), 10 runs across Mac ARM + Linux x86:

| Project | Deps | Archives p50 | Git p50 | Faster |
|---|---|---|---|---|
| spi-server | 67 | 68s | 100s | **25-40%** |
| swiftpm-large-project | 48 | 46s | 93s | **50-55%** |
| penny-bot | 47 | 44s | 78s | **40-47%** |
| container | 29 | 26s | 42s | **34-43%** |

`.build/` disk: **1.4-3.2x smaller** across all tested projects.


Sources/Basics/GitHubSourceArchiveProvider.swift
```diff
@@ -0,0 +1,134 @@
...
+        return set
+    }()
+
+    public func archiveURL(for tag: String) -> URL {
+        let escapedTag = tag.addingPercentEncoding(withAllowedCharacters: Self.urlPathComponentAllowed) ?? tag
+        // Use codeload.github.com directly to avoid a 302 redirect from
+        // github.com/archive/. Foundation's libcurl backend on Linux crashes
+        // following that redirect (CURLE_BAD_FUNCTION_ARGUMENT in
+        // configureEasyHandle during didCompleteRedirectCallback).
+        guard let url = URL(string: "https://codeload.github.com/\(owner)/\(repository)/zip/refs/tags/\(escapedTag)") else {
```
plemarquand: Using codeload.github.com vs the GitHub API has some important tradeoffs worth considering. While codeload requires no redirects, the URL format is also subject to change and doesn't seem to handle auth. Because of this, does it support private repos? What are the rate limit implications of using one over the other?

Regardless, in order to adopt the GitHub API we'd need to get to the bottom of https://github.com/swiftlang/swift-corelibs-foundation/issues/5445

Sources/Workspace/Workspace+ResolvedPackages.swift
```diff
@@ -177,10 +182,10 @@ extension ResolvedPackagesStore.ResolutionState {
         }
     }
 
-    func equals(_: Version) -> Bool {
+    func equals(_ other: Version) -> Bool {
         switch self {
         case .version(let version, _):
-            return version == version
+            return version == other
```
plemarquand, line 188: 😮‍💨

dschaefer2:
I was about to propose something similar looking at integrating C libraries from source. I'll take a deep study of your PR and offer suggestions to make sure it'll fit in with some of the things we've been thinking of. Thanks!

Sources/Workspace/PackageContainer/SourceArchivePackageContainer.swift
```diff
@@ -0,0 +1,499 @@
...
+                seen.insert(sha)
+                continue
+            }
+            seen.insert(sha)
+            shasToFetch.append(sha)
+        }
+
+        guard !shasToFetch.isEmpty else { return }
+
+        let swiftVersion = Version(
```
plemarquand: nit: Move this out to a private getter as its moved in a few places:
```
    private var swiftVersion: Version {                                                                                                                                                                                                            
        Version(                                                                                                                                                                                                                                   
            self.currentToolsVersion.major,                                                                                                                                                                                                        
            self.currentToolsVersion.minor,                                                                                                                                                                                                        
            self.currentToolsVersion.patch                                                                                                                                                                                                         
        )                                                                                                                                                                                                                                          
    } 
```

Sources/Basics/SourceArchiveMetadataCache.swift
```diff
@@ -0,0 +1,107 @@
...
+    private let cachePath: AbsolutePath
+
+    public init(fileSystem: any FileSystem, cachePath: AbsolutePath) {
+        self.fileSystem = fileSystem
+        self.cachePath = cachePath
+    }
+
+    public func getMetadata(owner: String, repo: String, sha: String) throws -> SourceArchiveMetadata? {
+        let metadataPath = self.metadataFilePath(owner: owner, repo: repo, sha: sha)
+        guard self.fileSystem.exists(metadataPath) else {
```
plemarquand: Better to attempt to read the file and catch the error when it fails instead of checking it exists first. Not only does checking first have slightly more overhead there is also the potential race where the file is deleted after the check, but before the read happens.

Sources/Basics/SourceArchiveMetadataCache.swift
```diff
@@ -0,0 +1,107 @@
...
+            encoder.outputFormatting = [.prettyPrinted, .sortedKeys]
+            let data = try encoder.encode(metadata)
+            try self.fileSystem.writeFileContents(metadataPath, data: data)
+        }
+    }
+
+    public func getManifest(owner: String, repo: String, sha: String, filename: String) throws -> String? {
+        let manifestPath = self.shaDirectoryPath(owner: owner, repo: repo, sha: sha)
+            .appending(component: filename)
+        guard self.fileSystem.exists(manifestPath) else {
```
plemarquand: same suggestion as above: read, catch, return nil if its a file not found error.

dschaefer2:
Part of what I wanted to accomplish in my work is the ability to hook up source archives from any URL (with an accompanying checksum for security). That will require Package Manifest additions which is fine.

I like this addition for repository based packages that we know are Github and have a fixed version where we can provide automatic acceleration to fetch the source archive and fallback to git when that doesn't work. I am wary about relying on GitHub interfaces that may change or disappear in the future, but with a fallback mechanism we can make sure we don't break if that happens.

I'll be thinking about the architecture of this solution and make sure it fits in cleanly with the more general vision and make suggestions as I work through that. But this is a fantastic start. Thank you!

vsarunas:
I wouldn't be worried about GitHub API changing, there is only 2 endpoints used:

Opening up https://github.com/apple/swift-nio/blob/2.97.1/Package.swift -> press Raw which requests to https://github.com/apple/swift-nio/raw/refs/tags/2.97.1/Package.swift -> which redirects to CDN domain https://raw.githubusercontent.com/apple/swift-nio/refs/tags/2.97.1/Package.swift

Varnish CDN:
```
curl -sI -L https://raw.githubusercontent.com/apple/swift-nio/558f24a4647193b5a0e2104031b71c55d31ff83a/Package.swift
HTTP/2 200
cache-control: max-age=300
content-security-policy: default-src 'none'; style-src 'unsafe-inline'; sandbox
content-type: text/plain; charset=utf-8
etag: "81a42c6ca70d6080049d98651658b83565e498db36af30466f0fd1fc3f7c79c9"
strict-transport-security: max-age=31536000
x-content-type-options: nosniff
x-frame-options: deny
x-xss-protection: 1; mode=block
x-github-request-id: D334:49769:1DB68:40274:69CD439E
accept-ranges: bytes
date: Wed, 01 Apr 2026 16:12:38 GMT
via: 1.1 varnish
x-served-by: cache-bma-essb1270025-BMA
x-cache: HIT
x-cache-hits: 0
x-timer: S1775059959.635458,VS0,VE2
vary: Authorization,Accept-Encoding
[...]
```

# Source Archive

https://github.com/apple/swift-nio/releases/tag/2.97.1 -> gives link to https://github.com/apple/swift-nio/archive/refs/tags/2.97.1.zip -> Redirects to https://codeload.github.com/apple/swift-nio/zip/refs/tags/2.97.1

Cannot follow redirects currently due to: https://github.com/swiftlang/swift-corelibs-foundation/issues/5445

https://docs.github.com/en/repositories/working-with-files/using-files/downloading-source-code-archives?apiVersion=2026-03-10

The tag move is affecting the git clone path currently as well; if tag is deleted and re-created will get `does not match previously recorded value` error.

Sources/Workspace/Workspace+SourceArchives.swift
```diff
@@ -0,0 +1,516 @@
...
+    func sourceArchiveAuthorizationProvider(
+        for provider: any SourceArchiveProvider
+    ) -> (any AuthorizationProvider)? {
+        provider is GitHubSourceArchiveProvider
+            ? GitHubSourceArchiveProvider.GitHubTokenAuthorizationProvider(
+                underlying: self.authorizationProvider)
+            : self.authorizationProvider
+    }
+
+    /// Creates a ``SourceArchiveDownloader`` using the workspace's shared HTTP client and file system.
```
plemarquand:
```suggestion
    /// Creates a `SourceArchiveDownloader` using the workspace's shared HTTP client and file system.
```

Sources/Workspace/Workspace+SourceArchives.swift
```diff
@@ -0,0 +1,516 @@
...
+        }
+        self.delegate?.didFetchPackage(
+            package: identity,
+            packageLocation: packageLocation,
+            result: .success(PackageFetchDetails(fromCache: false, updatedCache: paths.cachePath != nil)),
+            duration: fetchStart.distance(to: .now())
+        )
+    }
+
+    /// Downloads a source archive for a package (ZIP path, no submodules).
```
plemarquand, line 255: Packages with submodules fall back to a shallow clone; how do packages using git-lfs have their lfs files archived? Are they included in this downloaded archive or do we also need to fall back for repos using git-lfs?

Sources/Workspace/Workspace+SourceArchives.swift
```diff
@@ -0,0 +1,516 @@
...
+            )
+
+            guard self.fileSystem.exists(tempClonePath.appending("Package.swift")) else {
+                throw StringError("Package.swift not found in shallow clone at '\(tempClonePath)'")
+            }
+
+            try self.fileSystem.withLock(on: cachePath, type: .exclusive) {
+                if !self.fileSystem.exists(cachePath.appending("Package.swift")) {
+                    if !self.fileSystem.exists(cachePath.parentDirectory) {
+                        try self.fileSystem.createDirectory(cachePath.parentDirectory, recursive: true)
```
plemarquand: You can call createDirectory here without checking for existence first, as `recursive:true` acts like `mkdir -p` and only creates the directories if they don't exist.

Sources/Workspace/Workspace+SourceArchives.swift
```diff
@@ -0,0 +1,516 @@
...
+            try await gitProvider.shallowClone(
+                repository: repository, tag: tag, to: tempClonePath,
+                recurseSubmodules: true
+            )
+
+            guard self.fileSystem.exists(tempClonePath.appending("Package.swift")) else {
+                throw StringError("Package.swift not found in shallow clone at '\(tempClonePath)'")
+            }
+
+            try self.fileSystem.withLock(on: cachePath, type: .exclusive) {
```
plemarquand, line 383: Actually maybe the following is sufficient:

```swift
  try self.fileSystem.withLock(on: cachePath, type: .exclusive) {                                                                                                                                                                                   
      if !self.fileSystem.exists(cachePath.appending("Package.swift")) {                                                                                                                                                                            
          try self.fileSystem.createDirectory(cachePath.parentDirectory, recursive: true)                                                                                                                                                           
          try? self.fileSystem.removeFileTree(cachePath)                                                                                                                                                                                            
          try self.fileSystem.copy(from: tempClonePath, to: cachePath)                                                                                                                                                                              
      }                                                                                                                                                                                                                                             
  }   
```

plemarquand (commented):
This is great, and I'm optimistic about this approach to speeding up package resolution. I gave it a pass and had a few comments.

Sources/Basics/GitHubSourceArchiveProvider.swift
vsarunas: There are 3 endpoints:

1. https://api.github.com/repos/apple/swift-nio/contents/Package.swift?ref=2.76.1 , but this produces json output and has rate limits, and provides `"download_url": "https://raw.githubusercontent.com/apple/swift-nio/2.76.1/Package.swift",`
2. https://raw.githubusercontent.com/apple/swift-nio/2.76.1/Package.swift, which doesn't have rate-limits, and returns a redirect to the codelab. I was using the raw domain initially but had to switch due to the mentioned bug once started to test on Linux.

vsarunas:
>  https://github.com/apple/swift-nio/archive/refs/tags/2.97.1.zip

This is the endpoint that Brew [uses](https://github.com/search?q=repo%3AHomebrew%2Fbrew%20%22archive%2Frefs%2Ftags%22&type=code); looks like for formula building; not for downloading every binary.

npm package installer [first tries a tarball](https://github.com/npm/pacote/blob/v21.5.0/lib/git.js#L51-L58), and then resorts to git clone; all of the logic and translation is in [hosted-git-info](https://github.com/npm/hosted-git-info/blob/v9.0.2/lib/hosts.js#L55), which uses the codeload.github.com domain. With [11 million](https://www.npmjs.com/package/pacote) weekly downloads, would be fairly safe to use this host and avoid the redirect hop; but for downloads the extra 50-100ms doesn't really matter.

Sources/Workspace/Workspace+SourceArchives.swift
vsarunas: Need to figure this one out.

Example, hosts images in lfs: https://github.com/wireapp/wire-ios/blob/develop/.gitattributes

When downloading the release the zip file does contain:
```
wire-ios-appstore-4.16.0/wire-ios/Wire-iOS Tests/ReferenceImages/ConversationListCellTests/testThatItRendersGroupConversationWithIncomingCall.1.png
```

but unfortunately it just shows:
```
version https://git-lfs.github.com/spec/v1
oid sha256:4d78632c8f7439617a0ea6800754f0aa4f1a958cc8394c6c54b508a6d1ae2cc6
size 14538
```

And when opening https://github.com/wireapp/wire-ios/blob/develop/wire-ios/Wire-iOS%20Tests/ReferenceImages/ConversationListCellTests/testThatItRendersGroupConversationWithIncomingCall.1.png shows `Stored with Git LFS`

vsarunas:
@dschaefer2 , @plemarquand please review new changes in the PR:

 1. Tag-based to  SHA-based archive downloads - eliminates the race condition where a tag could move between `git ls-remote` and the archive download.
2. Symlinks from shared cache - workspace `.build/source-archives/` entries are symlinks to the shared cache (`~/Library/Caches/org.swift.swiftpm/source-archives/`), reducing per-project disk usage from hundreds of MB to effectively zero. Cached files are made read-only to prevent accidental modification through symlinks. Hundreds of MB saved across each project.
3. Parallel tag prefetch - *3x* `swift package update` improvement by pre-populating the tag memoizer before resolution starts and switching to a shared HTTP session.

## Speeding up tag discovery

The standard git path forks a `git fetch` process per dependency. Each process establishes its own TCP connection and TLS handshake - two network round-trips per repo, repeated ~50 times sequentially through the dependency graph.

The source archive path replaces these process forks with in-process HTTP requests using [git protocol v2](https://git-scm.com/docs/protocol-v2), reusing a single TCP/TLS session across all repos. Tags for all packages are fetched in parallel, before the resolver starts, so every `getContainer` → `getTags()` call during resolution is an instant memoizer hit.

Additionally, apply a filtering of what tags to receive via `ref-prefix` argument passed to GitHub, to avoid receiving hundreds of kilobytes of tags which will be a discarded, i.e remove tags that are not relevant went using source downloads:
```
009319121d5b302ad5c46e8db115fabe5f88e826638f refs/tags/swift-5.9-DEVELOPMENT-SNAPSHOT-2023-09-26-a peeled:d42e3db1c0f97f4a3be9a4552f7003c2959268d5
0093de29b46c1c2a4f92c894092c6bc5a99b14c153dd refs/tags/swift-5.9-DEVELOPMENT-SNAPSHOT-2023-09-27-a peeled:d42e3db1c0f97f4a3be9a4552f7003c2959268d5
0093dcc1b5ee99d037fa537b07e943732af47db220d7 refs/tags/swift-5.9-DEVELOPMENT-SNAPSHOT-2023-09-28-a peeled:d42e3db1c0f97f4a3be9a4552f7003c2959268d5
009355319ee1f174e48181c97191d510d008e26d1d19 refs/tags/swift-5.9-DEVELOPMENT-SNAPSHOT-2023-09-29-a peeled:d42e3db1c0f97f4a3be9a4552f7003c2959268d5
0093f26e86de6563debebe51d4569336678198b14dfc refs/tags/swift-5.9-DEVELOPMENT-SNAPSHOT-2023-09-30-a peeled:d42e3db1c0f97f4a3be9a4552f7003c2959268d5
```

The 402KB swift-syntax tag request example from https://github.com/swiftlang/swift-package-manager/issues/9845 reduces to 10KB.

Re-creating the resolve sample based on [apple/container](https://github.com/apple/containerization/blob/main/Package.swift) [bench-ls-refs.swift.zip](https://github.com/user-attachments/files/26460260/bench-ls-refs.swift.zip) :

```
./bench-ls-refs
Tag discovery benchmark (29 repos, 10 cores)
Platform: Version 26.3.1 (a) (Build 25D771280a)

Setup: ensuring bare mirrors exist...
  0 ms (0 if cached)

1. Standard git update path
   git fetch (7 slots):  3925 ms
   manifest read (seq):     1948 ms
   total:                   5873 ms

2. Source archives (parallel prefetch + memoized resolve)
   prefetch (8 slots):      2007 ms
   resolve (cache hits):    0 ms
   total:                   2007 ms
```

Proxy support is added, it is possible to use https://www.mitmproxy.org/ to inspect inside the encrypted requests coming out from SPM.

opsroller:
Nobody notice the obvious CODEX anti-patterns? CODEX is terrible at writing swift and fails miserably at handling basic abstractions, and makes mistakes like manually introducing PROXY support where it isn’t necessary because the OS handles it already.

vsarunas:
> manually introducing PROXY support where it isn’t necessary because the OS handles

Please spend 5 minutes in setting up mitmproxy and try it out for your self on how well macOS handles it.  Hint: when it is set globally in `System Preferences`, but ignoring the [common proxy variables](https://about.gitlab.com/blog/we-need-to-talk-no-proxy/). What about other platforms SPM is used on, like Linux and Windows?

For reference, proxy handling `anti-pattern` also exists in: [container](https://github.com/apple/container), [containerization](https://github.com/apple/containerization), [swift-evolution-metadata-extractor](https://github.com/swiftlang/swift-evolution-metadata-extractor), [swiftly](https://github.com/swiftlang/swiftly). And non-Swift project from Apple like: [foundationdb](https://github.com/apple/foundationdb), [corenet](https://github.com/apple/corenet), [ml-cvnets](https://github.com/apple/ml-cvnets)...

The PR switches from forking `git` command to making direct HTTP requests, existing behaviour needs to be preserved as there is [code handling of the proxy environment variables in git](https://github.com/git/git/blob/v2.50.1/http.c#L1160-L1255).

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

Sources/Basics/GitHubSourceArchiveProvider.swift
vsarunas: Using 6 parallel sessions works as per https://github.com/swiftlang/swift-corelibs-foundation/issues/5445; so I changed to it and changed the URL to be GitHub one instead of the codelab subdomain.

We can increase to 8 again once the URLSession is fixed.

dschaefer2:
This is finally looping around back into my view. As I mentioned earlier I may be asking for some architectural changes. I want to be able to support source archives from any URL using this mechanism so I'll be working through it with that in mind. We don't have to do that now, I just want to make sure we have APIs to make that happen. I should get back to you in the next day or so.

vsarunas:
@plemarquand, can you re-trigger CI please? I had fixed Linux test failure. 

Windows will be another few hours till it completes: https://ci-external.swift.org/job/pr-swiftpm-windows-self-hosted/4614/console

plemarquand:
@swift-ci test

plemarquand:
@swift-ci test windows

Sources/Workspace/Workspace+Dependencies.swift
```diff
@@ -80,6 +81,8 @@ extension Workspace {
         let rootManifestsMinimumToolsVersion = rootManifests.values.map(\.toolsVersion).min() ?? ToolsVersion.current
         let resolvedFileOriginHash = try self.computeResolvedFileOriginHash(root: root)
 
+        await self.prefetchTags(rootManifests: rootManifests, observabilityScope: observabilityScope)
```
dschaefer2, line 84: First of all, this is a huge change with its fingers all over the package loading architecture. That has me concerned it's trying to do too much. I'll start with this one. What tags are you prefetching and why are you doing that here so early in the loader?

dschaefer2: Are you feeding these version tags to the PubGrub algorithm for resolution?

dschaefer2:
My plan for generalized source archive support is to add a new PackageReference.Kind for source archive that has the URL and a checksum for the archive (similar to how we do binary archives). You would actually specify that in the manifest. Once it's downloaded it behaves like a fileSystem Kind.

I was hoping this would be similar. You start with a remoteSourceControl Kind and either an exact version requirement in the dependency, or use the version from the Package.resolved file if it exists and we use the force resolved versions flag and fetch the source archive using the appropriate calculated URL for that version tag. Once extracted it should again behave like a fileSystem Kind.

I was just wondering what hurdles you ran into that needed this to be as complex as it is. I could be missing something.

jakepetroules:
It would be great if the hosts mapping to a particular SourceArchiveProvider implementation could be specified via an external configuration file as well. This way corporate users could, for example, add their GitHub Enterprise domain(s) to a config file that would all map to a GitHubEnterpriseSourceArchiveProvider. IIRC the API surface is different than GitHub.com itself. Similar could be said for Bitbucket vs Bitbucket Datacenter, GitLab, and so on.

Sources/PackageFingerprint/FilePackageFingerprintStorage.swift
```diff
@@ -197,13 +197,16 @@ private enum StorageModel {
...
+                    case sourceArchive
                     case manifest
                     case versionSpecificManifest(toolsVersion: ToolsVersion)
 
                     static func from(contentType: Fingerprint.ContentType) -> ContentType {
                         switch contentType {
                         case .sourceCode:
                             return .sourceCode
+                        case .sourceArchive:
+                            return .sourceArchive
```
vsarunas, line 209: @dschaefer2 , @plemarquand , need some help here.

Fingerprints now for Source Archives don't end up in the same file as from git based, they go to `Library/org.swift.swiftpm/security/fingerprints/swift-collections.json` instead of `Library/org.swift.swiftpm/security/fingerprints/swift-collections-7bbf4661.json` currently.

I tried to extend the fingerprint to have more information saved so to have the url, checksum of the zip file to be globally saved instead of saving this in a build workspace:

```json
{
  "version" : 2,
  "versionFingerprints" : {
    "1.4.1" : {
      "sourceControl" : {
        "sourceArchive" : {
          "checksum" : "e54d90b91f9717ae58ec8bf8320ffcef2ea858811f6e9e5d616ceb4c1b9b1155",
          "contentType" : {
            "sourceArchive" : {

            }
          },
          "downloadURL" : "https://github.com/apple/swift-collections/archive/6675bc0ff86e61436e615df6fc5174e043e57924.zip",
          "fingerprint" : "6675bc0ff86e61436e615df6fc5174e043e57924",
          "origin" : "https://github.com/apple/swift-collections.git"
        }
      }
    }
  }
}
```

but unfortunately this results with older SPM not being able to read the fingerprint:
```
error: 'swift-system': Failed to get source control fingerprint for swift-system remoteSourceControl https://github.com/apple/swift-system.git version 1.6.4 from storage: DecodingError.typeMismatch: expected value of type ContentType. Path: versionFingerprints.`1.6.4`.sourceControl.sourceArchive.contentType. Debug description: Invalid number of keys found, expected one.
error: 'swift-system': DecodingError.typeMismatch: expected value of type ContentType. Path: versionFingerprints.`1.6.4`.sourceControl.sourceArchive.contentType. Debug description: Invalid number of keys found, expected one.
error: 'swift-configuration': Failed to get source control fingerprint for swift-configuration remoteSourceControl https://github.com/apple/swift-configuration.git version 1.2.0 from storage: DecodingError.typeMismatch: expected value of type ContentType. Path: versionFingerprints.`1.2.0`.sourceControl.sourceArchive.contentType. Debug description: Invalid number of keys found, expected one.
```

Options:
1. New fingerprint version?
2. Separate fingerprint file?
3. Avoid changing format?

Sources/Workspace/Workspace+Dependencies.swift
vsarunas: Correct, this is the part of the change that improves `swift package update` 3x times.

It was added separately under https://github.com/swiftlang/swift-package-manager/pull/9870/changes/898dcb6c491b4d92042bab5a051a71c715654590#diff-4aa6515031a746ddd2e4c61529d5484eaef6c4916f6cef9f8b65ba85beed976eR433

It exercises `makeSourceArchiveContainer()` for the known package dependencies (reading Package.resolved, or relying on root level manifest definitions if not present).

This essentially just prefetches the tags, so later PubGrub can make decisions faster instead of having to wait for the network sequentially for each package. When PubGrub later calls getContainer -> versionsAscending(), the tags are already cached - no network round-trip needed.

vsarunas:
> It would be great if the hosts mapping to a particular SourceArchiveProvider implementation could be specified via an external configuration file as well.

@jakepetroules , any suggestions, should it be a sub-command of:
```
swift package config
OVERVIEW: Manipulate configuration of the package

USAGE: swift package config <subcommand>

OPTIONS:
  --version               Show the version.
  -h, -help, --help       Show help information.

SUBCOMMANDS:
  set-mirror              Set a mirror for a dependency.
  unset-mirror            Remove an existing mirror.
  get-mirror              Print mirror configuration for the given package dependency.

  See 'swift help package config <subcommand>' for detailed help.
```

Creating a URL config to host mapping instead of having hardcoded defaults to be able to work with custom self hosted versions? `.swiftpm/configuration/source-archive.json`:
```json 
  {
    "version": 1,
    "providers": {
      "github.me.com": "github",
      "gitlab.me.com": "gitlab",
      "bitbucket.me.com": "bitbucket"
    }
  }
```

https://docs.github.com/en/enterprise-cloud@latest/rest/repos/contents?apiVersion=2026-03-10#download-a-repository-archive-tar API looks to be the same, but I don't have access to this to actually verify.

I'm also wondering the benefit and usages, if there is a self hosted SCM system with a large code base, maybe there would be a Package Registry as well that would be creating a large overlap and reducing benefit of this change?

vsarunas:
> My plan for generalized source archive support is to add a new PackageReference.Kind for source archive that has the URL and a checksum for the archive

Parts of this could be re-used for downloading; but approach similar to binaryTargets is much simpler.

> I was just wondering what hurdles you ran into that needed this to be as complex as it is. I could be missing something.

The architecture is slightly different and several/features/optimizations:

1. Reading Package.swift and version variations without git cloning.  To figure out which versions work together, SPM reads each package's Package.swift and sub-versions. Normally it clones the repo to get this file. In PR fetch just that Package.swift files over HTTP instead. This avoids cloning entirely but needs a whole new container implementation and subsequent global caching not to to do HTTP requests repeated for same packages per different SPM invocations.
2. Need a fallback for packages with sub-modules which are fairly common. Adding shallow clones. Being able to fallback to git if something went wrong. 
3. Tag/SHA discovery, to make the package update fast, needed to remove forking of git commands and parallelise HTTP session reuse to reduce round trips, had to implement simple native git smart protocol v2.
4. Since git command was executed previously, needed to handle keychain, ~/.netrc, GH_TOKEN for authentication and proxy environment as previously this was handled externally transparently.
5. Optimize disk space of .build, try to create soft links instead of full copies of the same source code from a global cache.
6. Try to be generic and have an option for future GitLab, BitBucket, etc if needed.

dschaefer2:
> Parts of this could be re-used for downloading; but approach similar to binaryTargets is much simpler.

The thing I'm mentioning is a replacement for binaryTargets which itself is not a good architecture. Trying to replicate them I don't believe is the right approach. Binary artifacts suffer similar issues of spreading through the Workspace and package resolution architecture. I plan on cleaning that up with a new external packages abstraction that lets us wrap source and binary archives with a package manifest so we can treat them as any other package

> 1. Reading Package.swift and version variations without git cloning.

So you are doing package resolution based on this. All of that should be encapsulated in the Package Container just as it is with the SourceControlPackageContainer. Also there is still hope that package registries are the future for the source control problem. It might be worth looking at that container since it seems to do similar things. I'm fine with a new container for source archives from git providers.

If there are things we need to add to the PackageContainer protocol, we can explore that. Ideally, once the code decides it can override the source control package with a source archive from git provider package, everything just happens through that protocol.

You may also want to break this up into multiple PRs. We're having a hard time reviewing this one and trying to analyze all the impacts it'll have on the rest of the system.

vsarunas:
> You may also want to break this up into multiple PRs

Ok, I can open up tomorrow ~3 support PRs for wiring up the independent/standalone changes and then this PR can be the final largest workspace integration.

vsarunas:
I had opened the following, each independent from each other that can be merged in any order:

1. https://github.com/swiftlang/swift-package-manager/pull/9909
2. https://github.com/swiftlang/swift-package-manager/pull/9910
3. https://github.com/swiftlang/swift-package-manager/pull/9911
4. https://github.com/swiftlang/swift-package-manager/pull/9912

Once these are merged, will be able to rebase this branch PR and potentially split into another smaller PR again.

johnno1962:
What happens if the Swift Package has sub modules which are currently also cloned by SPM. I have a repo that depends on this behaviour.

vsarunas:
> what happens if the Swift Package has sub modules which are currently also cloned by SPM.

If the repo that is imported is like https://github.com/apple/swift-protobuf/blob/main/.gitmodules; the repo and the sub-modules will be git shallow-cloned (protobuf and abseil in this case).

Is your project code public ?

johnno1962:
Excellent, thanks for getting back so quickly. The repo is very public: https://github.com/johnno1962/InjectionNext and has a .gitmodules file. If you're supporting that we should be good.

vsarunas:
This works:
```swift
// swift-tools-version: 5.9
import PackageDescription

let package = Package(
    name: "InjectDemo",
    products: [
        .library(name: "InjectDemo", targets: ["InjectDemo"]),
    ],
    dependencies: [
        .package(url: "https://github.com/johnno1962/InjectionNext", from: "1.5.4"),
    ],
    targets: [
        .target(
            name: "InjectDemo",
            dependencies: [
                .product(name: "InjectionNext", package: "InjectionNext"),
            ]
        ),
    ]
```

```
~/GitHub/swift-package-manager/.build/debug/swift-build  --experimental-source-archive-downloads
Fetching https://github.com/johnno1962/InjectionNext from cache
Fetched https://github.com/johnno1962/InjectionNext from cache (0.71s)
Computing version for https://github.com/johnno1962/InjectionNext
Computed https://github.com/johnno1962/InjectionNext at 1.5.4 (1.53s)
Creating working copy for https://github.com/johnno1962/InjectionNext
Working copy of https://github.com/johnno1962/InjectionNext resolved at 1.5.4
Building for debugging...
[...]
[94/100] InjectDemo 146 / 155
[100/100]
Build complete! (12,87 secs)
```

```
$ du -sh .build/*
180M	.build/arm64-apple-macosx
  0B	.build/artifacts
2.1M	.build/build.db
 11M	.build/checkouts
  0B	.build/debug
 56K	.build/debug.yaml
 56K	.build/plugin-tools.yaml
1.6M	.build/repositories
4.0K	.build/workspace-state.json
```

```
$ find shallow-clones -type d -maxdepth 5
shallow-clones
shallow-clones/github.com
shallow-clones/github.com/johnno1962
shallow-clones/github.com/johnno1962/InjectionNext
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea/App
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea/SwiftRegex5
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea/fishhook
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea/DLKit
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea/Tests
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea/.github
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea/Sources
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea/SwiftTrace
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea/.git
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea/.swiftpm
shallow-clones/github.com/johnno1962/InjectionNext/1.5.4-f257a3ea/InjectionLite
```

The repository is small, so shallow clone does not provide much benefit, 1MB.

johnno1962:
Looks good, it's less important but do the sub modules have the versions they had for the particular tag? e.g. does doing the same thing for, say tag 1.2.0 work and clone the old revisions of the sub modules as they were for that tag?

ShikiSuen:
I'm sharing my concerns about this PR: https://forums.swift.org/t/swift-package-manager-needs-its-own-nuget/85990/3

(Let's shift the discussion to the forum.)

Sources/Workspace/Workspace+Dependencies.swift
dschaefer2: We use memoization for caching. Even the manifests are cached that way. That should be applied here. Keeping this functionality localized to the package container is critical.

dschaefer2:
Thanks @ShikiSuen! I'll reply on the forum post. I appreciate your words there.

Sources/Workspace/Workspace+Dependencies.swift
vsarunas: Correct, this will use existing memoization for caching. I moved this inside to resolveDependencies().

I did some analysis on the timings and requests for the repo in https://github.com/swiftlang/swift-package-manager/issues/9845 (mitmproxy capture and analyse).

HTTP request with the prefetch vs without, PubGrub is essentially waiting for network to proceed and working sequentially:

spm_archive

Current default behaviour is working sequentially for the most part, hence it takes 30+ seconds:
spm_gitfetch

Overall, 10x less bandwidth needed if tags semantic version tags are requested:

spm_summary

It is slightly slower since I reduced from 8 to 6 parallel requests to workaround the issue in  https://github.com/swiftlang/swift-corelibs-foundation/issues/5445

vsarunas: @dschaefer2 , I made this more generic and applied it to the existing git pattern in https://github.com/swiftlang/swift-package-manager/pull/9942

vsarunas:
> Looks good, it's less important but do the sub modules have the versions they had for the particular tag? e.g. does doing the same thing for, say tag 1.2.0 work and clone the old revisions of the sub modules as they were for that tag?

Structurally in this PR, only an additional flag is[ passed to git to perform a shallow clone](https://github.com/swiftlang/swift-package-manager/pull/9910/changes#diff-f1b14ecdb63c95b7b9b9633396c9e256d41ecb5e038eccb1ff890f0d7664b60bR378-R400) instead of a full clone (via `git clone --depth 1 --recurse-submodules --shallow-submodules`): 

The internals of how git resolves that that 1.2.1 should use a particular version of DLKit remains the same: checked out version in the git history has a special commit with version of the submodule to check out:
```
InjectionNext git:(1.2.1) ✗ git ls-tree HEAD DLKit
160000 commit f12cebd9b0a39962b6b3f94b7b4946c903da0027	DLKit
```

johnno1962:
Thanks, I'm sure git will do the right thing. Good luck on this PR, You're doing God's work. I know folks are waiting for it.

Issue #9172, opened 2025-09-22
Output falsely says a package was fetched from cache when it wasn't


jakepetroules:
### Is it reproducible with SwiftPM command-line tools: `swift build`, `swift test`, `swift package` etc?

- [x] Confirmed reproduction steps with SwiftPM CLI. The description text _must_ include reproduction steps with either of command-line SwiftPM commands, `swift build`, `swift test`, `swift package` etc.

### Description

CustomWorkspaceDelegate.didFetchPackage fails to check the 'fromCache' flag of the PackageFetchDetails it receives as a parameter, and therefore unconditionally prints "from cache" in the output message regardless of whether the package was fetched from cache.

Only didFetchPackage has this issue, willFetchPackage is correctly checking the flag.

ChrisWeiss:
We've been fighting with this for a while now. Claude composed the following fix (since this is a work issue, I'm not allowed to edit code directly). If AI-generated contributions are not allowed, please delete this comment:

CommandWorkspaceDelegate.didFetchPackage unconditionally prints "from cache" in its completion message, regardless of whether the package was actually served from the shared cache or fetched from the network.

Line 84:

self.outputHandler("Fetched \(packageLocation ?? package.description) from cache (\(duration.descriptionInSeconds))", .always)
The method receives a result: Result<PackageFetchDetails, Error> parameter that contains accurate fromCache and updatedCache booleans populated by RepositoryManager.fetchAndPopulateCache, but these values are never read. The result is only checked for .success vs .failure on line 69.

Expected behavior:

The message should reflect the actual fetch source, e.g.:

"Fetched <package> from cache (0.32s)" when fromCache == true
"Fetched <package> (0.32s)" when fromCache == false
This is the pattern already used correctly by didDownloadBinaryArtifact (lines 146-167) and didDownloadPrebuilt (lines 195-210) in the same file, both of which unpack their result and branch on fromCache.

Impact:

This makes it impossible to determine from CLI output whether SPM actually used the shared dependency cache or hit the network. This is particularly problematic when debugging cache reliability issues (e.g., shared caches on NFS mounts) since the output always claims cache was used even when it was not.

Reproduction:

Run swift package resolve on a project with dependencies, with the shared cache disabled (--disable-dependency-cache) or with no cached repositories present.
Observe that the output still says "Fetched <package> from cache" for every package.
Suggested fix:

[...]

PR #9872, opened 2026-04-01
Fix didFetchPackage incorrectly reporting package was cached

plemarquand:
### Motivation:

`CustomWorkspaceDelegate.didFetchPackage` wasn't checking the `fetchDetails.fromCache` flag and was always reporting that packages were fetched from the cache even when they weren't.

### Modifications:

Follow the same pattern as `willFetchPackage` and check the `fromCache` flag, and report correctly.

### Result:

Logs report correctly whether the package was cached or not

Issue: #9172


plemarquand:
@swift-ci please test

PR #9878, opened 2026-04-02
Explicitly set SUPPORTED_PLATFORMS on macro targets

owenv:
Previously, we relied on the hostBuildTool product type's OverridingBuildSettings to set SUPPORTED_PLATFORMS=$(HOST_PLATFORM) for macros. This worked for builtin SDKs, but was subtly wrong because the product type properties aren't pushed at the very beginning of platform/SDK resolution. As a result we'd sometimes make decisions based on the SUPPORTED_PLATFORMS=$(AVAILABLE_PLATFORMS) project level setting instead and specialize macros for the destination platform when cross-compiling with a Swift SDK. Set SUPPORTED_PLATFORMS in the target PIF instead, like we do for build plugins.

### DIFF
--- a/Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
+++ b/Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
@@ -966,4 +966,24 @@ struct PIFBuilderTests {
             )
         }
     }
+
+    @Test func macroPackageSupportedPlatforms() async throws {
+        try await withGeneratedPIF(fromFixture: "Macros/MinimalMacroPackage") { pif, observabilitySystem in
+            #expect(observabilitySystem.diagnostics.filter { $0.severity == .error }.isEmpty)
+            let project = try pif.workspace.project(named: "MinimalMacroPackage")
+            let projectPlatforms = try project.buildConfig(named: .debug).settings[.SUPPORTED_PLATFORMS]
+            #expect(projectPlatforms == ["$(AVAILABLE_PLATFORMS)"])
+            let targets = project.underlying.targets
+            for target in targets {
+                let id = target.common.id.value
+                let config = try target.buildConfig(named: .debug)
+                let platforms = config.settings[.SUPPORTED_PLATFORMS]
+                if id == "PACKAGE-TARGET:MacroImpl" {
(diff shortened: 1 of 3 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test

Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
```diff
@@ -966,4 +966,24 @@ struct PIFBuilderTests {
...
+            let project = try pif.workspace.project(named: "MinimalMacroPackage")
+            let projectPlatforms = try project.buildConfig(named: .debug).settings[.SUPPORTED_PLATFORMS]
+            #expect(projectPlatforms == ["$(AVAILABLE_PLATFORMS)"])
+            let targets = project.underlying.targets
+            for target in targets {
+                let id = target.common.id.value
+                let config = try target.buildConfig(named: .debug)
+                let platforms = config.settings[.SUPPORTED_PLATFORMS]
+                if id == "PACKAGE-TARGET:MacroImpl" {
+                    #expect(platforms == ["$(HOST_PLATFORM)"])
```
bkhouri: suggestion: consider including a custom error message that will help determine the failure should the test fail. 
```suggestion
                    #expect(platforms == ["$(HOST_PLATFORM)"], "target \(id) did not have the expected supported platform setting")
```

Tests/SwiftBuildSupportTests/PIFBuilderTests.swift
```diff
@@ -966,4 +966,24 @@ struct PIFBuilderTests {
...
+                } else {
+                    #expect(platforms == nil)
```
bkhouri: suggestion: consider including a custom error message that will help determine the failure should the test fail. 


```suggestion
                    #expect(platforms == nil, "target \(id) has supported platforms set, unexpectedly")
```

owenv:
@swift-ci test

owenv:
@swift-ci test windows

owenv:
@swift-ci test macOS

owenv:
@swift-ci test self hosted

PR #9883, opened 2026-04-02
[CMake] Use a separate ABI name for the SwiftSyntax used in SwiftPM

DougGregor:
SwiftPM can end up linking against shared libraries containing other versions of SwiftSyntax, causing ABI conflicts. Just like we do for the toolchain, ABI-namespace the SwiftSyntax modules built as part of SwiftPM so they don't collide.

### Motivation:

The [Manifest parsing PR](https://github.com/swiftlang/swift-package-manager/pull/9759) is blocked because it introduces the use of SwiftSyntax in a new place within the SwiftPM build. This uncovers some existing issues in the CI where we mix-and-match ABI-incompatible SwiftSyntax versions between SwiftPM, swift-build, and the compiler, causing some crashes when (e.g.) loading dylibs from the toolchain into SwiftPM or swift-build.

### Modifications:

Always build SwiftSyntax specifically for SwiftPM, and do so with the prefix `_PM` so it doesn't collide with anything. This is effectively the same approach we used in the toolchain, but separated out from the toolchain so we can test a new SwiftPM without rebuilding the whole toolchain.

### Result:

We'll now have _PM versions of the SwiftSyntax libraries for SwiftPM.

### DIFF
--- a/BuildSupport/SwiftSyntax/CMakeLists.txt
+++ b/BuildSupport/SwiftSyntax/CMakeLists.txt
@@ -28,15 +28,13 @@ set_property(DIRECTORY "${binary_dir}" PROPERTY EXCLUDE_FROM_ALL TRUE)
 # Install SwiftPM versions of the swift-syntax libraries we use.
 set(SWIFTPM_SWIFT_SYNTAX_MODULES
   _PMSwiftBasicFormat
-  _PMSwiftIfConfig
+  _PMSwiftDiagnostics
+  _PMSwiftIDEUtils
   _PMSwiftParser
   _PMSwiftParserDiagnostics
-  _PMSwiftDiagnostics
-  _PMSwiftSyntax
-  _PMSwiftOperators
-  _PMSwiftSyntaxBuilder
-  _PMSwiftIDEUtils
   _PMSwiftRefactor
+  _PMSwiftSyntax
+  _PMSwiftSyntaxBuilder
 )
 
 set(SWIFT_HOST_LIBRARIES_SUBDIRECTORY "swift/host")


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Linux

owenv:
I think the package-based build is probably going to hit similar issues

DougGregor:
@swift-ci please test

DougGregor:
> I think the package-based build is probably going to hit similar issues

Maybe. The compiler dylibs that get dlopen'd use the `Compiler`-prefixed version. Before this PR, the CMake build of SwiftPM also uses the `Compiler`-prefixed version, which is why we get a collision. The package-based build of SwiftPM isn't using any prefix, so it won't conflict with the compiler dylibs. It's possible there's something *else* being dlopen'd that it does conflict with, but I don't know what that would be.

DougGregor:
@swift-ci please smoke test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:

DougGregor:
@swift-ci please test Windows

DougGregor:

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

DougGregor:
@swift-ci please test Windows

DougGregor:
Got the all clear on the Windows bot that was detecting this failure all along. Just waiting on that one straggler and we'll be good to go

DougGregor:

DougGregor:

DougGregor:
@swift-ci please test Windows

DougGregor:
I swear these two Windows bots never agree

DougGregor:

DougGregor:
@swift-ci please test Windows

DougGregor:

DougGregor:

DougGregor:
https://github.com/swiftlang/swift-build/pull/1281#pullrequestreview-4058894471

DougGregor:
https://github.com/swiftlang/swift-build/pull/1281#pullrequestreview-4058894471

DougGregor:
https://github.com/swiftlang/swift-build/pull/1281

DougGregor:
https://github.com/swiftlang/swift-build/pull/1281

PR #9885, opened 2026-04-03
Adopt session-based buildTargetInfo API

owenv:
Accessing this info using the existing session ensures it uses a core which is initialized with the right resource search paths, avoiding a crash that could occur when running builds on macOS with an Xcode install including the metal toolchain.

### DIFF
--- a/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
+++ b/Sources/SwiftBuildSupport/SwiftBuildSystem.swift
@@ -759,16 +759,16 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
         }
     }
 
-    private func buildTargetInfo(service: SWBBuildService) async throws -> SWBBuildTargetInfo {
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

rconnell9 (approved):

Sources/SwiftBuildSupport/SwiftBuildSystem.swift
```diff
@@ -759,16 +759,16 @@ public final class SwiftBuildSystem: SPMBuildCore.BuildSystem {
         }
     }
 
-    private func buildTargetInfo(service: SWBBuildService) async throws -> SWBBuildTargetInfo {
+    private func buildTargetInfo(session: SWBBuildServiceSession) async throws -> SWBBuildTargetInfo {
```
jakepetroules, line 762: This function should've just been eliminated since it doesn't need to call toolchainDeveloperPathInfo nor does the logic differ based on whether or not isEmbeddedInXcode is true.

PR #9888, opened 2026-04-03
package collection docs

### LINKED ISSUE #8828: Publish API documentation for PackageCollectionsModel
### Description

Used by package collection tooling, the PackageCollectionsModel should have API reference documentation published akin to what's available for [PackageDescription](https://developer.apple.com/documentation/packagedescription)

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

_No response_

_No response_

_No response_

heckj:
API reference content for PackageCollectionsModel, which provides the model classes for Package Collections as supported by Swift Package Manager.

resolves #8828 

### Motivation:

We've been missing the API from published content - it's primarily model focused, but I still wanted to complete it out.

### Modifications:

- Curation (organization) for the public API types within PackageCollectionsModel
- Updates the doc comments to flesh out the abstracts, align with our patterns and styles for documentation.
- Fixes for spelling, grammar, punctuation, and so forth
- Some revisions of passive voice to active voice for abstracts and discussions on symbols

### Result:

Preview with the command `swift package --disable-sandbox preview-documentation --target PackageCollectionsModel`

Screenshot 2026-04-03 at 3 37 06 PM

The curation is heavily embedded under multiple enums with the existing structure, so I leaned into that for organization rather than replicating content at the top level of the catalog. I'm not convinced that's great, but it felt less awkward to read and explore.

### DIFF
--- a/Sources/PackageCollectionsModel/PackageCollectionModel+v1.swift
+++ b/Sources/PackageCollectionsModel/PackageCollectionModel+v1.swift
@@ -26,10 +26,10 @@ extension PackageCollectionModel.V1 {
         /// A description of the package collection.
         public let overview: String?
 
-        /// An array of keywords that the collection is associated with.
+        /// An array of keywords associated with the collection.
         public let keywords: [String]?
 
-        /// An array of package metadata objects
+        /// An array of package metadata objects.
         public let packages: [PackageCollectionModel.V1.Collection.Package]
 
         /// The version of the format to which the collection conforms.
@@ -38,7 +38,7 @@ extension PackageCollectionModel.V1 {
         /// The revision number of this package collection.
         public let revision: Int?
 
-        /// When the package collection was generated.
+        /// The generation date for this package collection.
         public let generatedAt: Date
 
         /// The author of this package collection.
@@ -83,16 +83,16 @@ extension PackageCollectionModel.V1 {
 extension PackageCollectionModel.V1.Collection {
     /// A package entry in a collection.
     public struct Package: Equatable, Codable {
-        /// The URL of the package. Currently only Git repository URLs are supported.
+        /// The URL of the package. The collection format currently supports only Git repository URLs.
         public let url: URL
         
-        /// Package identity for registry (https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/Registry.md#36-package-identification).
+        /// The package identity for a registry.
         public let identity: String?
 
         /// A description of the package.
         public let summary: String?
 
-        /// An array of keywords that the package is associated with.
+        /// An array of keywords associated with the package.
         public let keywords: [String]?
 
         /// An array of version objects representing the most recent and/or relevant releases of the package.
@@ -101,7 +101,7 @@ extension PackageCollectionModel.V1.Collection {
         /// The URL of the package's README.
         public let readmeURL: URL?
 
-        /// The package's current license info
+        /// The package's current license information.
         public let license: PackageCollectionModel.V1.License?
 
         /// Creates a `Package`.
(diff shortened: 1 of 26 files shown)


plemarquand:
@swift-ci test

Sources/PackageCollectionsModel/PackageCollectionModel+v1.swift
```diff
@@ -443,63 +493,71 @@ extension PackageCollectionModel.V1.ProductType: Codable {
...
-    /// Package collection signature and associated metadata
+    /// A cryptographic signature and the certificate metadata used to verify a
+    /// package collection.
     public struct Signature: Equatable, Codable {
-        /// The signature
+        /// The signature.
         public let signature: String
 
-        /// Details about the certificate used to generate the signature
+        /// Details about the certificate that generates the signature.
```
plemarquand: I think the tense was correct here. The information contained in this variable _was_ used to generate the signature, but this instance of `Certificate` wont be used to generate a signature in the future.

Sources/PackageCollectionsModel/PackageCollectionModel+v1.swift
```diff
@@ -78,17 +91,26 @@ extension PackageCollectionModel.V1 {
...
+        /// and may contain a `.git` suffix.
         public let url: URL
-        
-        /// Package identity for registry (https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/Registry.md#36-package-identification).
+
+        /// An optional package identity that overrides the identity derived from the URL.
+        ///
+        /// When `nil`, the package identity is derived from ``url``.
+        /// Set this when the package is published to a registry or when the
+        /// URL-derived identity is not appropriate.
```
plemarquand, line 107: This behaviour happens in the client, not the initializer, so it'd be more accurate to say

```suggestion
        /// When `nil`, consumers should typically derive the identity from ``url``.
        /// Set this when the package is published to a registry or when the
        /// URL-derived identity is not appropriate.
```

Sources/PackageCollectionsModel/PackageCollectionModel+v1.swift
```diff
@@ -218,20 +252,23 @@ extension PackageCollectionModel.V1.Collection.Package {
...
 
 extension PackageCollectionModel.V1 {
+    /// A target within a package, with an optional module name for importable targets.
     public struct Target: Equatable, Codable {
         /// The target name.
         public let name: String
 
-        /// The module name if this target can be imported as a module.
+        /// The module name if you can import this target as a module, or `nil` for
+        /// targets that are not importable (such as test targets or resource targets).
```
plemarquand: nit: test targets are importable, just with `@testable`, so it'd be more accurate to say "The module name if the target can be imported as a module, or `nil` otherwise."

plemarquand (commented):
Thanks @heckj! This cleans up a lot of gaps in the docs and makes things much cleaner. I had a few comments; after those are addressed this looks good to me.

Sources/PackageCollectionsModel/PackageCollectionModel+v1.swift
heckj: We try to keep everything present tense, but since this is all about past actions, I switched it over to "that generated" instead. I want to avoid the phrasing "used to generate" only because it's passive voice that implies an actor that we're not identifying that did the generating, and who or what it was isn't relevant to the content (beyond the details that are in the property, naturally).

Sources/PackageCollectionsModel/PackageCollectionModel+v1.swift
heckj: Dropped "typically", and used the rest

Sources/PackageCollectionsModel/PackageCollectionModel+v1.swift
heckj: simplified down to "The module name if you can import this target as a module; `nil` otherwise."

plemarquand:
@swift-ci please test

plemarquand (approved):

plemarquand:
@swift-ci test windows

heckj:

PR #9895, opened 2026-04-06
Fix path escaping for Swift Build request overrides

owenv:
These overrides run through the build settings parser, which mangles windows paths. For now, convert these to POSIX style separators, similar to the build system test suite. I'm looking into modifying the xcconfig parser to just allow backslashes in these positions instead in the medium term.


owenv:
@swift-ci test

owenv:
@swift-ci test windows

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test self hosted

PR #9904, opened 2026-04-08
Tests: Make testSDKTracking more robust

bkhouri:
The `testSDKTracking` test was failing in some envorinments as the `newSdkPathStr` did not exist.  Update the test to try remove duplicate `//` in the new SDK Root path name.


bkhouri:
@swift-ci test

PR #9905, opened 2026-04-08
Prebuilts for all host targets

dschaefer2:
Add support for prebuilts for all targets that build for host. Remove the previous attempt that used platform constraints on the modules.

Creates a new BinaryModule kind for Prebuilts so they can be explicit in the module graph instead of adding flags to all dependents of the prebuilts. We also remove those `PREBUILT_` flags since they are no longer needed.

A hook is added to the PlatformsCondition to model whether it is satisfied by a platform that supports the prebuilts. Dependents of the prebuilts module mark the dependency as true (i.e. for host) while leaving the dependencies on the source modules as false (i.e. for non-host). This allows the module to depend on both and the build system to decide which one to use to match the prebuilts platform.

For native, this translates into the host destination builds, i.e. host tools, using the prebuilts and the target destination (even if it's the same as the host) to use the corresponding modules from source.

For SwiftBuild, this uses a new extension to the PlatformFilter to exclude the named platform/environment. The prebuilt dependencies have a platform filter of host and the corresponding source module to exclude the host.

Also cleaned up how the prebuilts are passed into the resolution code. The product name to library isn't really required since there is only one prebuilt library for the foreseeable future.


Sources/PackageGraph/ModulesGraph+Loading.swift
```diff
@@ -30,7 +30,7 @@ extension ModulesGraph {
         requiredDependencies: [PackageReference] = [],
         unsafeAllowedPackages: Set<PackageReference> = [],
         binaryArtifacts: [PackageIdentity: [String: BinaryArtifact]],
-        prebuilts: [PackageIdentity: [String: PrebuiltLibrary]], // Product name to library mapping
+        prebuilts: [PackageIdentity: [PrebuiltLibrary]], // Product name to library mapping
```
dschaefer2: That comment should be fixed for the change. I'll fix if we need a respin. This isn't worth losing another day over.

dschaefer2:
@swift-ci please test

dschaefer2:
@swift-ci please test windows

dschaefer2:
I am writing up an Evolution proposal for this. I'll be looking for reviews once that is complete.

PR #9906, opened 2026-04-08
When running multiple test products, report which ones had failures at the end

owenv:
Currently we emit a note at the end of the log if any XCTests failed, since those run before Swift Testing tests. However, this doesn't scale to situations where we have more than one test product. With this change, we instead emit a note like:
```
Note: Some test targets reported failures:                                                                                                                                                      
     - TargetATests (Swift Testing)
     -  TargetBTests (XCTest)
```

Which will hopefully provide a better high level overview

### DIFF
--- a/Fixtures/Miscellaneous/MultipleXCTestSuitesWithFailures/Package.swift
+++ b/Fixtures/Miscellaneous/MultipleXCTestSuitesWithFailures/Package.swift
@@ -1,0 +1,13 @@
+// swift-tools-version: 5.9
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "MultipleXCTestSuitesWithFailures",
+    targets: [
+        .testTarget(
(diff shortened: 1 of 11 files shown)


owenv:
@swift-ci test

stmontgomery (approved):

dschaefer2 (approved):
Like it. This will make it easier for advanced progress tracking we hope to do in the future.

owenv:
@swift-ci test

owenv:
@swift-ci test windows

owenv:
@swift-ci test self hosted

owenv:
@swift-ci test macOS

owenv:
@swift-ci test windows

dabrahams:
This doesn't appear to be in the currently released Xcode.  Any idea when it will be available?  Just did a huge amount of work without noticing failures because of it.

stmontgomery:
> This doesn't appear to be in the currently released Xcode.  Any idea when it will be available?  Just did a huge amount of work without noticing failures because of it.

This should be included in Xcode 27 Beta.

PR #9910, opened 2026-04-09
Source archive download: Git protocol v2 tag discovery and shallow clone support [part 2]

vsarunas:
Git protocol v2 tag discovery and shallow clone support

### Motivation:

Second in a series of PRs adding source archive download support to SwiftPM. https://github.com/swiftlang/swift-package-manager/pull/9870

SPM currently forks a git process per dependency for tag discovery. Each process establishes its own TCP connection and TLS handshake - two network round-trips per repo, repeated sequentially through the dependency graph. For a project with ~30 dependencies this adds up to several seconds of wall time just for tag discovery.

Additionally, when a package has submodules and cannot use a ZIP archive, there is no way to do a shallow clone - SPM always clones the full history.

### Modifications:

- Git HTTP smart protocol v2 implementation (GitProtocolV2.swift) for in-process tag discovery without forking git. Implements pkt-line encoding/decoding and `ls-refs` command per https://git-scm.com/docs/protocol-v2. Uses `ref-prefix` filtering to request only semver-like tags, reducing response size significantly - e.g. swift-syntax tag response drops from 402KB to 10KB by filtering out hundreds of DEVELOPMENT-SNAPSHOT tags.
- ResolvedTag type that pairs a tag name with its peeled commit SHA and parsed semantic version, avoiding repeated parsing downstream.
- Shallow clone and clone-for-editing methods on GitRepositoryProvider for packages that need git (submodules, git-lfs) but not full history.

### Result:

- Tag discovery can be done via HTTP without forking git, enabling connection reuse and parallelism across repos in later PRs
- Packages with submodules can be shallow cloned instead of full cloned
- Benchmark on ~30 dependency project (apple/containerization): tag discovery drops from ~3900ms (sequential git fetch) to ~2000ms (parallel HTTP with session reuse)

### DIFF
--- a/Sources/Basics/GitProtocolV2.swift
+++ b/Sources/Basics/GitProtocolV2.swift
@@ -1,0 +1,184 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+// Git HTTP smart protocol and protocol v2 implementation for tag discovery.
+// https://git-scm.com/docs/http-protocol
+// https://git-scm.com/docs/protocol-v2
+
+import Foundation
+
+import struct TSCUtility.Version
+
+/// A resolved git tag with its name, the commit SHA it points to, and the
+/// parsed semantic version (so callers don't need to re-parse it).
+public struct ResolvedTag: Equatable, Sendable {
+    public let name: String
+    public let commitSHA: String
+    public let version: Version
+
+    public init(name: String, commitSHA: String, version: Version) {
+        self.name = name
+        self.commitSHA = commitSHA
+        self.version = version
+    }
+}
+
+// MARK: - Git HTTP Protocol v2
+
+/// HTTP-based git protocol v2 helpers for tag discovery without forking
+/// a `git` process.
+package enum GitHTTPProtocolV2 {
+    /// Builds the pkt-line encoded request body for a protocol v2 `ls-refs`
+    /// command with ref-prefix hints for semver-like tags.
+    ///
+    /// Uses 20 `ref-prefix` lines (`refs/tags/0` through `refs/tags/9` plus
+    /// `refs/tags/v0` through `refs/tags/v9`) as an optimization hint to
+    /// reduce the response size. Per the spec, `ref-prefix` is advisory —
+    /// servers MAY return refs not matching the prefix, so callers must
+    /// still filter results (which `resolvedTags(from:)` does via
+    /// `Version(tag:)`).
+    package static func makeTagRefsRequestBody(serverCapabilities: [String] = []) -> Data {
+        var body = Data()
+        body.append(PktLine.encode("command=ls-refs\n"))
+        if serverCapabilities.contains(where: { $0.hasPrefix("agent=") }) {
+            body.append(PktLine.encode("agent=SwiftPackageManager/\(SwiftVersion.current.displayString)\n"))
+        }
+        body.append(PktLine.delimiter)
+        body.append(PktLine.encode("peel\n"))
+        for digit in 0...9 {
+            body.append(PktLine.encode("ref-prefix refs/tags/\(digit)\n"))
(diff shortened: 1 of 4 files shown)


vsarunas:
Not to get lost inside the [comments](https://github.com/swiftlang/swift-package-manager/pull/9870#issuecomment-4183313606), duplicating this here as well.

Re-creating the resolve sample based on [apple/containerization](https://github.com/apple/containerization/blob/main/Package.swift) (29 dependencies); full benchmark script: [bench-ls-refs.swift.zip](https://github.com/user-attachments/files/26460260/bench-ls-refs.swift.zip)

## Standard git (current SPM behaviour)

SPM's [RepositoryManager.lookup()](https://github.com/swiftlang/swift-package-manager/blob/release/6.3/Sources/SourceControl/RepositoryManager.swift#L129) dispatches through an AsyncOperationQueue limited to 3/4 * cores concurrent slots. Each lookup forks a git fetch subprocess, which establishes its own TCP connection and TLS handshake - two network round-trips per repo, with no connection reuse between them. After fetching, the resolver reads manifests sequentially: git tag -l to find the latest version, then [git ls-tree](https://github.com/swiftlang/swift-package-manager/blob/release/6.3/Sources/SourceControl/GitRepository.swift#L537-L549) + git cat-file to extract Package.swift from the bare mirror - one process fork per step, per dependency.

## HTTP protocol v2 (this PR)

Uses in-process git [HTTP smart protocol v2](https://git-scm.com/docs/protocol-v2) `ls-refs` request over a shared URLSession. A single session reuses TCP connections and TLS state across all repos (HTTP/2 multiplexing where supported). Tag discovery is dispatched through 6/8 concurrent slots matching the prefetchTagsIfEnabled() path. The `ref-prefix` refs/tags/ filter is sent as a hint to the server, reducing response payloads significantly(swift-syntax drops from 402 KB to 10 KB by filtering out hundreds of DEVELOPMENT-SNAPSHOT tags).
Results are stored in a Memoizer actor, so the subsequent resolver getContainer() -> getTags() calls are instant cache hits (0 ms).

#### Results

```
./bench-ls-refs
Tag discovery benchmark (29 repos, 10 cores)
Platform: Version 26.3.1 (a) (Build 25D771280a)

Setup: ensuring bare mirrors exist...
  0 ms (0 if cached)

1. Standard git update path
   git fetch (7 slots):  3925 ms
   manifest read (seq):     1948 ms
   total:                   5873 ms

2. Source archives (parallel prefetch + memoized resolve)
   prefetch (8 slots):      2007 ms
   resolve (cache hits):    0 ms
   total:                   2007 ms
```

The example script timing from the forum post; fetching in parallel Package.swift files over CDN is fast:
```bash
git clone --mirror https://github.com/apple/swift-nio.git /tmp/nio
cd /tmp/nio
git tag -l '2.*' | sort -V | tail -20 > /tmp/nio-tags.txt
while read t; do git rev-parse "$t^{commit}"; done < /tmp/nio-tags.txt > /tmp/nio-shas.txt

# Git path (~600ms)
time (while read t; do git ls-tree "$t" -- Package.swift > /dev/null; git cat-file -p "$t":Package.swift > /dev/null; done < /tmp/nio-tags.txt)

# HTTP concurrent (~120ms)
time (while read sha; do curl -sL -o /dev/null "https://raw.githubusercontent.com/apple/swift-nio/$sha/Package.swift" & [ $(jobs -r | wc -l) -ge 8 ] && wait -n; done < /tmp/nio-shas.txt; wait)
```

## Why the HTTP path is faster

| Factor | git fetch| HTTP ls-refs |
|---|---|---|
| Connection setup | New TCP + TLS per repo, per git command (29 handshakes * 3) | Shared URLSession - significant connection reuse |
| Process overhead | Fork + exec git per operation (~120 forks: 29 fetch + 29 tag + 29 ls-tree + 29 cat-file) | Zero subprocesses - all in-process HTTP |
| Payload size | Full ref advertisement (swift-syntax 402 KB) | ref-prefix filtered ls-refs only (semver tags) (swift-syntax 10KB) |
| Manifest reading | Sequential git ls-tree + git cat-file per dep (1948 ms) | Not needed at tag discovery phase - 0 ms (memoized) |
| Concurrency | 7 slots (3/4 * cores) | 8 slots (matches prefetchTagsIfEnabled) |

The 3925 ms git fetch time is dominated by per-repo TLS handshakes and process fork overhead - even with pre-warmed bare mirrors, each git fetch re-establishes a connection. The 1948 ms sequential manifest read adds roughly 67 ms per dependency in process spawn and git object lookup. The HTTP path eliminates both bottlenecks by keeping everything in-process with connection reuse.

PR #9925, opened 2026-04-09
Add environment variable-based authentication for registries and HTTP downloads

plemarquand:
Introduce `EnvironmentAuthorizationProvider` and `InMemoryNetrcAuthorizationProvider` to support using environment variables to authenticate with registries and to download binary artifacts from authenticated locations. This should better enable CI/CD workflows without requiring a keychain (macOS only) or .netrc files, which is the only option on for auth on non macOS platforms.

Environment variables:
- `SWIFTPM_REGISTRY_TOKEN`: Bearer token for all registries
- `SWIFTPM_REGISTRY_LOGIN/PASSWORD`: Basic auth for all registries
- `SWIFTPM_SOURCE_CONTROL_TOKEN`: Token for binary artifact/prebuilt downloads
- `SWIFTPM_NETRC_DATA`: Inline netrc content for per-host credentials

Environment variables take highest precedence over existing credential sources. When no explicit auth type is configured in `registries.json`, the `RegistryClient` infers Bearer vs Basic from the credential shape, allowing env var auth to work without running `swift package-registry login` first.

Issue: #9920

### DIFF
--- a/Sources/Workspace/Workspace+Configuration.swift
+++ b/Sources/Workspace/Workspace+Configuration.swift
@@ -309,6 +309,22 @@ extension Workspace.Configuration {
         ) throws -> AuthorizationProvider? {
             var providers = [AuthorizationProvider]()
 
+            let env = Environment.current
+            if let token = env[.swiftpmSourceControlToken], !token.isEmpty {
+                providers.append(EnvironmentAuthorizationProvider(kind: .sourceControl))
+            }
+
+            if let netrcData = env[.swiftpmNetrcData], !netrcData.isEmpty {
+                do {
+                    providers.append(try InMemoryNetrcAuthorizationProvider(content: netrcData))
+                } catch {
+                    observabilityScope.emit(
+                        warning: "Failed to parse SWIFTPM_NETRC_DATA environment variable",
+                        underlyingError: error
+                    )
+                }
+            }
+
             switch self.netrc {
             case .custom(let path):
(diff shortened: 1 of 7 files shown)


plemarquand:
@swift-ci test

plemarquand:
@swift-ci test linux

Sources/Workspace/Workspace+Configuration.swift
```diff
@@ -354,6 +370,33 @@ extension Workspace.Configuration {
             fileSystem: FileSystem,
             observabilityScope: ObservabilityScope
         ) throws -> AuthorizationProvider? {
+            let env = Environment.current
```
owenv, line 373: I wonder if we should consider giving EnvironmentAuthorizationProvider a snapshot of Environment.current, and then remove auth env vars from the process environment to reduce the chance they leak into the build. It's not a perfect defense against credential leaks, but might be valuable in combination with other safeguards

plemarquand: Altering the Environment means  that `makeRegistryAuthorizationProvider` could only ever be called once. 

I think that would work for SwiftPM as its architected to only create one Workspace, but other clients that use SwiftPM as a package may create multiple Workspaces and Workspace.Configurations. In that case only the first configuration would create an EnvironmentAuthorizationProvider.

owenv (approved):

PR #9933, opened 2026-04-12
Extended manifest parsing

DougGregor:
Extends the parsing manifest loader introduced in https://github.com/swiftlang/swift-package-manager/pull/9759 to handle more package manifests and have a more regular internal structure.

### Motivation:

The parsing manifest loader handles declarative package manifests. By teaching it to handle more common patterns in real-world package manifests, it can provide its optimizations for more use cases.

### Modifications:

There are three major changes here:
* Addition of support for global variables in package manifests, so if you pull out some common logic into a global variable and use it in your package manifest, the parsing manifest loader can still handle it.
* Addition of support for appending to package targets/products/dependencies/etc. after the initial package declaration.
* Reworked the implementation to make it more internally consistent and declarative, collapsing if-else chains into switch statements, using error handling to ensure that we record limitations consistently, and generally improving the quality and maintainability of the code.

### Result:

A bunch more package manifests can be handled by the parsing manifest loader over the original PR.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -348,7 +348,16 @@ let package = Package(
                 "Basics",
                 "PackageModel",
                 "SourceControl",
-            ],
+            ] + swiftSyntaxDependencies(
+                [
+                    "SwiftDiagnostics",
+                    "SwiftIfConfig",
+                    "SwiftOperators",
+                    "SwiftParser",
+                    "SwiftParserDiagnostics",
+                    "SwiftSyntax",
+                ]
+            ),
             exclude: ["CMakeLists.txt", "README.md"],
             swiftSettings: commonExperimentalFeatures
         ),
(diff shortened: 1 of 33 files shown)


DougGregor:
@swift-ci please test

0xpablo:
This is great! I had some time to do a bit more testing.
Out of our 210 local packages at Goodnotes, just 9 fell back to the executor path under the old implementation and this work made a couple of them more work in the parsing path so we are down to 7. 
I took a look at the kind of patterns that remain and they are loops or transforming operations so we expect those to remain in the executor path.

In terms of timing, our warm resolution times went from 2:45s (6.2.4 SPM) to 0:30s. We will try to switch our remote dependencies over to a package registry in the coming weeks which we are hoping it will help even further.

Thanks a lot for this work! 👏

dschaefer2:
Thanks Doug! I'll take a look at the manifest loading PRs once I get some progress on my current prebuilts work. Is this something that can wait until after the 6.4 branch?

PR #9934, opened 2026-04-13
swift package experimental-build-server should return compiler args for plugin scripts

owenv:
This is roughly the same as https://github.com/swiftlang/swift-package-manager/pull/9583, but for `swift package experimental-build-server`


owenv:
@swift-ci test

jakepetroules (approved):

PR #9945, opened 2026-04-14
Fix hang in URLSession HTTP client on Linux/Windows when server returns 401

plemarquand:
The underlying networking behaviour of URLSession seems to differ between macOS and Linux/Windows. If a server responds with an authorization challenge (401) and the Linux/Windows client doesn't provide an implementation of `urlSession(_:task:didReceive:challenge:)` then the request hangs forever waiting for the client to answer the auth challenge. On macOS if the client doesn't implement this delegate method then the 401 error is propagated up via the
`urlSession(_:task:didCompleteWithError` delegate method.

Implement the challenge delegate in both `DataTaskManager` and `DownloadTaskManager` to explicitly cancel HTTP auth challenges and deliver the original response, producing the expected badResponseStatusCode(401) error consistent with macOS behavior.

This prevents a client with no authentication credentials saved from hanging forever during a `swift package resolve` when it tries to fetch binary artifacts from a registry that requires authentication.

### DIFF
--- a/Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
+++ b/Sources/Basics/HTTPClient/URLSessionHTTPClient.swift
@@ -184,11 +184,37 @@ private final class DataTaskManager: NSObject, URLSessionDataDelegate {
         }
     }
 
-    public func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
-        guard let task = self.tasks.removeValue(forKey: task.taskIdentifier) else {
+    public func urlSession(
+        _ session: URLSession,
+        task: URLSessionTask,
+        didReceive challenge: URLAuthenticationChallenge,
+        completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void
+    ) {
+        switch challenge.protectionSpace.authenticationMethod {
+        case NSURLAuthenticationMethodHTTPBasic,
+            NSURLAuthenticationMethodHTTPDigest,
+            NSURLAuthenticationMethodNTLM:
+            if var dataTask = self.tasks[task.taskIdentifier] {
+                dataTask.response = challenge.failureResponse as? HTTPURLResponse
+                dataTask.authChallengeCancelled = true
+                self.tasks[task.taskIdentifier] = dataTask
+            }
+            completionHandler(.cancelAuthenticationChallenge, nil)
+        default:
+            completionHandler(.performDefaultHandling, nil)
+        }
+    }
+
+    public func urlSession(
+        _ session: URLSession,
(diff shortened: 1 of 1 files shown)


plemarquand:
@swift-ci test

dschaefer2 (approved):
Nice!

jakepetroules:
I fixed something like this upstream in Foundation: swiftlang/swift-corelibs-foundation#5022

I also see that the methods you implemented have a default implementation already:

https://github.com/swiftlang/swift-corelibs-foundation/blob/0b23e798e921fd14ff59b37c9145021cd60e0aa9/Sources/FoundationNetworking/URLSession/URLSessionDelegate.swift#L148

Before we merge this, can we make sure we fully understand the failure cause and whether this should instead be a fix upstream in Foundation?

Users should never hit a hang by default, so maybe those default implementations are missing something compared to the Darwin implementation.

plemarquand:
@jakepetroules Thanks, this was helpful and pointed me towards the real cause. 

My first attempt in SwiftPM was implementing the delegate methods the same way as you did in swift-corelibs-foundation, simply doing `completionHandler(.performDefaultHandling, nil)`. This still hung, which initially confused me but I now realize makes sense since this was the implementation that was already being used.

Consequently I went digging in swift-corelibs-foundation and I think the root bug tracks back to [attemptProceedingWithDefaultCredential](https://github.com/swiftlang/swift-corelibs-foundation/blob/0b23e798e921fd14ff59b37c9145021cd60e0aa9/Sources/FoundationNetworking/URLSession/URLSessionTask.swift#L1270) in URLSessionTask.swift. 

```swift
        @Sendable func attemptProceedingWithDefaultCredential() {
            if let credential = challenge.proposedCredential {
                let last = task._protocolLock.performLocked { task._lastCredentialUsedFromStorageDuringAuthentication }
                
                if last?.credential != credential {
                    proceed(using: credential)
                } else {
                    task.cancel()
                }
            }
        }
```

This method is called after the delegate method does `completionHandler(.performDefaultHandling, nil)`. In the failure case I have no credentials saved; I'm making the request expecting a 401. `credential` evaluates to `nil`, and then nothing happens. The task isn't ever cancelled or completed. In the `nil` case I think we just need to call `task.cancel()` to unblock the request.

I'm going to drive this PR upstream and come back and close out this one.

plemarquand:
Potentially related forum thread: https://forums.swift.org/t/urlsession-works-on-macos-fail-on-linux/35903/

plemarquand:
Closing this in favour of https://github.com/swiftlang/swift-corelibs-foundation/pull/5457

jakepetroules:
Awesome work, @plemarquand. Thanks for taking the time to track down the root cause!

